PR #6056, opened 2025-07-01
fix: preserve transaction type of batched transactions

micaelae:
## Checklist

### DIFF
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Preserve type of nested transactions in batch if specified in `TransactionBatchSingleRequest` ([]())
+
 ## [58.1.0]
 
 ### Added
--- a/packages/transaction-controller/src/utils/batch.test.ts
+++ b/packages/transaction-controller/src/utils/batch.test.ts
@@ -834,6 +834,110 @@ describe('Batch Utils', () => {
         });
       });
 
+      it('calls publish batch hook with requested transaction type', async () => {
+        const publishBatchHook: jest.MockedFn<PublishBatchHook> = jest.fn();
+
+        addTransactionMock
+          .mockResolvedValueOnce({
+            transactionMeta: {
+              ...TRANSACTION_META_MOCK,
+              id: TRANSACTION_ID_MOCK,
+            },
+            result: Promise.resolve(''),
+          })
+          .mockResolvedValueOnce({
+            transactionMeta: {
+              ...TRANSACTION_META_MOCK,
+              id: TRANSACTION_ID_2_MOCK,
+            },
+            result: Promise.resolve(''),
+          });
+
+        publishBatchHook.mockResolvedValue({
+          results: [
+            {
+              transactionHash: TRANSACTION_HASH_MOCK,
+            },
+            {
+              transactionHash: TRANSACTION_HASH_2_MOCK,
+            },
+          ],
+        });
+
+        addTransactionBatch({
+          ...request,
+          publishBatchHook,
+          request: {
+            ...request.request,
+            transactions: [
+              {
+                ...request.request.transactions[0],
+                type: TransactionType.swap,
+              },
+              {
+                ...request.request.transactions[1],
+                type: TransactionType.bridge,
+              },
+            ],
+            disable7702: true,
+          },
+        }).catch(() => {
+          // Intentionally empty
+        });
+
+        await flushPromises();
+
+        const publishHooks = addTransactionMock.mock.calls.map(
+          ([, options]) => options.publishHook,
+        );
+
+        publishHooks[0]?.(
+          TRANSACTION_META_MOCK,
+          TRANSACTION_SIGNATURE_MOCK,
+        ).catch(() => {
+          // Intentionally empty
+        });
+
+        publishHooks[1]?.(
+          TRANSACTION_META_MOCK,
+          TRANSACTION_SIGNATURE_2_MOCK,
+        ).catch(() => {
+          // Intentionally empty
+        });
+
+        await flushPromises();
+
+        expect(publishBatchHook).toHaveBeenCalledTimes(1);
+        expect(publishBatchHook).toHaveBeenCalledWith({
+          from: FROM_MOCK,
+          networkClientId: NETWORK_CLIENT_ID_MOCK,
+          transactions: [
+            {
+              id: TRANSACTION_ID_MOCK,
+              params: {
+                ...TRANSACTION_BATCH_PARAMS_MOCK,
+                gas: undefined,
+                maxFeePerGas: undefined,
+                maxPriorityFeePerGas: undefined,
+              },
+              signedTx: TRANSACTION_SIGNATURE_MOCK,
+              type: TransactionType.swap,
+            },
+            {
+              id: TRANSACTION_ID_2_MOCK,
+              params: {
+                ...TRANSACTION_BATCH_PARAMS_MOCK,
+                gas: undefined,
+                maxFeePerGas: undefined,
+                maxPriorityFeePerGas: undefined,
+              },
+              signedTx: TRANSACTION_SIGNATURE_2_MOCK,
+              type: TransactionType.bridge,
+            },
+          ],
+        });
+      });
+
       it('resolves individual publish hooks with transaction hashes from publish batch hook', async () => {
         const publishBatchHook: jest.MockedFn<PublishBatchHook> = jest.fn();
 
--- a/packages/transaction-controller/src/utils/batch.ts
+++ b/packages/transaction-controller/src/utils/batch.ts
@@ -246,7 +246,7 @@ async function getNestedTransactionMeta(
   ethQuery: EthQuery,
 ): Promise<NestedTransactionMetadata> {
   const { from } = request;
-  const { params } = singleRequest;
+  const { params, type: requestType } = singleRequest;
 
   const { type } = await determineTransactionType(
     { from, ...params },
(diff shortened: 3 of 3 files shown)


packages/transaction-controller/src/utils/batch.ts
```diff
@@ -595,6 +595,7 @@ async function processTransactionWithHook(
   const { transactionMeta } = await addTransaction(
     transactionMetaForGasEstimates.txParams,
     {
+      type,
```
micaelae: Does line 603 override the `requireApproval` value passed to `addTransactionBatch`? We conditionally set this to true for mobile clients that use hardware wallets so need to preserve the requested value of this as well

bfullam (approved):

bfullam:
@metamaskbot publish-preview

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -255,7 +255,7 @@ async function getNestedTransactionMeta(
 
   return {
     ...params,
-    type,
+    type: requestType ?? type,
```
matthewwalsh0: Minor, can we assign this to a variable rather than do it inline?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -246,7 +246,7 @@ async function getNestedTransactionMeta(
   ethQuery: EthQuery,
 ): Promise<NestedTransactionMetadata> {
   const { from } = request;
-  const { params } = singleRequest;
+  const { params, type: requestType } = singleRequest;
 
   const { type } = await determineTransactionType(
```
matthewwalsh0: Minor, could we assign this to `determinedType` for clarity?

packages/transaction-controller/src/utils/batch.ts
matthewwalsh0: To clarify, this change will only influence 7702 transactions which will still have their own top-level `batch` type as it would be a single transaction.

So the individual call types would have to be checked inside the `nestedTransactions` property.

packages/transaction-controller/src/utils/batch.ts
matthewwalsh0: Minor, alphabetical.

matthewwalsh0: When using `addTransactionBatch`, the intent is to show a single batch confirmation if approval is needed, though this is not yet implemented in extension so currently only supports no approvals.

The `requireApproval` on like 603 suppresses further approvals for each individual transaction / call as we already have approval.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -620,11 +621,16 @@ async function processTransactionWithHook(
     value,
   };
 
-  log('Processed new transaction with hook', { id, params: newParams });
+  log('Processed new transaction with hook', {
+    id,
+    params: newParams,
+    type: nestedTransaction.type,
```
matthewwalsh0: Minor, can we use `type` directly?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -620,11 +621,16 @@ async function processTransactionWithHook(
...
+  });
 
   return {
     id,
     params: newParams,
+    type: nestedTransaction.type,
```
matthewwalsh0: Minor, also here?

packages/transaction-controller/src/utils/batch.test.ts
```diff
@@ -834,6 +834,110 @@ describe('Batch Utils', () => {
...
+          TRANSACTION_META_MOCK,
+          TRANSACTION_SIGNATURE_2_MOCK,
+        ).catch(() => {
+          // Intentionally empty
+        });
+
+        await flushPromises();
+
+        expect(publishBatchHook).toHaveBeenCalledTimes(1);
+        expect(publishBatchHook).toHaveBeenCalledWith({
```
matthewwalsh0: Can we remove the duplicate property assertions here and check only the `type` as that's the scope of the test?

packages/transaction-controller/src/utils/batch.test.ts
```diff
@@ -834,6 +834,110 @@ describe('Batch Utils', () => {
         });
       });
 
+      it('calls publish batch hook with requested transaction type', async () => {
```
matthewwalsh0: Can we add an alternate test for the 7702 flow to ensure the request types are retained in the `nestedTransactions` property?

packages/transaction-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Preserve type of nested transactions in batch if specified in `TransactionBatchSingleRequest` ([#6056](https://github.com/MetaMask/core/pull/6056))
```
matthewwalsh0:
```suggestion
- Preserve provided `type` in `transactions` when calling `addTransactionBatch` ([#6056](https://github.com/MetaMask/core/pull/6056))
```

matthewwalsh0 (changes requested):

packages/transaction-controller/src/utils/batch.test.ts
micaelae: Good call - removing these duplicates

packages/transaction-controller/src/utils/batch.ts
micaelae: That makes sense! We check the individual call types in our controller and in the clients so the top-level batch type + individual call types for nested transactions works

packages/transaction-controller/src/utils/batch.ts
micaelae: Will update this

packages/transaction-controller/src/utils/batch.ts
micaelae: Oh that makes sense! We want just the single approval for the whole batch. Sounds like that's already the case so we don't need any changes here

vinistevam (dismissed):

packages/transaction-controller/src/types.ts
```diff
@@ -1722,6 +1722,9 @@ export type PublishBatchHookTransaction = {
 
   /** Signed transaction data to publish. */
   signedTx: Hex;
+
+  /** Type of the nested transaction. */
+  type?: TransactionType;
```
matthewwalsh0: Why do we need the `type` here?

The intent of providing the `id` is so the hook can require any specific transaction metadata it needs via the transaction controller state and that reference.

packages/transaction-controller/src/utils/batch.test.ts
```diff
@@ -838,6 +884,89 @@ describe('Batch Utils', () => {
...
+          // Intentionally empty
+        });
+
+        await flushPromises();
+
+        expect(publishBatchHook).toHaveBeenCalledTimes(1);
+        expect(publishBatchHook.mock.calls[0][0].transactions[0].type).toBe(
+          TransactionType.swap,
+        );
+        expect(publishBatchHook.mock.calls[0][0].transactions[1].type).toBe(
```
matthewwalsh0: Rather than asserting the hook is called with the type, ideally the hook wouldn't know, and instead we could verify that `addTransaction` is called multiple times with the correct types?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/types.ts
micaelae: Will remove this!

packages/transaction-controller/src/utils/batch.test.ts
micaelae: Good call - I updated the test to do this! I also removed the `type` key from  the `PublishBatchHookTransaction` type, which I added for this test.

packages/transaction-controller/src/utils/batch.test.ts
```diff
@@ -837,6 +932,50 @@ describe('Batch Utils', () => {
           PUBLISH_BATCH_HOOK_PARAMS,
         );
       });
+      //   const publishBatchHook: jest.MockedFn<PublishBatchHook> = jest.fn();
```
matthewwalsh0: Accident?

micaelae: not sure how this happened - will update

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -606,6 +607,7 @@ async function processTransactionWithHook(
       networkClientId,
       publishHook,
       requireApproval: false,
+      type,
```
OGPoyraz, line 610: Hey @micaelae , I was debugging the metric issues due to swaps and bridge transaction events and `origin` was also not preserving to the individual batched transactions. If you don't mind we could add `origin` here too. Otherwise source will be `dapp` in metrics which is incorrect.

OGPoyraz, line 610: On another note if we are correctly preserving the type with this fix now - do we want to remove this temporary type update from `bridge-status-controller`? https://github.com/MetaMask/core/blob/main/packages/bridge-status-controller/src/utils/transaction.ts#L355

matthewwalsh0: Given the age of this PR, are we okay to do that in a separate PR @OGPoyraz ?

matthewwalsh0 (approved):

packages/transaction-controller/src/utils/batch.ts
micaelae: Yes! I have a draft for removing this, will just need to re-test once this version is published

micaelae: The PR was merged before I could add this. Can you open a new PR for this change?

OGPoyraz: All good, we will do it in a separate fix!

PR #6060, opened 2025-07-02
feat: persist access token and metadata access token in seedless controller

himanshuchawla009:
## Explanation

- This update introduces two new tokens, accessToken and metadataAccessToken, to the SeedlessOnboardingController. 

- These tokens are essential for pairing with the profile sync auth service and accessing the metadata service before the vault is created or unlocked.

-  The changes include updates to the controller state, methods for handling these tokens, and corresponding tests to ensure proper functionality and error handling when tokens are missing. 

- Additionally, the vault data structure has been modified to include the new accessToken. Tests have been added to verify the correct behavior of the controller with respect to these tokens.

## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
@@ -3,7 +3,6 @@ import type { StateMetadata } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
 import type {
   KeyPair,
-  NodeAuthTokens,
   RecoverEncryptionKeyResult,
   SEC1EncodedPublicKey,
 } from '@metamask/toprf-secure-backup';
@@ -123,6 +122,17 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
       persist: false,
       anonymous: true,
     },
+    // stays in vault
+    accessToken: {
+      persist: false,
+      anonymous: true,
+    },
+    // stays outside of vault as this token is accessed by the metadata service
+    // before the vault is created or unlocked.
+    metadataAccessToken: {
+      persist: false,
+      anonymous: true,
+    },
     encryptedSeedlessEncryptionKey: {
       persist: true,
       anonymous: true,
@@ -219,6 +229,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
    * @param params.socialLoginEmail - The user email from Social login.
    * @param params.refreshToken - refresh token for refreshing expired nodeAuthTokens.
    * @param params.revokeToken - revoke token for revoking refresh token and get new refresh token and new revoke token.
+   * @param params.accessToken - access token for pairing with profile sync auth service and to access other services.
+   * @param params.metadataAccessToken - metadata access token for accessing the metadata service before the vault is created or unlocked.
    * @param params.skipLock - Optional flag to skip acquiring the controller lock. (to prevent deadlock in case the caller already acquired the lock)
    * @returns A promise that resolves to the authentication result.
    */
@@ -231,6 +243,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
     socialLoginEmail?: string;
     refreshToken?: string;
     revokeToken?: string;
+    accessToken?: string;
+    metadataAccessToken?: string;
     skipLock?: boolean;
   }) {
     const doAuthenticateWithNodes = async () => {
@@ -244,6 +258,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
           socialLoginEmail,
           refreshToken,
           revokeToken,
+          accessToken,
+          metadataAccessToken,
         } = params;
 
         const authenticationResult = await this.toprfClient.authenticate({
@@ -266,6 +282,12 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
           if (revokeToken) {
             // Temporarily store revoke token in state for later vault creation
             state.revokeToken = revokeToken;
+          }
+          if (accessToken) {
+            state.accessToken = accessToken;
+          }
+          if (metadataAccessToken) {
+            state.metadataAccessToken = metadataAccessToken;
           }
         });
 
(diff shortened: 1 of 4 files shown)


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -123,6 +122,17 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
...
+    accessToken: {
+      persist: false,
+      anonymous: true,
+    },
+    // stays outside of vault as this token is accessed by the metadata service
+    // before the vault is created or unlocked.
+    metadataAccessToken: {
+      persist: false,
+      anonymous: true,
+    },
```
matthiasgeihs: is it correct that this item should **not** be persisted even though it lives outside the vault?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -231,6 +243,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
     socialLoginEmail?: string;
     refreshToken?: string;
     revokeToken?: string;
+    accessToken?: string;
+    metadataAccessToken?: string;
```
matthiasgeihs: why is it optional here but then in many/most/all? places we require this to be defined?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -219,6 +229,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
    * @param params.socialLoginEmail - The user email from Social login.
    * @param params.refreshToken - refresh token for refreshing expired nodeAuthTokens.
    * @param params.revokeToken - revoke token for revoking refresh token and get new refresh token and new revoke token.
+   * @param params.accessToken - access token for pairing with profile sync auth service and to access other services.
+   * @param params.metadataAccessToken - metadata access token for accessing the metadata service before the vault is created or unlocked.
```
matthiasgeihs:
```suggestion
   * @param params.accessToken - Access token for pairing with profile sync auth service and to access other services.
   * @param params.metadataAccessToken - Metadata access token for accessing the metadata service before the vault is created or unlocked.
```

matthiasgeihs (commented):
Access token and Metadata access token are being added but not used anywhere.
I'm assuming they will be used in a future version of the toprf client.
I think it would be more logical if we first update the client and then add the new tokens.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: I also wondered the same but this is not something i added in this PR, same is being done for other tokens because authenticate function is used at two different places and at one of the place these tokens are not required to be passed.  Not the best way to do it, we should have defined separate functions for that.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: fixed [62be127](https://github.com/MetaMask/core/pull/6060/commits/62be127e9423e6150b9877cd6923ff0d45ab4d70)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: fixed here: [62be127](https://github.com/MetaMask/core/pull/6060/commits/62be127e9423e6150b9877cd6923ff0d45ab4d70)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: fixed here [62be127](https://github.com/MetaMask/core/pull/6060/commits/62be127e9423e6150b9877cd6923ff0d45ab4d70)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: this is a breaking change then.
how would this change be reflected in the frontend code? where would the new tokens be retrieved from?

it is usually required to create a draft PR in at least one of the frontend repositories (extension/mobile) to ensure a proper integration is possible.

matthiasgeihs:
> > Access token and Metadata access token are being added but not used anywhere. I'm assuming they will be used in a future version of the toprf client. I think the purpose of this PR would be clearer if we first update the client and then add the new tokens. (This way we can make sure that the integration is right.)
> 
> Yes that's correct, access_token will be used by profile sync team in auth controller. Metadata access token will be used by sdk.

I think a draft PR in extension would help to understand how this change is integrated.

I am thinking that we should release it as a separate version to unblock the merging of the other seedless onboarding PRs that rely on controller v2 and the store keyring key functionality.

mirceanis (approved):
This looks good to me.
This PR is a prerequisite for the backup & sync pairing.
But @matthiasgeihs is right. For proper integration testing we do also require equivalent PRs in the clients

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: Nothing will break even if you don't pass these tokens, so its technically not a breaking change.

himanshuchawla009:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: if you add a required parameter to an existing API, that is a breaking change!?

lwin-kyaw:
@metamaskbot publish-preview

himanshuchawla009:
> ### Bug: Token Handling and Error Management Issues
> ### Bug: Vault Data Incompatibility Issue
> ### Bug: Inconsistent Token Validation Causes Errors
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

1. This is not issue , as token refresh token api always returns the access and metadata but the way tokens being handled currently are not optimal. We need another PR to fix the optional usage of tokens.

2. Again not a issue, validation is only done on non vault items in this function and access tokens stays in vault.

3. Same comments as 1, we need refactor in the way tokens are being handled in state to reduce the conflicting cases which is not the purpose of this PR.

PR #6063, opened 2025-07-02
Add fallback to sequential hook when `publishBatchHook` returns empty

vinistevam:
## Explanation
This PR aims to add a fallback mechanism to use the sequential hook when `publishBatchHook` returns empty. This will help us to handle case when `publishBatchHook` is defined by the client as it's for smart transaction but smart transaction is disabled.

## References

Fixes https://github.com/MetaMask/MetaMask-planning/issues/5301

# Changelog

## Checklist

### DIFF
--- a/packages/transaction-controller/src/types.ts
+++ b/packages/transaction-controller/src/types.ts
@@ -1742,7 +1742,7 @@ export type PublishBatchHookRequest = {
 export type PublishBatchHookResult =
   | {
       /** Result data for each transaction in the batch. */
-      results: {
+      results?: {
         /** Hash of the transaction on the network. */
         transactionHash: Hex;
       }[];
--- a/packages/transaction-controller/src/utils/batch.ts
+++ b/packages/transaction-controller/src/utils/batch.ts
@@ -488,17 +488,22 @@ async function addTransactionBatchWithHook(
       signedTx: signedTransactions[index],
     }));
 
-    log('Calling publish batch hook', { from, networkClientId, transactions });
+    const hookParams = { from, networkClientId, transactions };
 
-    const result = await publishBatchHook({
-      from,
-      networkClientId,
-      transactions,
-    });
+    log('Calling publish batch hook', hookParams);
+
+    let result = await publishBatchHook(hookParams);
 
     log('Publish batch hook result', result);
+    console.log('Publish batch hook result', result);
 
-    if (!result) {
+    if (result && !Array.isArray(result?.results)) {
+      log('Fallback to sequential publish batch hook due to empty results');
+      const sequentialBatchHook = sequentialPublishBatchHook.getHook();
+      result = await sequentialBatchHook(hookParams);
+    }
+
+    if (!result || !Array.isArray(result.results)) {
       throw new Error('Publish batch hook did not return a result');
     }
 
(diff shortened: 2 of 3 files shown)


vinistevam:
@metamaskbot publish-preview

vinistevam:
Draft PR extension https://github.com/MetaMask/metamask-extension/pull/34019

OGPoyraz (dismissed):

packages/transaction-controller/src/types.ts
```diff
@@ -1742,7 +1742,7 @@ export type PublishBatchHookRequest = {
 export type PublishBatchHookResult =
   | {
       /** Result data for each transaction in the batch. */
-      results: {
+      results?: {
```
matthewwalsh0: Should we instead be explicit and just check for an `undefined` top level result, but insist the `results` should be populated to make the data more consistent?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -488,17 +488,21 @@ async function addTransactionBatchWithHook(
...
-      transactions,
-    });
+    log('Calling publish batch hook', hookParams);
+
+    let result = await publishBatchHook(hookParams);
 
     log('Publish batch hook result', result);
 
-    if (!result) {
+    if (result && !Array.isArray(result?.results)) {
```
matthewwalsh0: Could this just be `!result` to make it explicit?

matthewwalsh0: Do we need to check this hook wasn't already the sequential hook so we don't try it twice?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -488,17 +488,21 @@ async function addTransactionBatchWithHook(
...
+      log('Fallback to sequential publish batch hook due to empty results');
+      const sequentialBatchHook = sequentialPublishBatchHook.getHook();
+      result = await sequentialBatchHook(hookParams);
+    }
+
+    if (!result || !Array.isArray(result.results)) {
```
matthewwalsh0: Minor, could we simplify to `!result?.results?.length`?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -488,17 +488,21 @@ async function addTransactionBatchWithHook(
       signedTx: signedTransactions[index],
     }));
 
-    log('Calling publish batch hook', { from, networkClientId, transactions });
+    const hookParams = { from, networkClientId, transactions };
```
matthewwalsh0, line 493: On line 438 above, should we also check that `requestPublishBatchHook` was provided?

vinistevam: Yes, I was sure it was like this before we changed the fallback sequence. In any case, added it into the check on line 438.

packages/transaction-controller/src/utils/batch.ts
vinistevam: Included a check to verify if  `publishBatchHook` is not sequential.

vinistevam: Applied suggestion.

packages/transaction-controller/src/types.ts
vinistevam: Sure, pushed changes.

PR #6066, opened 2025-07-03
feat: add rate limiting support to controllers leveraging the accounts API

mathieuartu:
## Explanation

@MetaMask/identity recently deployed a `profileId` based (with IP fallback) rate limiting service for the Accounts API.
This PR adds the correct Authentication headers to everywhere in the code the Accounts API was called.

It modifies three controllers, and adds `@metamask/profile-sync-controller` to their respective `package.json` file as a `peerDependency`.

The three controllers are: 
- `TokenDetectionController`
- `MultichainNetworkController`
- `TransactionController`

Those are **breaking** changes, since we need clients to make changes in order to accommodate for this new logic. 
An extension test-drive PR can be found below, and act as a reference on how to adapt client code for those breaking changes:
- https://github.com/MetaMask/metamask-extension/pull/34064

## Checklist

### DIFF
--- a/packages/assets-controllers/src/TokenDetectionController.test.ts
+++ b/packages/assets-controllers/src/TokenDetectionController.test.ts
@@ -2616,11 +2616,7 @@ describe('TokenDetectionController', () => {
             category: 'Wallet',
             properties: {
               tokens: [`${sampleTokenA.symbol} - ${sampleTokenA.address}`],
-              // TODO: Either fix this lint violation or explain why it's necessary to ignore.
-
               token_standard: 'ERC20',
-              // TODO: Either fix this lint violation or explain why it's necessary to ignore.
-
               asset_type: 'TOKEN',
             },
           });


packages/assets-controllers/src/multi-chain-accounts-service/multi-chain-accounts.ts
```diff
@@ -25,33 +26,56 @@ const getBalancesUrl = (
...
 ) {
   const url = getBalancesUrl(address, {
     networks: options?.networks?.join(),
   });
+  const authenticationControllerBearerToken =
+    await options.getAuthenticationControllerBearerToken();
   const response: GetBalancesResponse = await handleFetch(url, {
     headers: {
       'x-metamask-clientproduct': `metamask-${platform}`,
+      Authorization: `Bearer ${authenticationControllerBearerToken}`,
```
Prithpal-Sooriya, line 78: Very interesting!

Open discussion - could we pass the authorization header as a request/response cookie? All `.metamask.io` APIs will receive this cookie and can choose to use it?

It also prevents us from updating all the areas in the codebase?

Prithpal-Sooriya: This is not a blocker but an open discussion.

fabiobozzo (approved):
LGTM

packages/assets-controllers/src/multi-chain-accounts-service/multi-chain-accounts.ts
mathieuartu: Very valid! As discussed on Slack, we cannot implement this ATM, for many reasons (Hydra, PKCE, parallel authentication sessions...). But this sparked very interesting discussions for the future!

Prithpal-Sooriya: Yep great discussion, we can consider something once we support more authenticated APIs.

Prithpal-Sooriya (approved):

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -1041,6 +1081,14 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
     );
     return account?.address || '';
   }
+
+  async #getAuthenticationControllerBearerToken(): ReturnType<
+    AuthenticationControllerGetBearerToken['handler']
+  > {
+    return await this.messagingSystem.call(
+      'AuthenticationController:getBearerToken',
```
matthewwalsh0, line 1089: If all these controllers ultimately call the same action / method, why can't this just be the default where needed in a central accounts API service to avoid all these breaking changes and new dependencies?

Maybe in `@metamask/utils` or `@metamask/controller-utils` to avoid a new dependency?

Or even a util such as `fetchWithBearerToken`?

Or could the clients just implement a proxy to inject this token automatically to avoid any core changes at all and decouple all this from the controllers?

matthewwalsh0 (changes requested):

packages/assets-controllers/src/TokenDetectionController.ts
Prithpal-Sooriya: Great callout. Do we only use the Accounts API in the accounts controller?

matthewwalsh0: I think multiple controllers have their own partial abstractions over the accounts API.

My fear is if we keep casually adding peer dependencies between the controllers, we make the release process harder, breaking changes more common, and eventually we'll just have one interdependent bundle meaning no single controller can have a major release without it cascading through the others.

mathieuartu: @matthewwalsh0 I'm not sure I understand your comment correctly 😄 AFAIK, `AccountsController` does not make Accounts API requests, but each of those controllers directly do in their own "bubble". They all have their own logic and wrappers around `fetch` and also often attach their own headers to these requests. This is why we added this logic at this level.

LMK if that clarifies it - or if I completely missed your point!

matthewwalsh0: Apologies, I initially mentioned the AccountsController by mistake, but updated it to clarify how we could abstract this complexity in various ways.

Prithpal-Sooriya: 100% agreed, controller dependencies feel a bit "tight", the Assets team are already feeling this by how frequently our controllers get upgraded from peer deps 😅 

The cookie approach is a clean abstraction, however there are flaws discussed from this.
1. Maybe we can instead use method or constructor parameters - pass in the bearer token than relying on the messaging system?
2. Create a utility to inject the bearer token in fetch calls? Hmm, this may be tricky as the bearer token lives inside a controller/state and will need to be provided to a util... happy to brainstorm.

ccharly: I guess @matthewwalsh0's point is that, if we need to always pass down that bearer token, it would be great that we would have a "centric way" (like a service maybe?) to just call the endpoints we need without taking care of the token ourselves?

Hence the `fetchWithBearerToken`. We would not need to "get" the token and pass it down, we would just "fetch" the endpoint and get the result we need.

Hope I got this right 😄 and I do agree with @matthewwalsh0 on this, it would be better that way IMO.

matthewwalsh0: Exactly, right now it's a lot of duplication and new dependencies on three new controllers, to ultimately do the same logic. So a central mechanism would be great, either client proxy, util, or shared service.

packages/assets-controllers/src/TokenDetectionController.test.ts
```diff
@@ -2616,10 +2617,10 @@ describe('TokenDetectionController', () => {
             properties: {
               tokens: [`${sampleTokenA.symbol} - ${sampleTokenA.address}`],
               // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.test.ts
```diff
@@ -2616,10 +2617,10 @@ describe('TokenDetectionController', () => {
...
-              // eslint-disable-next-line @typescript-eslint/naming-convention
+
               token_standard: 'ERC20',
               // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -879,10 +918,10 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
             properties: {
               tokens: eventTokensDetails,
               // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire, line 920: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -879,10 +918,10 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
...
-              // eslint-disable-next-line @typescript-eslint/naming-convention
+
               token_standard: ERC20,
               // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire, line 923: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -1009,10 +1049,10 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
           properties: {
             tokens: eventTokensDetails,
             // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire, line 1051: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -1009,10 +1049,10 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
...
-            // eslint-disable-next-line @typescript-eslint/naming-convention
+
             token_standard: ERC20,
             // TODO: Either fix this lint violation or explain why it's necessary to ignore.
```
mcmire, line 1054: Should we remove this comment too?

``` suggestion
```

packages/assets-controllers/src/TokenDetectionController.ts
mathieuartu: Thanks everyone for your input!
I think I see the pattern here, and I agree. Nobody wants a new peerDep. 

It's just unfortunate that I'll need to juggle around 3 different fetch implementations 😅 
- `handleFetch` for `TokenDetectionController`
- `successfulFetch` for `TransactionController`
- `fetch: window.fetch.bind(window)` for `MultichainNetworkController` (as seen on [extension](https://github.com/MetaMask/metamask-extension/blob/9404b57a09d70c9863b6599aa3183210cb13d2eb/app/scripts/controller-init/multichain/multichain-network-service-init.ts#L10-L12))

matthewwalsh0: How do we feel about the proxy / hook idea, to automatically inject this from any client request, meaning no controller changes needed at all? And rather the clients themselves are conceptually "authorised" and so any requests they generate are automatically authorised also?

packages/assets-controllers/src/TokenDetectionController.test.ts
```diff
@@ -2616,10 +2617,10 @@ describe('TokenDetectionController', () => {
             properties: {
               tokens: [`${sampleTokenA.symbol} - ${sampleTokenA.address}`],
               // TODO: Either fix this lint violation or explain why it's necessary to ignore.
-              // eslint-disable-next-line @typescript-eslint/naming-convention
```
Gudahtt, line 2619: Was this a mistake? It looks like the line below would still violate this lint rule.

If the rule doesn't apply for some reason, we should ensure the accompanying TODO comment is also removed (as well as this empty line).

mathieuartu: `autofixOnSave` was the culprit here 🕵️  Fixed with [0ce01be](https://github.com/MetaMask/core/pull/6066/commits/0ce01be8bb091fa174d6f99ac942ffeb970632d8)

mcmire (commented):
Not sure if we will end up making changes to this PR, but if not, here is a shallow pass.

packages/assets-controllers/src/TokenDetectionController.ts
mcmire: What would automatically injecting the token from any client request look like? Why would it mean no controller changes?

cryptodev-2s: Instead of modifying individual service/controller functions. We can also use a rule-based fetch interceptor that automatically adds authentication headers to matching URLs.

Note: The interceptor below only works with globalThis.fetch, so it has limitations when used with other HTTP clients. Additionally, it may need to be adapted to support both extension and mobile environments. We should also consider addressing known issues related to differences in polyfill timing, global scope variations, and Metro bundler behavior. The example is only for idea sharing.

```typescript
// packages/controller-utils/src/fetch-interceptor.ts
type AuthRule = {
  pattern: string | RegExp;
  getToken: () => Promise<string>;
  headerName?: string;
  enabled?: boolean;
};

class FetchInterceptor {
  private authRules: AuthRule[] = [];
  private originalFetch: typeof globalThis.fetch;
  private isSetup = false;

  constructor() {
    this.originalFetch = globalThis.fetch;
  }

  setup() {
    if (this.isSetup) return;
    
    globalThis.fetch = async (input: RequestInfo | URL, init?: RequestInit) => {
      const url = this.extractUrl(input);
      const matchedRule = this.findMatchingRule(url);
      
      if (matchedRule && matchedRule.enabled !== false) {
        try {
          const token = await matchedRule.getToken();
          const headerName = matchedRule.headerName || 'Authorization';
          const headerValue = headerName === 'Authorization' ? `Bearer ${token}` : token;
          
          init = {
            ...init,
            headers: {
              ...init?.headers,
              [headerName]: headerValue,
            },
          };
        } catch (error) {
          console.warn(`Failed to get auth token for ${url}:`, error);
          // Continue with original request
        }
      }
      
      return this.originalFetch(input, init);
    };
    
    this.isSetup = true;
  }

  addAuthRule(rule: AuthRule) {
    this.authRules.push(rule);
  }

  teardown() {
    if (this.isSetup) {
      globalThis.fetch = this.originalFetch;
      this.isSetup = false;
    }
  }

  private extractUrl(input: RequestInfo | URL): string {
    if (typeof input === 'string') return input;
    if (input instanceof URL) return input.href;
    return input.url || '';
  }

  private findMatchingRule(url: string): AuthRule | null {
    return this.authRules.find(rule => {
      if (typeof rule.pattern === 'string') {
        return url.includes(rule.pattern);
      }
      return rule.pattern.test(url);
    }) || null;
  }
}

export const fetchInterceptor = new FetchInterceptor();
```

That we can use in the clients side
```typescript
import { fetchInterceptor } from '@metamask/controller-utils';

export async function initializeControllers() {
  // Initialize auth controller first
  const authController = new AuthenticationController({...});
  
  // Setup authentication rules
  fetchInterceptor.addAuthRule({
    pattern: 'accounts.api.cx.metamask.io',
    getToken: () => authController.getBearerToken(),
  });
  
  // Activate the interceptor
  fetchInterceptor.setup();
  
  // All subsequent fetch calls are automatically authenticated
  const tokenDetectionController = new TokenDetectionController({...});
}
```

PR #6072, opened 2025-07-07
fix: use scopes to retrieve chain ID

darioAnongba:
## Explanation

We currently use the address to identify what network the account belongs to, yet that method is incorrect as:
- accounts span multiple networks 
- addresses can be the same on different networks.

The current implementation leads to a bug where the network will always be Mainnet for Solana and Bitcoin, preventing testnets to work. It would also fail to work on EVM networks if this function was used.

The solution, using scopes, fixes the problem for Bitcoin and introduces no regression for Solana.

## References

* Fixes [#12345
* Related to #67890](https://consensyssoftware.atlassian.net/browse/NWNT-397

## Changelog

Fix bug where network is not switched to testnets when creating a new Solana or Bitcoin account.

## Checklist


montelaidev (approved):
LGTM

PR #6075, opened 2025-07-08
feat: remove storage schema enforcement

mathieuartu:
## Explanation

Currently, public `UserStorageController` methods that interact with the storage SDK implicitly enforce schema validation (`validateAgainstSchema: true`). This is not suitable for all scenarios, as some consumers may need to store data that does not conform to a predefined schema or manage schema validation externally.

As a DX improvement, This PR removes the schema enforcement for user storage paths, so that developers are relieved from having to update the schema before using the SDK or Controllers for their features.

## References

Fixes: https://consensyssoftware.atlassian.net/browse/IDENTITY-164

Test-drive PRs:
- https://github.com/MetaMask/metamask-extension/pull/34137
- https://github.com/MetaMask/metamask-mobile/pull/16982

## Checklist

### DIFF
--- a/packages/bridge-controller/CHANGELOG.md
+++ b/packages/bridge-controller/CHANGELOG.md
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add an optional `isSingleSwapBridgeButtonEnabled` feature flag that indicates whether Swap and Bridge entrypoints should be combined ([#6078](https://github.com/MetaMask/core/pull/6078))
+
 ### Changed
 
 - **BREAKING:** Bump peer dependency `@metamask/assets-controllers` from `^69.0.0` to `^70.0.0` ([#6061](https://github.com/MetaMask/core/pull/6061))
 - **BREAKING:** Bump peer dependency `@metamask/snaps-controllers` from `^12.0.0` to `^14.0.0` ([#6035](https://github.com/MetaMask/core/pull/6035))
+- **BREAKING** Remove `isSnapConfirmationEnabled` feature flag from `ChainConfigurationSchema` validation ([#6077](https://github.com/MetaMask/core/pull/6077))
 - Bump `@metamask/controller-utils` from `^11.10.0` to `^11.11.0` ([#6069](https://github.com/MetaMask/core/pull/6069))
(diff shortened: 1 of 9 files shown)


mirceanis:
This closes #6074

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Remove schema enforcement for user storage paths. This will improve DX by preventing developers from having to update the schema before using the SDK or Controllers for their features.
```
mathieuartu: This is not a breaking change, as it relaxes type enforcement. No changes are required in the clients.

mathieuartu: This is in fact actually a breaking change. `UserStoragePathWithFeatureAndKey` is not exported anymore, and was used in the extension.

packages/profile-sync-controller/src/shared/storage-schema.ts
```diff
@@ -2,55 +2,20 @@ import { createSHA256Hash } from './encryption';
...
- *
- * Adding ALLOW_ARBITRARY_KEYS as the first key in the array allows for any key to be used for this feature.
- * This can be useful for features where keys are not deterministic (eg. accounts addresses).
+ * Any user storage path should be in the form of `feature.key`.
  */
-const ALLOW_ARBITRARY_KEYS = 'ALLOW_ARBITRARY_KEYS' as const;
 
+/**
+ * Helper object that contains the feature names used in the controllers and SDK.
+ * Developers don't need to add new feature names to this object anymore, as the schema enforcement has been deprecated.
```
fabiobozzo, line 10: > anymore

Nit-nit pick: this comment refers to code that no longer exists. Perhaps we could leave the old `USER_STORAGE_SCHEMA` and mark it as `@deprecated` for the time being.

fabiobozzo (approved):
Nice 💯

packages/profile-sync-controller/src/shared/storage-schema.ts
mathieuartu: Ha! I see what you mean, but this refers to `USER_STORAGE_FEATURE_NAMES`.
Personally, I think we (Identity) should continue using this object as it helps us with not using magic strings. We're just removing the _need_ to use it.

PR #6086, opened 2025-07-09
Release/461.0.0

jpuri:
## Explanation

Releasing change in PreferenceController to cleanup preference `smartAccountOptInForAccounts` as it is not going to be used.

## References

* Related to [#67890](https://github.com/MetaMask/MetaMask-planning/issues/5262)

## Checklist


cryptodev-2s (approved):
LGTM!

cryptodev-2s (dismissed):
LGTM!

packages/assets-controllers/package.json
```diff
@@ -113,7 +113,7 @@
     "@metamask/network-controller": "^24.0.0",
     "@metamask/permission-controller": "^11.0.0",
     "@metamask/phishing-controller": "^12.5.0",
-    "@metamask/preferences-controller": "^18.0.0",
+    "@metamask/preferences-controller": "^19.0.0",
```
mcmire, line 116: Since the peer dependency got bumped in this package, we will need to include this package in the release as well. You'll want to document that the peer dependency on `preferences-controller` was bumped in the changelog for `assets-controllers` and release a new version of this package. 

However because many controllers depend on `assets-controllers` and we would be bumping the major on this package, you will also need to include any other packages that rely on it. That's why I was saying that it might be easier to deprecate the method for now. Just putting that option out there in case you don't have time right now.

packages/preferences-controller/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [19.0.0]
+
 ### Changed
```
mcmire: Should we use the "Removed" category here?

``` suggestion
### Removed

mcmire (changes requested):

packages/assets-controllers/package.json
jpuri: Hey @mcmire : `preferences-controller` is devDependency of `assets-controllers`, thus it may not need a release.

mcmire: It's included in `devDependencies` for development purposes, but it's also included in `peerDependencies`.

PR #6090, opened 2025-07-09
Improve create-update-issues workflow

### LINKED ISSUE #6089: `create-update-issues` workflow may not work correctly if a label is missing from a client repo
With the team topology changes, some team labels in the extension and mobile repos have been renamed. This breaks the workflow that automatically creates issues in client repos when a new major version of a package is published.

mcmire:
## Explanation

We have a workflow to automatically create tickets in both Extension and Mobile whenever a new major version of a package is published. However, there are a few of issues with this workflow:

- The command to create the issue can fail if a team label doesn't exist in either Extension or Mobile. This seems to prevent other issues from being created.
- It's actually difficult to tell which issues get successfully created and which don't, because there isn't much output from this workflow.
- The description in the issue is very terse. It says "Please update this package", assigns it to a team, and doesn't explain why that should care or what they should do next.
- The script in the workflow is difficult to maintain because it's inline.
- It's difficult to test the script without forking the repo.

This commit extracts the script to a separate file and addresses the above issues by adding better failure handling and logging, expanding the issue description, and adding a dry-run mode.

To test the changes here, run:

```
./scripts/create-update-issues.sh --ref v460.0.0

./scripts/create-update-issues.sh --ref v459.0.0
```

## References

Fixes #6089.

## Checklist

### DIFF
--- a/scripts/create-update-issues.sh
+++ b/scripts/create-update-issues.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+DEFAULT_LABEL="client-controller-update"
+
+run-create-issue-command() {
+  local dry_run="$1"
+  local repo="$2"
+  local package_name="$3"
+  local version="$4"
+  local labels="$5"
+
+  if [[ $dry_run -eq 1 ]]; then
+    echo "> gh issue create --title \"$title\" --body \"$body\" --repo \"$repo\" --label \"$labels"\"
+  else
+    gh issue create --title "$title" --body "$body" --repo "$repo" --label "$labels"
+  fi
+}
+
+create-issue() {
+  local dry_run="$1"
+  local repo="$2"
+  local package_name="$3"
+  local version="$4"
+  local labels="$5"
+
+  local title="Upgrade ${package_name} to version ${version}"
+  local body="A new major version of \`${package_name}\`, ${version}, is now available. This issue has been assigned to you and your team because you code-own this package in the \`core\` repo. If this package is present in this project, please prioritize upgrading it soon to unblock new features and bugfixes."
+
+  echo "Creating issue in ${repo} with labels: ${labels}"
+
+  local exitcode
+
+  run-create-issue-command "$dry_run" "$repo" "$package_name" "$version" "$labels"
+  exitcode=$?
+
+  if [[ $exitcode -ne 0 ]]; then
+    echo "That didn't work, trying to create issue in ${repo} for ${package_name} ${version} without labels"
+
+    run-create-issue-command "$dry_run" "$repo" "$package_name" "$version"
+    exitcode=$?
+  fi
+
+  if [[ $exitcode -eq 0 ]]; then
+    echo "Successfully created issue!"
+  else
+    echo "That didn't work, please create the issue manually"
+  fi
+
+  return $exitcode
+}
+
+main() {
+  local tag_array
+  local exitcode
+  local package_name
+  local package_name_without_leading_at
+  local version
+  local team_labels
+  local labels
+
+  local dry_run=1
+  local ref="HEAD"
+
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      --ref)
+        ref="$2"
+        shift 2
+        ;;
+      --no-dry-run)
+        dry_run=0
+        shift
+        ;;
+      *)
+        echo "Unknown argument: $1"
+        echo "Usage: $0 [--ref=REF] [--no-dry-run]"
+        exit 1
+        ;;
+    esac
+  done
+
+  if [[ $dry_run -eq 1 ]]; then
+    echo "[[[ DRY-RUN MODE ]]]"
+    echo
+  fi
+
+  local full_ref="$(git rev-parse "$ref")"
+
+  echo "Looking for release tags pointing to $full_ref for major-bumped packages..."
+  IFS=$'\n' read -r -d '' -a tag_array < <(git tag --points-at "$full_ref" | grep -E '^@metamask/[a-z0-9-]+@\d+\.0\.0$' && printf '\0')
+
+  if [[ "${#tag_array[@]}" -eq 0 ]]; then
+    echo "No tags to process, nothing to do."
+    exit
+  fi
+
+  for tag in "${tag_array[@]}"; do
+    # The tag name looks like "<package_name>@<version>",
+    # and "<package_name>" looks like "@metamask/*"
+    package_name="${tag%@*}"
+    package_name_without_leading_at="${package_name#@}"
+    version="${tag##*@}"
+
+    echo
+    echo "=== ${package_name} ${version} ==="
+    echo
+
+    # Use teams.json to determine which teams code-own this package, and what their labels are
+    team_labels=$(jq -r --arg key "$package_name_without_leading_at" '.[$key]' teams.json)
+    labels="$DEFAULT_LABEL"
+    if [[ $team_labels == "null" ]]; then
+      echo "Did not find team labels for ${package_name}, will create issues anyway"
+    else
+      echo "Found team labels for ${package_name}: ${team_labels}"
+      labels+=",$team_labels"
+    fi
+
+    # Create the issues
+    echo
+    create-issue "$dry_run" "MetaMask/metamask-extension" "$package_name" "$version" "$labels"
+    echo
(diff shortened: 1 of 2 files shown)


scripts/create-update-issues.sh
```diff
@@ -0,0 +1,134 @@
...
+        ;;
+    esac
+  done
+
+  if [[ $dry_run -eq 1 ]]; then
+    echo "[[[ DRY-RUN MODE ]]]"
+    echo
+  fi
+
+  local full_ref="$(git rev-parse "$ref")"
```
Gudahtt: Nit: This fixes ShellCheck warning
```suggestion
  local full_ref
  full_ref="$(git rev-parse "$ref")"
```

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
```
Gudahtt, line 2: Nit: Thoughts on adding our standard Bash script flags here? They weren't there before I know, but it would be preferred to halt upon error I would think.

```suggestion

set -e
set -u
set -o pipefail

```

Gudahtt: You'd need to move the exit code capturing code if we did this though.

e.g.
```
command  || exitcode="$?"
```

instead of 

```
command
exitcode="$?"
```

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,134 @@
...
+  echo "Creating issue in ${repo} with labels: ${labels}"
+
+  local exitcode
+
+  run-create-issue-command "$dry_run" "$repo" "$title" "$body" "$labels"
+  exitcode=$?
+
+  if [[ $exitcode -ne 0 ]]; then
+    echo "That didn't work, trying to create issue in ${repo} for ${package_name} ${version} without labels"
```
Gudahtt: Is this behavior something that we want? I'd think that we'd want to know if a label was incorrect. That's a problem we'd want to fix, and which is very easy to fix. An issue with no labels won't be seen, so it won't really serve its intended purpose.

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,134 @@
...
+
+    echo
+    echo "=== ${package_name} ${version} ==="
+    echo
+
+    # Use teams.json to determine which teams code-own this package, and what their labels are
+    team_labels=$(jq -r --arg key "$package_name_without_leading_at" '.[$key]' teams.json)
+    labels="$DEFAULT_LABEL"
+    if [[ $team_labels == "null" ]]; then
+      echo "Did not find team labels for ${package_name}, will create issues anyway"
```
Gudahtt: Similar question here, is this what we want to happen? I would prefer an error here so we know something is broken.

teams.json
Gudahtt, line 1: Nit: Can we make these changes in a separate PR? This seems distinct from the script rewrite, and the sooner we get this in the better.

scripts/create-update-issues.sh
Gudahtt: Maybe we could do both; create the issues so it's easier to fix manually, but also return an error.

teams.json
mcmire: That's fair. Done here: https://github.com/MetaMask/core/pull/6107

scripts/create-update-issues.sh
mcmire: You're right, this isn't correct. We still need to add the ticket with a `client-controller-update` label at least.

scripts/create-update-issues.sh
mcmire: We still add the `client-controller-update` label, so may not be seen by a team, but it will be added to the Package Updates board.

My thought here was that we could add a view within the board for tickets that are missing team labels. Then every so often we'd check this tab and assign the right teams.

But you're right, we could make this workflow error so we could see it when observing a deployment and fix it right there.

scripts/create-update-issues.sh
mcmire: Done here: [0d2ebd8](https://github.com/MetaMask/core/pull/6090/commits/0d2ebd817bf511fce845537da99730d6fd7dea33)

scripts/create-update-issues.sh
mcmire: Fixed here: [5f5714c](https://github.com/MetaMask/core/pull/6090/commits/5f5714cf2b9dfe83a43af6b8047392c87cb026a3)

scripts/create-update-issues.sh
mcmire: Fixed here: [5f5714c](https://github.com/MetaMask/core/pull/6090/commits/5f5714cf2b9dfe83a43af6b8047392c87cb026a3)

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,240 @@
...
+EOT
+}
+
+existing-issue-found() {
+  local repo="$1"
+  local package_name="$2"
+  local version="$3"
+
+  local all_issues
+  if ! all_issues=$(gh issue list --repo "$repo" --label "$DEFAULT_LABEL" --state open --json number,title,url 2>&1); then
```
Gudahtt: Hmm. This seems a bit inefficient, fetching this list once per published package.

Maybe we could fetch the list of existing issues just twice (for extension and mobile respective) and pass it into this function instead. Something like this: https://github.com/MetaMask/core/pull/6143

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,240 @@
...
+
+  if [[ -z "$all_issues" || "$all_issues" == "[]" ]]; then
+    echo "Found no issues in the repo."
+    return 1
+  fi
+
+  local package_name_without_at="${package_name#@}"
+  local package_name_escaped="${package_name//./\\.}"
+  local package_name_without_at_escaped="${package_name_without_at//./\\.}"
+  local regex_pattern="^(Update|Upgrade) (${package_name_escaped}|${package_name_without_at_escaped}) to version ${version//./\\.}$"
```
Gudahtt, line 40: Curious, did we change the issue names at some point? I see we're matching with and without the `@`, and with either Update or Upgrade

mcmire: Yes, this PR improves the issue names as well, here: https://github.com/MetaMask/core/pull/6090/files#diff-0317ced59688181bdf7140d859f78c8eb7d2ecb14f8a7e29916e82602397ba94R82. I've always found the existing terminology to be slightly annoying. I feel that "upgrade" is a better word to use than "update" (which could imply changing the package, adding a patch, or something like that). Also, `@` was somehow missing from the issue title. So you'd get something like "Update metamask/foo to version 1.2.3" which seemed strange. Now it says "Upgrade @metamask/foo to 1.2.3". Maybe minor, but I figure it could help.

scripts/create-update-issues.sh
mcmire: Thanks, I've merged this.

scripts/create-update-issues.sh
```diff
@@ -0,0 +1,270 @@
...
+  local team_labels
+
+  local exitcode=0
+  local dry_run=1
+  local ref="$DEFAULT_REF"
+
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      --ref|-r)
+        if [[ -n "${2:-}" ]] && ! [[ "$2" =~ ^- ]]; then
```
Gudahtt, line 133: What is the purpose of the `&& ! [[ "$2" =~ ^- ]]` part of this condition? Why does it matter if the ref starts with a dash?

mcmire: Very minor, but I guess it depends on how we want to handle a case like this:

```
./scripts/create-update-issues.sh --ref --some-other-option
```

Should the script interpret `--some-other-option` as another option entirely? Or should it treat it as a value for `--ref`? Currently, it does the former.

Gudahtt: Ahh I see. Personally I'd expect it to treat `--some-other-option` in that example as a ref. But it's definitely ambiguous, and I can't see this coming up very often.

Not sure this is ideal, but not a blocker either way

cryptodev-2s (approved):
The changes look good to me — I tested dry mode locally and it worked as expected.

Quick question: have we ever considered using [google/zx](https://google.github.io/zx/) for scripting?

mcmire:
@cryptodev-2s I haven't heard of that, but I will check it out!

PR #6097, opened 2025-07-10
feat: KDF performance test 2: keep scrypt but dumb down parameters

mathieuartu:
## Checklist

### DIFF
--- a/packages/profile-sync-controller/src/sdk/user-storage.ts
+++ b/packages/profile-sync-controller/src/sdk/user-storage.ts
@@ -1,7 +1,6 @@
 import type { IBaseAuth } from './authentication-jwt-bearer/types';
 import { NotFoundError, UserStorageError } from './errors';
 import encryption, { createSHA256Hash } from '../shared/encryption';
-import { SHARED_SALT } from '../shared/encryption/constants';
 import type { Env } from '../shared/env';
 import { getEnvUrls } from '../shared/env';
 import type {
@@ -318,10 +317,13 @@ export class UserStorage {
         options?.nativeScryptCrypto,
       );
 
-      // Re-encrypt the entry if it was encrypted with a random salt
-      const salt = encryption.getSalt(encryptedData);
-      if (salt.toString() !== SHARED_SALT.toString()) {
-        await this.#upsertUserStorage(path, decryptedData, options);
+      // Migrate data from v1 to v2 encryption
+      if (JSON.parse(encryptedData).v === '1') {
+        const reEncryptedData = await encryption.encryptString(
+          decryptedData,
+          storageKey,
+        );
+        await this.#upsertUserStorage(path, reEncryptedData, options);
       }
 
       return decryptedData;
@@ -388,17 +390,13 @@ export class UserStorage {
           );
           decryptedData.push(data);
 
-          // Re-encrypt the entry was encrypted with a random salt
-          const salt = encryption.getSalt(entry.Data);
-          if (salt.toString() !== SHARED_SALT.toString()) {
-            reEncryptedEntries.push([
-              entry.HashedKey,
-              await encryption.encryptString(
-                data,
-                storageKey,
-                options?.nativeScryptCrypto,
-              ),
-            ]);
+          // Migrate data from v1 to v2 encryption
+          if (JSON.parse(entry.Data).v === '1') {
+            const reEncryptedData = await encryption.encryptString(
+              data,
+              storageKey,
+            );
+            reEncryptedEntries.push([entry.HashedKey, reEncryptedData]);
           }
         } catch {
           // do nothing
--- a/packages/profile-sync-controller/src/shared/encryption/constants.ts
+++ b/packages/profile-sync-controller/src/shared/encryption/constants.ts
@@ -6,6 +6,7 @@ export const ALGORITHM_KEY_SIZE = 16; // 16 bytes
 // see: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt
 export const SCRYPT_SALT_SIZE = 16; // 16 bytes
 export const SCRYPT_N = 2 ** 17; // CPU/memory cost parameter (must be a power of 2, > 1)
+export const SCRYPT_N_V2 = 2;
 export const SCRYPT_r = 8; // Block size parameter
 export const SCRYPT_p = 1; // Parallelization parameter
 
--- a/packages/profile-sync-controller/src/shared/encryption/encryption.ts
+++ b/packages/profile-sync-controller/src/shared/encryption/encryption.ts
@@ -13,6 +13,7 @@ import {
   ALGORITHM_KEY_SIZE,
   ALGORITHM_NONCE_SIZE,
   SCRYPT_N,
+  SCRYPT_N_V2,
   SCRYPT_p,
   SCRYPT_r,
   SCRYPT_SALT_SIZE,
@@ -28,7 +29,7 @@ import type { NativeScrypt } from '../types/encryption';
 
 export type EncryptedPayload = {
   // version
-  v: '1';
+  v: '1' | '2';
 
   // key derivation function algorithm - scrypt
   t: 'scrypt';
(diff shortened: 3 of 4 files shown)


mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Dumb down the `N` `scrypt` parameter in order to increase performance on clients
+  - This decreases the impact of the KDF on all clients to around 99%
+  - Add backwards compatible migration logic
+  - Add encryption callbacks so we can measure the impact of future migrations through analytics
```
mirceanis:
```suggestion
- - Reduce the `N` `scrypt` parameter used to compute the encryption key in order to significantly increase performance on clients.
  - The input to the KDF is already a long, high entropy non-user-generated string so it does not make sense to use a KDF with high compute cost.
  - This decreases the impact of the KDF on all clients by around 99%
  - Add backwards compatible migration logic
  - Add encryption callbacks so we can measure the impact of future migrations through analytics
```

mirceanis (commented):
Looks good!
Perhaps we can rephrase the changelog to reenforce the reason for this change.

Since this PR deals with encryption "soft" migration I'd take the opportunity to make sure that future "hard" migrations fail gracefully, meaning that clients don't fail entire operations if one of the entries does not decrypt. See lines like `if (encryptedData.v === '1').... else throw`.
WDYT @mathieuartu ?

packages/profile-sync-controller/CHANGELOG.md
mathieuartu: I like this! Fixed with [ea7474e](https://github.com/MetaMask/core/pull/6097/commits/ea7474e2f69a4ac072e21e70ecc3aa8e73b42b3e)

packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -69,16 +86,27 @@ class EncryptorDecryptor {
...
-    nativeScryptCrypto?: NativeScrypt,
+    options?: {
+      nativeScryptCrypto?: NativeScrypt;
+      onDecrypt?: (encryptedData: Omit<EncryptedPayload, 'd'>) => Promise<void>;
+    },
   ): Promise<string> {
     try {
       const encryptedData: EncryptedPayload = JSON.parse(encryptedDataStr);
+
+      await options?.onDecrypt?.({
```
mirceanis, line 100: Comment for a future PR here: having so many callbacks that are essentially used for analytics makes me thing we should replace them with a single callback that can pipe back multiple types of events.

mirceanis (dismissed):
looks good

packages/profile-sync-controller/src/shared/encryption/encryption.ts
mathieuartu: This is very valid! Yep! Adding to the board :)

mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/src/shared/encryption/constants.ts
```diff
@@ -6,6 +6,7 @@ export const ALGORITHM_KEY_SIZE = 16; // 16 bytes
 // see: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt
 export const SCRYPT_SALT_SIZE = 16; // 16 bytes
 export const SCRYPT_N = 2 ** 17; // CPU/memory cost parameter (must be a power of 2, > 1)
+export const SCRYPT_N_V2 = 2;
```
Mrtenz, line 9: This is well below the recommended minimum. Have you considered the security implications of changing this?

mathieuartu: @Mrtenz Good point! Yes we did, as the input to the KDF is already a long, high entropy non-user-generated string so it does not make sense to use a KDF with high compute cost.

We even considered completely removing the KDF, but kept it in order to maintain backwards compatibility.
LMK if that makes sense for you!

cc @danroc @mirceanis

mirceanis: Using `scrypt` here as KDF was a mistake in the first place because the input is a 32 byte string generated from a signature. The input will soon be replaced with a randomly generated 32 byte string, so it will make even less sense in the future.

However, `scrypt` with the params recommended for regular user provided passwords is way too expensive to use here, taking 500ms on desktop and 2000ms on mobile even on high end devices, blocking the UI for this time on firefox and mobile and potentially making things janky on chrome.

For backwards compatibility we will continue to use scrypt as a KDF with relaxed parameters until time comes for a real migration at which point we will remove it completely.

Mrtenz: Ok, makes sense. Though changing the `N` value is not backwards compatible either way, so maybe consider removing it altogether?

mathieuartu: @Mrtenz It is not backwards compatible per se, you're right. Different `N` means different derived keys. But since we save the encryption parameters alongside the encrypted data in the payload, keeping `scrypt` with the lowest `N` means that the old code can still function while receiving "New `N`" payloads.

packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -238,7 +263,7 @@ class EncryptorDecryptor {
     salt?: Uint8Array,
     nativeScryptCrypto?: NativeScrypt,
   ) {
-    const hashedPassword = createSHA256Hash(password);
+    const hashedPassword = `${createSHA256Hash(password)}.${o.N}.${o.r}.${o.p}.${o.dkLen}`;
```
mirceanis: I know this is a nit but why not
```suggestion
    const hashedPassword = createSHA256Hash(`${password}.${o.N}.${o.r}.${o.p}.${o.dkLen}`);
```
?

mirceanis: Actually, since this is only used as an in-memory cache, there's actually no point to hashing anything
```suggestion
    const hashedPassword = `${password}.${o.N}.${o.r}.${o.p}.${o.dkLen}`;
```

mirceanis (dismissed):

mirceanis (approved):

mirceanis (dismissed):

packages/profile-sync-controller/src/sdk/user-storage.ts
```diff
@@ -381,24 +383,23 @@ export class UserStorage {
...
-                options?.nativeScryptCrypto,
-              ),
-            ]);
+          // Data migration
+          if (encryption.doesEntryNeedReEncryption(entry.Data)) {
+            const reEncryptedData = await encryption.encryptString(
+              data,
+              storageKey,
+              {
+                nativeScryptCrypto: options?.nativeScryptCrypto,
```
fabiobozzo, line 398: I don't fully get it. The data is re-encrypted with.... the same `storageKey`? I guess then `nativeScryptCrypto` is what includes the lower `N` value when re-encrypting? @mathieuartu

fabiobozzo (dismissed):

packages/profile-sync-controller/src/sdk/user-storage.ts
mathieuartu: Yeah, it's not obvious at first glance, but the default `N` is now `2` so if we re-encrypt data that had `N: 2**17`, it will become `N: 2`

mathieuartu:
@metamaskbot publish-preview

fabiobozzo (dismissed):

mathieuartu:
Going back to draft again as it seems to be working fine now but may have introduced a UI bug on mobile when adding or renaming accounts. Investigating.

PR #6104, opened 2025-07-10
fix: handle snaps error

salimtb:
## Explanation

The `MultichainAssetsRatesController` was experiencing crashes when Snap requests failed due to parameter validation errors. 

the error handling was insufficient - when Snap requests failed, the controller will send a sentry error rather than gracefully handling the error and continuing operation.

## Checklist

### DIFF
--- a/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.test.ts
+++ b/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.test.ts
@@ -598,32 +598,6 @@ describe('MultichainAssetsRatesController', () => {
       consoleErrorSpy.mockRestore();
     });
 
-    it('handles unknown error types gracefully', async () => {
-      const { controller, messenger } = setupController();
-
-      const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation();
-      const unknownError = 'String error instead of Error object';
-
-      const snapHandler = jest.fn().mockRejectedValue(unknownError);
-      messenger.registerActionHandler(
-        'SnapController:handleRequest',
-        snapHandler,
-      );
-
-      await controller.updateAssetsRates();
-
-      // Should have logged the unknown error
-      expect(consoleErrorSpy).toHaveBeenCalledWith(
-        'Unknown error in snap request:',
-        unknownError,
-      );
-
-      // Should not update state when snap request fails
-      expect(controller.state.conversionRates).toStrictEqual({});
-
-      consoleErrorSpy.mockRestore();
-    });
-
     it('handles mixed success and failure scenarios', async () => {
       const { controller, messenger } = setupController({
         accountsAssets: [fakeNonEvmAccount, fakeEvmAccount2],


packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts
```diff
@@ -417,14 +422,22 @@ export class MultichainAssetsRatesController extends StaticIntervalPollingContro
...
+            origin: 'metamask',
+            handler: HandlerType.OnAssetHistoricalPrice,
+            request: {
+              jsonrpc: '2.0',
+              method: HandlerType.OnAssetHistoricalPrice,
+              params: {
+                from: asset,
+                to: currentCaipCurrency,
+              },
+            },
```
FrederikBolding, line 438: You shouldn't need to nest it like this. I would just revert back to just passing the params as you had before

packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts
```diff
@@ -635,26 +652,40 @@ export class MultichainAssetsRatesController extends StaticIntervalPollingContro
...
+      if (error instanceof Error) {
+        console.error(`Snap request failed for ${handler}:`, {
+          snapId,
+          handler,
+          message: error.message,
+          params,
+        });
+      } else {
+        console.error('Unknown error in snap request:', error);
+      }
```
FrederikBolding: I think the error should always be an instance of `Error`, we can probably simplify this

packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts
```diff
@@ -417,14 +422,22 @@ export class MultichainAssetsRatesController extends StaticIntervalPollingContro
...
         );
       try {
-        const historicalPricesResponse = await this.#handleSnapRequest({
-          snapId: selectedAccount?.metadata.snap?.id as SnapId,
-          handler: HandlerType.OnAssetHistoricalPrice,
-          params: {
-            from: asset,
-            to: currentCaipCurrency,
+        const historicalPricesResponse = await this.messagingSystem.call(
+          'SnapController:handleRequest',
```
FrederikBolding, line 426: Is the expectation on the UI side that this should throw if the data is unavailable? Because it seems like it is also valid to return nothing. Can we use the same error handling that we've introduced for the other handlers?

packages/assets-controllers/CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **BREAKING:** Bump peer dependency `@metamask/phishing-controller` to `^13.0.0` ([#6098](https://github.com/MetaMask/core/pull/6098))
 
+### Fixed
```
cryptodev-2s, line 14: Nit: can we move this CHANGELOG under the `Unreleased` section instead of version `71.0.0`, and remove the `no-changelog` label.

PR #6106, opened 2025-07-11
fix: TAT-1263 pooled-staking data fails to refresh when on unsupported chainId

Matt561:
## Explanation

Currently, when the GNS active chainId isn't supported by pooled-staking we cannot fetch pooled-staking data. Supported chains are Ethereum mainnet and Hoodi Testnet.

To fix this, we fallback to Ethereum mainnet when the active chain isn't supported.
 
## Fixes
- [TAT-1263](https://consensyssoftware.atlassian.net/browse/TAT-1263): Fix staking data refresh issue when switching accounts in GNS


packages/earn-controller/src/EarnController.ts
```diff
@@ -500,6 +495,25 @@ export class EarnController extends BaseController<
...
+    if (
+      !activeChainId ||
+      !this.#supportedPooledStakingChains.includes(toHex(activeChainId))
+    ) {
+      // Fallback to Ethereum Mainnet if chainId is not supported.
+      return convertHexToDecimal(CHAIN_IDS.MAINNET);
+    }
+
+    return activeChainId;
+  }
```
ameliejyc, line 515: I'm still getting to grips with what we need to do on Earn side regarding GNS removal, but it looks like if we are removing the `chainId` params of the pooled staking calls in favour of using this function within them, then we are increasing our coupling to `getCurrentChainId()` and therefore `NetworkController:getState` which is something we want to avoid? 

If instead we want to always pass in a `chainId` to the pooled staking calls then we can still use this logic to check if it's a supported chain but without calling `NetworkController:getState`. Let me know if I'm getting this right and if this is an interim fix to get it working without having to change the mobile code also!

Matt561: Your thought process is 100% correct. This does increase coupling to the GNS' active network and it's something we don't want moving forward. This set of changes is meant as a stopgap to fix a [bug](https://consensyssoftware.atlassian.net/browse/TAT-1263) discovered while testing the `7.51.0` mobile release.

We will want to pass in the `chainId` most likely when making calls to EarnController methods and handle cases for when they're supported/unsupported.

I will quickly refactor this to revert removal of the `chainId` param and instead use `getActivePooledStakingChainId` (I'll likely rename) to be a fallback/validation function.

ameliejyc: All good, I think this as a stopgap is fine as it is so I'll approve to unblock you!

ameliejyc (approved):

PR #6115, opened 2025-07-15
feat!: use `@metamask/account-api` in the `AccountTreeController`

ccharly:
## Explanation

Use the new `@metamask/account-api` that introduces the concept of account wallet and account group into a generic account API/SDK package.

This controller was already using those concepts, but it now properly implements well-defined types that will also be used by multichain accounts.

## References

- https://github.com/MetaMask/accounts/pull/307

## Changelog

N/A

## Checklist


packages/account-tree-controller/src/rules/EntropySourceWalletRule.ts
```diff
@@ -0,0 +1,129 @@
...
+      this.#wallets.set(wallet.id, wallet);
+    }
+
+    // This will automatically creates the group if it's missing.
+    const group = wallet.addAccount(account);
+
+    return {
+      wallet,
+      group,
+    };
```
ccharly, line 127: This code is being repeated several times and could be moved to the `BaseWalletRule` class maybe. For now, I decided to keep this simple, since the rules might be refactored even more once we link this controller with the new `MultichainAccount{Controller,Service}`

ccharly:
@metamasbot publish-preview

ccharly:
@metamaskbot publish-preview

PatrykLucka (approved):

ccharly:
For some reason, the preview builds seem not up-to-date... I'll re-build them now.

ccharly:
@metamaskbot publish-preview

PatrykLucka (dismissed):

packages/account-tree-controller/src/AccountTreeGroup.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.#accounts = new Set();
+  }
+
+  get accounts(): AccountId[] {
+    return Array.from(this.#accounts); // FIXME: Should we force the copy here?
+  }
+
+  getAccounts(): InternalAccount[] {
+    const accounts = [];
+
```
montelaidev: Maybe the `accounts` getter should be `accountIds` because `getAccounts` returns an array of InternalAccounts

ccharly: Yes.. I thought of this too actually

Also, I really wonder if I should go with `getAccountIds` instead. Given that we're still making an actual "copy" of the sets, I don't want to "hide" this copy behind a "false property" access

montelaidev (approved):

Issue #6113, opened 2025-07-14
RpcService: Making a request to an endpoint with basic auth credentials in the URL does not work


mcmire:
`RpcService` has where fetching an RPC endpoint inside of  where the URL contains a username and password (e.g. `https://username:password@example.rpc.chain`) will produce the following error:

```
Request cannot be constructed from a URL that includes credentials
```

This is happening because while `RpcService` converts the basic auth credentials to an `Authorization` header [here](https://github.com/MetaMask/core/blob/525d6629ecc1bb5b7b01bb1e2da1cb15f05cba8a/packages/network-controller/src/rpc-service/rpc-service.ts#L444) it does not remove them from the URL. This was happening in the `fetch` middleware where the code for `RpcService` was taken indirectly by normalizing the request URL [here](https://github.com/MetaMask/eth-json-rpc-middleware/blob/d2578c5e92c53fb08257a6d2be7979f5a19182ed/src/fetch.ts#L331). We may or may not have to do something similar.

This is also logged on the extension side here: https://github.com/MetaMask/metamask-extension/issues/34241

mcmire:
This is reproduceable with the following test:

``` typescript
    it.only('extracts a username and password from the URL to the Authorization header', async () => {
      const scope = nock('https://rpc.example.chain', {
        reqheaders: {
          Authorization: 'Basic dXNlcm5hbWU6cGFzc3dvcmQ=',
        },
      })
        .post('/', {
          id: 1,
          jsonrpc: '2.0',
          method: 'eth_chainId',
          params: [],
        })
        .reply(200, {
          id: 1,
          jsonrpc: '2.0',
          result: '0x1',
        });
      const promiseForRequestUrl = new Promise<string>((resolve) => {
        scope.on('request', (request) => {
          resolve(request.options.href);
        });
      });
      const service = new RpcService({
        fetch,
        btoa,
        endpointUrl: 'https://username:password@rpc.example.chain',
      });

      const response = await service.request({
        id: 1,
        jsonrpc: '2.0',
        method: 'eth_chainId',
        params: [],
      });

      expect(response).toStrictEqual({
        id: 1,
        jsonrpc: '2.0',
        result: '0x1',
      });
      expect(await promiseForRequestUrl).toBe('https//rpc.example.chain');
    });
```

PR #6116, opened 2025-07-15
fix: strip credentials from RPC endpoint URL after extracting for auth header

cryptodev-2s:
## Explanation

This PR fixes the error "Request cannot be constructed from a URL that includes credentials" that occurs when using RPC endpoints with embedded credentials.

When an RPC endpoint URL contains credentials (e.g., `https://username:password@rpc.example.com`), the fetch API throws an error:
```
TypeError: Request cannot be constructed from a URL that includes credentials: https://username:password@rpc.example.com/
```

This is because modern fetch implementations do not allow credentials to be passed directly in the URL for security reasons.

The `RpcService` already extracts credentials from the URL to create a Basic Authorization header. This fix ensures the credentials are also stripped from the URL before it's used in fetch requests, preventing the error while maintaining the authentication functionality.

## References

https://github.com/MetaMask/core/issues/6113

## Checklist

### DIFF
--- a/packages/network-controller/src/rpc-service/rpc-service.test.ts
+++ b/packages/network-controller/src/rpc-service/rpc-service.test.ts
@@ -758,7 +758,7 @@ describe('RpcService', () => {
     });
 
     it('extracts a username and password from the URL to the Authorization header', async () => {
-      nock('https://rpc.example.chain', {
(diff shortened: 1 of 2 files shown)


cryptodev-2s:
@metamaskbot publish-preview

mcmire (approved):
LGTM!

Issue #5626, opened 2025-04-10
Implement messenger delegation


Gudahtt:
Implement messenger delegation as described in https://github.com/MetaMask/decisions/pull/53. This would include removal of the` RestrictedMessenger` class/type, and replacing `RestrictedMessenger` type declarations with Messenger type declarations.

This will be a breaking change for `@metamask/base-controller`, and for anything that exports a type declaration using `RestrictedMessenger` (including all controllers)

PR #6132, opened 2025-07-17
feat: Add messenger `delegate` and `revoke` methods

Gudahtt:
## Explanation

The methods `delegate` and `revoke` have been added to the Messenger class. These methods replace the need for the `RestrictedMessenger`. The `getRestricted` method, and the `RestrictedMessenger` class, have been removed as obsolete.

Note that the `parent` constructor parameter described in the ADR is not implemented yet, that will come in the next PR.

## References

See this ADR PR for details: MetaMask/decisions#53
Relates to #5626

## Checklist


packages/messenger/src/Messenger.ts
```diff
@@ -101,36 +99,64 @@ export type NotNamespacedBy<
...
 > {
-  readonly #actions = new Map<Action['type'], unknown>();
+  readonly #namespace: Namespace;
+
+  readonly #actions = new Map<Action['type'], Action['handler']>();
 
   readonly #events = new Map<Event['type'], EventSubscriptionMap<Event>>();
 
+  /**
+   * The set of messengers we've delegated events to, by event type.
```
Gudahtt, line 142:
```suggestion
   * The set of messengers we've delegated events to and their event handlers, by event type.
```

packages/messenger/src/Messenger.ts
```diff
@@ -149,6 +175,16 @@ export class Messenger<
...
+   *
+   * @param args - Constructor arguments
+   * @param args.namespace - The messenger namespace.
+   */
+  constructor({ namespace }: { namespace: Namespace }) {
+    this.#namespace = namespace;
+  }
+
   /**
    * Register an action handler.
```
mcmire, line 236: To better distinguish this method from `registerDelegatedActionHandler`, should we clarify that the action here is one that the messenger "owns"?

``` suggestion
   * Register a handler for an action under this messenger's namespace.
```

Or, perhaps literally:

``` suggestion
   * Register an action owned by this messenger.
```

packages/messenger/src/Messenger.ts
```diff
@@ -160,7 +196,42 @@ export class Messenger<
    * @throws Will throw when a handler has been registered for this action type already.
    * @template ActionType - A type union of Action type strings.
```
mcmire: Maybe this should say:

``` suggestion
   * @template ActionType - A union of type strings for actions under this messenger's namespace.
```

packages/messenger/src/Messenger.ts
```diff
@@ -160,7 +196,42 @@ export class Messenger<
...
+    this.#registerActionHandler(actionType, handler);
+  }
+
+  /**
+   * Register a delegated action handler.
+   *
+   * This will make the registered function available to call via the `call` method.
+   *
+   * @deprecated Do not call this directly, instead use the `delegate` method.
+   * @param actionType - The action type. This is a unqiue identifier for this action.
```
mcmire: Typo:

``` suggestion
   * @param actionType - The action type. This is a unique identifier for this action.

File: packages/messenger/src/Messenger.ts [hunk 5]
```diff
@@ -160,7 +196,42 @@ export class Messenger<
...
+        }:'`,
+      );
+    }
+    this.#registerActionHandler(actionType, handler);
+  }
+
+  /**
+   * Register a delegated action handler.
+   *
+   * This will make the registered function available to call via the `call` method.
```
mcmire: Is it worth expanding this description slightly to clarify what "delegated" means? My thought is that it might seem odd that we are defining a method that is very similar to `registerActionHandler`, but doesn't require that the action has this messenger's namespace or that the action is one that this messenger recognizes.

``` suggestion
   * Register a delegated action handler (that is, a handler which comes from another messenger
   * that this messenger is permitted to call).
   *
   * This will make the registered function available to call via the `call` method.
```

packages/messenger/src/Messenger.ts
```diff
@@ -180,7 +251,7 @@ export class Messenger<
    * handlers
    */
   registerMethodActionHandlers<
-    MessengerClient extends { name: string },
+    MessengerClient extends { name: Namespace },
```
mcmire, line 284: Ah, good catch.

packages/messenger/src/Messenger.ts
```diff
@@ -200,10 +271,48 @@ export class Messenger<
    * @param actionType - The action type. This is a unique identifier for this action.
    * @template ActionType - A type union of Action type strings.
    */
-  unregisterActionHandler<ActionType extends Action['type']>(
+
```
mcmire: Extra line? ✂️ 

``` suggestion
```

packages/messenger/src/Messenger.ts
mcmire: I guess if we make this change, we may have to clarify what "delegated" means for other methods that handle delegation, so there is that.

Alternatively, we could tell people to read the docs for `delegate` if they want more context. I know we mention this method in the `@deprecate` tag, but it might be missed.

packages/messenger/src/Messenger.ts
```diff
@@ -160,7 +196,42 @@ export class Messenger<
...
+   * @param handler - The action handler. This function gets called when the `call` method is
+   * invoked with the given action type.
+   * @throws Will throw when a handler has been registered for this action type already.
+   * @template ActionType - A type union of Action type strings.
+   */
+  registerDelegatedActionHandler<ActionType extends Action['type']>(
+    actionType: ActionType,
+    handler:
+      | ActionHandler<Action, ActionType>
+      | ActionHandler<ActionConstraint, ActionType>,
```
mcmire: This seems to allow any kind of action handler, but if `ActionType` must refer to an action that this messenger recognizes, then shouldn't the handler be connected to that action?

``` suggestion
    handler: ActionHandler<Action, ActionType>,
```

packages/messenger/src/Messenger.ts
```diff
@@ -277,6 +438,40 @@ export class Messenger<
...
+    if (!this.#isInCurrentNamespace(eventType)) {
+      throw new Error(
+        `Only allowed publishing events prefixed by '${this.#namespace}:'`,
+      );
+    }
+    this.#publish(eventType, ...payload);
+  }
+
+  /**
+   * Publish a delegated event.
```
mcmire: Similar as above — is it worth adding more language here?

``` suggestion
   * Publish a delegated event (that is, an event published by another messenger that this
   * messenger is permitted to subscribe to).
```

packages/messenger/src/Messenger.ts
```diff
@@ -419,62 +614,141 @@ export class Messenger<
...
-   * allowed to subscribe to.
-   * @template Namespace - The namespace for this messenger. Typically this is the name of the
-   * module that this messenger has been created for. The authority to publish events and register
-   * actions under this namespace is granted to this restricted messenger instance.
-   * @template AllowedAction - A type union of the 'type' string for any allowed actions.
-   * This must not include internal actions that are in the messenger's namespace.
-   * @template AllowedEvent - A type union of the 'type' string for any allowed events.
-   * This must not include internal events that are in the messenger's namespace.
-   * @returns The restricted messenger.
+   * Delegate actions and/or events to another messenger.
```
mcmire, line 667: Is it worth explaining what this does?

``` suggestion
   * Delegate actions and/or events to another messenger.
   *
   * - For all actions specified, this method will register an action handler
   * on the given messenger that calls the same action within this messenger.
   * - For all events specified, this method will register an event handler
   * on this messenger that publishes the same event within that messenger.
   *
   * In other words:
   *
   * - The delegated messenger is able to call any of the specified actions
   * known to this messenger as though they had been registered directly; when
   * they are called there, they are called here.
   * - The delegated messenger is able to subscribe to any of the specified
   * events known to this messenger; when they are published here they will
   * be published there as well.
   *
   * Note that to delegate an action to a messenger, it needs to be already
   * present in the type union of all Action types for that messenger; and
   * similarly for an event.
```

packages/messenger/src/Messenger.ts
```diff
@@ -419,62 +614,141 @@ export class Messenger<
...
+        );
+      }
+      delegatedEventSubscriptions.set(messenger, subscriber);
+      const getPayload = this.#initialEventPayloadGetters.get(eventType);
+      if (getPayload) {
+        messenger.registerDelegatedInitialEventPayload({
+          eventType,
+          getPayload,
+        });
+      }
```
mcmire: It seems like we are already calling `registerDelegatedInitialEventPayload` for all delegation targets within `#registerInitialEventPayload`. Do we need this?

``` suggestion
```

packages/messenger/CHANGELOG.md
```diff
@@ -10,5 +10,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add `delegate` and `revoke` methods ([#6132](https://github.com/MetaMask/core/pull/6132))
+  - These allow delegating or revoking capabilities (actions or events) from one `Messenger` instance to another.
+  - This allows passing capabilities through chains of messengers of arbitrary length
+  - See this ADR for details: https://github.com/MetaMask/decisions/blob/main/decisions/core/0012-messenger-delegation.md
+
+### Changed
+
+- **BREAKING:** Add `Namespace` type parameter and required `namespace` constructor parameter ([#6132](https://github.com/MetaMask/core/pull/6132))
+  - All published events and registered actions should fall under the given namespace. Typically the namespace is the controller or service name. This is the equivalent to the `Namespace` parameter from the old `RestrictedMessenger` class.
+- **BREAKING:** Remove `RestrictedMessenger` class ([#6132](https://github.com/MetaMask/core/pull/6132))
```
mcmire, line 27: Should this be in "Removed"?

packages/messenger/src/Messenger.ts
```diff
@@ -101,36 +99,64 @@ export type NotNamespacedBy<
...
+/**
+ * A messenger that actions and/or events can be delegated to.
+ *
+ * This is a minimal type interface to avoid complex incompatibilities resulting from generics.
+ */
+type DelegatedMessenger<
+  Action extends ActionConstraint,
+  Event extends EventConstraint,
+> = Pick<
+  Messenger<string, Action | ActionConstraint, Event | EventConstraint>,
```
mcmire: Similar question to https://github.com/MetaMask/core/pull/6132/files#r2226392393 — why do we need to allow any kind of action or event?

packages/messenger/src/Messenger.ts
Gudahtt: The idea was that these `delegated____` methods are internal-only (i.e. called only from within this class). I wasn't too concerned with making the intended usage clear, so long as we understand the intended usage. Everyone else just needs to understand _not_ to call this, ever.

Maybe there is something better than deprecating we can do to achieve this though, e.g. move it under a namespace or something.

mcmire: > The idea was that these `delegated____` methods are internal-only (i.e. called only from within this class). I wasn't too concerned with making the intended usage clear, so long as we understand the intended usage. Everyone else just needs to understand _not_ to call this, ever.

Yeah, that's fair.

> Maybe there is something better than deprecating we can do to achieve this though, e.g. move it under a namespace or something.

Maybe we could prefix the methods with something ugly like `__INTERNAL__` or `__PRIVATE__` or just `_`? But then again, VSCode will highlight the usage of the method as deprecated with an underline, so if you were to use it, I feel like you would know. Maybe the `@deprecated` is good enough.

packages/messenger/src/Messenger.ts
Gudahtt: Yes; this is for the case where the initial payload is registered before the delegation. My intent was to support both scenarios (see the `correctly registers initial event payload when delegated before payload is set` and `correctly registers initial event payload when delegated after payload is set` tests)

packages/messenger/src/Messenger.ts
Gudahtt: I'll try something different out, I wasn't totally happy with relying on just the method name prefix + `@deprecated` tag. I'll attempt to revise the description as well.

packages/messenger/src/Messenger.ts
mcmire: (Also, if any of that is wrong, feel free to correct me)

packages/messenger/src/Messenger.ts
mcmire: Nice!

packages/messenger/src/Messenger.ts
Gudahtt: Good question. I ran into problems with covariance/contravariance here (see https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)), but I can't recall the specifics. I'll investigate.

packages/messenger/src/Messenger.ts
Gudahtt: The idea was to keep this as the _sole_ way of registering an action handler, with the delegated one being an internal method.

I'll add something to clarify this. I am hesitant to add a qualifier to the description like in your suggestion though, as it might send the wrong message, and I'd like to keep the first sentence as simple as possible to convey the purpose

Gudahtt: Done here: https://github.com/MetaMask/core/pull/6132/commits/9e67b2bb47056fdec20d4b7d2c4d652844983880

packages/messenger/src/Messenger.ts
Gudahtt: Good catch! Fixed here: https://github.com/MetaMask/core/pull/6132/commits/a5cf4fa9934d0a78fe7f99c3fb954914b2643683

packages/messenger/CHANGELOG.md
Gudahtt: Definitely! Done here: https://github.com/MetaMask/core/pull/6132/commits/912daa46435a5fdeaafca5ade30375185bc974a5

packages/messenger/src/Messenger.ts
Gudahtt: I have renamed the methods and somewhat improved the descriptions here: d6a6222ba

Let me know what you think!

packages/messenger/src/Messenger.ts
Gudahtt: I'll resolve this since it's discussed more in depth in this other thread: https://github.com/MetaMask/core/pull/6132#discussion_r2258474529

packages/messenger/src/Messenger.ts
Gudahtt: Good point. I have expanded the description in this commit: 2c7be1fb4

I tried to keep it concise, but still getting at the main points you've raised here. Please let me know if something important still seems to be missing.

packages/messenger/src/Messenger.ts
Gudahtt: I had to widen the type because of contravarient types. Essentially, TypeScript was complaining in some situations because the messenger we were delegating to had _additional_ actions registered that the delegating messenger's type was not aware of. This doesn't present a problem for us in practice, but it did cause a type mismatch.

I added a comment to explain this.

Also I experimented with adding type tests in a separate branch: https://github.com/MetaMask/core/tree/demonstrate-type-error

Notice the two commits: first I narrowed this type, and wrote type tests to demonstrate the error. Then the next commit (the last one) makes the minimal changes necessary to fix the errors. Try experimenting with it locally, every single change in that last commit was necessary.

packages/messenger/src/Messenger.ts
Gudahtt: I was able to reproduce the errors that motivated this change. I've added a comment to explain, and I've experimented with type tests in a separate branch to demonstrate the problem and solution. See here for more details: https://github.com/MetaMask/core/pull/6132#discussion_r2258489334

packages/messenger/src/Messenger.ts
```diff
@@ -249,18 +331,52 @@ export class Messenger<
    * Registering a function for getting the payload allows event selectors to have a point of
    * comparison the first time state changes.
    *
+   * The event type must be under the same namespace as the messenger.
+   *
    * @param args - The arguments to this function
    * @param args.eventType - The event type to register a payload for.
    * @param args.getPayload - A function for retrieving the event payload.
+   * @template EventType - A type union of Event type strings under this messenger's namespace.
```
Gudahtt, line 384: I'll create a separate PR to add this, it was missing before.

Gudahtt: Done here: https://github.com/MetaMask/core/pull/6257

packages/messenger/src/Messenger.ts
```diff
@@ -212,7 +292,9 @@ export class Messenger<
    * This prevents all actions from being called.
    */
   clearActions() {
-    this.#actions.clear();
+    for (const actionType of this.#actions.keys()) {
+      this.#unregisterActionHandler(actionType);
+    }
```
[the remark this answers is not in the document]
Gudahtt: It's deleting the current key, that's it. This is perfectly predictable and safe to do with an iterator. Also this is pre-existing.

packages/messenger/src/Messenger.ts
Gudahtt: Rebased to resolve conflict on this line

packages/messenger/src/Messenger.ts
mcmire: Looks much better!

packages/messenger/src/Messenger.ts
mcmire: Agreed about keeping it concise. I am good with your changes!

packages/messenger/src/Messenger.ts
mcmire: Interesting! Thanks for making that branch and writing up those tests, that helped a lot. (`tstyche` is pretty cool, too.) I am not fully convinced that this is the best solution long-term, but it does make sense, and in the end I don't think it impacts type-safety.

(On a long-term solution: I am curious whether the fact that we divorce the type of an action from its handler throughout the entire Messenger class may be the real obstacle here. I feel like if we received and passed the `Action` or `Event` type as one unit then it could work better and TypeScript would be able to understand that we are comparing a subset of all actions and types that are involved. But I haven't tested that out yet, so we might end up in the same place.)

(On type-safety: One drawback from widening these types is that now you won't get a type error if you try to delegate an action or event to a messenger that doesn't actually support it. It's not a big deal because if you tried to use the delegated messenger to call the action or subscribe to the event later it wouldn't work anyway, but it's something to keep in mind.)

packages/messenger/src/Messenger.ts
mcmire: Thanks!

Doesn't `Action | ActionConstraint` resolve to `ActionConstraint`, since all `Action`s are also `ActionConstraint`s, and similar for `Event`?

I'm curious whether we could simplify this to:

``` typescript
type DelegatedMessenger = Pick<
  Messenger<string, ActionConstraint, EventConstraint>,
  | '_internalPublishDelegated'
  | '_internalRegisterDelegatedActionHandler'
  | '_internalRegisterDelegatedInitialEventPayload'
  | '_internalUnregisterDelegatedActionHandler'
>;

Or:

``` typescript
type DelegatedMessenger<
  Action extends ActionConstraint = ActionConstraint,
  Event extends EventConstraint = EventConstraint,
> = Pick<
  Messenger<string, Action, Event>,
  | '_internalPublishDelegated'
  | '_internalRegisterDelegatedActionHandler'
  | '_internalRegisterDelegatedInitialEventPayload'
  | '_internalUnregisterDelegatedActionHandler'
>;
```

If you don't want to mess with this then that's fine too :)

mcmire (dismissed):
I had one more comment, but it is non-blocking.

LGTM!

packages/messenger/src/Messenger.ts
Gudahtt: >One drawback from widening these types is that now you won't get a type error if you try to delegate an action or event to a messenger that doesn't actually support it

Hmm. Good point. This does mean `delegate` isn't really type-safe. I will revisit this. If we can't get type safety here, this could at least be simplified. 

>I am curious whether the fact that we divorce the type of an action from its handler throughout the entire Messenger class may be the real obstacle here. I feel like if we received and passed the Action or Event type as one unit then it could work better and TypeScript would be able to understand that we are comparing a subset of all actions and types that are involved

Could you elaborate on what you had in mind here? I'm not sure I follow what you mean when you talk about passing them as "one unit". Are you talking about how we're using the type union, or how we're referencing individual properties of an action or event?

The problem as I see it is that "comparing a subset" fundamentally doesn't work with contravarient types. But maybe I'm missing something.

mcmire: The type for `registerActionHandler`, `publish`, `subscribe`, `call`, `_internalRegisterDelegatedActionHandler`, etc. take just the name of an action or event and then derive an action/event handler type from it if necessary. I was suggesting that if it is possible for these methods to take the entire action or event object type then perhaps TypeScript could compare the subset of actions and events more easily between the `DelegatedMessenger` type and the `Messenger` type. But perhaps you are right; at the end of the day it would probably still have to compare function types, so perhaps it wouldn't matter.

Gudahtt:
I still need to address some of these problems highlighted by Cursor, but I'm marking this as ready for review because the earlier issues with type safety have been resolved

packages/messenger/src/Messenger.ts
Gudahtt: Agreed! This has been simplified

packages/messenger/src/Messenger.ts
Gudahtt: I managed to find a way to type check the `delegate` and `revoke` parameters using a conditional type. The `DelegatedMesseger` type is still broadened for internal references past that point, but at least this type safety issue should be resolved. I've pushed updated tests to this branch as well: https://github.com/MetaMask/core/tree/demonstrate-type-error

packages/messenger/src/Messenger.ts
```diff
@@ -422,62 +552,251 @@ export class Messenger<
...
+        // Nothing to revoke
+        continue;
+      }
+      this.unsubscribe(eventType, delegatedSubscriber);
+      delegationTargets.delete(messenger);
+      if (delegationTargets.size === 0) {
+        this.#subscriptionDelegationTargets.delete(eventType);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Gudahtt: Resolved here: https://github.com/MetaMask/core/pull/6132/commits/b064fad9e6886980860515a5496f821a563b2c3f

packages/messenger/src/Messenger.ts
```diff
@@ -422,62 +552,251 @@
...
+          eventType,
+          getPayload,
+        });
+      }
+      this.subscribe(eventType, subscriber);
+    }
+  }
+
+  /**
+   * Revoke delegated actions and/or events from another messenger.
```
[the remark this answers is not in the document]
Gudahtt: Resolved here: https://github.com/MetaMask/core/pull/6132/commits/240680815508ae805f88b75da884cc971a057ef6

Gudahtt:
Rebased to resolve conflicts. All feedback has now been addressed

packages/messenger/src/Messenger.ts
```diff
@@ -447,83 +624,329 @@ export class Messenger<
...
+    for (const [handler, metadata] of subscriptions.entries()) {
+      if (metadata.delegation) {
+        continue;
+      }
+      subscriptions.delete(handler);
+    }
+
+    if (subscriptions.size === 0) {
+      this.#events.delete(eventType);
+    }
```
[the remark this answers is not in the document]
Gudahtt: Deleting an element from a collection being iterated over is not unpredictable behavior, this is fine.

packages/messenger/src/Messenger.ts
Gudahtt: I had to refactor it again. The old branch I used with the type tests wasn't easy to rebase onto this implementation because the way I produced an error was no longer applicable, so I squashed the tests together and moved them to this new branch to demonstrate that the types still work:

https://github.com/MetaMask/core/tree/add-messenger-type-tests

packages/messenger/src/Messenger.ts
```diff
@@ -447,83 +625,324 @@ export class Messenger<
...
     eventType: EventType,
   ) {
-    this.#events.delete(eventType);
+    const subscriptions = this.#events.get(eventType);
+    if (!subscriptions) {
+      return;
+    }
+
+    for (const [handler, metadata] of subscriptions.entries()) {
+      if (metadata.delegation) {
```
mcmire, line 643: I don't know if this logic was already there and I missed it before or if this is a recent change to this PR, but either way, good catch.

mcmire (approved):
This looks so good now. Fantastic work!

packages/messenger/src/Messenger.ts
Gudahtt: This was the fix one of the last two bugs that Cursor found, it's a new addition from last time you reviewed.

packages/messenger/src/Messenger.ts
mcmire: Out of curiosity, did you plan on merging these type tests? Or were they just for exploring a path forward? The `tstyche` package you found seems very useful at least.

Gudahtt: I wanted to test whether vite was a better alternative or not first. And an obstacle to that would be figuring out whether vite itself is a suitable replacement for Jest or not in this repo.

I guess we could add tstyche directly in the short term at least, maybe I'll propose that instead

PR #6136, opened 2025-07-17
Fix/revoke token validation

himanshuchawla009:
## Explanation
- This PR changes the usage of revokeToken from mandatory to optional in vault.
- The reason why its done because revokeToken is not accessible in a device if max key chain length exceeds. Although this condition should never happen for a general user. Making it optional will prevent user from getting stuck in dead end login situation.

## Checklist


himanshuchawla009:
@metamaskbot publish-preview

himanshuchawla009:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - remove `access_token` validation when the wallet is locked. ([#6133](https://github.com/MetaMask/core/pull/6133))
+- remove `revoke_token` validation from `#parseVault` and `createNewVaultWithAuthData` to handle the case when
+  max key chain length exceeds. ([#6136](https://github.com/MetaMask/core/pull/6136))
```
matthiasgeihs: nit: We should start changelog entries with a capital letter, just like in normal English.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -452,7 +452,7 @@ async function createMockVault(
       sk: `0x${authKeyPair.sk.toString(16)}`,
       pk: bytesToBase64(authKeyPair.pk),
     }),
-    revokeToken: mockRevokeToken,
+    revokeToken: mockRevokeToken || undefined,
```
matthiasgeihs: i don't get this. if the string is empty, you set undefined? why?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -4946,4 +4913,95 @@ describe('SeedlessOnboardingController', () => {
...
+        mockToprfEncryptor.deriveAuthKeyPair(MOCK_PASSWORD);
+
+      // Create vault without revokeToken
+      const { encryptedMockVault, vaultEncryptionKey, vaultEncryptionSalt } =
+        await createMockVault(
+          MOCK_ENCRYPTION_KEY,
+          MOCK_PW_ENCRYPTION_KEY,
+          MOCK_AUTH_KEY_PAIR,
+          MOCK_PASSWORD,
+          '', // No revokeToken
```
matthiasgeihs: use undefined here, instead of using `revokeToken || undefined` above.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1755,6 +1723,12 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
       toprfAuthKeyPair,
     } = await this.#unlockVaultAndGetBackupEncKey();
 
+    // revoke token can be undefined if the max key chain length is reached.
+    // TODO: remove this once we have better solution to handle max key chain length.
+    if (!revokeToken) {
+      return;
+    }
```
matthiasgeihs: we can't do this!
this is a public method called `revokeRefreshToken` and the caller's expectation is that the refresh token is revoked when this method returns successfully.

There is two internal locations where we call `revokeRefreshToken`: `submitPassword`, `submitGlobalPassword`. Afaik, this method is not called publicly yet.

We can either check before calling if `revokeToken` is present. Or we can rename this function to `tryRevokeRefreshToken` and return a boolean that signals if the revocation was successful. We can also make it private to ensure that it's really not called externally, if there is no use case for it currently.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -4946,4 +4913,95 @@ describe('SeedlessOnboardingController', () => {
...
+      await withController(
+        {
+          state,
+        },
+        async ({ controller, mockRevokeRefreshToken }) => {
+          await controller.submitPassword(MOCK_PASSWORD);
+
+          // Verify the method doesn't throw and returns without calling the revoke function
+          expect(
+            await controller.revokeRefreshToken(MOCK_PASSWORD),
```
matthiasgeihs: this should fail if the revokeToken is not present

packages/seedless-onboarding-controller/src/assertions.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (
+    !value || // value is not defined
+    typeof value !== 'object' || // value is not an object
+    !('toprfEncryptionKey' in value) || // toprfEncryptionKey is not defined
+    typeof value.toprfEncryptionKey !== 'string' || // toprfEncryptionKey is not a string
+    !('toprfPwEncryptionKey' in value) || // toprfPwEncryptionKey is not defined
+    typeof value.toprfPwEncryptionKey !== 'string' || // toprfPwEncryptionKey is not a string
+    !('toprfAuthKeyPair' in value) || // toprfAuthKeyPair is not defined
+    typeof value.toprfAuthKeyPair !== 'string' || // toprfAuthKeyPair is not a string
+    // throw error if revoke token exist but is not a string and is not undefined
```
matthiasgeihs: nit
```suggestion
    // revoke token exists but is not a string and is not undefined
```

packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -334,8 +334,9 @@ export type VaultData = {
   toprfAuthKeyPair: string;
   /**
    * The revoke token to revoke refresh token and get new refresh token and new revoke token.
+   * Revoke is only available when max password chain length doesn't exceeds on a device.
```
matthiasgeihs: This explanation is not easy to understand without context.
It's probably leading to more confusion than clarity and should be removed.

matthiasgeihs: PS: If we still wanna keep it, at least we should revise the comment.
```suggestion
   * The revoke token may no longer be available after a large number of password changes. In this case, re-authentication is advised.
```

tuna1207 (approved):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
himanshuchawla009: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -754,10 +767,24 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+
+      if (revokeToken) {
+        // revoke and recyle refresh token after unlock to keep refresh token fresh, avoid malicious use of leaked refresh token
+        await this.revokeRefreshToken(revokeToken);
+        await this.#createNewVaultWithAuthData({
+          password: globalPassword,
+          rawToprfEncryptionKey: toprfEncryptionKey,
+          rawToprfPwEncryptionKey: toprfPwEncryptionKey,
+          rawToprfAuthKeyPair: toprfAuthKeyPair,
+        });
```
matthiasgeihs, line 788: ideally we shouldn't change the vault on unlock. the common expectation is that a `submit` functions unlocks the vault, but does not change it.

i assume you do it here because you want to persist the fresh tokens, but this is quite non-obvious.
why don't you move this back into `revokeRefreshToken`? then it's located next to the state update. still probably worth adding a comment `// recreating vault to persist fresh tokens`.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
himanshuchawla009: fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1740,20 +1735,12 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
    * Revoke the refresh token and get new refresh token and new revoke token.
-   * This method is to be called after unlock
+   * This method is to be called after user is authenticated.
    *
-   * @param password - The password to re-encrypt new token in the vault.
+   * @param revokeToken - The revoke token to use for revoking the refresh token.
    */
-  async revokeRefreshToken(password: string) {
-    this.#assertIsUnlocked();
+  async revokeRefreshToken(revokeToken: string) {
```
matthiasgeihs: should we make this function private?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -631,10 +631,23 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+        toprfAuthKeyPair,
+        revokeToken,
+      } = await this.#unlockVaultAndGetBackupEncKey(password);
       this.#setUnlocked();
-      // revoke and recyle refresh token after unlock to keep refresh token fresh, avoid malicious use of leaked refresh token
-      await this.revokeRefreshToken(password);
+
+      if (revokeToken) {
+        await this.revokeRefreshToken(revokeToken);
+        await this.#createNewVaultWithAuthData({
```
matthiasgeihs, line 645: same comment [as below](https://github.com/MetaMask/core/pull/6136/files#r2213531932).
i think at least a comment why you recreate the vault here would be good. or move vault creation back into `revokeRefreshToken`.

himanshuchawla009: added a comment for now.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -631,10 +631,24 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+        toprfAuthKeyPair,
+        revokeToken,
+      } = await this.#unlockVaultAndGetBackupEncKey(password);
       this.#setUnlocked();
-      // revoke and recyle refresh token after unlock to keep refresh token fresh, avoid malicious use of leaked refresh token
-      await this.revokeRefreshToken(password);
+
+      if (revokeToken) {
+        await this.#revokeRefreshTokenAndUpdateState(revokeToken);
+        // re-creating with to persist the new revoke token
```
matthiasgeihs: nit
```suggestion
        // re-creating vault to persist the new revoke token
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -754,10 +768,25 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+        toprfAuthKeyPair,
+      } = await this.#unlockVaultAndGetBackupEncKey(undefined, vaultKey);
       this.#setUnlocked();
-      // revoke and recyle refresh token after unlock to keep refresh token fresh, avoid malicious use of leaked refresh token
-      await this.revokeRefreshToken(globalPassword);
+
+      if (revokeToken) {
+        // revoke and recyle refresh token after unlock to keep refresh token fresh, avoid malicious use of leaked refresh token
+        await this.#revokeRefreshTokenAndUpdateState(revokeToken);
+        // re-creating with to persist the new revoke token
```
matthiasgeihs: nit
```suggestion
        // re-creating vault to persist the new revoke token
```

matthiasgeihs (dismissed):

matthiasgeihs (dismissed):

matthiasgeihs (approved):

PR #6137, opened 2025-07-17
Create shield controller

matthiasgeihs:
## Explanation

New controller for MetaMask Shield.

## Checklist

### DIFF
--- a/packages/shield-controller/package.json
+++ b/packages/shield-controller/package.json
@@ -1,0 +1,59 @@
+{
+  "name": "@metamask/shield-controller",
+  "version": "0.0.0",
+  "description": "Controller handling shield transaction coverage logic.",
+  "keywords": [
+    "MetaMask",
+    "Ethereum"
+  ],
+  "homepage": "https://github.com/MetaMask/core/tree/main/packages/shield-controller#readme",
+  "bugs": {
+    "url": "https://github.com/MetaMask/core/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MetaMask/core.git"
+  },
+  "license": "MIT",
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js",
+      "types": "./dist/types/index.d.ts"
+    },
+    "./package.json": "./package.json"
+  },
+  "main": "./dist/index.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "dist/"
+  ],
+  "scripts": {
+    "build": "ts-bridge --project tsconfig.build.json --verbose --clean --no-references",
+    "build:docs": "typedoc",
+    "changelog:validate": "../../scripts/validate-changelog.sh @metamask/shield-controller",
+    "publish:preview": "yarn npm publish --tag preview",
+    "test": "jest --reporters=jest-silent-reporter",
+    "test:clean": "jest --clearCache",
+    "test:verbose": "jest --verbose",
+    "test:watch": "jest --watch"
+  },
+  "devDependencies": {
+    "@metamask/auto-changelog": "^3.4.4",
+    "@types/jest": "^27.4.1",
+    "deepmerge": "^4.2.2",
+    "jest": "^27.5.1",
+    "ts-jest": "^27.1.4",
+    "typedoc": "^0.24.8",
+    "typedoc-plugin-missing-exports": "^2.0.0",
+    "typescript": "~5.2.2"
+  },
+  "engines": {
+    "node": "^18.18 || >=20"
+  },
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://registry.npmjs.org/"
+  }
+}
--- a/packages/shield-controller/package.json
+++ b/packages/shield-controller/package.json
@@ -39,7 +39,11 @@
     "test:verbose": "jest --verbose",
     "test:watch": "jest --watch"
   },
+  "dependencies": {
+    "@metamask/base-controller": "^8.0.1"
+  },
   "devDependencies": {
+    "@lavamoat/allow-scripts": "^3.3.4",
     "@metamask/auto-changelog": "^3.4.4",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
@@ -55,5 +59,10 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "lavamoat": {
+    "allowScripts": {
+      "@lavamoat/allow-scripts>@lavamoat/preinstall-always-fail": false
+    }
   }
 }
--- a/packages/shield-controller/src/ShieldController.ts
+++ b/packages/shield-controller/src/ShieldController.ts
@@ -1,0 +1,150 @@
+import { BaseController } from '@metamask/base-controller';
+import type { RestrictedMessenger } from '@metamask/base-controller';
+import type {
+  TransactionControllerUnapprovedTransactionAddedEvent,
+  TransactionMeta,
+  TransactionParams,
+} from '@metamask/transaction-controller';
+
+export type Transaction = {
+  from: string;
+  to: string;
+  value: string;
+  data: string;
+  nonce: string;
+};
+
+export type CoverageStatus = 'success' | 'error';
+
+export type CoverageResult = {
+  txId: string;
+  status: CoverageStatus;
+};
+
+export type ShieldControllerState = {
+  coverageResults: CoverageResult[];
+};
+
+/**
+ * Get the default state for the ShieldController.
+ *
+ * @returns The default state for the ShieldController.
+ */
+function getDefaultShieldControllerState(): ShieldControllerState {
+  return {
+    coverageResults: [],
+  };
+}
+
+/**
+ * The name of the {@link ShieldController}.
+ */
+const controllerName = 'ShieldController';
+
+export type ShieldControllerCheckCoverageAction = {
+  type: `${typeof controllerName}:checkCoverage`;
+  handler: ShieldController['checkCoverage'];
+};
+
+/**
+ * The internal actions available to the ShieldController.
+ */
+export type ShieldControllerActions = ShieldControllerCheckCoverageAction;
+
+export type ShieldControllerCoverageResultReceivedEvent = {
+  type: `${typeof controllerName}:coverageResultReceived`;
+  payload: [coverageResult: CoverageResult];
+};
+
+/**
+ * The internal events available to the ShieldController.
+ */
+export type ShieldControllerEvents =
+  ShieldControllerCoverageResultReceivedEvent;
+
+/**
+ * The external actions available to the ShieldController.
+ */
+export type AllowedActions = never;
+
+/**
+ * The external events available to the ShieldController.
+ */
+export type AllowedEvents =
+  TransactionControllerUnapprovedTransactionAddedEvent;
+
+/**
+ * The messenger of the {@link ShieldController}.
+ */
+export type ShieldControllerMessenger = RestrictedMessenger<
+  typeof controllerName,
+  ShieldControllerActions | AllowedActions,
+  ShieldControllerEvents | AllowedEvents,
+  AllowedActions['type'],
+  AllowedEvents['type']
+>;
+
+/**
+ * Metadata for the ShieldController state, describing how to "anonymize"
+ * the state and which parts should be persisted.
+ */
+const metadata = {
+  coverageResults: {
+    persist: true,
+    anonymous: false,
+  },
+};
+
+export type ShieldControllerOptions = {
+  messenger: ShieldControllerMessenger;
+  state?: ShieldControllerState;
+};
+
+export class ShieldController extends BaseController<
+  typeof controllerName,
+  ShieldControllerState,
+  ShieldControllerMessenger
+> {
+  constructor(options: ShieldControllerOptions) {
+    const { messenger, state } = options;
+    super({
+      name: controllerName,
+      metadata,
+      messenger,
+      state: {
+        ...getDefaultShieldControllerState(),
+        ...state,
+      },
+    });
+  }
+
+  start() {
+    this.messagingSystem.subscribe(
+      'TransactionController:unapprovedTransactionAdded',
+      this.handleUnapprovedTransactionAdded.bind(this),
+    );
+  }
+
+  async handleUnapprovedTransactionAdded(transactionMeta: TransactionMeta) {
+    await this.checkCoverage(transactionMeta.txParams);
+  }
+
+  async checkCoverage(txParams: TransactionParams): Promise<CoverageResult> {
+    // Check subscription status
+    const subscriptionStatus = await this.checkSubscriptionStatus();
+    if (subscriptionStatus !== 'subscribed') {
+      throw new Error('Not subscribed');
+    }
+
+    // Check coverage
+    const coverageResult = await this.fetchCoverageResult(txParams);
+
+    // Publish coverage result
+    this.messagingSystem.publish(
+      `${controllerName}:coverageResultReceived`,
+      coverageResult,
+    );
+
+    return coverageResult;
+  }
+}
--- a/packages/shield-controller/src/index.ts
+++ b/packages/shield-controller/src/index.ts
@@ -1,0 +1,9 @@
+/**
+ * Example function that returns a greeting for the given name.
+ *
+ * @param name - The name to greet.
+ * @returns The greeting.
+ */
+export default function greeter(name: string): string {
+  return `Hello, ${name}!`;
+}
--- a/packages/shield-controller/src/index.ts
+++ b/packages/shield-controller/src/index.ts
@@ -1,9 +1,1 @@
-/**
- * Example function that returns a greeting for the given name.
- *
- * @param name - The name to greet.
- * @returns The greeting.
- */
-export default function greeter(name: string): string {
-  return `Hello, ${name}!`;
-}
+export { ShieldController } from './ShieldController';
--- a/scripts/create-package/cli.ts
+++ b/scripts/create-package/cli.ts
@@ -20,8 +20,6 @@ export default async function cli(
     // Disable --version. This is an internal tool and it doesn't have a version.
     .version(false)
     .usage('$0 [args]')
-    // @ts-expect-error: The CommandModule<T, U>[] signature does in fact exist,
-    // but it is missing from our yargs types.
     .command(commands)
     .strict()
     .check((args) => {
@@ -43,6 +41,5 @@ export default async function cli(
     .showHelpOnFail(false)
     .help()
     .alias('help', 'h')
-    // @ts-expect-error: This does in fact exist, but it is missing from our yargs types.
     .parseAsync();
 }
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Each package in this repository has its own README where you can find installati
 - [`@metamask/sample-controllers`](packages/sample-controllers)
 - [`@metamask/seedless-onboarding-controller`](packages/seedless-onboarding-controller)
 - [`@metamask/selected-network-controller`](packages/selected-network-controller)
+- [`@metamask/shield-controller`](packages/shield-controller)
 - [`@metamask/signature-controller`](packages/signature-controller)
 - [`@metamask/token-search-discovery-controller`](packages/token-search-discovery-controller)
 - [`@metamask/transaction-controller`](packages/transaction-controller)
@@ -121,6 +122,7 @@ linkStyle default opacity:0.5
   sample_controllers(["@metamask/sample-controllers"]);
   seedless_onboarding_controller(["@metamask/seedless-onboarding-controller"]);
   selected_network_controller(["@metamask/selected-network-controller"]);
+  shield_controller(["@metamask/shield-controller"]);
   signature_controller(["@metamask/signature-controller"]);
   token_search_discovery_controller(["@metamask/token-search-discovery-controller"]);
   transaction_controller(["@metamask/transaction-controller"]);
@@ -162,7 +164,6 @@ linkStyle default opacity:0.5
   bridge_status_controller --> base_controller;
   bridge_status_controller --> controller_utils;
   bridge_status_controller --> polling_controller;
-  bridge_status_controller --> user_operation_controller;
   bridge_status_controller --> accounts_controller;
   bridge_status_controller --> bridge_controller;
   bridge_status_controller --> gas_fee_controller;
@@ -243,7 +244,6 @@ linkStyle default opacity:0.5
   profile_sync_controller --> base_controller;
   profile_sync_controller --> accounts_controller;
   profile_sync_controller --> keyring_controller;
-  profile_sync_controller --> network_controller;
   rate_limit_controller --> base_controller;
   remote_feature_flag_controller --> base_controller;
   remote_feature_flag_controller --> controller_utils;
--- a/packages/shield-controller/CHANGELOG.md
+++ b/packages/shield-controller/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+[Unreleased]: https://github.com/MetaMask/core/
--- a/packages/shield-controller/LICENSE
+++ b/packages/shield-controller/LICENSE
@@ -1,0 +1,20 @@
+MIT License
+
+Copyright (c) 2025 MetaMask
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
--- a/packages/shield-controller/README.md
+++ b/packages/shield-controller/README.md
@@ -1,0 +1,15 @@
+# `@metamask/shield-controller`
+
+Controller handling shield transaction coverage logic.
+
+## Installation
+
+`yarn add @metamask/shield-controller`
+
+or
+
+`npm install @metamask/shield-controller`
+
+## Contributing
+
+This package is part of a monorepo. Instructions for contributing can be found in the [monorepo README](https://github.com/MetaMask/core#readme).
--- a/packages/shield-controller/jest.config.js
+++ b/packages/shield-controller/jest.config.js
@@ -1,0 +1,26 @@
+/*
+ * For a detailed explanation regarding each configuration property and type check, visit:
+ * https://jestjs.io/docs/configuration
+ */
+
+const merge = require('deepmerge');
+const path = require('path');
+
+const baseConfig = require('../../jest.config.packages');
+
+const displayName = path.basename(__dirname);
+
+module.exports = merge(baseConfig, {
+  // The display name when running multiple projects
+  displayName,
+
+  // An object that configures minimum threshold enforcement for coverage results
+  coverageThreshold: {
+    global: {
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
+    },
+  },
+});
--- a/packages/shield-controller/src/index.test.ts
+++ b/packages/shield-controller/src/index.test.ts
@@ -1,0 +1,9 @@
+import greeter from '.';
+
+describe('Test', () => {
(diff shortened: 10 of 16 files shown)


packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export type ShieldControllerCheckCoverageAction = {
+  type: `${typeof controllerName}:checkCoverage`;
+  handler: ShieldController['checkCoverage'];
+};
+
+/**
+ * The internal actions available to the ShieldController.
+ */
+export type ShieldControllerActions = ShieldControllerCheckCoverageAction;
```
lwin-kyaw, line 49: Would be nice if we put all the types in separate file, e.g. `types.ts`

scripts/create-package/cli.ts
```diff
@@ -20,8 +20,6 @@ export default async function cli(
     // Disable --version. This is an internal tool and it doesn't have a version.
     .version(false)
     .usage('$0 [args]')
-    // @ts-expect-error: The CommandModule<T, U>[] signature does in fact exist,
```
lwin-kyaw: May I know why do we need to update this?
If it's not working, I think we should do it in separate PR :)

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: moved some types, can potentially move more

scripts/create-package/cli.ts
matthiasgeihs: it wasn't working when i was running the command. but yeah, should probably do that separately.

matthiasgeihs: I think this was fixed in the meantime. this change is no longer part of this PR after rebasing.

packages/shield-controller/src/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { TransactionMeta } from '@metamask/transaction-controller';
+
+export type CoverageResult = {
+  status: CoverageStatus;
+};
+
+export type CoverageStatus = 'covered' | 'malicious' | 'unsupported';
```
chaitanyapotti: rename unsupported to unknown

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,217 @@
...
+      previousTransactions?.map((tx) => [tx.id, tx]) ?? [],
+    );
+    for (const transaction of transactions) {
+      const previousTransaction = previousTransactionsById.get(transaction.id);
+
+      // Check coverage if the transaction is new or if the simulation data has
+      // changed.
+      if (
+        !previousTransaction ||
+        previousTransaction.simulationData !== transaction.simulationData
```
chaitanyapotti, line 184: pls add a comment reg. this logic

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,117 @@
...
+export const BASE_URL = 'https://rule-engine.metamask.io';
+
+export type InitCoverageCheckRequest = {
+  txParams: {
+    from: string;
+    to?: string;
+    value?: string;
+    data?: string;
+    nonce?: string;
+    chainId?: string;
```
chaitanyapotti: sync on this

chaitanyapotti: research on batch tx atomic

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,117 @@
...
+      coverageId,
+    };
+
+    const startTime = Date.now();
+    while (Date.now() - startTime < timeout) {
+      const res = await fetch(`${BASE_URL}/api/v1/coverage/result`, {
+        method: 'POST',
+        headers: await this.#createHeaders(),
+        body: JSON.stringify(reqBody),
+      });
```
chaitanyapotti: need a per req n global timeout

packages/shield-controller/tests/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  };
+}
+
+/**
+ * Get a random coverage status.
+ *
+ * @returns A random coverage status.
+ */
+export function getRandomCoverageStatus(): CoverageStatus {
+  return ['covered', 'malicious', 'unsupported'][
```
chaitanyapotti: change code here

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    // Update state
+    if (coverageResultEntry.results.length >= this.#coverageHistoryLimit) {
+      coverageResultEntry.results.pop();
+    }
+    coverageResultEntry.results.unshift(coverageResult);
+
+    // Write state
+    this.update((draft) => {
+      draft.coverageResults[txId] = coverageResultEntry;
```
chaitanyapotti: check if tx controller is limiting the no. of transactions

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,117 @@
+import type { TransactionMeta } from '@metamask/transaction-controller';
+
+import type { CoverageResult, CoverageStatus, ShieldBackend } from './types';
+
+export const BASE_URL = 'https://rule-engine.metamask.io';
```
chaitanyapotti, line 5: endpoint should be injected in controller config

packages/shield-controller/src/types.ts
matthiasgeihs: done

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: Done. Note that (while shallow equality is sufficient to check for changes) this does only trigger if simulation data has changed. That is, if a simulation didn't change any of the simulation data, this will not trigger.
```ts
// Checking reference equality is sufficient because this object if the
// simulation data has changed.
previousTransaction.simulationData !== transaction.simulationData
```

packages/shield-controller/src/backend.ts
matthiasgeihs: Synced and updated type.

Batch transactions are handled separately by the TransactionController using `addTransactionBatch` and `state.transactionBatches`. Transaction batches (`TransactionBatchMeta`) have a different type than normal transactions (`TransactionMeta`).

packages/shield-controller/src/backend.ts
matthiasgeihs: done, please check

packages/shield-controller/tests/utils.ts
matthiasgeihs: done

packages/shield-controller/src/backend.ts
matthiasgeihs: TranasctionController is doing it like this for Sentinel URL, but will change it. I also think that it's good to have it configurable.

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: Checked once again how the TransactionController does it:
- It trims the total number of transactions if it goes above a certain number: https://github.com/MetaMask/core/blob/258c03b00d6ce5ee63845454f86ea404c48dbde8/packages/transaction-controller/src/TransactionController.ts#L3178
- It compress the transaction history if it gets too large: https://github.com/MetaMask/core/blob/258c03b00d6ce5ee63845454f86ea404c48dbde8/packages/transaction-controller/src/utils/history.ts#L94

You requested that we should purge state like the TransactionController does it:
- For the overall number of transactions: We currently do not trim the overall amount of transactions. I will change that so that old transactions are discarded whenever their number reaches a certain configurable limit.
- For the coverage history: We are currently [simply discarding](https://github.com/MetaMask/core/blob/258c03b00d6ce5ee63845454f86ea404c48dbde8/packages/shield-controller/src/ShieldController.ts#L209) old items from the coverage history if the number of entries goes beyond a certain configurable threshold. I'm not sure if we can compress the coverage history in a similar way to how the TransactionController compresses the transaction history.

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+          const startTime = Date.now();
+          const res = await fetch(`${this.#baseUrl}/api/v1/coverage/result`, {
+            method: 'POST',
+            headers,
+            body: JSON.stringify(reqBody),
+          });
+          if (res.status === 200) {
+            return (await res.json()) as GetCoverageResultResponse;
+          }
+          await sleep(pollInterval - (Date.now() - startTime));
```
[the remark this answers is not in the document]
matthiasgeihs: This is intended

packages/shield-controller/src/ShieldController.ts
chaitanyapotti: Our current approach for coverage history is fine

packages/shield-controller/src/types.ts
```diff
@@ -4,7 +4,8 @@ export type CoverageResult = {
   status: CoverageStatus;
 };
 
-export type CoverageStatus = 'covered' | 'malicious' | 'unsupported';
+export const coverageStatuses = ['covered', 'malicious', 'unknown'] as const;
+export type CoverageStatus = (typeof coverageStatuses)[number];
```
chaitanyapotti, line 8: `(typeof coverageStatuses)[keyof typeof coverageStatuses];`

matthiasgeihs: Why, if I do that
```ts
export type CoverageStatus =
  (typeof coverageStatuses)[keyof typeof coverageStatuses];
```
, the type resolves to
```ts
type CoverageStatus = "covered" | "malicious" | "unknown" | 3 | (() => ArrayIterator<"covered" | "malicious" | "unknown">) | {
    readonly [x: number]: boolean | undefined;
    readonly length?: boolean | undefined;
    ... 24 more ...;
    at?: boolean | undefined;
} | ... 22 more ... | ((index: number) => "covered" | ... 2 more ... | undefined)
```

while with
```ts
export type CoverageStatus = (typeof coverageStatuses)[number];
```
the type resolves to
```ts
type CoverageStatus = "covered" | "malicious" | "unknown"
```
which looks more like what we want

chaitanyapotti: Oh.. i see the problem. We usually do 
```ts
export const COVERAGE_STATUS = {
COVERED: 'covered', 
MALICIOUS: 'malicious', 
UNKNOWN: 'unknown'
} as const;
export type COVERAGE_STATUS_TYPE = (typeof COVERAGE_STATUS)[keyof typeof COVERAGE_STATUS];
```

so that while using in code we can write 
```ts
status: COVERAGE_STATUS.COVERED
```

matthiasgeihs: I see. However, I don't see the benefit. The main thing I want here is an array of coverageStatus values.
Now I need to write `Object.values(COVERAGE_STATUS)` to get this, instead of just `coverageStatuses`.

chaitanyapotti (approved):

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,259 @@
...
+      previousTransactions?.map((tx) => [tx.id, tx]) ?? [],
+    );
+    for (const transaction of transactions) {
+      const previousTransaction = previousTransactionsById.get(transaction.id);
+
+      // Check coverage if the transaction is new or if the simulation data has
+      // changed.
+      if (
+        !previousTransaction ||
+        // Checking reference equality is sufficient because this object if the
```
mcmire: Nit: Typo in this comment here? Not sure what this should say.

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,259 @@
...
+      ) {
+        this.checkCoverage(transaction).catch(
+          // istanbul ignore next
+          (error) => log('Error checking coverage:', error),
+        );
+      }
+    }
+  }
+
+  async checkCoverage(txMeta: TransactionMeta): Promise<CoverageResult> {
```
mcmire, line 200: Nit: Thoughts on providing JSDoc for this method to help others know what this does and what it's useful for?

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,259 @@
...
+    // Read state
+    let newEntry = false;
+    let coverageResultEntry = this.state.coverageResults[txId];
+    if (!coverageResultEntry) {
+      newEntry = true;
+      coverageResultEntry = {
+        results: [],
+      };
+    } else {
+      // Clone object to avoid mutation
```
mcmire: Nit: Not a big deal, but if you put this logic in the `this.update` call below you don't need to clone objects and arrays when updating them. We use the [Immer](https://immerjs.github.io/immer/#a-quick-example-for-comparison) library internally and it will allow you to treat the state object as though it's mutable, but under the hood it will automatically and intelligently clone objects as it goes.

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+        timeoutReached = true;
+        reject(new Error('Timeout waiting for coverage result'));
+      }, timeout);
+
+      const poll = async (): Promise<GetCoverageResultResponse> => {
+        // The timeoutReached variable is modified in the timeout callback.
+        // eslint-disable-next-line no-unmodified-loop-condition
+        while (!timeoutReached) {
+          const startTime = Date.now();
+          const res = await fetch(`${this.#baseUrl}/api/v1/coverage/result`, {
```
mcmire: This file seems to rely on the global `fetch` function. We've had problems with this in the past across clients, and we've found it's better to pass in the `fetch` function so that clients can pass whatever implementation is specific to the platform. Thoughts on having your `ShieldRemoteBackend` class take a `fetch` option? (Alternatively, the "data service" pattern I mention in an earlier comment already does this.)

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const headers = await this.#createHeaders();
+    return await new Promise((resolve, reject) => {
+      let timeoutReached = false;
+      setTimeout(() => {
+        timeoutReached = true;
+        reject(new Error('Timeout waiting for coverage result'));
+      }, timeout);
+
+      const poll = async (): Promise<GetCoverageResultResponse> => {
```
mcmire, line 117: It looks like you are implementing an automatic retry mechanism in this class. Are you familiar with the [Cockatiel](https://github.com/connor4312/cockatiel) library? This implements a range of different strategies for gracefully handling failures when interacting with external systems via patterns that they call policies.

We've developed a "data service" pattern that makes use of two policies from this library: one that automatically retries when there are failures and another that uses the circuit breaker policy to automatically pause requests when the API is perceived to be down.  that makes use of this utility. What are your thoughts on refactoring this file to follow this pattern? You can see an example here: https://github.com/MetaMask/core/blob/b3f97967a33681a44880ea845f03128137f9e505/packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts

packages/shield-controller/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export type { CoverageStatus } from './types';
+export type {
+  ShieldControllerActions,
+  ShieldControllerEvents,
+  AllowedActions as ShieldControllerAllowedActions,
+  AllowedEvents as ShieldControllerAllowedEvents,
```
mcmire: Is there a reason why you want to export `AllowedActions` and `AllowedEvents`? These types contain `ShieldControllerActions` and `ShieldControllerEvents` (which are already being exported) and types from other packages, which are already available. We don't typically recommend this and have some guidance around controllers: https://github.com/MetaMask/core/blob/bc2755e933d10ae4e261b94453246652938594b6/docs/controller-guidelines.md#define-but-do-not-export-a-type-union-for-external-action-types

packages/shield-controller/src/logger.ts
```diff
@@ -0,0 +1,5 @@
+import { createProjectLogger } from '@metamask/utils';
+
+import { controllerName } from './constants';
+
+export const projectLogger = createProjectLogger(controllerName);
```
mcmire, line 5: Nit: It looks like you are using the name of the controller class as the namespace for this logger but the idea is that the whole package has a logger and then individual files have loggers derived from that. This way you can prepend a command with `DEBUG=metamask:shield-controller:*` to see log messages for the package or e.g. `DEBUG=metamask:shield-controller:ShieldController` for an individual file.

It might be best to say 

``` suggestion
export const projectLogger = createProjectLogger('shield-controller');
```

Then, in `ShieldController.ts` you can have:

``` typescript
import { projectLogger, createModuleLogger } from './logger';

const log = createModuleLogger(projectLogger, 'ShieldController');
```

packages/shield-controller/src/ShieldController.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  return {
+    controller,
+    messenger,
+    baseMessenger,
+    backend,
+  };
+}
+
+describe('ShieldController', () => {
+  it('should trigger checkCoverage when a new transaction is added', async () => {
```
mcmire: Nit: It looks like you are adding tests directly to the describe for `ShieldController`?What are your thoughts on grouping these tests by method or event? It might be easier to maintain in the long run.

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,259 @@
...
+  >;
+  orderedTransactionHistory: string[]; // List of txIds ordered by time, latest first
+};
+
+/**
+ * Get the default state for the ShieldController.
+ *
+ * @returns The default state for the ShieldController.
+ */
+function getDefaultShieldControllerState(): ShieldControllerState {
```
mcmire: Can you export this function? Occasionally it is useful in tests for constructing mock state objects. We provide some guidance on this here: https://github.com/MetaMask/core/blob/bc2755e933d10ae4e261b94453246652938594b6/docs/controller-guidelines.md#provide-a-default-representation-of-state

packages/shield-controller/src/index.ts
```diff
@@ -0,0 +1,10 @@
...
+  ShieldControllerMessenger,
+  ShieldControllerState,
+} from './ShieldController';
+export { ShieldController } from './ShieldController';
```
mcmire: It looks like the controller defines one action and one event, can you export those here so that other clients and packages can make use of them?

Also can you export the "get default state" function so that it can be used in other places as well?

``` suggestion
  ShieldControllerState,
  ShieldControllerCheckCoverageAction,
  ShieldControllerCoverageResultReceivedEvent
} from './ShieldController';
export {
  ShieldController,
  getDefaultShieldControllerState,
} from './ShieldController';
```

packages/shield-controller/src/ShieldController.ts
```diff
@@ -0,0 +1,259 @@
...
+    this.#transactionHistoryLimit = transactionHistoryLimit;
+    this.#transactionControllerStateChangeHandler =
+      this.#handleTransactionControllerStateChange.bind(this);
+  }
+
+  start() {
+    this.messagingSystem.subscribe(
+      'TransactionController:stateChange',
+      this.#transactionControllerStateChangeHandler,
+      (state) => state.transactions,
```
mcmire, line 157: Good use of a selector function, nice!

packages/shield-controller/package.json
```diff
@@ -0,0 +1,95 @@
...
+    "uuid": "^8.3.2"
+  },
+  "peerDependencies": {
+    "@metamask/accounts-controller": "^32.0.2",
+    "@metamask/approval-controller": "^7.1.3",
+    "@metamask/gas-fee-controller": "^24.0.0",
+    "@metamask/network-controller": "^24.1.0",
+    "@metamask/remote-feature-flag-controller": "^1.7.0",
+    "@metamask/transaction-controller": "^59.2.0"
+  },
```
mcmire: Do all of these need to be peer dependencies? Usually we use peer dependencies to represent controllers that this controller communicates with via the messaging system. But the only package I could find was `transaction-controller`.

``` suggestion
  "devDependencies": {
    "@babel/runtime": "^7.23.9",
    "@lavamoat/allow-scripts": "^3.0.4",
    "@lavamoat/preinstall-always-fail": "^2.1.0",
    "@metamask/transaction-controller": "^59.2.0",
    "@ts-bridge/cli": "^0.6.1",
    "@types/jest": "^27.4.1",
    "deepmerge": "^4.2.2",
    "jest": "^27.5.1",
    "ts-jest": "^27.1.4",
    "typedoc": "^0.24.8",
    "typedoc-plugin-missing-exports": "^2.0.0",
    "typescript": "~5.2.2",
    "uuid": "^8.3.2"
  },
  "peerDependencies": {
    "@metamask/transaction-controller": "^59.2.0"
  },
```

mcmire: It just occurred to me that you might have added the other peer dependencies because you got warnings locally. If so, I don't think they need to be added here in order to satisfy the warnings. `transaction-controller` has its own peer dependencies and a developer should see warnings in the extension or mobile repo if any of `transaction-controller`'s peer deps are incorrect (and same for any other controller with peer deps).

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: fixed

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: added

matthiasgeihs: Wondering: shouldn't the linter complain about this!?

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: OK, thanks. New to this library.

I wasn't sure whether it is OK to put logic inside the `update` callback. I assume there is a limit on what you should do within that function. Probably shouldn't be blocking for a long time. Any general advice on that?

matthiasgeihs: Moved the whole thing into the update callback refactored it a bit.

packages/shield-controller/src/backend.ts
matthiasgeihs: I'm ok with passing in an optional `fetch` function, if this appeared useful in the past.

For context: I've been following the approach of TransactionController here and it seems like it's just relying on the default fetch as well.

I'll address the other minor comments first, and then potentially come back to this one, including your ask to use the "data service" pattern.

packages/shield-controller/src/backend.ts
matthiasgeihs: Not familiar. Note that the backend service we are querying here is a bit special in the sense that is designed to be polled intentionally until the result is there. (First we initiate the backend process, and then we repeatedly check until the result is there or a timeout occurs.) Not sure if this falls into one of the existing use cases for the "data service" pattern that you are mentioning here. I'll have a closer look later.

In general, I'm a fan of keeping dependencies minimal. I don't necessarily see a reason to use a library here, since we can write the polling logic on a couple tens of lines.

packages/shield-controller/src/index.ts
matthiasgeihs: Thanks for the reference.

Reason for exporting, was 1) for the tests and 2) I believe for setting up the controller messenger in extension.
Anyhow, saw that you have the type extractor available `ExtractAvailableAction` for the tests, and I believe in the end I actually did the same thing in the extension POC where I tested things out: [shield-controller-messenger.ts](https://github.com/MetaMask/metamask-extension/blob/7c666875f9935ed84d6a37c889f2951c592a1673/app/scripts/controller-init/messengers/shield/shield-controller-messenger.ts#L4)

Remove these exports now.

packages/shield-controller/src/logger.ts
matthiasgeihs: I see. Again, I was following the pattern used in [TransactionController](https://github.com/MetaMask/core/blob/99a49634c373c119832930c77ff1bbd81387ed36/packages/transaction-controller/src/TransactionController.ts#L89).

Will update my code according to your suggestion.

packages/shield-controller/src/ShieldController.test.ts
matthiasgeihs: Sure. Wasn't aware that we should group. Thanks for the reminder. Added grouping. However, there is currently only one group `checkCoverage`.

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: done, thx for doc reference.

packages/shield-controller/src/index.ts
matthiasgeihs: Two events, actually, `ShieldControllerCoverageResultReceivedEvent` and `ShieldControllerStateChangeEvent`.

Exported controller actions and events now.

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: Was a recommendation by @matthewwalsh0 ;)

packages/shield-controller/package.json
matthiasgeihs: That is exactly right, I think `yarn lint` (?) complained about these not being listed as peer dependencies... I'll remove them.

matthiasgeihs: I've left `"@metamask/auto-changelog": "^3.4.4",` in there because I've seen it is also included in other controllers.

matthiasgeihs: Btw: Any idea why these were reported as missing?

packages/shield-controller/src/backend.ts
matthiasgeihs: Had a look at your example. It's good to know about this option but I'd prefer not using the Cockatiel library to keep dependencies minimal and have more manual control.

packages/shield-controller/src/backend.ts
matthiasgeihs: Can you say a bit more about what problems you were previously facing with using `fetch`? It seems to work fine in the context of TransactionController. Anyhow, I can see that relying on platform-dependent globals can be an issue. Added a fetch parameter.

packages/shield-controller/src/backend.ts
matthiasgeihs: Actually, experimented a bit with `createServicePolicy`. In fact, I believe using that could potentially simplify the code. However, I believe it currently does not support the policy that we want here, which is: "retry an arbitrary number of times (with a given constant pause between attempts) until a fixed time threshold is met". 

The other thing I'm not sure about: How can we control which errors `execute` throws? In particular, how can I distinguish between an error where fetch gave a 500 (unexpected), and where fetch gave a 404 (expected), or where `fetch` returned 200 but `await res.json()` failed?

mcmire: > However, I believe it currently does not support the policy that we want here, which is: "retry an arbitrary number of times (with a given constant pause between attempts) until a fixed time threshold is met".

Hmm, I see. You're right, `createServicePolicy` doesn't quite do that. It sounds like essentially what you want is a combination of at least the retry policy with the [timeout policy](https://github.com/connor4312/cockatiel/blob/f87137b9bb55e0cdfb02e7d5c401e47c6f5fcc9f/src/Policy.ts#L314), and perhaps the circuit breaker policy if you want that. However, the retry policy is intended to take a maximum of retries and it seems like you don't want that. You could specify a very high maximum if you wanted, but essentially it sounds like you want something custom and you'd have to put it together yourself.

The implementation you have here is pretty small so I would agree that perhaps using Cockatiel in this case might be overkill.

packages/shield-controller/src/ShieldController.ts
mcmire: Yes, agreed. I believe this one may be simple to add, I will look into adding it.

packages/shield-controller/src/logger.ts
mcmire: Gotcha. Hmm, perhaps the documentation is lacking here, I will investigate further.

packages/shield-controller/src/ShieldController.ts
mcmire: > I assume there is a limit on what you should do within that function. Probably shouldn't be blocking for a long time. Any general advice on that?

Good intuition — you shouldn't call anything in an `update` function that's computationally expensive, produces side effects, or is asynchronous.

packages/shield-controller/src/backend.ts
mcmire: > Can you say a bit more about what problems you were previously facing with using `fetch`?

I believe that we last encountered issues with `fetch` in 2023, and at the time, Mobile was using Node 14 in development, and we were using a version of React Native that didn't support Node 16 (and perhaps tools like LavaMoat didn't support Node 16 either). Therefore we couldn't assume that global `fetch` worked in all platforms. I believe we tried to resolve the differences using `cross-fetch`, but there were security issues with that package. Simliarly, `node-fetch` didn't play nice with Mobile and `isomorphic-fetch` didn't play nice with Snaps. I think we ultimately decided that injecting `fetch` was the most bulletproof path going forward. ([Here](https://github.com/MetaMask/eth-json-rpc-middleware/pull/190) is a PR removing `node-fetch` from a repo and [here](https://github.com/MetaMask/core/pull/1106) is a PR removing `isomorphic-fetch`, with [related Slack thread](https://consensys.slack.com/archives/C01V1L10W2E/p1676626042328509)).

However, since then, various parts of our stack has been upgraded (dev version of Node on both platforms, React Native in Mobile, LavaMoat, etc.). In fact I see that Snaps is actively using global `fetch` [here](https://github.com/MetaMask/snaps/blob/bd2ad5d9120b4776ab559dabe6bdd4f381ed1a82/packages/snaps-execution-environments/src/common/endowments/network.ts#L229), and of course you point out that TransactionController uses it. So perhaps the advice I gave you is outdated. I'll look into this more tomorrow.

packages/shield-controller/package.json
mcmire: `transaction-controller` has its own peer dependencies, so if you don't install them you will get a warning from Yarn to do so. Typically the process would be to add all peer deps of `transaction-controller` as peer deps of `shield-controller` — sort of passing them along — except that in reality that isn't necessary. When a developer adds `shield-controller` to either Extension or Mobile, they will see a warning saying that they need to install the right version of `transaction-controller`. If that isn't present, once they do that, then they will see that they may need to install the right versions of whatever other controllers `transaction-controller` uses. In other words, all of the peer dependencies need to be satisfied on the consuming end and not here.

chaitanyapotti (approved):

packages/shield-controller/src/backend.ts
```diff
@@ -0,0 +1,161 @@
...
+          await sleep(pollInterval - (Date.now() - startTime));
+        }
+        // The following line will not have an effect as the upper level promise
+        // will already be rejected by now.
+        throw new Error('unexpected error');
+      };
+
+      poll().then(resolve).catch(reject);
+    });
+  }
```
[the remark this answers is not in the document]
matthiasgeihs: This is all intended for now

PR #6141, opened 2025-07-17
feat: add `multichain-account-service` (readonly)

ccharly:
## Explanation

Introducing a new service to support multichain accounts/wallets.

## References

N/A

## Changelog

N/A

## Checklist


package.json
```diff
@@ -47,7 +47,11 @@
   "resolutions": {
     "elliptic@6.5.4": "^6.5.7",
     "fast-xml-parser@^4.3.4": "^4.4.1",
-    "ws@7.4.6": "^7.5.10"
+    "ws@7.4.6": "^7.5.10",
+    "@metamask/keyring-utils@^3.0.0": "npm:@metamask-previews/keyring-utils@3.0.0-e86c8ec",
+    "@metamask/keyring-api@^18.0.0": "npm:@metamask-previews/keyring-api@18.0.0-e86c8ec",
+    "@metamask/keyring-internal-api@^6.2.0": "npm:@metamask-previews/keyring-internal-api@6.2.0-e86c8ec",
+    "@metamask/account-api@^0.1.0": "npm:@metamask-previews/account-api@0.1.0-e86c8ec"
```
ccharly: I'll update this one those packages got released and published.

hmalik88: Should we just wait? You said you were waiting on a core release?

ccharly: I was waiting for the next `accounts` monorepo release, but this has been published. I'll update everything now!

ccharly: Should be ok now!

packages/multichain-account-controller/src/providers/BaseAccountProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.messenger = messenger;
+  }
+
+  #getAccounts(
+    filter: (account: InternalAccount) => boolean = () => true,
+  ): Bip44Account<InternalAccount>[] {
+    const accounts: Bip44Account<InternalAccount>[] = [];
+
+    for (const account of this.messenger.call(
+      'AccountsController:listMultichainAccounts',
```
shane-t: It might be worth adding a comment saying what this actually does (until such time as the action is changed)
i.e. it fetches all internal accounts

packages/multichain-account-controller/src/MultichainAccountController.ts
```diff
@@ -0,0 +1,118 @@
...
+  #getWallet(
+    entropySource: EntropySourceId,
+  ): MultichainAccountWallet<InternalAccount> {
+    const wallet = this.#wallets.get(
+      toMultichainAccountWalletId(entropySource),
+    );
+
+    if (!wallet) {
+      throw new Error(
+        'Unknown wallet, not wallet matching this entropy source',
```
cryptodev-2s: Nit: tiny typo
```suggestion
        'Unknown wallet, no wallet matching this entropy source',

File: packages/multichain-account-controller/src/MultichainAccountController.ts [hunk 4]
```diff
@@ -0,0 +1,118 @@
...
+      );
+    }
+
+    return wallet;
+  }
+
+  getMultichainAccount({
```
cryptodev-2s: Can we add some jsdoc here ? Do we want this to be messenger based also ?

packages/multichain-account-controller/src/MultichainAccountController.ts
```diff
@@ -0,0 +1,118 @@
...
+      this.#getWallet(entropySource).getMultichainAccount(groupIndex);
+
+    if (!multichainAccount) {
+      throw new Error(`No multichain account for index: ${groupIndex}`);
+    }
+
+    return multichainAccount;
+  }
+
+  getMultichainAccounts({
```
cryptodev-2s: Can we add some jsdoc here ? Do we want this to be messenger based also ?

packages/multichain-account-controller/src/MultichainAccountController.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.#messenger = messenger;
+    this.#wallets = new Map();
+    // TODO: Rely on keyring capabilities once the keyring API is used by all keyrings.
+    this.#providers = [
+      new EvmAccountProvider(this.#messenger),
+      new SolAccountProvider(this.#messenger),
+    ];
+  }
+
+  init(): void {
```
cryptodev-2s: Can we add some jsdoc here ? Do we want this to be messenger based also ?

packages/multichain-account-controller/tsconfig.build.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.packages.build.json",
+  "compilerOptions": {
+    "baseUrl": "./",
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
+  "references": [
+    { "path": "../base-controller/tsconfig.build.json" },
+    { "path": "../account-tree-controller/tsconfig.build.json" }
```
cryptodev-2s: Nit: I don't see any reference for account-tree-controller in this new controller ? however I believe it would have to reference account-controller and keyring-controller?

packages/multichain-account-controller/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.packages.json",
+  "compilerOptions": {
+    "baseUrl": "./"
+  },
+  "references": [
+    { "path": "../base-controller" },
+    { "path": "../accounts-controller" }
```
cryptodev-2s: Nit: do we need to add keyring-controller too ?

packages/multichain-account-controller/src/MultichainAccountController.ts
```diff
@@ -0,0 +1,119 @@
...
+    // Gather all entropy sources first.
+    const { keyrings } = this.#messenger.call('KeyringController:getState');
+
+    const entropySources = [];
+    for (const keyring of keyrings) {
+      if (keyring.type === (KeyringTypes.hd as string)) {
+        entropySources.push(keyring.metadata.id);
+      }
+    }
```
hmalik88: nit: I think you can use a reduce here and condense this code block.

packages/multichain-account-controller/src/providers/BaseAccountProvider.ts
hmalik88: Agreed, some JS doc would be good here.

packages/multichain-account-controller/src/MultichainAccountController.ts
ccharly: Yup, I did refactor this a bit indeed 😅 I saw that when handling `KeyringController:stateChange`, so it should be simpler now.

packages/multichain-account-controller/tsconfig.build.json
ccharly: Indeed, those are old configs, it used to depend on the `account-tree-controller` but I actually changed that and forgot to update this, thanks for catching it! 😁

packages/multichain-account-controller/src/providers/BaseAccountProvider.ts
ccharly: Just added a small comment explaining what the action does, and also added a comment to explain that we should probably rename this at some point!

packages/multichain-account-controller/tsconfig.json
ccharly: Fixed here: https://github.com/MetaMask/core/pull/6141/commits/ed97c85d9e5b099ef2b8e41b969ce7340899d12d

packages/multichain-account-controller/tsconfig.build.json
ccharly: Fixed here: https://github.com/MetaMask/core/pull/6141/commits/ed97c85d9e5b099ef2b8e41b969ce7340899d12d

packages/multichain-account-controller/src/providers/BaseAccountProvider.ts
ccharly: Fixed here: https://github.com/MetaMask/core/pull/6141/commits/de60d19f6127a8320caffecca9facfec2ed41117

shane-t (approved):

cryptodev-2s:
@ccharly What do you think about renaming this to a Service instead of a Controller?

ccharly:
> ### Bug: Bip44Account Function Error Handling
> The `isBip44Account` function may throw a `TypeError` if `account.options` is `undefined` when attempting to access `account.options.entropy`. A defensive check for `account.options?.entropy` is recommended. Additionally, the function uses `console.warn` directly, which should be replaced with a proper logging mechanism.
> 
> `packages/multichain-account-service/src/providers/BaseAccountProvider.ts#L26-L33`
> https://github.com/MetaMask/core/blob/b26e62946d91ef99d6450128d483dd8caa1c549d/packages/multichain-account-service/src/providers/BaseAccountProvider.ts#L26-L33
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjMyNWVkNzk3LTQ4N2EtNDVhYy04NTE3LWYxYzJiYjcwNTA1MyIsImVuY3J5cHRpb25LZXkiOiJYc2d5eHduR3FfQmxua2QtQl9iZ2NwSWZQNnFxVml2QThHVGh2OHZHdkVJIiwiYnJhbmNoIjoiZmVhdC9tdWx0aWNoYWluLWFjY291bnRzLWNvbnRyb2xsZXItcmVhZG9ubHkifSwiaWF0IjoxNzUyODU0MzAyLCJleHAiOjE3NTM0NTkxMDJ9.PczagpvPmCp4YYrIDWvKKtM40FVJTTe0WgtQJCNoJaW0Z_4QpRYrfSlUgZi5uOOuB9mHO6r6jRumTBrKh3CtY7kginSEx7Pqj_beMrFUlLLPOO2oGxBL6Q1YKVXQSiqwsHrN3c30I6eVsrbVbeUOYq79wIS6rYLuxOt8SboAeV51X2ZV45cyC8ucx5BJfh6-I4RjzkYtBStzBoad6tGvnt9kzrmhm7te2BAzFGjxBBuBwYRf8Ym9ky0taowFDo1PRgm-VbN75zD_sAody9gMIQrXbd8tOxoIDjsKNSfmiSDd4e_8oTFY8TBkNwiRflcPNL6Wbh7TSy9iuuAYoy7L8Q) • [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjMyNWVkNzk3LTQ4N2EtNDVhYy04NTE3LWYxYzJiYjcwNTA1MyIsImVuY3J5cHRpb25LZXkiOiJYc2d5eHduR3FfQmxua2QtQl9iZ2NwSWZQNnFxVml2QThHVGh2OHZHdkVJIiwiYnJhbmNoIjoiZmVhdC9tdWx0aWNoYWluLWFjY291bnRzLWNvbnRyb2xsZXItcmVhZG9ubHkiLCJyZXBvT3duZXIiOiJNZXRhTWFzayIsInJlcG9OYW1lIjoiY29yZSIsInByTnVtYmVyIjo2MTQxLCJjb21taXRTaGEiOiJiMjZlNjI5NDZkOTFlZjk5ZDY0NTAxMjhkNDgzZGQ4Y2FhMWM1NDlkIn0sImlhdCI6MTc1Mjg1NDMwMiwiZXhwIjoxNzUzNDU5MTAyfQ.fKTbMGzbSLnVbwahYah4NMfL4clNatldqIJn_iLFtD0Mqqke9TUydNDROGhtpZOnxUUYSV9BturKgkfKtuyyFHWXC26VY7rb51QCmgrlxXwW1S_z7KuL_4lU5ibNdRqMSxJWpzMv6AgzmBCXuP6bwfq7jK0CWalGhermSWHvgW9YrGRDQ0Wfa8F37FPlfHh_DW7kIly8_rZjuN1X9PyTv3P1-XKmwjihZexCK3ER3J-LdUjDE1W53VCCp_tkcPQ_gnrLsivjlRfrmIKdQdxfdr85VjNkNyhvlPrSuWz7qnPxmi7x1w79silWJeyEXntKznxbsmEveVeCIDwjZdJReQ)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

Doing this defensive check sounds a bit too much to me given that `Account extends KeyringAccount` and `options` should always be defined.

hmalik88 (approved):

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -0,0 +1,151 @@
...
+    // Gather all entropy sources first.
+    const { keyrings } = this.#messenger.call('KeyringController:getState');
+    this.#setMultichainAccountWallets(keyrings);
+
+    // TODO: For now, we to every `KeyringController` state change to detect when
+    // new entropy sources/SRPs are being added. Having a dedicated event when
+    // new keyrings are added would make this more efficient.
+    this.#messenger.subscribe('KeyringController:stateChange', (state) => {
+      this.#setMultichainAccountWallets(state.keyrings);
+    });
```
mcmire: You mention having a dedicated event, but there's already another way to do this that's more efficient.

The `subscribe` method takes a third argument, a function that works in spirit similar to [input selectors in Redux](https://redux.js.org/usage/deriving-data-selectors#createselector-overview) (that is, they should be simple and only return a single property of the state object). This function should return the part of state that you are interested in observing. 

What about this:

``` suggestion
    this.#messenger.subscribe(
      'KeyringController:stateChange',
      (keyringsFromState) => {
        this.#setMultichainAccountWallets(keyringsFromState);
      },
      selectKeyringControllerKeyrings,
    );
```

And then above the `class MultichainAccountService` line, you could add:

``` typescript
/**
 * Select keyrings from keyring controller state.
 *
 * @param state - The keyring controller state.
 * @returns The keyrings.
 */
function selectKeyringControllerKeyrings(state: KeyringControllerState) {
  return state.keyrings;
}
```

Because the entire `keyrings` array is changed if a new keyring is added or removed, it will detect that `state.keyrings` is a new object reference and your subscription function will be called. And if no keyrings are added or removed, your subscription function will not be called.

packages/multichain-account-service/src/providers/BaseAccountProvider.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { Json } from '@metamask/utils';
+
+import { isBip44Account } from './BaseAccountProvider';
+import { MOCK_HD_ACCOUNT_1 } from '../tests';
+
+describe('isBip44Account', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
```
mcmire: Nit: Jest is already configured to do this automatically, so this shouldn't be needed:

``` suggestion
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -0,0 +1,14 @@
...
+import { KeyringTypes } from '@metamask/keyring-controller';
+import type { InternalAccount } from '@metamask/keyring-internal-api';
+
+import { BaseAccountProvider } from './BaseAccountProvider';
+
+export class EvmAccountProvider extends BaseAccountProvider {
+  isAccountCompatible(account: InternalAccount): boolean {
+    return (
+      account.type === EthAccountType.Eoa &&
+      account.metadata.keyring.type === (KeyringTypes.hd as string)
```
mcmire, line 11: Is this type assertion needed?

``` suggestion
      account.metadata.keyring.type === KeyringTypes.hd
```

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { BaseAccountProvider } from './BaseAccountProvider';
+
+export class SolAccountProvider extends BaseAccountProvider {
+  static SOLANA_SNAP_ID = 'npm:@metamask/solana-wallet-snap' as SnapId;
+
+  isAccountCompatible(account: InternalAccount): boolean {
+    return (
+      account.type === SolAccountType.DataAccount &&
+      account.metadata.keyring.type === (KeyringTypes.snap as string)
```
mcmire, line 14: Is this type assertion needed?

``` suggestion
      account.metadata.keyring.type === KeyringTypes.snap
```

packages/multichain-account-service/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export type { MultichainAccountServiceMessenger } from './types';
```
mcmire: Nit: I know that currently this class does not expose any actions, but for consistency with other controllers and services, should we go ahead and export the action and event types now as well?

``` suggestion
export type {
  MultichainAccountServiceActions,
  MultichainAccountServiceEvents,
  MultichainAccountServiceMessenger,
} from './types';
```

packages/multichain-account-service/src/tests/accounts.test.ts
mcmire, line 1: Nit: Are these tests for tests? :)

Typically if a test helper needs tests it indicates it's too complex. Maybe we can rely on the fact it is being tested indirectly and so we don't need this file?

mcmire (commented):
Looks great! Just some minor comments.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
ccharly: So, it's not "truly needed" but we do get a warning if we're not doing it that way, cause `account.metadata.keyring.type` is actually a `string` and we compare it with an `enum` value... I'm also a bit surprised by this TBH, but I have seen this same warning in some `AccountsController` comparisons too:

```
The two values in this comparison do not have a shared enum type.eslint[@typescript-eslint/no-unsafe-enum-comparison](https://typescript-eslint.io/rules/no-unsafe-enum-comparison)
```

packages/multichain-account-service/src/index.ts
ccharly: Definitely!

packages/multichain-account-service/src/tests/accounts.test.ts
ccharly: Actually, I can probably remove this yes. I had to add a tests because some `MockAccountBuilder` methods are not yet in use, but they will be used in a future PR. For the moment I guess I can just remove them, thus, they should be covered "implicitly" yes 👍

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Sounds like a good idea! I always forgot we have built-in selectors for controllers too! 🙈

Let me do this!

packages/multichain-account-service/src/providers/SolAccountProvider.ts
mcmire: Oh weird. For some reason I wasn't getting that warning locally. But I do know about it. Never mind then :)

packages/multichain-account-service/src/tests/accounts.test.ts
ccharly: Fixed here: https://github.com/MetaMask/core/pull/6141/commits/18a63581c8fa1501b08003b3e6e4b0d13e8586a7

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Fixed here: https://github.com/MetaMask/core/pull/6141/commits/7399ee396697f01296585b0f2b67147acac0731f

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: Similar answer: https://github.com/MetaMask/core/pull/6141#discussion_r2216951281

PatrykLucka (approved):

PR #6142, opened 2025-07-17
feat: Add Messenger `parent` parameter

Gudahtt:
## Explanation

Add the `parent` constructor parameter to the `Messenger` class for automatic delegation of all capabilities under that messenger's namespace. This significantly reduces boilerplate, making the messenger easier to use in a similar manner to how the old `RestrictedMessenger` class wored.

See this ADR PR for details: MetaMask/decisions#53

## References

Depends upon #6132
Relates to #5626

## Checklist

### DIFF
--- a/packages/messenger/src/Messenger.ts
+++ b/packages/messenger/src/Messenger.ts
@@ -184,6 +184,16 @@ export class Messenger<
 > {
   readonly #namespace: Namespace;
 
+  /**
+   * The parent messenger. All actions/events under this namespace are automatically delegated to
+   * the parent messenger.
+   */
+  readonly #parent?: Messenger<
+    string,
+    Action | ActionConstraint,
+    Event | EventConstraint
+  >;
+
   readonly #actions = new Map<Action['type'], Action['handler']>();
 
   readonly #events = new Map<Event['type'], EventSubscriptionMap<Event>>();
@@ -225,11 +235,26 @@ export class Messenger<
   /**
    * Construct a messenger.
    *
+   * If a parent messenger is given, all actions and events under this messenger's namespace will
+   * be delegated to the parent automatically.
+   *
    * @param args - Constructor arguments
    * @param args.namespace - The messenger namespace.
+   * @param args.parent - The parent messenger.
    */
-  constructor({ namespace }: { namespace: Namespace }) {
+  constructor({
+    namespace,
+    parent,
+  }: {
+    namespace: Namespace;
+    parent?: Messenger<
+      string,
+      Action | ActionConstraint,
+      Event | EventConstraint
+    >;
+  }) {
     this.#namespace = namespace;
+    this.#parent = parent;
   }
 
   /**
@@ -257,6 +282,9 @@ export class Messenger<
       );
     }
     this.#registerActionHandler(actionType, handler);
+    if (this.#parent) {
+      this.delegate({ actions: [actionType], messenger: this.#parent });
+    }
   }
 
   #registerActionHandler<ActionType extends Action['type']>(
@@ -400,6 +428,12 @@ export class Messenger<
         }:'`,
       );
     }
+    if (
+      this.#parent &&
+      !this.#subscriptionDelegationTargets.get(eventType)?.has(this.#parent)
+    ) {
+      this.delegate({ events: [eventType], messenger: this.#parent });
+    }
     this.#registerInitialEventPayload({ eventType, getPayload });
   }
 
@@ -448,6 +482,12 @@ export class Messenger<
       throw new Error(
         `Only allowed publishing events prefixed by '${this.#namespace}:'`,
       );
+    }
+    if (
+      this.#parent &&
+      !this.#subscriptionDelegationTargets.get(eventType)?.has(this.#parent)
+    ) {
+      this.delegate({ events: [eventType], messenger: this.#parent });
     }
     this.#publish(eventType, ...payload);
   }
@@ -805,6 +845,9 @@ export class Messenger<
     events?: DelegatedEvents;
     messenger: Delegatee;
   }) {
+    if (messenger === this.#parent) {
+      throw new Error('Cannot revoke from parent');
+    }
     for (const actionType of actions || []) {
       const delegationTargets = this.#actionDelegationTargets.get(actionType);
       if (!delegationTargets || !delegationTargets.has(messenger)) {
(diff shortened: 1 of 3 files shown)


packages/messenger/src/Messenger.ts
```diff
@@ -134,6 +134,16 @@ export class Messenger<
...
 
+  /**
+   * The parent messenger. All actions/events under this namespace are automatically delegated to
+   * the parent messenger.
+   */
+  readonly #parent?: Messenger<
+    string,
+    Action | ActionConstraint,
+    Event | EventConstraint
+  >;
```
mcmire: Is it true that 1) all of the actions and events in this messenger are eligible for delegation, and because of this, 2) all of the actions and events that the parent knows about must be known to this messenger too? If so, can we say this?

``` suggestion
  readonly #parent?: DelegatedMessenger<Action, Event>;
```

packages/messenger/src/Messenger.ts
```diff
@@ -178,11 +188,26 @@ export class Messenger<
...
+  constructor({
+    namespace,
+    parent,
+  }: {
+    namespace: Namespace;
+    parent?: Messenger<
+      string,
+      Action | ActionConstraint,
+      Event | EventConstraint
+    >;
```
mcmire: Similar as above, can we say this?

``` suggestion
    parent?: DelegatedMessenger<Action, Event>;
```

packages/messenger/src/Messenger.ts
Gudahtt: For 2 you must have meant the reverse? That the parent must know about all of the child actions/events. But yes, both of those are correct.

Good suggestion, I've made the change here: 0a293cd28bc79b318351f3ae7186e1e9b87e5b60

packages/messenger/src/Messenger.ts
Gudahtt: Done here: 0a293cd28bc79b318351f3ae7186e1e9b87e5b60

packages/messenger/src/Messenger.ts
```diff
@@ -257,6 +287,11 @@ export class Messenger<
       );
     }
     this.#registerActionHandler(actionType, handler);
+    if (this.#parent) {
+      // @ts-expect-error The parent type isn't constructed in a way that proves it supports this
+      // action, but this is OK because it's validated in the constructor.
+      this.delegate({ actions: [actionType], messenger: this.#parent });
+    }
```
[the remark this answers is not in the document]
Gudahtt: >Duplicate action delegation error

This is fine; it's OK for there to be an error if someone manually delegates to a parent. Just don't do that.


The transitive delegation issue is interesting though. I hadn't thought too much about that. We could disallow it for simplicity, but it may come in handy in some cases so I'll take a closer look.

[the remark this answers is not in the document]
Gudahtt: Wait no, delegation to parents only applies to capabilities under the current namespace. Nothing delegated to a messenger would be under its namespace. Therefore delegation to a parent is inherently not transitive. It's intentional and OK for it not to be transitive.

packages/messenger/src/Messenger.ts
```diff
@@ -181,9 +181,24 @@ export class Messenger<
   Namespace extends string,
   Action extends ActionConstraint,
   Event extends EventConstraint,
+  Parent extends Messenger<
```
Gudahtt, line 184: This is a new addition from what was described in the ADR. As you can see in the tests, it does make using this a bit more verbose, but it's not too bad IMO. This was the only way I could find to validate that the parent had all necessary actions/events (this was validated in the constructor params, but the method of validation relied on there being a generic type param to validate against).

packages/messenger/src/Messenger.ts
```diff
@@ -181,9 +181,24 @@ export class Messenger<
...
+    string,
+    ActionConstraint,
+    EventConstraint,
+    // Use `any` to avoid preventing a parent from having a parent. `any` is harmless in a type
```
mcmire, line 188: This does work, but wouldn't leaving this type parameter out also work? Then the default would be `never` which seems more accurate.

packages/messenger/src/Messenger.ts
```diff
@@ -257,6 +287,11 @@ export class Messenger<
       );
     }
     this.#registerActionHandler(actionType, handler);
+    if (this.#parent) {
+      // @ts-expect-error The parent type isn't constructed in a way that proves it supports this
```
mcmire, line 291: I feel like there ought to be a way to align the types so this isn't necessary, but I don't know how. I tried changing the type of `#parent` from `DelegatedMessenger` to `Parent` but then the `if (messenger === this.#parent) {` statement you added at the beginning of `revoke` failed because TypeScript thought there was no overlap between the two values. So it seems that we need `@ts-expect-error` somewhere...

I do agree this seems type-safe though given the check we have in the constructor.

packages/messenger/src/Messenger.ts
mcmire: Makes sense given how we've designed `Messenger`. I can't think of another way to do this. We can use the `typeof` trick you used in the tests, though, to make providing this type easier.

packages/messenger/src/Messenger.ts
```diff
@@ -181,9 +181,24 @@ export class Messenger<
   Namespace extends string,
   Action extends ActionConstraint,
```
mcmire, line 182: Side note, but should `Action` and `Event` default to `never` so that if you don't supply any type parameters to `Messenger`, allowing TypeScript to infer them, you don't end up with a messenger that recognizes all actions and events?

mcmire (approved):
LGTM!

packages/messenger/src/Messenger.ts
Gudahtt: Good idea. I will try this separately

packages/messenger/src/Messenger.ts
Gudahtt: I was hesitant to do this because there is nothing really stopping someone from writing a messenger that is both a parent and has a parent. It might even be a reasonably good design. For example, if the `TransactionController` wanted to have internal messengers for each helper class it uses, they might delegate to the main transaction controller messenger so their actions/events can be used internally in the transaction controller.

mcmire: Ah, I missed the double negative. "_Avoid_ preventing". Got it :)

Yes I agree we should make it possible to essentially create a delegation chain.

packages/messenger/src/Messenger.ts
Gudahtt: Done here: https://github.com/MetaMask/core/pull/6311

PR #6147, opened 2025-07-18
feat(accounts-controller): add new typed options for `InternalAccount`s

ccharly:
## Explanation

Properly populate entropy options for EVM accounts and also add new typed options for every `InternalAccount`s.

Test PR:
- https://github.com/MetaMask/metamask-extension/pull/34468

## References

N/A

## Changelog

N/A

## Checklist


ccharly:
@metamaskbot publish-preview

packages/accounts-controller/src/utils.ts
```diff
@@ -99,3 +127,71 @@ export function isHdKeyringType(keyringType: KeyringTypes): boolean {
...
+}
+
+/**
+ * HD keyring account for Snap accounts that handles non-EVM HD accounts.
+ */
+export const HdSnapKeyringAccountOptionsStruct = object({
+  entropySource: string(),
+  index: number(),
+  derivationPath: string(),
+});
```
montelaidev, line 183: This isn't following the typed option. `entropySource` should be `id` now right?

packages/accounts-controller/src/utils.ts
```diff
@@ -99,3 +127,71 @@ export function isHdKeyringType(keyringType: KeyringTypes): boolean {
...
+  if (!isHdKeyringType(keyring.type)) {
+    // We cannot extract the group index from non-HD keyrings.
+    return undefined;
+  }
+
+  // We need to find the account index from its HD keyring. We assume those
+  // accounts are ordered, thus we can use their index to compute their
+  // derivation path and group index.
+  const groupIndex = keyring.accounts.findIndex(
+    (accountAddress) => accountAddress === address,
```
montelaidev: We should do a case insensitive compare here?

packages/accounts-controller/src/utils.ts
```diff
@@ -99,3 +127,71 @@ export function isHdKeyringType(keyringType: KeyringTypes): boolean {
...
+  );
+
+  // If for some reason, we cannot find this address, then the caller made a mistake
+  // and it did not use the proper keyring object. For now, we do not fail and just
+  // consider this account as "simple account".
```
montelaidev, line 166: `undefined` would be equivalent to a simple account? That doesn't seem correct.

packages/accounts-controller/src/utils.ts
montelaidev: We should add a TODO here. Currently, we rely on the order in which addresses are added to correspond to the group index. However, this may not hold true in the future if accounts can be added out of order.

packages/accounts-controller/src/utils.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        expect(isNormalKeyringType(keyringType)).toBe(true);
+      }
+    });
+
+    it('returns false for snap keyring type', () => {
+      expect(isNormalKeyringType(snapKeyringType)).toBe(false);
+    });
+  });
+
+  describe('getGroupIndexFromAddressIndex', () => {
```
montelaidev, line 22: We should add scenarios where the user inputs a checksummed address as well.

packages/accounts-controller/src/tests/mocks.ts
```diff
@@ -111,12 +117,22 @@ export const createMockInternalAccountOptions = (
   keyringIndex: number,
   keyringType: KeyringTypes,
   groupIndex: number,
-): Record<string, string | number> => {
+): KeyringAccount['options'] => {
   if (keyringType === KeyringTypes.hd) {
+    const entropySource = `mock-keyring-id-${keyringIndex}`;
+    const derivationPath = `m/44'/60'/0'/0/${groupIndex}`;
```
montelaidev, line 123: [Coin type `60`](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#user-content-Coin_type) is only for eth. Should we have a mapping of scope -> cointype ?

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -540,58 +549,60 @@ export class AccountsController extends BaseController<
...
+
+        const keyringAccountIndex =
+          keyringAccountIndexes.get(keyringTypeName) ?? 0;
         if (keyringAccountIndex) {
-          keyringTypes.set(keyringTypeName, keyringAccountIndex + 1);
+          keyringAccountIndexes.set(keyringTypeName, keyringAccountIndex + 1);
         } else {
-          keyringTypes.set(keyringTypeName, 1);
+          keyringAccountIndexes.set(keyringTypeName, 1);
         }
```
montelaidev: Don't we start at 0 index?

packages/accounts-controller/src/utils.ts
```diff
@@ -99,3 +127,71 @@ export function isHdKeyringType(keyringType: KeyringTypes): boolean {
 export function getDerivationPathForIndex(index: number): string {
   return `m/44'/60'/0'/0/${index}`;
```
montelaidev: Same comment about cointype.

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -609,20 +620,79 @@ export class AccountsController extends BaseController<
...
+    };
+
+    let options: InternalAccount['options'] = {};
+    if (isHdKeyringType(keyring.type)) {
+      // We need to find the account index from its HD keyring.
+      const groupIndex = getGroupIndexFromAddressIndex(keyring, address);
+
+      // If for some reason, we cannot find this address, then the caller made a mistake
+      // and it did not use the proper keyring object. For now, we do not fail and just
+      // consider this account as "simple account".
```
montelaidev, line 657: How are we handling undefined groupIndex on the ui side?

packages/accounts-controller/src/utils.ts
ccharly: You are right, those are the current `options` used by the Solana snap. 😅 I should probably make this a bit more clear.

But I use this small `struct` so I can verify the `options` shape of those Snap accounts before "migrating" them to the new typed options.

packages/accounts-controller/src/AccountsController.ts
ccharly: Well, for "account naming" we do start at 1 (and I just re-used the previous logic, I just renamed the variables to make it more clear.

That being said, with the current logic, the `keyringAccountIndexes` will always be 1 and I think the `if` is not necessary actually, because:
```
const keyringAccountIndex = 0;
keyringAccountIndexes.set(keyringTypeName, keyringAccountIndex + 1);
// Would set `keyringAccountIndexes` to 1.

const keyringAccountIndex = 0;
if (keyringAccountIndex) {
  keyringTypes.set(keyringTypeName, keyringAccountIndex + 1);
} else {
  // This branch will get executed.
  keyringTypes.set(keyringTypeName, 1);
  // Would set `keyringAccountIndexes` to 1.
}
```

So they do seem similar.

Lastly, we mostly use those indexes to compute the account name, but I do believe we never end up using this one, cause we ALWAYS set a name when creating an account 🤔 So this fallback is probably never executed and probably buggy too 🙃

I'll try to fix it still.

packages/accounts-controller/src/AccountsController.ts
ccharly: This should not really be used, or at least, it won't be used directly with multichain accounts. Though, if the `groupIndex` is somehow missing, this account won't be grouped under a multichain account group (because we would have no way to identify its index).

packages/accounts-controller/src/utils.test.ts
ccharly: Given your [other comment](https://github.com/MetaMask/core/pull/6147/files#r2220796962), we probably should yes.

packages/accounts-controller/src/utils.ts
ccharly: We only use this function for EVM HD accounts. That being said, I can probably rename this option to something more specific yes!

packages/accounts-controller/src/utils.ts
ccharly: Yes, that's why I added this comment: `We assume those accounts are ordered, thus we can use their index to compute their derivation path and group index.`

But I can make this a bit more clear if needed. All this kind of "hacky logic" would be gone once the HD keyring start using the new unified keyring API.

packages/accounts-controller/src/utils.ts
ccharly: The other option would be to throw here. But that could be really risky and could brick wallets if not handled I guess 🤔

Also, this should NEVER happen, truly 😅 but I need to handle the error case.

Do you have anything else in mind for this case?

packages/accounts-controller/src/tests/mocks.ts
ccharly: Nope, given that we only care about EVM derivation path here.

packages/accounts-controller/src/AccountsController.ts
montelaidev: Maybe we can add a deprecation notice to the name property in the internal account. The account name will come from the account group.

packages/accounts-controller/src/AccountsController.ts
montelaidev: Maybe we can put hd accounts with undefined group indexes into a "malformed account group"? It won't be shown in the ui anyways but as a user we can dump the state to see missing accounts there.

ccharly:
@metamaskbot publish-preview

packages/accounts-controller/src/AccountsController.ts
ccharly: > Maybe we can add a deprecation notice to the name property in the internal account. The account name will come from the account group.

Sounds like a good idea, but IMO we should do that once we truly introduce account group naming!

montelaidev (approved):

PR #6149, opened 2025-07-18
fix: don't poll indefinitely for bridge status if 500 errors

infiniteflower:
## Explanation

This PR fixes an issue where we would poll indefinitely if the Bridge API keeps returning a 500 error. The original bug was that the tx was submitted on the wrong chain and therefore the txHash would never appear on the chain we are querying for, so we would poll the Bridge API indefinitely.

Screenshot 2025-07-18 at 5 04 19 PM

## References

Original issue: https://consensyssoftware.atlassian.net/browse/SWAPS-655
Mobile implementation: https://github.com/MetaMask/metamask-mobile/pull/17396

## Checklist


infiniteflower:
@metamaskbot publish-preview

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -467,14 +590,21 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
         state.txHistory[bridgeTxMetaId] = newBridgeHistoryItem;
       });
 
+      // Reset attempts counter on successful fetch
+      // A successful fetch does not mean the the tx is in a final state
+      this.update((state) => {
+        state.txHistory[bridgeTxMetaId].attempts = undefined;
```
micaelae: What's the reason for updating this separately instead of doing so when `newBridgeHistoryItem` is set?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -177,6 +179,10 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       `${BRIDGE_STATUS_CONTROLLER_NAME}:submitTx`,
       this.submitTx.bind(this),
     );
+    this.messagingSystem.registerActionHandler(
+      `${BRIDGE_STATUS_CONTROLLER_NAME}:resetAttempts`,
```
micaelae: Minor: what do you think of renaming this to `restartStatusPolling` to make its purpose easier to understand when reading the client code?

packages/bridge-status-controller/src/bridge-status-controller.ts
infiniteflower: Done in 6d9f6be173b22790148e7637cac1e4c3e07cd073

packages/bridge-status-controller/src/bridge-status-controller.ts
infiniteflower: Good point, I called it `restartPollingForFailedAttempts` since there's also `restartPollingForIncompleteHistoryItems`.

infiniteflower: Done in 0e927982e6abdb62e82633b0964dd530fbc5ba17

micaelae (approved):

micaelae (approved):

PR #6153, opened 2025-07-21
Tat 1264 mobile remove usage s of global chain id in earn experiences

ameliejyc:
## Checklist


packages/earn-controller/src/EarnController.ts
```diff
@@ -323,43 +312,34 @@ export class EarnController extends BaseController<
...
-        }
+        this.#initializeSDK(
+          networkControllerState.selectedNetworkClientId,
+        ).catch(console.error);
+
+        // refresh pooled staking data
+        this.refreshPooledStakingVaultMetadata().catch(console.error);
+        this.refreshPooledStakingVaultDailyApys().catch(console.error);
+        this.refreshPooledStakingVaultApyAverages().catch(console.error);
+        this.refreshPooledStakes().catch(console.error);
```
ameliejyc, line 349: I've left these calls the same here but I don't know if it brings much value to refresh PS data on each network change given PS isn't multi-chain? For the Lending calls below it seems to make more sense.

ameliejyc:
@metamaskbot publish-preview

packages/earn-controller/src/EarnController.ts
salimtb: the first `NetworkController:stateChange` capture all changes on the network state ( example: rpc change , network name change ... etc ) , the new one `NetworkController:networkDidChange` capture changes on the selected network only.
this is much better for perf, good change

packages/earn-controller/src/EarnController.ts
```diff
@@ -522,31 +455,36 @@ export class EarnController extends BaseController<
    * @param options - Optional arguments
    * @param [options.resetCache] - Control whether the BE cache should be invalidated (optional).
    * @param [options.address] - The address to refresh pooled stakes for (optional).
+   * @param [options.chainId] - The chain id to refresh pooled stakes for (optional).
    * @returns A promise that resolves when the stakes data has been updated
    */
   async refreshPooledStakes({
     resetCache = false,
     address,
+    chainId = ChainId.ETHEREUM,
```
salimtb, line 475: can you help me to understand why you have eth mainnet as default value here

ameliejyc: Earn has two main products, Pooled Staking and Lending, and Pooled Staking is only available for Ethereum mainnet and Hoodi. The fallback here is just to make sure that the Earn Service API call that happens later succeeds with a valid chainID should `refreshPooledStakes` be called without the parameter. Whether or not `refreshPooledStakes` should actually be called without the chainID though is a question I have related to my earlier comment - hoping @nickewansmith might have an opinion on that one.

packages/earn-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING:** Removed usages of `NetworkController:getState` for GNS removal. Added mandatory `chainId` parameter to `executeLendingTokenApprove`, `executeLendingWithdraw`, `executeLendingDeposit`, `getLendingMarketDailyApysAndAverages` and `getLendingPositionHistory` methods. Added mandatory `selectedNetworkClientId` parameter to constructor. Added optional `chainId` parameter to `refreshPooledStakingVaultApyAverages`, `refreshPooledStakingVaultDailyApys`, `refreshPooledStakingVaultMetadata` and `refreshPooledStakes`. ([#6153](https://github.com/MetaMask/core/pull/6153))
```
mcmire: There's a lot of things in this changelog entry. Ideally, each changelog entry should describe one change, or at least one kind of change. What are your thoughts on breaking this up? For instance you could say:

``` suggestion
## Added

- **BREAKING:** Added mandatory parameter `selectedNetworkClientId` to `EarnController` constructor ([#6153](https://github.com/MetaMask/core/pull/6153))
- **BREAKING:** Added mandatory `chainId` parameter to `executeLendingTokenApprove`, `executeLendingWithdraw`, `executeLendingDeposit`, `getLendingMarketDailyApysAndAverages` and `getLendingPositionHistory` methods ([#6153](https://github.com/MetaMask/core/pull/6153))
- **BREAKING:** Introduce `chainId` as first argument to `refreshPooledStakingVaultDailyApys` (defaults to Ethereum) ([#6153](https://github.com/MetaMask/core/pull/6153))
- Added optional `chainId` parameter to `refreshPooledStakingVaultApyAverages`, `refreshPooledStakingVaultMetadata` and `refreshPooledStakes` (defaults to Ethereum) ([#6153](https://github.com/MetaMask/core/pull/6153))

## Changed

- **BREAKING:** `EarnController` messenger must now allow `NetworkController:getSelectedNetworkClient` and `NetworkController:networkDidChange` and must not allow `NetworkController:getState` and `NetworkController:stateChange` ([#6153](https://github.com/MetaMask/core/pull/6153))
- `refreshPooledStakingData` now refreshes for all supported chains, not just global chain ([#6153](https://github.com/MetaMask/core/pull/6153))
```

I left out the removals of `EarnControllerGetStateAction` and `EarnControllerActions` because I'm not sure we want to do that.

packages/earn-controller/src/EarnController.ts
```diff
@@ -323,43 +310,34 @@ export class EarnController extends BaseController<
...
+        this.refreshPooledStakingVaultMetadata().catch(console.error);
+        this.refreshPooledStakingVaultDailyApys().catch(console.error);
+        this.refreshPooledStakingVaultApyAverages().catch(console.error);
+        this.refreshPooledStakes().catch(console.error);
+
+        // refresh lending data for all chains
+        this.refreshLendingMarkets().catch(console.error);
+        this.refreshLendingPositions().catch(console.error);
+
         this.#selectedNetworkClientId =
```
mcmire: Nit: Is there a reason why this is being stored last in the handler? If other methods need to use it then it won't be up to date until after those methods are called.

packages/earn-controller/src/EarnController.ts
nickewansmith: @ameliejyc the idea behind this was moreso to keep the data fresh as the PS data can be is visible in the popular networks filtered view while users are switching to other networks unrelated to PS

packages/earn-controller/src/EarnController.ts
```diff
@@ -285,25 +270,27 @@ export class EarnController extends BaseController<
...
-  readonly #supportedPooledStakingChains: Hex[];
+  readonly #supportedPooledStakingChains: number[];
 
   readonly #env: EarnEnvironments;
 
   constructor({
     messenger,
     state = {},
     addTransactionFn,
+    selectedNetworkClientId,
```
mcmire, line 300: Hmm, we may have followed this pattern in the past, but this will require that NetworkController be initialized before EarnController. We are trying to eliminate dependencies on controllers like this and make it so that it doesn't matter which order controllers are initialized, they will "just work".

I was going to suggest that we make `selectedNetworkClientId` optional, then add an `init` method where we pull the selected network via `NetworkController:getState`. But that's a bit similar to what we were doing before, and access to `NetworkController:getState` seems to be getting removed in this PR. The PR description claims that this is due to removing reliance on the global network, but `selectedNetworkClientId` _is_ the global network. So I'm a bit confused on what this change in particular achieves. Are we also trying to increase performance?

packages/earn-controller/src/EarnController.ts
nickewansmith: This should be ok, as the PS data is kept in the EarnController state per network. The default mainnet refreshed data will be available for users using the popular network filter view which can show mainnet pooled staking data.

packages/earn-controller/src/EarnController.ts
```diff
@@ -219,25 +217,12 @@ export function getDefaultEarnControllerState(): EarnControllerState {
...
- */
-export type EarnControllerGetStateAction = ControllerGetStateAction<
-  typeof controllerName,
-  EarnControllerState
->;
-
-/**
- * All actions that EarnController registers, to be called externally.
- */
-export type EarnControllerActions = EarnControllerGetStateAction;
```
mcmire: It's standard for all controllers to export an `*Actions` type, even if nothing is there. What are your thoughts on keeping this but setting the right hand side to `never`?

packages/earn-controller/src/EarnController.ts
```diff
@@ -219,25 +217,12 @@ export function getDefaultEarnControllerState(): EarnControllerState {
 
 // === MESSENGER ===
 
-/**
- * The action which can be used to retrieve the state of the EarnController.
- */
-export type EarnControllerGetStateAction = ControllerGetStateAction<
```
mcmire: Can you explain why we are removing this event? It's standard for all controllers to export this event (so that other controllers can listen for state changes).

packages/earn-controller/src/EarnController.ts
```diff
@@ -611,27 +555,32 @@ export class EarnController extends BaseController<
    * Refreshes pooled staking vault daily apys for the current chain.
    * Updates the pooled staking vault daily apys controller state.
    *
+   * @param chainId - The chain id to refresh pooled staking vault daily apys for (optional).
    * @param days - The number of days to fetch pooled staking vault daily apys for (defaults to 365).
    * @param order - The order in which to fetch pooled staking vault daily apys. Descending order fetches the latest N days (latest working backwards). Ascending order fetches the oldest N days (oldest working forwards) (defaults to 'desc').
    * @returns A promise that resolves when the pooled staking vault daily apys have been updated.
    */
   async refreshPooledStakingVaultDailyApys(
+    chainId: number = ChainId.ETHEREUM,
```
mcmire: This is a breaking change. In the future, perhaps consider grouping these arguments into an options bag so that a new property can be added to the options without affecting the method signature?

packages/earn-controller/src/EarnController.ts
ameliejyc: I see, okay makes sense to keep it in then. Thanks for confirming Nick!

packages/earn-controller/src/EarnController.ts
```diff
@@ -866,7 +827,7 @@ export class EarnController extends BaseController<
    * Gets the lending market daily apys and averages for the current chain.
    *
    * @param options - Optional arguments
-   * @param [options.chainId] - The chain id to get lending market daily apys and averages for (optional).
+   * @param [options.chainId] - The chain id to get lending market daily apys and averages for.
```
mcmire: Nit

``` suggestion
   * @param options.chainId - The chain id to get lending market daily apys and averages for.

File: packages/earn-controller/CHANGELOG.md (hunk 3 above)
Comment (ameliejyc) [reply packages/earn-controller/CHANGELOG.md]: That's definitely clearer, thanks for pointing this out! Will incorporate those suggestions.

File: packages/earn-controller/src/EarnController.ts (hunk 7 above)
Comment (ameliejyc) [reply packages/earn-controller/src/EarnController.ts]: This was an oversight, I think I mistook it for the NetworkControllerGetStateAction. Thanks for the explanation, I'm new to this codebase and this is really helpful.

File: packages/earn-controller/src/EarnController.ts (hunk 5 above)
Comment (salimtb) [reply packages/earn-controller/src/EarnController.ts:300]: Hey @mcmire,

The main objective here is to remove the direct dependency on the globalNetwork selector, which currently relies on NetworkController.state.selectedNetworkClientId.

Instead of accessing the network directly from the `NetworkController`, we want to pass `selectedNetworkClientId` as a parameter to the EarnController constructor during initialization.

For now, since the globalNetwork selector is still in place on mobile, it will continue to be used. But in the upcoming PR, once it's removed, we’ll update the initialization to pass the enabledNetwork value instead.

So on the client side, the initialization will look like this:

```
new EarnController({
  ...,
  selectedNetworkClientId: Engine.context.NetworkController.state.selectedNetworkClientId,
  ...
})
```
Passing the network value from the client has two main benefits:

- It allows us to easily swap or override the value as needed.
- It eliminates the tight coupling with NetworkController.selectedNetworkClientId, improving modularity and flexibility.

File: packages/earn-controller/src/EarnController.ts (hunk 6 above)
Comment (ameliejyc) [reply packages/earn-controller/src/EarnController.ts]: Got it, I think I can just completely revert this change then rather than setting it to `never`?

File: packages/earn-controller/src/EarnController.ts (hunk 4 above)
Comment (ameliejyc) [reply packages/earn-controller/src/EarnController.ts]: No reason. The above methods don't need to use it but that's a great callout - I'll change it to be the first thing updated so that if in future methods rely on it we shouldn't create a bug.

File: packages/earn-controller/src/EarnController.ts (hunk 8 above)
Comment (ameliejyc) [reply packages/earn-controller/src/EarnController.ts]: Also a great callout. And it's more in line with how we call most of these other similar methods so I think I'll go ahead and do that.

File: packages/earn-controller/src/EarnController.ts (hunk 6 above)
Comment (mcmire) [reply packages/earn-controller/src/EarnController.ts]: Ah, yes, sorry, that is what I meant.

File: packages/earn-controller/src/EarnController.ts (hunk 5 above)
Comment (mcmire) [reply packages/earn-controller/src/EarnController.ts:300]: Okay, thanks for the context. How will enabled networks work on mobile? Will multiple enabled networks be allowed like on extension or will only one be allowed? 

Also, you mention that this eliminates the tight coupling to `NetworkController.selectedNetworkClientId`, however, it seems that we are still listening to `networkDidChange` below, so we are still coupled to the global network. Do we have plans to change that down the road as well?

**Comment by ameliejyc** [comment]
@metamaskbot publish-preview

File: packages/earn-controller/src/EarnController.ts (hunk 9 above)
Comment (ameliejyc) [reply packages/earn-controller/src/EarnController.ts]: Resolved these and I'll follow up with another separate PR to fix a lot of other inaccuracies in the jsdocs

File: packages/earn-controller/src/EarnController.ts (hunk 5 above)
Comment (Gudahtt) [reply packages/earn-controller/src/EarnController.ts:300]: Since we're still getting the `selectedNetworkClientId` from the `NetworkController`, this doesn't seem to change how coupled this is with the `NetworkController` at all. I don't follow how this makes it easier to swap or override the value either - do you mean in a test? The messenger class was designed to be easily used in tests for mocking external code.

File: packages/earn-controller/CHANGELOG.md [hunk 10]
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Added mandatory parameter `selectedNetworkClientId` to `EarnController` constructor ([#6153](https://github.com/MetaMask/core/pull/6153))
+- **BREAKING:** Added mandatory `chainId` parameter to `executeLendingTokenApprove`, `executeLendingWithdraw`, `executeLendingDeposit`, `getLendingMarketDailyApysAndAverages` and `getLendingPositionHistory` methods ([#6153](https://github.com/MetaMask/core/pull/6153))
+- **BREAKING:** Change `refreshPooledStakingVaultDailyApys` to accept an options object with `chainId`, `days`, and `order` properties instead of separate parameters ([#6153](https://github.com/MetaMask/core/pull/6153))
```
mcmire: Should we note that `chainId` is a new option? 

``` suggestion
- **BREAKING:** Change `refreshPooledStakingVaultDailyApys` to accept an options object with instead of separate parameters ([#6153](https://github.com/MetaMask/core/pull/6153))
  - `days` and `order` are existing options; `chainId` is a new option
```

packages/earn-controller/jest.config.js
```diff
@@ -23,4 +23,6 @@ module.exports = merge(baseConfig, {
       statements: 100,
     },
   },
+
+  coveragePathIgnorePatterns: ['constants.ts'],
```
mcmire: We don't usually exclude files from the test coverage report unless there's a good reason to. In this case, it seems that we aren't importing this file anywhere in a test, or a file used by a test. Maybe we don't need this file anymore?

packages/earn-controller/src/EarnController.ts
salimtb: @mcmire This is just the initial step. The next phase will involve listening to events from the new `NetworkEnablementController` and dropping the dependency on `networkDidChange`. That said, the related PR is still under review and the controller hasn’t been published yet.

On mobile, the enabled network behavior will align with the extension.

salimtb: @Gudahtt Hmm, I’m not sure I fully follow. Currently, we’re still using `selectedNetworkClientId` from the `NetworkController` and passing it in param because the global network selector hasn’t been removed on mobile yet. Once that’s done, we’ll update this value directly from the mobile UI codebase at the initialisation.

packages/earn-controller/CHANGELOG.md
ameliejyc: Updated!

packages/earn-controller/jest.config.js
ameliejyc: Ah great spot! Updated

packages/earn-controller/src/EarnController.ts
Gudahtt: My point was that if you are passing `selectedNetworkClientId` from the `NetworkController` to the `EarnController`, doing so via a constructor parameter is not less coupled than doing so through the messenger, it's equally coupled. The goals you stated above are not achieved by this PR.

But that was before I heard about your intention to pass in a network client ID from UI state. I can see how this might be a step towards the goals you referenced even if it doesn't achieve them alone. This makes a bit more sense with that context, at least if there is some obstacle to implementing the network enablement controller on mobile (otherwise why not go straight to using that?).

nickewansmith (approved):

packages/earn-controller/src/index.ts
```diff
@@ -5,9 +5,7 @@ export type {
   LendingPositionWithMarket,
   LendingPositionWithMarketReference,
   EarnControllerState,
-  EarnControllerGetStateAction,
```
mcmire: Should we also add back these imports?

mcmire (dismissed):
LGTM!

nickewansmith (approved):

PR #6154, opened 2025-07-21
Release/471.0.0

ccharly:
Release of the `account-tree-controller` to introduce the use of the new `@metamask/account-api` package in clients.


cryptodev-2s (approved):
LGTM!

PR #6155, opened 2025-07-21
fix: Update access token validation and retrieval from vault and state 

lwin-kyaw:
## Explanation

This PR includes fix and update to get the `accessToken` value from the controller. 
`accessToken` data is saved in both vault and state (as non-persisted value) from the `OAuth Authentication` and previously, we only read and validate it from the state when updating the SeedlessOnboarding vault data.

However, there's an edge case in the client (especially in the extension) where `acessToken` is persisted to vault but removed from the state. Please refer to this extension bug (https://github.com/MetaMask/metamask-extension/issues/34437) as an example.
In that case, `accessToken` value is persisted to vault and set (temporarily in mem) to state after the rehydration (unlock), however when the user refresh the page before the onboarding is finished, the `accessToken` value is removed from the state(memory) and when user tries to rehydrate, the controller throw the `InvalidAccessToken` error as it's not available in the in-memory state anymore.

This PR fixes the issue by checking the access-token in both in-memory state and in the encrypted vault.

## References

Fixes: https://github.com/MetaMask/metamask-extension/issues/34437

## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/CHANGELOG.md
+++ b/packages/seedless-onboarding-controller/CHANGELOG.md
@@ -7,8 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [2.4.0]
-
 ### Added
 
 - Added `getAccessToken` public method in the seedless-onboarding-controller. ([#6155](https://github.com/MetaMask/core/pull/6155))
--- a/packages/seedless-onboarding-controller/CHANGELOG.md
+++ b/packages/seedless-onboarding-controller/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- Added `getAccessToken` public method in the seedless-onboarding-controller. ([#6155](https://github.com/MetaMask/core/pull/6155))
-
 ### Fixed
 
 - Retrieve `accessToken` from the encrypted vault if it's not available as an in-memory state. ([#6155](https://github.com/MetaMask/core/pull/6155))
--- a/packages/seedless-onboarding-controller/CHANGELOG.md
+++ b/packages/seedless-onboarding-controller/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.4.0]
+
+### Added
+
+- Added `getAccessToken` public method in the seedless-onboarding-controller. ([#6155](https://github.com/MetaMask/core/pull/6155))
+
+### Fixed
+
+- Retrieve `accessToken` from the encrypted vault if it's not available as an in-memory state. ([#6155](https://github.com/MetaMask/core/pull/6155))
+
 ## [2.3.0]
 
 ### Uncategorized
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
@@ -415,7 +415,7 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
       const performBackup = async (): Promise<void> => {
         // verify the password and unlock the vault
         const { toprfEncryptionKey, toprfAuthKeyPair } =
-          await this.#unlockVaultAndGetBackupEncKeyAndTokens();
+          await this.#unlockVaultAndGetVaultData();
 
         // encrypt and store the secret data
         await this.#encryptAndStoreSecretData({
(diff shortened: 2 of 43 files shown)


lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -867,6 +871,34 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+  /**
+   * Get the access token from the state or the vault.
+   * If the access token is not in the state, it will be retrieved from the vault by decrypting it with the password.
+   *
+   * If both the access token and the vault are not available, an error will be thrown.
+   *
+   * @param password - The password to unlock the vault.
+   * @returns The access token.
+   */
+  async getAccessToken(password: string): Promise<string> {
```
matthiasgeihs: why is this a public function? if there is no good reason for making this public, i think it should be an internal function.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1197,13 +1229,15 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+   * - revokeToken: The decrypted revoke token
+   * - accessToken: The decrypted access token
    * @throws {Error} If:
    * - The password is invalid or empty
    * - The vault is not initialized
    * - The password is incorrect (from encryptor.decrypt)
    * - The decrypted vault data is malformed
    */
-  async #unlockVaultAndGetBackupEncKey(
+  async #unlockVaultAndGetBackupEncKeyAndTokens(
```
himanshuchawla009: this can be simply renamed to `unlockVaultAndGetVaultData`

himanshuchawla009 (changes requested):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1404,11 +1438,7 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
   }): Promise<void> {
     this.#assertIsAuthenticatedUser(this.state);
 
-    if (!this.state.accessToken) {
-      throw new Error(
-        SeedlessOnboardingControllerErrorMessage.InvalidAccessToken,
-      );
-    }
+    const accessToken = await this.getAccessToken(password);
```
matthiasgeihs: i think the controller logic is flawed.
this controller unlocks automatically whenever the keyring controller unlocks.
however, that doesn't mean that the vault is unlocked and the transient controller state is populated. this happens only separately after calling one of the `submit` functions.

i believe this attempted fix is also flawed.
it's assuming that password is provided or encryption key is cached. however, in the issue attached to this PR, afaics the password is not provided and the encryption key was also cleared. hence, this will still fail.

the resolution is not to attempt to load the access token from the vault, but to properly unlock the controller and to ensure that whenever the controller is unlocked, the vault is loaded into the transient state.

matthiasgeihs: @lwin-kyaw maybe i'm missing something. could you explain how the data can be read from the vault here if the state was cleared (that's what you are describing) and the vault hasn't been properly unlocked?

matthiasgeihs: also, if it really works, shouldn't the same logic be applied for `revokeToken`?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Make that method as private in this commit, https://github.com/MetaMask/core/pull/6155/commits/0494abfd5d83ebe65e63ca856b94235ddb63e203

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: To populate the controller state, when the vault is unlocked, we should probably create a new PR as that needs changes in the multiple places.

lwin-kyaw: > i believe this attempted fix is also flawed.
it's assuming that password is provided. however, in the issue attached to this PR, afaics the password is not provided and this will still fail.

I'm pretty sure that in the video in the issues, the password is provided.
Or can you give more context for that?

matthiasgeihs: had a call with @lwin-kyaw. he explained that the issue occurs when `fetchAllSecretData` is called with password.
we don't really know why the state is cleared in the situation described in the issue, but the fix here should work.

packages/seedless-onboarding-controller/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `getAccessToken` public method in the seedless-onboarding-controller. ([#6155](https://github.com/MetaMask/core/pull/6155))
```
matthiasgeihs:
```suggestion
- Added `getAccessToken` method in the seedless-onboarding-controller. ([#6155](https://github.com/MetaMask/core/pull/6155))
```
Either update or remove, as this is no longer public-facing.

matthiasgeihs (changes requested):

packages/seedless-onboarding-controller/CHANGELOG.md
lwin-kyaw: This changelog item is removed in this commit, https://github.com/MetaMask/core/pull/6155/commits/0d65444425fa8220d65c02489328d32197e102da

matthiasgeihs (approved):

matthiasgeihs:
@metamaskbot publish-preview

PR #6156, opened 2025-07-21
feat: Stop persisting `permissionActivityLog` state

Gudahtt:
## Explanation

The `permissionActivityLog` property of the `PermissionLogController` state tracks all calls to restricted methods. This was intended as a diagnostic aid, but in practice it is rarely used. This log is causing performance issues related to disk writes because it can be triggered while the wallet is idle (by snaps calling restricted methods).

To reduce the performance burden of this property, it is no longer persisted. This list will still be available in-memory for diagnosing problems, but will not be preserved across restarts.

## References

Relates to https://github.com/MetaMask/metamask-extension/issues/33879

Here is the draft PR of this update for `metamask-extension` that addresses the breaking changes: https://github.com/MetaMask/metamask-extension/pull/34465
This controller is not used by mobile, so there is no draft PR for mobile.

## Checklist

### DIFF
--- a/packages/permission-log-controller/CHANGELOG.md
+++ b/packages/permission-log-controller/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


Gudahtt:
@metamaskbot publish-preview

GuillaumeRx (approved):

Mrtenz (approved):

PR #6165, opened 2025-07-22
feat(multichain-account-service): re-sync multichain account and wallets on account events

ccharly:
## Explanation

Re-sync multichain accounts and wallets upon `AccountsController` events.

## References

N/A

## Changelog

N/A

## Checklist


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -89,6 +103,72 @@ export class MultichainAccountService {
...
+      return;
+    }
+
+    const wallet = this.#wallets.get(
+      toMultichainAccountWalletId(account.options.entropy.id),
+    );
+    if (wallet) {
+      // This new account might be a new multichain account, and the wallet might not
+      // know it yet, so we need to force-sync here.
+      wallet.sync();
```
gantunesr: Not sure if I follow what is happening here. If it is a new account then the wallet should not now it, I don't understand why it is a "might".

And the `sync` operation seems quite expensive to be executed for every new account added, at least the second part of it,

```ts
for (const [groupIndex, multichainAccount] of this.#accounts.entries()) {
  multichainAccount.sync();

  // Clean up old multichain accounts.
  if (!multichainAccount.hasAccounts()) {
    this.#accounts.delete(groupIndex);
  }
}
```

gantunesr (commented):
Overall it looks good. Left a question and I still need to see the rest of the code to understand better these changes.

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: For the moment, we don't control the account creation through the multichain account "SDK".

So you might have the following scenario:
- You have "Wallet 1"
  * It has 1 "Multichain Account 1" (index 0)
- You now create a new EVM account for the next index (index 1)
- This will trigger a `:accountAdded` event
- Currently "Wallet 1" lives in memory and it does not know about this new account (index 1) yet
- Thus, the service `sync` the appropriate wallet (by checking the `options.entropy.id`):
  * To let it know that a new account got added
    - So the wallet can check all accounts (from its providers) and create (a `MultichainAccount` instance) for the missing "Multichain Account 2" (index 1)


I initially had another idea where `AccountProvider` would implement a sort pub/sub pattern to let the `MultichainAccountWallet` or `MultichainAccount` instances that something got added:
- https://github.com/MetaMask/accounts/pull/320

But the implementation felt more complex, and I opted for the "manual sync" + controller events instead (at least, for the moment).

> And the sync operation seems quite expensive to be executed for every new account added

I'd say, it's ok for the moment. We can always change this if needed (it won't change the public API part). Also, we might not use events anymore once we introduce operations to create "a multichain account" entirely through the SDK/account providers.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -38,17 +40,28 @@ function selectKeyringControllerKeyrings(state: KeyringControllerState) {
   return state.keyrings;
 }
 
+/** Reverse mapping object used to map account IDs and their wallet/multichain account. */
+type AccountReverseMapping<Account extends Bip44Account<KeyringAccount>> = {
+  wallet: MultichainAccountWallet<Account>;
+  multichainAccount: MultichainAccount<Account>;
+};
+
```
hmalik88:
```suggestion
type AccountContext<Account extends Bip44Account<KeyringAccount>> = {
  wallet: MultichainAccountWallet<Account>;
  multichainAccount: MultichainAccount<Account>;
};

```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -38,17 +40,28 @@ function selectKeyringControllerKeyrings(state: KeyringControllerState) {
...
 
   readonly #wallets: Map<
     MultichainAccountWalletId,
-    MultichainAccountWallet<InternalAccount>
+    MultichainAccountWallet<Bip44Account<InternalAccount>>
+  >;
+
+  readonly #reverse: Map<
+    InternalAccount['id'],
+    AccountReverseMapping<Bip44Account<InternalAccount>>
```
hmalik88:
```suggestion
  readonly #accountIdToContext: Map<
    Bip44<InternalAccount>['id'],
    AccountReverseMapping<Bip44Account<InternalAccount>>
```

Can we be more descriptive here with the type and do `Bip44Account<InternalAccount>['id']` instead?

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: That being said, I can maybe improve this 🤔

Given that we only filter for `Bip44Account`, we can do this:
```ts
// `account` is a `Bip44Account`.
const wallet = this.#wallets.get(
  toMultichainAccountWalletId(account.options.entropy.id),
);
if (wallet) {
  let multichainAccount = wallet.getMultichainAccount(
    account.options.entropy.groupIndex
  );
  if (multichainAccount) {
    // This new account is part of an existing multichain account, let it
    // re-sync with its providers.
    multichainAccount.sync();
  } else {
    // This new account is a new multichain account, let the wallet know
    // it has to re-sync with its providers.
    wallet.sync();
    
    multichainAccount = wallet.getMultichainAccount(
      account.options.entropy.groupIndex,
    );
  }
  
  if (multichainAccount) {
    // Same here, this account should have been already grouped in that
    // multichain account.
    this.#reverse.set(account.id, {
      wallet,
      multichainAccount,
    });
  }
}
```

Thus we avoid doing a `wallet.sync` unnecessarily.

hmalik88: Agreed on the sync method feeling like overkill for a single account added. Wondering if there's a more surgical way of syncing this new account?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -89,6 +103,72 @@ export class MultichainAccountService {
...
+    }
+  }
+
+  #handleOnAccountRemoved(id: InternalAccount['id']): void {
+    // Force sync of the appropriate wallet if an account got removed.
+    const found = this.#reverse.get(id);
+    if (found) {
+      const { wallet } = found;
+
+      wallet.sync();
```
hmalik88, line 179: Same sentiment as above re: adding an account.

packages/multichain-account-service/src/tests/accounts.ts
```diff
@@ -170,6 +169,11 @@ export class MockAccountBuilder {
     return new MockAccountBuilder(account);
   }
 
+  withId(id: InternalAccount['id']) {
```
hmalik88, line 172: Can this be `Bip44Account<InternalAccount>['id']`?

ccharly: Since we're dealing with `InternalAccount` in the tests, I'll keep this as `InternalAccount['id']` (Though, I made the change for other `InternalAccount['id']`, cause I also think it's better that way!) Thanks

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Yup, delegating the synchronization on the wallet and multichain accounts themselves might be better indeed.

I do have a draft PR for this, but this makes the responsibility of the "account providers". So for now, I'd just keep the `sync` method until we find a better pattern for this. That being said, I do agree that syncing the entire wallet for 1 account can be avoided, so I'll change the logic for now.

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: I did refactor this, lemme know what you think. (I still believe we could do it in a more surgical way, but that should be ok for now).

hmalik88 (approved):

Issue #4892, opened 2024-11-01
[NetworkController] Give consumers a way to get the selected chain ID purely from state


mcmire:
(I will fill this in later, but using state to retrieve the current chain ID is a common pattern in extension and right now this is difficult.)

PR #6168, opened 2025-07-22
Improve sample-controllers to be more useful examples

### LINKED ISSUE #6166: The controllers and services in `sample-controllers` are not as useful as examples as they could be
The controllers and services in `sample-controllers` are not as useful as examples as they could be. They do not follow the best practices that we want others to follow, so we cannot ask teams to copy them, and they do not follow patterns that are standard in the product.

Here is a more exhaustive list of issues we need to fix:

- The `SampleGasPricesService` does not make use of the messenger system.
- The `AbstractSampleGasPricesService` is no longer needed if `SampleGasPricesService` uses the messenger system.
- `SampleGasPricesService` does not follow the patterns described in the [data services pattern](https://github.com/MetaMask/core/blob/5b7f6739e42cc1faf39d9a648f154a309078cb7f/docs/data-services.md). Specifically, it does not use `createServicePolicy` internally.
- The tests for `SampleGasPricesController` and `SamplePetnamesController` do not use the `withController` pattern.
- The tests for `SampleGasPricesController` and `SamplePetnamesController` set up a "root" messenger. We don't use this terminology anywhere else (we either use "global" messenger or "unrestricted" messenger).
- `@metamask/network-controller` types are copied instead of being used directly.
- `fetchGasPrices` in `SampleGasPricesController` does not follow the standard multichain pattern and operates against the globally selected network, so contributors learning how to implement a feature might find this strange.
- `getDefaultGasPricesControllerState` and `getDefaultPetnamesControllerState` are placed in the wrong section in their respective files.
- `getDefaultPetnamesControllerState` is not exported.

mcmire:
## Explanation

The controllers and services in `sample-controllers` are not as useful as examples as they could be. They do not follow the best practices that we want others to follow, so we cannot ask teams to copy them, and they do not follow patterns that are standard in the product.

This commit makes the following changes:

- The `SampleGasPricesService` now makes use of the messenger system.
- `SampleGasPricesController` and `SamplePetnamesController` now exposes their methods as actions, which follows the typical controller.
- The `AbstractSampleGasPricesService` has been removed, and `SampleGasPricesService` is now expected to be used via the messaging system.
- `SampleGasPricesService` now uses the data services pattern, specifically, `createServicePolicy`.
- The tests for `SampleGasPricesController` and `SamplePetnamesController` now use the `withController` pattern.
- The tests for `SampleGasPricesController` and `SamplePetnamesController` now set up a "unrestricted" messenger rather than a "root" messenger.
- `@metamask/network-controller` types are now used directly instead of being copied (it was already a peer dependency).
- `fetchGasPrices` in `SampleGasPricesController` is now called automatically when the globally selected chain changes, and also now takes a chain ID to match.
- `getDefaultGasPricesControllerState` and `getDefaultPetnamesControllerState` are now placed in the "STATE" section.
- `getDefaultPetnamesControllerState` is now exported.

## References

Closes #6166.

## Checklist


packages/sample-controllers/src/index.test.ts
mcmire, line 1: I chose to delete this file because it seemed extraneous. It has never been an official best practice, and it doesn't do a good job of ensuring that changes to exports are highlighted, as it omits TypeScript (and now that we ask teams to not use barrel exports, any changes to exports will show up in the diff for `index.ts`).

packages/sample-controllers/src/index.ts
```diff
@@ -1,25 +1,33 @@
...
   SampleGasPricesControllerGetStateAction,
   SampleGasPricesControllerMessenger,
   SampleGasPricesControllerState,
   SampleGasPricesControllerStateChangeEvent,
+  SampleGasPricesControllerUpdateGasPricesAction,
 } from './sample-gas-prices-controller';
 export {
-  getDefaultSampleGasPricesControllerState,
   SampleGasPricesController,
+  getDefaultSampleGasPricesControllerState,
```
mcmire, line 18: This export was already present, but it was not ordered alphabetically.

packages/sample-controllers/src/network-controller-types.ts
mcmire, line 1: This file was added originally before `sample-controllers` was published to NPM as it wasn't a requirement that SampleGasPricesController be a working example, just illustrative. Since then, `network-controller` has been added as a peer dependency, so we can simply use the types from there.

packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -17,7 +21,7 @@ import type { SampleAbstractGasPricesService } from './sample-gas-prices-service
  * controller's actions and events and to namespace the controller's state data
  * when composed with other controllers.
  */
-export const controllerName = 'SampleGasPricesController';
+export const CONTROLLER_NAME = 'SampleGasPricesController';
```
mcmire: This isn't in line with how we usually name constants. I don't think we need to make an exception for controller/service names.

This variable isn't documented in the controller guidelines at all (yet), so we should be okay to change it and enforce it from now on.

packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -66,85 +69,89 @@ const gasPricesControllerMetadata = {
...
 
-/**
- * Constructs the default {@link SampleGasPricesController} state. This allows
- * consumers to provide a partial state object when initializing the controller
- * and also helps in constructing complete state objects for this controller in
- * tests.
- *
- * @returns The default {@link SampleGasPricesController} state.
- */
-export function getDefaultSampleGasPricesControllerState(): SampleGasPricesControllerState {
```
mcmire, line 142: Not sure why this was added to the "MESSENGER" section originally (probably just an oversight). Regardless this now lives under "STATE".

packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -204,49 +224,77 @@ export class SampleGasPricesController extends BaseController<
...
+    selectedNetworkClientId: NetworkClientId,
+  ) {
+    const {
+      configuration: { chainId },
+    } = this.messagingSystem.call(
+      'NetworkController:getNetworkClientById',
+      selectedNetworkClientId,
+    );
+
+    if (chainId !== this.#selectedChainId) {
```
mcmire, line 295: Ideally we would have a selector in NetworkController that would allow us to listen for when the globally selected _chain_ changes rather than the _network client ID_, but we don't have that, so this is the next best thing.

packages/sample-controllers/src/sample-gas-prices-service/sample-abstract-gas-prices-service.ts
mcmire, line 1: We don't need an abstract type anymore because nothing takes this service anymore; we expect it to be instantiated ahead of time.

packages/sample-controllers/src/index.test.ts
Gudahtt: I thought it was best practice. But that was before the expectation was set around barrel exports, good point. Without barrel exports, removing this sounds good to me.

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
```diff
@@ -11,46 +75,91 @@ type GasPricesResponse = {
...
    *
-   * @param args - The arguments.
-   * @param args.fetch - A function that can be used to make an HTTP request.
-   * If your JavaScript environment supports `fetch` natively, you'll probably
-   * want to pass that; otherwise you can pass an equivalent (such as `fetch`
-   * via `node-fetch`).
+   * @param options - The options. See {@link SampleGasPricesServiceOptions}.
    */
-  constructor({ fetch: fetchFunction }: { fetch: typeof fetch }) {
+  constructor(options: SampleGasPricesServiceOptions) {
```
Gudahtt: Nit: Why declare the type separately here? I know we don't have any guidance against doing this (yet), but I find this pattern really disruptive for reading. It also makes the intellisense less useful (when hovering over the constructor, you need to click this type and review it in a separate tab rather than the parameters being shown directly in the tooltip)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
```diff
@@ -11,46 +75,91 @@ type GasPricesResponse = {
...
+    const { messenger, fetch: fetchFunction, policyOptions = {} } = options;
+
+    this.#messenger = messenger;
     this.#fetch = fetchFunction;
+    this.#policy = createServicePolicy(policyOptions);
+
+    this.#messenger.registerActionHandler(
```
Gudahtt: Nit: should we update these examples to use the new `registerMethodActionHandlers` method?

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
```diff
@@ -61,12 +170,19 @@ export class SampleGasPricesService {
...
-    // Type assertion: We have to assume the shape of the response data.
-    const gasPricesResponse =
-      (await response.json()) as unknown as GasPricesResponse;
-    return gasPricesResponse.data;
+    return await this.#policy.execute(async () => {
+      const response = await this.#fetch(
+        `https://example.com/gas-prices/${chainId}.json`,
+      );
+      if (response.ok) {
+        // Type assertion: We have to assume the shape of the response data.
```
Gudahtt: Nit (pre-existing issue): We can validate, we definitely don't need to assume the shape. Assuming the shape of the response without validation should be strongly discouraged. We've had plenty of examples even with internal APIs of them not returning the expected shape (especially with error responses, which sometimes can be returned even by HTTP 200 responses)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
```diff
@@ -61,12 +170,19 @@ export class SampleGasPricesService {
...
+        const gasPricesResponse =
+          (await response.json()) as unknown as GasPricesResponse;
+        return gasPricesResponse.data;
+      }
+      throw new Error(
```
Gudahtt: This seems broken. The service policy expects HTTP errors to have a `httpStatus` property, so we should be using the `HttpError` error class here or the policy won't work correctly

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
```diff
@@ -11,46 +75,91 @@ type GasPricesResponse = {
...
+  }
+
+  /**
+   * Listens for when the request is retried due to failures.
+   *
+   * @param listener - The callback to be called when the retry occurs.
+   * @returns The same value that {@link ServicePolicy.onRetry} returns.
+   * @see {@link createServicePolicy}
+   */
+  onRetry(listener: Parameters<ServicePolicy['onRetry']>[0]) {
```
Gudahtt, line 177: Should we expose `onDegraded` and `onBreak` as well?

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
Gudahtt: Specifically it's broken in that all HTTP failures are treated like indications of an outage. There is no test of HTTP 400 errors for example, nor is there a test that ensures the circuit doesn't break when it shouldn't.

Gudahtt: We should probably encourage the use of something like `successfulFetch` here. Unfortunately that utility function doesn't accept a custom `fetch` function right now, so we can't do that if we still want to accept that as a parameter.

Gudahtt: It's also very problematic to do the response parsing and validation within the `execute` callback, because it means any validation failure would be treated like a service outage. Essentially if we do anything beyond what `sucessfulFetch` does in the policy execution callback, it's a bug.

packages/sample-controllers/package.json
```diff
@@ -58,6 +58,7 @@
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
     "nock": "^13.3.1",
+    "sinon": "^9.2.4",
```
Gudahtt, line 61: Darn. We should be able to get rid of sinon when we update jest, but until then we need `getFakeTimers` from `sinon` 👍

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
Gudahtt: As is, it seems that we're asking people to duplicate the TSDoc block for each action/method in two places: once for the action type, once for the method. This seems like a bad recommendation.

packages/sample-controllers/src/sample-petnames-controller.test.ts
```diff
@@ -19,160 +22,237 @@ describe('SamplePetnamesController', () => {
...
-      expect(controller.state).toMatchInlineSnapshot(`
-        Object {
-          "namesByChainIdAndAddress": Object {},
-        }
-      `);
     });
   });
 
-  describe('assignPetname', () => {
+  describe.each([
```
Gudahtt: This seems a bit convoluted 🤔 Is it helpful to test the method in both ways? If we test just the method, certainly the tests would miss a problem with the action registration. But if we have tests for the action, it's unclear what value the tests for the method have. Especially if we decide to disallow calling methods directly at some point in the future.

packages/sample-controllers/src/sample-gas-prices-controller.ts
Gudahtt: Something like this should work as the selector:

```
      (networkControllerState) => {
        const entries = Object.entries(
          networkControllerState.networkConfigurationsByChainId,
          // Cast needed because `Object.entries` returns `string` as index type
        ) as [Hex, NetworkConfiguration][];
        const selectedEntry = entries.find(([, networkConfiguration]) =>
          networkConfiguration.rpcEndpoints.some(
            (rpcEndpoint) =>
              rpcEndpoint.networkClientId ===
              networkControllerState.selectedNetworkClientId,
          ),
        );
        if (!selectedEntry) {
          return undefined;
        }
        return selectedEntry[0];
      }
```

It's pretty convoluted though. This does highlight a design problem with the structure of the NetworkController state (common access patterns should be simple and fast)

mcmire: Agreed. I forgot I already made a ticket for this a while back: https://github.com/MetaMask/core/issues/4892

packages/sample-controllers/src/sample-gas-prices-controller.ts
mcmire: I decided to change this back ([d73a7ba](https://github.com/MetaMask/core/pull/6168/commits/d73a7ba32fc1819b196757909715a592e10be1cb)). Ideally of course we would follow a stricter rule here but lots of controllers already use camelcase for the controller or service name and we'd have to change this across the board which I'd rather avoid for now.

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: Usually options are stored in private properties and those properties need to be typed as well. If a separate type exists for options my thought was that I could more easily reuse that type when typing the private properties.

This is pretty minor though and I see your point about making the Intellisense less useful. Updated in [7d53513](https://github.com/MetaMask/core/pull/6168/commits/7d535139e7dc06e3a9f150ed1d660b20c7d466c9).

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: Yes, we might as well. Other teams are starting to hand-write similar patterns. Done here: [7e967d2](https://github.com/MetaMask/core/pull/6168/commits/7e967d20853aa613f6799d91e8929db7db7d1114)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: You're right, but we haven't been validating responses elsewhere as far as I know. What is the pattern that we want to promote here, should we use Superstruct to do so or do it manually? And what should happen if the validation fails, should we throw an error?

packages/sample-controllers/src/sample-petnames-controller.test.ts
mcmire: I have noticed that it is easy when adding an action which directly maps to a method to forget to test the action. (I notice, in fact, that I forgot to do this for `SampleGasPricesService`.) This approach seemed like the best way to ensure that both the method and the action do exactly the same thing.

But if you think this is convoluted, and you think that at some point we will disallow calling methods at some point in the future, perhaps we should promote a pattern where the action is tested comprehensively and the method is tested more simply? (We could of course just test the action, but that seems inconsistent, as there are some cases where actions _don't_ map to methods, and the methods are still technically callable, at least for now.)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: You're right, I misunderstood what the correct pattern should be going forward. Fixed in [31fc78b](https://github.com/MetaMask/core/pull/6168/commits/31fc78b00ea8c2c220ed2e4a66478df3fac68073).

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: I guess if we expect developers to copy and paste this example then that seems fine. Added in [31fc78b](https://github.com/MetaMask/core/pull/6168/commits/31fc78b00ea8c2c220ed2e4a66478df3fac68073).

packages/sample-controllers/src/sample-petnames-controller.test.ts
mcmire: I have updated the tests to primarily test the messenger actions here. Let me know if that fits more with what you were thinking or if you had something else in mind: [b3f9796](https://github.com/MetaMask/core/pull/6168/commits/b3f97967a33681a44880ea845f03128137f9e505)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
Gudahtt: We don't have any prescriptive guidance right now on how to best validate, so that would be up to the team or contributor.

Either throwing an error for invalid inputs, or stripping the data down to the valid parts/normalizing it, may be appropriate depending on the circumstance. Personally I'd default to throwing an error if I wasn't sure how to handle something.

packages/sample-controllers/CHANGELOG.md
```diff
@@ -7,10 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Export `getDefaultPetnamesControllerState` ([#6168](https://github.com/MetaMask/core/pull/6168))
+
 ### Changed
 
-- Bump `@metamask/utils` from `^11.2.0` to `^11.4.2` ([#6054](https://github.com/MetaMask/core/pull/6054))
+- **BREAKING:** The messenger for `SampleGasPricesController` now expects `NetworkController:getNetworkClientById` to be allowed, and no longer expects `NetworkController:getState` to be allowed ([#6168](https://github.com/MetaMask/core/pull/6168))
+- **BREAKING:** `SampleGasPricesController.updateGasPrices` now takes a required `chainId` option ([#6168](https://github.com/MetaMask/core/pull/6168))
+- `SampleGasPricesController` will now automatically update gas prices when the globally selected chain changes ([#6168](https://github.com/MetaMask/core/pull/6168))
 - Bump `@metamask/base-controller` from `^8.0.1` to `^8.1.0` ([#6284](https://github.com/MetaMask/core/pull/6284))
+- Bump `@metamask/controller-utils` from `^11.10.0` to `^11.11.0` ([#6069](https://github.com/MetaMask/core/pull/6069))
```
Gudahtt: This is a dev dependency, and doesn't seem to be changed in this PR. Maybe this was left behind after a branch update somehow?

```suggestion
```

packages/sample-controllers/src/index.ts
```diff
@@ -18,8 +26,8 @@ export type {
   SamplePetnamesControllerState,
   SamplePetnamesControllerStateChangeEvent,
 } from './sample-petnames-controller';
-export { SamplePetnamesController } from './sample-petnames-controller';
 export {
-  SampleGasPricesService,
-  type SampleAbstractGasPricesService,
```
Gudahtt, line 24: Maybe worth mentioning this removal in the changelog?

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
Gudahtt: Looks great, thanks!

packages/sample-controllers/src/sample-petnames-controller.test.ts
Gudahtt: It just seems like _a lot_ for us to ask teams to test both redundantly, when there isn't really a reason for both to exist, you know. The way we wire up controllers, in theory nothing should be able to call any method directly anyway.

Certainly today there are methods that are not actions, but if this is an example of what we're recommending as best practices for new controllers, we need not let those existing inconsistencies restrict our choices here.


We haven't really agreed on preventing direct method calls though, that's just something I had in mind to propose in the future to simplify things. Maybe until we agree on that we should test both? I just didn't feel great about it, you know.

mcmire: Yeah, I know what you mean. We could certainly ask people to stop testing methods directly that are being exposed via the messenger. I worry that this guidance would be out of place at the moment, though. As it is, we _do_ use methods directly, both in the background API and in Mobile. So at the moment, it is useful, even if it's not ideal. But I see your point. I'll think about this some more.

packages/sample-controllers/src/index.ts
```diff
@@ -1,3 +1,10 @@
+export type {
+  SampleGasPricesServiceActions,
+  SampleGasPricesServiceEvents,
+  SampleGasPricesServiceMessenger,
+} from './sample-gas-prices-service/sample-gas-prices-service';
+export type { SampleGasPricesServiceFetchGasPricesAction } from './sample-gas-prices-service/sample-gas-prices-service-method-action-types';
```
mcmire, line 6: These import paths are a bit annoying, now that I look at it. I'm curious whether flattening them is a good idea. Maybe I'll make another PR that organizes these files better.

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.test.ts
```diff
@@ -26,3 +294,65 @@ describe('SampleGasPricesService', () => {
     });
   });
 });
+
+/**
+ * The type of the messenger where all actions and events will be registered.
+ */
+type UnrestrictedMessenger = Messenger<
```
mcmire: I need to follow the same changes in https://github.com/MetaMask/core/pull/6335 with regard to setting up the messenger in tests.

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
```diff
@@ -59,101 +48,356 @@ describe('SampleGasPricesController', () => {
...
+    });
+  });
 });
 
 /**
- * The union of actions that the root messenger allows.
+ * The type of the messenger where all actions and events will be registered.
  */
-type RootAction = ExtractAvailableAction<SampleGasPricesControllerMessenger>;
+type UnrestrictedMessenger = Messenger<
```
mcmire: I need to follow the same changes in https://github.com/MetaMask/core/pull/6335 with regard to setting up the messenger in tests.

packages/sample-controllers/src/sample-petnames-controller.test.ts
```diff
@@ -19,162 +19,248 @@ describe('SamplePetnamesController', () => {
...
+      );
     });
   });
 });
 
 /**
- * The union of actions that the root messenger allows.
+ * The type of the messenger where all actions and events will be registered.
+ */
+type UnrestrictedMessenger = Messenger<
```
mcmire: I need to follow the same changes in https://github.com/MetaMask/core/pull/6335 with regard to setting up the messenger in tests.

mcmire:
The changes in https://github.com/MetaMask/core/pull/6335 will probably cause conflicts here, so I will wait until that's merged and then revisit this PR.

packages/sample-controllers/src/index.ts
mcmire: You're right. Added changelog entry here: [525c4d8](https://github.com/MetaMask/core/pull/6168/commits/525c4d8ae053506bd67fea92587a2f120115b384)

packages/sample-controllers/CHANGELOG.md
mcmire: Hmm, yes, probably a bad merge. Removed in [9d0799e](https://github.com/MetaMask/core/pull/6168/commits/9d0799eedd91e10cf91558953ed80001dcff3d53).

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts
mcmire: Okay, I've added some basic validation here: [8f9a6b9](https://github.com/MetaMask/core/pull/6168/commits/8f9a6b9dac2696a022c269cac96d369e379f941e)

packages/sample-controllers/src/sample-petnames-controller.test.ts
mcmire: I've updated the JSDocs to encourage people to use the controllers and service through the messenger, but left the method and its test in place for now since they have practical use: [0abf441](https://github.com/MetaMask/core/pull/6168/commits/0abf44122754292c172a1cd2c34b46b214010af9). The tests for the methods are extremely simple on purpose, so it shouldn't be a huge burden if people were to copy those examples. 

Once we have a plan for simplifying the background API as well as guidance for using the messenger from the UI, we should be able to completely drop methods along with their tests and promote the messenger as the one true way to use controllers and services.

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
mcmire: Done here: [0240763](https://github.com/MetaMask/core/pull/6168/commits/024076357229fde64bb0a0d5e79478edbb22c9d6)

packages/sample-controllers/src/sample-petnames-controller.test.ts
mcmire: Done here: [0240763](https://github.com/MetaMask/core/pull/6168/commits/024076357229fde64bb0a0d5e79478edbb22c9d6)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.test.ts
mcmire: Done here: [0240763](https://github.com/MetaMask/core/pull/6168/commits/024076357229fde64bb0a0d5e79478edbb22c9d6)

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service-method-action-types.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * This file is auto generated by `scripts/generate-method-action-types.ts`.
```
mcmire, line 2: One thing I noticed is that method action types for services are not generated automatically. I had to add this one myself. I will make a ticket to fix the script.

mcmire:
Ready for review again.

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service-method-action-types.ts
mcmire: I've made a ticket for this here: https://github.com/MetaMask/core/issues/6407

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.test.ts
```diff
@@ -1,22 +1,283 @@
...
+      );
+
+      expect(gasPricesResponse).toStrictEqual({
+        low: 5,
+        average: 10,
+        high: 15,
+      });
+    });
+
+    it.each([
```
Gudahtt, line 51: Nice. This test is satisfyingly compact. Great to see response validation in the example service as well.

Gudahtt (dismissed):
LGTM!

packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.test.ts
```diff
@@ -26,3 +287,68 @@ describe('SampleGasPricesService', () => {
...
+  ExtractAvailableEvent<SampleGasPricesServiceMessenger>
+>;
+
+/**
+ * Constructs the messenger populated with all external actions and events
+ * required by the service under test.
+ *
+ * @returns The root messenger.
+ */
+function buildRootMessenger(): RootMessenger {
```
mcmire: Dang, I just noticed that this should be `getRootMessenger` to align with controller tests. I will change this tomorrow.

mcmire: This is now fixed: [7a97b62](https://github.com/MetaMask/core/pull/6168/commits/7a97b625081284b1927ff7c65d098668410d9ee8)

PR #6169, opened 2025-07-23
feat: configuration passwordOutdatedCache in the controller

lwin-kyaw:
## Explanation

This PR adds optional `passwordOutdatedCache` (in ms) to the `SeedlessOnboardingController` constructor params.
If user provides `passwordOutdatedCache`, we will use this value when setting cache for the `passwordOutdatedCheck`, otherwise the default value (10 seconds) will be used.

## Checklist


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -656,6 +656,29 @@ describe('SeedlessOnboardingController', () => {
         },
       );
     });
+
+    it('should throw an error if the password outdated cache TTL is not a valid number or string', () => {
```
matthiasgeihs:
```suggestion
    it('should throw an error if the password outdated cache TTL is not a valid number', () => {
```

packages/seedless-onboarding-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added an optional param, `passwordOutdatedCacheTTL` to the constructor params and exported `SecretMetadata` class from the controller.([#6169](https://github.com/MetaMask/core/pull/6169))
```
matthiasgeihs: nit
```suggestion
- Added an optional parameter, `passwordOutdatedCacheTTL` to the constructor and exported `SecretMetadata` class from the controller.([#6169](https://github.com/MetaMask/core/pull/6169))
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -198,7 +208,11 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
       messenger,
     });
 
+    assertIsPasswordOutdatedCacheValid(passwordOutdatedCacheTTL);
```
matthiasgeihs, line 211: nit:
i feel like it's a bit overkill to verify the type in the assert function.
can't we rely on the typescript compiler here?

packages/seedless-onboarding-controller/src/index.ts
```diff
@@ -21,4 +21,5 @@ export {
   AuthConnection,
   SecretType,
 } from './constants';
+export { SecretMetadata } from './SecretMetadata';
```
matthiasgeihs, line 24: why this?

matthiasgeihs (approved):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: We can't rely on the typescript complier if the user is not using typescript. For example, in the extension, where controller initialization is done from the js env.

packages/seedless-onboarding-controller/src/index.ts
lwin-kyaw: This is exported, so that user can use this in testings (and mockings).

tuna1207 (approved):

matthiasgeihs (approved):

PR #6178, opened 2025-07-24
fix: Preserve `origin` and `type` when adding batch with hook

OGPoyraz:
## Explanation

This PR aims to preserve `origin` and `type` property when adding a batch with hook.

As part of this PR; also adding an optional parameter `fiatValue` which then will be later to matched with metrics for transactions like swaps and bridge.

Note: `origin` is already preserved while adding batch 7702.

## References

* Fixes part of: https://github.com/MetaMask/MetaMask-planning/issues/5353 and
* Fixes part of: https://github.com/MetaMask/MetaMask-planning/issues/5355

* Related to (will be consumed by): https://github.com/MetaMask/metamask-extension/pull/34594/

## Checklist


packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1132,6 +1133,7 @@ export class TransactionController extends BaseController<
       batchId?: Hex;
       deviceConfirmedOn?: WalletDevice;
       disableGasBuffer?: boolean;
+      fiatValue?: string;
```
OGPoyraz: This is part of https://github.com/MetaMask/MetaMask-planning/issues/5355

packages/transaction-controller/src/hooks/ExtraTransactionsPublishHook.ts
```diff
@@ -88,9 +88,13 @@ export class ExtraTransactionsPublishHook {
...
 
     const extraTransactions: TransactionBatchSingleRequest[] =
-      this.#transactions.map((transaction) => ({
-        params: transaction,
-      }));
+      this.#transactions.map((transaction) => {
+        const { type, ...rest } = transaction;
+        return {
+          params: rest,
+          type,
```
OGPoyraz, line 96: This is preserving `type` for batch transactions with hook.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -455,6 +456,10 @@ async function addTransactionBatchWithHook(
 
   let txBatchMeta: TransactionBatchMeta | undefined;
   const batchId = generateBatchId();
+  const nestedTransactions = requestedTransactions.map((tx) => ({
+    ...tx,
+    origin,
```
OGPoyraz, line 461: Missing `origin` in every transactions inside batch transactions.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -604,7 +610,9 @@ async function processTransactionWithHook(
     {
       batchId,
       disableGasBuffer: true,
+      fiatValue,
```
OGPoyraz: `fiatValue` also needs to be passed individually - so swaps can add batch transactions and define `fiatValue` per batched transactions.

OGPoyraz:
@metamaskbot publish-preview

packages/transaction-controller/src/TransactionController.ts
matthewwalsh0: We should definitely add a dedicated `type` for this in future so we can add JSDoc on each property.

matthewwalsh0: Do we definitely want a string? And would that be pre-formatted so `$1.23` or do we want Hex or a decimal string etc?

packages/transaction-controller/src/types.ts
```diff
@@ -170,6 +170,11 @@ export type TransactionMeta = {
    */
   estimateUsed?: string;
 
+  /**
+   * The fiat value of the transaction to be used to passed metrics.
+   */
```
matthewwalsh0: Can we be explicit about the expected format here as mentioned in the other comment?

packages/transaction-controller/src/types.ts
```diff
@@ -170,6 +170,11 @@ export type TransactionMeta = {
    */
   estimateUsed?: string;
 
+  /**
+   * The fiat value of the transaction to be used to passed metrics.
```
matthewwalsh0:
```suggestion
   * The total fiat value of the transaction, to support client metrics.
```

packages/transaction-controller/src/types.ts
```diff
@@ -170,6 +170,11 @@ export type TransactionMeta = {
...
+   */
+  fiatValue?: string;
```
matthewwalsh0: Do we need two properties, one for receiving and one for sending?

packages/transaction-controller/src/types.ts
```diff
@@ -1556,6 +1561,9 @@ export type BatchTransactionParams = {
   /** Address of the target contract or EOA. */
   to?: Hex;
 
+  /** Type of the transaction. */
+  type?: TransactionType;
```
matthewwalsh0: Rather than adding this to this parameter specific type, should we instead change `batchTransactions` on `TransactionMeta` to be an array of `NestedTransactionMetadata`?

As that is already this type and the transaction type and it keeps the parameters themselves separate.

We could rename `NestedTransactionMetadata` in future but not now, to avoid the breaking change.

packages/transaction-controller/src/types.ts
```diff
@@ -1585,6 +1593,12 @@ export type TransactionBatchSingleRequest = {
     signedTransaction: Hex;
   };
 
+  /** Fiat value of the nested transaction. */
+  fiatValue?: string;
+
+  /** Origin of the nested transaction. */
+  origin?: string;
```
matthewwalsh0: Why do we need this per transaction, as we already have `origin` on the parent `TransactionBatchRequest`?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: Agreed, this is already a big type as `addTransaction` is one of the most used functionality.

OGPoyraz: As I understood we directly want to map it to `simulation_receiving_assets_total_value` so that's why I left it as direct string. (also assumed a fiat value so not a hex)

In personal opinion - if we are always sending `simulation_sending_assets_total_value` metric values as in USD, then we should've tell it in the property names. But Barbara said it's fine to keep it as is.

packages/transaction-controller/src/types.ts
OGPoyraz: Adding `assetsFiatValues` object consist of `sending` `recieving` props.

packages/transaction-controller/src/types.ts
OGPoyraz: Good call - removed this from `TransactionBatchSingleRequest` and mapped to `userRequest.origin` instead of reading from `nestedTransaction`

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: Moved this into types and exported

packages/transaction-controller/src/types.ts
```diff
@@ -1586,7 +1594,7 @@ export type TransactionBatchSingleRequest = {
   };
 
   /** Parameters of the single transaction. */
-  params: BatchTransactionParams;
+  params: NestedTransactionMetadata;
```
matthewwalsh0: This is still the parameters only so we shouldn't need to change the type?

packages/transaction-controller/src/types.ts
```diff
@@ -1877,3 +1885,61 @@ export type BeforeSignHook = (request: {
     }
   | undefined
 >;
+
+/**
+ * The total fiat values of the transaction, to support client metrics.
+ */
+export type AssetFiatValues = {
+  /**
+   * The fiat value of the receiving asset.
```
matthewwalsh0: Minor, `assets`?

packages/transaction-controller/src/types.ts
```diff
@@ -1877,3 +1885,61 @@ export type BeforeSignHook = (request: {
...
+  /**
+   * The fiat value of the sending asset.
+   */
+  sending?: string;
+};
+
+/**
+ * Options for adding a transaction.
+ */
+export type AddTransactionOptions = {
```
matthewwalsh0: Could we create this type in a dedicated PR so we can take the time to check each of the descriptions, and unblock this PR?

There is lots of them 😄

matthewwalsh0 (changes requested):

packages/transaction-controller/src/types.ts
OGPoyraz: Sure no problem :)

packages/transaction-controller/src/types.ts
OGPoyraz: Done

packages/transaction-controller/src/types.ts
OGPoyraz: Unintended one - changed back

PR #6184, opened 2025-07-24
call revoke token in background

himanshuchawla009:
## Explanation

- This PR moves revoke token call in background. Revoking refresh token  is not a must after submit password so can be done in background and even when it fails it should affect user wallet unlock.

## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
@@ -654,16 +654,27 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
         revokeToken,
       } = await this.#unlockVaultAndGetVaultData(password);
       this.#setUnlocked();
-
+      await this.#createNewVaultWithAuthData({
+        password,
+        rawToprfEncryptionKey: toprfEncryptionKey,
+        rawToprfPwEncryptionKey: toprfPwEncryptionKey,
+        rawToprfAuthKeyPair: toprfAuthKeyPair,
+      });
       if (revokeToken) {
-        await this.#revokeRefreshTokenAndUpdateState(revokeToken);
-        // re-creating vault to persist the new revoke token
-        await this.#createNewVaultWithAuthData({
-          password,
-          rawToprfEncryptionKey: toprfEncryptionKey,
-          rawToprfPwEncryptionKey: toprfPwEncryptionKey,
-          rawToprfAuthKeyPair: toprfAuthKeyPair,
-        });
+        // this call is not critical for unlocking, so we can do it in the background without await.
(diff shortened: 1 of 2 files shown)


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -654,16 +654,27 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
         revokeToken,
       } = await this.#unlockVaultAndGetVaultData(password);
       this.#setUnlocked();
-
+      await this.#createNewVaultWithAuthData({
+        password,
+        rawToprfEncryptionKey: toprfEncryptionKey,
+        rawToprfPwEncryptionKey: toprfPwEncryptionKey,
+        rawToprfAuthKeyPair: toprfAuthKeyPair,
+      });
```
matthiasgeihs, line 662: note that `#createNewVaultWithAuthData` is potentially an expensive operation.
it runs `encryptWithDetail(pw, vaultData)`, which involves the intentionally slow pbkdf2 key derivation step.

as you are increasing the usage of vault encryption, you are driving up cpu usage.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -654,16 +654,27 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+            return this.#createNewVaultWithAuthData({
+              password,
+              rawToprfEncryptionKey: toprfEncryptionKey,
+              rawToprfPwEncryptionKey: toprfPwEncryptionKey,
+              rawToprfAuthKeyPair: toprfAuthKeyPair,
+            });
+          })
+          .catch((error) => {
+            log('Error revoking refresh token', error);
+          });
```
matthiasgeihs, line 677: writing the vault asynchronously like this is dangerous as it can lead to vault corruption.
imagine another piece of code updates the vault in the meantime. when this code here is executed after that, the vault will be overwritten with some old data.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -759,18 +770,30 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+            return this.#createNewVaultWithAuthData({
+              password: globalPassword,
+              rawToprfEncryptionKey: latestEncKey,
+              rawToprfPwEncryptionKey: latestPwEncKey,
+              rawToprfAuthKeyPair: latestAuthKeyPair,
+            });
+          })
+          .catch((error) => {
+            log('Error revoking refresh token', error);
+          });
```
matthiasgeihs, line 795: see comment above, we can't do this

PR #6185, opened 2025-07-24
feat(account-tree-controller): multichain account support + use 1 group per account

ccharly:
## Explanation

TODO

## References

N/A

## Changelog

N/A

## Checklist


packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -0,0 +1,111 @@
...
+    return {
+      walletId,
+      groupId,
+    };
+  }
+
+  getDefaultAccountWalletName(wallet: AccountTreeWallet): string {
+    // Precondition: This method is invoked only if there was a match for
+    // a BIP-44 account, so we can type-cast here.
+    const account = wallet.getAnyAccount() as Bip44Account<InternalAccount>;
```
ccharly: Given that the renaming phase must come after we have constructed the account tree, we have to re-get some "common information" back from the accounts (e.g `groupIndex` for `Entropy` wallets, `snapId` for `Snap` wallets, etc...)

For the moment, I query any account to compute those information, but a better pattern could be to introduce "typed options" under account wallet and account group (similarly to `account.options`) so we could easily access those informations from those nodes without relying on the accounts themselves.

It would also align with the way the "rules" are grouping accounts together. You cannot have diverging options, otherwise this would be a different wallet/group.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -207,14 +250,39 @@ export class AccountTreeController extends BaseController<
...
+            const group = wallet.groups[groupId];
+            if (group) {
+              const groupInstance = walletInstance.getAccountGroup(group.id);
+              if (groupInstance) {
+                if (group.metadata.name === '') {
+                  this.#renameAccountGroup(groupInstance, group);
+                }
+              }
+            }
+          }
```
ccharly: This feels really wrong to me to have all those `if`. Another alternative would be to `throw` if we cannot get the wallet/group instance. Normally instances and the state are always in sync, so this should be safe.

packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -0,0 +1,111 @@
...
+        `! Found an unknown entropy ID: "${entropySource}", account "${account.id}" won't be grouped by entropy.`,
+      );
+      return undefined;
+    }
+
+    const walletId = toMultichainAccountWalletId(account.options.entropy.id);
+    const groupId = toMultichainAccountId(
+      walletId,
+      account.options.entropy.groupIndex,
+    );
```
ccharly: I initially wanted to wire up the `MultichainAccountService` with this, to avoid dealing with crafting those IDs here.

This would also abstract away the "BIP-44" implementation that is the responsibility of that service.

Though, wallet names belongs here. So we somehow still need to resolve the entropy source and their indexes.

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -0,0 +1,111 @@
...
+export class EntropyRule extends Rule {
+  readonly category = AccountWalletCategory.Entropy;
+
+  readonly #entropySourcesToIndex: Map<EntropySourceId, number>;
+
+  constructor(messenger: AccountTreeControllerMessenger) {
+    super(messenger);
+
+    this.#entropySourcesToIndex = new Map();
+    this.#syncEntropySources();
```
ccharly: Actually, this might be a mistake. Keyrings might not be available at that time (if extension is still locked).

I'll move this to a runtime call to solve that issue.

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/rules/rule.ts
```diff
@@ -0,0 +1,50 @@
+import type { AccountWalletCategory } from '@metamask/account-api';
+import type { InternalAccount } from '@metamask/keyring-internal-api';
+
+import type { AccountTreeGroup, AccountTreeWallet } from '..';
+import type { AccountTreeControllerMessenger } from '../AccountTreeController';
+import type { AccountContext } from '../types';
+
+export abstract class Rule {
```
gantunesr: The API of this class looks weird and it may be because I don't fully follow what is happening with it. Specifically the usage of `getDefaultAccountWalletName` and `getDefaultAccountGroupName`

packages/account-tree-controller/src/rules/rule.ts
```diff
@@ -0,0 +1,50 @@
...
+  abstract match(account: InternalAccount): AccountContext | undefined;
+
+  /**
+   * Gets default name for a wallet.
+   *
+   * @param wallet - Wallet associated to this rule.
+   * @param context - Rule context.
+   * @returns The default name for that wallet.
+   */
+  abstract getDefaultAccountWalletName(wallet: AccountTreeWallet): string;
```
gantunesr: Why default? This implies that we want to leave room for a `getAccountWalletName` and `getAccountGroupName`

packages/account-tree-controller/src/rules/keyring.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+}
+
+export class KeyringRule extends AccountTreeRule {
+  readonly category = AccountWalletCategory.Keyring;
+
+  match(account: InternalAccount): AccountTreeRuleResult | undefined {
+    // We assume that `type` is really a `KeyringTypes`.
+    const type = account.metadata.keyring.type as KeyringTypes;
```
gantunesr, line 63: why assume and not return undefined in the edge case that is not?

packages/account-tree-controller/src/rules/snap.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  isSnapAccount(
+    account: InternalAccount,
+  ): account is SnapAccount<InternalAccount> {
+    return (
+      account.metadata.keyring.type === (KeyringTypes.snap as string) &&
+      account.metadata.snap !== undefined &&
+      account.metadata.snap.enabled
+    );
+  }
```
gantunesr, line 38: Don't we have already a method that does this? Its probably just in the clients

packages/account-tree-controller/src/rules/snap.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  getDefaultAccountWalletName(wallet: AccountTreeWallet): string {
+    // Precondition: We assume the AccountTreeController will always use
+    // the proper wallet instance.
+    const options = wallet.options as AccountTreeWalletSnapOptions;
+
+    const snap = this.messenger.call('SnapController:get', options.snap.id);
+    const snapName = snap
+      ? // TODO: Handle localization here, but that's a "client thing", so we don't have a `core` controller
+        // to refer to.
```
gantunesr, line 75: What do you want to localize? The snap name?

packages/account-tree-controller/src/AccountTreeController.ts
gantunesr: I also prefer to `throw`/`return` instead of deep nested ifs

packages/account-tree-controller/src/rules/keyring.ts
gantunesr: nvm just saw why

packages/account-tree-controller/src/rules/entropy.ts
ccharly: Ended-up making a direct call everytime. That feels a bit overkill to listen to the `KeyringController` state change anyway. Much more concise that way.

packages/account-tree-controller/src/rules/rule.ts
ccharly: Actually, I don't really like those `abstract` either... I might have another idea for this.

Also, the idea of having the "naming logic" on the rules was to have a central place for each logic (since we do not accounts the same way for multichain accounts compared to the other account groups)

But I also feel like the responsibility/split isn't perfect either... I'll see what I can do!

packages/account-tree-controller/src/rules/rule.ts
ccharly: I went for "Default" because we might persist names in the state at some point, meaning we would need default names only when a new wallet/group is being added, afterward their names will be persisted.

packages/account-tree-controller/src/rules/snap.ts
ccharly: We actually don't have one in the "core world" yes! I thought of adding a `isSnapInternalAccount` on our new `account-api` cause that could be useful elsewhere too! (and yes "Internal" because we need the `metadata` and `KeyringAccount` does not have any)

packages/account-tree-controller/src/rules/snap.ts
ccharly: Yes. We have ways of localizing names with Snaps, but this requires some "client logic" to be applied here, since we don't know what's the current locale in the "core world".

packages/account-tree-controller/src/rules/snap.ts
ccharly: One of the main problem too, is that `KeyringTypes` is not part of our packages (it lives in the `keyring-controller` package). We might want to move it someday so we can use it!

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -207,14 +246,39 @@ export class AccountTreeController extends BaseController<
...
+        }
+      }
```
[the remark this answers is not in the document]
ccharly: This is valid concern. And IMO, we could split the `#insert` logic a bit more (this is not new, we were already using messengers when grouping accounts).

I believe we could (outside of `this.update`):
1. Run all rules to get an `AccountContext` (no insertion)
2. Get names for groups and wallets if they have to be renamed
3. (Inside `this.update`) Make the insertion of the account/wallet/group if they are missing

We can do this later when refactoring the rule logic

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamasbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1099,3 +1495,73 @@ describe('AccountTreeController', () => {
...
+      'AccountsController:getAccount',
+      () => undefined,
+    );
+
+    expect(() => rule.getOnlyAccountFrom(group)).toThrow(
+      `Unable to get account with ID: "${account.id}"`,
+    );
+  });
+
+  it('throws if there is not enough account', () => {
```
gantunesr:
```suggestion
  it('throws if there is not enough accounts', () => {
```

gantunesr (approved):

PR #6186, opened 2025-07-24
feat(account-tree): add `selectedAccountGroup`

PatrykLucka:

The AccountTreeController organizes accounts into groups but has no concept of which group is currently selected. Meanwhile, AccountsController tracks the selected individual account. This creates a disconnect between account selection at the individual level and group level.

Added `selectedAccountGroup` state to AccountTreeController with bidirectional synchronization:

- **Initialization**: Sets selectedAccountGroup based on AccountsController's currently selected account
- **AccountsController → AccountTreeController**: Listens to `selectedAccountChange` events and updates selectedAccountGroup to match the selected account's group
- **AccountTreeController → AccountsController**: When `setSelectedAccountGroup` is called, selects the first account in that group via AccountsController

The sync is idempotent to prevent infinite loops - changes only occur when the target state differs from current state.

- `#getFirstAccountInGroup` method - when selecting a group, we need to pick a specific account to tell AccountsController about, so we choose the first one
## Checklist


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +332,111 @@ export class AccountTreeController extends BaseController<
...
+        const group = (wallet.groups as Record<string, AccountGroupObject>)[
+          accountGroupId
+        ];
+        if (group && group.accounts.length > 0) {
+          return group.accounts[0];
+        }
+      }
+    }
+    return undefined;
+  }
```
ccharly: NOTE: This will be changed to `AccountTreeGroup.getOnlyAccount` once this PR is merged:
- https://github.com/MetaMask/core/pull/6185

For now, we can keep that one, I'll make the change 👍

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -194,6 +228,27 @@ export class AccountTreeController extends BaseController<
     this.update((state) => {
       state.accountTree.wallets = wallets;
     });
+
+    // Initialize selectedAccountGroup based on currently selected account
+    this.#initializeSelectedAccountGroup();
```
ccharly: To avoid 2 `update` blocks. Maybe we could do something like this instead, also, we should keep the currently selected account group if it's set already? 🤔:

```ts
    this.update((state) => {
      state.accountTree.wallets = wallets;
      
      if (state.selectedAccountGroup === '') {
        // No group is selected yet, re-sync with the AccountsController.
        state.selectedAccountGroup = this.#getDefaultSelectedAccountGroup();
      }
    });
```

And `#getDefaultSelectedAccountGroup` (btw, we can pick a better name):
```ts
  /**
   * Initializes the selectedAccountGroup based on the currently selected account from AccountsController.
   */
  #getDefaultSelectedAccountGroup(): AccountGroupId | '' {
    const selectedAccount = this.messagingSystem.call(
      'AccountsController:getSelectedAccount',
    );
    if (selectedAccount && selectedAccount.id) {
      const accountMapping = this.#reverse.get(selectedAccount.id);
      if (accountMapping) {
        const { groupId } = accountMapping;
        
        return groupId;
      }
    }
    
    // QUESTION: Should we default to the first wallet, first group in case of errors?
    return '';
  }
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `selectedAccountGroup` state and bidirectional synchronization with AccountsController ([#XXXX](https://github.com/MetaMask/core/pull/6186))
```
```suggestion
- Add `selectedAccountGroup` state and bidirectional synchronization with AccountsController ([#6186](https://github.com/MetaMask/core/pull/6186))
```

ccharly (dismissed):

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -193,9 +223,40 @@ export class AccountTreeController extends BaseController<
...
+      const accountMapping = this.#reverse.get(selectedAccount.id);
+      if (accountMapping) {
+        const { groupId } = accountMapping;
+
+        return groupId;
+      }
+    }
+
+    // Default to the first wallet, first group in case of errors.
+    return Object.values(Object.values(wallets)[0]?.groups ?? {})[0]?.id ?? '';
```
ccharly: We could re-use `#findFirstNonEmptyGroup` here?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +353,130 @@ export class AccountTreeController extends BaseController<
...
+   *
+   * @param accountGroupId - The account group ID.
+   * @returns The first account ID in the group, or undefined if no accounts found.
+   */
+  #getFirstAccountInGroup(
+    accountGroupId: AccountGroupId,
+  ): AccountId | undefined {
+    for (const wallet of Object.values(
+      this.state.accountTree.wallets,
+    ) as AccountWalletObject[]) {
```
ccharly: Do we really need to type cast here? 🤔

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +353,130 @@ export class AccountTreeController extends BaseController<
...
+      if (Object.hasOwnProperty.call(wallet.groups, accountGroupId)) {
+        const group = (wallet.groups as Record<string, AccountGroupObject>)[
```
ccharly: Same here? That does not seem necessary

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +353,130 @@ export class AccountTreeController extends BaseController<
...
+   * @param accountGroupId - The account group ID.
+   * @returns The first account ID in the group, or undefined if no accounts found.
+   */
+  #getFirstAccountInGroup(
+    accountGroupId: AccountGroupId,
+  ): AccountId | undefined {
+    for (const wallet of Object.values(
+      this.state.accountTree.wallets,
+    ) as AccountWalletObject[]) {
+      if (Object.hasOwnProperty.call(wallet.groups, accountGroupId)) {
```
ccharly: I feel like `wallet.groups[accountGroupId]` feels more natural, WDYT?

PatrykLucka: agree

ccharly: Though, `hasOwnProperty` might be more "correct" in this context, I don't think we're too strict about this, and group IDs are not valid JS/TS identifiers, so it cannot be mapped to an existing fields on `Object` IMO

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +353,130 @@ export class AccountTreeController extends BaseController<
...
+  }
+
+  /**
+   * Gets the first account ID in the specified group.
+   *
+   * @param accountGroupId - The account group ID.
+   * @returns The first account ID in the group, or undefined if no accounts found.
+   */
+  #getFirstAccountInGroup(
+    accountGroupId: AccountGroupId,
```
ccharly: WDYT about renaming it to `groupId` for consistency?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,4 +354,124 @@ export class AccountTreeController extends BaseController<
...
+  getSelectedAccountGroup(): AccountGroupId | '' {
+    return this.state.accountTree.selectedAccountGroup;
+  }
+
+  /**
+   * Sets the selected account group and updates the AccountsController selectedAccount accordingly.
+   *
+   * @param accountGroupId - The account group ID to select.
+   */
+  setSelectedAccountGroup(accountGroupId: AccountGroupId): void {
```
ccharly: Cosmetic: We could use `groupId` here too, to be consistent?

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -706,4 +744,376 @@ describe('AccountTreeController', () => {
...
+
+      expect(() => {
+        messenger.publish(
+          'AccountsController:accountRemoved',
+          'definitely-not-a-real-account-id',
+        );
+      }).not.toThrow();
+
+      expect(controller.getSelectedAccountGroup()).not.toBe('');
+    });
```
ccharly: Those 2 are duplicates no?

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `selectedAccountGroup` state and bidirectional synchronization with AccountsController ([#6186](https://github.com/MetaMask/core/pull/6186))
+  - New `getSelectedAccountGroup()` and `setSelectedAccountGroup()` methods
+  - Automatic synchronization when selected account changes in AccountsController
+  - New action types `AccountTreeControllerGetSelectedAccountGroupAction` and `AccountTreeControllerSetSelectedAccountGroupAction`
```
ccharly: Very nit: We can fix that during release.

```suggestion
  - New `getSelectedAccountGroup()` and `setSelectedAccountGroup()` methods.
  - Automatic synchronization when selected account changes in `AccountsController`.
  - New action types `AccountTreeControllerGetSelectedAccountGroupAction` and `AccountTreeControllerSetSelectedAccountGroupAction`.
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `selectedAccountGroup` state and bidirectional synchronization with AccountsController ([#6186](https://github.com/MetaMask/core/pull/6186))
```
```suggestion
- Add `selectedAccountGroup` state and bidirectional synchronization with `AccountsController` ([#6186](https://github.com/MetaMask/core/pull/6186))
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -29,6 +30,23 @@ import {
...
 import { getAccountWalletNameFromKeyringType } from './rules/KeyringWalletRule';
 
+// Local mock of EMPTY_ACCOUNT to avoid circular dependency
+const EMPTY_ACCOUNT_MOCK: InternalAccount = {
+  id: '',
+  address: '',
+  options: {},
+  methods: [],
+  type: EthAccountType.Eoa,
+  scopes: [EthScope.Eoa],
```
Vannatay8, line 40: Yggh

PR #6187, opened 2025-07-24
feat: adds revokeRefreshToken public function.

himanshuchawla009:
## Explanation

- This PR moves revoke token call in a separate public function. Revoking refresh token  is not a must after submit password so can be done post unlock and even when it fails it shouldn't affect user wallet unlock.

## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
@@ -4904,7 +4904,7 @@ describe('SeedlessOnboardingController', () => {
     });
   });
 
-  describe('revokeRefreshTokenAndUpdateVault', () => {
+  describe('revokeRefreshToken', () => {
     const MOCK_PASSWORD = 'mock-password';
     const CURRENT_REVOKE_TOKEN = 'current-revoke-token';
     const NEW_REVOKE_TOKEN = 'new-revoke-token';
@@ -4956,7 +4956,7 @@ describe('SeedlessOnboardingController', () => {
 
           const encryptorSpy = jest.spyOn(encryptor, 'encryptWithDetail');
 
-          await controller.revokeRefreshTokenAndUpdateVault(MOCK_PASSWORD);
+          await controller.revokeRefreshToken(MOCK_PASSWORD);
 
           // Verify that revokeRefreshToken was called with correct parameters
           expect(mockRevokeRefreshToken).toHaveBeenCalledWith({
@@ -5011,7 +5011,7 @@ describe('SeedlessOnboardingController', () => {
         },
         async ({ controller }) => {
           await expect(
-            controller.revokeRefreshTokenAndUpdateVault(MOCK_PASSWORD),
+            controller.revokeRefreshToken(MOCK_PASSWORD),
           ).rejects.toThrow(
             SeedlessOnboardingControllerErrorMessage.InvalidRevokeToken,
           );
@@ -5037,7 +5037,7 @@ describe('SeedlessOnboardingController', () => {
           );
 
           await expect(
-            controller.revokeRefreshTokenAndUpdateVault(MOCK_PASSWORD),
+            controller.revokeRefreshToken(MOCK_PASSWORD),
           ).rejects.toThrow('Failed to revoke refresh token');
 
           expect(mockRevokeRefreshToken).toHaveBeenCalledWith({
@@ -5065,7 +5065,7 @@ describe('SeedlessOnboardingController', () => {
             .mockRejectedValueOnce(new Error('Failed to decrypt vault'));
 
           await expect(
-            controller.revokeRefreshTokenAndUpdateVault(MOCK_PASSWORD),
+            controller.revokeRefreshToken(MOCK_PASSWORD),
           ).rejects.toThrow('Failed to decrypt vault');
         },
       );
@@ -5094,7 +5094,7 @@ describe('SeedlessOnboardingController', () => {
             .mockRejectedValueOnce(new Error('Failed to encrypt vault'));
 
           await expect(
-            controller.revokeRefreshTokenAndUpdateVault(MOCK_PASSWORD),
+            controller.revokeRefreshToken(MOCK_PASSWORD),
           ).rejects.toThrow('Failed to encrypt vault');
 
           expect(mockRevokeRefreshToken).toHaveBeenCalled();
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
@@ -3742,7 +3742,6 @@ describe('SeedlessOnboardingController', () => {
     let initialAuthKeyPair: KeyPair; // Store initial keypair for vault creation
     let initialEncKey: Uint8Array; // Store initial encKey for vault creation
     let initialPwEncKey: Uint8Array; // Store initial pwEncKey for vault creation
-    let initialSeedlessEncKey: Uint8Array; // Store initial seedlessEncKey for vault creation
     let initialEncryptedSeedlessEncryptionKey: Uint8Array; // Store initial encryptedSeedlessEncryptionKey for vault creation
 
     // Generate initial keys and vault state before tests run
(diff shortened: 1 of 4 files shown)


himanshuchawla009:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -647,24 +647,15 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-          rawToprfPwEncryptionKey: toprfPwEncryptionKey,
-          rawToprfAuthKeyPair: toprfAuthKeyPair,
-        });
-      }
+      await this.#createNewVaultWithAuthData({
+        password,
+        rawToprfEncryptionKey: toprfEncryptionKey,
+        rawToprfPwEncryptionKey: toprfPwEncryptionKey,
+        rawToprfAuthKeyPair: toprfAuthKeyPair,
+      });
```
matthiasgeihs: why move this down? controller should be unlocked only after this!

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -785,24 +776,17 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-          rawToprfPwEncryptionKey: toprfPwEncryptionKey,
-          rawToprfAuthKeyPair: toprfAuthKeyPair,
-        });
-      }
+      await this.#createNewVaultWithAuthData({
+        password: globalPassword,
+        rawToprfEncryptionKey: latestEncKey,
+        rawToprfPwEncryptionKey: latestPwEncKey,
+        rawToprfAuthKeyPair: latestAuthKeyPair,
+      });
```
matthiasgeihs: is this even necessary? if yes, should do it above `this.#setUnlocked();`?

matthiasgeihs: @lwin-kyaw any thoughts?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1841,21 +1825,64 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+        revokeToken: newRevokeToken,
+        accessToken,
+      });
+
+      // this function acquires the vault lock
+      await this.#updateVault({
+        password,
+        serializedVaultData,
+        pwEncKey: rawToprfPwEncryptionKey,
+      });
```
matthiasgeihs: are you duplicating the vault serialization logic here?
this is dangerous as we risk having duplicate implementations deviate from each other.
one key aspect of the keyring vault and our vault should be that it always goes through exactly the same code to update the vault.

Please use the existing function `#createNewVaultWithAuthData` if you want to write to the vault.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1841,21 +1825,64 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
    * Revoke the refresh token and get new refresh token and new revoke token.
    * This method is to be called after user is authenticated.
    *
-   * @param revokeToken - The revoke token to use for revoking the refresh token.
+   * @param password - The password to encrypt the vault.
+   * @returns A Promise that resolves to void.
    */
-  async #revokeRefreshTokenAndUpdateState(revokeToken: string) {
-    this.#assertIsAuthenticatedUser(this.state);
+  async revokeRefreshTokenAndUpdateVault(password: string) {
```
matthiasgeihs: thought: can simply call it `revokeRefreshToken` and mention in the function doc that it also updates the vault.

matthiasgeihs (changes requested):
I generally think this is a good change:
Make it more explicit to call the `revokeRefreshToken`, then the client can control when the supposedly high-latency revoke call is made.

however, there are some issues with the current implementation. we should not duplicate the vault update logic!

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: as discussed in call, we don't need to call `#createNewVaultWithAuthData` here

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: as discussed in call, we don't need to call `#createNewVaultWithAuthData` here

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -3731,13 +3742,17 @@ describe('SeedlessOnboardingController', () => {
     let initialAuthKeyPair: KeyPair; // Store initial keypair for vault creation
     let initialEncKey: Uint8Array; // Store initial encKey for vault creation
     let initialPwEncKey: Uint8Array; // Store initial pwEncKey for vault creation
+    let initialSeedlessEncKey: Uint8Array; // Store initial seedlessEncKey for vault creation
```
matthiasgeihs: not need to expose this here.

matthiasgeihs: if you absolutely want to expose it, you should use consistent naming `encryptionKey` for vault encryption keys. (`encKey` is used for keys coming from toprf.)
```suggestion
    let initialVaultEncryptionKey: Uint8Array; // Store initial seedlessEncKey for vault creation
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -3747,6 +3762,13 @@ describe('SeedlessOnboardingController', () => {
         OLD_PASSWORD,
       );
 
+      initialSeedlessEncKey =
+        mockToprfEncryptor.deriveSeedlessEncKey(OLD_PASSWORD);
+
+      const aes = managedNonce(gcm)(mockResult.pwEncKey);
+      initialEncryptedSeedlessEncryptionKey = aes.encrypt(
+        initialSeedlessEncKey,
+      );
```
matthiasgeihs: this needs to be an encryption of an exported key, not a raw key.
see `encryptWithDetail`.

matthiasgeihs (changes requested):

lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: removed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: removed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: it's still there!?

matthiasgeihs (commented):
the function call to `#createNewVaultWithAuthData` is still there in `#submitGlobalPassword`. Otherwise looks good.

packages/seedless-onboarding-controller/tests/mocks/toprfEncryptor.ts
matthiasgeihs, line 1: Sorry, one more comment.

It seems like `deriveSeedlessEncKey` is actually not needed.
It is used only in `mockcreateLocalKey`, but the return value is not used anywhere as far as i can see.
if it's not needed, we should remove it.

matthiasgeihs (changes requested):

matthiasgeihs (approved):

PR #6188, opened 2025-07-24
Add `error` property to "degraded" events

### LINKED ISSUE #6190: Add `error` property to `NetworkController:rpcEndpointDegraded`
When the `NetworkController:rpcEndpointDegraded` event is published, it does not include the error from the last response (if the event was the result of exceeding the max number of retries). We would include to include the RPC error code in the metrics event on the client side so we need access to this error in order to do so.

mcmire:
## Explanation

We would like to start tracking the HTTP status code in metrics when the `NetworkController:rpcEndpointDegraded` event is published because the maximum number of retries for a request has been exceeded. The HTTP status is stored in the error that Cockatiel captures, so we just need to include it in the event payload.

## References

Closes #6190.

## Checklist

  - https://github.com/MetaMask/metamask-extension/pull/34605
  - https://github.com/MetaMask/metamask-mobile/pull/18059


mcmire:
@metamaskbot publish-previews

packages/network-controller/src/rpc-service/shared.ts
```diff
@@ -9,8 +9,5 @@ export type FetchOptions = RequestInit;
  */
 export type AddToCockatielEventData<EventListener, AdditionalData> =
   EventListener extends (data: infer Data) => void
-    ? // Prevent Data from being split if it's a type union
-      [Data] extends [void]
-      ? (data: AdditionalData) => void
-      : (data: Data & AdditionalData) => void
+    ? (data: Data extends void ? AdditionalData : Data & AdditionalData) => void
```
mcmire, line 12: This was an interesting one to figure out.

The purpose of this utility type is to extend the type for a Cockatiel event with additional payload data. We need to do this because `NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointAvailable` are based on existing Cockatiel events but with additional properties we add.

However, this type did not properly account for when `EventListener` was `CockatielEvent<FailureReason<unknown> | void>` (which is the same thing as saying `(data: void | { error: Error } | { value: unknown }) => void`).

If we say the following

``` typescript
AddToCockatielEventData<
  (data: void | { error: Error } | { value: unknown }) => void,
  { foo: 'bar' }
>
```

we want this to resolve to:

``` typescript
(data: { foo: 'bar' } | ({ error: Error } & { foo: 'bar' }) | ({ value: unknown } & { foo: 'bar' })) => void
```

(Essentially, we want the `void` to be treated as `{}`.)

But instead this was resolving as:

``` typescript
(data: ((void | { error: Error } | { value: unknown }) & { foo: 'bar' }) => void
```

which distributes to:

``` typescript
(data: (void & { foo: 'bar' }) | ({ error: Error } & { foo: 'bar' }) | ({ value: unknown } & { foo: 'bar' })) => void
```

This is incorrect, because `void & { foo: 'bar' }` doesn't make sense.

What we want is for TypeScript to distribute `void | { error: Error } | { value: unknown }` over the condition. So first, despite the comment here, we _don't_ want `[Data] extends [void]`, because that prevents TypeScript from applying the distribution.

But the second thing is that even if we say `Data extends void` then TypeScript still won't apply the distribution, because the condition is actually in the wrong place. What we want is not a type union of three function types, but a single type with a type union for the arguments.

Even if the condition _were_ in the place we wanted it, TypeScript seems to distribute the type union only in a particular circumstance. Looking a bit deeper at the TypeScript 2.8 release notes where this concept was introduced, it [says](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types): "Distributive conditional types are automatically distributed over union types during instantiation." In this case, we also need to place the condition so it becomes a part of the type that we want to "return" from this utility type and is not used to simply determine the return type. This way TypeScript evaluates the condition when that return type is used and not beforehand.

packages/controller-utils/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
```
cryptodev-2s, line 10: Should this be Changed ? because it feels a bit strange to have under Added section Update as a first keyword ?

packages/network-controller/src/rpc-service/shared.ts
cryptodev-2s: The change make sense.

packages/controller-utils/CHANGELOG.md
mcmire: The [Keep a Changelog](https://keepachangelog.com/en/1.1.0/#how) spec says that "Added" is for new features and "Changed" is for changes in existing functionality. I've always found this to be a bit lacking, though.

The way I've been treating "Changed" is that it is for changes in existing behavior or functionality that do not come with changes to the API itself. For instance, maybe a method was updated so that the logic is different but returns data in the same shape. "Added", on the other hand, is for changes that extend the API in some way, so this could be something larger like a new class, function, method, etc. but also something smaller like a new argument or a new property on a type.

I guess I could have said something like:

``` markdown
- `createServicePolicy`: Add `error` property to the payload for `onDegraded` ([#6188](https://github.com/MetaMask/core/pull/6188))
  - This can be used to access the error produced by the last request when the maximum number of retries is exceeded
  - This property will not be present if the degraded event merely represents a slow request
```

Would that sound like it fits better? Or what is your view on Added vs. Changed?

cryptodev-2s: Yes looks good

PR #6198, opened 2025-07-28
Release/480.0.0

salimtb:
Optimized RPC performance for account and asset tracking by batching requests with a multicall contract, with fallbacks for unsupported chains. Also fixed a caching issue in TokenBalancesController to ensure up-to-date balances.


Prithpal-Sooriya (approved):

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [73.1.0]
+
 ### Changed
 
+- Upgrade `AccountsController` version ([#6194](https://github.com/MetaMask/core/pull/6194))
```
cryptodev-2s: Nit: Could we remove this entry, I see it has only been a dev dependency bump 
```Suggestion

File: packages/assets-controllers/CHANGELOG.md [hunk 2]
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [73.1.0]
```
cryptodev-2s: should this be a patch ? I see we have nothing new compared to previous version only some fixes ?

cryptodev-2s (approved):
LGTM!

PR #6205, opened 2025-07-29
chore: exports seedless-onboarding allowed actions/events

lwin-kyaw:
## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/src/index.ts
+++ b/packages/seedless-onboarding-controller/src/index.ts
@@ -12,6 +12,8 @@ export type {
   SeedlessOnboardingControllerStateChangeEvent,
   SeedlessOnboardingControllerActions,
   SeedlessOnboardingControllerEvents,
+  SeedlessOnboardingControllerAllowedEvents,
+  SeedlessOnboardingControllerAllowedActions,
   ToprfKeyDeriver,
   RecoveryErrorData,
 } from './types';
--- a/packages/seedless-onboarding-controller/src/types.ts
+++ b/packages/seedless-onboarding-controller/src/types.ts
@@ -183,6 +183,10 @@ export type SeedlessOnboardingControllerActions =
 
 export type AllowedActions = never;
 
+export type SeedlessOnboardingControllerAllowedActions =
+  | SeedlessOnboardingControllerActions
+  | AllowedActions;
+
 // Events
 export type SeedlessOnboardingControllerStateChangeEvent =
   ControllerStateChangeEvent<
@@ -196,11 +200,15 @@ export type AllowedEvents =
   | KeyringControllerLockEvent
   | KeyringControllerUnlockEvent;
 
+export type SeedlessOnboardingControllerAllowedEvents =
+  | SeedlessOnboardingControllerEvents
+  | AllowedEvents;
+
 // Messenger
 export type SeedlessOnboardingControllerMessenger = RestrictedMessenger<
   typeof controllerName,
-  SeedlessOnboardingControllerActions | AllowedActions,
-  SeedlessOnboardingControllerEvents | AllowedEvents,
+  SeedlessOnboardingControllerAllowedActions,
+  SeedlessOnboardingControllerAllowedEvents,
   AllowedActions['type'],
   AllowedEvents['type']
 >;


lwin-kyaw:
@metamaskbot publish-preview

lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -513,7 +530,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
    */
   async changePassword(newPassword: string, oldPassword: string) {
     return await this.#withControllerLock(async () => {
-      this.#assertIsUnlocked();
+      // assert that the controller is unlocked before changing the password
+      this.#assertIsUnlocked(this.state);
```
matthiasgeihs: i'm surprised that this now gets a state. why is that?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -956,8 +981,9 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
    * @param keyringEncryptionKey - The keyring encryption key.
    */
   async storeKeyringEncryptionKey(keyringEncryptionKey: string) {
-    const { toprfPwEncryptionKey: encKey } =
-      await this.#unlockVaultAndGetVaultData();
+    this.#assertIsUnlocked(this.state);
+    const { toprfPwEncryptionKey } = this.state;
+    const encKey = base64ToBytes(toprfPwEncryptionKey);
```
matthiasgeihs: what do you think about having a dedicated function for loading the key from the state and deserializing it?
```suggestion
    const encKey = this.getToprfPwEncryptionKey();
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1243,64 +1275,52 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+      if (params?.encryptionKey) {
+        vaultEncryptionKey = params.encryptionKey;
       }
 
       let decryptedVaultData: unknown;
       const updatedState: Partial<SeedlessOnboardingControllerState> = {};
-
-      if (password) {
-        assertIsValidPassword(password);
+      if (params?.password) {
```
matthiasgeihs: could we first check for encryption key and then for password?
if we have the encryption key, we don't need to derive it again.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -105,10 +111,6 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
       persist: false,
       anonymous: true,
     },
-    vaultEncryptionSalt: {
-      persist: false,
-      anonymous: true,
-    },
```
matthiasgeihs: does removing an element from the state cause any issue when upgrading?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1421,17 +1447,16 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-   * @param params.rawToprfEncryptionKey - The encryption key to encrypt the vault.
-   * @param params.rawToprfPwEncryptionKey - The encryption key to encrypt the password.
-   * @param params.rawToprfAuthKeyPair - The authentication key pair to encrypt the vault.
+   * @param params.rawToprfEncryptionKey - The TOPRF encryption key.
+   * @param params.rawToprfPwEncryptionKey - The TOPRF password encryption key.
+   * @param params.rawToprfAuthKeyPair - The TOPRF authentication key pair.
+   * @returns A promise that resolves to the updated vault.
    */
-  async #createNewVaultWithAuthData({
+  async #updateVault({
```
matthiasgeihs: it would be great if this function would use `encryptWithKey` if the encryption key is available.
then we don't need to derive the key from the password everytime we update the vault. (key derivation is intentionally slow to make brute force attacks harder.)

lwin-kyaw: yes, that's the reason I still keep the `vaultEncryptionKey` in the state. I will update the encryptor in the next PR to avoid making bulk PR.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: No, previously `vaultEncryptionSalt` is only useful when we decrypt the vault with `encryptionKey` to get the decrypted vault data during the vault is unlocked.
But with the state changes (after vault unlock), we don't need to consider about that anymore as the required vault data is always available in the state as long as the vault is unlocked.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: To make sure that the seedless controller/vault is unlocked, we will now also need to assert that `toprfEncryptionKey`, `toprfPwEncryptionKey` and `toprfAuthKeyPair` are available in the state.

That's why we need to pass the state.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Adding `encryptWithKey` to `Encryptor` interface will be **breaking** change to the client, so wanna keep it in separate PR.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Updated in this commit, https://github.com/MetaMask/core/pull/6205/commits/d084fe8a51bd9669f3308010608c980871f727a4

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Updated here, https://github.com/MetaMask/core/pull/6205/commits/d084fe8a51bd9669f3308010608c980871f727a4

packages/seedless-onboarding-controller/src/utils.ts
```diff
@@ -33,3 +39,30 @@ export function decodeJWTToken(token: string): DecodedBaseJWTToken {
...
+ * @param authKeyPair - The authentication key pair to deserialize.
+ * @returns The deserialized authentication key pair.
+ */
+export function deserializAuthKeyPair(authKeyPair: string): KeyPair {
+  const { publicKey, privateKey } = JSON.parse(authKeyPair);
+  return {
+    pk: base64ToBytes(publicKey),
+    sk: hexToBigInt(privateKey),
+  };
+}
```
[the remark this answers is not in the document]
lwin-kyaw: Fixed spelling in here, https://github.com/MetaMask/core/pull/6205/commits/86b227a42c8f505c1a41719993934da18d4d1d82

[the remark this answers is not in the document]
lwin-kyaw: Update the serialize function to have consistency with original data here, https://github.com/MetaMask/core/pull/6205/commits/5dfaf898d0fd0b84bc457b3feb486c7c9cffbebb

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1636,12 +1680,39 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+
+    if (
+      !value ||
+      typeof value !== 'object' ||
+      !('toprfEncryptionKey' in value) ||
+      !('toprfPwEncryptionKey' in value) ||
+      !('toprfAuthKeyPair' in value)
+    ) {
+      throw new Error(SeedlessOnboardingControllerErrorMessage.VaultLocked);
+    }
```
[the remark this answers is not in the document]
lwin-kyaw: Updated the assertion in this commit, https://github.com/MetaMask/core/pull/6205/commits/9c1f7ef1c84b7e52fb26d856bd5a1946107aac27

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: but you can directly access this.state inside `assertIsUnlocked` , why passing it explicitly?

packages/seedless-onboarding-controller/src/index.ts
```diff
@@ -12,6 +12,8 @@ export type {
   SeedlessOnboardingControllerStateChangeEvent,
   SeedlessOnboardingControllerActions,
   SeedlessOnboardingControllerEvents,
+  SeedlessOnboardingControllerAllowedEvents,
```
mcmire: Can you explain the purpose of exporting these types? We don't typically need to do this in controller packages. `SeedlessOnboardingControllerActions` and `SeedlessOnboardingControllerEvents` are already available, and the other action and event types that `SeedlessOnboardingMessenger` uses are already available from other packages.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1391,6 +1391,9 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
         state.vaultEncryptionKey = vaultEncryptionKey;
         state.revokeToken = revokeToken;
         state.accessToken = accessToken;
+        state.toprfEncryptionKey = bytesToBase64(toprfEncryptionKey);
+        state.toprfPwEncryptionKey = bytesToBase64(toprfPwEncryptionKey);
+        state.toprfAuthKeyPair = serializeAuthKeyPair(toprfAuthKeyPair);
```
matthiasgeihs: As we saw with the recent state log vulnerability, putting sensitive information in the state can lead to major issues.
After that recent finding, I no longer think we should put these keys into the state.

I had a look how they KeyringController does it: When it loads the keyrings, it only puts the keyring metadata into the state. The SRPs and private keys are never added to the state. Instead they are only held in a private variable `#keyrings`. https://github.com/MetaMask/core/blob/17999dc66c1070fb6af8a9603b9caa3ce182a637/packages/keyring-controller/src/KeyringController.ts#L660

I think we can do the same thing here: We can add private variables for these keys (or one private variable that holds all of them). If the variable is not set, we try to load it from the vault.

lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/index.ts
lwin-kyaw: Hi, thanks for the review.

The reason for exporting `SeedlessOnboardingControllerAllowedEvents` is to directly use the type when initializing the messenger, in the clients.
e.g.
```ts
import {
  SeedlessOnboardingControllerGetStateAction,
  SeedlessOnboardingControllerAllowedEvents,
  SeedlessOnboardingControllerAllowedActions,
  SeedlessOnboardingControllerEvents,
} from '@metamask/seedless-onboarding-controller';

type MessengerActions =
  | SeedlessOnboardingControllerGetStateAction
  | SeedlessOnboardingControllerAllowedActions;

type MessengerEvents =
  | SeedlessOnboardingControllerEvents
  | SeedlessOnboardingControllerAllowedEvents;
```

Otherwise, in the clients, we will have to re-create the type and listed all the allowed events/actions as same as the controller does.
This could lead to accidentally defining incorrect types, missing types or always needing to update the type in the clients whenever the controller allowed events/actions changes.
One example of this in extension: https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/controller-init/messengers/transaction-controller-messenger.ts#L45

wdyt?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1588,87 +1600,28 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-    };
-
-    return {
-      toprfEncryptionKey: rawToprfEncryptionKey,
-      toprfPwEncryptionKey: rawToprfPwEncryptionKey,
-      toprfAuthKeyPair: rawToprfAuthKeyPair,
-      revokeToken: parsedVaultData.revokeToken,
-      accessToken: parsedVaultData.accessToken,
-    };
+    return parsedVaultData;
```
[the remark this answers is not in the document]
lwin-kyaw: `VaultDataError` - for unexpected vault data (such as non-string values, or non-json strings)

`InvalidVaultData` - for validating required values in the decrypted vault data.

packages/seedless-onboarding-controller/jest.config.js
```diff
@@ -17,10 +17,10 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 100,
+      branches: 98.82,
```
lwin-kyaw: I had to update the coverage threshold coz I can't make 100% test coverage (without breaking the hard private class members).

And I don't think we should break the controller logic (exposing `private` fields for mock) to compensate test coverage.

lwin-kyaw:
@metamaskbot publish-preview

lwin-kyaw:
> I notice that you've applied the `no-changelog` label. Is this merely a refactor PR or are there changes to the controller state type that are worth noting?

Hi @mcmire, Yes, this is a refactor PR and some optimization (caching) for internal vault logics

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1260,81 +1280,62 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
       });
 
-      return {
-        toprfEncryptionKey,
-        toprfPwEncryptionKey,
-        toprfAuthKeyPair,
-        revokeToken,
-        accessToken,
-      };
+      const deserializedVaultData = deserializeVaultData(vaultData);
```
matthiasgeihs, line 1302: Hm, I don't understand why you separate parsing and deserialization? Why don't you merge deserialization into `#decryptAndParseVaultData`?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -2344,6 +2188,65 @@ describe('SeedlessOnboardingController', () => {
...
+          }),
+        },
+        async ({ controller, encryptor }) => {
+          jest
+            .spyOn(encryptor, 'decryptWithKey')
+            .mockResolvedValueOnce(mockVault);
+          await expect(
+            controller.submitPassword(MOCK_PASSWORD),
+          ).rejects.toThrow(
+            SeedlessOnboardingControllerErrorMessage.InvalidVaultData,
```
matthiasgeihs, line 2211: Are you throwing a different error here than before? If so, this can be considered a breaking change (depending on your definition of breaking change).

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -182,6 +189,11 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
 
   readonly #revokeRefreshToken: RevokeRefreshToken;
 
+  /**
+   * The TTL of the password outdated cache in milliseconds.
+   */
+  readonly #passwordOutdatedCacheTTL: number;
+
```
matthiasgeihs, line 196: Why did you move that up?

TransactionController is maintaining an alphabetical ordering on properties and I kind of like it.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1351,12 +1338,25 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+      // Note that vault decryption using the password is a very costly operation as it involves deriving the encryption key
+      // from the password using an intentionally slow key derivation function.
+      // We should make sure that we only call it very intentionally.
+      const result = await this.#vaultEncryptor.decryptWithDetail(
+        params.password,
+        encryptedVault,
+      );
+      decryptedVaultData = result.vault;
+      vaultEncryptionKey = result.exportedKeyString;
+      vaultEncryptionSalt = result.salt;
```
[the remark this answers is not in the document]
matthiasgeihs: It's not supposed to validate the password, if the key is given, right?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1588,87 +1600,28 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
     }
 
     let parsedVaultData: unknown;
     try {
       parsedVaultData = JSON.parse(data);
     } catch {
-      throw new Error(
-        SeedlessOnboardingControllerErrorMessage.InvalidVaultData,
-      );
+      throw new Error(SeedlessOnboardingControllerErrorMessage.VaultDataError);
```
matthiasgeihs, line 1619: Changing the error type can be considered a breaking change.
Why did you change it?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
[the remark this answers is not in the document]
matthiasgeihs: Hm, not sure if this differentiation is necessary or clear.

packages/seedless-onboarding-controller/src/index.ts
matthiasgeihs: ping @mcmire 

@lwin-kyaw Btw, there is a contributor docs entry for this.
https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#define-but-do-not-export-a-type-union-for-external-action-types
(I don't have a strong opinion on this, just noting.)

packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -328,10 +330,6 @@ export type MutuallyExclusiveCallback<Result> = ({
  * The structure of the data which is serialized and stored in the vault.
  */
 export type VaultData = {
-  /**
-   * The node auth tokens from OAuth User authentication after the Social login.
-   */
-  authTokens: NodeAuthTokens;
```
matthiasgeihs, line 334: Why are these removed?

packages/seedless-onboarding-controller/src/index.ts
matthiasgeihs: Btw2, the upcoming messenger migration will make some of this obsolete.
See https://github.com/MetaMask/core/pull/6364.
 https://github.com/MetaMask/core/blob/15d3a73ea75fec24f472debc5cae0a0cb9ef8685/packages/seedless-onboarding-controller/tests/__fixtures__/mockMessenger.ts#L18-L21

matthiasgeihs: Btw3, you can also extract these types manually.
https://github.com/MetaMask/core/blob/42df019122b90ee813110a09b6e11c8c9d589e4b/packages/shield-controller/tests/mocks/messenger.ts#L21-L22

packages/seedless-onboarding-controller/jest.config.js
```diff
@@ -17,10 +17,10 @@ module.exports = merge(baseConfig, {
...
       functions: 100,
-      lines: 100,
-      statements: 100,
+      lines: 99.78,
+      statements: 99.78,
```
matthiasgeihs: Striving for less than perfection 😭

This appears to be the part not covered.

https://github.com/MetaMask/core/blob/a3d8559e74dc370e5ecdff33acf55036bd361df4/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts#L1341-L1348

Couldn't you force the state to hold an inconsistent salt?

matthiasgeihs (approved):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: I just wanna group the `readonly` variables in one block. That's why I moved that up.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Previously, it throws `Invalid vault data` error.
But I think the error should be `The decrypted vault has an unexpected shape.` since this is the parse error.

packages/seedless-onboarding-controller/src/types.ts
lwin-kyaw: `authTokens` are persisted outside the vault and so, it's not required to be inside the vault and it's redundant.

We also never add this value to vault before.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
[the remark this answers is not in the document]
lwin-kyaw: yes.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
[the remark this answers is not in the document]
lwin-kyaw: We gonna throw the `VaultDataError` (error message: `The decrypted vault has an unexpected shape.`) when vault is missing, or it's failed to parse (due to invalid json strings).

We will throw the `InvalidVaultData` (message: `Invalid vault data`) error when the decrypted vault is missing required data (e.g. tokens, encKeys).

packages/seedless-onboarding-controller/src/index.ts
lwin-kyaw: Thanks @matthiasgeihs for highlighting that. I did missed that documentation.
I have reverted the exports and updated the messenger types here,  6feeb1e4d193864a58a4a539e2cbbf40c6f9210d

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -115,7 +118,11 @@ type WithControllerCallback<ReturnValue, EKey> = ({
   encryptor: VaultEncryptor<EKey>;
   initialState: SeedlessOnboardingControllerState;
   messenger: SeedlessOnboardingControllerMessenger;
-  baseMessenger: Messenger<AllowedActions, AllowedEvents>;
+  baseMessenger: Messenger<
+    ExtractAvailableAction<SeedlessOnboardingControllerMessenger>,
+    | SeedlessOnboardingControllerEvents
+    | ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>
```
mcmire: Nit: `SeedlessOnboardingControllerEvents` is already included in `ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>`, so you should be able to simplify this:

``` suggestion
    ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>

File: packages/seedless-onboarding-controller/tests/__fixtures__/mockMessenger.ts [hunk 20]
```diff
@@ -1,18 +1,26 @@
...
  * creates a custom seedless onboarding messenger, in case tests need different permissions
  *
  * @returns base messenger, and messenger. You can pass this into the mocks below to mock messenger calls
  */
 export function createCustomSeedlessOnboardingMessenger() {
-  const baseMessenger = new Messenger<AllowedActions, AllowedEvents>();
+  const baseMessenger = new Messenger<
+    SeedlessOnboardingControllerActions | never,
+    | SeedlessOnboardingControllerEvents
+    | ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>
```
mcmire: Nit: `SeedlessOnboardingControllerEvents` is already included in `ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>`, so you should be able to simplify this:

``` suggestion
    ExtractAvailableEvent<SeedlessOnboardingControllerMessenger>

packages/seedless-onboarding-controller/src/index.ts
mcmire: Thank you!

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
lwin-kyaw: Thanks for the review. I've addressed it in this commit, https://github.com/MetaMask/core/pull/6205/commits/d7be4d5e78bae8024740f471450755b113bfc374

packages/seedless-onboarding-controller/tests/__fixtures__/mockMessenger.ts
lwin-kyaw: Addressed here, https://github.com/MetaMask/core/pull/6205/commits/d7be4d5e78bae8024740f471450755b113bfc374

packages/seedless-onboarding-controller/jest.config.js
lwin-kyaw: Ah sorry. My bad. I found a valid way to make the different salt. Added a test case here, https://github.com/MetaMask/core/pull/6205/files#diff-81c3df63ce0f2eb695178ae8b64b431ab1a75a03ee6d128d3867d2b75bc69fcdR3931

lwin-kyaw: I've reverted the threshold back to 100% 🙈

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: I see. I think the previous error message was fine, but the new one is more differentiated. I think the error variable names are not expressive enough to reflect that meaning, so this can lead to them not being used correctly.

matthiasgeihs (approved):

lwin-kyaw:
> Another worthwhile optimization would be to use the cached encryption key for encryption as well. (Currently it's only used for decryption.) And we may want to rotate the key after each controller unlock, like the KeyringController does it.

Yes definitely. I will address this in the new PR, coz that includes **breaking change** to the encryptor interface.

PR #6207, opened 2025-07-29
fix: dont fetch asset rates when no assets

xavier-brochard:
## Explanation

This PR adds an early return in `MultichainAssetsRatesController` when method is being called with an empty list of assets.

In some situations, for instance when the user sends all their balance of a given asset, the `MultichainAssetsRatesController` can attempt to fetch conversions with an empty list of assets. This results in a validation error crashing the snap (plus an unnecessary network call), since the [snap validation structure](https://github.com/MetaMask/snaps/blob/main/packages/snaps-execution-environments/src/common/validation.ts#L312-L323) expects at least 1 conversion argument.

## Checklist


Prithpal-Sooriya (dismissed):

xavier-brochard:
Closing. This case should never happen, because the snap should always return at least the native asset.

Mrtenz (approved):

PR #6212, opened 2025-07-30
feat: create new utility function to fetch balances for multiples acc…

salimtb:
feat: create new utility function to fetch balances for multiple account addresses

### Current State
The existing multicall functionality only provided basic contract call batching using the `tryAggregate` function. When applications needed to fetch token balances for multiple addresses and tokens, they had to make individual calls or create their own batching logic, leading to inefficient RPC usage and slower user experiences.

### Solution
This PR introduces a new specialized utility function `getTokenBalancesForMultipleAddresses` that leverages Multicall3's `aggregate3` function to efficiently batch balance queries for multiple token addresses and user addresses in a single operation.

**Key additions:**

1. **`aggregate3` function**: A new lower-level function that provides direct access to Multicall3's `aggregate3` method, allowing for more granular control over batch calls with individual failure handling.

2. **`getTokenBalancesForMultipleAddresses` function**: The main utility that:
   - Fetches ERC20 token balances for multiple tokens and addresses in a single batch
   - Optionally includes native token balances (mapped to zero address)
   - Automatically handles call encoding/decoding for both ERC20 `balanceOf` and Multicall3 `getEthBalance`
   - Processes up to 300 calls per batch to avoid gas/size limits

3. **Comprehensive fallback support**: 
   - Falls back to individual balance calls when Multicall3 is not supported on a chain
   - Uses `eth_getBalance` for native token balances when multicall fails
   - Maintains the same interface regardless of the underlying implementation

4. **Helper functions**:
   - `createERC20BalanceCalls`: Generates encoded balance calls for ERC20 tokens
   - `createNativeTokenBalanceCalls`: Generates encoded calls for native token balances
   - `processBalanceResults`: Decodes and organizes the batch results
   - `getNativeBalancesFallback`: Handles native balance fetching for unsupported chains

### Technical Details
- Supports all chains where Multicall3 is deployed (extensive chain support already configured)
- Returns a nested object structure: `Record<tokenAddress, Record<userAddress, BN>>`
- Native token balances are accessible via the zero address key
- Includes proper error handling and logging for debugging
- Maintains backwards compatibility with existing multicall functionality

## References

This enhancement will significantly improve performance for applications that need to:
- Display portfolio balances across multiple tokens
- Batch balance checks for multiple user addresses
- Reduce RPC calls when fetching balance data

## Checklist


packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +466,402 @@ export const multicallOrFallback = async (
...
+  // Create a temporary contract instance to encode the balanceOf function calls
+  const tempContract = new Contract(
+    ZERO_ADDRESS,
+    ERC20_BALANCE_OF_ABI,
+    provider,
+  );
+
+  tokenAddresses.forEach((tokenAddress) => {
+    userAddresses.forEach((userAddress) => {
+      calls.push({
```
Prithpal-Sooriya: Nice!!

Do we know if there is a limit to the number of call data we can use?
1. Maybe is there an RPC max size request?
2. Does the multicall contract have a max request size?

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +466,402 @@ export const multicallOrFallback = async (
...
+/**
+ * Processes and decodes balance results from aggregate3 calls
+ *
+ * @param results - Array of results from aggregate3 calls
+ * @param callMapping - Array mapping call indices to token and user addresses
+ * @param chainId - The hexadecimal chain id
+ * @param provider - An ethers rpc provider
+ * @returns Map of token address to map of user address to balance
+ */
+const processBalanceResults = (
```
Prithpal-Sooriya, line 540: Really nice mapping here

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +466,402 @@ export const multicallOrFallback = async (
...
+
+    // Add native token balance calls if requested
+    if (includeNative) {
+      const nativeCallData = createNativeTokenBalanceCalls(
+        userAddresses,
+        chainId,
+        provider,
+      );
+      allCalls = [...allCalls, ...nativeCallData.calls];
+      allCallMapping = [...allCallMapping, ...nativeCallData.callMapping];
```
Prithpal-Sooriya: nit - thoughts on `.concat` or `.push` instead of spreading?
```

// concat approach
allCalls = allCalls.concat(nativeCallData.calls)

// push approach
allCalls.push(...nativeCallData.calls)
```

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +466,402 @@ export const multicallOrFallback = async (
...
+    }
+
+    // Execute all calls in batches
+    const maxCallsPerBatch = 300; // Limit calls per batch to avoid gas/size limits
+    const allResults: Aggregate3Result[] = [];
```
Prithpal-Sooriya, line 974: Aha very nice work with this limit checks.

packages/assets-controllers/src/multicall.ts
Prithpal-Sooriya: NVM I see below we make batches, nice!

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +466,402 @@ export const multicallOrFallback = async (
...
+
+    await reduceInBatchesSerially<Aggregate3Call, void>({
+      values: allCalls,
+      batchSize: maxCallsPerBatch,
+      initialResult: undefined,
+      eachBatch: async (_, batch) => {
+        const batchResults = await aggregate3(batch, chainId, provider);
+        allResults.push(...batchResults);
```
Prithpal-Sooriya, line 982: optional nit - the `reduceInBatchesSerially` could return the array of results?
But TBH I'm fine with just pushing the results into this array.

packages/assets-controllers/src/multicall.ts
```diff
@@ -131,7 +133,7 @@ const MULTICALL_CONTRACT_BY_CHAINID = {
   '0x96f': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0x3cc5': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0x4571': '0xcA11bde05977b3631167028862bE2a173976CA11',
-  '0xe99': '0xcA11bde05977b3631167028862bE2a173976CA11',
+  '0xe99': '0xca11bde05977b3631167028862be2a173976ca11',
```
Prithpal-Sooriya: Was this meant to be lowercased instead of checksummed?

packages/assets-controllers/src/multicall.ts
```diff
@@ -193,30 +195,11 @@ const MULTICALL_CONTRACT_BY_CHAINID = {
   '0xa1337': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0x1f2b': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0xf63': '0xcA11bde05977b3631167028862bE2a173976CA11',
-  '0x144': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
```
Prithpal-Sooriya: Ooooh interesting that this network changed address 👀

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +479,539 @@ export const multicallOrFallback = async (
...
+ * @param includeNative - Whether to include native token balances (default: true)
+ * @param maxCallsParallel - Maximum number of parallel calls (default: 20)
+ * @returns Promise resolving to map of token address to map of user address to balance
+ */
+const getTokenBalancesFallback = async (
+  tokenAddresses: string[],
+  userAddresses: string[],
+  provider: Web3Provider,
+  includeNative: boolean,
+  maxCallsParallel: number,
```
Prithpal-Sooriya, line 696: Great work including this max calls in parallel.

Prithpal-Sooriya (dismissed):

amitabh94 (approved):

PR #6214, opened 2025-07-30
refactor(account-tree-controller): add type field for each tree node + improve typing overall

ccharly:
## Explanation

Add new `type` field for each wallets/groups.

This is used as a "tag", so proper node types can be inferred by the compiler.

e.g:
```ts
// Using wallet as root node:
if (wallet.type === AccountWalletType.Entropy) {
  console.log(wallet.metadata.entropy.id);

  // This `wallet` is narrowed down to `AccountWalletEntropyObject`, it knows that
  // `wallet.groups` are `AccountGroupMultichainAccountObject`, so we can use
  // the `groupIndex` safely here too.
    console.log(group.metadata.entropy.groupIndex);
  }
}

if (wallet.type === AccountWalletType.Snap) {
  console.log(wallet.metadata.snap.id);
}

if (wallet.type === AccountWalletType.Keyring) {
  console.log(wallet.metadata.keyring.type);
}

// Using group as root node:
if (group.type === AccountGroupType.SingleAccount) {
  // No real metadata for this, however, we enforced the `accounts` to be
  // a `[AccountId]` tuple, so it is safe to do:
  console.log(group.accounts[0]);
  
  // And this should not build:
  console.log(group.accounts[1]); // ERROR: Tuple type '[string]' of length '1' has no element at index '1'.
}

if (group.type === AccountGroupType.MultichainAccount) {
  console.log(group.metadata.groupIndex);
  
  // For multichain accounts group we have "at least 1" account, but we could have more.
  console.log(group.accounts[0]);
  console.log(group.accounts[1]); // Assuming we have 2 accounts here.
}
```

## References

N/A

## Changelog

N/A

## Checklist


packages/account-tree-controller/src/rule.ts
```diff
@@ -0,0 +1,105 @@
...
+    id: AccountWalletIdOf<WalletType>;
+    // Omit `name` since it will get computed after the tree is built.
+    metadata: Omit<AccountWalletObjectOf<WalletType>['metadata'], 'name'>;
+  };
+  group: {
+    type: GroupType;
+    // TODO: Maybe re-use `WalletType` in the ID definition here.
+    id: AccountGroupIdOf<WalletType>;
+    // Omit `name` since it will get computed after the tree is built.
+    metadata: Omit<AccountGroupObjectOf<GroupType>['metadata'], 'name'>;
```
ccharly, line 29: I'm not sure we should omit the `name` actually..

For `SingleAccount` we could just name them as we "group" accounts, given that 1 group === 1 account (this would also avoid having a "fallback name" for those)

So I might make that change...

ccharly: And for `MultichainAccount`, we would just leave it empty, so it will be resolved once the tree has been built (only if we don't have a name for this group already)

packages/account-tree-controller/src/rule.ts
```diff
@@ -0,0 +1,105 @@
...
+export type RuleResult<
+  WalletType extends AccountWalletType,
+  GroupType extends AccountGroupType,
+> = {
+  wallet: {
+    type: WalletType;
+    // TODO: Maybe re-use `WalletType` in the ID definition here.
+    id: AccountWalletIdOf<WalletType>;
+    // Omit `name` since it will get computed after the tree is built.
+    metadata: Omit<AccountWalletObjectOf<WalletType>['metadata'], 'name'>;
```
ccharly, line 23: We could also, not omit it here (the `Omit` feels kinda out-of-place TBH)

And we could use `''` as a default value...

My initial idea was only use the "specific part" of `metadata` here, in a generic way, but we cannot index this "specific part" using the `WalletType` type unfortunately.

We could pass it as a third parameter though...

packages/account-tree-controller/package.json
```diff
@@ -53,7 +53,7 @@
     "lodash": "^4.17.21"
   },
   "devDependencies": {
-    "@metamask/account-api": "^0.3.0",
+    "@metamask/account-api": "^0.5.0",
```
fabiobozzo, line 56:
```^0.5.0
type AccountWalletObject = {
  type: AccountWalletType; // new REQUIRED field?
  ...
}
```
Isn't that a breaking change?

fabiobozzo (approved):
LGTM. Question about your discriminated union example in _Explanation_:

You access `group.metadata.entropy.groupIndex` inside an 
`AccountWalletType.Entropy` check. Does this mean groups inherit the wallet's 
type for their metadata structure? This line suggests groups must match their parent wallet's type: `console.log(group.metadata.entropy.groupIndex);`.

Is this intentional, or should groups have their own independent type discrimination?

packages/account-tree-controller/package.json
ccharly: Good point! It is yes! I also forgot to update the changelog, so I'll make sure to add that there. (NOTE: It is a breaking change because we bump the peer dep itself)

packages/account-tree-controller/src/group.ts
```diff
@@ -0,0 +1,149 @@
...
+  getOnlyAccount(): InternalAccount {
+    const accountIds = this.getAccountIds();
+
+    if (accountIds.length > 1) {
+      throw new Error('Group contains more than 1 account');
+    }
+
+    // A group always have at least one account.
+    return this.#getAccount(accountIds[0]);
+  }
```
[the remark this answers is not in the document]
ccharly: That's a very good catch, and that's true!

Right now, we're not removing empty groups and empty wallets.

We should definitely do that to enforce our runtime state and be aligned with our type signatures!

Though, I think we should fix this a separate PR, since this one is pretty big already.

ccharly:
> LGTM. Question about your discriminated union example in _Explanation_:
> 
> You access `group.metadata.entropy.groupIndex` inside an `AccountWalletType.Entropy` check. Does this mean groups inherit the wallet's type for their metadata structure? This line suggests groups must match their parent wallet's type: `console.log(group.metadata.entropy.groupIndex);`.
> 
> Is this intentional, or should groups have their own independent type discrimination?

I did edit the example to give a better explanations on how the "wallet/group typing" works now. Thanks for pointing me that 😁

fabiobozzo (approved):

PR #6215, opened 2025-07-30
feat: add multichain account syncing checks & logic branches

mathieuartu:
## Explanation

This PR paves the way for Multichain Account Syncing by adding a new callback to `UserStorageController` that will be wired to the `enableMultichainAccounts` remote feature flag on clients.
When `enableMultichainAccounts` is `true`, then the current (and soon to be legacy) Account Syncing method will stop sending updates to the user storage, and only act as a legacy restoration vessel.

The reason behind all those changes is that `AccountTreeController` will be responsible for Multichain Account Syncing.

This PR also removes an unused public method.

Changelog below:

```md
### Added

- `UserStorageController` optional config callback `getIsMultichainAccountSyncingEnabled`, and `getIsMultichainAccountSyncingEnabled` public method / messenger action
  - This callback needs to be wired to client specific selectors in order to fetch the value of the feature flag dynamically
  - If `true`, Account syncing will stop pushing new data to the user storage and only act as an account restoration method that will be fired before multichain account syncing for legacy compatibility
  - This is done because `AccountTreeController` will become responsible for Multichain Account syncing

### Removed

- Unused `UserStorageController:saveInternalAccountToUserStorage` public method
```

## Checklist


fabiobozzo:
> The reason behind all those changes is that AccountTreeController will be responsible for Multichain Account Syncing.
Not `MultichainAccountService` ?

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
```diff
@@ -606,6 +616,13 @@ export default class UserStorageController extends BaseController<
     });
   }
 
+  public getIsMultichainAccountSyncingEnabled(): boolean {
+    return (
+      this.#config.accountSyncing?.getIsMultichainAccountSyncingEnabled?.() ??
```
fabiobozzo, line 621: Does that mean extension/mobile shall **inject** a `getIsMultichainAccountSyncingEnabled` which likely read from feature flags?

fabiobozzo (approved):

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
mathieuartu: Yes exactly :) 
And to answer your first question, multichain account syncing will live in the account-tree-controller package because of its immediate proximity to the tree. But it will of course leverage `MultichainAccountService` for the group operations.

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
```diff
@@ -473,6 +478,11 @@ export default class UserStorageController extends BaseController<
       'UserStorageController:getStorageKey',
       this.getStorageKey.bind(this),
     );
+
+    this.messagingSystem.registerActionHandler(
+      'UserStorageController:getIsMultichainAccountSyncingEnabled',
+      this.getIsMultichainAccountSyncingEnabled.bind(this),
+    );
```
mathieuartu, line 485: Exposing the new method as an action handler since `AccountTreeController` will also leverage it

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/setup-subscriptions.ts
```diff
@@ -19,7 +19,10 @@ export function setupAccountSyncingSubscriptions(
       if (
         !canPerformAccountSyncing(options) ||
         !getUserStorageControllerInstance().state
-          .hasAccountSyncingSyncedAtLeastOnce
+          .hasAccountSyncingSyncedAtLeastOnce ||
+        // If multichain account syncing is enabled, we do not push account syncing V1 data anymore.
+        // AccountTreeController handles proper multichain account syncing
+        getUserStorageControllerInstance().getIsMultichainAccountSyncingEnabled()
```
ccharly, line 25: Cosmetic: quick question, any reason why we're duplicating the logic here for both `:account{Added,Renamed}`?

Can't we just make a common function and use it both `subscribe`? (It could be scoped into `setupAccountSyncingSubscriptions` too, so we can capture the `options`)

That would allow to not duplicate the logic and comment. (but yes, my suggestion is mainly "cosmetic")

Something like:

```ts
const syncInternalAccountToUserStorage = (account: InternalAccount) => {
  if (
    !canPerformAccountSyncing(options) ||
    !getUserStorageControllerInstance().state
      .hasAccountSyncingSyncedAtLeastOnce ||
    // If multichain account syncing is enabled, we do not push account syncing V1 data anymore.
    // AccountTreeController handles proper multichain account syncing
    getUserStorageControllerInstance().getIsMultichainAccountSyncingEnabled()
  ) {
    return;
  }
  
  ...
};

...

getMessenger().subscribe('AccountsController:accountAdded', syncInternalAccountToUserStorage);
getMessenger().subscribe('AccountsController:accountRenamed', syncInternalAccountToUserStorage);


```

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/controller-integration.ts
```diff
@@ -30,6 +30,14 @@ export async function saveInternalAccountToUserStorage(
   const saveAccount = async () => {
     const { getUserStorageControllerInstance } = options;
 
+    if (
+      getUserStorageControllerInstance().getIsMultichainAccountSyncingEnabled()
```
ccharly, line 34: Can't we just put this check in `canPerformAccountSyncing` directly? This way all/most of the code would be unchanged?

It seems we are using the function a lot before doing any write operation, so this might work?

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/setup-subscriptions.ts
ccharly: Maybe this suggestion won't make sense if my previous comment about `canPerformAccountSyncing` is enough.

fabiobozzo (dismissed):

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/controller-integration.ts
fabiobozzo: Afaics, `canPerformAccountSyncing` is "one level above" `isMultichainAccountSyncingEnabled`. 

If it can't, well, it can't.  
If it can, then it depends on that feature flag who's responsible for it:
- new scenario: account-tree-controller
- previous scenario: user-storage-controller

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/setup-subscriptions.ts
mathieuartu: While agreeing with your suggestion, since we're going to deprecate this version of account syncing soon, I don't want to introduce potentially confusing diffs in this PR. Does that make sense to you?

packages/profile-sync-controller/src/controllers/user-storage/account-syncing/controller-integration.ts
mathieuartu: Exactly what @fabiobozzo said ❤️

ccharly: But then, that would still be compatible with this no? Like a lower-priority compared to the previous conditions? E.g:
```ts
if (it can't) {
  // Well, it can't.
  return false;
}

if (it can) {
  return getUserStorageControllerInstance().getIsMultichainAccountSyncingEnabled();
}
```

Unless the `canPerformAccountSyncing` is sometimes used in some other context that are not impacted by multichain accounts? 🤔

PR #6218, opened 2025-07-31
perf: stop updating `selectedAccount` unnecessarily

FrederikBolding:
## Explanation

Return early when handling network switches if the account ID is already selected to prevent unnecessary state updates.

### DIFF
--- a/packages/accounts-controller/src/AccountsController.ts
+++ b/packages/accounts-controller/src/AccountsController.ts
@@ -604,7 +604,6 @@ export class AccountsController extends BaseController<
     }
 
     this.#update((state) => {
-      // @ts-expect-error "Type instantiation is excessively deep"
       state.internalAccounts.accounts = internalAccounts;
     });
   }


packages/accounts-controller/src/AccountsController.ts
```diff
@@ -600,6 +604,7 @@ export class AccountsController extends BaseController<
     }
 
     this.#update((state) => {
+      // @ts-expect-error "Type instantiation is excessively deep"
```
FrederikBolding: Unsure why this shows up now 🫠

ccharly: Well yes... I do have the same problem everytime I run the tests locally... It fails on locally without it, but it makes the CI fails if you leave it there 🙃

So you'll have to remove it unfortunately 😅

FrederikBolding: Not in CI apparently 😓

PR #6221, opened 2025-07-31
feat(account-tree-controller): add persistent metadata for account groups and wallets

fabiobozzo:
## Explanation

This PR adds persistence capabilities to the `AccountTreeController` to support user customizations like custom names, pinning/hiding account groups, and sync metadata. Previously, the controller was stateless and would lose all user customizations on each tree rebuild.

### Current State & Problem
The `AccountTreeController` currently rebuilds the account tree from scratch using rules every time it initializes, resulting in:
- Loss of custom names set by users for wallets and account groups
- No support for pinning/hiding account groups in the UI
- No persistence of sync metadata required for Account Syncing V2
- Poor user experience as customizations don't survive app restarts or tree rebuilds

### Solution Overview
This implementation adds **persistent metadata storage** with two new state properties that store user customizations separately from the tree structure:

- `accountGroupsMetadata: Record<AccountGroupId, AccountGroupMetadata>`
- `accountWalletsMetadata: Record<AccountWalletId, AccountWalletMetadata>`

### Key Changes

**Persistent State Structure**:
- Added `AccountGroupMetadata` type supporting `name`, `pinned`, `hidden`, and `lastUpdatedAt` fields
- Added `AccountWalletMetadata` type supporting `name` and `lastUpdatedAt` fields
- Metadata is stored in dedicated maps indexed by stable IDs

**Enhanced Tree Building Process**:
- Modified existing `#renameAccountWalletIfNeeded()` and `#renameAccountGroupIfNeeded()` methods to prioritize persisted custom names
- Customizations are injected during tree construction rather than requiring separate tree traversals
- Maintains the existing two-phase approach: structure building + customization application

**New Public Methods**:
- `setAccountGroupName(groupId, name)` - Sets custom name for account groups
- `setAccountWalletName(walletId, name)` - Sets custom name for wallets  
- `setAccountGroupPinned(groupId, pinned)` - Controls group pinning state
- `setAccountGroupHidden(groupId, hidden)` - Controls group visibility

**Optimization Strategy**:
- Leverages stable wallet/group IDs for O(1) metadata lookups during tree building
- Avoids expensive tree traversals by applying customizations inline during construction
- Automatic tree rebuilding when metadata changes to ensure consistency

### Implementation Details

The persistence follows a **metadata overlay pattern** where:
1. Base tree structure is built using existing rules (Entropy → Snap → Keyring)
2. Persisted customizations are applied during the naming phase if they exist
3. Default rule-generated names are used as fallbacks
4. All metadata updates trigger automatic tree rebuilds to maintain consistency

This approach ensures backward compatibility while adding the persistence layer needed for Account Syncing V2 and advanced UI features like pinning/hiding.

## References

- Supports Account Syncing V2 metadata requirements (`lastUpdatedAt` timestamps for conflict resolution)
- Enables pinning and hiding account groups functionality
- Foundation for persistent user customizations across app sessions

## Checklist


packages/account-tree-controller/src/types.ts
```diff
@@ -22,6 +22,34 @@ import type {
...
+ */
+export type AccountGroupMetadata = {
+  /** Custom name set by user, overrides default naming logic */
+  name?: string;
+  /** Whether this group is pinned in the UI */
+  pinned?: boolean;
+  /** Whether this group is hidden in the UI */
+  hidden?: boolean;
+  /** Timestamp of last metadata update for sync conflict resolution */
+  lastUpdatedAt?: number;
```
mathieuartu: Question: shouldn't we want to expand on the existing metadata types and define those properties there? 
For example:
- [Here](https://github.com/MetaMask/core/blob/1a9757aa6f15c5326e93f8453fa974daef9f58c7/packages/account-tree-controller/src/wallet.ts#L32-L34) for `WalletObject`
- [Here](https://github.com/MetaMask/core/blob/1a9757aa6f15c5326e93f8453fa974daef9f58c7/packages/account-tree-controller/src/group.ts#L24-L26) for `AccountGroupObject`

packages/account-tree-controller/src/types.ts
```diff
@@ -30,6 +58,10 @@ export type AccountTreeControllerState = {
     };
     selectedAccountGroup: AccountGroupId | '';
   };
+  /** Persistent metadata for account groups (names, pinning, hiding, sync timestamps) */
+  accountGroupsMetadata: Record<AccountGroupId, AccountGroupMetadata>;
```
mathieuartu: And then here have 
- `Record<AccountGroupId, AccountGroupObject['metadata']>`
- `Record<AccountWalletId, AccountWalletObject['metadata']>`

packages/account-tree-controller/src/types.ts
```diff
@@ -22,6 +22,34 @@ import type {
...
+  lastUpdatedAt?: number;
+};
+
+/**
+ * Metadata for persisting account wallet customizations.
+ * This includes user customizations like names, UI states, and sync timestamps.
+ */
+export type AccountWalletMetadata = {
+  /** Custom name set by user, overrides default naming logic */
+  name?: string;
```
mathieuartu: Question: Should those properties be required instead?

mathieuartu (commented):
Will review more in-depth tomorrow, added comments in the meantime :) 

Also, we will need to change the shape of metadata properties following the discussions we recently had: 
- [Slack message](https://consensys.slack.com/archives/C08R8HGNFDH/p1753979641534629)

So that they are all objects that contain a `value` and a `lastUpdatedAt` property.
Example (only a suggestion, let's discuss with everybody :)) :

```ts
export type UpdatableField<T> = { value: T; lastUpdatedAt: number };

...

type IsAccountWalletObject<
  Type extends {
    type: AccountWalletType;
    id: AccountWalletId;
    groups: {
      [groupId: AccountGroupId]: AccountGroupObject;
    };
    metadata: {
      name: UpdatableField<string>;
      pinned: UpdatableField<boolean>;
      // ...etc
    };
  },
> = Type;
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+   * @param groupId - The account group ID.
+   * @param name - The custom name to set.
+   */
+  setAccountGroupName(groupId: AccountGroupId, name: string): void {
+    this.update((state) => {
+      state.accountGroupsMetadata[groupId] = {
+        ...state.accountGroupsMetadata[groupId],
+        name,
+        lastUpdatedAt: Date.now(),
+      };
```
ccharly: What about just getting a reference to the metadata object and just mutate what we need?

This notation here would probably do "unnecessary mutations" for other fields (speaking of `immer` mutations). I'm not sure we would notice performance-wise, but I still think we should only mutate the `name` and `lastUpdatedAt` fields here.

```suggestion
      const metadata = state.accountGroupsMetadata[groupId];
      metadata.name = name;
      metadata.lastUpdatedAt = Date.now();
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
 
+  /**
+   * Sets a custom name for an account group.
+   *
+   * @param groupId - The account group ID.
+   * @param name - The custom name to set.
+   */
+  setAccountGroupName(groupId: AccountGroupId, name: string): void {
+    this.update((state) => {
+      state.accountGroupsMetadata[groupId] = {
```
ccharly: We should probably check that the `groupId` also exists in this map before using it.

IMO, throwing if the `groupId` is not "valid" would make sense here.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+        ...state.accountGroupsMetadata[groupId],
+        name,
+        lastUpdatedAt: Date.now(),
+      };
+    });
+
+    // Rebuild tree to apply the new name
+    this.init();
```
ccharly: This seems "too much" to me. Though, we don't have any "contexts" or "mapping" using a `groupId`.

Maybe we should add one 🤔 so we could update the group node directly.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+    this.update((state) => {
+      state.accountWalletsMetadata[walletId] = {
+        ...state.accountWalletsMetadata[walletId],
+        name,
+        lastUpdatedAt: Date.now(),
+      };
+    });
+
+    // Rebuild tree to apply the new name
+    this.init();
```
ccharly: Same here.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+   * @param groupId - The account group ID.
+   * @param pinned - Whether the group should be pinned.
+   */
+  setAccountGroupPinned(groupId: AccountGroupId, pinned: boolean): void {
+    this.update((state) => {
+      state.accountGroupsMetadata[groupId] = {
+        ...state.accountGroupsMetadata[groupId],
+        pinned,
+        lastUpdatedAt: Date.now(),
+      };
```
ccharly: Same here.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+   * @param groupId - The account group ID.
+   * @param hidden - Whether the group should be hidden.
+   */
+  setAccountGroupHidden(groupId: AccountGroupId, hidden: boolean): void {
+    this.update((state) => {
+      state.accountGroupsMetadata[groupId] = {
+        ...state.accountGroupsMetadata[groupId],
+        hidden,
+        lastUpdatedAt: Date.now(),
+      };
```
ccharly: Same here.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -516,6 +538,95 @@ export class AccountTreeController extends BaseController<
...
+  setAccountWalletCollapsed(
+    walletId: AccountWalletId,
+    collapsed: boolean,
+  ): void {
+    this.update((state) => {
+      state.accountWalletsMetadata[walletId] = {
+        ...state.accountWalletsMetadata[walletId],
+        collapsed,
+        lastUpdatedAt: Date.now(),
+      };
```
ccharly: Same here.

packages/account-tree-controller/src/types.ts
```diff
@@ -22,6 +22,34 @@ import type {
...
+  /** Whether this wallet is collapsed in the UI */
+  collapsed?: boolean;
```
ccharly: Not sure we have a need for this yet. Or at least, we don't need to persist it on a controller's state IMO.

packages/account-tree-controller/src/types.ts
ccharly: Yes, the "metadata" might becomes ambiguous... I think we should have 2 types like:
```ts
type Account{Group,Wallet}PersistedMetadata = { ... };
type Account{Group,Wallet}Metadata = { ... };
```

and intersect the 2 types together on the `*Object` types.

This way, [we could avoid having this `Omit<..., 'name'>`](https://github.com/MetaMask/core/blob/main/packages/account-tree-controller/src/rule.ts#L23-L29) which kinda feels out-of-place (I was not a big fan of it when I added it, but I think it makes sense to "move" the `name` to some other types now).

packages/account-tree-controller/src/types.ts
ccharly: If we go with having `Account{Group,Wallet}PersistedMetadata`, we could use those here instead.

fabiobozzo: Pardon me, perhaps I don't understand correctly. As far as I understand: 

Option 1 (me, Charly): separate persistent and "tree building" types
```
// Controller state
type AccountGroupPersistedMetadata = {
  name?: UpdatableField<string>;
  pinned?: UpdatableField<boolean>;
  hidden?: UpdatableField<boolean>;
};

// During tree building, objects could extract .value from UpdatableField.value
type AccountGroupTreeMetadata = {
  name: string;           
  pinned: boolean;     
  hidden: boolean;     
};
```

Option 2 (Mathieu): unified
```
// Tree objects store UpdatableField directly
type AccountGroupMetadata = {
  name: UpdatableField<string>;
  entropy: { groupIndex: number };
  pinned: UpdatableField<boolean>;
  hidden: UpdatableField<boolean>;
};
```
In the latter option, the controller doesn't use a different type, so: `accountGroupsMetadata: Record<AccountGroupId, AccountGroupObject['metadata']>;`

BUT... the UI should be adjusted to access `group.metadata.pinned.value`.. no? @ccharly @mathieuartu

packages/account-tree-controller/src/types.ts
fabiobozzo: Optional makes sense to me because not all groups/wallets will have custom names initially. @ccharly do we prefer null values by default (empty strings or so) ?

packages/account-tree-controller/src/types.ts
fabiobozzo: Not needed indeed, I got confused 👎

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: I introduced a `readonly #groupIdToWalletId: Map<AccountGroupId, AccountWalletId>;` for fast lookup. Wdyt?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Fixed in all setters ✔️

packages/account-tree-controller/src/types.ts
ccharly: Here's was I had in mind to "extract" a "normal object" out from a `*PersistedMetadata` type:
- https://www.typescriptlang.org/play/?#code/C4TwDgpgBAqmAmBDYiBGAbCAxAlhd8APAGqLoCuEAfFALxQDeAUFKwG5mUBcUpFEAbhat0iAM7A4SYBHgBBYDwB25ALaoIAJyEBfJk1CQoAFQD2fSoVz54NetYJQIADxlL4Y2AmRpMDouRKANZKpgDuSjQA-FD+ANoA5Bz8CQC6UDz+Qgbg0GYA8qgAVhAAxsCEwoUl5RbQLm4eUABKZaaaRBKaOEoA5gA0UIEh4ZH9THaMwnEA0hAgUD1QQfOmAGZQ1WXAdak8ZnWEW7WcELPzqVRMejlGAApaYjgSsgCyECjSiHRTrFBKiFUEB4Uh8GGweAIhC6PV6VCEfxkEh4qFMpkwiCUun0hmg70+Ph+BWK20ID00Txe8HxiC+8KAA

Which is:
```ts
type UpdatableField<Value> = {
   value: Value;
   lastUpdatedAt: number;
}

type ToValue<Field> = Field extends UpdatableField<unknown> ? Field['value'] : Field;

type ToObject<
  ObjectValue extends Record<string, unknown>,
> = {
  [Key in keyof ObjectValue]: ToValue<ObjectValue[Key]>
}

type PersistedMetadata = {
   name: UpdatableField<string>;
   test: boolean;
}

type Metadata = ToObject<PersistedMetadata>;
```

Which then could be used:
```ts
type AccountGroupPersistedMetadata = {
  name: UpdatableField<string>;
  pinned: UpdatableField<boolean>;
  hidden: UpdatableField<boolean>;
}

type AccountGroupMultichainAccountObject = {
  type: AccountGroupType.MultichainAccount;
  id: MultichainAccountGroupId;
  // Blockchain Accounts (at least 1 account per multichain-accounts):
  accounts: [AccountId, ...AccountId[]];
  metadata: ToObject<AccountGroupPersistedMetadata> & {
    entropy: {
      groupIndex: number;
    };
  };
```

> [!NOTE]
> The `ToObject` name is "too generic" IMO, but you get the idea!

fabiobozzo:
Verifier draft PR:
- https://github.com/MetaMask/metamask-extension/pull/34830

packages/account-tree-controller/src/types.ts
```diff
@@ -20,8 +20,81 @@ import type {
...
+export type ExtractValues<ObjectValue extends Record<string, unknown>> = {
+  [Key in keyof ObjectValue]: ToValue<ObjectValue[Key]>;
+};
+
+/**
+ * Type utility to extract plain values and make all properties required (for tree objects).
+ */
+export type ExtractRequiredValues<ObjectValue extends Record<string, unknown>> =
+  {
+    [Key in keyof ObjectValue]-?: ToValue<ObjectValue[Key]>;
```
ccharly: WDYT about renaming to:

```suggestion
export type ExtractFieldValue<Field> =
  Field extends UpdatableField<unknown> ? Field['value'] : Field;

/**
 * Type utility to extract plain values from an object with UpdatableField properties.
 */
export type ExtractFieldValues<ObjectValue extends Record<string, unknown>> = {
  [Key in keyof ObjectValue]: ExtractFieldValue<ObjectValue[Key]>;
};

/**
 * Type utility to extract plain values and make all properties required (for tree objects).
 */
export type ExtractRequiredFieldValues<ObjectValue extends Record<string, unknown>> =
  {
    [Key in keyof ObjectValue]-?: ExtractFieldValue<ObjectValue[Key]>;
```

packages/account-tree-controller/src/types.ts
```diff
@@ -20,8 +20,81 @@ import type {
...
+ * Type utility to extract plain values from an object with UpdatableField properties.
+ */
+export type ExtractValues<ObjectValue extends Record<string, unknown>> = {
+  [Key in keyof ObjectValue]: ToValue<ObjectValue[Key]>;
+};
+
+/**
+ * Type utility to extract plain values and make all properties required (for tree objects).
+ */
+export type ExtractRequiredValues<ObjectValue extends Record<string, unknown>> =
```
ccharly: Maybe we could move all those utility types to a new file like `src/typing.ts` maybe?

packages/account-tree-controller/src/types.ts
```diff
@@ -20,8 +20,81 @@ import type {
...
+// Backward compatibility aliases using indexed access types
+/**
+ * @deprecated Use AccountGroupTreeMetadata for tree objects or AccountGroupPersistedMetadata for controller state
+ */
+export type AccountGroupMetadata = AccountGroupObject['metadata'];
+
+/**
+ * @deprecated Use AccountWalletTreeMetadata for tree objects or AccountWalletPersistedMetadata for controller state
+ */
+export type AccountWalletMetadata = AccountWalletObject['metadata'];
```
ccharly: I'd move each of those types to their proper files:
```
Account*Group* -> group.ts
Account*Wallet* -> wallet.ts
```

WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -336,6 +376,26 @@ export class AccountTreeController extends BaseController<
...
+
+  /**
+   * Validates that a wallet exists in the current account tree.
+   *
+   * @param walletId - The account wallet ID to validate.
+   * @returns True if the wallet exists, false otherwise.
+   */
+  #walletExistsInTree(walletId: AccountWalletId): boolean {
+    return Boolean(this.state.accountTree.wallets[walletId]);
+  }
```
ccharly: WDYT about making those method like "assert methods" instead, like:

```suggestion
  /**
   * Validates that a group exists in the current account tree.
   *
   * @param groupId - The account group ID to validate.
   * @returns True if the group exists, false otherwise.
   */
  #assertAccountGroupExists(groupId: AccountGroupId): void {
    const exists = this.#groupIdToWalletId.has(groupId);
    if (!exists) {
      throw new Error(`Account group with ID "${groupId}" not found in tree`);
    }
  }

  /**
   * Validates that a wallet exists in the current account tree.
   *
   * @param walletId - The account wallet ID to validate.
   * @returns True if the wallet exists, false otherwise.
   */
  #assertAccountWalletExists(walletId: AccountWalletId): void {
    const exists = Boolean(this.state.accountTree.wallets[walletId]);
    if (!exists) {
      throw new Error(`Account wallet with ID "${walletId}" not found in tree`);
    }
  }
```

This way, we don't repeat error messages + having strong preconditions (like expecting a group/wallet to exist) make sense to me in many contexts.

WDYT?

packages/account-tree-controller/src/types.ts
fabiobozzo: It makes more sense, yeah! 😄 Domain-specific types live with their related logic.

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: 💯 Agreed! I like the fail-fast approach too, it makes debugging easier. Those thrown errors also go to... Sentry, right?

packages/account-tree-controller/src/types.ts
fabiobozzo: It is sorted out @mathieuartu if you wanna have a 2nd look, I followed Charly's approach on this.

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: They might end up in Sentry if not being catch yes. In anyway, the way I see this is, we should never really hit those frequently, and I also prefer this "fail-fast approach".

You have some preconditions, if the caller/consumer fails to honor/respect them, then you just throw right away, which works really well here IMO.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -314,11 +349,16 @@ export class AccountTreeController extends BaseController<
...
+          pinned: false, // Default UI state
+          hidden: false, // Default UI state
         },
         // We do need to type-cast since we're not narrowing `result` with
         // the union tag `result.group.type`.
       } as AccountGroupObject;
       group = wallet.groups[groupId];
+
+      // Map group ID to its containing wallet ID for efficient direct access
+      this.#groupIdToWalletId.set(groupId, walletId);
```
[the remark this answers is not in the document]
fabiobozzo: Issue 1 was definitely a real bug that needed fixing.  

Issue 2 now has defensive programming that ensures the mapping is always consistent, even in edge cases we might not have considered.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -166,46 +185,62 @@ export class AccountTreeController extends BaseController<
...
+        group.metadata.name = this.#getSnapRule().getDefaultAccountGroupName(
+          // Same here.
+          wallet.groups[group.id],
+        );
+      } else {
+        group.metadata.name = this.#getKeyringRule().getDefaultAccountGroupName(
+          // Same here.
+          wallet.groups[group.id],
+        );
+      }
```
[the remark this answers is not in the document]
fabiobozzo: Ok. I allowed for user-defined empty-string account names.

fabiobozzo:
@metamaskbot publish-preview

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add setter methods for setting custom account group names, wallet names and their pinning state and visibility ([#6221](https://github.com/MetaMask/core/pull/6221))
 - Add `group.type` tag ([#6214](https://github.com/MetaMask/core/pull/6214))
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
 
 ### Changed
 
+- Enhanced customization priority hierarchy in tree building ([#6221](https://github.com/MetaMask/core/pull/6221))
+
```
```suggestion
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Custom user names now take priority over default rule-generated names
+
 - **BREAKING:** Bump peer dependency `@metamask/account-api` from `^0.3.0` to `^0.7.0` ([#6214](https://github.com/MetaMask/core/pull/6214)), ([#6216](https://github.com/MetaMask/core/pull/6216)), ([#6222](https://github.com/MetaMask/core/pull/6222))
+
```
```suggestion
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add persistence support for user customizations ([#6221](https://github.com/MetaMask/core/pull/6221))
+  - New `accountGroupsMetadata` (of new type `AccountGroupMetadata`) and `accountWalletsMetadata` (of new type `AccountWalletMetadata`) state properties to persist custom names, pinning, and hiding states
+  - Custom names and metadata survive controller initialization and tree rebuilds
+  - Support for `lastUpdatedAt` timestamps for Account Syncing V2 compatibility
```
```suggestion
  - New `accountGroupsMetadata` (of new type `AccountGroupPersistedMetadata`) and `accountWalletsMetadata` (of new type `AccountWalletPersistedMetadata`) state properties to persist custom names, pinning, and hiding states.
  - Custom names and metadata survive controller initialization and tree rebuilds.
  - Support for `lastUpdatedAt` timestamps for Account Syncing V2 compatibility.
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Add `group.type` tag ([#6214](https://github.com/MetaMask/core/pull/6214))
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
 
 ### Changed
 
+- Enhanced customization priority hierarchy in tree building ([#6221](https://github.com/MetaMask/core/pull/6221))
+
+  - Custom user names now take priority over default rule-generated names
```
```suggestion
  - Custom user names now take priority over default rule-generated names.
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1568,10 +1598,357 @@ describe('AccountTreeController', () => {
...
+      }).toThrow(
+        'Account group with ID "non-existent-group-id" not found in tree',
+      );
+
+      // Should throw for non-existent wallet operations
+      expect(() => {
+        controller.setAccountWalletName(nonExistentWalletId, 'Test Wallet');
+      }).toThrow(
+        'Account wallet with ID "non-existent-wallet-id" not found in tree',
+      );
```
ccharly: Nit:

```suggestion
      expect(() => {
        controller.setAccountGroupName(nonExistentGroupId, 'Test Name');
      }).toThrow(
        `Account group with ID "${nonExistentGroupId}" not found in tree`,
      );

      expect(() => {
        controller.setAccountGroupPinned(nonExistentGroupId, true);
      }).toThrow(
        `Account group with ID "${nonExistentGroupId}" not found in tree`,
      );

      expect(() => {
        controller.setAccountGroupHidden(nonExistentGroupId, true);
      }).toThrow(
        `Account group with ID "${nonExistentGroupId}" not found in tree`,
      );

      // Should throw for non-existent wallet operations
      expect(() => {
        controller.setAccountWalletName(nonExistentWalletId, 'Test Wallet');
      }).toThrow(
        `Account wallet with ID "${nonExistentWalletId}" not found in tree`,
      );
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -92,6 +105,8 @@ export class AccountTreeController extends BaseController<
     // Reverse map to allow fast node access from an account ID.
     this.#accountIdToContext = new Map();
 
+    this.#groupIdToWalletId = new Map();
```
ccharly, line 108: Very nit: Just for consistency:

```suggestion
    // Reverse map to allow fast wallet node access from a group ID.
    this.#groupIdToWalletId = new Map();
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -63,6 +73,9 @@ export class AccountTreeController extends BaseController<
 > {
   readonly #accountIdToContext: Map<AccountId, AccountContext>;
 
+  // Mapping for direct tree node access
```
ccharly: Cosmetic: We could remove this, if we add the other comment I made

```suggestion
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -313,14 +348,26 @@ export class AccountTreeController extends BaseController<
...
     } else {
       group.accounts.push(account.id);
+
+      // Assertion: mapping should always exist for existing groups
+      /* istanbul ignore if: This should never happen in normal operation */
+      if (!this.#groupIdToWalletId.has(groupId)) {
+        throw new Error(
+          `Internal inconsistency: mapping missing for group ${groupId}`,
+        );
+      }
```
ccharly: Why do we have this here? I think it makes sense to "assert" preconditions when we "consume" our state/parameters, but I don't think we need this here

packages/account-tree-controller/src/group.ts
```diff
@@ -6,9 +6,31 @@ import type { AccountGroup, AccountGroupId } from '@metamask/account-api';
...
+  /** Whether this group is pinned in the UI */
+  pinned?: UpdatableField<boolean>;
+  /** Whether this group is hidden in the UI */
+  hidden?: UpdatableField<boolean>;
+};
+
+/**
+ * Tree metadata for account groups (required plain values extracted from persisted metadata).
+ */
+export type AccountGroupTreeMetadata =
```
ccharly: Should be named that way to map the "main group type" `AccountTreeGroup`

```suggestion
export type AccountTreeGroupMetadata =
```

packages/account-tree-controller/src/wallet.ts
```diff
@@ -15,8 +15,26 @@ import type {
...
+ */
+export type AccountWalletPersistedMetadata = {
+  /** Custom name set by user, overrides default naming logic */
+  name?: UpdatableField<string>;
+};
+
+/**
+ * Tree metadata for account wallets (required plain values extracted from persisted metadata).
+ */
+export type AccountWalletTreeMetadata =
```
ccharly: Same comment here:

```suggestion
export type AccountTreeWalletMetadata =
```

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: I introduce that to silence cursor: https://github.com/MetaMask/core/pull/6221/files#r2251702319 
But I agree w/ you, if we keep the `this.#groupIdToWalletId` properly in sync as I did in #insert we shouldn't enforce it randomly like I do here.

hmalik88 (dismissed):
LGTM!

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,13 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
 
 ### Changed
 
+- Enhanced customization priority hierarchy in tree building ([#6221](https://github.com/MetaMask/core/pull/6221))
+
+  - Custom user names now take priority over default rule-generated names.
+
```
ccharly: We usually don't leave any empty lines in the changelogs:

```suggestion
- Enhanced customization priority hierarchy in tree building ([#6221](https://github.com/MetaMask/core/pull/6221))
  - Custom user names now take priority over default rule-generated names.
```

packages/account-tree-controller/src/group.ts
```diff
@@ -6,9 +6,31 @@ import type { AccountGroup, AccountGroupId } from '@metamask/account-api';
...
+  UpdatableField,
+  ExtractRequiredFieldValues,
+} from './type-utils.js';
 import type { AccountTreeControllerMessenger } from './types';
 import type { AccountTreeWallet } from './wallet';
 
+/**
+ * Persisted metadata for account groups (stored in controller state for persistence/sync).
+ */
+export type AccountGroupPersistedMetadata = {
```
ccharly: Actually, I think we should also rename this too. Since the `AccountGroup` type is the "interface"/"concept" of the new account model. The metadata belongs to the `AccountTreeGroup` itself. WDYT?


```suggestion
export type AccountTreeGroupPersistedMetadata = {
```

packages/account-tree-controller/src/group.ts
```diff
@@ -6,9 +6,31 @@ import type { AccountGroup, AccountGroupId } from '@metamask/account-api';
...
+  pinned?: UpdatableField<boolean>;
+  /** Whether this group is hidden in the UI */
+  hidden?: UpdatableField<boolean>;
+};
+
+/**
+ * Tree metadata for account groups (required plain values extracted from persisted metadata).
+ */
+export type AccountTreeGroupMetadata =
+  ExtractRequiredFieldValues<AccountGroupPersistedMetadata>;
```
ccharly: Actually, I think we should rely on the built-in `Required` utility type here:

```suggestion
  Required<ExtractFieldValues<AccountTreeGroupPersistedMetadata>>;
```

packages/account-tree-controller/src/type-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+/**
+ * Type utility to extract plain values and make all properties required (for tree objects).
+ */
+export type ExtractRequiredFieldValues<
+  ObjectValue extends Record<string, unknown>,
+> = {
+  [Key in keyof ObjectValue]-?: ExtractFieldValue<ObjectValue[Key]>;
+};
```
ccharly: I don't think we really need this, since we can combine `Required` and `ExtractFieldValues` for this.

packages/account-tree-controller/src/wallet.ts
```diff
@@ -15,8 +15,26 @@ import type {
...
+import type {
+  UpdatableField,
+  ExtractRequiredFieldValues,
+} from './type-utils.js';
 import { type AccountTreeControllerMessenger } from './types';
 
+/**
+ * Persisted metadata for account wallets (stored in controller state for persistence/sync).
+ */
+export type AccountWalletPersistedMetadata = {
```
ccharly: Same here, I think `AccounTreeWalletPersistedMetadata` would be better.

```suggestion
export type AccountTreeWalletPersistedMetadata = {
```

packages/account-tree-controller/src/wallet.ts
```diff
@@ -15,8 +15,26 @@ import type {
...
+  /** Custom name set by user, overrides default naming logic */
+  name?: UpdatableField<string>;
+};
+
+/**
+ * Tree metadata for account wallets (required plain values extracted from persisted metadata).
+ */
+export type AccountTreeWalletMetadata =
+  ExtractRequiredFieldValues<AccountWalletPersistedMetadata>;
```
ccharly: And same here, we can use `Required` here too:

```suggestion
  Required<ExtractFieldValues<AccountTreeWalletPersistedMetadata>?;
```

packages/account-tree-controller/src/type-utils.ts
fabiobozzo: Good catch! I didn't know about that `Required` 😑 my bad.

ccharly (approved):
LGTM! Thanks for addressing all my comments! 😄

PR #6222, opened 2025-07-31
feat(multichain-account-service): add multichain account create operations support

ccharly:
## Explanation

Adding new methods to create a multichain account for a given index or the next index.

## References

N/A

## Changelog

N/A

## Checklist

### DIFF
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -190,8 +190,13 @@ export class MultichainAccountWallet<
    * @returns The next group index of this wallet.
    */
   getNextGroupIndex(): number {
-    // Assuming we cannot have indexes gaps.
-    return this.#accounts.size; // No +1 here, group indexes starts at 0.
+    // We do not check for gaps.
+    return (
+      Math.max(
+        -1, // So it will default to 0 if no groups.
+        ...this.#accounts.keys(),
+      ) + 1
+    );
   }
 
   /**
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -92,7 +92,10 @@ export class MultichainAccountWallet<
     }
 
     // Now force-sync all remaining multichain accounts.
-    for (const [groupIndex, multichainAccount] of this.#accountGroups.entries()) {
+    for (const [
+      groupIndex,
+      multichainAccount,
+    ] of this.#accountGroups.entries()) {
       multichainAccount.sync();
 
       // Clean up old multichain accounts.
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -248,9 +248,17 @@ export class MultichainAccountWallet<
     // just create the missing accounts, and previously created accounts will just be
     // re-used as is.
     if (results.some((result) => result.status === 'rejected')) {
-      throw new Error(
-        `Unable to create multichain account group for index: ${groupIndex}`,
-      );
+      const error = `Unable to create multichain account group for index: ${groupIndex}`;
+
+      let warn = `${error}:`;
+      for (const result of results) {
+        if (result.status === 'rejected') {
+          warn += `\n- ${result.reason}`;
+        }
+      }
+      console.warn(warn);
+
+      throw new Error(error);
     }
 
     // Because of the :accountAdded automatic sync, we might already have created the
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -234,25 +234,33 @@ export class MultichainAccountWallet<
       ),
     );
 
+    // --------------------------------------------------------------------------------
     // READ THIS CAREFULLY:
+    //
     // Since we're not "fully supporting multichain" for now, we still rely on single
     // :accountCreated events to sync multichain account groups and wallets. Which means
-    // that even if of the provider fails, some accouns (on the `AccountsController`
-    // might be showing up AND that some multichain account wallets/groups might have
-    // been automatically sync'd too.
+    // that even if of the provider fails, some accounts will still be created on some
+    // other providers and will become "available" on the `AccountsController`, like:
     //
-    // We will need to unsubscribe from :accountAdded once we "fully support multichain"
-    // (meaning, the entire codebase relies on this new service for multichain account
-    // creations).
+    // 1. Creating a multichain account group for index 1
+    // 2. EvmAccountProvider.createAccounts returns the EVM account for index 1
+    //   * AccountsController WILL fire :accountCreated for this account
+    //   * This account WILL BE "available" on the AccountsController state
+    // 3. SolAccountProvider.createAccounts fails to create a Solana account for index 1
+    //   * AccountsController WON't fire :accountCreated for this account
+    //   * This account WON'T be "available" on the Account
+    // 4. MultichainAccountService will receive a :accountCreated for the EVM account from
+    // step 2 and will create a new multichain account group for index 1, but it won't
+    // receive any event for the Solana account of this group. Thus, this group won't be
+    // "aligned" (missing "blockchain account" on this group).
+    //
     // --------------------------------------------------------------------------------
 
     // If any of the provider failed to create their accounts, then we consider the
     // multichain account group to have failed too.
-    // NOTE: Though, we don't rollback existing created accounts and that's totally fine
-    // because account creations is assumed to be imdepotent, thus, trying again should
-    // just create the missing accounts, and previously created accounts will just be
-    // re-used as is.
     if (results.some((result) => result.status === 'rejected')) {
+      // NOTE: Some accounts might still have been created on other account providers. We
+      // don't rollback them.
       const error = `Unable to create multichain account group for index: ${groupIndex}`;
 
       let warn = `${error}:`;
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -1,23 +1,23 @@
+import {
+  getGroupIndexFromMultichainAccountGroupId,
+  isMultichainAccountGroupId,
+  toMultichainAccountWalletId,
+} from '@metamask/account-api';
+import { toDefaultAccountGroupId } from '@metamask/account-api';
+import { AccountWalletType } from '@metamask/account-api';
 import type {
   Bip44Account,
   MultichainAccountWalletId,
   MultichainAccountWallet as MultichainAccountWalletDefinition,
 } from '@metamask/account-api';
 import type { AccountGroupId } from '@metamask/account-api';
-import {
-  getGroupIndexFromMultichainAccountId as getGroupIndexFromMultichainAccountGroupId,
-  isMultichainAccountGroupId,
-  toMultichainAccountWalletId,
-} from '@metamask/account-api';
-import { toDefaultAccountGroupId } from '@metamask/account-api';
-import { AccountWalletType } from '@metamask/account-api';
+import type { AccountProvider } from '@metamask/account-api';
 import {
   type EntropySourceId,
   type KeyringAccount,
 } from '@metamask/keyring-api';
 
 import { MultichainAccountGroup } from './MultichainAccountGroup';
-import type { AccountProvider } from './providers/BaseAccountProvider';
 
 /**
  * A multichain account wallet that holds multiple multichain accounts (one multichain account per
--- a/packages/multichain-account-service/src/MultichainAccountWallet.ts
+++ b/packages/multichain-account-service/src/MultichainAccountWallet.ts
@@ -4,7 +4,6 @@ import type {
   MultichainAccountWallet as MultichainAccountWalletDefinition,
 } from '@metamask/account-api';
 import type { AccountGroupId } from '@metamask/account-api';
-import type { AccountProvider } from '@metamask/account-api';
 import {
   getGroupIndexFromMultichainAccountId as getGroupIndexFromMultichainAccountGroupId,
   isMultichainAccountGroupId,
@@ -18,6 +17,7 @@ import {
 } from '@metamask/keyring-api';
 
 import { MultichainAccountGroup } from './MultichainAccountGroup';
+import type { AccountProvider } from './providers/BaseAccountProvider';
 
 /**
  * A multichain account wallet that holds multiple multichain accounts (one multichain account per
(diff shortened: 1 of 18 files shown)


packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -1,12 +1,28 @@
...
+export type AccountProvider<Account extends KeyringAccount> =
+  AccountProviderDefinition<Account> & {
+    createAccounts({
+      entropySource,
+      groupIndex,
+    }: {
+      entropySource: EntropySourceId;
+      groupIndex: number;
+    }): Promise<Account['id'][]>;
+  };
```
ccharly: TODO: Move to `@metamask/account-api`.

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -54,5 +70,36 @@ export abstract class BaseAccountProvider
...
+
   abstract isAccountCompatible(account: Bip44Account<InternalAccount>): boolean;
+
+  abstract createAccounts({
+    entropySource,
+    groupIndex,
+  }: {
+    entropySource: EntropySourceId;
+    groupIndex: number;
+  }): Promise<Bip44Account<InternalAccount>['id'][]>;
```
gantunesr: Is the return type an array to cover the discovery accounts use case?

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -15,4 +17,52 @@ export class SolAccountProvider extends BaseAccountProvider {
...
+    // Create account without any confirmation nor selecting it.
+    const keyringAccount = await createAccount(
+      SolAccountProvider.SOLANA_SNAP_ID,
+      opts,
+      {
+        displayAccountNameSuggestion: false,
+        displayConfirmation: false,
+        setSelectedAccount: false,
+      },
+    );
```
gantunesr, line 62: Do you think we should move this class outside the package? I see that we could just add a flag to manage the capabilities. After thinking about it a bit more, ideally these classes would be part of each snap monorepo but that could limit our ownership over them and create some management complexity around multichain accounts

ccharly: As we discussed yesterday, yes, I do believe with some effort that we could do this.

I might need to think about it a bit more, but here's a brain dump of my ideas:
- We should heavily rely on `Account['id']`, since we're using `KeyringAccount` on the Snap side, but we use `InternalAccount` more often on the extension
  * Using those IDs can easily allow us to map one or the other (and that's why most operations are returning IDs here
- We could use `keyring_listAccounts` instead of filtering the entire `AccountsController` accounts list. The Snap have a smaller list of accounts, which could make more sense as a "provider", something like `getBip44Accounts` would be great.
- We could maintain the list of account providers in another component and have add operations on top of it, to disable/enable providers if we need this kind of control too
  * That being said, once a multichain account group holds an account IDs in its list, we would have to either remove it, or have a way to know that this ID should not be considered anymore based on its provider

I could try to POC something around this and see how that fits with Bitcoin integration 👍

> [!CAUTION]
> I do believe though, that moving to the unified keyring API would/could solve some of the specific implementations
> we have on the `AccountProvider` side. So we could maybe adds even more capabilities/state into
> it to achieve a similar pattern here (no real idea on this yet though)

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
ccharly: I know this looks weird, but one of the problem I faced during my first POC was that, we're ONLY dealing with `KeyringAccount` on the Snap side.

And since I wanted to use `InternalAccount` on the multichain account groups, I decided to use IDs to "reference" accounts so I could resolve them using the `AccountsController`.

Lastly (and probably a bigger problem), when we use `keyring_createAccount`, we get a `KeyringAccount`, and this account will get registered as an internal account later on, though, this part of the flow is "truly async". When we get a `KeyringAccount`, that does not mean we can "fetch" its `InternalAccount` right away! 🙃
- https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/lib/snap-keyring/snap-keyring.ts#L444-L448
- https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/lib/snap-keyring/snap-keyring.ts#L330-L332

To "workaround" this, I decided to rely solely on account ID and fetch the account from the `AccountsController` only when we use "accounts getter" on a multichain account group. (Yes, there could be a race condition right after calling `keyring_createAccount` and those "accounts getter" if the internal account is not yet persisted, but we kinda have this problem today everywhere we deal with account creation + use of this new account in the codebase).

An other alternative would be to solely rely on `KeyringAccount` on the multichain account group and fetch them from their "providers" (Snaps or keyrings), but we would not have access to the account's metadata in that case.

(Sorry, thinking out loud for now: I actually really wonder if we need to use `InternalAccount` anymore with our new groups 🤔 since my last improvements, groups now have metadata too, so maybe we don't need account's metadata anymore 🤔)

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
ccharly: See:
- https://github.com/MetaMask/accounts/pull/337

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
ccharly: So, after discussing with @danroc, everything seems to fall into place now 💪

We should go with `KeyringAccount` (even if we only use the `KeyringAccount.id` under the hood).

And since I'm replacing `InternalAccount` by `KeyringAccount`, I can now use the returned account in the Solana provider too 🎉 (which was not possible beforehand because of the "async flow" for Snap account creation, or it would have been much more difficult to implement it).

- https://github.com/MetaMask/core/pull/6227

Thanks for starting that discussion 😁

gantunesr: Changing `InternalAccount` for `KeyringAccount` seems like the right move here 🫶 

I think you didn't really answer my original question haha

ccharly: Oh yeah sorry, so the reason why I was using account IDs instead accounts was mostly related to the `InternalAccount` case actually 😅

For Snaps, with `keyring_createAccount` we get a `KeyringAccount`, not an `InternalAccount` and the equivalent `InternalAccount` will become available at some point on the `AccountsController` (but the "persistance step" of this account is "truly async" and there's no easy way to `await` it as of now, unless you're listening to `:accountAdded`, which is kinda overkill here).

So that's why I initially went with account IDs instead, so that I can "get" the `InternalAccount` later on when we call `group.getAccount` or `group.getAccounts`.

But now that we are using `KeyringAccount`, we can return them (which looks better too IMO 😄). Internally, we still uses account IDs for now, but we could have more use of them later on too 👍

With this new release, the `AccountProvider` now use `KeyringAccount` for both the `createAccounts` method and the `discoverAndCreateAccounts` method:
- https://github.com/MetaMask/accounts/pull/338

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -173,4 +173,69 @@ export class MultichainAccountWallet<
...
+    // idempotent, so we should get the same accounts and potentially some new accounts (if
+    // some account providers decide to return more of them this time).
+    return this.#createMultichainAccountGroup(groupIndex);
+  }
+
+  async createNextMultichainAccountGroup(): Promise<
+    MultichainAccountGroup<Account>
+  > {
+    return this.createMultichainAccountGroup(this.getNextGroupIndex());
+  }
```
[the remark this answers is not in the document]
ccharly: 1. Yes, true, but we should not allow gap "by design" for the moment
2. I still think `<=` is better here, you can call `createMultichainAccountGroup` with the next index (`=`) or a lower index for existing account (`<`)
3. Yes, also true, and I think I could change the logic to run it in parallel with `Promise.allSettled` and check for statuses.
<br>If none accounts got created, then we also throw (since a multichain account group with no account is considered invalid), though IDK how to report that something went wrong to the consumer 🤔
<br>Maybe having a `isSynced` flag on the `MultichainAccountGroup` could be helpful in that case, so you know that you can force re-sync this group if needed?

packages/multichain-account-service/src/providers/SolAccountProvider.ts
danroc: These providers shouldn't be necessary anymore once we migrate to the new `Keyring` interface, so it might not be worth to move them to the Snaps' repos just to remove them later.

gantunesr: > That being said, once a multichain account group holds an account IDs in its list, we would have to either remove it, or have a way to know that this ID should not be considered anymore based on its provider

Hmm not sure if we should consider this scenario, a provider should either be ON or OFF from build time, allowing it to be changed at run time can create weird edge cases that will add more complexity

gantunesr: ➕ to the keyring refactor point

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -173,4 +173,69 @@ export class MultichainAccountWallet<
...
+      // FIXME: What to do if any provider fails to create accounts?
+      await provider.createAccounts({
+        entropySource: this.#entropySource,
+        groupIndex,
+      });
+    }
+
+    // Re-create and "refresh" the multichain account (we assume all account creations are
+    // idempotent, so we should get the same accounts and potentially some new accounts (if
+    // some account providers decide to return more of them this time).
```
hmalik88: Not sure if I understand this comment, if we're at this point wouldn't the accounts created from this group all be "new"?

packages/multichain-account-service/src/MultichainAccountWallet.ts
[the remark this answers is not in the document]
hmalik88: Maybe we can emit an event stating what accounts were failed to be created and have a manual sync button somewhere that is based on the flag you mentioned above? I do think we should do some error handling here though.

packages/multichain-account-service/src/providers/SolAccountProvider.test.ts
```diff
@@ -38,25 +103,62 @@ function setup({
...
+    'SnapController:handleRequest',
+    mockHandleRequest,
+  );
+
+  messenger.registerActionHandler(
+    'KeyringController:withKeyring',
+    async (_, operation) =>
+      operation({
+        // We type-cast here, since `withKeyring` defaults to `EthKeyring` and the
+        // Snap keyring does really implement this interface (this is expected).
```
hmalik88:
```suggestion
        // Snap keyring doesn't really implement this interface (this is expected).
```

hmalik88 (commented):
Mind adding JSDocs where they are missing?

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Yes, actually this is an old comment that does not make sense anymore. I did remove it.

Basically, a multichain account group could be "partially" composed of BIP-44 accounts, like, it could have 1 EVM account, but no Solana account yet. This is true only for the moment because we're migrating our current account model.

Before I introduced the `sync` method, I was generally "re-building" the same group again to let it re-populate its internal list of accounts, but now we can just use `sync` to re-sync with the account providers.

Lastly, even after re-sync we might still miss a Solana account (in my previous example). This issue will be fixed once we introduce the "alignment mechanism" which will automatically create missing account for a given wallet and group index combo.

packages/multichain-account-service/src/providers/SolAccountProvider.test.ts
ccharly: It's now outdated, thanks for the suggestion though.

packages/multichain-account-service/src/MultichainAccountWallet.ts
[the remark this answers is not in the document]
ccharly: Well, after discussing with @danroc, here's what we thought:
- Run all account creations asynchronously (use `Promise.allSettled`)
- If one of the provider fails to create their accounts
  * Then we throw an error

The reasoning being, creating accounts should never really fail since we don't need any real computations here. If it fails, it's most likely a bug.

However... After checking the code a bit more, I realized that "not creating the multichain account group if a provider fails" is not really possible for the moment. Because we rely on `:accountAdded` to sync the groups/wallets (because of the existing account model). We would need to remove this event handling once the `MultichainAccountService` becomes the only "entrypoint" to create multichain account groups. For the moment, we heavily rely on the `AccountsController` and the existing system.

I'll add some more logs too.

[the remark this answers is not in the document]
hmalik88: Yeah error works too, I was thinking we can display a toast or something in the UI to indicate a failure so that the user is aware and they can try to sync again at some point. Not sure if there's any extra steps to account creation for the currently supported chains (or future ones) e.g. balance check through a 3rd party API or something -- that can possibly be a failure point.

packages/multichain-account-service/src/providers/SolAccountProvider.test.ts
ccharly: Actually, I was wrong, it was still in there... Cursor made the same comment as you 🙈 I applied your suggestion manually, thanks!

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -173,4 +183,116 @@ export class MultichainAccountWallet<
...
+    // (meaning, the entire codebase relies on this new service for multichain account
+    // creations).
+    // --------------------------------------------------------------------------------
+
+    // If any of the provider failed to create their accounts, then we consider the
+    // multichain account group to have failed too.
+    // NOTE: Though, we don't rollback existing created accounts and that's totally fine
+    // because account creations is assumed to be imdepotent, thus, trying again should
+    // just create the missing accounts, and previously created accounts will just be
+    // re-used as is.
```
hmalik88: Do you mind re-writing this comment? It's a bit hard to follow (multi-line comment might be better here). Would appreciate if you expand on these two statements as well:

1. "that even if of the provider fails, some accouns (on the `AccountsController` might be showing up AND that some multichain account wallets/groups might have been automatically sync'd too."
2. "NOTE: Though, we don't rollback existing created accounts and that's totally fine because account creations is assumed to be imdepotent, thus, trying again should just create the missing accounts, and previously created accounts will just be re-used as is.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -173,4 +183,116 @@ export class MultichainAccountWallet<
...
+   * Gets next group index for this wallet.
+   *
+   * @returns The next group index of this wallet.
+   */
+  getNextGroupIndex(): number {
+    // We do not check for gaps.
+    return (
+      Math.max(
+        -1, // So it will default to 0 if no groups.
+        ...this.#accounts.keys(),
```
hmalik88: Nit: We should probably rename this to `accountGroups` as the name is more descriptive. Accounts can be misunderstood to be a singular address.

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Yes... I tried to not "over-comment" this cause this is kinda tricky, but I do agree it's pretty hard to follow.

I'll try to come up with a better explanation 👍

ccharly: I tried something different, LMK what you think: https://github.com/MetaMask/core/pull/6222/commits/e9fbdff5c2ae772299a28aeab360568604d58d40

hmalik88: Yes, better -- thanks!

PR #6226, opened 2025-08-01
controller-level selector for all tokens

bergarces:
## Explanation

Adds new selector for list of tokens and balances based on the selected account group id (BIP44, works with stage 1).

Selectors are made with the same pattern used for controller-level BridgeController selectors.

## Checklist

### DIFF
--- a/eslint-warning-thresholds.json
+++ b/eslint-warning-thresholds.json
@@ -193,11 +193,11 @@
     "n/no-unsupported-features/node-builtins": 1
   },
   "packages/keyring-controller/src/KeyringController.test.ts": {
-    "import-x/namespace": 14,
-    "jest/no-conditional-in-test": 8
+    "import-x/namespace": 5,
+    "jest/no-conditional-in-test": 2
   },
   "packages/keyring-controller/src/KeyringController.ts": {
-    "@typescript-eslint/no-unsafe-enum-comparison": 4,
+    "@typescript-eslint/no-unsafe-enum-comparison": 2,
     "@typescript-eslint/no-unused-vars": 1
   },
   "packages/keyring-controller/tests/mocks/mockKeyring.ts": {
--- a/packages/keyring-controller/CHANGELOG.md
+++ b/packages/keyring-controller/CHANGELOG.md
@@ -14,6 +14,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Bump `@metamask/keyring-api` from `^20.0.0` to `^20.1.0`
   - Bump `@metamask/keyring-internal-api` from `^8.0.0` to `^8.1.0`
 
+### Removed
+
+- **BREAKING:** Removed QR keyring methods ([#6031](https://github.com/MetaMask/core/pull/6031))
+  - The following methods have been removed:
+    - `cancelQRSignRequest`
+    - `cancelQRSynchronization`
+    - `connectQRHardware`
+    - `forgetQRDevice`
+    - `getOrAddQRKeyring`
+    - `getQRKeyring`
+    - `getQRKeyringState`
+    - `resetQRKeyringState`
+    - `restoreQRKeyring`
+    - `submitQRCryptoHDKey`
+    - `submitQRCryptoAccount`
+    - `submitQRSignature`
+    - `unlockQRHardwareWalletAccount`
+  - Consumers can use the `withKeyring` method to select a QR keyring and execute a callback with it as argument.
+- **BREAKING:** Removed `KeyringController:qrKeyringStateChange` event ([#6031](https://github.com/MetaMask/core/pull/6031))
+
 ## [22.1.1]
 
 ### Changed
--- a/packages/keyring-controller/jest.config.js
+++ b/packages/keyring-controller/jest.config.js
@@ -17,10 +17,10 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 94.31,
+      branches: 95.78,
       functions: 100,
-      lines: 98.79,
-      statements: 98.8,
+      lines: 98.68,
+      statements: 98.69,
     },
   },
 
--- a/packages/keyring-controller/package.json
+++ b/packages/keyring-controller/package.json
@@ -48,7 +48,6 @@
   },
   "dependencies": {
     "@ethereumjs/util": "^9.1.0",
-    "@keystonehq/metamask-airgapped-keyring": "^0.14.1",
     "@metamask/base-controller": "^8.1.0",
     "@metamask/browser-passworder": "^4.3.0",
     "@metamask/eth-hd-keyring": "^12.0.0",
@@ -66,7 +65,6 @@
   "devDependencies": {
     "@ethereumjs/common": "^4.4.0",
     "@ethereumjs/tx": "^5.4.0",
-    "@keystonehq/bc-ur-registry-eth": "^0.19.0",
     "@lavamoat/allow-scripts": "^3.0.4",
     "@lavamoat/preinstall-always-fail": "^2.1.0",
     "@metamask/auto-changelog": "^3.4.4",
--- a/packages/keyring-controller/src/KeyringController.test.ts
+++ b/packages/keyring-controller/src/KeyringController.test.ts
@@ -1,8 +1,6 @@
 import { Chain, Common, Hardfork } from '@ethereumjs/common';
 import type { TypedTxData } from '@ethereumjs/tx';
 import { TransactionFactory } from '@ethereumjs/tx';
-import { CryptoHDKey, ETHSignature } from '@keystonehq/bc-ur-registry-eth';
-import { MetaMaskKeyring as QRKeyring } from '@keystonehq/metamask-airgapped-keyring';
 import { Messenger } from '@metamask/base-controller';
 import { HdKeyring } from '@metamask/eth-hd-keyring';
 import {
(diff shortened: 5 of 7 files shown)


packages/assets-controllers/src/selectors/stringify-balance.ts
bergarces, line 1: This comes from extension code (I just changed the input type to use native bigint and replaced the deprecated `substr` to `slice`).

Since this does not convert it to a `Number` at any part of the process, it avoids float representation issues.

This is only used for EVM, as multichain balances already come with the decimals applied.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,517 @@
...
+        const rawBalance = accountBalance.balance || '0x0';
+
+        // TODO: This is just a placeholder that will be removed once this whole selector is removed
+        const nativeToken = {
+          address: getNativeTokenAddress(chainId),
+          decimals: 18,
+          // These fields need to be filled at client level for now
+          image: '',
+          name: 'Ethereum',
+          symbol: 'ETH',
```
bergarces: This metadata still needs to be handled at the FE, since we don't have the necessary information here to determine the correct values (e.g. MATIC or BNB).

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+export const selectAssetsBySelectedAccountGroup = createAssetListSelector(
+  [selectAllAssets, (state) => state.accountTree],
+  (groupAssets, accountTree) => {
+    const { selectedAccountGroup } = accountTree;
+    if (!selectedAccountGroup) {
+      return {};
+    }
+    return groupAssets[selectedAccountGroup] || {};
+  },
+);
```
bergarces, line 403: This is how this selector is being called in the client. Same pattern as Bridge Controller.

https://github.com/MetaMask/metamask-extension/pull/34828/files#diff-4b3624e1f5d0c3944513e937350779a84fedd2082919a08cdd38632c52837119

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+
+        if (!groupAssets[accountGroupId]) {
+          groupAssets[accountGroupId] = {};
+        }
+
+        let groupChainAssets = groupAssets[accountGroupId][chainId];
+        if (!groupChainAssets) {
+          groupChainAssets = [];
+          groupAssets[accountGroupId][chainId] = groupChainAssets;
+        }
```
Prithpal-Sooriya: Thoughts on using nullish coalescing assignment to shortcut setting a fallback/default value?

```suggestion
        groupAssets[accountGroupId] ??= {};
        groupAssets[accountGroupId][chainId] ??= [];
        const groupChainAssets = groupAssets[accountGroupId][chainId];
```

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+
+          if (!groupAssets[accountGroupId]) {
+            groupAssets[accountGroupId] = {};
+          }
+
+          let groupChainAssets = groupAssets[accountGroupId][chainId];
+          if (!groupChainAssets) {
+            groupChainAssets = [];
+            groupAssets[accountGroupId][chainId] = groupChainAssets;
+          }
```
Prithpal-Sooriya: Same nullish coalescing assignment for default/fallback vals

```suggestion
          groupAssets[accountGroupId] ??= {};
          groupAssets[accountGroupId][chainId] ??= [];
          const groupChainAssets = groupAssets[accountGroupId][chainId];
```

we can add a comment if parts need more explaining.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+  ) => {
+    const groupAssets: AssetsByAccountGroup = {};
+
+    for (const [accountId, accountAssets] of Object.entries(multichainTokens)) {
+      for (const assetId of accountAssets) {
+        // TODO: We need a safe way to extract each part of the id (in case of multiple / characters)
+        const [chainId, asset] = assetId.split('/') as [
+          `${string}:${string}`,
+          `${string}:${string}`,
+        ];
```
Prithpal-Sooriya: Hmm I think we can use `parseCaipAssetType`, but keep in mind this can throw.

```ts
// maybe we should wrap in try/catch? or YOLO
const caipAsset = parseCaipAssetType(assetId);
const { chainId } = caipAsset;
const asset = `${caipAsset.assetNamespace}:${caipAsset.assetReference}`;
```

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+
+        if (!groupAssets[accountGroupId]) {
+          groupAssets[accountGroupId] = {};
+        }
+
+        let groupChainAssets = groupAssets[accountGroupId][chainId];
+        if (!groupChainAssets) {
+          groupChainAssets = [];
+          groupAssets[accountGroupId][chainId] = groupChainAssets;
+        }
```
Prithpal-Sooriya: Same nullish coalescing assignment for default/fallback vals

```suggestion
        groupAssets[accountGroupId] ??= {};
        groupAssets[accountGroupId][chainId] ??= [];
        const groupChainAssets = groupAssets[accountGroupId][chainId];
```

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,512 @@
...
+      for (const [network, chainAssets] of Object.entries(accountAssets)) {
+        const existingNetworkAssets = existingAccountGroupAssets[network];
+        if (!existingNetworkAssets) {
+          existingAccountGroupAssets[network] = chainAssets;
+        } else {
+          existingAccountGroupAssets[network] = [
+            ...existingNetworkAssets,
+            ...chainAssets,
+          ];
+        }
```
Prithpal-Sooriya: I think we can use some shorthand to clean this up.

```suggestion
        existingAccountGroupAssets[network] ??= [];
        existingAccountGroupAssets[network].push(...chainAssets);
```

packages/assets-controllers/src/selectors/stringify-balance.ts
Prithpal-Sooriya: fantastic, good idea to add a comment to the lib.

packages/assets-controllers/src/selectors/token-selectors.ts
Prithpal-Sooriya: Yeah sucks. Ideally the popular network logic all exists in CORE, sucks that this is in each individual platform....

Prithpal-Sooriya: We could start migrating by moving a small hard-coded list of EVM native tokens to CORE (with a chainID lookup)?

Prithpal-Sooriya: However images def cannot be handled in core (extension and mobile handle image file paths completely differently)... unless we make them remote URLs 🤪

packages/assets-controllers/src/selectors/token-selectors.ts
bergarces: On one side, it is coming from the controller state, and therefore it should be ok.

On the other, we all know that guarantees nothing. So I'd prefer to try/catch and skip rather than having the whole extension going into a meltdown.

Ideally we would want to report the error, but right now, from core, it does not seem to be an option.

When I was exploring how to make the selectors more memoization-friendly, I came across the option of wrapping all selectors in a factory function in which the client passes the method to extract the state. If we end up using something like that, we can inject a method to report errors as well.

Prithpal-Sooriya: Yep valid concerns!

I see some places in CORE use `console.error`, however I'm unsure if these are effective on both platforms (IIRC mobile uses a custom logging system).

Good callout though. We can do this as a fast follow PR.

packages/assets-controllers/src/selectors/token-selectors.ts
bergarces: I'm going to do the same as extension and just use ticket from network state for the symbol.

If the symbol is `'ETH'`, the name becomes `'Ethereum'`, otherwise the name is the same as the symbol.

If there's no network configuration, the symbol defaults to `'NATIVE'` in both extension and mobile, so I did the same here.

This is also a good idea for another reason, if the native token doesn't match the network configuration ticket, the client shows a scam warning, so better keep it like this.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -0,0 +1,491 @@
...
+          const rawBalance =
+            tokenBalances[accountAddress]?.[chainId]?.[tokenAddress];
+
+          if (!rawBalance) {
+            continue;
+          }
+
+          groupAssets[accountGroupId] ??= {};
+          groupAssets[accountGroupId][chainId] ??= [];
+          const groupChainAssets = groupAssets[accountGroupId][chainId];
```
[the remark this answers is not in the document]
bergarces: This should not be able to happen at all.

packages/assets-controllers/src/selectors/token-selectors.ts
Prithpal-Sooriya: Yep sounds good! 👍🏾

PR #6232, opened 2025-08-04
feat: use aggregate3 or accountAPI to fetch balances

salimtb:
You're absolutely right! Let me update that section to be more precise about the fallback strategy:

### Current State and Problem
Previously, the TokenBalancesController fetched token balances using individual RPC calls for each token-account combination. This approach was inefficient and slow, especially for users with multiple accounts and many tokens across different networks. Each balance check required a separate `balanceOf` call, leading to:

- High number of RPC requests
- Slower balance updates
- Potential rate limiting issues
- Poor user experience during balance refreshes

### Solution
This PR implements a **two-tier balance fetching strategy** with intelligent fallback:

1. **Primary**: **Accounts API** - Uses MetaMask's external API service for supported networks (faster, more efficient)
2. **Fallback**: **RPC with Aggregate3** - Uses Multicall3's `aggregate3` function for efficient batch RPC calls when:
   - The chain is not supported by the Accounts API
   - The Accounts API service is unavailable or fails
   - Any other API-related errors occur

#### Smart Fallback Architecture
- `TokenBalancesController` first attempts to fetch balances via the **Accounts API** for supported networks
- For chains **not supported by the API** or when the **API fails**, it automatically falls back to **RPC calls using `aggregate3`**
- Successfully processed chains are removed from subsequent fetchers, ensuring no duplicate work
- Each fetcher only processes chains it can handle

#### Accounts API Integration
- Leverages MetaMask's backend infrastructure for faster balance retrieval
- Supports a curated list of networks (`SUPPORTED_NETWORKS_ACCOUNTS_API_V4`)
- Handles multi-account and multi-chain requests efficiently

#### Aggregate3 Fallback Implementation
When the Accounts API can't handle certain networks, the system uses Multicall3's `aggregate3` function to:
- Batch multiple `balanceOf` calls into a single RPC request
- Handle both ERC20 token balances and native token balances
- Support staking contract balance queries
- Process results efficiently with individual failure handling

#### Performance Optimizations
- **API-first strategy**: Fastest possible retrieval for supported networks
- **Efficient RPC fallback**: Up to 300 calls per batch to avoid gas/size limits when API isn't available
- **Serial batch execution**: Processes batches sequentially to prevent overwhelming the RPC
- **Smart chain routing**: Each chain uses the most appropriate fetching method
- **Selective updates**: Only updates state when balance changes are detected

### Technical Details

The fallback mechanism ensures comprehensive network coverage:
- **Supported networks** → Accounts API (fastest)
- **Unsupported networks or API failures** → RPC with `aggregate3` (still much faster than individual calls)

The `aggregate3` function provides significant improvements over individual RPC calls:
- Each call can fail individually without affecting others (`allowFailure: true`)
- Supports different call types (ERC20, native, staking) in the same batch
- Automatically falls back to individual calls if multicall fails entirely

### Impact
- **Optimal performance**: Uses fastest available method for each network
- **Complete network coverage**: No network is left without efficient balance fetching
- **Reduced RPC calls**: For non-API networks, reduces from N individual calls to 1 call per batch of up to 300 operations
- **Better reliability**: Multiple fetching strategies ensure balance data is always available
- **Graceful degradation**: Seamless fallback when services are unavailable

## References

Addresses performance issues with token balance fetching by implementing Accounts API integration with aggregate3 RPC fallback for comprehensive and efficient balance retrieval across all networks.

## Checklist

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -18,6 +18,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump `@metamask/account-tree-controller` from `^0.7.0` to `^0.9.0` ([#6310](https://github.com/MetaMask/core/pull/6310))
+- **BREAKING**: Improved `TokenBalancesController` performance with two-tier balance fetching strategy ([#6232](https://github.com/MetaMask/core/pull/6232))
+  - Implements Accounts API as primary fetching method for supported networks (faster, more efficient)
+  - Falls back to RPC calls using Multicall3's `aggregate3` for unsupported networks or API failures
+  - Significantly reduces RPC calls from N individual requests to batched calls of up to 300 operations
+  - Provides comprehensive network coverage with graceful degradation when services are unavailable
 - Bump `@metamask/base-controller` from `^8.0.1` to `^8.1.0` ([#6284](https://github.com/MetaMask/core/pull/6284))
 - Bump `@metamask/controller-utils` from `^11.11.0` to `^11.12.0` ([#6303](https://github.com/MetaMask/core/pull/6303))
 
@@ -29,15 +34,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- **BREAKING**: Improved `TokenBalancesController` performance with two-tier balance fetching strategy ([#6232](https://github.com/MetaMask/core/pull/6232))
-  - Implements Accounts API as primary fetching method for supported networks (faster, more efficient)
-  - Falls back to RPC calls using Multicall3's `aggregate3` for unsupported networks or API failures
-  - Significantly reduces RPC calls from N individual requests to batched calls of up to 300 operations
-  - Provides comprehensive network coverage with graceful degradation when services are unavailable
+- Bump `@metamask/keyring-api` from `^19.0.0` to `^20.0.0` ([#6248](https://github.com/MetaMask/core/pull/6248))
+
+### Fixed
+
+- Correct the polling rate for the DeFiPositionsController from 1 minute to 10 minutes. ([#6242](https://github.com/MetaMask/core/pull/6242))
+- Fix `AccountTrackerController` to force block number update to avoid stale cached native balances ([#6250](https://github.com/MetaMask/core/pull/6250))
+
+## [73.0.2]
 
 ### Fixed
 
 - Use a narrow selector when listening to `CurrencyRateController:stateChange` ([#6217](https://github.com/MetaMask/core/pull/6217))
+- Fixed an issue where attempting to fetch asset conversions for accounts without assets would crash the snap ([#6207](https://github.com/MetaMask/core/pull/6207))
 
 ## [73.0.1]
 
@@ -1836,7 +1845,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use Ethers for AssetsContractController ([#845](https://github.com/MetaMask/core/pull/845))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.0.1...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.1.0...HEAD
+[73.1.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.0.2...@metamask/assets-controllers@73.1.0
+[73.0.2]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.0.1...@metamask/assets-controllers@73.0.2
 [73.0.1]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.0.0...@metamask/assets-controllers@73.0.1
 [73.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@72.0.0...@metamask/assets-controllers@73.0.0
 [72.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@71.0.0...@metamask/assets-controllers@72.0.0


packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -326,13 +340,20 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
...
   }: AccountTrackerPollingInput): Promise<void> {
-    // TODO: Either fix this lint violation or explain why it's necessary to ignore.
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.refresh(networkClientIds);
+    if (includeStakedAssets) {
+      await this.refreshStakedBalances(networkClientIds);
+    }
+    if (includeNativeAssets) {
+      await this.refreshNativeBalances(networkClientIds);
+    }
```
Prithpal-Sooriya: Nice separation here. Only caveat is that both of these have independent state updates. Meaning we could end up with multiple state updates (if we need to update staked assets then native assets).

This is a minor nit, we can clean up any additional state updates in a future design.

packages/assets-controllers/src/AccountTrackerController.ts
Prithpal-Sooriya, line 1: (discuss) why are we including changes in the AccountTrackerController? I thought the new design would have everything in the TokenBalancesController?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -26,79 +20,76 @@ import { StaticIntervalPollingController } from '@metamask/polling-controller';
...
 import type {
   TokensControllerGetStateAction,
   TokensControllerState,
   TokensControllerStateChangeEvent,
 } from './TokensController';
+// endregion
 
-const DEFAULT_INTERVAL = 180000;
+// ────────────────────────────────────────────────────────────────────────────
+// region: Type aliases & constants
```
Prithpal-Sooriya: Do we need these? TBH I'm not against it if it helps segment and improves readability.

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -26,79 +20,76 @@ import { StaticIntervalPollingController } from '@metamask/polling-controller';
...
   TokenBalancesControllerState
 >;
 
 export type TokenBalancesControllerActions =
   TokenBalancesControllerGetStateAction;
 
+export type TokenBalancesControllerStateChangeEvent =
+  ControllerStateChangeEvent<typeof CONTROLLER, TokenBalancesControllerState>;
+
+export type NativeBalanceEvent = {
```
Prithpal-Sooriya, line 81: nit - can this be renamed to `TokenBalancesControllerNativeBalanceEvent` for consistency?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
-    // Refresh when flipped off -> on
-    const refresh = queryMultipleAccounts && !this.#queryMultipleAccounts;
-    this.#queryMultipleAccounts = queryMultipleAccounts;
-
-    if (refresh) {
-      this.updateBalances().catch(console.error);
-    }
-  };
+  // ───────── private helpers ─────────
+  get #chainIdsWithTokens(): ChainIdHex[] {
```
Prithpal-Sooriya: nit - hmm I think I prefer a method over a `get`, esp since this returns a new array on call.

E.g.
```ts
// Getter
const targetChains = chainIds ?? this.#chainIdsWithTokens;

// Method
const targetChains = chainIds ?? this.#chainIdsWithTokens();
```

Prithpal-Sooriya: Meh, its a nit, can ignore.

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
-    );
-  }
+    // Try each fetcher in order, removing successfully processed chains
+    for (const fetcher of this.#balanceFetchers) {
+      const supportedChains = remainingChains.filter((c) =>
+        fetcher.supports(c),
+      );
+      if (!supportedChains.length) {
+        continue;
+      }
```
Prithpal-Sooriya, line 266: Clean!

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
-      for (const currentChain of Object.keys(allChains)) {
-        if (chainIds?.length && !chainIdsSet.has(currentChain as Hex)) {
-          continue;
+      try {
+        const balances = await fetcher.fetch({
+          chainIds: supportedChains,
+          queryAllAccounts: this.#queryAllAccounts,
+          selectedAccount: selected as ChecksumAddress,
+          allAccounts,
+        });
```
Prithpal-Sooriya: Clean!!

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
-    chainId,
-    pairs,
-  }: {
-    chainId: Hex;
-    pairs: { accountAddress: Hex; tokenAddress: Hex }[];
-  }): Promise<MulticallResult[]> {
-    if (!pairs.length) {
-      return [];
+    if (!isEqual(prev, next)) {
+      this.update(() => next);
```
Prithpal-Sooriya, line 352: Hmm, do other controllers follow this pattern.

I think we can just abstract and call this inside `this.update` - `this.update` uses immer and some shallow equality check internally.

Prithpal-Sooriya:
```ts
this.update((state) => {
  // mutate state if new vals
  initialiseEmptyTokens(state);
  
  // mutate state for new balances
  updateBalances(state)
  
  return state; // internally this uses immer to see if new properties were added
})
```

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
+          this.messagingSystem.call(
+            'AccountTrackerController:updateNativeToken',
+            r.account,
+            r.chainId,
+            r.value?.toString() ?? '0',
+          );
+        }
+      });
+    } else {
+      this.#log('No balance changes detected');
```
Prithpal-Sooriya: Nice, we can attach this to debug logs (as long as we aren't logging in prod 😄)

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
+
+      aggregated.forEach((r) => {
+        if (r.success && r.token === ZERO_ADDRESS) {
+          this.messagingSystem.call(
+            'AccountTrackerController:updateNativeToken',
+            r.account,
+            r.chainId,
+            r.value?.toString() ?? '0',
+          );
+        }
```
Prithpal-Sooriya: Hmm since we are doing multichain architecture changes + performance, do you think we can modify this event to take in a list of chains and values.

This way we only emit 1 event to process all updated native tokens.

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
-          ] = toHex(value as BN);
+    // Handle both cleanup and updates in a single state update
+    this.update((s) => {
+      for (const chainId of relevantChainIds) {
+        const id = chainId as ChainIdHex;
+        const hasTokensNow =
+          (state.allTokens[id] &&
+            Object.keys(state.allTokens[id]).length > 0) ||
+          (state.allDetectedTokens[id] &&
+            Object.keys(state.allDetectedTokens[id]).length > 0);
```
Prithpal-Sooriya, line 463:
```suggestion
        const hasTokensNow =
          (Object.keys(state.allTokens[id] ?? {}).length > 0) ||
          (Object.keys(state.allDetectedTokens[id] ?? {}).length > 0;
```

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
+        const hadTokensBefore =
+          (this.#allTokens[id] &&
+            Object.keys(this.#allTokens[id]).length > 0) ||
+          (this.#detectedTokens[id] &&
+            Object.keys(this.#detectedTokens[id]).length > 0);
```
Prithpal-Sooriya, line 468:
```suggestion
        const hadTokensBefore =
          (Object.keys(this.#allTokens[id] ?? {}).length > 0) ||
          (Object.keys(this.#detectedTokens[id] ?? {}).length > 0;
```

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -107,526 +98,405 @@ export type AllowedEvents =
...
+
+        if (
+          !isEqual(state.allTokens[id], this.#allTokens[id]) ||
+          !isEqual(state.allDetectedTokens[id], this.#detectedTokens[id])
```
Prithpal-Sooriya, line 472: Wondering if we need deep-eq on state change? Probably yes...

packages/assets-controllers/src/constants.ts
```diff
@@ -3,3 +3,15 @@ export enum Source {
   Dapp = 'dapp',
   Detected = 'detected',
 }
+
+export const SUPPORTED_NETWORKS_ACCOUNTS_API_V4 = [
```
Prithpal-Sooriya, line 8: Nice use of this supported network.

I thought the accounts API also returns an endpoint to get dynamic list of supported networks?

packages/assets-controllers/src/multicall.ts
```diff
@@ -193,30 +195,11 @@ const MULTICALL_CONTRACT_BY_CHAINID = {
   '0xa1337': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0x1f2b': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0xf63': '0xcA11bde05977b3631167028862bE2a173976CA11',
-  '0x144': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
+  '0x144': '0xcA11bde05977b3631167028862bE2a173976CA11',
   '0x118': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
   '0x12c': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
   '0x18995f': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
   '0x2b74': '0xF9cda624FBC7e059355ce98a31693d299FACd963',
-  '0xfc': '0xcA11bde05977b3631167028862bE2a173976CA11',
```
Prithpal-Sooriya: was this meant to be nuked?

packages/assets-controllers/src/multicall.ts
```diff
@@ -416,3 +479,539 @@ export const multicallOrFallback = async (
...
+ * Processes and decodes balance results from aggregate3 calls
+ *
+ * @param results - Array of results from aggregate3 calls
+ * @param callMapping - Array mapping call indices to token and user addresses
+ * @param chainId - The hexadecimal chain id
+ * @param provider - An ethers rpc provider
+ * @param includeStaked - Whether to include staked balances
+ * @returns Map of token address to map of user address to balance
+ */
+const processBalanceResults = (
```
Prithpal-Sooriya: Guessing these changes are the same as this PR:
https://github.com/MetaMask/core/pull/6212

packages/assets-controllers/src/rpc-service/rpc-balance-fetcher.ts
```diff
@@ -0,0 +1,185 @@
...
+
+export type ProcessedBalance = {
+  success: boolean;
+  value?: BN;
+  account: ChecksumAddress;
+  token: ChecksumAddress;
+  chainId: ChainIdHex;
+};
+
+export type BalanceFetcher = {
```
Prithpal-Sooriya, line 23: Is this meant to be a shared interface between different BalanceFetchers?

packages/assets-controllers/src/selectors.ts
```diff
@@ -0,0 +1,100 @@
...
+      for (const [tokenAddress, balance] of Object.entries(chainBalances)) {
+        // tokenRates is organized by chainId -> tokenAddress -> MarketDataDetails
+        // We need to find the market data for this token across all chains
+        let marketData;
+        for (const chainData of Object.values(tokenRates)) {
+          if (chainData[tokenAddress as Hex]) {
+            marketData = chainData[tokenAddress as Hex];
+            break;
+          }
+        }
```
Prithpal-Sooriya:
```suggestion
        const marketData = Object.values(tokenRates).find(r => r[tokenAddress]);
```

packages/assets-controllers/src/selectors.ts
```diff
@@ -0,0 +1,100 @@
...
+                  (token: { address: string; decimals: number }) =>
+                    token.address.toLowerCase() === tokenAddress.toLowerCase(),
+                );
+                if (foundToken && foundToken.decimals !== undefined) {
+                  tokenDecimals = foundToken.decimals;
+                  break;
+                }
+              }
+            }
+          }
```
Prithpal-Sooriya: Makes sense, just a tad gnarly to navigate through. Wondering if it is worth converting this into a indexible map object selector.

packages/assets-controllers/src/selectors.ts
```diff
@@ -0,0 +1,100 @@
...
+          // If token has different decimals, convert to 18 decimals
+          if (tokenDecimals !== 18) {
+            const decimalDifference = 18 - tokenDecimals;
+            normalizedBalance *= Math.pow(10, decimalDifference);
+          }
+
+          // Now divide by 10^18 since everything is normalized to 18 decimals
+          const balanceInTokens = normalizedBalance / Math.pow(10, 18);
+          const tokenValue = price * balanceInTokens * rateToUse;
+          const roundedTokenValue = Math.round(tokenValue * 100) / 100; // Round to 2 decimal places
```
Prithpal-Sooriya: I might need you to break down the formula again.

- `balanceInTokens` - e.g. 1MNT
- `rateToUse` - ETH conversion rate - e.g. 0.0032...ETH = 1MNT.
- `price` - USD/Currency Price - e.g. $1234 = 1ETH

So combining all these gives you the price in USD?

Prithpal-Sooriya: Does this cover other native tokens?

Prithpal-Sooriya (approved):

packages/assets-controllers/src/TokenBalancesController.ts
salimtb: we're using deep equal on this case yes

packages/assets-controllers/src/constants.ts
salimtb: we don't have this endpoint for v4 api , i just added a todo here

packages/assets-controllers/src/TokenBalancesController.ts
salimtb: yes i like this idea , i'll do this instead

amitabh94 (dismissed):

amitabh94 (approved):

PR #6233, opened 2025-08-04
Feat/shield subscription controller

tuna1207:
## Explanation

Create Shield Subscription controller
- initialize controller with architecture doc
- handle basic CRUD with subscription service

## References

- [Jira](https://consensyssoftware.atlassian.net/browse/SUBS-25?atlOrigin=eyJpIjoiMDkyMjE1ZGVhN2NhNDdlMGE3MzY2NmM5NjNlNGY1ZTYiLCJwIjoiaiJ9)

## Checklist


packages/subscription-controller/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+
+## [Unreleased]
+
+### Added
+
+- Initial release of the subscription controller ([#6233](https://github.com/MetaMask/core/pull/6233))
+  - This controller allows MM extension and mobile users to manage user subscription, create new subscription, check subscription status.
+  - The current controller contains the following methods:
+    - `getSubscription`: Retrieve current user subscription info if exist.
+    - `cancelSubscription`: Cancel user active subscription.
```
matthiasgeihs: No need to describe controller functionality here.
```suggestion
```

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,724 @@
...
+      // The controller should have created a SubscriptionService internally
+      expect(controller.state).toStrictEqual(
+        getDefaultSubscriptionControllerState(),
+      );
+    });
+  });
+
+  describe('getSubscription', () => {
+    it('should fetch and store subscription successfully', async () => {
+      await withController(async ({ controller, mockService }) => {
```
matthiasgeihs, line 212: Personally, I don't like the way these tests are written.

Using `withController` creates unnecessary nesting and makes the code hard to read:

```ts
      await withController(
        {
          // params
        },
        async ({ controller }) => {
          // test code
        },
      );
```

Instead we should use the following style:
```ts
const { controller } = await setup({params});
// test code
```

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+    },
+    pendingPaymentTransactions: {
+      persist: true,
+      anonymous: true,
+    },
+    authTokenRef: {
+      persist: true,
+      anonymous: true,
+    },
+  };
```
matthiasgeihs: These are not `anonymous`. Set `anonymous` to false.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+   * @param options.state - Initial state to set on this controller.
+   * @param options.config - Configuration for this controller.
+   * @param options.subscriptionService - The subscription service for communicating with subscription server.
+   */
+  constructor({
+    messenger,
+    state,
+    config,
+    subscriptionService,
+  }: SubscriptionControllerOptions) {
```
matthiasgeihs: I don't understand why you are wrapping some of the config in a separate type. It's very unconventional. (Moreover, this type currently has only one entry.) Just use a flat constructor options type instead and put the items as separate properties on the class.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+export class SubscriptionController extends BaseController<
+  typeof controllerName,
+  SubscriptionControllerState,
+  SubscriptionControllerMessenger
+> {
+  readonly #subscriptionService: ISubscriptionService;
+
+  readonly #config: SubscriptionControllerConfig = {
+    env: Env.PRD,
+  };
```
matthiasgeihs: Remove the default here and construct the object in the constructor. Then it's easier to understand what's going on.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+        ...getDefaultSubscriptionControllerState(),
+        ...state,
+      },
+      messenger,
+    });
+
+    this.#config = {
+      ...this.#config,
+      ...config,
+    };
```
matthiasgeihs: Huh? Why this? We are in the constructor here. Just construct the object the way you want it here in one place. No need to use a default above.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+      subscriptionService ??
+      new SubscriptionService({
+        env: this.#config.env,
+        auth: {
+          getAccessToken: () =>
+            this.messagingSystem.call(
+              'AuthenticationController:getBearerToken',
+            ),
+        },
+      });
```
matthiasgeihs: We could make the `subscriptionService` parameter mandatory. Then we can get rid on the dependency on the `env` variable here in the class. The only thing we currently use the variable for is in the subscription service.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.#assertIsUserSubscribed();
+
+    await this.#subscriptionService.cancelSubscription({ subscriptionId });
+  }
+
+  // #assertIsUserNotSubscribed() {
+  //   if (this.state.subscription) {
+  //     throw new Error(SubscriptionControllerErrorMessage.UserAlreadySubscribed);
+  //   }
+  // }
```
matthiasgeihs: Remove if not needed

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  async getSubscription() {
+    const subscription = await this.#subscriptionService.getSubscription();
+
+    this.update((state) => {
+      state.subscription = subscription ?? undefined;
+    });
+
+    return subscription;
+  }
```
matthiasgeihs: We should support multiple subscriptions.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+  //   if (this.state.subscription) {
+  //     throw new Error(SubscriptionControllerErrorMessage.UserAlreadySubscribed);
+  //   }
+  // }
+
+  #assertIsUserSubscribed() {
+    if (!this.state.subscription) {
+      throw new Error(SubscriptionControllerErrorMessage.UserNotSubscribed);
+    }
+  }
```
matthiasgeihs: This should get the product as a parameter.

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,481 @@
...
+        'https://subscription-service.dev-api.cx.metamask.io/api/v1/subscription/123/cancel',
+      );
+    });
+  });
+
+  describe('getSubscription', () => {
+    it('should fetch subscription successfully', async () => {
+      const config = createMockConfig();
+      const service = new SubscriptionService(config);
+      const testUrl = getTestUrl(Env.DEV);
```
matthiasgeihs: Please avoid code duplication. Wrap this in a setup function and use it consistently across this file. It will make it much easier to apply changes.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,133 @@
...
+import { SubscriptionService } from './SubscriptionService';
+import type {
+  ISubscriptionService,
+  SubscriptionControllerConfig,
+  SubscriptionControllerMessenger,
+  SubscriptionControllerOptions,
+  SubscriptionControllerState,
+} from './types';
+
+// const log = createModuleLogger(projectLogger, controllerName);
```
matthiasgeihs: remove commented out code!

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly #config: SubscriptionServiceConfig;
+
+  readonly #env: Env;
+
+  constructor(config: SubscriptionServiceConfig) {
+    this.#env = config.env;
+    this.#config = config;
+  }
+
+  async getSubscription(): Promise<Subscription | null> {
```
matthiasgeihs:
```suggestion
  async getSubscriptions(): Promise<Subscription[]> {
```

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+      const errorMessage =
+        e instanceof Error ? e.message : JSON.stringify(e ?? '');
+
+      throw new SubscriptionServiceError(
+        `failed to get subscription. ${errorMessage}`,
+      );
+    }
+  }
+
+  async cancelSubscription(params: { subscriptionId: string }): Promise<void> {
```
matthiasgeihs, line 41: nit
```suggestion
  async cancelSubscription(subscriptionId: string): Promise<void> {
```

packages/subscription-controller/src/constants.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      expect(controllerName).toBe('SubscriptionController');
+    });
+  });
+
+  describe('Env enum', () => {
+    it('should have correct values', () => {
+      expect(Env.DEV).toBe('dev');
+      expect(Env.UAT).toBe('uat');
+      expect(Env.PRD).toBe('prd');
+    });
```
matthiasgeihs: Really? We need a test for that? If you are worried about that, I suggest you use the following syntax instead:
```ts
export const environments = ['dev', 'uat', 'prd'] as const;
export type Env = (typeof environments)[number];
```

matthiasgeihs (changes requested):
Please be mindful that more tests is not always better.
Having a lot of tests also means that applying changes to the functionality will require a lot of changes to the tests, which costs time.
When writing tests, make sure that you avoid code duplication and try to use a single setup function across a suite of tests, so that it's easy to change things in the setup without having to change it across all tests individually.

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,112 @@
...
+      chainId: string;
+      hash?: string;
+    };
+  };
+};
+
+export type SubscriptionControllerState = Partial<AuthUserData> &
+  Partial<PendingPaymentTransactionData> & {
+    subscription?: Subscription;
+  };
```
matthiasgeihs: Instead of including partial types, the fields should be explicitly listed here. Otherwise it's unclear which fields are supposed to be included, which can be a problem. For example, when we define metadata, we must know for which fields the state metadata is to be defined.

packages/subscription-controller/src/SubscriptionController.ts
MajorLift: For subclasses of `BaseController`, the recommended/enforced pattern is to prefer options bags over positional arguments.

> https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#use-single-options-bag-for-constructor-arguments

packages/subscription-controller/src/SubscriptionController.ts
MajorLift: This seems like an architectural change that would require a fair bit of refactoring to implement. Is the team aligned on this controller managing state for multiple subscriptions instead of one?

packages/subscription-controller/src/SubscriptionService.ts
MajorLift: Point of consideration for extensibility: Applying this suggestion will make any future change to the method signature a breaking change. An options bag allows more latitude for arguments to be added or modified as a non-breaking change.

packages/subscription-controller/src/constants.test.ts
MajorLift: The surface area covered by compile-time type checks does not completely overlap with that of runtime unit/regression tests.

In any case, there's no need to newly derive this union type since we already have access to it via the `Env` enum.

```ts
import { Env } from './constants';

export type Environments = `${Env}` // "dev" | "uat" | "prd"
```

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,112 @@
...
+
+// Actions
+export type SubscriptionControllerGetStateAction = ControllerGetStateAction<
+  typeof controllerName,
+  SubscriptionControllerState
+>;
+export type SubscriptionControllerActions =
+  SubscriptionControllerGetStateAction;
```
MajorLift: Should we also define `SubscriptionController{Get,Cancel}SubscriptionAction`, and enable consumers to access the corresponding class methods via messenger actions?

packages/subscription-controller/src/SubscriptionService.ts
MajorLift: Same as https://github.com/MetaMask/core/pull/6233#discussion_r2291177089

packages/subscription-controller/src/SubscriptionService.test.ts
MajorLift: Agreed, but avoid using `beforeEach`: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#avoid-the-use-of-beforeeach

packages/subscription-controller/src/SubscriptionController.test.ts
MajorLift: `withController` is written as a wrapper function because it's used for teardown as well as setup.

More info and example code available by clicking "Read more" at the bottom of this docs entry: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#avoid-the-use-of-beforeeach

packages/subscription-controller/src/SubscriptionController.ts
MajorLift: Wouldn't keeping this default instantiation logic for `SubscriptionService`  reduce code duplication in downstream contexts where `SubscriptionController` is instantiated?

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: I was referring to `config`, which is an option bag inside an option bag.

packages/subscription-controller/src/SubscriptionController.test.ts
matthiasgeihs: OK, wasn't aware that `withController` is the suggested style. I've seen other controllers follow a different pattern (e.g., [TransactionController.test.ts](https://github.com/MetaMask/core/blob/main/packages/transaction-controller/src/TransactionController.test.ts)) and I definitely think it's more readable. Also note that `withController` implemented here does not have any teardown code. Moreover, using the `setup` pattern would be perfectly consistent with the "avoid the use of beforeEach" rule.

In any case, if the `withController` pattern is a common convention used here, I guess I can't argue against that.

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: I think one can argue both ways here. I'd tend to make it mandatory, to fully remove the dependency on the instantiation. Anyways, either way has pros and cons.

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: The goal was to design the controller in a way that it can be used for other products in the future as well. My interpretation of that is that we should be able to support multiple subscriptions. But, good point, we should ensure we are aligned here.

My worry is that if we don't do the change now, it will be a lot more work down the road.

packages/subscription-controller/src/SubscriptionService.ts
matthiasgeihs: Good point. I guess that's more my personal preference to not use an option bag for a single parameter.

Besides, I was missing some consistency in the naming. In a follow up PR, for a different request, the `params` are called `request` [#6300](https://github.com/MetaMask/core/pull/6300/files#diff-f06809e9719ecb0a54f60af8215079c06f4becdfca03a611167b269b87ff0891). I think it would be nice if we use a consistent naming here. (Unless there is a particular reason not to.)

packages/subscription-controller/src/constants.test.ts
matthiasgeihs: Sorry, I was unaware that this was an `enum` type. (I somehow thought we shouldn't use `enum` types in `core`, but apparently this is wrong. I believe we followed this convention in TOPRF Client SDK.)

In any case, is this test really necessary?

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type ErrorMessage = {
+  message: string;
+  error: string;
+};
+
+export const SUBSCRIPTION_URL = (env: Env, path: string) =>
+  `${getEnvUrls(env).subscriptionApiUrl}/api/v1/${path}`;
+
+export class SubscriptionService implements ISubscriptionService {
```
matthiasgeihs, line 23: There is a bunch of code duplication in this class.

The request and error handling logic for all requests is very similar.
You should reduce the lines of code significantly by creating and using a dedicated function that holds the core request logic:
```ts
async #makeRequest<Result>(
  path: string,
  method: 'GET' | 'POST' = 'GET',
): Promise<Result> {
  try {
    const headers = await this.#getAuthorizationHeader();
    const url = new URL(SUBSCRIPTION_URL(this.#env, path));

    const response = await fetch(url.toString(), {
      method,
      headers: {
        'Content-Type': 'application/json',
        ...headers,
      },
    });

    const responseBody = (await response.json()) as ErrorMessage;
    if (!response.ok) {
      throw new Error(
        `HTTP error message: ${responseBody.message}, error: ${responseBody.error}`,
      );
    }

    return responseBody as Result;
  } catch (e) {
    const errorMessage =
      e instanceof Error ? e.message : JSON.stringify(e ?? '');

    throw new SubscriptionServiceError(
      `failed to make request. ${errorMessage}`,
    );
  }
}
```

Then your individual request functions become almost 1-liners:
```ts
  async cancelSubscription(params: { subscriptionId: string }): Promise<void> {
    const path = `subscription/${params.subscriptionId}/cancel`;
    await this.#makeRequest(path, 'POST');
  }
```

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  async cancelSubscription(params: { subscriptionId: string }): Promise<void> {
+    const path = `subscription/${params.subscriptionId}/cancel`;
+    try {
+      const headers = await this.#getAuthorizationHeader();
+      const url = new URL(SUBSCRIPTION_URL(this.#env, path));
+      const response = await fetch(url.toString(), {
+        method: 'POST',
```
matthiasgeihs: Why is this a POST request?

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+        throw new Error(
+          `HTTP error message: ${responseBody.message}, error: ${responseBody.error}`,
+        );
+      }
+
+      const res = (await response.json()) as { data: Subscription | null };
+      return res.data;
+    } catch (e) {
+      const errorMessage =
+        e instanceof Error ? e.message : JSON.stringify(e ?? '');
```
matthiasgeihs:
```suggestion
        e instanceof Error ? e.message : JSON.stringify(e ?? 'unknown error');
```

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      if (!response.ok) {
+        const responseBody = (await response.json()) as ErrorMessage;
+        throw new Error(
+          `HTTP error message: ${responseBody.message}, error: ${responseBody.error}`,
+        );
+      }
+    } catch (e) {
+      const errorMessage =
+        e instanceof Error ? e.message : JSON.stringify(e ?? '');
```
matthiasgeihs:
```suggestion
      const errorMessage =
        e instanceof Error ? e.message : JSON.stringify(e ?? 'unknown error');
```

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,481 @@
...
+
+    it('should handle null/undefined exceptions', async () => {
+      const config = createMockConfig();
+      const service = new SubscriptionService(config);
+      const testUrl = getTestUrl(Env.DEV);
+
+      nock(testUrl).get('/api/v1/subscription').replyWithError('Network error');
+
+      await expect(service.getSubscription()).rejects.toThrow(/Network error/u);
+    });
```
matthiasgeihs: If you create a dedicated `makeRequest` function, then you don't have to test these general cases for each individual request type separately.

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs, line 1: It was previously suggested to me by @mcmire (see https://github.com/MetaMask/core/pull/6137#discussion_r2279103688) that we should inject `fetch`.

packages/subscription-controller/src/SubscriptionService.test.ts
matthiasgeihs: Aren't these tests duplicates of each other? (Same comments applies to other instances of these tests for the other request types.)

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: Oh i was following this controller (packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts) 's config pattern, i think it's to separate the config for environment, but we can set it as separate option instead of inside another config

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: agree, updated 👍

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: updated to have direct env option instead of inside config object

packages/subscription-controller/CHANGELOG.md
tuna1207: removed

packages/subscription-controller/src/SubscriptionController.test.ts
tuna1207: Also agree that the `setup` pattern make reading easier, I think we will keep `withController` pattern for now to follow the convention

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: removed

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: removed

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: updated naming to `request`

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: It's following server API definition, it's also a cancel request, not delete, so i think POST is more suitable

packages/subscription-controller/src/constants.test.ts
MajorLift: TypeScript will apparently be [phasing out `enum`s](https://www.totaltypescript.com/erasable-syntax-only), but as of now they're used widely throughout the codebase and we don't have any rules or conventions against them. 

> In any case, is this test really necessary?

I don't really see a scenario where this test would fail, no. At least none that I'd expect a unit test to cover.😅

packages/subscription-controller/src/SubscriptionService.ts
MajorLift: Thanks to @davidmurdoch, I recently learned that `String()` can be better for handling unknown errors than `JSON.stringify()`:

```suggestion
        e instanceof Error ? e.message : String(e);
```

> `JSON.stringify([new Error("string")])` returns `'["","",{}]'` but `String(["","",new Error("string")])` returns `',,Error: string'`

> `String(error)` better handles edge cases:
If error is null or undefined, `String()` returns `"null"` or `"undefined"`.
If error is an object without a proper `toString()`, `String()` still works
`JSON.stringify` doesn't like lots of things, like circular objects (throws), Error (returns nothing), and some other weirdness.

> -- https://github.com/MetaMask/metamask-extension/pull/34305#discussion_r2239290239

packages/subscription-controller/src/SubscriptionService.ts
MajorLift:
```suggestion
      const errorMessage =
        e instanceof Error ? e.message : String(e);
```

See https://github.com/MetaMask/core/pull/6233#discussion_r2297287778

packages/subscription-controller/src/SubscriptionController.ts
MajorLift: Since `SubscriptionService` still relies on `env` as a dependency, and the service is itself a required dependency of `SubscriptionController`, wouldn't `env` always need to be supplied to the controller wherever it is instantiated?

For context, these controllers are only consumed and instantiated in the context of singleton injector classes that are aware of and supply all of their dependencies at runtime (`MetamaskController` in extension, `Engine` in mobile). So generally speaking, considerations for those specific downstream consumers tend to take priority over internal maintainability concerns.

I'm not saying removing `env` in the class is necessarily a bad idea. Just wanted to be sure that the change would represent an improvement that aligns with the intended purpose of the package.

packages/subscription-controller/src/SubscriptionController.ts
MajorLift: Agreed that this is a change that would be better to make sooner rather than later. Not saying that it's a bad idea at all -- just that it should be implemented consistently. :)

packages/subscription-controller/src/types.ts
MajorLift: Rule of thumb: If making the methods private doesn't seem like a good idea, and they interact with controller state, it probably is a good idea to expose them via the messenger.

> https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#treat-state-mutating-methods-as-actions

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: Updated

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: Updated per API change to DELETE

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: Updated

packages/subscription-controller/src/types.ts
tuna1207: Updated

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: Updated

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: Updated

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: Updated to list of subscriptions

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: Updated

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: Agree, updated

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: Updated inject `fetch`

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,756 @@
...
+  const mockGetBearerToken = jest.fn().mockResolvedValue(MOCK_ACCESS_TOKEN);
+
+  jest.spyOn(messenger, 'call').mockImplementation((...args) => {
+    const [actionType] = args as [string, ...unknown[]];
+
+    if (actionType === 'AuthenticationController:getBearerToken') {
+      return mockGetBearerToken();
+    }
+    throw new Error(`MOCK_FAIL - unsupported messenger call: ${actionType}`);
+  });
```
matthiasgeihs: We shouldn't spy on the messenger's call function to mock messenger functionality. The call function is an implementation detail of the `messenger` that may change and then our tests break.

Instead you can register an action handler on the base messenger. This is also much simpler.
```ts
  baseMessenger.registerActionHandler(
    'AuthenticationController:getBearerToken',
    mockGetBearerToken,
  );
```

packages/subscription-controller/src/types.ts
tuna1207: Agreed, exposed the actions

packages/subscription-controller/src/SubscriptionController.test.ts
tuna1207: Agree, updated

packages/subscription-controller/src/constants.test.ts
tuna1207: Agree, removed the test

.github/CODEOWNERS
```diff
@@ -78,6 +78,7 @@
 
 ## Web3Auth Team
 /packages/seedless-onboarding-controller @MetaMask/web3auth
+/packages/subscription-controller        @MetaMask/web3auth
```
matthiasgeihs: nit: order alphabetically

packages/subscription-controller/src/SubscriptionService.test.ts
matthiasgeihs: This has not been addressed as expected.
The idea was to move all setup code into a single function so that you can do
```ts
const { config, service, testUrl } = setup();
```

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,84 @@
...
+      return responseBody as Result;
+    } catch (e) {
+      const errorMessage =
+        e instanceof Error ? e.message : JSON.stringify(e ?? 'unknown error');
+
+      throw new SubscriptionServiceError(
+        `failed to make request. ${errorMessage}`,
+      );
+    }
+  }
```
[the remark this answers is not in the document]
matthiasgeihs: Oh, this is my bad. I've suggested this piece of code. Even though it's technically doing the right thing, it looks odd to first cast the body to `ErrorMessage` and then, if the status is OK, recast it to the result type.

We could instead cast it within the if block:
```ts
      const responseBody = await response.json();
      if (!response.ok) {
        const { message, error } = responseBody as ErrorMessage;
        throw new Error(`HTTP error message: ${message}, error: ${error}`);
      }

      return responseBody as Result;
```

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+
+/**
+ * Jest Mock Utility to generate a mock Subscription Messenger
+ *
+ * @param overrideMessengers - override messengers if need to modify the underlying permissions
+ * @param overrideMessengers.baseMessenger - base messenger to override
+ * @param overrideMessengers.messenger - messenger to override
+ * @returns series of mocks to actions that can be called
+ */
+function mockSubscriptionMessenger(overrideMessengers?: {
```
matthiasgeihs, line 78: It would be nice to move all mock messenger related code (L57-L126) into a separate file under `tests`. Then we can focus on the actual subscription controller tests here in this file.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,242 @@
...
+  };
+
+export class SubscriptionController extends BaseController<
+  typeof controllerName,
+  SubscriptionControllerState,
+  SubscriptionControllerMessenger
+> {
+  readonly #subscriptionService: ISubscriptionService;
+
+  readonly #env: Env;
```
matthiasgeihs: We are not using this `#env` variable anywhere and should remove it.
If it becomes relevant, we can add it back any time.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  /**
+   * Subscription service to use for the subscription controller.
+   */
+  subscriptionService?: ISubscriptionService;
+
+  /**
+   * Fetch function to use for the subscription controller.
+   */
+  fetchFn: typeof globalThis.fetch;
```
matthiasgeihs: This is another variable (in addition to `env`) that is only relevant for the SubscriptionService.
The SubscriptionService dependency is hence bleeding into this controller. Again, I think it would be cleaner to remove `fetchFn` and `env` here and instead make `subscriptionService` mandatory.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+      expect(mockFetch).toHaveBeenCalledWith(
+        expect.stringContaining('subscription-service.api.cx.metamask.io'),
+        expect.objectContaining({
+          headers: expect.objectContaining({
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${MOCK_ACCESS_TOKEN}`,
+          }),
+        }),
+      );
+    });
```
matthiasgeihs: This is testing functionality of the `SubscriptionService`. For the tests you should not have to rely on the actual SubscriptionController implementation, but use a mock instead.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+        env: Env.PRD,
+        fetchFn: global.fetch,
+      });
+
+      expect(controller).toBeDefined();
+      // The controller should have created a SubscriptionService internally
+      expect(controller.state).toStrictEqual(
+        getDefaultSubscriptionControllerState(),
+      );
+    });
```
matthiasgeihs: This test does not test what it says it tests. You are not actually checking that the `SubscriptionService` has been initialized. I think you can simply remove this test.

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,60 @@
...
+export type PendingPaymentTransaction = {
+  type: 'subscription_approval' | 'subscription_payment';
+  status: 'pending' | 'confirmed' | 'failed';
+  chainId: string;
+  hash?: string;
+};
+
+export type GetSubscriptionsResponse = {
+  customerId: string;
+  subscriptions: Subscription[] | null;
```
matthiasgeihs: I can't see a good reason why this optionally return `null`.
You can simply return an empty array when there is not subscriptions.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+            newSubscription,
+          ]);
+          expect(controller.state.subscriptions[0]?.id).toBe('sub_new');
+        },
+      );
+    });
+  });
+
+  describe('cancelSubscription', () => {
+    it('should cancel subscription successfully when user is subscribed', async () => {
```
matthiasgeihs: nit
```suggestion
    it('should cancel subscription successfully', async () => {
```

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+            }),
+          ).rejects.toThrow(
+            SubscriptionControllerErrorMessage.UserNotSubscribed,
+          );
+
+          // Verify the subscription service was not called
+          expect(mockService.cancelSubscription).not.toHaveBeenCalled();
+        },
+      );
+    });
```
matthiasgeihs: You can merge these two tests.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+          expect(mockService.cancelSubscription).toHaveBeenCalledWith({
+            subscriptionId: 'sub_123456789',
+            type: ProductType.SHIELD,
+          });
+          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
+        },
+      );
+    });
+
+    it('should handle network errors during cancellation', async () => {
```
matthiasgeihs: I think you can remove one of the tests. These are basically duplicates of each other and having two of them doesn't add value in my opinion. Or am I missing something?

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+      });
+
+      expect(controller.state.subscriptions).toStrictEqual([MOCK_SUBSCRIPTION]);
+      expect(controller.state.authTokenRef).toStrictEqual(MOCK_AUTH_TOKEN_REF);
+      expect(controller.state.pendingPaymentTransactions).toStrictEqual(
+        MOCK_PENDING_PAYMENT_TRANSACTIONS,
+      );
+    });
+
+    it('should update state correctly through getSubscription', async () => {
```
matthiasgeihs: I think you are already testing this in the getSubscription test suite. You can remove the test here.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+        trialedProducts: [],
+      });
+
+      await controller.getSubscriptions();
+
+      expect(controller.state.subscriptions).toStrictEqual([newSubscription]);
+      expect(controller.state.subscriptions[0]?.id).toBe('new_sub_id');
+    });
+
+    it('should handle partial state updates through initial state', () => {
```
matthiasgeihs: Since you already have the constructor tests, I think you can move this partial state initialization test there.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+            type: ProductType.SHIELD,
+          });
+          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
+        },
+      );
+    });
+  });
+
+  describe('state management', () => {
+    it('should properly initialize with default state', () => {
```
matthiasgeihs: This seems to fit well into the constructor test suite. Are you maybe already testing the state initialization there?

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+        env: Env.PRD,
+        fetchFn: global.fetch,
+      });
+
+      expect(controller.state).toStrictEqual(
+        getDefaultSubscriptionControllerState(),
+      );
+    });
+
+    it('should merge initial state with default state', () => {
```
matthiasgeihs: This seems to fit well into the constructor test suite. Are you maybe already testing the state initialization there?

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,751 @@
...
+            subscriptionId: 'sub_123456789',
+            type: ProductType.SHIELD,
+          });
+          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
+        },
+      );
+    });
+  });
+
+  describe('state management', () => {
```
matthiasgeihs: This test suite is partially covered by other test suites already. I don't see why you need a separate test suite to check the effect of state updates. I think you can perfectly cover the cases in the existing test suites above, if not already done so.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,242 @@
...
+  ProductType,
+  ISubscriptionService,
+  Subscription,
+  AuthTokenRef,
+  PendingPaymentTransaction,
+} from './types';
+
+export type SubscriptionControllerState = {
+  subscriptions: Subscription[];
+  authTokenRef?: AuthTokenRef;
```
matthiasgeihs: I don't understand why we need this field. It's certainly not used yet. And I'm also not sure if we need it in the future, because the access token is managed by the authentication controller. I'd prefer removing it here. We can always add it back if we need it.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,242 @@
...
+  pendingPaymentTransactions?: {
+    [transactionId: string]: PendingPaymentTransaction;
+  };
```
matthiasgeihs: Similar comment as for `authTokenRef`. We are not using `pendingPaymentTransactions` at the moment and I'd prefer not adding it until we actually need it.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,242 @@
...
+    return subscriptions;
+  }
+
+  async cancelSubscription(request: {
+    subscriptionId: string;
+    type: ProductType;
+  }) {
+    this.#assertIsUserSubscribed({ productType: request.type });
+
+    await this.#subscriptionService.cancelSubscription(request);
```
matthiasgeihs: That's odd. You are accepting and forwarding a product type here but the `cancelSubscription` API actually only requires a `subscriptionId`. I think you don't have to check for the product type and can drop the product type parameter. (And then you can also remove `#assertIsUserSubscribed`.)

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+  });
+
+  afterEach(() => {
+    cleanAll();
+  });
+
+  describe('constructor', () => {
+    it('should create instance with valid config', () => {
+      const config = createMockConfig();
+      const service = new SubscriptionService(config);
```
matthiasgeihs, line 94: The idea was that we have a common setup function so that we can write
```ts
const { config, service } = setup();
```
instead of the two lines here. (An additional benefit is that it's easy to add something to the setup later.)

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+}
+
+describe('SubscriptionService', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    cleanAll();
+  });
```
matthiasgeihs: Not sure why, but @MajorLift mentioned that we should avoid `beforeEach` (and probably also `afterEach`).

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL)
+        .get('/api/v1/subscriptions')
+        .reply(500, MOCK_ERROR_RESPONSE);
+
+      await expect(service.getSubscriptions()).rejects.toThrow(
+        SubscriptionServiceError,
+      );
+    });
```
matthiasgeihs: Looks like these tests are duplicates of each other?

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL)
+        .get('/api/v1/subscriptions')
+        .replyWithError('Network error');
+
+      await expect(service.getSubscriptions()).rejects.toThrow(
+        SubscriptionServiceError,
+      );
+    });
```
matthiasgeihs: another duplicate

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL)
+        .get('/api/v1/subscriptions')
+        .replyWithError('String error');
+
+      await expect(service.getSubscriptions()).rejects.toThrow(
+        SubscriptionServiceError,
+      );
+    });
```
matthiasgeihs: another duplicate

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+        .reply(200, {
+          customerId: 'cus_1',
+          subscriptions: [MOCK_SUBSCRIPTION],
+          trialedProducts: [],
+        });
+
+      await service.getSubscriptions();
+
+      // Verify the correct headers were sent
+      expect(isDone()).toBe(true);
```
matthiasgeihs: This is not verifying the headers!?

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      jest.spyOn(global, 'fetch').mockImplementation().mockRejectedValue(null);
+
+      try {
+        await expect(
+          service.cancelSubscription({ subscriptionId: 'sub_123456789' }),
+        ).rejects.toThrow(SubscriptionServiceError);
+      } finally {
+        global.fetch = originalFetch;
+      }
+    });
```
matthiasgeihs: This test is not doing what it is supposed to be doing.
The mocked fetch function is not used here because the `fetch` function is already passed in the constructor.
The following works (and is simpler):
```ts
    it('should handle null exceptions in catch block', async () => {
      const fetchMock = jest.fn().mockRejectedValueOnce(null);
      const config = createMockConfig({fetchFn: fetchMock});
      const service = new SubscriptionService(config);

      await expect(
        service.cancelSubscription({ subscriptionId: 'sub_123456789' }),
      ).rejects.toThrow(SubscriptionServiceError);
    });
```
(And you have to update `createMockConfig` accordingly.)

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL)
+        .delete('/api/v1/subscriptions/sub_123456789')
+        .reply(200, {});
+
+      await service.cancelSubscription({ subscriptionId: 'sub_123456789' });
+
+      // Verify the correct headers and method were used
+      expect(isDone()).toBe(true);
```
matthiasgeihs: I don't see how this test checks the headers!?

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+    it('should handle empty subscription ID', async () => {
+      const config = createMockConfig();
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL).delete('/api/v1/subscriptions/').reply(200, {});
+
+      await service.cancelSubscription({ subscriptionId: '' });
+
+      expect(isDone()).toBe(true);
+    });
```
matthiasgeihs: Why this test?

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      const service = new SubscriptionService(config);
+
+      nock(TEST_URL)
+        .delete('/api/v1/subscriptions/sub_123-456_789')
+        .reply(200, {});
+
+      await service.cancelSubscription({ subscriptionId: 'sub_123-456_789' });
+
+      expect(isDone()).toBe(true);
+    });
```
matthiasgeihs: Why this test?

Also, these characters are not particularly special looking. An actually interesting test would be to check what happens if you pass in characters that are problematic in URLs such as `?,=,/,&,:`.

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+
+      nock(TEST_URL)
+        .delete('/api/v1/subscriptions/sub_123456789')
+        .reply(200, {});
+
+      await service.getSubscriptions();
+      await service.cancelSubscription({ subscriptionId: 'sub_123456789' });
+
+      expect(config.auth.getAccessToken).toHaveBeenCalledTimes(2);
+    });
```
matthiasgeihs: I don't understand the reason for this test.
If you already tested that `getSubscriptions` and `cancelSubscription` works as expected and asks for an access token, why do you need a test that checks that they also ask for access tokens when used sequentially?

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,359 @@
...
+
+      nock(TEST_URL)
+        .delete('/api/v1/subscriptions/sub_123456789')
+        .reply(200, {});
+
+      await service.getSubscriptions();
+      await service.cancelSubscription({ subscriptionId: 'sub_123456789' });
+
+      expect(isDone()).toBe(true);
+    });
```
matthiasgeihs: Again, not sure why this test is relevant.

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (!response.ok) {
+        throw new Error(
+          `HTTP error message: ${responseBody.message}, error: ${responseBody.error}`,
+        );
+      }
+
+      return responseBody as Result;
+    } catch (e) {
+      const errorMessage =
+        e instanceof Error ? e.message : JSON.stringify(e ?? 'unknown error');
```
matthiasgeihs, line 71: The coverage report shows that the `'unknown error'` branch is not covered.
See https://github.com/MetaMask/core/pull/6233#discussion_r2299066642 for the reason.

packages/subscription-controller/src/constants.test.ts
```diff
@@ -0,0 +1,59 @@
...
+          'https://subscription-service.uat-api.cx.metamask.io',
+      });
+    });
+
+    it('should return correct URLs for prd environment', () => {
+      const result = getEnvUrls(Env.PRD);
+      expect(result).toStrictEqual({
+        subscriptionApiUrl: 'https://subscription-service.api.cx.metamask.io',
+      });
+    });
```
matthiasgeihs, line 26: Again, not sure if these tests add any value. You still have 100% coverage even if you remove them.

packages/subscription-controller/src/constants.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    it('should have correct error messages', () => {
+      expect(SubscriptionControllerErrorMessage.UserAlreadySubscribed).toBe(
+        'SubscriptionController - User is already subscribed',
+      );
+      expect(SubscriptionControllerErrorMessage.UserNotSubscribed).toBe(
+        'SubscriptionController - User is not subscribed',
+      );
+    });
+  });
+});
```
matthiasgeihs: Please remove these. They don't add any value and are just making maintenance harder.

packages/subscription-controller/tsconfig.build.json
```diff
@@ -0,0 +1,20 @@
...
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
+  "references": [
+    {
+      "path": "../base-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../message-manager/tsconfig.build.json"
+    },
```
matthiasgeihs: Why this? Remove?

packages/subscription-controller/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "compilerOptions": {
+    "baseUrl": "./"
+  },
+  "references": [
+    {
+      "path": "../base-controller"
+    },
+    {
+      "path": "../message-manager"
+    },
```
matthiasgeihs: Why this? Remove?

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,84 @@
...
+          ...headers,
+        },
+      });
+
+      const responseBody = (await response.json()) as ErrorMessage;
+      if (!response.ok) {
+        throw new Error(
+          `HTTP error message: ${responseBody.message}, error: ${responseBody.error}`,
+        );
+      }
```
matthiasgeihs: See my reply to the bot. The bot pointed out that it looks odd that we are always casting to `ErrorMessage` here. Instead, you can do the following:
```ts
      const responseBody = await response.json();
      if (!response.ok) {
        const { message, error } = responseBody as ErrorMessage;
        throw new Error(`HTTP error message: ${message}, error: ${error}`);
      }

      return responseBody as Result;
```

matthiasgeihs (changes requested):
Sorry, a bunch more comments. Some are nits, some are a bit more substantial. I think we can further trim the code by a bit to make sure that we start with a lean code base.

(Last time I only commented on things that stood out to me while working on the follow up feature. This time I had a look at the entire code, so I don't expect many more comments after that.)

packages/subscription-controller/src/SubscriptionService.test.ts
MajorLift: The `beforeEach` hook is unnecessary here, since `resetMocks` and `restoreMocks` are set to true in core,
> https://github.com/MetaMask/core/blob/main/jest.config.packages.js#L116-L118
> https://github.com/MetaMask/core/blob/main/jest.config.packages.js#L126-L129

as recommended by the docs.
> https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#restore-function-mocks-after-each-test

```suggestion
  afterEach(() => {
    cleanAll();
  });
```

MajorLift: @matthiasgeihs The reasons that we recommend avoiding the `beforeEach` hook are discussed in detail in the following section of the docs under "Read more":
> https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#avoid-the-use-of-beforeeach

matthiasgeihs: OK, to summarize what the docs say:
Using `beforeEach`...
- ... makes tests harder to read.
- ... makes writing tests for new scenarios difficult.

packages/subscription-controller/src/SubscriptionController.ts

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: updated mandatory `subscriptionService`

packages/subscription-controller/src/SubscriptionService.ts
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  SubscriptionController,
+  getDefaultSubscriptionControllerState,
+} from './SubscriptionController';
+export type * from './types';
+export * from './errors';
+export * from './constants';
```
cryptodev-2s: Nit: What do you think about switching to named exports? We usually prefer explicitly defining exports so it’s clear which ones are intended for external usage versus internal only

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: yes, it's for future reference only, removed for now

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: updated, using `withSubscriptionService` pattern for consistency with subscriptionController.test.ts and following convention

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: they tested different http error code, but not needed in this case, removed

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: @matthiasgeihs it's to check if user already subscribed to the product in local state before making request to server, do you mean we don't need to check that ?

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: removed comment

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: removed

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: removed, since we would need to make request to server to test those special character

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: removed

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: removed

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,199 @@
...
+  async getSubscriptions() {
+    const { subscriptions } =
+      await this.#subscriptionService.getSubscriptions();
+
+    this.update((state) => {
+      state.subscriptions = subscriptions ?? [];
+    });
+
+    return subscriptions;
+  }
```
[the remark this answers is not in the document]
tuna1207: the method doesn't return null

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+  customerId: string;
+  subscriptions: Subscription[];
+  trialedProducts: ProductType[];
+};
+
+export type AuthUtils = {
+  getAccessToken: () => Promise<string>;
+};
+
+export type ISubscriptionService = {
```
cryptodev-2s, line 48: Nit: Could we drop this type? It’s effectively acting as an interface for the service, but since we only have a single implementation, we could reference the class directly in the types instead. What do you think?

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+// state
+export type Subscription = {
+  id: string;
+  products: Product[];
+  currentPeriodStart: string; // ISO 8601
+  currentPeriodEnd: string; // ISO 8601
+  billingCycles?: number;
+  status: 'active' | 'inactive' | 'trialing' | 'cancelled';
```
cryptodev-2s, line 33: Just to clarify, how are cases like `past_due` handled? Do we auto-cancel the subscription if a renewal payment fails? same for `incomplete` ?

packages/subscription-controller/ARCHITECTURE.md
cryptodev-2s, line 1: General question: Do we plan to merge this, and could it be included in the README? Also, it looks like the controller is currently only used to fetch and cancel subscriptions — are the other features planned for the future as described in this file? if we plan to merge it can we just include what it does currently ?

packages/subscription-controller/src/constants.ts
```diff
@@ -0,0 +1,42 @@
...
+  DEV = 'dev',
+  UAT = 'uat',
+  PRD = 'prd',
+}
+
+type EnvUrlsEntry = {
+  subscriptionApiUrl: string;
+};
+
+const ENV_URLS: Record<Env, EnvUrlsEntry> = {
```
cryptodev-2s, line 13: Nit: I tried to use these APIs they don't seem to work yet ?

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: Yes, I think we don't have to check for that. In fact, it might be a bad idea to rely on local state here. Maybe local state is outdated and the user is actually subscribed. But because the state is outdated, you fail to make the request and cancel the subscription.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,211 @@
...
```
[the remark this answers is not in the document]
matthiasgeihs: Since this doesn't return null anymore, you can use `state.subscriptions = subscriptions;` instead of `state.subscriptions = subscriptions ?? [];`.

packages/subscription-controller/src/types.ts
matthiasgeihs: Personally, I think it's helpful to have this concise interface description.

packages/subscription-controller/ARCHITECTURE.md
matthiasgeihs: I think this document might also be a bit out of date by now. Although this is a nice document, the quickest solution might be to simply remove it from the repo and keep the architecture documentation separately.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,201 @@
...
+      messenger,
+    });
+
+    this.#subscriptionService = subscriptionService;
+    if (!this.#subscriptionService.hasAuthUtils()) {
+      this.#subscriptionService.setAuthUtils({
+        getAccessToken: () =>
+          this.messagingSystem.call('AuthenticationController:getBearerToken'),
+      });
+    }
```
matthiasgeihs: You can make the controller agnostic of the fact that the subscription service requires an access token:
If you make `authUtils` mandatory in the subscription service constructor, then you can simplify things as follows:
- You can remove these lines here.
- You can remove the functions `hasAuthUtils` and `setAuthUtils` from SubscriptionService.
- You can remove `AuthenticationControllerGetBearerToken` from the AllowedActions.

The main point being that the controller doesn't need to know that the service requires an access token. Hence, we can remove that complexity completely from the controller.

packages/subscription-controller/src/constants.ts
tuna1207: oh right, the service hasn't been deployed yet, this is following the pre defined interface

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,577 @@
...
+
+          expect(controller.state.subscriptions).toStrictEqual([
+            { ...MOCK_SUBSCRIPTION, status: 'cancelled' },
+          ]);
+
+          expect(mockService.cancelSubscription).toHaveBeenCalledWith({
+            subscriptionId: 'sub_123456789',
+            type: ProductType.SHIELD,
+          });
+          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -0,0 +1,249 @@
...
+            customerId: 'cus_1',
+            subscriptions: [MOCK_SUBSCRIPTION],
+            trialedProducts: [],
+          });
+
+        await service.getSubscriptions();
+
+        expect(isDone()).toBe(true);
+      });
+    });
```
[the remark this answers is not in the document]
tuna1207: corrected nock header test

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: updated to use mandtory `authUtils` in `SubscriptionService`

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,195 @@
...
+  #assertIsUserSubscribed(request: { productType: ProductType }) {
+    if (
+      !this.state.subscriptions.find((subscription) =>
+        subscription.products.some(
+          (product) => product.name === request.productType,
+        ),
+      )
+    ) {
+      throw new Error(SubscriptionControllerErrorMessage.UserNotSubscribed);
+    }
```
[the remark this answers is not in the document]
tuna1207: this logic is correct

packages/subscription-controller/src/types.ts
tuna1207: Hi yes, `past_due` will become `cancelled` when retries time exceeded
you can refer to the state here
image

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,195 @@
...
+    });
+
+    this.update((state) => {
+      state.subscriptions = state.subscriptions.map((subscription) =>
+        subscription.id === request.subscriptionId
+          ? { ...subscription, status: 'cancelled' }
+          : subscription,
+      );
+    });
+  }
```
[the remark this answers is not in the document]
tuna1207: currently we are only having user subscribed to 1 product type at a time so this check is sufficient

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  interval: 'month' | 'year';
+  paymentMethod: PaymentMethod;
+};
+
+export type GetSubscriptionsResponse = {
+  customerId: string;
+  subscriptions: Subscription[];
```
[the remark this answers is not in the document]
tuna1207: it is not null

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  trialedProducts: ProductType[];
+};
```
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type AuthUtils = {
+  getAccessToken: () => Promise<string>;
```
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/types.ts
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (
+      !this.state.subscriptions.find((subscription) =>
+        subscription.products.some(
+          (product) => product.name === request.productType,
+        ),
+      )
+    ) {
+      throw new Error(SubscriptionControllerErrorMessage.UserNotSubscribed);
+    }
+  }
```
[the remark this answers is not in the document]
tuna1207: atm user can only have 1 subscription of a type

packages/subscription-controller/src/SubscriptionController.ts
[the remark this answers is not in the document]
tuna1207: updated

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      throw new SubscriptionServiceError(
+        `failed to make request. ${errorMessage}`,
+      );
+    }
+  }
+
+  async #getAuthorizationHeader(): Promise<{ Authorization: string }> {
+    const accessToken = await this.authUtils?.getAccessToken();
+    return { Authorization: `Bearer ${accessToken}` };
```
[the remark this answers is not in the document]
tuna1207: removed chaining

packages/subscription-controller/src/SubscriptionService.test.ts
tuna1207: updated

packages/subscription-controller/jest.config.js
```diff
@@ -0,0 +1,26 @@
...
+  displayName,
+
+  // An object that configures minimum threshold enforcement for coverage results
+  coverageThreshold: {
+    global: {
+      branches: 92,
+      functions: 98,
+      lines: 99,
+      statements: 99,
+    },
```
matthiasgeihs: Coverage is actually higher than these numbers here. This is what I get:
```
---------------------------|---------|----------|---------|---------|-------------------
File                       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------------|---------|----------|---------|---------|-------------------
All files                  |     100 |    94.44 |     100 |     100 |                   
 SubscriptionController.ts |     100 |    66.66 |     100 |     100 | 174               
 SubscriptionService.ts    |     100 |      100 |     100 |     100 |                   
 constants.ts              |     100 |      100 |     100 |     100 |                   
 errors.ts                 |     100 |      100 |     100 |     100 |                   
 types.ts                  |     100 |      100 |     100 |     100 |                   
---------------------------|---------|----------|---------|---------|-------------------
```
Only one branch not covered. If you address my other comment, you can achieve 100% coverage and update the file here accordingly.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -0,0 +1,434 @@
...
+            { ...MOCK_SUBSCRIPTION, status: 'cancelled' },
+          ]);
+
+          expect(mockService.cancelSubscription).toHaveBeenCalledWith({
+            subscriptionId: MOCK_SUBSCRIPTION.id,
+          });
+          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
+        },
+      );
+    });
```
matthiasgeihs: If you do this, you have 100% coverage. This case should be tested.
```suggestion
    it('should cancel subscription successfully', async () => {
      const mockSubscription2 = { ...MOCK_SUBSCRIPTION, id: 'sub_2' };
      await withController(
        {
          state: {
            subscriptions: [MOCK_SUBSCRIPTION, mockSubscription2],
          },
        },
        async ({ controller, mockService }) => {
          mockService.cancelSubscription.mockResolvedValue(undefined);

          expect(
            await controller.cancelSubscription({
              subscriptionId: MOCK_SUBSCRIPTION.id,
            }),
          ).toBeUndefined();

          expect(controller.state.subscriptions).toStrictEqual([
            { ...MOCK_SUBSCRIPTION, status: 'cancelled' },
            mockSubscription2,
          ]);

          expect(mockService.cancelSubscription).toHaveBeenCalledWith({
            subscriptionId: MOCK_SUBSCRIPTION.id,
          });
          expect(mockService.cancelSubscription).toHaveBeenCalledTimes(1);
        },
      );
    });
```
Edit: This looks like it's a lot of changes. It's really just updating 3 lines of code. (Adding a second subscription to the state and assert that this one is not altered when the other subscription is canceled.)

tuna1207: Thanks, updated

matthiasgeihs (dismissed):

chaitanyapotti (dismissed):

packages/subscription-controller/src/index.ts
tuna1207: Agree, updated 👍

matthiasgeihs (approved):

PR #6234, opened 2025-08-04
feat(account-tree-controller): automatically removes empty groups and wallets

hmalik88:
## Explanation

* What is the current state of things and why does it need to change? Account removal can leave empty nodes in the Account tree controller.
* What is the solution your changes offer and how does it work? My solution prunes the tree when there is an empty group and/or wallet.

## Checklist

### DIFF
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -13,8 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
-- Add `pruneEmptyGroupAndWallet` helper method ([#6234](https://github.com/MetaMask/core/pull/6234))
-  - This method ensures that there aren't any empty nodes in the `AccountTreeController` upon account removal.
+- Automatically prune empty groups and wallets upon account removal ([#6234](https://github.com/MetaMask/core/pull/6234))
+  - This ensures that there aren't any empty nodes in the `AccountTreeController` state.
 
 ### Changed
 
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
+- Add `pruneEmptyGroupAndWallet` helper method ([#6234](https://github.com/MetaMask/core/pull/6234))
+  - This method ensures that there aren't any empty nodes in the `AccountTreeController` upon account removal.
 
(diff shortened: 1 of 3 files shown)


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -270,6 +270,9 @@ export class AccountTreeController extends BaseController<
                 this.#getDefaultAccountGroupId(state.accountTree.wallets);
             }
           }
+          if (accounts.length === 0) {
+            this.#pruneEmptyGroupAndWallet(state, walletId, groupId);
+          }
```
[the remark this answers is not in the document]
hmalik88: That's incorrect, this logic is executed at the point that a group has been found and its accounts array has been operated on (account removal). The method is only ever called with the `walletId` and `groupId` of the removed account. We would never reach a state where the `selectedAccountGroup` points to a non-existent group, because there would have to be at least one account in the group for it to be selected (the pruning takes place AFTER the account removal).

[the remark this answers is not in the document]
hmalik88: Incorrect, see above.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -278,6 +281,30 @@ export class AccountTreeController extends BaseController<
...
+   */
+  #pruneEmptyGroupAndWallet(
+    state: AccountTreeControllerState,
+    walletId: AccountWalletId,
+    groupId: AccountGroupId,
+  ) {
+    delete state.accountTree.wallets[walletId].groups[groupId];
+    if (Object.keys(state.accountTree.wallets[walletId].groups).length === 0) {
+      delete state.accountTree.wallets[walletId];
+    }
```
ccharly: Very nit: WDYT about:

```suggestion
    const { wallets } = state.accountTree;

    delete wallets[walletId].groups[groupId];
    if (Object.keys(wallets[walletId].groups).length === 0) {
      delete wallets[walletId];
    }
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - This `type` can be used as a tag to strongly-type (tagged-union) the `AccountGroupObject`.
 - Add `group.metadata` metadata object ([#6214](https://github.com/MetaMask/core/pull/6214))
   - Given the `group.type` you will now have access to specific metadata information (e.g. `groupIndex` for multichain account groups)
+- Add `pruneEmptyGroupAndWallet` helper method ([#6234](https://github.com/MetaMask/core/pull/6234))
+  - This method ensures that there aren't any empty nodes in the `AccountTreeController` upon account removal.
```
ccharly: Since it's not "consumer-facing", we could maybe say:

```suggestion
- Automatically prune empty groups and wallets upon account removal ([#6234](https://github.com/MetaMask/core/pull/6234))
  - This ensures that there aren't any empty nodes in the `AccountTreeController` state.
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -313,6 +316,34 @@ export class AccountTreeController extends BaseController<
...
+    walletId: AccountWalletId,
+    groupId: AccountGroupId,
+  ) {
+    const { wallets } = state.accountTree;
+
+    delete wallets[walletId].groups[groupId];
+    this.#groupIdToWalletId.delete(groupId);
+
+    if (Object.keys(wallets[walletId].groups).length === 0) {
+      delete state.accountTree.wallets[walletId];
```
ccharly: You should be able to re-use `wallets` reference even there no?

```suggestion
      delete wallets[walletId];
```

hmalik88: Ah yes, missed that! Fixed.

PR #6246, opened 2025-08-06
feat(account-tree-controller): fallback naming for newly created groups

fabiobozzo:
## Explanation

**Problem:**  
When creating new account groups/wallets, naming fails because the rule-based logic depends on existing accounts that don't exist yet for new wallets.

**Solution:**  
Added fallback naming logic that generates "Account 1", "Account 2", etc. within each wallet when rule-based naming returns empty strings. 

The strategy we agreed upon prescribes "indexes per wallet": 
- Wallet 1 → Account 1, Account 2 
- Wallet 2 → Account 1, Account 2

**Changes:**
- Added `#getFallbackAccountGroupName()` method to `AccountTreeController`
- Modified `#applyAccountGroupMetadata()` to use fallback when rule-based naming fails
- Added test coverage for the fallback logic

## References

Fixes naming bug reported by @monte.lai for new account groups/wallets:  
https://consensys.slack.com/archives/C08R8HGNFDH/p1754062266574059

## Checklist


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -206,6 +206,34 @@ export class AccountTreeController extends BaseController<
...
+   * @returns A fallback name like "Account 1", "Account 2", etc.
+   */
+  #getFallbackAccountGroupName(
+    wallet: AccountWalletObject,
+    group: AccountGroupObject,
+  ): string {
+    // Get all groups in this wallet, sorted by their IDs for consistent ordering
+    const walletGroups = Object.values(wallet.groups).sort((a, b) =>
+      a.id.localeCompare(b.id),
+    );
```
ccharly: While this can work for multichain account groups, it might not for other groups 🤔 (like single-account groups)

Since we use the unique-account address to create the `group.id` for those groups, then the ordering cannot be guaranteed the same way that multichain account groups (which uses "group indexes" in their group ids).

So I think for those, we would need to use the `wallet.groups.length` as the index here.

The fallback should only be used for new groups mainly, so I think this should always work 🤔

fabiobozzo: Ok, I tried a different strategy now (please check again). 

That should work because it gives consistent ordering regardless of ID format, and since this fallback is only used for new groups, the position-based numbering should give us the sequential "Account 1, Account 2" names as we want. wdyt @ccharly ?

ccharly: That looks better indeed, however I honestly wonder if we need to `sort` at all 🤔

I wonder if we should re-purpose the `getDefaultAccountGroupName` from each rule to do this. Like we could have 2 variants:
- `getLegacyAccountGroupName` (not sure for "legacy" here) or `getComputedAccountGroupName` which would compute the group name based on the underlying account names
- `getDefaultAccountGroupName` which would compute the actual default name for this rule.
  * This could allow to have even better naming based on the metadata, like for "single-accounts", we could repeat the keyring name in there, so instead of having `Account X` we would have `Ledger Account X` for example
  * We could pass the index of the group within the wallet, so we don't have to "compute it" ourselves in the rule

WDYT?

ccharly: Though, we still to find a trick to know which accounts are newly created, because they would always have a `name`, thus we would always use the `getComputedAccountGroupName` for those too, resulting in a potentially wrong name for the account tree 🤔

Maybe a simple timestamp when the service boots + compare it with the `importTime` of the newly created account could work 🤔

fabiobozzo: I did a deeper refactor, in my last commit you can see now. 

1. I accepted your suggestion of splitting rules' naming methods: 
  - `getComputedAccountGroupName` for existing groups (uses underlying account names)
  - `getDefaultAccountGroupName` for new groups (index-based)
2. I detect new vs existing accounts based on their creation timestamp, against `#serviceStartTime` (controller boot)
3. New naming logic:
  - new groups -> `getDefaultAccountGroupName` -> Account 1, Account 2 ...
  - existing groups -> try `getComputedAccountGroupName` first -> fallback to `getDefaultAccountGroupName`
4. I updated `BaseRule` to handle both naming methods and overridden `EntropyRule`'s implementation. 

Do you think my test case covers it all? 👁️‍🗨️ @ccharly

packages/account-tree-controller/src/AccountTreeController.test.ts
hmalik88, line 1: I think it would be better if we move these tests into `rule.test.ts`, `entropy.test.ts`, `keyring.test.ts` and `snap.test.ts`.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -206,6 +214,42 @@ export class AccountTreeController extends BaseController<
...
+        'AccountsController:getAccount',
+        id,
+      );
+      if (account && account.metadata.importTime > this.#serviceStartTime) {
+        return true;
+      }
+    }
+    return false;
+  }
+
```
hmalik88: We should probably just add that piece of information when the group is created in the `match` method of each rule instead of doing repeated calls for each account again.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -216,23 +260,20 @@ export class AccountTreeController extends BaseController<
...
-          wallet.groups[group.id],
-        );
-      } else {
-        group.metadata.name = this.#getKeyringRule().getDefaultAccountGroupName(
-          // Same here.
-          wallet.groups[group.id],
-        );
-      }
+      // Sort group IDs for consistent ordering
+      const sortedGroupIds = Object.keys(wallet.groups).sort();
```
hmalik88: Is this needed since by design the group id is sequential? In JS, I'm pretty sure `Object.keys` maintains the order in which keys are added.

packages/account-tree-controller/src/rules/keyring.ts
```diff
@@ -93,4 +94,17 @@ export class KeyringRule
   ): string {
     return getAccountWalletNameFromKeyringType(wallet.metadata.keyring.type);
   }
+
+  getComputedAccountGroupName(
```
david0xd, line 98: We should have granular unit tests targeting testing for these functions specifically. I know this is being called somewhere else in the process and covered, but to ensure the integrity of these we should still have them covered in their own test files like `keyring.test.ts`.

packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -105,4 +105,11 @@ export class EntropyRule
 
     return candidate;
   }
+
+  getDefaultAccountGroupName(
```
david0xd: I know that unfortunately `entropy.test.ts` doesn't exist, but we should take care of it. Create it and test these new functions there at least.

packages/account-tree-controller/src/rules/snap.ts
```diff
@@ -92,4 +93,17 @@ export class SnapRule
 
     return snapName;
   }
+
+  getComputedAccountGroupName(
```
david0xd, line 97: It would be good to have these covered in newly created `snap.test.ts` file as well.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -206,6 +214,42 @@ export class AccountTreeController extends BaseController<
     }
   }
 
+  #getRuleForWallet<WalletType extends AccountWalletType>(
```
david0xd, line 270: Please add docstring with params and return to these new functions. We usually do that, for consistence and maintenance when calling these even within these kind of controllers given that they're getting pretty bloated with functional stuff quickly.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -206,6 +214,42 @@ export class AccountTreeController extends BaseController<
...
+      default:
+        return this.#getKeyringRule() as unknown as Rule<
+          WalletType,
+          AccountGroupType
+        >;
+    }
+  }
+
+  #isNewGroup(group: AccountGroupObject): boolean {
+    // Check if any account in the group was created after service start
```
david0xd: Please add docstring and explain what this function is doing, and remove this comment (move it in docstring). That way we could have more consistent approaches of documenting code and can easier reuse it later on.

packages/account-tree-controller/src/rule.ts
```diff
@@ -89,18 +101,32 @@ export class BaseRule {
...
+
+  /**
+   * Gets default name for a group based on its position in the wallet.
+   *
+   * @param _group - Group associated to this rule.
+   * @param index - The group's position within its wallet.
+   * @returns The default name for that group.
+   */
+  getDefaultAccountGroupName(
+    _group: AccountGroupObject,
```
david0xd: I'm not sure why `_group` param is needed here?

packages/account-tree-controller/src/rule.ts
```diff
@@ -89,18 +101,32 @@ export class BaseRule {
...
+  }
+
+  /**
+   * Gets default name for a group based on its position in the wallet.
+   *
+   * @param _group - Group associated to this rule.
+   * @param index - The group's position within its wallet.
+   * @returns The default name for that group.
+   */
+  getDefaultAccountGroupName(
```
david0xd: We should have tests for this function as well somewhere in `rule.test.ts`.

packages/account-tree-controller/src/rules/entropy.ts
ccharly: I think this also makes sense we start splitting those tests, especially that now this controller is getting a bit "packed" with more and more logic 😅

david0xd: We should do it while it's easy to do. And it is still easy for now 😂

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: So, I did a bit of research on this: 

Group IDs are sequential by design (wallet123/0, wallet123/1, wallet123/2, etc.) and as you pointed out, Object.keys() does maintain insertion order since ES2015. So explicit sorting is unnecessary overhead. 

I tried another approach which should be more robust: direct numeric extraction. Please let me know what you think about it 👁️‍🗨️ 
```
const groupIdParts = group.id.split('/');
const groupIndex = parseInt(groupIdParts[groupIdParts.length - 1], 10);
```

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Method removed as per @hmalik88 comment above ☝️

packages/account-tree-controller/src/rules/keyring.ts
fabiobozzo: I added test coverage for SnapRule methods:

`getComputedAccountGroupName`:
- happy case: returns account metadata name when account exists
- edge case: returns empty string when account not found

`getDefaultAccountGroupName`:
- index-based naming: "Account 1", "Account 2", etc.

`getDefaultAccountWalletName tests`:
- proposed name: returns snap's proposed name when available
- no proposed name: undefined when snap exists but no proposed name (current behavior)
- snap not found: Returns cleaned snap ID when snap is not found

packages/account-tree-controller/src/rule.ts
fabiobozzo: Good catch, I decided to keep it for interface consistency. I marked it with `_` to hint the linter that it's not used her locally (... yet? not sure how it could be useful in the future)

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Added JSDoc ✔️

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: @hmalik88 Good question! While Object.keys() does guarantee insertion order (ES2015+), we do need .sort() here for a specific reason **related to Entropy wallets**.

Problem:
Entropy wallet group IDs contain logical indices: entropy:abc/0, entropy:abc/1, entropy:abc/2
Without sorting: If accounts are processed out of order (e.g., groupIndex 1 before groupIndex 0), insertion order would be:

```typescript
Object.keys(wallet.groups) // ["/1", "/0", "/2"]
// Result: groupIndex 0 → "Account 2" ❌ (should be "Account 1")
```

I solve that with sorting: alphabetical sort gives us logical order.
```typescript
Object.keys(wallet.groups).sort() // ["/0", "/1", "/2"]
// Result: groupIndex 0 → "Account 1" ✅
```

The sorting ensures that "Account N" names match the logical group indices rather than arbitrary processing order, imho!
- Entropy wallets: Sort gives logical order (/0, /1, /2)
- Snap/Keyring wallets: Sort gives consistent alphabetical order
- All wallet types: Deterministic, predictable naming

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -270,23 +308,24 @@ export class AccountTreeController extends BaseController<
...
+      // - Snap/Keyring: group IDs like "keyring:ledger/0xABC" get consistent alphabetical order
+      const sortedGroupIds = Object.keys(wallet.groups).sort();
+      const groupIndex = sortedGroupIds.indexOf(group.id);
+
+      // For new groups, use default naming. For existing groups, try computed name first
+      const isNewGroup = this.#newGroupsMap.get(group) || false;
+      group.metadata.name = isNewGroup
+        ? rule.getDefaultAccountGroupName(typedGroup, groupIndex)
+        : rule.getComputedAccountGroupName(typedGroup) ||
+          rule.getDefaultAccountGroupName(typedGroup, groupIndex);
```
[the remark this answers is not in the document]
fabiobozzo: FIXED - This issue has been resolved by the changes in this PR

packages/account-tree-controller/src/rules/snap.ts
```diff
@@ -92,4 +93,17 @@ export class SnapRule
 
     return snapName;
   }
+
```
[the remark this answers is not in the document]
fabiobozzo: FIXED - This issue has been resolved by the changes in this PR

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -270,23 +302,22 @@ export class AccountTreeController extends BaseController<
...
+      const rule = this.#getRuleForWallet(wallet);
+      const typedWallet = wallet as AccountWalletObjectOf<typeof wallet.type>;
+      const typedGroup = typedWallet.groups[group.id];
+
+      // For new groups, use default naming. For existing groups, try computed name first
+      const isNewGroup = this.#newGroupsMap.get(group) || false;
+      group.metadata.name = isNewGroup
+        ? rule.getDefaultAccountGroupName(typedGroup, groupIndex)
+        : rule.getComputedAccountGroupName(typedGroup) ||
+          rule.getDefaultAccountGroupName(typedGroup, groupIndex);
```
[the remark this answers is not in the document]
fabiobozzo: FIXED - This issue has been resolved by the changes in this PR

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -270,23 +312,20 @@
...
+      // Get the appropriate rule for this wallet type
+      const rule = this.#getRuleForWallet(wallet);
+      const typedWallet = wallet as AccountWalletObjectOf<typeof wallet.type>;
+      const typedGroup = typedWallet.groups[group.id];
+
+      // For new groups, use default naming. For existing groups, try computed name first
+      group.metadata.name = this.#isNewGroup(group)
+        ? rule.getDefaultAccountGroupName(typedGroup, groupIndex)
+        : rule.getComputedAccountGroupName(typedGroup) ||
+          rule.getDefaultAccountGroupName(typedGroup, groupIndex);
```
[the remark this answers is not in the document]
fabiobozzo: FIXED - This issue has been resolved by the changes in this PR

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: I added a weak map to keep track of "new" accounts and avoid roundtrips to AccountsController. 👍

packages/account-tree-controller/src/rules/entropy.ts
fabiobozzo: See comment above ☝️

packages/account-tree-controller/src/rules/snap.test.ts
```diff
@@ -0,0 +1,296 @@
+/* eslint-disable jest/no-conditional-in-test */
```
david0xd: Why is this line needed?

packages/account-tree-controller/src/rule.ts
david0xd: It looks redundant, no real usage now or for anything that is using that API. It should be removed.

fabiobozzo: Agreed and addressed. 👍 The group param was unused in all implementations and call sites, so I removed it for clarity. The API is now getDefaultAccountGroupName(index: number). If we need group context later, we can add it back as an options parameter in a backward-compatible way.

packages/account-tree-controller/src/rules/snap.test.ts
fabiobozzo: It was needed by some conditional logic in account retrieval, for those tests. I worked it around in a simpler way that makes the linter pass too. 👌

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -270,23 +308,24 @@ export class AccountTreeController extends BaseController<
...
+      // - Snap/Keyring: group IDs like "keyring:ledger/0xABC" get consistent alphabetical order
+      const sortedGroupIds = Object.keys(wallet.groups).sort();
+      const groupIndex = sortedGroupIds.indexOf(group.id);
+
+      // For new groups, use default naming. For existing groups, try computed name first
+      const isNewGroup = this.#newGroupsMap.get(group) || false;
+      group.metadata.name = isNewGroup
+        ? rule.getDefaultAccountGroupName(groupIndex)
+        : rule.getComputedAccountGroupName(typedGroup) ||
+          rule.getDefaultAccountGroupName(groupIndex);
```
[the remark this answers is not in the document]
fabiobozzo: Issue #1 
Group Index -1 Error: groupIndex could be -1 if group.id was not found in sortedGroupIds.
I added a defensive fallback to clamp groupIndex to 0 when not found.

Issue #2
WeakMap flags could persist across rebuilds, but new AccountGroupObject instances would cause lookups to fail
I added cleanup to delete the flag after use, making it strictly ephemeral per-insert.

PR #6251, opened 2025-08-06
feat(account-tree-controller): add `AccountsController:accountRenamed` event handling

mathieuartu:
## Explanation

This PR adds support for handling `AccountsController:accountRenamed` events in `AccountTreeController` to maintain compatibility with both legacy account syncing and the multichain accounts state 1 implementation.

`AccountTreeController` now subscribes to and handles `AccountsController:accountRenamed` events. When an EVM account is renamed, the controller automatically updates the corresponding account group name if the group has a default name pattern ("Account X") but the account now has a custom name. This ensures that custom account names are reflected at the group level for better user experience.

This change maintains support for legacy account syncing flows that depend on programmatic account renames while ensuring `AccountTreeController` state remains consistent with account renames even though the UI doesn't use the tree structure yet in multichain state 1.

## References

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-521

## Checklist


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -316,6 +324,29 @@ export class AccountTreeController extends BaseController<
...
+    // This method is only there to support legacy account syncing programmatic renames.
+    // Since legacy account syncing only syncs HD EVM accounts, we only
+    // handle HD EVM accounts here.
+    if (account.metadata.keyring.type !== (KeyringTypes.hd as string)) {
+      return;
+    }
+
+    const context = this.#accountIdToContext.get(account.id);
+
+    if (context) {
```
fabiobozzo, line 339: And if it's not found.... It's some kind of inconsistency isn't it? But I guess it's pointless to throw anything here.

fabiobozzo (approved):

fabiobozzo (dismissed):

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -316,6 +323,37 @@ export class AccountTreeController extends BaseController<
...
+      const { walletId, groupId } = context;
+
+      const wallet = this.state.accountTree.wallets[walletId];
+      if (wallet) {
+        const group = wallet.groups[groupId];
+        if (group) {
+          const isDefaultNameRegex = /^Account ([0-9]+)$/u;
```
ccharly: Can you put this in the top-level like `DEFAULT_ACCOUNT_NAME_REGEX` maybe?

I'll aso add a quick comment explaining that we both use the same naming convention for BIP-44 accounts and legacy HD accounts.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -316,6 +323,37 @@ export class AccountTreeController extends BaseController<
     }
   }
 
+  #handleAccountRenamed(account: InternalAccount) {
+    if (!isEvmAccountType(account.type)) {
```
ccharly, line 333:
```suggestion
    // We only consider EVM accounts for the moment as they have
    // an higher priority over others when it comes to naming.
    // (Similar logic than `EntropyRule.getDefaultAccountGroupName`).
    if (!isEvmAccountType(account.type)) {
```

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Also, I think we can add a comment explaining we only consider HD accounts and private key accounts for now.

Because `isEvmAccountType` would also be true for private keys/hardware wallet accounts for example.

But this regex will not match those. (Unless we're re-using the custom prefix we have with `keyringToTypeNames` which is used by the `AccountsController` default naming logic)

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for `AccountsController:accountRenamed` event handling for state 1 and legacy account syncing compatibility ([#6251](https://github.com/MetaMask/core/pull/6251))
```
ccharly: It should be breaking cause we need to declare this event on the controller-init of this controller.

```suggestion
- **BREAKING:** Add support for `AccountsController:accountRenamed` event handling for state 1 and legacy account syncing compatibility ([#6251](https://github.com/MetaMask/core/pull/6251))
```

PR #6262, opened 2025-08-07
fix: Update `unsubscribe` type to support selector subscriptions

### LINKED ISSUE #6200: [base-controller, messenger] `unsubscribe` type signature does not support subscriptions with selectors
The `unsubscribe` method of the `Messenger` and `RestrictedMessenger` classes does not support subscriptions that have selectors.

`subscribe` and `unsubscribe` accept a handler of the type `ExtractEventHandler<Event, EventType>`. However, `subscribe` also supports handlers of the type `SelectorEventHandler<SelectorReturnValue>` for selector subscriptions, but this type is not accepted by `unsubscribe`. `unsubscribe` still works if you call it with a selector handler anyway, but there is a type error.

We should update the type for the `unsubscribe` function to allow it to be called with selector subscription hanlders.

Gudahtt:
## Explanation

The `unsubscribe` type signature has been updated to allow passing in event handlers for subscriptions that use a selector function. This was missed when the selector feature was added.

## References

Fixes #6200

## Checklist


packages/base-controller/src/Messenger.ts
```diff
@@ -382,7 +382,7 @@ export class Messenger<
    */
   unsubscribe<EventType extends Event['type']>(
     eventType: EventType,
-    handler: ExtractEventHandler<Event, EventType>,
+    handler: ExtractEventHandler<Event, EventType> | SelectorEventHandler,
```
MajorLift: nit: Probably unnecessary, but would this make sense as a future-proofing measure?

```suggestion
  unsubscribe<EventType extends Event['type'], SelectorReturnValue = unknown>(
    eventType: EventType,
    handler: ExtractEventHandler<Event, EventType> | SelectorEventHandler<SelectorReturnValue>,
```

Gudahtt: Interesting idea! This would allow for a more precise internal type, but unfortunately this would currently introduce a type error because of how the type is structured.

I discovered this problem in the `subscribe` implementation while I was working on this, and I have a separate PR to address it that is based on this branch: https://github.com/MetaMask/core/pull/6264

If we add a type parameter as this suggests, we'd be forced to cast to throw that type specificity away to overcome the same problem. The cast is not a dealbreaker exactly, but given that there is no clear advantage either (no selector parameter to type check the handler against), it doesn't seem worthwhile to me.

mcmire: Should we add a test for this to confirm that the types would work? (Same goes for the changes to the `messenger` package.)

packages/messenger/src/RestrictedMessenger.ts
```diff
@@ -342,7 +342,10 @@ export class RestrictedMessenger<
     EventType extends
       | AllowedEvent
       | (Event['type'] & NamespacedName<Namespace>),
-  >(event: EventType, handler: ExtractEventHandler<Event, EventType>) {
+  >(
+    event: EventType,
+    handler: ExtractEventHandler<Event, EventType> | SelectorEventHandler,
```
mcmire: I see that changes were made to RestrictedMessenger in this package but not in `base-controller`, should we copy them there too?

Gudahtt: Whoops, great catch! Done here: 9cda93451

packages/base-controller/src/Messenger.ts
Gudahtt: Great idea. I had initially not bothered because I was just working on type tests for the delegated messenger PR, so I thinking "I can't add tests for this now, we don't have a type testing tool in this repo yet, I'll have to come back to that later".

But... a regular test would also suffice for validating the type 😅. I will work on that now.

Gudahtt: Nevermind, I've just found that the same type bug I'm addressing in #6264 impacts this method as well. And unlike in that case, it would impact callers as well (because there is no type override shielding callers from it). I believe I'll have to address it here using your suggestion of adding a ReturnType parameter.

Gudahtt: There was already a test, but it made extensive use of `sinon.stub` with the default `any` as the type parameter, so it wasn't much good for validating the type problem. I've updated the test to minimize the use of `sinon`, ensuring all relevant functions are typed normally.

Updated here: https://github.com/MetaMask/core/pull/6262/commits/000fb9f22d01cd059854a852ed17f6348b26f6bb

This revealed a new type error that I had to resolve as well (by taking @MajorLift 's suggestion above)

Gudahtt: Done here: https://github.com/MetaMask/core/pull/6262/commits/000fb9f22d01cd059854a852ed17f6348b26f6bb

MajorLift: Nice! Good to know that this was actually useful :)

PR #6275, opened 2025-08-08
Feat/separate revoke renew refresh token

tuna1207:
## Explanation

Previously revoke token api will revoke and create new refresh token at the same time.
This PR separate them into:
- `renewRefreshToken` method to renew refresh token from client
- `revokePendingRefreshTokens` method to revoke all pending old refresh tokens instead from client

## Checklist


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1867,6 +1886,63 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+   */
+  async revokePendingRefreshTokens() {
+    this.#assertIsAuthenticatedUser(this.state);
+    const { pendingToBeRevokedTokens } = this.state;
+    if (!pendingToBeRevokedTokens || pendingToBeRevokedTokens.length === 0) {
+      return;
+    }
+
+    for (const { revokeToken } of pendingToBeRevokedTokens) {
+      await this.#revokeRefreshToken({
```
lwin-kyaw: We should call it parallelly. e.g `Promise.allSettled`

tuna1207: Updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1867,6 +1886,71 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+  async revokePendingRefreshTokens() {
+    this.#assertIsAuthenticatedUser(this.state);
+    const { pendingToBeRevokedTokens } = this.state;
+    if (!pendingToBeRevokedTokens || pendingToBeRevokedTokens.length === 0) {
+      return;
+    }
+
+    // revoke all pending refresh tokens in parallel
+    const promises = pendingToBeRevokedTokens.map(({ revokeToken }) =>
+      (async () => {
```
lwin-kyaw: Why do we use IIFE here, instead of simply returning the promise and resolve it later?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1867,6 +1886,71 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+            connection: this.state.authConnection as AuthConnection,
+            revokeToken,
+          });
+          this.#removePendingRefreshToken({ revokeToken });
+        } catch (error) {
+          log('Error revoking refresh token', error);
+        }
+      })(),
+    );
+    await Promise.allSettled(promises);
```
lwin-kyaw: since nothing is returned from the `pendingToBeRevokedTokens.map` the promises here is an array of `undefined`.

lwin-kyaw:
```suggestion
    const settledResults =  await Promise.allSettled(
      pendingToBeRevokedTokens.map(({ revokeToken }) => {
        return this.#revokeRefreshToken({
          connection: this.state.authConnection as AuthConnection,
          revokeToken,
        });
      })
    );
    
    // check the settled result and update state
    settledResults.forEach((result, index) => {
      const revokeToken = pendingToBeRevokedTokens[index].revokeToken;
      if (result.status === 'fulfilled'  && revokeToken) {
        // if revoke request is successful, removed the token from the pending list
        this.#removePendingRefreshToken({ revokeToken });
      }
      // Otherwise, no need to do update the state
    });
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: We shouldn't use async callback in `Array.map()` since `Array.protoype.map()` is synchronous and even with using `async/await` in callback, it will still return promises.
e.g
```ts
const res = [1, 2, 3].map(async (v) => {
  const val = await someFunc(v)
  return val
})

console.log(res) // [Promise, Promise, Promise]
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1867,6 +1886,71 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
 
+  /**
+   * Revoke all pending refresh tokens.
+   *
+   * This method is to be called after user is authenticated.
+   */
+  async revokePendingRefreshTokens() {
+    this.#assertIsAuthenticatedUser(this.state);
+    const { pendingToBeRevokedTokens } = this.state;
+    if (!pendingToBeRevokedTokens || pendingToBeRevokedTokens.length === 0) {
```
lwin-kyaw: nit: we can skip this and asserts optional chaining (?) in below array.map.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: This make no affects since the promises are array of `undefined`.

lwin-kyaw (changes requested):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: @lwin-kyaw the mapping returned the result of IIFE which is the promise so these would be an array of promises, but you're right, IIFE is not needed, we should return the promise directly for better readbility

lwin-kyaw: In that case, `promise.allSettled` is kinda redundant since `try...catch` is already inside iife?

tuna1207: Agree, updated 👍

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1856,6 +1875,81 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+      return;
+    }
+
+    // revoke all pending refresh tokens in parallel
+    const promises = pendingToBeRevokedTokens.map(({ revokeToken }) => {
```
ieow: #removePendingRefreshToken remove all refresh token related to revoke token locally but we are looping all tokens pairs here?

Shouldn't we group to same revoke token and just call 1 api call for the same revoke token

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1856,6 +1875,81 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+
+  /**
+   * Remove a pending refresh, revoke token from the state.
+   *
+   * @param params - The parameters for removing a pending refresh, revoke token.
+   * @param params.revokeToken - The revoke token to remove.
+   */
+  #removePendingRefreshToken({ revokeToken }: { revokeToken: string }) {
+    this.update((state) => {
+      state.pendingToBeRevokedTokens = state.pendingToBeRevokedTokens?.filter(
```
himanshuchawla009: we should acquire lock on state before adding or removing revoked tokens since we are calling the revocation of all tokens parrallely. It can cause list having stale data due to race conditions.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -140,6 +141,10 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
       persist: false,
       anonymous: false,
     },
+    pendingToBeRevokedTokens: {
+      persist: true,
+      anonymous: true,
```
lwin-kyaw: These should not be sent to sentry.
Ref: https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#define-metadata-for-state-properties
```suggestion
      anonymous: false,
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -182,6 +187,8 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
 
   readonly #revokeRefreshToken: RevokeRefreshToken;
 
+  readonly #renewRefreshToken: RenewRefreshToken;
```
lwin-kyaw, line 197: At this point, I think the controller should accept the `OAuthService` instance, rather than adding multiple callbacks.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
himanshuchawla009: its not a already anonymized property right since it has tokens which contains user related info?  All other tokens are marked `anonymous` false so it should be as well.

himanshuchawla009 (changes requested):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1856,6 +1875,81 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+  }
+
+  /**
+   * Add a pending refresh, revoke token to the state to be revoked later.
+   *
+   * @param params - The parameters for adding a pending refresh, revoke token.
+   * @param params.refreshToken - The refresh token to add.
+   * @param params.revokeToken - The revoke token to add.
+   */
+  #addPendingRefreshToken({
```
lwin-kyaw: Can we make this method name clearer? This method basically does adding the token to the revokeList right?
```suggestion
  #addToRevokeList({
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1856,6 +1875,81 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
 
+  /**
+   * Revoke all pending refresh tokens.
+   *
+   * This method is to be called after user is authenticated.
+   */
+  async revokePendingRefreshTokens() {
+    this.#assertIsAuthenticatedUser(this.state);
+    const { pendingToBeRevokedTokens } = this.state;
+    if (!pendingToBeRevokedTokens || pendingToBeRevokedTokens.length === 0) {
```
lwin-kyaw: nit: can skip the length check.
```suggestion
    if (!pendingToBeRevokedTokens) {
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1856,6 +1875,81 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+  }
+
+  /**
+   * Remove a pending refresh, revoke token from the state.
+   *
+   * @param params - The parameters for removing a pending refresh, revoke token.
+   * @param params.revokeToken - The revoke token to remove.
+   */
+  #removePendingRefreshToken({ revokeToken }: { revokeToken: string }) {
+    this.update((state) => {
```
lwin-kyaw: We should not update the state concurently. This is being run in parallel from [`#pendingToBeRevokedTokens` method.](https://github.com/MetaMask/core/pull/6275/files#diff-7ec8aa1b0ce65286f76f0cc6467231ec9f79c7cefd409bb13bb0d20bd258b52aR1894)
It's possible that the state can be out of sync, having multiple updates at the same time.

Instead we can create an array of successfully revoked token-pairs and update the state after the concurrent operations.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -4022,6 +4045,69 @@ describe('SeedlessOnboardingController', () => {
...
+          }),
+        },
+        async ({ controller, toprfClient }) => {
+          // Mock fetchAuthPubKey to throw a non-token-related error
+          jest
+            .spyOn(toprfClient, 'fetchAuthPubKey')
+            .mockRejectedValue(new Error('Network error'));
+
+          // This should throw the wrapped error without retrying
+          await expect(controller.checkIsPasswordOutdated()).rejects.toThrow(
```
lwin-kyaw, line 4069: There's an issue in `checkIsPasswordOutdated()` method. 
We will never be able to catch the `TokenExpired` from `checkIsPasswordOutdated()` method coz we masked the original error and [throw a different error.](https://github.com/MetaMask/core/blob/main/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts#L873)

Not related to this PR, we can address this in another one.

packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -250,7 +259,15 @@ export type RefreshJWTToken = (params: {
 export type RevokeRefreshToken = (params: {
   connection: AuthConnection;
   revokeToken: string;
-}) => Promise<{ newRevokeToken: string; newRefreshToken: string }>;
+}) => Promise<void>;
+
+export type RenewRefreshToken = (params: {
+  connection: AuthConnection;
+  revokeToken: string;
```
ieow, line 264: @himanshuchawla009 , this look a bit weird
can ignore for now

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -5483,31 +5378,255 @@ describe('SeedlessOnboardingController', () => {
...
+        {
+          state: getMockInitialControllerState({
+            withMockAuthenticatedUser: true,
+            pendingToBeRevokedTokens: null as unknown as
+              | {
+                  refreshToken: string;
+                  revokeToken: string;
+                }[]
+              | undefined,
+          }),
```
lwin-kyaw: imo, this is cleaner. `null` is not an acceptable type for `pendingToBeRevokedTokens` state, so we no need to do hard casts.
```suggestion
            // @ts-expect-error - test `null` value case
            pendingToBeRevokedTokens: null,
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -5483,31 +5378,255 @@ describe('SeedlessOnboardingController', () => {
...
+        },
+        async ({ controller, mockRevokeRefreshToken }) => {
+          await controller.revokePendingRefreshTokens();
+
+          expect(mockRevokeRefreshToken).not.toHaveBeenCalled();
+        },
+      );
+    });
+
+    it('should handle case where revoke token is not found in pending list', async () => {
```
lwin-kyaw: I think the test case and title is very different. 😅

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: I think we should not update the state while in `Promise.all` at all.
We should create a list of items to be removed from the state and update the state after the concurrent operations.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: I think this make sense. It's better to handle this in the server side rather than in the frontend.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: empty array in js is still considered truthy value so it's different from `undefined/null`
e.g:
```javascript
![] === fase
!null === true
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
tuna1207: you're right, removed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: updated

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: agree, i think we can create a separate PR for this

lwin-kyaw (commented):
LGTM!

ieow:
@metamaskbot  publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
ieow: separate pr but should be in the same next controller version else will be another breaking.
also create as interface with the three functions signature instead of Oauth instance class

lwin-kyaw:
@metamaskbot publish-preview

ieow:
lgtm

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1812,6 +1831,74 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+            });
+            return revokeToken;
+          } catch (error) {
+            log('Error revoking refresh token', error);
+            return null;
+          }
+        };
+        return revokePromise();
+      });
+      const result = await Promise.all(promises); // no need to do Promise.allSettled because the promise already handle try catch
```
ieow, line 1865: Should use allSettled as if one of the request failed, it will fails all

ieow: noted that the inner promise already catch the exceptions

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1812,6 +1831,74 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
+      // filter out the null values
+      const revokedTokens = result.filter((token) => token !== null);
+      if (revokedTokens.length > 0) {
+        // update the state to remove the revoked tokens once all concurrent token revoke finish
+        this.update((state) => {
+          state.pendingToBeRevokedTokens =
```
himanshuchawla009, line 1871: what if `addRefreshTokenToRevokeList` is called when this state update is in progress.

himanshuchawla009 (changes requested):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
tuna1207: @himanshuchawla009 this is only being called inside `revokePendingRefreshTokens` which already using `withControllerLock` so there shouldn't be another state update in progress

PR #6281, opened 2025-08-11
Integrate Shield gateway in transaction controller

matthiasgeihs:
## Explanation

For MetaMask Transaction Shield we want to reroute all transaction simulation requests through a dedicated proxy.
This PR adds the following functionality to the TransactionController to accomplish that.

- Add optional `getSimulationConfig` parameter to constructor.
- If `getSimulationConfig` parameter is present, call the function to retrieve a new simulation URL and an authorization header. Replace the simulation URL and add the authorization header to the simulation request.

## References

[Transaction Shield Notion Page](https://www.notion.so/Shield-1c3f86d67d68801f8821dae7af4de4bc?source=copy_link)

## Checklist

### DIFF
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add optional `metamaskPay` and `requiredTransactionIds` properties to `TransactionMeta` ([#6361](https://github.com/MetaMask/core/pull/6361))
   - Add `updateRequiredTransactionIds` method.
+- Add `getSimulationConfig` constructor property ([#6281](https://github.com/MetaMask/core/pull/6281))
 
 ### Changed
 
@@ -21,7 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add `isGasFeeSponsored` property to `TransactionMeta` type ([#6244](https://github.com/MetaMask/core/pull/6244))
-- Add `getSimulationConfig` constructor property ([#6281](https://github.com/MetaMask/core/pull/6281))
 
 ### Changed
 


packages/transaction-controller/src/TransactionController.ts
```diff
@@ -450,6 +451,11 @@ export type TransactionControllerOptions = {
     ) => Promise<{ transactionHash: string }>;
     publishBatch?: PublishBatchHook;
   };
+
+  /**
+   * Optional parameters for the transaction shield.
+   */
+  shieldParams?: ShieldParams;
```
matthewwalsh0: Assuming we definitely want to go the client route, instead of a remote-only service / cache of some kind, then the `TransactionController` should not know about `Shield` at all.

It's a fundamental controller, so it's vital we abstract and decouple other features and domain for the sake of maintainability and coupling.

Could this instead be something like:

```
getSimulationConfig?: () => Promise<{url?: string; authorization?: string;}>;
```

matthewwalsh0 (dismissed):

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: will adapt the PR accordingly

matthiasgeihs: updated to
```ts
export type GetSimulationConfig = () => Promise<{
  baseUrl?: string;
  authorization?: string;
}>;
```

matthiasgeihs: realized that we need to provide the current URL as a request parameter included in the proxy URL.
update to
```ts
export type GetSimulationConfig = (url: string) => Promise<{
  newUrl?: string;
  authorization?: string;
}>;
```

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -843,6 +851,7 @@ export class TransactionController extends BaseController<
       publicKeyEIP7702,
       securityProviderRequest,
       sign,
+      getSimulationConfig,
```
OGPoyraz: Minor - alphabetical

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -4171,6 +4181,7 @@ export class TransactionController extends BaseController<
             ethQuery: this.#getEthQuery({ networkClientId }),
             nestedTransactions,
             txParams,
+            getSimulationConfig: this.#getSimulationConfig,
```
OGPoyraz: Minor alphabetical

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -4191,6 +4202,7 @@ export class TransactionController extends BaseController<
         messenger: this.messagingSystem,
         publicKeyEIP7702: this.#publicKeyEIP7702,
         transactionMeta,
+        getSimulationConfig: this.#getSimulationConfig,
```
OGPoyraz: Minor alphabetical

packages/transaction-controller/src/api/simulation-api.ts
```diff
@@ -111,6 +112,11 @@ export type SimulationRequest = {
    * Defaults to false.
    */
   withLogs?: boolean;
+
+  /**
+   * Optional function to modify the simulation request.
+   */
+  getSimulationConfig?: GetSimulationConfig;
```
OGPoyraz: Minor alphabetical

packages/transaction-controller/src/api/simulation-api.ts
```diff
@@ -274,7 +280,13 @@ export async function simulateTransactions(
   chainId: Hex,
   request: SimulationRequest,
 ): Promise<SimulationResponse> {
-  const url = await getSimulationUrl(chainId);
+  let url = await getSimulationUrl(chainId);
+
+  const { newUrl, authorization } =
+    (await request.getSimulationConfig?.(url)) || {};
```
OGPoyraz: Should this be a blocker in case of error? Would it make sense to wrap it in try-catch and fallback to let it work as is?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -899,6 +908,7 @@ export class TransactionController extends BaseController<
       hooks?.publish ?? (() => Promise.resolve({ transactionHash: undefined }));
     this.#publishBatchHook = hooks?.publishBatch;
     this.#securityProviderRequest = securityProviderRequest;
+    this.#getSimulationConfig = getSimulationConfig;
```
OGPoyraz: Shall we add a default property like we did it for other optional constructor params as preferred way? 

```suggestion
    this.#getSimulationConfig = getSimulationConfig ?? ((url) => Promise.resolve({newUrl: url});
```

packages/transaction-controller/src/api/simulation-api.ts
OGPoyraz: We can get rid of optional chaining and || {} if we apply [this suggestion](https://github.com/MetaMask/core/pull/6281/files/83695c3d022f6ac1d2fd482dcfd5ad80251e408d#r2287024863)

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: done

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: done

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: done

packages/transaction-controller/src/api/simulation-api.ts
matthiasgeihs: done

packages/transaction-controller/src/api/simulation-api.ts
matthiasgeihs: In our use case, it should throw. Otherwise our backend would not receive the necessary information for Shield transaction protection. I can see that there may be cases where it's OK to fallback to the default. However, I'd leave it to the implementation to do the error handling. What do you think?

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: If I understand correctly, you are basically suggesting that we make `getSimulationConfig` a mandatory parameter of the class and all the functions that use it. To be honest, I don't see a clear benefit of doing so. Maybe you can give some insights?

On the contrary, I do think that this can lead to unnecessary code bloat, as all callers are then forced to provide an additional parameter, even if they are perfectly fine with the default behavior.

packages/transaction-controller/src/api/simulation-api.ts
matthiasgeihs: replied above https://github.com/MetaMask/core/pull/6281#discussion_r2287362066

matthiasgeihs:
@metamaskbot publish-preview

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: Not a mandatory parameter but this will make `this.#getSimulationConfig` to be typed and defined all the time. 

There are severals params added this way if you have look at couple lines above so it's more like a convention  rather than a code bloat. Because a clear benefit is that we are not introducing any unnecessary untested branches and `undefined` types that optional chaining causes e.g `(await request.getSimulationConfig?.(url)) || {};`

packages/transaction-controller/src/api/simulation-api.ts
OGPoyraz: Just to clarify for my own understanding;
`getSimulationConfig` is generating proxy URL for simulation requests. And in case if `getSimulationConfig` can't create proxy url - we want to cut off this simulation feature as we couldn't generate proxy address?

matthiasgeihs: How this will work:
If the user is subscribed to Shield, this request will be routed through a proxy that monitors and stores all requests and their responses for later reference.

We can handle failures within `getSimulationConfig` gracefully and potentially fallback to the default. But if `getSimulationConfig` fails completely, then we probably also don't want to run simulation.

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: > Not a mandatory parameter but this will make this.#getSimulationConfig to be typed and defined all the time.

Yes, I understood that.

The issue I'm having is the following: If I make the parameter `getSimulationConfig` mandatory in `simulateTransactions`, I will now need to provide it everywhere I call this function. For example, `simulateTransactions` is also called from `simulateGas`, but there we currently don't require any proxying (as far as I know). We are mostly interested in recording requests that are related to balance changes.

I will clarify with the product owners what's the functional requirement here. If we need the proxy only for a subset of the calls, then I think having `getSimulationConfig` as a mandatory parameter on `simulateTransactions` is inconvenient. If we want the same `getSimulationConfig` to be used across all calls, then I have less of a problem with making the parameter mandatory.

matthiasgeihs: Thinking about this more: I can see that from a general viewpoint, using `getSimulationConfig` only for a subset of the `simulateTransactions` calls is probably not what we want. I'll apply it consistently and will double check if that meets the functional requirements.

packages/transaction-controller/src/api/simulation-api.ts
OGPoyraz: Thanks for taking time for explanation Matthias!

OGPoyraz (dismissed):

packages/transaction-controller/src/TransactionController.ts
matthiasgeihs: done now in 12739154fc5911cad40e67cb0c62d2e85ebbcf8b

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -748,10 +749,11 @@ async function prepareApprovalData({
     messenger,
     request: userRequest,
     isSimulationEnabled,
+    getChainId,
+    getEthQuery,
     getGasFeeEstimates,
+    getSimulationConfig,
     update,
-    getEthQuery,
-    getChainId,
```
OGPoyraz, line 754: ❤️

matthiasgeihs:
@metamaskbot publish-preview

PR #6283, opened 2025-08-11
Seedless onboarding: Set anonymous='false' for sensitive fields

matthiasgeihs:
## Explanation

Updating the state metadata to use `anonymous='false'` for sensitive fields.

## Checklist

### DIFF
--- a/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
+++ b/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
@@ -75,11 +75,11 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
     },
     socialBackupsMetadata: {
       persist: true,
-      anonymous: true,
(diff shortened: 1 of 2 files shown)


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -95,54 +95,54 @@ const seedlessOnboardingMetadata: StateMetadata<SeedlessOnboardingControllerStat
     },
     userId: {
       persist: true,
-      anonymous: true,
+      anonymous: false,
```
lwin-kyaw, line 98: I think we can do the same for `groupedAuthConnectionId` and `authConnectionId` too

matthiasgeihs: not necessary as discussed on call

himanshuchawla009 (approved):
lgtm

PR #6285, opened 2025-08-12
feat(assets): implement balance change calculator and network filtering

amitabh94:
## Explanation

The PR introduces change to remove the centralized selectors and instead places a centralized calculation logic that clients can use to create their own selectors. 

- Current: aggregated percent/amount change for balances lives in client selectors with platform-specific nuances.
- Change: introduce a pure calculator to centralize portfolio change computation and move the logic out of selectors. Callers (mobile/extension) can now request 1d/7d/30d change consistently.
- Details:
  - EVM: uses TokenRatesController.marketData price and pricePercentChange{1d|7d|30d}, converted via CurrencyRateController.
  - Non‑EVM: uses MultichainAssetsRatesController.conversionRates with marketData.pricePercentChange.{P1D|P7D|P30D}.
  - Previous totals reconstructed via current / (1 + percent/100); amount and percent change computed from that.
  - Honors enabled-network filtering map (same semantics as totals calculation).

## References

- Internal: ASSETS-1129
- Package: @metamask/assets-controllers
- Related PR: 
                Mobile : https://github.com/MetaMask/metamask-mobile/pull/18315
                Extension : https://github.com/MetaMask/metamask-extension/pull/35012**
## Checklist

- [ ] I've prepared draft pull requests for clients and consumer packages to adopt the new calculator


packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,506 @@
...
+                  : undefined;
+
+              if (
+                typeof nativeToUserRate !== 'number' ||
+                Number.isNaN(nativeToUserRate)
+              ) {
+                continue;
+              }
+
+              if (typeof percentRaw !== 'number' || Number.isNaN(percentRaw)) {
```
darkwing: This type and `isNan` check is used in a handful of places in this file; could this be a helper function?

amitabh94: I have extracted this into a helper function

amitabh94:
@metamaskbot publish-preview

amitabh94:
@metamaskbot publish-preview

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+
+const isChainEnabledByMap = (
+  map: Record<string, Record<string, boolean>> | undefined,
+  id: Hex | CaipChainId,
+): boolean => {
+  if (!map) {
+    return true;
+  }
+  const isHex = typeof id === 'string' && id.startsWith('0x');
+  if (isHex) {
```
Prithpal-Sooriya: Can we use the typed util `isStrictHexString` from `@metamask/utils`?
This should discriminate between hex and CAIP and we can then remove the type assertions.

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+};
+
+const getInternalAccountsForGroup = (
+  accountTreeState: AccountTreeControllerState,
+  accountsState: AccountsControllerState,
+  groupId: string,
+) => {
+  const walletId = groupId.split('/')[0] as EntropySourceId;
+  const wallet = (
+    accountTreeState.accountTree.wallets as Record<string, AccountWalletObject>
```
Prithpal-Sooriya: surprised there is no utility for splitting and grabbing this data. Lets keep as is for now.

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+  }
+  const group = (wallet.groups as Record<string, AccountGroupObject>)[groupId];
+  if (!group) {
+    return [] as InternalAccount[];
+  }
+  return group.accounts
+    .map(
+      (accountId: string) => accountsState.internalAccounts.accounts[accountId],
+    )
+    .filter(Boolean) as InternalAccount[];
```
Prithpal-Sooriya: Can we provide a return type to this function? This was we don't need to include all of these assertions.

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+          groupId,
+          totalBalanceInUserCurrency: 0,
+          userCurrency: currencyRateState.currentCurrency,
+        };
+        continue;
+      }
+
+      let groupTotalBalance = 0;
+
+      for (const account of accounts) {
```
Prithpal-Sooriya: holy nesting.... this feels really complex.
TBH we are dealing with very complex data.

Is there methods or ways we can flatten this out?
If not flatten immediately, can we break it down into smaller parts?

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+  accountTreeState: AccountTreeControllerState,
+  accountsState: AccountsControllerState,
+  tokenBalancesState: TokenBalancesControllerState,
+  tokenRatesState: TokenRatesControllerState,
+  multichainRatesState: MultichainAssetsRatesControllerState,
+  multichainBalancesState: MultichainBalancesControllerState,
+  tokensState: TokensControllerState,
+  currencyRateState: CurrencyRateState,
+  enabledNetworkMap: Record<string, Record<string, boolean>> | undefined,
+): AllWalletsBalance {
```
Prithpal-Sooriya, line 311: This scares me...
Ideally I would like our selectors to be as narrow as possible.

In this scenario, this might be as narrow are we can make it (its balances for all addresses). But yikes this will be incredibly volatile.

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,695 @@
...
+      userCurrency: currencyRateState.currentCurrency,
+    };
+    totalBalanceInUserCurrency += walletTotalBalance;
+  }
+
+  return {
+    wallets: walletBalances,
+    totalBalanceInUserCurrency,
+    userCurrency: currencyRateState.currentCurrency,
+  };
```
Prithpal-Sooriya: Hmmm, this feels like it is handling multiple responsibilities.
Do we need this 1 selector to return walletBalances AND totalBalanceInUserCurrency? This feels like it should be 2 separate selectors.

packages/assets-controllers/src/selectors/balanceSelectors.ts
Prithpal-Sooriya, line 1: team discuss - I kinda liked the idea of selectors being isolated in a folder.
We have too many things in the root level.

packages/assets-controllers/src/balances.ts
amitabh94: Totally agree on keeping “narrow.” We’ve removed selectors and centralized the logic as pure calculators, so consumers can build their own narrow selectors on top without coupling to app state shape

packages/assets-controllers/src/balances.ts
amitabh94: I have refactored the function to be easily readable.

packages/assets-controllers/src/selectors/balanceSelectors.ts
amitabh94: I couldn't point a finger to where I could place these core calculators and hence placed them at root. If there is a better place , happy to move this calculation there.

packages/assets-controllers/src/balances.ts
amitabh94: We removed selectors; the pure calculator returns a structured result with both wallets and the total. The clients can choose whether they want to use this total or not. Although they don't need it right now , but in the future if they want to display total user balance across all wallets then they know where to find it. 

We can have this as a separate function derived from this function but I feel this should be fine.

packages/assets-controllers/src/balances.ts
amitabh94: I have added a small wrapper util for this

packages/assets-controllers/src/balances.ts
amitabh94: I have used the util from @metamask/utils. Thanks!

amitabh94:
@metamaskbot publish-preview

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+  if (!map) {
+    return true;
+  }
+  const isHex = isStrictHexString(id as string);
+  if (isHex) {
+    const evm = map[String(KnownCaipNamespace.Eip155)];
+    return Boolean(evm?.[id as Hex]);
+  }
+  const { namespace } = parseCaipChainId(id as CaipChainId);
+  return Boolean(map[namespace]?.[id as CaipChainId]);
```
Prithpal-Sooriya: Better implicit types and removes assertions.

```suggestion
  if (isStrictHexString(id)) {
    return Boolean(map[KnownCaipNamespace.Eip155]?.[id]);
  }
  const { namespace } = parseCaipChainId(id);
  return Boolean(map[namespace]?.[id]);
```

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+};
+
+const getWalletIdFromGroupId = (groupId: string): EntropySourceId => {
+  return groupId.split('/')[0] as EntropySourceId;
+};
```
Prithpal-Sooriya: Stricter type

```suggestion
const getWalletIdFromGroupId = (groupId: string): AccountWalletId => {
  return groupId.split('/')[0] as AccountWalletId;
};
```

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+
+const getInternalAccountsForGroup = (
+  accountTreeState: AccountTreeControllerState,
+  accountsState: AccountsControllerState,
+  groupId: string,
+): InternalAccount[] => {
+  const walletId = getWalletIdFromGroupId(groupId);
+  const wallet = (
+    accountTreeState.accountTree.wallets as Record<string, AccountWalletObject>
+  )[walletId];
```
Prithpal-Sooriya: Remove assertion, not needed now.

```suggestion
  const wallet = accountTreeState.accountTree.wallets[walletId];
```

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+  if (!wallet) {
+    return [];
+  }
+  const group = (wallet.groups as Record<string, AccountGroupObject>)[groupId];
```
Prithpal-Sooriya: Type assert the index than the underlying data structure.

```suggestion
  const group = wallet.groups[groupId as AccountGroupId];
```

Prithpal-Sooriya: Ideally we would not need to assert here, maybe we can make the parameter the correct type?

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+  currencyRateState: CurrencyRateState,
+  isEvmChainEnabled: (chainId: Hex) => boolean,
+): number {
+  let total = 0;
+  const accountBalances =
+    tokenBalancesState.tokenBalances[account.address as Hex];
+  if (!accountBalances) {
+    return 0;
+  }
+  for (const [chainId, chainBalances] of Object.entries(accountBalances)) {
```
Prithpal-Sooriya: Still feels incredibly dense... I'll try to cook something

Prithpal-Sooriya: Thoughts on this?

```ts
/**
 * Sum EVM account token balances in user currency.
 *
 * @param account - Internal account.
 * @param tokenBalancesState - Token balances state.
 * @param tokensState - Tokens state.
 * @param tokenRatesState - Token rates state.
 * @param currencyRateState - Currency rate state.
 * @param isEvmChainEnabled - Predicate to check EVM chain enablement.
 * @returns Total value in user currency.
 */
function sumEvmAccountBalanceInUserCurrency(
  account: InternalAccount,
  tokenBalancesState: TokenBalancesControllerState,
  tokensState: TokensControllerState,
  tokenRatesState: TokenRatesControllerState,
  currencyRateState: CurrencyRateState,
  isEvmChainEnabled: (chainId: Hex) => boolean,
): number {
  const getValidTokenBalances = () => {
    const accountBalances =
      tokenBalancesState.tokenBalances[account.address as Hex] ?? {};
    const tokenBalances = Object.entries(accountBalances)
      .filter(([chainId]) => isEvmChainEnabled(chainId as Hex))
      .flatMap(([chainId, chainBalances]) =>
        Object.entries(chainBalances).map(([tokenAddress, balance]) => ({
          chainId: chainId as Hex,
          tokenAddress: tokenAddress as Hex,
          balance,
        })),
      );
    return tokenBalances;
  };

  const calcTokenValue = (
    tokenBalance: ReturnType<typeof getValidTokenBalances>[number],
  ) => {
    const { chainId, tokenAddress, balance } = tokenBalance;

    // Get Token Info
    const chainTokens = tokensState.allTokens[chainId];
    const accountTokens = chainTokens?.[account.address];
    const token = accountTokens?.find((t) => t.address === tokenAddress);
    if (!token) {
      return 0;
    }

    // Get market data
    const chainMarketData = tokenRatesState.marketData[chainId];
    const tokenMarketData = chainMarketData?.[tokenAddress as Hex];
    if (!tokenMarketData?.price) {
      return 0;
    }

    // Get conversion rate
    const nativeToUserRate =
      currencyRateState.currencyRates[tokenMarketData.currency]?.conversionRate;
    if (!nativeToUserRate) {
      return 0;
    }

    // Calculate value
    const decimals = isNonNaNNumber(token.decimals) ? token.decimals : 18;
    const balanceInSmallestUnit = parseInt(balance, 16);
    if (!isNonNaNNumber(balanceInSmallestUnit)) {
      return 0;
    }

    return (
      (balanceInSmallestUnit / Math.pow(10, decimals)) *
      tokenMarketData.price *
      nativeToUserRate
    );
  };

  const tokenBalances = getValidTokenBalances();
  const total = tokenBalances.reduce((a, b) => a + calcTokenValue(b), 0);
  return total;
}
```

- Decoupled filtering from calculating
- Flattened DS, easier to loop through.
- use returns instead of for loops and continues

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+/**
+ * Sum non‑EVM account balances in user currency from multichain sources.
+ *
+ * @param account - Internal account.
+ * @param multichainBalancesState - Multichain balances state.
+ * @param multichainRatesState - Multichain rates state.
+ * @param isAssetChainEnabled - Predicate to check asset chain enablement.
+ * @returns Total value in user currency.
+ */
+function sumNonEvmAccountBalanceInUserCurrency(
```
Prithpal-Sooriya, line 270: man, non-evm tokens are nice and easy to calculate 🙈

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+          groupId,
+          totalBalanceInUserCurrency: 0,
+          userCurrency: currencyRateState.currentCurrency,
+        };
+        continue;
+      }
+
+      let groupTotalBalance = 0;
+
+      for (const account of accounts) {
```
Prithpal-Sooriya: We still have 3 layers of nesting here, I wonder if we can traverse through a flattened structure. I'll try cooking something.

Prithpal-Sooriya: <details><summary>Similar cleanup</summary>

```ts

/**
 * Calculate balances for all wallets and groups.
 * Pure function – accepts controller states and returns aggregated totals.
 *
 * @param accountTreeState - AccountTreeController state
 * @param accountsState - AccountsController state
 * @param tokenBalancesState - TokenBalancesController state
 * @param tokenRatesState - TokenRatesController state
 * @param multichainRatesState - MultichainAssetsRatesController state
 * @param multichainBalancesState - MultichainBalancesController state
 * @param tokensState - TokensController state
 * @param currencyRateState - CurrencyRateController state
 * @param enabledNetworkMap - Map of enabled networks keyed by namespace
 * @returns Aggregated balances for all wallets
 */
export function calculateBalanceForAllWallets(
  accountTreeState: AccountTreeControllerState,
  accountsState: AccountsControllerState,
  tokenBalancesState: TokenBalancesControllerState,
  tokenRatesState: TokenRatesControllerState,
  multichainRatesState: MultichainAssetsRatesControllerState,
  multichainBalancesState: MultichainBalancesControllerState,
  tokensState: TokensControllerState,
  currencyRateState: CurrencyRateState,
  enabledNetworkMap: Record<string, Record<string, boolean>> | undefined,
): AllWalletsBalance {
  const isEvmChainEnabled = (chainId: Hex): boolean =>
    isChainEnabledByMap(enabledNetworkMap, chainId);

  const isAssetChainEnabled = (assetId: CaipAssetType): boolean =>
    isChainEnabledByMap(enabledNetworkMap, parseCaipAssetType(assetId).chainId);

  const getBalance = {
    evm: (account: InternalAccount) =>
      sumEvmAccountBalanceInUserCurrency(
        account,
        tokenBalancesState,
        tokensState,
        tokenRatesState,
        currencyRateState,
        isEvmChainEnabled,
      ),
    nonEvm: (account: InternalAccount) =>
      sumNonEvmAccountBalanceInUserCurrency(
        account,
        multichainBalancesState,
        multichainRatesState,
        isAssetChainEnabled,
      ),
  };

  const getFlatAccountBalances = () =>
    Object.entries(accountTreeState.accountTree.wallets ?? {})
      .flatMap(([walletId, wallet]) =>
        Object.keys(wallet?.groups || {}).flatMap((groupId) => {
          const accounts = getInternalAccountsForGroup(
            accountTreeState,
            accountsState,
            groupId,
          );

          return accounts.map((account) => ({
            walletId,
            groupId,
            account,
            isEvm: isEvmAccountType(account.type),
          }));
        }),
      )
      .map((flatAccount) => {
        const flatAccountWithBalance = flatAccount as typeof flatAccount & {
          balance: number;
        };
        flatAccountWithBalance.balance = flatAccount.isEvm
          ? getBalance.evm(flatAccount.account)
          : getBalance.nonEvm(flatAccount.account);
        return flatAccountWithBalance;
      });

  const getAggWalletBalance = (
    flatAccountBalances: ReturnType<typeof getFlatAccountBalances>,
  ): number => flatAccountBalances.reduce((a, b) => a + b.balance, 0);

  const getWalletBalances = (
    flatAccountBalances: ReturnType<typeof getFlatAccountBalances>,
  ): Record<string, WalletBalance> => {
    const wallets: Record<string, WalletBalance> = {};
    const defaultWalletBalance = (walletId: string): WalletBalance => ({
      walletId,
      groups: {},
      totalBalanceInUserCurrency: 0,
      userCurrency: currencyRateState.currentCurrency,
    });
    const defaultGroupBalance = (
      walletId: string,
      groupId: string,
    ): AccountGroupBalance => ({
      walletId,
      groupId,
      totalBalanceInUserCurrency: 0,
      userCurrency: currencyRateState.currentCurrency,
    });

    flatAccountBalances.forEach((flatAccount) => {
      const { walletId, groupId, balance } = flatAccount;
      wallets[walletId] ??= defaultWalletBalance(walletId);
      wallets[walletId].groups[groupId] ??= defaultGroupBalance(
        walletId,
        groupId,
      );
      wallets[walletId].groups[groupId].totalBalanceInUserCurrency += balance;
      wallets[walletId].totalBalanceInUserCurrency += balance;
    });

    // Ensure all groups (including empty ones) are represented
    Object.entries(accountTreeState.accountTree.wallets ?? {}).forEach(
      ([walletId, wallet]) => {
        if (!wallet) {
          return;
        }
        wallets[walletId] ??= defaultWalletBalance(walletId);
        Object.keys(wallet.groups || {}).forEach((groupId) => {
          wallets[walletId].groups[groupId] ??= defaultGroupBalance(
            walletId,
            groupId,
          );
        });
      },
    );

    return wallets;
  };

  const flatAccounts = getFlatAccountBalances();
  return {
    wallets: getWalletBalances(flatAccounts),
    totalBalanceInUserCurrency: getAggWalletBalance(flatAccounts),
    userCurrency: currencyRateState.currentCurrency,
  };
}
```

</details> 

- flattened DS, easier to loop through.
- split logic for the different pieces

Prithpal-Sooriya: Hmm I still feel like this util does "too much", it is aggregating wallet balance, group balance, and returning a nested obj.

Can this not just return an array of group balances? The UI can be responsible for performing a .reduce/sum over it to get wallet balance.

packages/assets-controllers/src/balances.ts
Prithpal-Sooriya: Commented a cleanup here:
https://github.com/MetaMask/core/pull/6285/files#r2277578982

It still feels like this util does too much. It even returns a deep nested DS. Is this really needed? Can't we just return an array of walletId + groupId + balance.

packages/assets-controllers/src/balances.ts
amitabh94: Thanks for this. I had considered various approaches of flattening and otherwise and somehow ended with this .Let me have a look as to  why I did it, otherwise I don't see any reason not to flatten it. 

> Hmm I still feel like this util does "too much", it is aggregating wallet balance, group balance, and returning a nested obj.
> 
> Can this not just return an array of group balances? The UI can be responsible for performing a .reduce/sum over it to get wallet balance.

we can do this or continue doing what we are doing now and have derived selectors from that one selector. I wanted to keep as much logic centralized as possible.

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,800 @@
...
+ * @param tokenBalancesState - TokenBalancesController state
+ * @param tokenRatesState - TokenRatesController state
+ * @param multichainRatesState - MultichainAssetsRatesController state
+ * @param multichainBalancesState - MultichainBalancesController state
+ * @param tokensState - TokensController state
+ * @param currencyRateState - CurrencyRateController state
+ * @param enabledNetworkMap - Map of enabled networks keyed by namespace
+ * @returns Aggregated balances for all wallets
+ */
+export function calculateBalanceForAllWallets(
```
salimtb, line 301: is there a reason why we don't use createSelector ?

packages/assets-controllers/src/balances.ts
```diff
@@ -0,0 +1,1103 @@
...
+ * @param tokenRatesState - TokenRatesController state.
+ * @param multichainRatesState - MultichainAssetsRatesController state.
+ * @param multichainBalancesState - MultichainBalancesController state.
+ * @param tokensState - TokensController state.
+ * @param currencyRateState - CurrencyRateController state.
+ * @param enabledNetworkMap - Map of enabled networks keyed by namespace.
+ * @param period - Period to compute change for ('1d' | '7d' | '30d').
+ * @returns Aggregated change details for the requested period.
+ */
+export function calculateBalanceChangeForAllWallets(
```
Prithpal-Sooriya, line 446: Heya, did some cleanup on these functions in this diff
https://github.com/MetaMask/core/pull/6319

Feel free to look and pick through the commits (based off of this branch)

packages/assets-controllers/src/balances.ts
Prithpal-Sooriya: Yep that is fair, only once we integrated and feel like it needs decoupling should we do it.

packages/assets-controllers/src/balances.ts
amitabh94: This is just a function for core logic and we have the selectors in clients

amitabh94:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #6287, opened 2025-08-12
feat: add isNetworkEnabled function

salimtb:
## Explanation

The NetworkEnablementController was missing a public method to check whether a specific network is enabled. This functionality is needed by consumers to query the enablement state of networks without having to access the internal state structure directly.
The solution adds the isNetworkEnabled method which:

- Accepts both Hex chain IDs (e.g., '0x1') and CAIP chain IDs (e.g., 'eip155:1')
- Returns a boolean indicating whether the network is enabled
- Gracefully handles edge cases like non-existent networks or namespaces by returning false
- Uses the existing deriveKeys utility to normalize different chain ID formats
- Leverages optional chaining to prevent errors when accessing undefined namespace properties

Additionally, this PR adds Palm network and HypeEVM network to the list of popular networks, expanding the supported network ecosystem.
The method includes comprehensive test coverage to ensure reliability across different blockchain namespaces (EVM, Solana, etc.) and various edge cases including invalid inputs and non-existent networks.

## Checklist

### DIFF
--- a/packages/network-enablement-controller/CHANGELOG.md
+++ b/packages/network-enablement-controller/CHANGELOG.md
@@ -7,17 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
+### Added
 
-- add `isNetworkEnabled` method to handle non-existent namespaces gracefully without throwing errors
+- add `isNetworkEnabled` method to check if network is enabled ([#6287](https://github.com/MetaMask/core/pull/6287))
+- add `Palm network` and `HypeEVM` network to list of popular network ([#6287](https://github.com/MetaMask/core/pull/6287))
 
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.0.1` to `^8.1.0` ([#6284](https://github.com/MetaMask/core/pull/6284))
-
-### Added
-
-- Comprehensive test coverage for `isNetworkEnabled` method including edge cases and multi-chain scenarios
 
 ## [0.1.0]
 


packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -247,6 +247,20 @@ export class NetworkEnablementController extends BaseController<
...
+   * @param chainId - The chain ID of the network to check. Can be either:
+   * - A Hex string (e.g., '0x1' for Ethereum mainnet) for EVM networks
+   * - A CAIP-2 chain ID (e.g., 'eip155:1' for Ethereum mainnet, 'solana:mainnet' for Solana)
+   * @returns True if the network is enabled, false otherwise
+   */
+  isNetworkEnabled(chainId: Hex | CaipChainId): boolean {
+    const derivedKeys = deriveKeys(chainId);
+    const { namespace, storageKey } = derivedKeys;
+    return this.state.enabledNetworkMap[namespace]?.[storageKey] ?? false;
+  }
```
Prithpal-Sooriya, line 262: Nice!
Technically this doesn't do any state manipulation stuff, could be a util or selector?

packages/network-enablement-controller/src/constants.ts
```diff
@@ -9,4 +9,6 @@ export const POPULAR_NETWORKS = [
   '0x89', // Polygon (137)
   '0x531', // Sei (Assuming 1329 used in EVM context)
   '0x144', // zkSync Era (324)
+  '0x2a15c308d', // Palm (11297108109)
+  '0x3e7', // HyperEVM (999)
```
Prithpal-Sooriya, line 13: Okie, nice. As you mentioned, we can have this list passed into the controller through a param, this way clients (if need to) can diverge their popular networks.

packages/network-enablement-controller/src/NetworkEnablementController.ts
Prithpal-Sooriya: Not a blocker, we can refactor out if this controller gets really large.

PR #6288, opened 2025-08-12
feat: social login authentication state

lwin-kyaw:
## Explanation

This PR adds new persisted state value, `isSeedlessOnboardingUserAuthenticated` to determine whether the user has successfully finished social login and TOPRF Authentication.
This is mainly for the UI state in the clients (extension/mobile) and to avoid querying the sensitive controller state data from the client side for determining the social login authentication state.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "496.0.0",
+  "version": "497.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/assets-controllers/package.json
+++ b/packages/assets-controllers/package.json
@@ -95,7 +95,7 @@
     "@metamask/preferences-controller": "^18.4.1",
     "@metamask/providers": "^22.1.0",
     "@metamask/snaps-controllers": "^14.0.1",
-    "@metamask/transaction-controller": "^59.1.0",
+    "@metamask/transaction-controller": "^59.2.0",
     "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.191",
     "@types/node": "^16.18.54",
--- a/packages/bridge-controller/package.json
+++ b/packages/bridge-controller/package.json
@@ -73,7 +73,7 @@
     "@metamask/remote-feature-flag-controller": "^1.7.0",
     "@metamask/snaps-controllers": "^14.0.1",
     "@metamask/superstruct": "^3.1.0",
-    "@metamask/transaction-controller": "^59.1.0",
+    "@metamask/transaction-controller": "^59.2.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
--- a/packages/bridge-status-controller/package.json
+++ b/packages/bridge-status-controller/package.json
@@ -63,7 +63,7 @@
     "@metamask/gas-fee-controller": "^24.0.0",
     "@metamask/network-controller": "^24.0.1",
     "@metamask/snaps-controllers": "^14.0.1",
-    "@metamask/transaction-controller": "^59.1.0",
+    "@metamask/transaction-controller": "^59.2.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
--- a/packages/controller-utils/CHANGELOG.md
+++ b/packages/controller-utils/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Update `onDegraded` in `createServicePolicy` so that its event payload now includes an `error` property which can be used to access the error produced by the last request when the maximum number of retries is exceeded ([#6188](https://github.com/MetaMask/core/pull/6188))
+  - This `error` property will be `undefined` if the degraded event merely represents a slow request
+
 ## [11.11.0]
 
 ### Added
--- a/packages/controller-utils/src/create-service-policy.ts
+++ b/packages/controller-utils/src/create-service-policy.ts
@@ -14,6 +14,7 @@ import {
 import type {
   CircuitBreakerPolicy,
   Event as CockatielEvent,
+  FailureReason,
   IBackoffFactory,
   IPolicy,
   Policy,
@@ -95,7 +96,7 @@ export type ServicePolicy = IPolicy & {
    * never succeeds before the retry policy gives up and before the maximum
    * number of consecutive failures has been reached.
    */
-  onDegraded: CockatielEvent<void>;
+  onDegraded: CockatielEvent<FailureReason<unknown> | void>;
   /**
    * A function which will be called by the retry policy each time the service
    * fails and the policy kicks off a timer to re-run the service. This is
@@ -229,10 +230,11 @@ export function createServicePolicy(
   });
   const onBreak = circuitBreakerPolicy.onBreak.bind(circuitBreakerPolicy);
 
-  const onDegradedEventEmitter = new CockatielEventEmitter<void>();
-  retryPolicy.onGiveUp(() => {
+  const onDegradedEventEmitter =
+    new CockatielEventEmitter<FailureReason<unknown> | void>();
+  retryPolicy.onGiveUp((data) => {
     if (circuitBreakerPolicy.state === CircuitState.Closed) {
-      onDegradedEventEmitter.emit();
+      onDegradedEventEmitter.emit(data);
     }
   });
   retryPolicy.onSuccess(({ duration }) => {
--- a/packages/earn-controller/package.json
+++ b/packages/earn-controller/package.json
@@ -58,7 +58,7 @@
     "@metamask/accounts-controller": "^32.0.2",
     "@metamask/auto-changelog": "^3.4.4",
     "@metamask/network-controller": "^24.0.1",
-    "@metamask/transaction-controller": "^59.1.0",
+    "@metamask/transaction-controller": "^59.2.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
--- a/packages/network-controller/CHANGELOG.md
+++ b/packages/network-controller/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The object in the `NetworkController:rpcEndpointDegraded` event payload now includes an `error` property, which can be used to access the error produced by the last request when the maximum number of retries is exceeded ([#6188](https://github.com/MetaMask/core/pull/6188))
+  - This `error` property will be `undefined` if the degraded event merely represents a slow request
+
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.0.1` to `^8.1.0` ([#6284](https://github.com/MetaMask/core/pull/6284))
--- a/packages/network-controller/src/NetworkController.ts
+++ b/packages/network-controller/src/NetworkController.ts
@@ -470,6 +470,7 @@ export type NetworkControllerRpcEndpointDegradedEvent = {
     {
       chainId: Hex;
       endpointUrl: string;
+      error: unknown;
     },
   ];
 };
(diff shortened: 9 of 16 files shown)


packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -57,6 +57,7 @@ const log = createModuleLogger(projectLogger, controllerName);
 export function getDefaultSeedlessOnboardingControllerState(): SeedlessOnboardingControllerState {
   return {
     socialBackupsMetadata: [],
+    isSeedlessOnboardingUserAuthenticated: false,
```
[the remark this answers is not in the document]
lwin-kyaw: Addressed it here, https://github.com/MetaMask/core/pull/6288/commits/1d75d1e03bd2190b5750b19cfbbfc4f053520cf8

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -318,19 +335,20 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
           state.userId = userId;
           state.authConnection = authConnection;
           state.socialLoginEmail = socialLoginEmail;
+          state.metadataAccessToken = metadataAccessToken;
+          state.accessToken = accessToken;
```
lwin-kyaw, line 338: `metadataAccessToken` and `accessToken` should be present in the state, in order to assert as an `authenticatedUser`. On calling `refreshAuthToken`, this method is called again and we have to make sure that we receive the valid data (i.e. with `metadataAccessToken` and `accessToken` from the server response data)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
```diff
@@ -669,6 +672,38 @@ describe('SeedlessOnboardingController', () => {
       );
     });
 
+    it('should be able to instantiate with as an authenticated user', () => {
```
matthiasgeihs:
```suggestion
    it('should be able to instantiate with an authenticated user', () => {
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -50,13 +51,28 @@ import { decodeJWTToken, decodeNodeAuthToken } from './utils';
...
-export function getDefaultSeedlessOnboardingControllerState(): SeedlessOnboardingControllerState {
+export function getInitialSeedlessOnboardingControllerStateWithDefaults(
+  overrides?: Partial<SeedlessOnboardingControllerState>,
+): SeedlessOnboardingControllerState {
+  let isSeedlessOnboardingUserAuthenticated =
+    overrides?.isSeedlessOnboardingUserAuthenticated ?? false;
+
+  try {
+    assertIsSeedlessOnboardingAuthenticatedUser(overrides);
+    isSeedlessOnboardingUserAuthenticated = true;
```
lwin-kyaw: In production, the controller can be re-initialized with `overrides state` many times, and we have to ensure that this state value, `isSeedlessOnboardingUserAuthenticated` is aligned with the provided state. 
We derive the `isSeedlessOnboardingUserAuthenticated` state value based on the `authentication data` from the provided state value so that we won't mistakenly set it as `AuthenticatedUser` if some of the authentication info are missing.

packages/seedless-onboarding-controller/src/assertions.ts
```diff
@@ -1,5 +1,53 @@
...
+    );
+  }
+  if (
+    !('metadataAccessToken' in value) ||
+    typeof value.metadataAccessToken !== 'string'
+  ) {
+    throw new Error(
+      SeedlessOnboardingControllerErrorMessage.InvalidMetadataAccessToken,
+    );
+  }
```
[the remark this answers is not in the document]
lwin-kyaw: `metadataAcessToken` is taken from the `getAuthToken` or `refreshAuthToken` response and if the value is missing, we can conclude it to the `AuthenticationError`.

For the debugging purpose, we add a log to check the actual error message in the controller, https://github.com/MetaMask/core/blob/7b9e9545cf22bbb16fe7c4845d196150733442f9/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts#L356

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -50,13 +51,28 @@ import { decodeJWTToken, decodeNodeAuthToken } from './utils';
...
+    isSeedlessOnboardingUserAuthenticated = false;
+  }
+
   return {
-    socialBackupsMetadata: [],
+    ...overrides,
+    socialBackupsMetadata: overrides?.socialBackupsMetadata ?? [],
+    isSeedlessOnboardingUserAuthenticated,
   };
 }
```
matthiasgeihs: I think you can simplify this to:
```ts
export function getInitialSeedlessOnboardingControllerState(
  overrides?: Partial<SeedlessOnboardingControllerState>,
): SeedlessOnboardingControllerState {
  const initialState = {
    socialBackupsMetadata: [],
    isSeedlessOnboardingUserAuthenticated: false,
    ...overrides,
  };

  // Ensure authenticated flag is set correctly.
  try {
    assertIsSeedlessOnboardingAuthenticatedUser(initialState);
    initialState.isSeedlessOnboardingUserAuthenticated = true;
  } catch {
    initialState.isSeedlessOnboardingUserAuthenticated = false;
  }
  return initialState;
}
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1676,42 +1695,13 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-    ) {
-      throw new Error(
-        SeedlessOnboardingControllerErrorMessage.InvalidMetadataAccessToken,
-      );
+    try {
+      assertIsSeedlessOnboardingAuthenticatedUser(value);
+    } catch (error) {
+      this.update((state) => {
+        state.isSeedlessOnboardingUserAuthenticated = false;
+      });
```
matthiasgeihs, line 1702: I'm torn about this one.
On the upside, this ensures that the state is correct.
On the downside, this updates the state at a place where we wouldn't expect state updates. (I think we should generally try to minimize the number of state updates.)

packages/seedless-onboarding-controller/src/index.ts
```diff
@@ -1,6 +1,6 @@
 export {
   SeedlessOnboardingController,
-  getDefaultSeedlessOnboardingControllerState,
+  getInitialSeedlessOnboardingControllerStateWithDefaults,
```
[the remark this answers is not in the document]
matthiasgeihs: This comment is correct.

Why are we even exposing this method? Is this used somewhere?
Should we maybe remove it or export a `getDefaultSeedlessOnboardingControllerState` as before?

packages/seedless-onboarding-controller/src/assertions.ts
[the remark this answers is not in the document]
matthiasgeihs: @lwin-kyaw, can you respond to this comment?

packages/seedless-onboarding-controller/src/assertions.ts
```diff
@@ -1,5 +1,53 @@
...
+}
```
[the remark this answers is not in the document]
lwin-kyaw: `accessToken` is only available when the wallet is unlocked or inside the encrypted vault, unlike others variables in the assertion.

`assertIsSeedlessOnboardingAuthenticatedUser` can be called even when the wallet is unlocked, for e.g. `passwordOutdatedCheck` and `submitGlobalPassword` where the wallet/vault is still locked.

That's why we left out the `acessToken assertion` in `assertIsSeedlessOnboardingAuthenticatedUser` coz it's not needed for TOPRF API calls.

cc: @matthiasgeihs

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Thanks, this is much simpler implementation.
Updated here, https://github.com/MetaMask/core/pull/6288/commits/ef6dc19e638bad545292d2ae5b0ba58417262bcd

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
lwin-kyaw: Addressed here,  https://github.com/MetaMask/core/pull/6288/commits/ef6dc19e638bad545292d2ae5b0ba58417262bcd

lwin-kyaw:
> @lwin-kyaw, can you respond to this comment?
Replied here, https://github.com/MetaMask/core/pull/6288#discussion_r2269757886

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.test.ts
matthiasgeihs: implemented

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: btw, i also shortened the function name dropping the `withDefaults` suffix.
not really short, but shorter.

packages/seedless-onboarding-controller/src/index.ts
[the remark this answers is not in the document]
lwin-kyaw: Mobile is using this, `getDefaultSeedlessOnboardingControllerState` function.

> export a getDefaultSeedlessOnboardingControllerState as before?

I thought of that, but I think with new implementation, it does more than loading the static default value.
Wdyt?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: I think it would be an appropriate place to update this state coz the state, `isSeedlessOnboardingUserAuthenticated` is really tied to the `user authentication info`.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: understood

lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/assertions.ts
```diff
@@ -1,5 +1,53 @@
...
-import type { VaultData } from './types';
+import type { AuthenticatedUserDetails, VaultData } from './types';
+
+/**
+ * Check if the provided value is a valid authenticated user.
+ *
+ * @param value - The value to check.
+ * @throws If the value is not a valid authenticated user.
+ */
+export function assertIsSeedlessOnboardingAuthenticatedUser(
```
matthiasgeihs: nit
```suggestion
export function assertIsSeedlessOnboardingUserAuthenticated(
```

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: ok

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -318,19 +334,20 @@ export class SeedlessOnboardingController<EncryptionKey> extends BaseController<
...
-            state.accessToken = accessToken;
-          }
-          if (metadataAccessToken) {
-            state.metadataAccessToken = metadataAccessToken;
-          }
+
+          // we will check if the controller state is properly set with the authenticated user info
+          // before setting the isSeedlessOnboardingUserAuthenticated to true
+          assertIsSeedlessOnboardingAuthenticatedUser(state);
+          state.isSeedlessOnboardingUserAuthenticated = true;
```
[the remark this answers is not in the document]
matthiasgeihs: Part of this comment is wrong. `metadataAccessToken` is not optional.

However, `refreshToken` is indeed optional, but it is required by `assertIsSeedlessOnboardingAuthenticatedUser`, so this is a breaking change.

What do you think about making `refreshToken` a mandatory parameter of `authenticate`? @lwin-kyaw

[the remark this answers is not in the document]
matthiasgeihs: update:
after more reviewing, turns out that the second part of the bot's comment is also not quite true.
If `refreshToken` is not provided, then it is not replaced. If it was present before, it will remain present here.

matthiasgeihs (dismissed):

matthiasgeihs (approved):

PR #6291, opened 2025-08-12
Release 497.0.0

matthewwalsh0:
Minor release of `@metamask/transaction-controller`.

### DIFF
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Uncategorized
+
+- Release/492.0.0 ([#6273](https://github.com/MetaMask/core/pull/6273))
+
 ### Added
 
 - Add optional `updateType` property to disable `type` update in `updateEditableParams` method ([#6289](https://github.com/MetaMask/core/pull/6289))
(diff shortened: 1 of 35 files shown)


packages/transaction-controller/CHANGELOG.md
```diff
@@ -7,13 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+## [59.2.0]
+
 ### Added
 
 - Add optional `updateType` property to disable `type` update in `updateEditableParams` method ([#6289](https://github.com/MetaMask/core/pull/6289))
 - Add `perpsDeposit` to `TransactionType` ([#6282](https://github.com/MetaMask/core/pull/6282))
 
 ### Changed
 
+- Bump `@metamask/accounts-controller` from `^32.0.1` to `^32.0.2` ([#6273](https://github.com/MetaMask/core/pull/6273))
```
cryptodev-2s: Nit: this could be removed, it was just a devDependency bump ?
```suggestion

cryptodev-2s:
@matthewwalsh0 any reason why did you use the `no-changelog` label ?

cryptodev-2s:
> > @matthewwalsh0 any reasons why did you use the `no-changelog` label ?
> 
> We previously had to, in order to avoid the changelog validation job failing, I assume that's been fixed now?

Yes indeed the edge cases where it uses to fail where it shouldn't has been fixed.

PR #6295, opened 2025-08-12
feat: calculate effective gasFee for swap and bridge quotes

micaelae:
## Checklist

### DIFF
--- a/packages/bridge-controller/src/types.ts
+++ b/packages/bridge-controller/src/types.ts
@@ -124,7 +124,7 @@ export type QuoteMetadata = {
    * total is the gas fee that is spent by the user, including refunds.
    * max is the max gas fee that will be used by the transaction.
    */
-  gasFee: Record<'total' | 'max', TokenAmountValues>;
+  gasFee: Record<'effective' | 'total' | 'max', TokenAmountValues>;
   totalNetworkFee: TokenAmountValues; // estimatedGasFees + relayerFees
   totalMaxNetworkFee: TokenAmountValues; // maxGasFees + relayerFees
   /**
--- a/packages/bridge-controller/src/types.ts
+++ b/packages/bridge-controller/src/types.ts
@@ -118,7 +118,13 @@ export type QuoteMetadata = {
    * If gas is included, this is the value of the src or dest token that was used to pay for the gas
    */
   includedTxFees?: TokenAmountValues | null;
-  gasFee: TokenAmountValues;
+  /**
+   * The gas fee for the bridge transaction.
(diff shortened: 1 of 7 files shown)


packages/bridge-controller/src/types.ts
```diff
@@ -118,7 +118,13 @@ export type QuoteMetadata = {
    * If gas is included, this is the value of the src or dest token that was used to pay for the gas
    */
   includedTxFees?: TokenAmountValues | null;
-  gasFee: TokenAmountValues;
+  /**
+   * The gas fee for the bridge transaction.
+   * effective is the gas fee that is shown to the user.
+   * total is the gas fee that is spent by the user, including refunds.
```
SteP-n-s, line 125: Won't `total` have the same value as `effective` since the user will need to see the total amount they'll pay i.e. `(baseFee + maxPriorityFeePerGas) * effectiveGas`

micaelae: `total` is the amount they need to execute the transaction. `effective` is the amount that actually gets spent after the user receives refunds. We'll only show the effective amount, but will use the max amount to determine whether the user has sufficient gas funds

SteP-n-s (approved):

infiniteflower:
> clients should, display the totalNetworkFee, but use the totalMaxNetworkFee value to determine whether the user has sufficient funds for gas

I remember we've had issues in the past with this approach, where people complain that the number we display for the estimated gas and the amount we block them on is not the same.

It would be helpful to find out what the other teams are doing about this, like Send.

micaelae:
@infiniteflower that's a good callout. With Sam's latest change that makes the "high" (max) network fee closer to the base fee, the discrepancy will be much lower than it used to be. Iirc the scenario you identified is why we used to show a range of network fees

Either way both values will be returned as quote metadata and the clients can use the value that makes sense

packages/bridge-controller/src/utils/quote.ts
```diff
@@ -174,23 +175,25 @@ export const calcRelayerFee = (
...
+  l1GasFeesInHexWei,
   feePerGasInDecGwei,
   priorityFeePerGasInDecGwei,
   nativeToDisplayCurrencyExchangeRate,
   nativeToUsdExchangeRate,
 }: {
-  bridgeQuote: QuoteResponse & L1GasFees;
+  approvalGasLimit?: number | null;
+  tradeGasLimit?: number | null;
+  l1GasFeesInHexWei?: string | null;
```
infiniteflower, line 188: Nice I like how it's less coupled to the quote, easier to test.

PR #6300, opened 2025-08-13
feat: subscription controller handle create with card

tuna1207:
## Explanation

Handle subscription controller start subscription with card

## Checklist

### DIFF
--- a/packages/subscription-controller/src/SubscriptionController.ts
+++ b/packages/subscription-controller/src/SubscriptionController.ts
@@ -120,10 +120,10 @@ export class SubscriptionController extends BaseController<
     await this.#subscriptionService.cancelSubscription({ subscriptionId });
   }
 
-  async startShieldSubscription() {
+  async startShieldSubscriptionWithCard() {
     this.#assertIsUserNotSubscribed();
 
-    return await this.#subscriptionService.startSubscription({
+    return await this.#subscriptionService.startSubscriptionWithCard({
       products: [SUBSCRIPTION_PRODUCTS.SHIELD],
       isTrialRequested: true,
     });
--- a/packages/subscription-controller/src/SubscriptionController.ts
+++ b/packages/subscription-controller/src/SubscriptionController.ts
@@ -3,6 +3,7 @@ import { BaseController, type StateMetadata } from '@metamask/base-controller';
 import {
   controllerName,
   Env,
+  SUBSCRIPTION_PRODUCTS,
   SubscriptionControllerErrorMessage,
 } from './constants';
 // import { projectLogger, createModuleLogger } from './logger';
@@ -119,11 +120,20 @@ export class SubscriptionController extends BaseController<
     await this.#subscriptionService.cancelSubscription({ subscriptionId });
   }
 
-  // #assertIsUserNotSubscribed() {
-  //   if (this.state.subscription) {
-  //     throw new Error(SubscriptionControllerErrorMessage.UserAlreadySubscribed);
-  //   }
-  // }
+  async startShieldSubscription() {
+    this.#assertIsUserNotSubscribed();
+
+    return await this.#subscriptionService.startSubscription({
+      products: [SUBSCRIPTION_PRODUCTS.SHIELD],
+      isTrialRequested: true,
+    });
+  }
+
+  #assertIsUserNotSubscribed() {
+    if (this.state.subscription) {
+      throw new Error(SubscriptionControllerErrorMessage.UserAlreadySubscribed);
+    }
+  }
 
   #assertIsUserSubscribed() {
     if (!this.state.subscription) {
--- a/packages/subscription-controller/src/SubscriptionService.test.ts
+++ b/packages/subscription-controller/src/SubscriptionService.test.ts
@@ -456,7 +456,7 @@ describe('SubscriptionService', () => {
         .post('/api/v1/subscriptions/card', MOCK_START_SUBSCRIPTION_REQUEST)
         .reply(200, MOCK_START_SUBSCRIPTION_RESPONSE);
 
-      const result = await service.startSubscription(MOCK_START_SUBSCRIPTION_REQUEST);
+      const result = await service.startSubscriptionWithCard(MOCK_START_SUBSCRIPTION_REQUEST);
 
       expect(result).toStrictEqual(MOCK_START_SUBSCRIPTION_RESPONSE);
       expect(config.auth.getAccessToken).toHaveBeenCalledTimes(1);
@@ -475,7 +475,7 @@ describe('SubscriptionService', () => {
         .post('/api/v1/subscriptions/card', request)
         .reply(200, MOCK_START_SUBSCRIPTION_RESPONSE);
 
-      const result = await service.startSubscription(request);
+      const result = await service.startSubscriptionWithCard(request);
 
       expect(result).toStrictEqual(MOCK_START_SUBSCRIPTION_RESPONSE);
     });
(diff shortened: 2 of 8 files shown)


packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -119,11 +120,20 @@ export class SubscriptionController extends BaseController<
     await this.#subscriptionService.cancelSubscription({ subscriptionId });
   }
 
-  // #assertIsUserNotSubscribed() {
-  //   if (this.state.subscription) {
-  //     throw new Error(SubscriptionControllerErrorMessage.UserAlreadySubscribed);
-  //   }
-  // }
+  async startShieldSubscriptionWithCard() {
```
matthiasgeihs: We should design the controller in a way that it supports subscriptions to different products.

```suggestion
  async startSubscriptionWithCard(product: string) {
```

chaitanyapotti (approved):

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -24,7 +30,9 @@ type CreateActionsObj<Controller extends keyof SubscriptionController> = {
     handler: SubscriptionController[K];
   };
 };
-type ActionsObj = CreateActionsObj<'getSubscriptions' | 'cancelSubscription'>;
+type ActionsObj = CreateActionsObj<
+  'getSubscriptions' | 'cancelSubscription' | 'startShieldSubscriptionWithCard'
```
cryptodev-2s: We’d prefer not to use this pattern, since it can be easy to miss when a controller action is added or removed. These changes need to be explicitly listed in the changelog and also be exported as Controller Actions/Events, as they can sometimes be breaking. Could you please follow our [controller guidelines](https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#use-standard-naming-scheme-for-action-identifiers) instead?

I know this can feel a bit verbose, but we’re planning to introduce a new way of auto-generating these types in the future. For now, could you update these to follow the standard naming scheme (e.g., `SubscriptionControllerGetSubscriptionsAction`...) to keep things consistent?

cryptodev-2s (changes requested):

packages/subscription-controller/src/SubscriptionController.ts
cryptodev-2s: If you’d prefer, you could handle this in a follow-up PR before releasing the controller. I’d be fine approving in that case

tuna1207: I see, thanks @cryptodev-2s , i have exported action type according to the guidelines

packages/subscription-controller/src/types.ts
```diff
@@ -10,8 +10,13 @@ export type Product = {
 };
 
 export enum PaymentType {
-  CARD = 'card',
-  CRYPTO = 'crypto',
+  byCard = 'card',
+  byCrypto = 'crypto',
+}
+
+export enum RecurringInterval {
```
matthiasgeihs, line 17: Is `RecurringInterval` a common term? Maybe change to `PaymentInterval` or `BillingCycle`? (Stripe seems to be using the term "billing cycle". [Source](https://docs.stripe.com/billing/subscriptions/billing-cycle))

packages/subscription-controller/src/types.ts
```diff
@@ -23,29 +28,62 @@ export type PaymentMethod = {
...
-  status: 'active' | 'inactive' | 'trialing' | 'cancelled';
-  interval: 'month' | 'year';
+  status: SubscriptionStatus;
+  interval: RecurringInterval;
   paymentMethod: PaymentMethod;
 };
 
 export type GetSubscriptionsResponse = {
-  customerId: string;
+  customerId?: string;
```
matthiasgeihs, line 64: Hm, why is this optional now?

packages/subscription-controller/src/types.ts
```diff
@@ -23,29 +28,62 @@ export type PaymentMethod = {
...
+  checkoutSessionUrl: string;
+};
+
 export type AuthUtils = {
   getAccessToken: () => Promise<string>;
 };
 
 export type ISubscriptionService = {
   getSubscriptions(): Promise<GetSubscriptionsResponse>;
   cancelSubscription(request: { subscriptionId: string }): Promise<void>;
```
matthiasgeihs, line 85: Might want to add a dedicated request type here for consistency.

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -220,4 +239,53 @@ describe('SubscriptionService', () => {
...
+      nock(testUrl)
+        .post('/api/v1/subscriptions/card', request)
+        .reply(200, MOCK_START_SUBSCRIPTION_RESPONSE);
+
+      const result = await service.startSubscriptionWithCard(request);
+
+      expect(result).toStrictEqual(MOCK_START_SUBSCRIPTION_RESPONSE);
+    });
+
+    it('should handle empty products array', async () => {
```
matthiasgeihs:
```suggestion
    it('throws when products array is empty', async () => {
```

matthiasgeihs (approved):
Only minor comments

packages/subscription-controller/src/types.ts
tuna1207: oh i was trying to follow the naming in [server](https://github.com/consensys-vertical-apps/va-mmcx-subscription-api/blob/main/src/subscriptions/subscriptions.constant.ts#L5) for consistency, i think we can update them both later if needed to make it more clear

packages/subscription-controller/src/types.ts
tuna1207: server updated, it's in case user profile didn't get a customer created in db yet
https://github.com/consensys-vertical-apps/va-mmcx-subscription-api/blob/main/src/subscriptions/subscriptions.service.ts#L375

packages/subscription-controller/src/SubscriptionController.ts
cryptodev-2s: Can we switch to using these explicit types for the controller actions? If we keep relying on the generic type, we run into the same issue, items can be added to the auto-generated union without defining a corresponding action/event type. These should also be exported at the controller level so that other controllers can reference them directly (they need to be added the index file).


```typescript
export type SubscriptionControllerGetSubscriptionsAction = {
  type: `${typeof controllerName}:getSubscriptions`;
  handler: SubscriptionController['getSubscriptions'];
};
export type SubscriptionControllerCancelSubscriptionAction = {
  type: `${typeof controllerName}:cancelSubscription`;
  handler: SubscriptionController['cancelSubscription'];
};
export type SubscriptionControllerStartShieldSubscriptionWithCardAction = {
  type: `${typeof controllerName}:startShieldSubscriptionWithCard`;
  handler: SubscriptionController['startShieldSubscriptionWithCard'];
};

export type SubscriptionControllerGetStateAction = ControllerGetStateAction<
  typeof controllerName,
  SubscriptionControllerState
>;
export type SubscriptionControllerActions =
  | SubscriptionControllerGetSubscriptionsAction
  | SubscriptionControllerStartShieldSubscriptionWithCardAction
  | SubscriptionControllerStartShieldSubscriptionWithCardAction
  | SubscriptionControllerGetStateAction;
```

We can then remove `CreateActionsObj` and `ActionsObj`. We also have a new feature coming that auto-generates messenger actions/events documentation, which assumes that each controller action/event is defined explicitly in this way (as defined in the controller guidelines)

tuna1207: got it, updated and added actions export in index file

packages/subscription-controller/src/types.ts
tuna1207: not needed yet i think since this one is pretty simple and not being reused

chaitanyapotti (approved):

cryptodev-2s (approved):
LGTM!

PR #6302, opened 2025-08-13
fix: ledger txs for swap and bridge

infiniteflower:
## Explanation

This PR ports the changes from an earlier patch into the `BridgeStatusController`. It fixes an issue where the user would be unable to submit Ledger txs on Mobile.

## Checklist


packages/bridge-status-controller/src/utils/transaction.ts
```diff
@@ -177,6 +177,24 @@ export const handleLineaDelay = async (
...
+ * Extension does not have this issue.
+ *
+ * @param requireApproval - Whether the delay should be applied
+ */
+export const handleMobileHardwareWalletDelay = async (
+  requireApproval: boolean,
+) => {
+  if (requireApproval) {
+    const mobileHardwareWalletDelay = new Promise((resolve) =>
+      setTimeout(resolve, 1000),
```
SteP-n-s, line 192: Is 1s enough or do we need to bump it to be extra safe?

infiniteflower: This is enough, because this patch exists in Mobile since at least the last release

SteP-n-s (approved):

PR #6304, opened 2025-08-13
Reduce global Jest timeout to 5 seconds

mcmire:
## Explanation

Currently, the global Jest timeout is 30 seconds. This was probably changed because some tests are complicated and may take a long time to resolve. But most tests are not this way, and using a large timeout leads to a poor developer experience.

## Checklist

### DIFF
--- a/packages/assets-controllers/src/AssetsContractControllerWithNetworkClientId.test.ts
+++ b/packages/assets-controllers/src/AssetsContractControllerWithNetworkClientId.test.ts
@@ -6,6 +6,7 @@ import {
   setupAssetContractControllers,
   mockNetworkWithDefaultChainId,
 } from './AssetsContractController.test';
+import { SECONDS } from '../../../tests/constants';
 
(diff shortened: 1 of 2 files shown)


jest.config.packages.js
```diff
@@ -180,7 +180,7 @@ module.exports = {
   // testRunner: "jest-circus/runner",
 
   // Default timeout of a test in milliseconds.
-  testTimeout: 30000,
+  // testTimeout: 5000,
```
cryptodev-2s, line 183: This block is commented out — was that intentional? Is the default timeout 5 seconds, or was it commented by mistake?

mcmire: Yes, exactly, the default timeout is 5s. I commented it out to match the other Jest options here which are commented out if we are just using their defaults. (I believe this config file was generated via Jest originally.)

cryptodev-2s (approved):
LGTM!

amitabh94 (approved):

PR #6311, opened 2025-08-14
feat: Default action/event type params to never

Gudahtt:
## Explanation

The action and event type parameters for the new `@metamask/messenger` package's Messenger class now default to `never`. This should make it harder to accidentally create a messenger with overly broad inferred types, making it less type safe than the author might expect.

The `Messenger` classes in `@metamask/base-controller` were not changed to avoid breaking changes. While this change itself wouldn't really be breaking by most definitions, it would require changes due to our use of the `@typescript-eslint/no-unnecessary-type-arguments` lint rule.

## References

This was a suggestion from here: https://github.com/MetaMask/core/pull/6142#discussion_r2274411959

## Checklist


mcmire (approved):
LGTM as well.

Gudahtt:
Note for reviewers on impact: this will add some "asymmetry" to messenger declarations, between messengers with _only_ actions and _only_ events.

For example, a messenger with _only_ actions would be declared as `Messenger<name, Actions>`, but a messenger with _only_ events would be declared as `Messenger<name, never, Events>`.

Not necessarily a problem, but I wanted to make sure this was known about. If we try to make these declarations more "symmetrical" by explicitly including `never` as the Events type for a messenger with no events, this will violate the `no-unnecessary-type-arguments` rule.

PR #6312, opened 2025-08-14
fix stale selectedNetworkClientId by converting to function pattern

salimtb:
## Explanation

**Problem**

The selectedNetworkClientId parameter was being passed as a static string value, which could become stale and frozen after controller reinitialization. This led to the controller potentially using outdated network client IDs, especially during network switches or controller restarts.

**Solution**

**Controller Changes:**

- Modified selectedNetworkClientId parameter from string to () => string function
- This ensures the network client ID is dynamically evaluated each time it's accessed
- Prevents stale values from being cached during controller lifecycle

**Test Updates:**

- Updated test setup function to expect function type instead of string
- Fixed 4 test cases that were incorrectly passing string literals
- Maintained 100% test coverage with all tests passing

## Checklist

### DIFF
--- a/packages/earn-controller/src/EarnController.ts
+++ b/packages/earn-controller/src/EarnController.ts
@@ -283,7 +283,7 @@ export class EarnController extends BaseController<
 > {
   #earnSDK: EarnSdk | null = null;
 
-  #selectedNetworkClientId: string;
+  #selectedNetworkClientId: () => string;
 
   readonly #earnApiService: EarnApiService;
 
@@ -303,7 +303,7 @@ export class EarnController extends BaseController<
     messenger: EarnControllerMessenger;
     state?: Partial<EarnControllerState>;
     addTransactionFn: typeof TransactionController.prototype.addTransaction;
-    selectedNetworkClientId: string;
+    selectedNetworkClientId: () => string;
     env?: EarnEnvironments;
   }) {
     super({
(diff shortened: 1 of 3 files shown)


ameliejyc (commented):
I've tested with alongside the mobile code and it looks good to me. But to understand this edge case better, when you say the network client ID "could become stale and frozen after controller reinitialization" can you give an example of how this would happen? The network client ID is always refreshed in the `networkDidChange` listener so I don't see it becoming stale there.

packages/earn-controller/src/EarnController.ts
```diff
@@ -303,7 +303,7 @@ export class EarnController extends BaseController<
     messenger: EarnControllerMessenger;
     state?: Partial<EarnControllerState>;
     addTransactionFn: typeof TransactionController.prototype.addTransaction;
-    selectedNetworkClientId: string;
+    selectedNetworkClientId: () => string;
```
mcmire, line 306: Hmm, this is not something we typically have to do in any other controller, so this is surprising to me that we have to do this. In your PR description you mention that it is necessary after the controller is reinitialized. When would this occur? I know that the Engine gets reinitialized in mobile in certain cases but usually _all_ controllers get reinitialized when this happens and so the NetworkController should be up to date. (Edit: I see that @ameliejyc expressed some similar concerns.)

If we still think this is necessary, have you considered moving the asynchronous operations in this controller to an `init` function? You should then be able to get the `selectedNetworkClientId` from NetworkController by going through its state.

mcmire (changes requested):
Had some concerns about the approach here.

packages/earn-controller/src/EarnController.ts
salimtb: @mcmire the main issue was race conditions, but you're right moving to the init function should resolve it. I’ll proceed with this approach instead on the client side and close this PR

salimtb: or maybe you mean init function inside the EarnController ?

mcmire: Either an `init` method in EarnController or using the init pattern on the client side should work, whichever makes sense.

PR #6314, opened 2025-08-14
fix: inaccurate swap Submitted event properties

micaelae:
## Explanation

Changes
- Update the implementation of `UnifiedSwapBridgeEventName.Submitted` to require event publishers to provide all properties. This is in needed because the Submitted event can be published after the BridgeController's state has been reset
- Parse event properties from the quote request if an event needs to be published prior to tx submission (i.e., Failed, Submitted)
- Calculate `actual_time_minutes` event property based on `txMeta.time` if available

## Checklist


packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -415,9 +416,11 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       targetContractAddress,
       approvalTxId,
       isStxEnabled,
+      accountAddress,
     } = startPollingForBridgeTxStatusArgs;
 
-    const accountAddress = this.#getMultichainSelectedAccountAddress();
+    const selectedAddress =
+      accountAddress ?? this.#getMultichainSelectedAccount()?.address ?? '';
```
infiniteflower: What's the reasoning for this change?

micaelae: New tx submissions pass in `accountAddress` to reduce getState calls on the AccountsController. However, the `startPollingForBridgeTxStatus` doesn't pass it in so this is a fallback

Now that I look at this again tho, maybe `accountAddress` should be required

infiniteflower (approved):

PR #6316, opened 2025-08-14
feat: Create new experimental `next` export

Gudahtt:
## Explanation

This export will be used to test upcoming breaking changes for the `@metamask/base-controller` package related to #5626

## References

Relates to #5626

## Checklist

### DIFF
--- a/packages/base-controller/next.js
+++ b/packages/base-controller/next.js
@@ -1,0 +1,3 @@
+// Re-exported for compatibility with Browserify.
+// eslint-disable-next-line
+module.exports = require('./dist/next/index.cjs');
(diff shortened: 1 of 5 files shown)


packages/base-controller/next.js
mcmire, line 1: Right, I forgot that Browserify doesn't support `exports`. How do we make it work now then? 🤔

Gudahtt: Browserify supports the `main` field in `package.json`, which matches what we export as `.` in every package.

We only needed a special case here because of the use of this additional export, at a different path. Browserify interprets it as an import of the file `next` at the root of the package, so here it is.

mcmire: Ah, makes sense!

PR #6326, opened 2025-08-15
feat: add alignment methods to group, wallet and service classes

hmalik88:
## Explanation

**What is the current state of things and why does it need to change?** An alignment mechanism didn't exist for multichain account groups. It needs to be added because we can have multiple scenarios where misalignment occurs:
1. State 1 -> State 2 - We will have to align existing account groups when they are created.
2. New protocols added - Existing groups will have to have an account associated with the new provider (chain).
3. Basic functionality toggle - With basic functionality off, we can have a scenario where only an EVM account is created when a group is created.
4. Discovery - When importing an SRP, the discovery process might not necessarily yield a discovered account for every provider and we have to align the group in those scenarios.

**What is the solution your changes offer and how does it work?** It adds an alignment mechanism at the group level and it runs through the list of providers and checks to see if any accounts exist for that particular group, if there aren't any, then it calls `provider.createAccounts` with the associated entropy id and group index. Alignment methods were added at the wallet level where a wallet would call align on all of its existing groups and similarly at the service level where it can call align on all of its wallets. Singular methods were also added at the wallet and service level where a particular group associated with a wallet can be aligned and a particular wallet can be aligned by the service.

Note: I also refactored the `withCreateAccount` in `SnapAccountProvider` to be a bit more readable and easier to understand.

## Checklist

### DIFF
--- a/packages/network-enablement-controller/CHANGELOG.md
+++ b/packages/network-enablement-controller/CHANGELOG.md
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `init()` method to safely initialize network enablement state from controller configurations ([#6329](https://github.com/MetaMask/core/pull/6329))
+
 ### Changed
 
+- Change transaction listener from `TransactionController:transactionConfirmed` to `TransactionController:transactionSubmitted` for earlier network enablement ([#6329](https://github.com/MetaMask/core/pull/6329))
+- Update transaction event handler to properly access chainId from nested transactionMeta structure ([#6329](https://github.com/MetaMask/core/pull/6329))
 - Bump `@metamask/controller-utils` from `^11.11.0` to `^11.12.0` ([#6303](https://github.com/MetaMask/core/pull/6303))
 
 ## [0.1.1]
--- a/packages/network-enablement-controller/package.json
+++ b/packages/network-enablement-controller/package.json
@@ -50,6 +50,7 @@
     "@metamask/auto-changelog": "^3.4.4",
     "@metamask/multichain-network-controller": "^0.11.1",
     "@metamask/network-controller": "^24.1.0",
+    "@metamask/transaction-controller": "^59.2.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
@@ -67,7 +68,8 @@
   },
   "peerDependencies": {
     "@metamask/multichain-network-controller": "^0.11.0",
-    "@metamask/network-controller": "^24.0.0"
+    "@metamask/network-controller": "^24.0.0",
+    "@metamask/transaction-controller": "^59.0.0"
   },
   "engines": {
     "node": "^18.18 || >=20"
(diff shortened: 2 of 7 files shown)


packages/multichain-account-service/src/MultichainAccountGroup.ts
```diff
@@ -188,4 +188,24 @@ export class MultichainAccountGroup<
...
+        if (!accounts || accounts.length === 0) {
+          return provider.createAccounts({
+            entropySource: this.wallet.entropySource,
+            groupIndex: this.groupIndex,
+          });
+        }
+        return Promise.resolve();
+      }),
+    );
+  }
```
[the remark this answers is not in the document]
hmalik88: Changed implementation to use `Promise.allSettled` and warn of a provider failure in [9775720](https://github.com/MetaMask/core/pull/6326/commits/9775720636e310f1388f22c8526d3ec7d9e05be2). Also, `sync` isn't needed to really be called in alignment. Unless we find a use case where we would need the group's state to be immediately updated for some subsequent action, then it is fine to keep as is.

packages/multichain-account-service/src/MultichainAccountGroup.ts
```diff
@@ -188,4 +188,30 @@ export class MultichainAccountGroup<
...
+   * This will create accounts for providers that don't have any accounts yet.
+   */
+  async align(): Promise<void> {
+    const results = await Promise.allSettled(
+      this.#providers.map((provider) => {
+        const accounts = this.#providerToAccounts.get(provider);
+        if (!accounts || accounts.length === 0) {
+          return provider.createAccounts({
+            entropySource: this.wallet.entropySource,
+            groupIndex: this.groupIndex,
```
fabiobozzo, line 204: @hmalik88 Can you help me understand `this.groupIndex` with an example please? 😇 

So, we have `account group #99` which has the following accounts: 
```
EVM #1  |  SOL #1
EVM #2  |  SOL #2
        |  SOL #3
        |
        |  SOL #5
```

I guess we need to create: EVM accounts (#3, 4, 5) and SOL #4.  
Does `provider.createAccounts` return all of them? Even that SOL account "in the gap"?

gantunesr: Why would a group have multiple EVM and SOL accounts? They should only have one per provider

packages/multichain-account-service/src/MultichainAccountGroup.ts
```diff
@@ -188,4 +188,30 @@ export class MultichainAccountGroup<
...
+
+  /**
+   * Align the multichain account group.
+   *
+   * This will create accounts for providers that don't have any accounts yet.
+   */
+  async align(): Promise<void> {
+    const results = await Promise.allSettled(
+      this.#providers.map((provider) => {
+        const accounts = this.#providerToAccounts.get(provider);
```
gantunesr, line 200: Is this getting all accounts for this provider? Should it just get the account for this specific group?

gantunesr: Like the method `getAccounts`

packages/multichain-account-service/src/MultichainAccountGroup.ts
hmalik88: It would be a wallet that would have multiple accounts where there could be misalignment e.g. in the discovery process and thus needs to be "aligned". And yeah what @gantunesr said.

packages/multichain-account-service/src/MultichainAccountGroup.ts
hmalik88: No, the `providerToAccounts` map just grabs the ones associated with this specific `entropySource` + `groupIndex` combo.

Issue #6327, opened 2025-08-15
Migrate JSON-RPC stack to `JsonRpcEngineV2`


rekmarks:
Following #6088, all existing uses of `JsonRpcEngine` and its middleware pattern must be migrated to `JsonRpcEngineV2`. This is the tracking issue for this process, which will have to be incremental. This issue does not yet provide a full account of every affected package, but at least tries to record some of the non-obvious requirements of the migration.

#6176, the implementation for #6088, provides backwards- and forwards-compatibility adapters in the form of `asLegacyMiddleware()` and `asV2Middleware()`, respectively. A small but possibly negligible performance hit should be expected whenever these adapters are used. The upshot is that a complex middleware pipeline, e.g. the MetaMask clients' "provider engines", can be migrated incrementally in arbitrary order. That being said, it's probably best to identify the distinct components of these middleware pipelines and migrate them one at a time. It may also make the most sense to start at the end of the engine / middleware stack and work back up to the "root" / initial engine. This would imply e.g. migration `@metamask/eth-json-rpc-middleware` first, then the method middleware, and so and and so forth until the initial engine is finally migrated. However, due to `asV2Middleware()`, the initial engine could be migrated at any time, which would begin to address one of the problems motivating the migration in the first place, namely mangled errors from the legacy engine module.

### TODO

This list is unordered and potentially inexhaustive.

- [ ] Migrate `'notification'` event notifications to some new pattern
   - `JsonRpcEngine` is also a `SafeEventEmitter`. To wit, it emits a single event, `'notification'`, which we use exclusively(?) to forward JSON-RPC notifications to streams. The engine does nothing with these notifications except include them as the payload of the event.
   - In other words, this is just a convenient if idiosyncratic way of sending notifications to different remotes, including dapps. Rather than (ab)using the engine in this manner, we could just write these notifications directly to their intended dispatch mechanisms (again, usually a stream).
   - If we want to leverage `json-rpc-engine` to emit notifications, the proper place to do it would be on `JsonRpcServer`. However, it would need to receive a function or stream to dispatch the notification, and that is probably the wrong solution for our stack.
- [ ] Migrate method middlewares
   - See `/methodmiddleware/giu` in extension and mobile.
   - The method middleware pattern expects a legacy middleware function as the method implementation, so either all method implementations per method middleware are migrated in one go, or they are migrated piecemeal.
- [ ] Migrate `@metamask/providers`
  - Our providers use `JsonRpcEngine` internally.
[...]

PR #6328, opened 2025-08-15
refactor!: Remove "data" event from SafeEventEmitterProvider

rekmarks:
## Explanation

The `SafeEventEmitterProvider` `'data'` event relies on the `JsonRpcEngine` `'notification'` event, which per #6327 we are trying to get rid of. After failing to find an instance outside of tests where we actually use the `'data'` event, we assess that it can be removed.

## References

* Ref: #6327

## Checklist

- [x] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes (really just checking that CI passes in this case)
   - https://github.com/MetaMask/metamask-extension/pull/35498  
   - https://github.com/MetaMask/metamask-mobile/pull/18907


packages/eth-json-rpc-provider/src/safe-event-emitter-provider.ts
```diff
@@ -64,12 +64,6 @@ export class SafeEventEmitterProvider extends SafeEventEmitter {
   constructor({ engine }: { engine: JsonRpcEngine }) {
     super();
     this.#engine = engine;
-
-    if (engine.on) {
-      engine.on('notification', (message: string) => {
```
FrederikBolding, line 69: Your PR seems to imply that this isn't in use. Have you tested the extension with these lines stripped out?

Gudahtt: We searched for examples in all of our repos. We found a couple of uses, but only in test helpers.

I don't think it has been tested, that is a good idea though. We could probably delete this quickly in a patch, put it up as a draft, then see if CI passes.

Gudahtt:
@metamaskbot publish-preview

FrederikBolding (approved):

PR #6332, opened 2025-08-19
feat(multichain-account-service): basic functionality management

fabiobozzo:

Currently, the `MultichainAccountService` has no mechanism to manage provider states based on user preferences like the "basic functionality" toggle. When users disable advanced features (basic functionality OFF), all providers (EVM, Solana, Bitcoin, etc.) remain active and continue creating multichain accounts, which goes against the user's preference for a simplified wallet experience.

The extension's `PreferencesController` manages the basic functionality toggle (`useExternalServices`), but there was no way to communicate this state change to the core `MultichainAccountService` to disable non-essential providers.

This PR introduces a clean provider state management system that allows clients to control which providers are active based on user preferences:

1. **Provider-Level Disable Mechanism**: Added `setDisabled(disabled: boolean)` method to `BaseAccountProvider`. Providers now **can / should** check `this.isDisabled` in their `createAccounts` methods and return empty arrays when disabled, preventing new account creation.

2. `EvmAccountProvider` could be disabled, for other reasons in the future, but does not depend on basic functionality, at all.

3. Added `setBasicFunctionality({ enabled: boolean })` method to `MultichainAccountService` that:
   - Calls `setDisabled(!enabled)` on all providers
   - **Triggers wallet alignment when basic functionality is enabled to ensure account groups are complete**

4. **Extension Integration**: The extension's `PreferencesController` now calls this method when the basic functionality toggle changes.

## References

Alignment method provided by:  https://github.com/MetaMask/core/pull/6326

Stories: 
- [MUL-343](https://consensyssoftware.atlassian.net/browse/MUL-343)
- [MUL-525](https://consensyssoftware.atlassian.net/browse/MUL-525)
- [MUL-524](https://consensyssoftware.atlassian.net/browse/MUL-524)

## Checklist

- [] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary

[MUL-343]: https://consensyssoftware.atlassian.net/browse/MUL-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -33,6 +33,12 @@ export class SolAccountProvider extends SnapAccountProvider {
     entropySource: EntropySourceId;
     groupIndex: number;
   }): Promise<Bip44Account<KeyringAccount>[]> {
+    if (this.isDisabled) {
+      console.log(
+        `${this.constructor.name} is disabled - skipping account creation`,
+      );
+      return [];
```
gantunesr: Should return an empty array or throw an error?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -359,6 +359,36 @@ export class MultichainAccountService {
...
+  }): Promise<void> {
+    console.log(
+      `MultichainAccountService: Setting basic functionality ${enabled ? 'enabled' : 'disabled'}`,
+    );
+
+    // Loop through all providers and set their disabled state
+    for (const provider of this.#providers) {
+      // When basic functionality is disabled, disable all providers.
+      provider.setDisabled(!enabled);
+    }
```
gantunesr: We shouldn't disable all providers, just the ones that are snap based

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -42,10 +42,25 @@ export abstract class BaseAccountProvider
...
+   * When disabled, the provider should not create new accounts.
+   *
+   * @param disabled - Whether the provider should be disabled.
+   */
+  setDisabled(disabled: boolean): void {
+    this.isDisabled = disabled;
+    console.log(
+      `Provider ${this.constructor.name} ${disabled ? 'disabled' : 'enabled'}`,
+    );
+  }
```
gantunesr: I had the impression that you were going to develop a new class to wrap the providers

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -42,10 +42,25 @@ export abstract class BaseAccountProvider
 {
   protected readonly messenger: MultichainAccountServiceMessenger;
 
+  protected isDisabled: boolean = false;
```
hmalik88: Wondering if we should just keep this on the `SnapAccountProvider`, unless there's a scenario where we see us disabling the EVM provider? cc: @gantunesr

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: This would also change the property on the EVM provider unnecessarily -- especially since we aren't using the `isDisabled` boolean for anything right now in the EVM provider's `createAccounts`, I think we should alter this to something like this:

```suggestion
    for (const provider of this.#providers) {
      // When basic functionality is disabled, disable all providers.
      if (provider.snapId) {
        provider.setDisabled(!enabled);
      }
    }
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -359,6 +359,36 @@ export class MultichainAccountService {
...
+   * When enabled, all providers are enabled and wallet alignment is triggered.
+   *
+   * @param options - Options.
+   * @param options.enabled - Whether basic functionality is enabled.
+   */
+  async setBasicFunctionality({
+    enabled,
+  }: {
+    enabled: boolean;
+  }): Promise<void> {
```
hmalik88: Any reason why this param is an object?

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
hmalik88: Actually, I think this might be okay here...since we would need to have this property available for when the provider calls `getAccount` or `getAccounts`, which exist at the `BaseAccountProvider` level.

hmalik88 (commented):
I think we also need to make sure we return an empty array or throw for `getAccounts` and `getAccount` based on `isDisabled`.

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: True! I thought: maybe there could be another reason to disable providers in the future, and it's up to each provider to decide what's best to do w/ that state. Rn, disabled means just `createAccounts` is inhibited, there Evm provider ignores that. But I see your point, that was my initial thought, I'll stick to that 👌

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
fabiobozzo: Also `SolAccountProvider` ? Moreover: how would you avoid type-checking tricks around 
```
provider.setDisabled(true)
```
in that case? 

In my case, I decided to just add `setDisabled` to the `AccountProvider` **interface**, but that's maybe too drastic?

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: Perfect 👍

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: None. The residue of my initial take on setBasicFunctionality.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -359,6 +360,34 @@ export class MultichainAccountService {
...
+   */
+  async setBasicFunctionality(enabled: boolean): Promise<void> {
+    console.log(
+      `MultichainAccountService: Setting basic functionality ${enabled ? 'enabled' : 'disabled'}`,
+    );
+
+    // Loop through providers and disable only snap-based ones when basic functionality is disabled
+    for (const provider of this.#providers) {
+      if ('snapId' in provider) {
+        (provider as SnapAccountProvider).setDisabled(!enabled);
```
fabiobozzo: @hmalik88 any better ideas on how to avoid that type casting ❓

packages/multichain-account-service/src/providers/SolAccountProvider.ts
fabiobozzo: Yes. Btw is SOL account provider subject to being enabled/disabled too? Or just Snap's?

gantunesr: The Solana provider is a snap provider so it should be

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: You can write a type guard:

In `SnapAccountProvider.ts`:

```ts
export const isSnapAccountProvider = (
  provider: unknown,
): provider is SnapAccountProvider => {
  return (
    provider !== null && typeof provider === 'object' && 'snapId' in provider
  );
};
```

and then in the service you can just do:

 ```ts
if (isSnapAccountProvider(provider) {
  provider.setDisabled(!enabled)
}
```

File: packages/multichain-account-service/src/providers/BaseAccountProvider.ts (hunk 3 above)
Comment (fabiobozzo) [reply packages/multichain-account-service/src/providers/BaseAccountProvider.ts]: @gantunesr I gave it a try! 😉 Now The Solana account provider is wrapped.  
@hmalik88 I initially named it `DisableableProviderWrapper` which is more correct but it sounds like a tongue twister 😵‍💫 wdyt?

File: packages/multichain-account-service/src/providers/SolAccountProvider.ts (hunk 1 above)
Comment (fabiobozzo) [reply packages/multichain-account-service/src/providers/SolAccountProvider.ts]: I didn't realize Solana provider IS A Snap's provider 😮‍💨

File: packages/multichain-account-service/src/providers/ProviderWrapper.ts [hunk 7]
```diff
@@ -0,0 +1,129 @@
...
+   */
+  async createAccounts(options: {
+    entropySource: EntropySourceId;
+    groupIndex: number;
+  }): Promise<Bip44Account<KeyringAccount>[]> {
+    if (this.isDisabled) {
+      console.log(
+        `Provider ${this.provider.constructor.name} is disabled - skipping account creation`,
+      );
+      return [];
```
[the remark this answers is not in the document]
fabiobozzo: I'll remove that before merging. It's just there for me to run manual tests atm.

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
hmalik88: I think what you have now is good `ProviderWrapper`, we might need different properties in the future that might conflict with the name `DisableableProviderWrapper`

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Get accounts, returns empty array when disabled.
+   *
+   * @returns Array of accounts, or empty array if disabled.
+   */
+  getAccounts(): Bip44Account<KeyringAccount>[] {
+    if (this.isDisabled) {
+      return [];
```
gantunesr: Should this throw an error?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
gantunesr: Similarly here, should this throw an error?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
...
+   */
+  async discoverAndCreateAccounts(options: {
+    entropySource: EntropySourceId;
+    groupIndex: number;
+  }): Promise<Bip44Account<KeyringAccount>[]> {
+    if (this.isDisabled) {
+      console.log(
+        `Provider ${this.provider.constructor.name} is disabled - skipping account discovery`,
+      );
+      return [];
```
gantunesr: Same here

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
+import type { AccountProvider, Bip44Account } from '@metamask/account-api';
+import type { EntropySourceId, KeyringAccount } from '@metamask/keyring-api';
+
+/**
+ * A simple wrapper that adds disable functionality to any AccountProvider.
+ * When disabled, the provider will not create new accounts and return empty results.
+ */
+export class ProviderWrapper
```
gantunesr: Looks great!

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: I'm not sure tbh 😐 Perhaps: 
- getAccounts: return empty array, **or maybe keep returning the accounts that could have been created prior disabling the provider** ❓ 
- createAccounts, discoverAndCreateAccounts: throw an error, to signal an invalid behavior to the caller?

gantunesr: > _maybe keep returning the accounts that could have been created prior disabling the provider❓_

This would require us to keep a cache of the accounts, right?

> _createAccounts, discoverAndCreateAccounts: throw an error, to signal an invalid behavior to the caller?_

I think it would make more sense since it would indicate that the request can't be fulfilled but interested in hearing others opinions

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -12,6 +12,14 @@ export type RestrictedSnapKeyringCreateAccount = (
   options: Record<string, Json>,
 ) => Promise<KeyringAccount>;
 
+export const isSnapAccountProvider = (
+  provider: unknown,
+): provider is SnapAccountProvider => {
+  return (
+    provider !== null && typeof provider === 'object' && 'snapId' in provider
```
danroc: Should we use `provider instanceof SnapAccountProvider` here?

packages/multichain-account-service/src/providers/SnapAccountProvider.test.ts
```diff
@@ -0,0 +1,31 @@
+import { isSnapAccountProvider } from './SnapAccountProvider';
+
+describe('SnapAccountProvider', () => {
+  describe('isSnapAccountProvider', () => {
+    it('returns true for object with snapId property', () => {
+      const mockProvider = { snapId: 'test-snap-id' };
+
+      expect(isSnapAccountProvider(mockProvider)).toBe(true);
```
danroc: It should return `false` for a plain object that doesn't implement the interface of `SnapAccountProvider`.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
danroc: IMO it's fine to return an empty list instead of throwing.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
...
+   * @returns True if the account is compatible.
+   */
+  isAccountCompatible(account: Bip44Account<KeyringAccount>): boolean {
+    // Check if the provider has the method (from BaseAccountProvider)
+    if (
+      'isAccountCompatible' in this.provider &&
+      typeof this.provider.isAccountCompatible === 'function'
+    ) {
+      return this.provider.isAccountCompatible(account);
+    }
```
danroc: @ccharly, maybe `BaseAccountProvider` wasn't named correctly since it expects the accounts to be BIP-44 accounts. Should it be `BaseBip44AccountProvider`?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
+import type { AccountProvider, Bip44Account } from '@metamask/account-api';
+import type { EntropySourceId, KeyringAccount } from '@metamask/keyring-api';
+
+/**
+ * A simple wrapper that adds disable functionality to any AccountProvider.
+ * When disabled, the provider will not create new accounts and return empty results.
+ */
+export class ProviderWrapper
+  implements AccountProvider<Bip44Account<KeyringAccount>>
```
danroc: Maybe it should extend `BaseAccountProvider` instead, since it expects accounts to be BIP-44 and implements the `isAccountCompatible` method.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
...
+  constructor(provider: AccountProvider<Bip44Account<KeyringAccount>>) {
+    this.provider = provider;
+  }
+
+  /**
+   * Set the disabled state for this provider.
+   *
+   * @param disabled - Whether the provider should be disabled.
+   */
+  setDisabled(disabled: boolean): void {
```
danroc: Nit/thought: Should we use `*Enabled` to avoid the "reverse logic" and align with `setBasicFunctionality(enabled)`?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -359,12 +377,54 @@ export class MultichainAccountService {
...
+    return this.#isAlignmentInProgress;
+  }
+
   /**
    * Align all multichain account wallets.
    */
   async alignWallets(): Promise<void> {
-    const wallets = this.getMultichainAccountWallets();
-    await Promise.all(wallets.map((w) => w.alignGroups()));
+    this.#isAlignmentInProgress = true;
```
danroc: If `alignWallets()` is called twice, it might run into a race condition: the first execution will set `#isAlignmentInProgress` to false before the second one has finished.

Maybe you should make `#isAlignmentInProgress` atomic and don't run the alignment if it's already in progress.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -373,7 +433,12 @@ export class MultichainAccountService {
    * @param entropySource - The entropy source of the multichain account wallet.
    */
   async alignWallet(entropySource: EntropySourceId): Promise<void> {
-    const wallet = this.getMultichainAccountWallet({ entropySource });
-    await wallet.alignGroups();
+    this.#isAlignmentInProgress = true;
```
danroc: I think we have the same problem here, plus it shares the same `#isAlignmentInProgress` variable for all wallets, which can also lead to a different race condition: calling `alignWallet` for different entropy sources will overwrite `#isAlignmentInProgress`.

Maybe this flag should be moved into the wallet itself instead of being kept at this level.

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -626,6 +627,59 @@ describe('MultichainAccountService', () => {
...
+
+      mocks.EvmAccountProvider.createAccounts.mockRejectedValueOnce(
+        new Error('Test error'),
+      );
+
+      try {
+        await service.alignWallets();
+      } catch {
+        // Expected to throw
+      }
```
ccharly: What about using this instead (not tested):
```suggestion
      await expect(service.alignWallets()).rejects.toThrow('Test error');
```

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Wouldn't extending BaseAccountProvider artificially constrain it and break the decorator pattern's core principle of being implementation-agnostic? I'd like to allow it to wrap any AccountProvider implementation, not just BaseAccountProvider subclasses. wdyt @ccharly ?

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: Ok. I'll scope alignment state to the resource being aligned (individual wallets) rather than being global. Concurrent operations on different wallets should then be safe and concurrent operations on the same wallet are prevented.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Yep, that makes more sense indeed 😊

packages/multichain-account-service/src/providers/SnapAccountProvider.test.ts
fabiobozzo: Oops! I updated the test to expect false for plain objects ✔️

fabiobozzo: And added a new test case with actual SolAccountProvider instance that returns true..

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
fabiobozzo: I think so... I don't see a valid reason for a provider to have `snapId` and NOT extend SnapAccountProvider 🤔 I think you're right then... ✔️

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: @danroc do you mean on each method? or as I discussed w/ Guto: 
- read methods -> return []
- write methods -> throw

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -359,12 +377,54 @@ export class MultichainAccountService {
...
 
+  /**
+   * Set basic functionality state and trigger alignment if enabled.
+   * When basic functionality is disabled, snap-based providers are disabled.
+   * When enabled, all snap providers are enabled and wallet alignment is triggered.
+   * EVM providers are never disabled as they're required for basic wallet functionality.
+   *
+   * @param enabled - Whether basic functionality is enabled.
+   */
+  async setBasicFunctionality(enabled: boolean): Promise<void> {
```
ccharly, line 381: IDK if we should have this as part of the public API 🤔

My initial idea was to have the "account provider wrapper" to "react" to some events (or maybe a runtime callback) and adapt each of their methods to return all accounts or no accounts (based on the "basic functionality flag").

Though, I haven't checked of this was done on both clients, so maybe it's not that easy to do

packages/multichain-account-service/src/providers/ProviderWrapper.ts
ccharly: Yes, and that's also the whole "idea" of that "wrapper" actually. The service will consume each providers as-is, and some providers will behave differently based on the basic functionality, thus, re-building different groups automatically upon "re-alignment".

packages/multichain-account-service/src/providers/ProviderWrapper.ts
ccharly: Based on our previous discussion, I think this would make sense to rename it yes 👍

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,129 @@
...
+   *
+   * @param account - The account to check.
+   * @returns True if the account is compatible.
+   */
+  isAccountCompatible(account: Bip44Account<KeyringAccount>): boolean {
+    // Check if the provider has the method (from BaseAccountProvider)
+    if (
+      'isAccountCompatible' in this.provider &&
+      typeof this.provider.isAccountCompatible === 'function'
+    ) {
```
ccharly: This looks a bit weird to rely on runtime check in this context. All providers here should be "bip-44 compatible" thus, if we go with [@danroc's suggestion](https://github.com/MetaMask/core/pull/6332/files#r2300392489) of relying on [`BaseBip44AccountProvider`](https://github.com/MetaMask/core/pull/6332/files#r2300389519), then this check won't be needed.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
ccharly: For consistency, I would also add a generic here and like `<Account extends KeyringAccount>` and re-use it in the other types that need one (thus we could use different account type if really needed).

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Indeed 👌 providers return different results when disabled, triggering automatic group rebuilds.

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Renamed ✔️

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Broken decorator pattern in order to have a `isAccountCompatible` impl that makes more sense.

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: See https://github.com/MetaMask/core/pull/6332#discussion_r2300543480 WDYT?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
fabiobozzo: Done ✔️

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: This is something to discuss @ccharly . It's an architectural decision, and I chose the simplest path (multichain account service was already available to both places where basicFunctionality is toggled in extension and mobile). 

Are you suggesting we use `MultichainAccountServiceMessenger` to dispatch a `MultichainAccountServiceSetBasicFunctionalityAction` ? I'm not sure how to get a ref to that messenger in mobile tbh.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -40,7 +42,7 @@ type AccountContext<Account extends Bip44Account<KeyringAccount>> = {
 export class MultichainAccountService {
   readonly #messenger: MultichainAccountServiceMessenger;
 
-  readonly #providers: AccountProvider<Bip44Account<KeyringAccount>>[];
+  readonly #providers: (BaseBip44AccountProvider | ProviderWrapper)[];
```
ccharly: I think we can keep `AccountProvider<Bip44Account<KeyringAccount>>` here, since `ProviderWrapper` should still be compatible with this.

```suggestion
  readonly #providers: AccountProvider<Bip44Account<KeyringAccount>>[];
```

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -37,7 +37,7 @@ export function assertAreBip44Accounts(
   accounts.forEach(assertIsBip44Account);
 }
 
-export abstract class BaseAccountProvider
+export abstract class BaseBip44AccountProvider
```
ccharly, line 40: Could you also rename the entire file to `BaseBip44AccountProvider` too?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { EntropySourceId, KeyringAccount } from '@metamask/keyring-api';
+
+import { BaseBip44AccountProvider } from './BaseAccountProvider';
+import type { MultichainAccountServiceMessenger } from '../types';
+
+/**
+ * A simple wrapper that adds disable functionality to any BaseAccountProvider.
+ * When disabled, the provider will not create new accounts and return empty results.
+ */
+export class ProviderWrapper extends BaseBip44AccountProvider {
```
ccharly: Could you rename this to `AccountProviderWrapper`? and also its file?

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Well, let's be pragmatic here and keep your implementation for now. It's simple enough and does get the job done, so that's ok! 👍

We can re-visit that later once we have a bit more time.

fabiobozzo:
@metamaskbot publish-preview

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
```diff
@@ -409,4 +409,93 @@ describe('MultichainAccountWallet', () => {
...
+
+      // Start second alignment while first is still running
+      const secondAlignment = wallet.alignGroup(0);
+
+      // Both should complete without error
+      await Promise.all([firstAlignment, secondAlignment]);
+
+      // Provider should only be called once (not twice due to concurrency protection)
+      expect(providers[1].createAccounts).toHaveBeenCalledTimes(1);
+    });
```
ccharly, line 499: That's a duplicate right?
```suggestion
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -307,12 +309,30 @@ export class MultichainAccountWallet<
...
   }
 
+  /**
+   * Gets whether alignment is currently in progress for this wallet.
+   *
+   * @returns True if alignment is in progress, false otherwise.
+   */
+  getIsAlignmentInProgress(): boolean {
+    return this.#isAlignmentInProgress;
+  }
```
ccharly, line 319: Should we have a similar logic for the groups themselves? 🤔

For the moment, we will only trigger alignments from the service level, but it could make sense to have this logic on the groups and compute `getIsAlignementInProgress` using a `.some` to check if at least one group is running alignment? 🤔

packages/multichain-account-service/src/index.ts
```diff
@@ -8,8 +8,9 @@ export type {
   MultichainAccountServiceGetMultichainAccountGroupsAction,
   MultichainAccountServiceCreateMultichainAccountGroupAction,
   MultichainAccountServiceCreateNextMultichainAccountGroupAction,
+  MultichainAccountServiceGetIsAlignmentInProgressAction,
 } from './types';
-export { BaseAccountProvider, SnapAccountProvider } from './providers';
+export { BaseBip44AccountProvider, SnapAccountProvider } from './providers';
```
ccharly: We should probably export the `AccountProviderWrapper` too, if we plan to externalize some of the providers on the client side. So those providers could also benefit from the "basic functionality" logic if needed, WDYT?

packages/multichain-account-service/src/providers/ProviderWrapper.ts
```diff
@@ -0,0 +1,120 @@
...
+   * @returns The account with the specified ID, or undefined if not found.
+   */
+  getAccount(
+    id: Bip44Account<KeyringAccount>['id'],
+  ): Bip44Account<KeyringAccount> | undefined {
+    if (!this.isEnabled) {
+      return undefined;
+    }
+    return this.provider.getAccount(id);
+  }
```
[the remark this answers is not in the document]
ccharly: NOTE: Cursor is wrong here.

That's not valid. The current `AccountProviderWrapper` states this:
```ts
/**
 * ...
 * @throws When disabled or account not found.
 */
```

And since the wrapper uses a `BaseBip44AccountProvider`, the `getAccount` will throw if the account is not found. So the return type is correct here, either we throw or we get a valid account.

packages/multichain-account-service/src/providers/AccountProviderWrapper.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Simple type guard to check if a provider is wrapped.
+ *
+ * @param provider - The provider to check.
+ * @returns True if the provider is an AccountProviderWrapper.
+ */
+export function isAccountProviderWrapper(
+  provider:
+    | AccountProvider<Bip44Account<KeyringAccount>>
+    | AccountProviderWrapper,
```
ccharly: Very nit: should we use `unknown` here instead? That's what we used for `isSnapAccountProvider` already.

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -10,13 +10,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
+- **BREAKING**: Rename `BaseAccountProvider` to `BaseBip44AccountProvider` for clarity ([#6332](https://github.com/MetaMask/core/pull/6332))
+  - `ProviderWrapper` now extends `BaseBip44AccountProvider` instead of implementing interface directly ([#6332](https://github.com/MetaMask/core/pull/6332))
```
ccharly: I think there's no need to mention this, since the "wrapper" is a new thing.
```suggestion
```

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -10,13 +10,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ## [0.5.0]
 
 ### Added
 
 - Allow for multichain account group alignment through the `align` method ([#6326](https://github.com/MetaMask/core/pull/6326))
   - You can now call alignment from the group, wallet and service levels.
+- Add `setBasicFunctionality` method to control providers state and trigger wallets alignment ([#6332](https://github.com/MetaMask/core/pull/6332))
+  - Add `ProviderWrapper` to handle Snap account providers behavior when disabled
```
ccharly: This has to be moved under `## Unrealeased` instead.

Also, we should rename `ProviderWrapper` to `AccountProviderWrapper` (and we usually ends "sub-entries" with a dot in the end.

Can you move this please? 😃

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -749,5 +802,164 @@ describe('MultichainAccountService', () => {
...
+
+  describe('setBasicFunctionality', () => {
+    it('accepts a boolean parameter instead of object', async () => {
+      const { service } = setup({ accounts: [MOCK_HD_ACCOUNT_1] });
+
+      // These should not throw errors
+      expect(await service.setBasicFunctionality(true)).toBeUndefined();
+      expect(await service.setBasicFunctionality(false)).toBeUndefined();
+    });
+
```
ccharly: Not sure if this is a left over test? We are covering the same things on the next 2 tests, so maybe we can remove it entirely?
```suggestion
```

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
fabiobozzo: Not really 😬 

1. prevents concurrent alignGroups calls - Tests race condition in wallet.alignGroups() (all groups)

2. prevents concurrent alignGroup calls - Tests race condition in wallet.alignGroup(0) (single group) -> separate race condition check on line 345

Do you see a way to make that more compact?

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo: In this particular case, I'd keep it simple (for now). I see three potential disadvantages if we move it to group-level: 

1. more state to manage and synchronize
2. race condition risks: if providers are shared between groups, concurrent alignments could conflict
3. As you noted, _we will only trigger alignments from the service level_, therefore I'd leave the group-level tracking additional complexity for later

Just my 2cents btw 🙃 I can refactor it quickly if you think it adds value!

packages/multichain-account-service/src/providers/ProviderWrapper.ts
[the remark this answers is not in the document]
fabiobozzo: Yep yep 😋 I thought so, I kept then consistency w/ the contract. Therefore I test: 
```
expect(() => wrapper.getAccount('test-id')).toThrow('Provider is disabled');
```

fabiobozzo:
@metamaskbot publish-preview

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,9 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `setBasicFunctionality` method to control providers state and trigger wallets alignment ([#6332](https://github.com/MetaMask/core/pull/6332))
+  - Add `AccountProviderWrapper` to handle Snap account providers behavior when disabled.
+  - Export `AccountProviderWrapper` from public API to allow external providers to use basic functionality logic
```
ccharly: What about?

```suggestion
  - Add `AccountProviderWrapper` to handle Snap account providers behavior according to the basic functionality flag.
```

packages/multichain-account-service/src/index.ts
```diff
@@ -8,8 +8,13 @@ export type {
   MultichainAccountServiceGetMultichainAccountGroupsAction,
   MultichainAccountServiceCreateMultichainAccountGroupAction,
   MultichainAccountServiceCreateNextMultichainAccountGroupAction,
+  MultichainAccountServiceGetIsAlignmentInProgressAction,
```
ccharly, line 11: We also miss the `MultichainAccountServiceSetBasicFunctionalityAction` no?

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: No that's ok! Let's keep it simple for now! Thanks 😄

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
ccharly: Oh I misread then, no let's keep it as-is for now! Thanks

ccharly (approved):
LGTM!

PR #6335, opened 2025-08-19
feat: Migrate controller guidelines and examples to new Messenger

Gudahtt:
## Explanation

Migrate the controller guidelines and the `sample-controllers` package to the `next` export of the `@metamask/base-controller` package and the new `Messenger` class from `@metamask/messenger`.

## References

Relates to https://github.com/MetaMask/core/issues/5626

## Checklist


packages/sample-controllers/src/sample-gas-prices-controller.test.ts
```diff
@@ -121,30 +124,57 @@ function getRootMessenger({
...
+    AllSampleGasPricesControllerEvents,
+    typeof rootMessenger
+  >({
+    namespace: 'SampleGasPricesController',
+    parent: rootMessenger,
+  });
+  // Delegate external actions/events
+  rootMessenger.delegate({
+    actions: ['NetworkController:getState'],
+    messenger,
```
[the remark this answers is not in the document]
Gudahtt: wat

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
```diff
@@ -121,30 +124,57 @@ function getRootMessenger({
...
   networkControllerGetStateActionHandler?: NetworkControllerGetStateAction['handler'];
-} = {}): Messenger<RootAction, RootEvent> {
-  const rootMessenger = new Messenger<RootAction, RootEvent>();
-  rootMessenger.registerActionHandler(
+} = {}): Messenger<
+  'Root',
+  AllSampleGasPricesControllerActions,
+  AllSampleGasPricesControllerEvents
+> {
+  const rootMessenger = new Messenger<
```
mcmire: I seem to recall that while discussing the ADR, two patterns came up for initializing the messenger. The existing way is to pass type parameters to the constructor, but I think we were also playing with using a type annotation instead:

``` typescript
  const rootMessenger: Messenger<
    'Root',
    AllSampleGasPricesControllerActions,
    AllSampleGasPricesControllerEvents
  > = new Messenger({ namespace: 'Root' });
```

As I see it, the advantage of using a type annotation is that if a type for the new messenger is already defined somewhere, we can simply reuse it, so for example:

``` typescript
  const rootMessenger: RootMessenger = new Messenger({ namespace: 'Root' });
```

Although we'd need a utility type to extend an existing messenger type with a parent, for instance:

``` typescript
  const messenger: ChildMessenger<
    SampleGasPricesControllerMessenger,
    typeof rootMessenger
  > = new Messenger({
    namespace: 'SampleGasPricesController',
    parent: rootMessenger,
  });
```

The downside is that we don't typically ask people to use type annotations, in fact it's something we advise against. 

I think your change is fine for now, but any thoughts on the direction we should take here?

docs/controller-guidelines.md
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
+  FooControllerEvents | AllowedEvents
 >;
+
+/* === packages/foo-controller/src/index.ts === */
+
+export type { AllowedEvents } from '@metamask/foo-controller';
 ```
 
-🚫 **`AllowedEvents['type']` is passed as the 5th type parameter, but `AllowedEvents` is exported**
+🚫 **External events are included in controller actieventn type**
```
mcmire: Typo:

``` suggestion
🚫 **External events are included in controller event type**

File: docs/controller-guidelines.md [hunk 4]
```diff
@@ -247,22 +245,40 @@ class FooController extends BaseController<
...
+  'FooController',
+  FooControllerActions,
+  FooControllerEvents | BarControllerStateChange,
+  typeof rootMessenger
+>({
+  namespace: 'FooController',
+  parent: rootMessenger,
+});
+rootMessenger.delegate({
+  events: ['BarController:stateChange' as const],
```
mcmire: Is the `as const` really needed, and if so, can it go outside the array?

``` suggestion
  events: ['BarController:stateChange'] as const,

File: docs/controller-guidelines.md [hunk 5]
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
+  FooControllerEvents
 >;
+
+/* === packages/foo-controller/src/index.ts === */
+
+export type { AllowedActions } from '@metamask/foo-controller';
 ```
 
-🚫 **`AllowedActions['type']` is passed as the 4th type parameter, but `AllowedActions` is exported**
+🚫 **External actions are included in controller action type**
```
mcmire, line 606: Perhaps this anti-pattern should be moved up to "Define and export a type union for internal action types"?

docs/controller-guidelines.md
mcmire: But also, perhaps this anti-pattern should be moved up to "Define and export a type union for internal event types"?

docs/controller-guidelines.md
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
+  FooControllerEvents
 >;
-
-/* === packages/foo-controller/src/index.ts === */
-
-export type { AllowedEvents } from '@metamask/foo-controller';
 ```
 
-✅ **`AllowedEvents['type']` is passed as the 5th type parameter, and `AllowedEvents` is _not_ exported**
+✅ **`AllowedEvents` is included in the 3nd type parameter but is _not_ exported**
```
mcmire: Since we no longer have an example where `AllowedActions` is included in a different type parameter, perhaps we should say:

``` suggestion
✅ **`AllowedEvents` is included in the events type union but is _not_ exported**
```

docs/controller-guidelines.md
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
-In this case, the controller should group these types into a type union so that they can be easily passed to the `RestrictedMessenger` type. However, it should not export this type, as it would then be re-exporting types that another package has already exported.
+In this case, the controller should group these types into a type union so that they can be easily passed to the `Messenger` type. However, it should not export this type, as it would then be re-exporting types that another package has already exported.
 
 The name of this type should be `AllowedEvents`.
 
-This type should not only be passed to `RestrictedMessenger` as the 3rd type parameter, but should also be included in its 5th type parameter.
+This type should be passed to `Messenger` as part of the 3rd type parameter, in a type union with internal events.
 
-🚫 **`never` is passed as the 5th type parameter**
+🚫 **`AllowedEvents` is exported**
```
mcmire: Perhaps for consistency this should say 

``` suggestion
🚫 **`AllowedEvents` is included in the actions type union but is exported**

File: docs/controller-guidelines.md [hunk 8]
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
-In this case, the controller should group these types into a type union so that they can be easily passed to the `RestrictedMessenger` type. However, it should not export this type, as it would then be re-exporting types that another package has already exported.
+In this case, the controller should group these types into a type union so that they can be easily passed to the `Messenger` type. However, it should not export this type, as it would then be re-exporting types that another package has already exported.
 
 The name of this type should be `AllowedActions`.
 
-This type should not only be passed to `RestrictedMessenger` as the 2nd type parameter, but should also be included in its 4th type parameter.
+This type should be passed to `Messenger` as part of the 2nd type parameter, in a type union with internal actions.
 
-🚫 **`never` is passed as the 4th type parameter**
+🚫 **`AllowedActions` is exported**
```
mcmire: Perhaps for consistency this should say:

``` suggestion
🚫 **`AllowedActions` is included in the actions type union but is exported**
```

docs/controller-guidelines.md
```diff
@@ -539,189 +543,178 @@ A controller should define and export a type union that holds all of its events.
...
+  FooControllerEvents
 >;
-
-/* === packages/foo-controller/src/index.ts === */
-
-export type { AllowedActions } from '@metamask/foo-controller';
 ```
 
-✅ **`AllowedActions['type']` is passed as the 4th type parameter, and `AllowedActions` is _not_ exported**
+✅ **`AllowedActions` is included in the 2nd type parameter but is _not_ exported**
```
mcmire: Since we no longer have an example where `AllowedActions` is included in a different type parameter, perhaps we should say:

``` suggestion
✅ **`AllowedActions` is included in the actions type union but is _not_ exported**
```

packages/sample-controllers/src/sample-gas-prices-controller.ts
mcmire, line 1: The controller class has some examples in JSDoc form so we should make sure to update those too.

packages/sample-controllers/src/sample-petnames-controller.ts
mcmire, line 1: The controller class has some examples in JSDoc form so we should make sure to update those too.

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
Gudahtt: Hmm, good point.

I could see that being a better option, though as you pointed out, the utility type would certainly be needed. Maybe we can consider working on this in a follow-up.

docs/controller-guidelines.md
Gudahtt: Oh? I think it makes more sense in this section about external events, since it's an example of putting these external events in the wrong place.

Maybe it would be simpler to combine these two sections? But it does seem like this example belongs here unless we do combine them.

docs/controller-guidelines.md
Gudahtt: Hmm, I don't think it's needed actually. Removed here: e59a0bb24

docs/controller-guidelines.md
Gudahtt: Related comment: https://github.com/MetaMask/core/pull/6335#discussion_r2308002924

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
```diff
@@ -300,8 +300,11 @@ describe('SampleGasPricesController', () => {
  * required by the controller under test.
  */
 type RootMessenger = Messenger<
-  ExtractAvailableAction<SampleGasPricesControllerMessenger>,
-  ExtractAvailableEvent<SampleGasPricesControllerMessenger>
+  // Use `string` rather than 'Root' here to allow registering actions and publishing events from
```
mcmire: Oh that's a good idea, I hadn't thought of that.

packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -151,56 +149,67 @@ export type SampleGasPricesControllerMessenger = RestrictedMessenger<
...
+ *   SampleGasPricesControllerEvents,
+ *   SampleGasPricesServiceActions,
+ *   SampleGasPricesServiceEvents
+ * } from '@metamask/sample-controllers';
  * import {
  *   SampleGasPricesController,
  *   SampleGasPricesService,
  *   selectGasPrices,
- * } from '@metamask/example-controllers';
+ * } from '@metamask/sample-controllers';
```
mcmire, line 169: Oops, forgot to update this reference in my PR. Good catch.

mcmire (approved):
LGTM!

docs/controller-guidelines.md
mcmire: Hmm, interesting. Perhaps I'll do another pass of this whole document later. Still, this is a good addition, so I approve.

docs/controller-guidelines.md
Gudahtt: Agreed, changed here: 9130d739f

docs/controller-guidelines.md
Gudahtt: Good idea, addressed here: 1a74ac854

docs/controller-guidelines.md
Gudahtt: Good idea, addressed here: 1a74ac854

docs/controller-guidelines.md
Gudahtt: Good idea, fixed here: 9130d739fd97c21fc7668f15d0891b47c53a5f53

packages/sample-controllers/src/sample-gas-prices-controller.ts
Gudahtt: I'll fix this separately as well, since this PR will be hanging around a while before we merge it

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
Gudahtt: It would have been a good idea if it worked 😅 I forgot there are runtime checks for this, tests are failing. Looking into it now.

packages/sample-controllers/src/sample-gas-prices-controller.ts
Gudahtt: Done here: https://github.com/MetaMask/core/pull/6421

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
Gudahtt: Alright, it works now.

cryptodev-2s (approved):
LGTM!

mikesposito:
nit: we should run `yarn update-readme-content` to refresh the README dependency graph

PR #6339, opened 2025-08-19
feat: Rename `ListenerV2` export to `StateChangeListener`

### LINKED ISSUE #6338: [base-controller] Rename `ListenerV2` to `StateChangeListener`
The state change listener type is still exported as `ListenerV2`, back from when all of our exports were suffixed with `V2` because of the two different `BaseController` versions. We forgot to drop the `V2` from this export when we dropped it from the others. Also this name is less descriptive than our other exports (it's not clear which lister it represents).

We should rename it to `StateChangeListener`, so that it's more clear what it is, and so that the obsolete "V2" is gone.

This will be a breaking change, so we will be grouping it with the breaking changes from https://github.com/MetaMask/core/issues/5626

Gudahtt:
## Explanation

The `ListenerV2` export of the `next` version of `BaseController` has been renamed to `StateChangeListener`.

## References

Fixes #6338

## Checklist


cryptodev-2s (approved):
LGTM!

mcmire (approved):
Makes sense.

PR #6343, opened 2025-08-19
Document recent breaking change to controller-utils

mcmire:
## Explanation

Commit 5e541a266d268d0306b1eac72fd25b47cb307adb introduced inadvertent breaking changes to `controller-utils` and `network-controller`. Specifically:

- `createServicePolicy` returns an object with type `ServicePolicy`, and this object has an `onDegraded` method on it. Previous to this commit, the event listener that `onDegraded` took expected no arguments, but the commit changed the signature of the listener such that it now expected either no arguments _or_ an object with either an `error` or `value` property.
- The `AbstractRpcService` type in `network-controller` builds on `ServicePolicy` for its `on*` hooks. That means that this commit also changed the signature of the listener for `onDegraded` so it now expected an object with either `endpointUrl`, `error` + `endpointUrl` or `value` + `endpointUrl`.

For posterity, we discovered these breaking changes within `eth-json-rpc-middleware`. This package contains a copy of `AbstractRpcService`, and we found that it was not compatible with `network-controller` when upgrading its dependency on `controller-utils` to the newest version.

Re-releasing these breaking changes under major versions would be painful as it would affect nearly all packages under the `core` monorepo. So instead, we merely document them in the changelogs here.

## Checklist


Gudahtt (approved):
LGTM!

cryptodev-2s (approved):
LGTM!

PR #6344, opened 2025-08-20
feat: add backup & sync capabilities to `AccountTreeController`

mathieuartu:
## Checklist


mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

packages/account-tree-controller/package.json
```diff
@@ -73,6 +76,8 @@
     "@metamask/account-api": "^0.9.0",
     "@metamask/accounts-controller": "^33.0.0",
     "@metamask/keyring-controller": "^23.0.0",
+    "@metamask/multichain-account-service": "^0.6.0",
```
ccharly: We should have it as a peer dep too.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -82,24 +101,45 @@ export class AccountTreeController extends BaseController<
...
   readonly #rules: [EntropyRule, SnapRule, KeyringRule];
 
+  readonly #trace: TraceCallback;
+
+  readonly #backupAndSyncConfig: {
+    emitBackupAndSyncEvent: (
+      event: BackupAndSyncEmitAnalyticsEventParams,
+    ) => void;
+    enableDebugLogging: boolean;
+  };
```
ccharly: Would be great to have a type for this. So we can re-use it here and in the tests IMO.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -82,24 +101,45 @@ export class AccountTreeController extends BaseController<
...
+
+  // Temporary: ensures we can release updates to AccountTreeController without
+  // breaking changes while we transition to the new multichain syncing approach.
+  readonly #disableMultichainAccountSyncing: boolean = false;
```
ccharly: Nit (maybe?): But I think it's simpler to use "enable*" over "disable*". Mostly to avoid the "double-negative" expressions (if "not disabled" instead if "enabled"). But that could personal preference maybe 😅

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -991,6 +1068,24 @@ export class AccountTreeController extends BaseController<
...
+  }
+
+  /**
+   * Clears all persisted metadata and syncing state.
+   *
+   * This will reset the account groups and wallets metadata, as well as
+   * the syncing state. This should be used when we want to
+   * completely reset the controller's state.
+   */
+  clearPersistedMetadataAndSyncingState(): void {
```
ccharly: I've seen other controllers using `clearState` too.. But maybe this is very specific to B&S here?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -991,6 +1068,24 @@ export class AccountTreeController extends BaseController<
...
+    this.update((state) => {
+      state.accountGroupsMetadata = {};
+      state.accountWalletsMetadata = {};
```
ccharly: We're clearing groups/wallets metadata, but we're not "resetting" the account-tree itself. Which means each nodes might still hold some previous names (or old pinned/hidden state values) that we won't be able to replicate anymore (since we've dropped their associated persisted values).

If we plan to keep method. We should probably fire a `this.init()` afterward to reset the tree metadata? 🤔

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1032,4 +1127,36 @@ export class AccountTreeController extends BaseController<
...
+   */
+  #createBackupAndSyncContext(): BackupAndSyncContext {
+    return {
+      controller: this,
+      messenger: this.messagingSystem,
+      controllerStateUpdateFn: this.update.bind(this),
+      traceFn: this.#trace.bind(this),
+      emitAnalyticsEventFn:
+        this.#backupAndSyncConfig.emitBackupAndSyncEvent.bind(this),
+      enableDebugLogging: this.#backupAndSyncConfig.enableDebugLogging,
```
ccharly: We could have passed the `#backupAndSyncConfig` object directly? 😅

packages/account-tree-controller/src/backup-and-sync/analytics/segment.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  BackupAndSyncAnalyticsEvents,
+  formatAnalyticsEvent,
+  type BackupAndSyncAnalyticsEvent,
+  type BackupAndSyncEmitAnalyticsEventParams,
+  type BackupAndSyncAnalyticsEventPayload,
+} from './segment';
+
+describe('BackupAndSyncAnalytics - Segment', () => {
+  describe('BackupAndSyncAnalyticsEvents', () => {
+    it('should contain all expected event names', () => {
```
ccharly: We usually use the [imperative form for all tests now](https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#use-it-to-specify-the-desired-behavior-for-the-code-under-test).

(I know that a lot of old tests are still using this though 😅)

packages/account-tree-controller/src/backup-and-sync/analytics/segment.ts
```diff
@@ -0,0 +1,51 @@
+export const BackupAndSyncAnalyticsEvents = {
+  WALLET_RENAMED: 'wallet_renamed',
+  GROUP_ADDED: 'group_added',
+  GROUP_RENAMED: 'group_renamed',
+  GROUP_HIDDEN_STATUS_CHANGED: 'group_hidden_status_changed',
+  GROUP_PINNED_STATUS_CHANGED: 'group_pinned_status_changed',
+  LEGACY_SYNCING_DONE: 'legacy_syncing_done',
+  LEGACY_GROUP_ADDED_FROM_ACCOUNT: 'legacy_group_added_from_account',
+  LEGACY_GROUP_RENAMED: 'legacy_group_renamed',
+} as const;
```
ccharly: Any reason to not use an `enum` here? I guess that has to do with the event names being just `string` for this kind of segment event?

But maybe an `enum` would have still worked in this context? 🤔

ccharly: Also, we might want to use the same naming convention here, so use of `PascalCase` for event key too? (Similarly to your `TraceName` const object)

ccharly: And lastly, we could use "singular" for the "enum" name too: `BackupAndSyncAnalyticsEvent`.

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
```diff
@@ -0,0 +1,126 @@
...
+    return this.#queue.length;
+  }
+
+  /**
+   * Checks if queue processing is currently in progress.
+   *
+   * @returns True if processing is in progress.
+   */
+  get isProcessing(): boolean {
+    return this.#isProcessingInProgress;
```
ccharly, line 116: I wonder if we could just use `this.#queue.length > 0` here?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,408 @@
...
+   *
+   * @throws Will throw if the sync operation encounters unrecoverable errors
+   */
+  async performFullSync(): Promise<void> {
+    if (this.#context.disableMultichainAccountSyncing) {
+      if (this.#context.enableDebugLogging) {
+        contextualLogger.warn(
+          'Multichain account syncing is disabled. Skipping full sync operation.',
+        );
+      }
```
ccharly: If we plan to keep this logger, I think it would be better to have the `enableDebugLogging` at logger-level. This way we would just "log" everytime, but depending on the flag itself, the logger would display something or not.

WDYT?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,408 @@
...
+            'Error during multichain account syncing:',
+            error,
+          );
+        }
+        throw error;
+      } finally {
+        this.#context.controllerStateUpdateFn(
+          (state: AccountTreeControllerState) => {
+            state.isAccountTreeSyncingInProgress = false;
+            state.hasAccountTreeSyncingSyncedAtLeastOnce = true;
```
ccharly: I think we would hit the `finally` no matter what. Which means that we would still set `state.hasAccountTreeSyncingSyncedAtLeastOnce = true;` even for errors.

Is that intended? If yes, I'd add a comment to explain this.

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -9,10 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add `syncInternalAccountsWithUserStorage` action handler so it can be used through the messaging system ([#6344](https://github.com/MetaMask/core/pull/6344))
```
mathieuartu: I need to remove this in the next commit

packages/account-tree-controller/src/backup-and-sync/authentication/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Retrieves the profile ID from AuthenticationController.
+ *
+ * @param context - The backup and sync context.
+ * @param entropySourceId - The optional entropy source ID.
+ * @returns The profile ID associated with the session, if available.
+ */
+export const getProfileId = async (
+  context: BackupAndSyncContext,
+  entropySourceId?: string,
+) => {
```
ccharly: We should probably enforce the return type here. Normally it should be inferred as `string | undefined` (IDK if we have a `ProfileId` type-alias somewhere for this?)

Also, I've have seen some use of this function and we do not check for the `undefined` case, are we always assuming the profile ID to be always available?

If we need to enforce this, then should we check for it there and `throw` if it's not available? 🤔

packages/profile-sync-controller/src/controllers/user-storage/contact-syncing/controller-integration.test.ts
```diff
@@ -57,7 +57,6 @@ const baseState = {
   isBackupAndSyncUpdateLoading: false,
   hasAccountSyncingSyncedAtLeastOnce: false,
   isAccountSyncingReadyToBeDispatched: false,
```
mathieuartu: Need to remove this as well

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,408 @@
...
+              !walletFromUserStorage ||
+              !walletFromUserStorage.isLegacyAccountSyncingDisabled
+            ) {
+              // 2.2 Perform legacy account syncing
+              // This will migrate legacy account data to the new structure.
+              // This operation will only be performed once.
+              await performLegacyAccountSyncing(
+                this.#context,
+                entropySourceId,
+                walletProfileId,
```
ccharly, line 346: This is what I meant by using the profile ID without checking for `undefined`, cause technically speaking, it could be `undefined` here, right?

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: This is actually a leftover and will be removed in the next commit. Thanks for flagging!

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: Let's discuss, because if we add a .init() call at the end like you suggested in your next comment, then we probably can get away with clearState(). I just need to make sure what exactly does this method do in terms of default values

packages/account-tree-controller/src/backup-and-sync/analytics/segment.ts
mathieuartu: It's the old `enum` vs `object as const` debate. While I don't have a strong opinion, I found out that we tended to want to remove enums from the codebase recently. WDYT?
Regarding your other points, agreed! Will do

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
mathieuartu: Probably. Good point!

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: Right, that would unclutter a lot of UTs also, agreed! Fair point

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: It is indeed. This flag exists in order to prevent any atomic updates before the first sync has completed (wether successfully or not) so that we don't end up with weird race conditions

packages/account-tree-controller/package.json
mathieuartu: Is it not already in those? 😅

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: Valid! Adding an init call in here. This method is supposed to be called from within the clients when resetting the wallet. I'll add more info in the JSDOC also.

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
mathieuartu: After investigating, this flag serves as a mutex. If we used `this.#queue.length > 0`, as soon as the first item is processed and removed from the queue, a new call to process() could start processing the remaining items concurrently.
Also the flag is set to true at the start of processing and only set to false in the finally block. During the processing loop, the queue length changes as items are removed, but we still want to prevent new processing from starting.

packages/account-tree-controller/src/backup-and-sync/authentication/utils.ts
mathieuartu: It's not necessary for service execution, since `profileId` is only used to log segment events. I've added a try catch block so that we return `undefined` in case the underlying call throws. Also added better types.

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
```diff
@@ -0,0 +1,120 @@
...
+    // Block enqueueing if big sync is running or if no initial sync has occurred
+    if (
+      this.#context.controller.state.isAccountTreeSyncingInProgress ||
+      !this.#context.controller.state.hasAccountTreeSyncingSyncedAtLeastOnce
+    ) {
+      return;
+    }
+
+    const syncEvent: AtomicSyncEvent = {
+      id: `sync-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`,
```
fabiobozzo: Double **nitpick** ✌️ 
- `substr` is deprecated afaics, maybe `substring` instead?
- random id is crypto-weak here

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
```diff
@@ -0,0 +1,120 @@
...
+          break;
+        }
+
+        try {
+          await event.execute();
+        } catch (error) {
+          this.#context.contextualLogger.error(
+            `Failed to process atomic sync event ${event.id}`,
+            error,
+          );
```
fabiobozzo: Would exponential backoff help here w/ repeated failures?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,364 @@
...
+   */
+  enqueueSingleWalletSync(walletId: AccountWalletId): void {
+    if (
+      !this.#context.controller.state.hasAccountTreeSyncingSyncedAtLeastOnce
+    ) {
+      return;
+    }
+
+    this.#atomicSyncQueue.enqueue(() =>
+      this.#performSingleWalletSync(walletId),
```
fabiobozzo: should call `#assertAccountWalletExists(walletId)` first?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,364 @@
...
+   * @param groupId - The group ID to sync.
+   */
+  enqueueSingleGroupSync(groupId: AccountGroupId): void {
+    if (
+      !this.#context.controller.state.hasAccountTreeSyncingSyncedAtLeastOnce
+    ) {
+      return;
+    }
+
+    this.#atomicSyncQueue.enqueue(() => this.#performSingleGroupSync(groupId));
```
fabiobozzo: you can also validate groupId existence here

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -118,6 +151,24 @@ export class AccountTreeController extends BaseController<
...
+      },
+      contextualLogger: new ContextualLogger({
+        isEnabled: config?.backupAndSync?.enableDebugLogging ?? false,
+      }),
+    };
+
+    // Initialize the syncing service
+    this.#syncingService = new BackupAndSyncService(
+      this.#createBackupAndSyncContext(),
+    );
```
fabiobozzo: Isn't that too early? Line 1168 passes `this.#groupIdToWalletId` to the context, but isn't that map used before the constructor is fully finished? Perhaps it's safer to init the service at the end of the constructor?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -953,6 +1017,9 @@ export class AccountTreeController extends BaseController<
       // Update tree node directly
       state.accountTree.wallets[walletId].metadata.name = name;
     });
+
+    // Trigger atomic sync for wallet rename
+    this.#syncingService.enqueueSingleWalletSync(walletId);
```
fabiobozzo: What if the sync operation fails (e.g. network errors, whatever)? wouldn't we end up in an out-of-sync situation, since local is updated and remote isn't? or will the next sync try again and rectify that?

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
mathieuartu: Great spot! But... I'm going to remove this as it does not serve any meaningful purpose 😄

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
mathieuartu: If we're here, it means that exponential backoff has already been tried without success

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: Since this is triggered from the `AccountTreeController.#insert` in case a new wallet was added, we can be sure it exists at this point. It can also be triggered from `AccountTreeController.setAccountWalletName` where `#assertAccountWalletExists` is called.

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: Same answer as above 🙂

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: It's passing down references so it's kept in "sync" the whole time. You're right to point out the order question, but `#groupIdToWalletId` already exists at this point in time.
Nevertheless, I'm moving those declarations a little bit below in the constructor, thanks!

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: Yes, but it will get picked up at the next wallet unlock / refresh, so it's not that critical.

packages/account-tree-controller/src/backup-and-sync/syncing/wallet.ts
```diff
@@ -0,0 +1,84 @@
...
+  // Track if we need to push this wallet to user storage
+  let shouldPushWallet = false;
+
+  // Compare and sync name metadata
+  const shouldPushForName = await compareAndSyncMetadata({
+    context,
+    localMetadata: walletPersistedMetadata?.name,
+    userStorageMetadata: walletFromUserStorage.name,
+    validateUserStorageValue: (value) =>
+      UserStorageSyncedWalletSchema.schema.name.schema.value.is(value),
```
ccharly, line 46: Is that equivalent to `is(UserStorageSyncedWalletSchema, value)`? (`is` being imported from `@metamask/superstruct`)

packages/account-tree-controller/src/backup-and-sync/syncing/wallet.ts
```diff
@@ -0,0 +1,84 @@
...
+  localWallet: AccountWalletEntropyObject,
+  walletFromUserStorage: UserStorageSyncedWallet | null | undefined,
+  profileId: ProfileId,
+): Promise<boolean> {
+  const walletPersistedMetadata =
+    context.controller.state.accountWalletsMetadata[localWallet.id];
+
+  if (!walletFromUserStorage) {
+    context.contextualLogger.warn(
+      `Wallet ${localWallet.id} did not exist in user storage, pushing to user storage...`,
```
ccharly, line 33: Very nit: Should this really be a `warn`? That could be a complete normal behavior if a user only has a local state and then enable B&S?

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
```diff
@@ -0,0 +1,76 @@
...
+    action: BackupAndSyncAnalyticsAction;
+    profileId: ProfileId;
+  };
+}): Promise<boolean> {
+  const localValue = localMetadata?.value;
+  const localTimestamp = localMetadata?.lastUpdatedAt;
+  const userStorageValue = userStorageMetadata?.value;
+  const userStorageTimestamp = userStorageMetadata?.lastUpdatedAt;
+
+  const isValueDifferent = localValue !== userStorageValue;
```
ccharly: Just checking, but we're comparing primitive types right? We never use more complex objects here? Just to make sure `!==` will cover it all (like, no need for deep-equality basically)

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  if (!isValueDifferent) {
+    return false; // No sync needed, values are the same
+  }
```
ccharly: Very nit: but I think we could have "flipped" the logic, like `if (isSameValue)` instead, to avoid the double negation here.

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const isUserStorageMoreRecent =
+    localTimestamp &&
+    userStorageTimestamp &&
+    localTimestamp < userStorageTimestamp;
```
ccharly, line 57: 🤔 if we have something in the user storage, but nothing locally, this will still be `false` right?

ccharly: But I guess, it's being covered with the `if ((... || !localMetadata) ...)` below?

packages/account-tree-controller/src/backup-and-sync/syncing/legacy.ts
```diff
@@ -0,0 +1,106 @@
...
+        )}`,
+      );
+      continue;
+    }
+    const localGroupId = getUUIDFromAddressOfNormalAccount(a);
+
+    if (n) {
+      // Find the local group that corresponds to this EVM address
+      const localGroup = localAccountGroups.find((group) =>
+        group.accounts.some((accountId) => accountId === localGroupId),
```
ccharly:
```suggestion
    const localAccountId = getUUIDFromAddressOfNormalAccount(a);

    if (n) {
      // Find the local group that corresponds to this EVM address
      const localGroup = localAccountGroups.find((group) =>
        group.accounts.some((accountId) => accountId === localAccountId),
```

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
fabiobozzo: Hooo.. right, I get it now, sorry 👀

packages/account-tree-controller/src/backup-and-sync/syncing/legacy.ts
```diff
@@ -0,0 +1,106 @@
...
+        profileId,
+        BackupAndSyncAnalyticsEvent.LegacyGroupAddedFromAccount,
+      );
+    }
+  }
+
+  // 3. Rename account groups if needed
+  const localAccountGroups = getLocalGroupsForEntropyWallet(
+    context,
+    `entropy:${entropySourceId}`,
```
ccharly: Nit: We do have a `toMultichainAccountWalletId` which already creates this kind of ID, but this is fine too (especially if you keep the naming `*EntropyWallet` IMO

(Just wanted to share just in case you preferred to mention "multichain account groups")

packages/account-tree-controller/src/backup-and-sync/syncing/legacy.ts
```diff
@@ -0,0 +1,106 @@
...
+      context.contextualLogger.warn(
+        `Legacy account data is missing name or address, skipping account: ${JSON.stringify(
+          legacyAccount,
+        )}`,
+      );
+      continue;
+    }
+    const localGroupId = getUUIDFromAddressOfNormalAccount(a);
+
+    if (n) {
```
ccharly: Very nit: But we can move this:

```suggestion
    if (n) {
      const localGroupId = getUUIDFromAddressOfNormalAccount(a);
      
```

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,293 @@
...
+    const didGroupAlreadyExist = Object.values(
+      context.controller.state.accountTree.wallets[`entropy:${entropySourceId}`]
+        .groups,
+    ).some((group) => group.metadata.entropy.groupIndex === groupIndex);
+
+    if (didGroupAlreadyExist) {
+      context.contextualLogger.warn(
+        `Group with index ${groupIndex} (wallet entropy ${entropySourceId}) already exists, skipping creation`,
+      );
+      continue; // Skip creating group if it already exists
```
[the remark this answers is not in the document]
mathieuartu: True, although this shouldn't happen in this PR context. Nevertheless, this can happen if a developer imports the `createLocalGroupsFromUserStorage` and uses it with an `entropySourceId` that doesn't correspond to a local wallet. Fixed with [8a5b6ca](https://github.com/MetaMask/core/pull/6344/commits/8a5b6ca33c7aacec6494978fe3fa062f07f89619)

packages/profile-sync-controller/package.json
```diff
@@ -132,7 +131,6 @@
     "webextension-polyfill": "^0.12.0"
   },
   "peerDependencies": {
-    "@metamask/accounts-controller": "^33.0.0",
```
ccharly, line 135: We're removing this controller, but I've seen some reference to the `address-book-controller` and `remote-feature-flag-controller`, they might have to be peer dep too now?

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Add `#deferredLogin` method that ensures only one login operation executes at a time using Promise map caching
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.3.0` ([#6355](https://github.com/MetaMask/core/pull/6355), [#6465](https://github.com/MetaMask/core/pull/6465))
 
+### Removed
+
+- **BREAKING:** Remove `@metamask/accounts-controller` peer dependency ([#6344](https://github.com/MetaMask/core/pull/6344))
+- **BREAKING:** Remove `@metamask/remote-feature-flag-controller` peer dependency ([#6344](https://github.com/MetaMask/core/pull/6344))
```
ccharly: Should this be `Add` instead? And add another one for `address-book-controller`?

packages/account-tree-controller/src/backup-and-sync/syncing/wallet.ts
mathieuartu: From JSDOC, my logic does `Check if a value passes the struct's validation.`, whereas `is` does `Check if a value passes a struct.` There's a slight difference, but the logic I've used is IMO more appropriate to what we want to do here

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2744,4 +2827,127 @@ describe('AccountTreeController', () => {
...
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+
+      await controller.syncWithUserStorage();
+
+      expect(performFullSyncSpy).toHaveBeenCalledTimes(1);
+
+      performFullSyncSpy.mockRestore();
```
ccharly: Nit: I think those mocks are automatically "restored" no? Or you had to manually restore them? Pretty sure someone told me that, but this might be hidden in some configuration somewhere 🤔

Anyway, would `resetAllMocks` do the trick instead of manual restores here?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -131,6 +164,25 @@ export class AccountTreeController extends BaseController<
...
     ];
 
+    // Initialize trace function
+    this.#trace = config?.trace ?? traceFallback;
+
+    // Initialize backup and sync config
+    this.#backupAndSyncConfig = {
+      emitAnalyticsEventFn: (event: BackupAndSyncEmitAnalyticsEventParams) => {
+        const formattedEvent = formatAnalyticsEvent(event);
+        return config?.backupAndSync?.onBackupAndSyncEvent?.(formattedEvent);
```
ccharly: (Very nit maybe?): We could maybe flip this and check for `config?.backupAndSync?.onBackupAndSyncEvent` first? If it's defined, then we call it with the formatted event?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -929,6 +991,9 @@ export class AccountTreeController extends BaseController<
           name;
       }
     });
+
+    // Trigger atomic sync for group rename
+    this.#syncingService.enqueueSingleGroupSync(groupId);
```
ccharly: Should this be guarded for entropy wallet only? 🤔

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -953,6 +1018,9 @@ export class AccountTreeController extends BaseController<
       // Update tree node directly
       state.accountTree.wallets[walletId].metadata.name = name;
     });
+
+    // Trigger atomic sync for wallet rename
+    this.#syncingService.enqueueSingleWalletSync(walletId);
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -981,6 +1049,9 @@ export class AccountTreeController extends BaseController<
           pinned;
       }
     });
+
+    // Trigger atomic sync for group pinning
+    this.#syncingService.enqueueSingleGroupSync(groupId);
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1009,6 +1080,25 @@ export class AccountTreeController extends BaseController<
           hidden;
       }
     });
+
+    // Trigger atomic sync for group hiding
+    this.#syncingService.enqueueSingleGroupSync(groupId);
```

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
mathieuartu: Yes... For now. We could probably reinforce the comparison here by using `fast-deep-equal` like other packages in the monorepo.

packages/profile-sync-controller/package.json
mathieuartu: Right... So it appears we let something slide in the past, and never added `address-book-controller` as a peerDependency... Will fix. `accounts-controller` and `remote-feature-flag-controller` are not needed anymore though.

packages/profile-sync-controller/CHANGELOG.md
mathieuartu: Nah, we're removing those. But you're right about adding an entry for `address-book-controller` as I mentioned in the comment above 🙂

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: I'm struggling to find the benefit of doing this here. Would it be to skip the const assignment in memory? Why would we want to do this? Genuine asks! 🙂

ccharly: I guess, it's more my old way of seeing things in different languages 😅 but basically, avoiding to "construct" (here the `formattedEvent`) if we don't need to construct it.

I think that's ok here, cause constructing is pretty cheap. But let's take an extreme example where formatting it would take up to 1s, then you would not want to compute it for nothing (nothing being here: `config?.backupAndSync?.onBackupAndSyncEvent` being evaluated to `undefined`)

But that's ok, I've checked the `formatAnalyticsEvent` and that's a simple construction, so no worries.

packages/profile-sync-controller/package.json
ccharly: Then I think we should remove it from the `tsconfig.*` file too? Pretty sure I saw `address-book` one + the `remote-feature-flag` (`accounts-controller` got removed though).

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
ccharly: Sounds like a good idea to me, given that everything has been written in a generic way!

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -83,24 +104,36 @@ export class AccountTreeController extends BaseController<
 
   readonly #groupIdToWalletId: Map<AccountGroupId, AccountWalletId>;
 
+  /**
+   * Service responsible for all backup and sync operations.
+   */
+  readonly #syncingService: BackupAndSyncService;
```
ccharly: WDYT about renaming this `#backupAndSyncService`? This would match its type, its context type and its config

packages/profile-sync-controller/package.json
mathieuartu: Added `address-book-controller` to the `package.json` as it should have been from the beginning (slipped when reviewing the contact syncing PR 🙄) and removed all mentions to `remote-feature-flag` + `accounts` controllers

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,364 @@
...
+  get isInProgress(): boolean {
+    return this.#context.controller.state.isAccountTreeSyncingInProgress;
+  }
+
+  /**
+   * Enqueues a single wallet sync operation.
+   *
+   * @param walletId - The wallet ID to sync.
+   */
+  enqueueSingleWalletSync(walletId: AccountWalletId): void {
```
ccharly, line 168: WDYT about using `MultichainAccountWalletId` here instead? This way this narrow down the `wallet.id` that can be used.

So all calls from the controller would have to be using `wallet.id` instead, and that `wallet.type === AccountWalletType.Entropy`. Each of those nodes should be strongly-typed accordingly! (Unless I messed it up somewhere 😅)

We can do the same with `AccountGroupId` -> `MultichainAccountGroupId`.

The reasoning for those 2 small changes here, is to enforce that we're dealing with multichain account wallet/group. (We already check this at controller level, but enforcing the types here, would "naturally" enforce this even here).

WDYT?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,364 @@
...
+      this.#performSingleWalletSync(walletId),
+    );
+  }
+
+  /**
+   * Enqueues a single group sync operation.
+   *
+   * @param groupId - The group ID to sync.
+   */
+  enqueueSingleGroupSync(groupId: AccountGroupId): void {
```
ccharly, line 190: Same here than:
- https://github.com/MetaMask/core/pull/6344/files#r2331213916

packages/account-tree-controller/src/backup-and-sync/analytics/segment.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param params.action - The action being performed.
+ * @param params.profileId - The profile ID associated with the event.
+ * @param params.additionalDescription - Optional additional description for the event.
+ *
+ * @returns The formatted event payload.
+ */
+export const formatAnalyticsEvent = ({
+  action,
+  profileId,
+  additionalDescription = '',
```
ccharly: If we use this, then we would always have an empty description if we don't provide it to `formatAnalyticsEvent`, is that ok? Should we keep it `undefined`? Should we do something like:

```ts
return {
  feature_name: BACKUP_AND_SYNC_EVENT_FEATURE_NAME,
  action,
  profile_id: profileId,
  ...(additionalDescription !== undefined ? { additional_description: additionalDescription } : {}),
};
```

This could be overkill, but I honestly don't know Segment events that much 😅

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,308 @@
...
+      context.contextualLogger.warn(
+        `Wallet with entropy ${entropySourceId} does not exist, skipping group creation`,
+      );
+      continue;
+    }
+
+    const didGroupAlreadyExist = Object.values(
+      context.controller.state.accountTree.wallets[
+        toMultichainAccountWalletId(entropySourceId)
+      ].groups,
```
ccharly:
```suggestion
      wallet.groups,
```

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: As discussed this evening and as we tried to do it, it creates some coercion issues. We agreed to work on this on a subsequent PR as there are runtime logic gates in place already.

packages/account-tree-controller/src/backup-and-sync/service/index.ts
mathieuartu: See https://github.com/MetaMask/core/pull/6344#discussion_r2331266286

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,308 @@
...
+    applyLocalUpdate: (name: string) => {
+      context.controller.setAccountGroupName(localGroup.id, name);
+    },
+    analytics: {
+      action: BackupAndSyncAnalyticsEvent.GroupRenamed,
+      profileId,
+    },
+  });
+
+  shouldPushGroup = shouldPushGroup || shouldPushForName;
```
ccharly: Nit:
```suggestion
  shouldPushGroup ||= shouldPushForName;
```

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,308 @@
...
+    applyLocalUpdate: (pinned: boolean) => {
+      context.controller.setAccountGroupPinned(localGroup.id, pinned);
+    },
+    analytics: {
+      action: BackupAndSyncAnalyticsEvent.GroupPinnedStatusChanged,
+      profileId,
+    },
+  });
+
+  shouldPushGroup = shouldPushGroup || shouldPushForPinned;
```
ccharly: Nit:

```suggestion
  shouldPushGroup ||= shouldPushForPinned;
```

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,308 @@
...
+    applyLocalUpdate: (hidden: boolean) => {
+      context.controller.setAccountGroupHidden(localGroup.id, hidden);
+    },
+    analytics: {
+      action: BackupAndSyncAnalyticsEvent.GroupHiddenStatusChanged,
+      profileId,
+    },
+  });
+
+  shouldPushGroup = shouldPushGroup || shouldPushForHidden;
```
ccharly:
```suggestion
  shouldPushGroup ||= shouldPushForHidden;
```

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,308 @@
...
+  profileId: ProfileId,
+): Promise<void> {
+  // Sort groups from user storage by groupIndex in ascending order
+  groupsFromUserStorage.sort((a, b) => a.groupIndex - b.groupIndex);
+
+  let previousGroupIndex = -1;
+  for (const groupFromUserStorage of groupsFromUserStorage) {
+    const { groupIndex } = groupFromUserStorage;
+
+    if (typeof groupIndex !== 'number' || groupIndex < 0) {
```
ccharly: Isn't `groupIndex` already of type `number`? The `UserStorageSyncedWalletGroup.groupIndex` should already been inferred correctly IMO

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,306 @@
...
+      context.contextualLogger.warn(
+        `Invalid group index ${groupIndex} found in user storage, skipping`,
+      );
+      continue;
+    }
+
+    const isGroupIndexOutOfSequence =
+      groupIndex <= previousGroupIndex || groupIndex !== previousGroupIndex + 1;
```
ccharly: This will always be true at the first time no? Since `previousGroupIndex` is `-1`? 🤔

mathieuartu: If groupIndex is 0 (the expected first valid index):

- `groupIndex <= previousGroupIndex` → `0 <= -1` → false
- `groupIndex !== previousGroupIndex + 1` → `0 !== (-1 + 1)` → `0 !== 0` → false

ccharly: I read the `groupIndex + 1` sorry 🙈

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,306 @@
...
+    try {
+      await createMultichainAccountGroup(
+        context,
+        entropySourceId,
+        groupIndex,
+        profileId,
+        BackupAndSyncAnalyticsEvent.GroupAdded,
+      );
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (_) {
```
ccharly: I think you can just `catch` for this one:

```suggestion
    } catch {
```

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,306 @@
...
+ * @param context - The sync context containing controller and messenger.
+ * @param localGroup - The local group to sync.
+ * @param groupFromUserStorage - The group from user storage to compare against.
+ * @param profileId - The profile ID for analytics.
+ * @returns A promise that resolves to true if the group needs to be pushed to user storage.
+ */
+async function syncGroupMetadataAndCheckIfPushNeeded(
+  context: BackupAndSyncContext,
+  localGroup: AccountGroupMultichainAccountObject,
+  groupFromUserStorage: UserStorageSyncedWalletGroup | null | undefined,
```
ccharly, line 116: WDYT about not having `| null | undefined` in here, and make the check above in the main loop that iterates through groups?

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,306 @@
...
+ * @param context - The sync context containing controller and messenger.
+ * @param localGroup - The local group to sync.
+ * @param groupFromUserStorage - The group from user storage to compare against (or null if it doesn't exist).
+ * @param entropySourceId - The entropy source ID.
+ * @param profileId - The profile ID for analytics.
+ */
+export async function syncSingleGroupMetadata(
+  context: BackupAndSyncContext,
+  localGroup: AccountGroupMultichainAccountObject,
+  groupFromUserStorage: UserStorageSyncedWalletGroup | null,
```
ccharly, line 202: Same here, removing the `| null`

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,306 @@
...
+    if (!wallet) {
+      context.contextualLogger.warn(
+        `Wallet with entropy ${entropySourceId} does not exist, skipping group creation`,
+      );
+      continue;
+    }
+
+    const didGroupAlreadyExist = Object.values(wallet.groups).some(
+      (group) => group.metadata.entropy.groupIndex === groupIndex,
+    );
```
ccharly: Slightly faster approach, we should be able to also do:

```suggestion
    const didGroupAlreadyExist = wallet.groups[toMultichainAccountGroupId(wallet.id, groupIndex)] !== undefined;
```

packages/account-tree-controller/src/backup-and-sync/syncing/legacy.ts
```diff
@@ -0,0 +1,107 @@
...
+        )}`,
+      );
+      continue;
+    }
+
+    if (n) {
+      // Find the local group that corresponds to this EVM address
+      const localAccountId = getUUIDFromAddressOfNormalAccount(a);
+      const localGroup = localAccountGroups.find((group) =>
+        group.accounts.some((accountId) => accountId === localAccountId),
```
ccharly: Nit: Assuming `group` is a similar to a `AccountGroupObject`, then `accounts` is a list of account IDs, so `includes` should be enough?

```suggestion
        group.accounts.includes(localAccountId),
```

packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts
ccharly: WDYT if we move the `!localMetadata` inside the `isUserStorageMoreRecent` then?

We could have a comment saying "If we have not local metadata, then user storage is guaranteed to be more recent" (or something like this..)?

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
ccharly: Actually, I saw the other pattern with wallet (similar pattern), and I guess it ok to have it like this. This way we log + return `true` so we know we have to push to the user storage. 👍

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
ccharly: See my other comment. Resolving this too.

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,364 @@
...
+      // Get the specific group from user storage
+      const groupFromUserStorage = await getGroupFromUserStorage(
+        this.#context,
+        entropySourceId,
+        group.metadata.entropy.groupIndex,
+      );
+
+      await syncSingleGroupMetadata(
+        this.#context,
+        group as AccountGroupMultichainAccountObject,
```
ccharly: Do we need this type cast? Or maybe that's the issue we were discussing internally? The CI complains about this, while it works locally (I'm still suprised by this kind of thing 🥹)

If yes, then I'd add a comment explaining why we need to force the type-cast here if possible?

packages/account-tree-controller/src/backup-and-sync/user-storage/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const USER_STORAGE_WALLETS_FEATURE_KEY = 'multichain_accounts_wallets';
+export const USER_STORAGE_WALLETS_FEATURE_ENTRY_KEY = 'wallet';
+export const USER_STORAGE_GROUPS_FEATURE_KEY = 'multichain_accounts_groups';
```
ccharly: Maybe nit: Isn't `multichain_account_*` (singular for `account` here) more appropriate?

packages/account-tree-controller/src/backup-and-sync/user-storage/validation.ts
```diff
@@ -0,0 +1,88 @@
...
+  walletData: unknown,
+): asserts walletData is UserStorageSyncedWallet {
+  try {
+    assert(walletData, UserStorageSyncedWalletSchema);
+  } catch (error) {
+    if (error instanceof StructError) {
+      const validationFailures = error
+        .failures()
+        .map(({ path, message }) => `[${path.join('.')}] ${message}`)
+        .join(', ');
```
ccharly: We're using this 3 times here. Maybe we could have a `formatValidationErrorMessages` (or any other appropriate) helper function (does not have to be exported) so we can do:

```ts
      throw new Error(
        `Invalid user storage wallet data: ${formatValidationErrorMessages(error)}`,
      );
```

And then we use it for all `assert*` functions from this file?

packages/account-tree-controller/src/backup-and-sync/utils/contextual-logger.ts
```diff
@@ -0,0 +1,39 @@
+const LOG_PREFIX = '[AccountTreeController - Backup and sync] ';
+
+export class ContextualLogger {
```
ccharly: As discussed together, another alternative would be to use the logging system from `@metamask/utils`.

It would give use similar prefixes/naming.

packages/account-tree-controller/src/backup-and-sync/user-storage/constants.ts
mathieuartu: I thought about it in terms of `feature_name_feature_key` where the feature we're working on is `multichain_accounts`

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
mathieuartu: Hmmm no because if `null | undefined` then we need to push the local data, so it's still valuable

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
mathieuartu: same answer as https://github.com/MetaMask/core/pull/6344#discussion_r2331391854

packages/account-tree-controller/src/backup-and-sync/user-storage/network-utils.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Executes a network operation with retry logic for transient failures.
+ *
+ * @param operation - The async operation to execute.
+ * @param maxRetries - Maximum number of retry attempts.
+ * @param baseDelayMs - Base delay between retries in milliseconds.
+ * @returns Promise that resolves with the operation result.
+ */
+export async function executeWithRetry<T>(
```
ccharly, line 9: NOTE for later: We could move this kind of helpers on `@metamask/utils` someday maybe? 🤔

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -7,12 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Changed
 
 - Implement deferred login pattern in `SRPJwtBearerAuth` to prevent race conditions during concurrent authentication attempts ([#6353](https://github.com/MetaMask/core/pull/6353))
   - Add `#deferredLogin` method that ensures only one login operation executes at a time using Promise map caching
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.3.0` ([#6355](https://github.com/MetaMask/core/pull/6355), [#6465](https://github.com/MetaMask/core/pull/6465))
 
+### Removed
+
+- **BREAKING:** Remove `@metamask/accounts-controller` peer dependency ([#6344](https://github.com/MetaMask/core/pull/6344))
+- **BREAKING:** Remove `@metamask/remote-feature-flag-controller` peer dependency ([#6344](https://github.com/MetaMask/core/pull/6344))
```
ccharly: I don't see any change related to this controller on the `profile-sync-controller` actually? 🤔

mathieuartu: Good catch! I thought this was published, but not yet. Removing this entry!

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2744,4 +2827,123 @@ describe('AccountTreeController', () => {
...
+
+      const { controller } = setup({
+        accounts: [MOCK_HARDWARE_ACCOUNT_1], // Use hardware account to avoid entropy calls
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+
+      await expect(controller.syncWithUserStorage()).rejects.toThrow(
+        'Sync failed',
```
ccharly: Nit:

```suggestion
        syncError.message,
```

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,274 @@
...
+    const formattedGroup = formatGroupForUserStorageUsage(context, group);
+
+    context.contextualLogger.info(
+      `Pushing group to user storage: ${JSON.stringify(formattedGroup)}`,
+    );
+
+    return await context.messenger.call(
+      'UserStorageController:performSetStorage',
+      `${USER_STORAGE_GROUPS_FEATURE_KEY}.${formattedGroup.groupIndex}`,
+      JSON.stringify(formattedGroup),
```
ccharly: Just stringify-ing once:

```suggestion
    const formattedGroupPayload = JSON.stringify(formattedGroup);

    context.contextualLogger.info(
      `Pushing group to user storage: ${formattedGroupPayload}`,
    );

    return await context.messenger.call(
      'UserStorageController:performSetStorage',
      `${USER_STORAGE_GROUPS_FEATURE_KEY}.${formattedGroup.groupIndex}`,
      formattedGroupPayload,
```

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,274 @@
...
+    const formattedGroups = groups.map((group) =>
+      formatGroupForUserStorageUsage(context, group),
+    );
+    const entries: [string, string][] = formattedGroups.map((group) => [
+      String(group.groupIndex),
+      JSON.stringify(group),
+    ]);
+
+    context.contextualLogger.info(
+      `Pushing groups to user storage: ${JSON.stringify(formattedGroups)}`,
```
ccharly: Wondering if 1 `JSON.stringify` is better than re-using the ones we have in `entries` 🤔  IIRC we had pretty bad perf with `stringify` on mobile (but this might have been resolved, or maybe this was strictly related to the Snap platform...).

Any opinion?

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,274 @@
...
+    const entropySourceId = wallet.metadata.entropy.id;
+
+    context.contextualLogger.info(
+      `Pushing wallet to user storage: ${JSON.stringify(formattedWallet)}`,
+    );
+
+    return await context.messenger.call(
+      'UserStorageController:performSetStorage',
+      `${USER_STORAGE_WALLETS_FEATURE_KEY}.${USER_STORAGE_WALLETS_FEATURE_ENTRY_KEY}`,
+      JSON.stringify(formattedWallet),
```
ccharly: Same here, we could stringify only once.

packages/account-tree-controller/src/backup-and-sync/user-storage/constants.ts
ccharly: Oh ok indeed. I don't remember seeing any `multichain_accounts` feature constant. But we could make one to emphasis on this maybe? And re-use it like:

```ts
export const USER_STORAGE_MULTICHAIN_ACCOUNTS_FEATURE_KEY = 'multichain_accounts';
export const USER_STORAGE_MULTICHAIN_ACCOUNTS_WALLETS_FEATURE_KEY =
  `${USER_STORAGE_MULTICHAIN_ACCOUNTS_FEATURE_KEY}_wallets`;
...
```

Just a small suggestion, but I won't fight for it if you don't want that one! 😄

ccharly: Also, I've seen that we were using `.` sometimes as a separator, but maybe you want to keep wallets and groups in a different "hierarchy" thus, a different "parent feature key"?

Anyway, again, I'm ok with anything here.

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    if (!accountsData) {
+      return [];
+    }
+
+    const allAccounts = accountsData
+      .map((accountStringifiedJSON) => {
+        try {
+          return parseLegacyAccountFromUserStorageResponse(
+            accountStringifiedJSON,
```
ccharly: Very nit: But we could follow the same naming convention you used before? Not sure what's best (and again, that's nit)

```suggestion
      .map((stringifiedAccount) => {
        try {
          return parseLegacyAccountFromUserStorageResponse(
            stringifiedAccount,
```

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,277 @@
...
+  UserStorageSyncedWalletGroup,
+} from '../types';
+
+/**
+ * Retrieves the wallet from user storage.
+ *
+ * @param context - The backup and sync context.
+ * @param entropySourceId - The entropy source ID.
+ * @returns The wallet from user storage or null if not found or invalid.
+ */
```
ccharly, line 34: Technically, all those functions (from this file) calling `executeWithRetry` can throw, so we could add a `@throw` entry maybe?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,363 @@
...
+              error instanceof Error ? error.message : String(error),
+            );
+
+            // Attempt to rollback state changes for this wallet
+            try {
+              if (!stateSnapshot) {
+                throw new Error(
+                  `State snapshot is missing for wallet ${wallet.id}`,
+                );
+              }
```
ccharly, line 417: If anything goes wrong with the legacy sincing, we will always throw an error, cause we won't have any snapshot at this point in time.

Should we split the `try catch` into 2?
1. Around legacy syncing (if any)
2. Around multichain account syncing (where snapshot should always be guaranteed to be defined, as this is the first step we're doing before doing any sync).

IDK if there's any benefit doing this though.. I just noticed the outer `try catch` (which does re-throw).

Maybe the reported error could be be better with this split?

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,363 @@
...
+   *
+   * @param walletId - The wallet ID to sync.
+   */
+  async #performSingleWalletSync(walletId: AccountWalletId): Promise<void> {
+    try {
+      const wallet =
+        this.#context.controller.state.accountTree.wallets[walletId];
+      if (!wallet || wallet.type !== AccountWalletType.Entropy) {
+        return; // Only sync entropy wallets
+      }
```
ccharly: Very nit: We have this pattern twice, maybe we could wrap this in a `#getEntropyWallet(walletId): AccountWalletEntropyObject` so we just have to do `if (!wallet)` afterward? (Thus, enforcing the wallet type in one place).

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,363 @@
...
+      const walletId = this.#context.groupIdToWalletId.get(groupId);
+      if (!walletId) {
+        return;
+      }
+
+      const wallet =
+        this.#context.controller.state.accountTree.wallets[walletId];
+      if (!wallet || wallet.type !== AccountWalletType.Entropy) {
+        return; // Only sync entropy wallets
+      }
```
ccharly: Similar comment here than:
- https://github.com/MetaMask/core/pull/6344/files#r2332256764

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    // Encapsulate the sync logic in a function to allow tracing
+    const bigSyncFn = async () => {
+      try {
+        this.#context.controllerStateUpdateFn(
+          (state: AccountTreeControllerState) => {
+            state.isAccountTreeSyncingInProgress = true;
+          },
+        );
```
ccharly: IDK if this is really concurrent safe 🤔

I also don't know if we should have concurrent `performFullSync`, but:

(I'll name thing "threads" even though, they're not "true" threads 🙈)
1. (Thread 1) Calls `performFullSync`
  - Check `isInProgress` is false
  - `await traceFn(bigSync);`
    * The scheduler could change to a different thread
2. (Thread 2) Calls `performFullSync`
  - Check `isInProgress` is still false
  - `await traceFn(bigSync);`
    * We now have 2 on-going concurrent full-syncs
    * The scheduler resume Thread 1 work

Maybe a simple fix would be to move the state updates outside of `bigSyncFn`?

packages/profile-sync-controller/CHANGELOG.md
```diff
@@ -7,12 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Implement deferred login pattern in `SRPJwtBearerAuth` to prevent race conditions during concurrent authentication attempts ([#6353](https://github.com/MetaMask/core/pull/6353))
   - Add `#deferredLogin` method that ensures only one login operation executes at a time using Promise map caching
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.3.0` ([#6355](https://github.com/MetaMask/core/pull/6355), [#6465](https://github.com/MetaMask/core/pull/6465))
 
+### Removed
+
+- **BREAKING:** Remove `@metamask/accounts-controller` peer dependency ([#6344](https://github.com/MetaMask/core/pull/6344))
+- **BREAKING:** Remove all account syncing code & logic ([#6344](https://github.com/MetaMask/core/pull/6344))
+  - `UserStorageController` now only holds the account syncing enablement status, but the logic itself has been moved to `@metamask/account-tree-controller`
+- Remove `UserStorageController` optional config callback `getIsMultichainAccountSyncingEnabled`, and `getIsMultichainAccountSyncingEnabled` public method / messenger action
```
ccharly: Not really reported as an error by our tool, but this entry miss the PR number:

```suggestion
- Remove `UserStorageController` optional config callback `getIsMultichainAccountSyncingEnabled`, and `getIsMultichainAccountSyncingEnabled` public method / messenger action ([#6344](https://github.com/MetaMask/core/pull/6344))
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -11,6 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Bidirectional metadata synchronization for wallets and groups with user storage
+  - Automatic sync triggers on metadata changes (rename, pin/hide operations) with superstruct validation
+  - New `isBackupAndSyncInProgress` state property to track sync status
+  - Analytics event tracking and performance tracing for sync operations
+  - Rollback mechanism for failed sync operations with state snapshot/restore capabilities
+  - Support for entropy-based wallets with multichain account syncing
+  - Legacy account syncing compatibility for seamless migration
+  - Optional configuration through new `AccountTreeControllerConfig.backupAndSync` options
+  - Add `@metamask/superstruct` for data validation
+  - Add `@metamask/profile-sync-controller` and `@metamask/multichain-account-service` peer dependencies
```
ccharly: Can you move this to the `## [Unreleased]` section under a new `### Added` sub-section? That should fix the CI.

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -252,23 +280,26 @@ export class BackupAndSyncService {
...
+          name: TraceName.AccountSyncFull,
+        },
+        bigSyncFn,
+      );
+    } finally {
+      // Always reset state, regardless of success or failure
+      this.#context.controllerStateUpdateFn(
+        (state: AccountTreeControllerState) => {
+          state.isAccountTreeSyncingInProgress = false;
+          state.hasAccountTreeSyncingSyncedAtLeastOnce = true;
```
ccharly: This might not be true if we early fails with legacy syncing no?

mathieuartu: `hasAccountTreeSyncingSyncedAtLeastOnce` is more of a convenience flag that prevents groups & wallets from being pushed before we at least tried to perform one big sync

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1110,19 +1110,15 @@ export class AccountTreeController extends BaseController<
   }
 
   /**
-   * Clears all persisted metadata and syncing state.
-   *
-   * This will reset the account groups and wallets metadata, as well as
-   * the syncing state. It also calls `init()` to reinitialize the controller.
-   * This should be used when we want to completely reset the controller's state.
+   * Clears the controller state and resets to default values.
```
mathieuartu, line 1107: TODO: Also add a comment that explains it also resets `BackupAndSyncService`'s state

mathieuartu:
@metamaskbot publish-preview

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
```diff
@@ -0,0 +1,123 @@
...
+   */
+  enqueue(syncFunction: () => Promise<void>): Promise<void> {
+    let resolvePromise: (() => void) | undefined;
+    let rejectPromise: ((error: unknown) => void) | undefined;
+
+    // Create promise that resolves when the sync function completes
+    const promise = new Promise<void>((resolve, reject) => {
+      resolvePromise = resolve;
+      rejectPromise = reject;
+    });
```
ccharly: Nit: We could have used https://github.com/MetaMask/utils/blob/main/src/promise.ts#L43

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -0,0 +1,554 @@
...
+        this.#firstOngoingFullSyncPromise,
+      );
+    }
+
+    // Create a new ongoing sync (sequential calls after previous completed)
+    const newSyncPromise = this.#atomicSyncQueue.clearAndEnqueue(() =>
+      this.#performFullSyncInner(),
+    );
+
+    return this.#setupOngoingPromiseCleanup(newSyncPromise);
```
ccharly: Nit: But I think we could have use this instead:

```suggestion
    // Create a new ongoing sync (sequential calls after previous completed)
    const newSyncPromise = this.#atomicSyncQueue.clearAndEnqueue(() =>
      this.#performFullSyncInner(),
    );

    // First sync setup - create and cache the first sync promise
    if (!this.#firstOngoingFullSyncPromise) {
      this.#firstOngoingFullSyncPromise = newSyncPromise;
    }

    return this.#setupOngoingPromiseCleanup(newSyncPromise);
```

This should be same, it's just that `#firstOngoingFullSyncPromise` will hold a reference to the very first `newSyncPromise` until its cleared.

packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -0,0 +1,267 @@
...
+    ...groupsFromUserStorage.map((g) => g.groupIndex),
+  );
+
+  for (
+    let groupIndex = 0;
+    groupIndex <= numberOfAccountGroupsToCreate;
+    groupIndex++
+  ) {
+    try {
+      await createMultichainAccountGroup(
```
ccharly, line 85: Nit: We could maybe add a small comment to mention idempotency? Something like:

```suggestion
      // Creating multichain account group is idempotent, so we can safely
      // re-create every groups starting from 0.
      await createMultichainAccountGroup(
```

packages/account-tree-controller/src/backup-and-sync/user-storage/network-operations.ts
```diff
@@ -0,0 +1,289 @@
...
+      entropySourceId,
+    );
+    if (!groupData) {
+      return [];
+    }
+
+    const allGroups = groupData
+      .map((groupStringifiedJSON) => {
+        try {
+          return parseGroupFromUserStorageResponse(groupStringifiedJSON);
```
ccharly: Very super nit: The "legacy" counterpart (for accounts) of this function uses a different name (despite being super similar), so just in case you want, we could use the same here:

```suggestion
      .map((stringifiedGroup) => {
        try {
          return parseGroupFromUserStorageResponse(stringifiedGroup);
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Bidirectional metadata synchronization for wallets and groups with user storage.
+  - Automatic sync triggers on metadata changes (rename, pin/hide operations) with superstruct validation.
+  - New `isBackupAndSyncInProgress` state property to track sync status.
+  - Analytics event tracking and performance tracing for sync operations.
+  - Rollback mechanism for failed sync operations with state snapshot/restore capabilities.
+  - Support for entropy-based wallets with multichain account syncing.
+  - Legacy account syncing compatibility for seamless migration.
+  - Optional configuration through new `AccountTreeControllerConfig.backupAndSync` options.
+  - Add `@metamask/superstruct` for data validation.
+  - Add `@metamask/profile-sync-controller` and `@metamask/multichain-account-service` peer dependencies.
```
ccharly: Those are peer dep, IMO they should be on a their own changelog entry (as breaking too)

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Add backup and sync capabilities ([#6344](https://github.com/MetaMask/core/pull/6344))
+  - New `syncWithUserStorage()` method for manual sync triggers, replacing `UserStorageController:syncInternalAccountsWithUserStorage` usage in clients.
+  - `BackupAndSyncService` with full and atomic sync operations for account tree data persistence.
+  - Bidirectional metadata synchronization for wallets and groups with user storage.
+  - Automatic sync triggers on metadata changes (rename, pin/hide operations) with superstruct validation.
```
ccharly: Nit: But I don't think we need to mention `superstruct` here?
```suggestion
  - Automatic sync triggers on metadata changes (rename, pin/hide operations).
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Add backup and sync capabilities ([#6344](https://github.com/MetaMask/core/pull/6344))
+  - New `syncWithUserStorage()` method for manual sync triggers, replacing `UserStorageController:syncInternalAccountsWithUserStorage` usage in clients.
```
```suggestion
  - New `syncWithUserStorage()` and `syncWithUserStorageAtLeastOnce()` methods for manual sync triggers, replacing `UserStorageController:syncInternalAccountsWithUserStorage` usage in clients.
```

packages/account-tree-controller/src/backup-and-sync/service/atomic-sync-queue.ts
mathieuartu: How tf do you know all those exist? 😄 Indeed!

PR #6355, opened 2025-08-21
Release 508.0.0

Gudahtt:
This is the initial release of the `@metamask/messenger` package and the experimental `next` export of `@metamask/base-controller`.

### DIFF
--- a/packages/base-controller/package.json
+++ b/packages/base-controller/package.json
@@ -57,7 +57,7 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
   "dependencies": {
-    "@metamask/messenger": "^0.0.0",
+    "@metamask/messenger": "^0.1.0",
     "@metamask/utils": "^11.4.2",
     "immer": "^9.0.6"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2705,7 +2705,7 @@ __metadata:
   dependencies:
     "@metamask/auto-changelog": "npm:^3.4.4"
     "@metamask/json-rpc-engine": "npm:^10.0.3"
-    "@metamask/messenger": "npm:^0.0.0"
+    "@metamask/messenger": "npm:^0.1.0"
     "@metamask/utils": "npm:^11.4.2"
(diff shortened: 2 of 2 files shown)


packages/messenger/CHANGELOG.md
```diff
@@ -32,8 +34,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Existing `RestrictedMessenger` instances should be replaced with a `Messenger` with the `parent` constructor parameter set to the global messenger. We can now use the same class everywhere, passing capabilities using `delegate`.
   - See this ADR for details: https://github.com/MetaMask/decisions/blob/main/decisions/core/0012-messenger-delegation.md
 
-### Fixed
-
-- Update `unsubscribe` type signature to support selector event handlers ([#6262](https://github.com/MetaMask/core/pull/6262))
```
Gudahtt, line 37: I removed this because it was already published under `base-controller`, so wasn't really "new" anymore, considering the migration from base-controller as the starting point.

mathieuartu (approved):
LGTM!

cryptodev-2s:
> Adding `no-changelog` because the changelog check job is complaining two unreleased packages not referencing the version bump in their changelogs. But they're not released yet, so there is no need to list any "changes" from the previous non-existent release.
> 
> Also it's complaining about the `base-controller` changelog not including the messenger version bump from 0.0.0 to 0.1.0 for some reason. But this is not really a version bump since it was not added until this release (and the addition of the dependency _is_ documented), so the job is wrong here as well I think.

On the unreleased packages:
Our recent practice is to include an “Initial release” section in CHANGELOG.md even before the first publish, and keep appending PRs (including dependency bumps) under that section. That lets us track all changes from the initial commit up to the first release.

On the base-controller check:
Agreed—the job only detects “dependency updated” and doesn’t distinguish “newly added.” While 0.0.0 → 0.1.0 isn’t really a bump, we can either:
- add a brief line under Added noting the new dependency at 0.1.0 (to satisfy the check), or
- skip it and follow up by improving the check’s logic to recognize first-time additions. (I can create an issue for this)

PR #6358, opened 2025-08-21
fix: token selector issues

bergarces:
## Explanation

The format for EVM addresses for AccountTrackerController in mobile differs from extension and from the other controller states. This was causing it to miss native account balances.

It also uses the correct internal account type in the assets, so that BTC account types can be identified.

Finally, it fixes an issue with the `mergeAssets` function mutating internal parts of cached objects.

## Checklist

### DIFF
--- a/packages/assets-controllers/src/selectors/token-selectors.ts
+++ b/packages/assets-controllers/src/selectors/token-selectors.ts
@@ -299,8 +299,6 @@ const selectAllEvmAssets = createAssetListSelector(
       }
     }
 
-    console.log('EVMMM', JSON.stringify(groupAssets, null, 2));
-
     return groupAssets;
   },
 );
--- a/packages/assets-controllers/src/selectors/token-selectors.ts
+++ b/packages/assets-controllers/src/selectors/token-selectors.ts
@@ -100,7 +100,7 @@ const selectAccountsToGroupIdMap = createAssetListSelector(
           accountsMap[
             // TODO: We would not need internalAccounts if evmTokens state had the accountId
             internalAccount.type.startsWith('eip155')
-              ? internalAccount.address
+              ? internalAccount.address.toLowerCase()
               : accountId
           ] = accountGroupId;
         }
@@ -137,7 +137,7 @@ const selectAllEvmAccountNativeBalances = createAssetListSelector(
       for (const [accountAddress, accountBalance] of Object.entries(
         chainAccounts,
       )) {
-        const accountGroupId = accountsMap[accountAddress];
+        const accountGroupId = accountsMap[accountAddress.toLowerCase()];
         groupAssets[accountGroupId] ??= {};
         groupAssets[accountGroupId][chainId] ??= [];
         const groupChainAssets = groupAssets[accountGroupId][chainId];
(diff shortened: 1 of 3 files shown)


packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -420,7 +445,10 @@ function mergeAssets(
     const existingAccountGroupAssets = existingAssets[accountGroupId];
 
     if (!existingAccountGroupAssets) {
-      existingAssets[accountGroupId] = accountAssets;
+      existingAssets[accountGroupId] = {};
+      for (const [network, chainAssets] of Object.entries(accountAssets)) {
+        existingAssets[accountGroupId][network] = [...chainAssets];
+      }
```
bergarces, line 457: This was tricky to spot. By reusing the array from the memoized object and later pushing to it, the new test was failing unless it was checked on isolation. This could have caused problems in the clients as well.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -137,7 +147,13 @@ const selectAllEvmAccountNativeBalances = createAssetListSelector(
       for (const [accountAddress, accountBalance] of Object.entries(
         chainAccounts,
       )) {
-        const accountGroupId = accountsMap[accountAddress];
+        const account = accountsMap[accountAddress.toLowerCase()];
+        if (!account) {
```
bergarces, line 156: This should not be possible, but better not to take any chances that can crash the app.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -224,7 +240,12 @@ const selectAllEvmAssets = createAssetListSelector(
       ) as [Hex, Token[]][]) {
         for (const token of addressTokens) {
           const tokenAddress = token.address as Hex;
-          const accountGroupId = accountsMap[accountAddress];
+          const account = accountsMap[accountAddress];
```
bergarces, line 243: This should not be possible, but better not to take any chances that can crash the app.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -224,7 +240,12 @@ const selectAllEvmAssets = createAssetListSelector(
...
+          if (!account) {
```
bergarces, line 250: This should not be possible, but better not to take any chances that can crash the app.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -30,15 +31,21 @@ const MULTICHAIN_NATIVE_ASSET_IDS = [
   `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501`,
 ];
 
+type EvmAccountType = Extract<InternalAccount['type'], `eip155:${string}`>;
+type MultichainAccountType = Exclude<
+  InternalAccount['type'],
+  `eip155:${string}`
+>;
```
bergarces, line 38: Going to use this type from the account instead of the generic `evm` and `multichain`. This will allow us to determine the type of bitcoin account the asset belongs to.

This future issue was raised recently on slack.

bergarces:
@metamaskbot publish-preview

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -137,7 +152,13 @@ const selectAllEvmAccountNativeBalances = createAssetListSelector(
       for (const [accountAddress, accountBalance] of Object.entries(
         chainAccounts,
       )) {
-        const accountGroupId = accountsMap[accountAddress];
+        const account = accountsMap[accountAddress.toLowerCase()];
```
bergarces, line 155: `toLowerCase()` here because `accountsByChainId` in mobile uses checksumed addresses, unlike every other controller state.

This is only temporary though, because as soon as native addresses are returned in token balances, all this code can go.

PR #6363, opened 2025-08-22
feat: Add metrics for gasless 7702, refactoring

dan437:
## Explanation
This PR adds metrics for gasless 7702 and includes refactoring (renaming, updated gasIncluded7702 checks). It also makes the `getBridgeHistoryItemByTxMetaId` function available via messaging system.

## Checklist


packages/bridge-status-controller/src/types.ts
```diff
@@ -215,6 +217,10 @@ export type SourceChainTxMetaId = string;
 
 export type BridgeStatusControllerState = {
   txHistory: Record<SourceChainTxMetaId, BridgeHistoryItem>;
+  currentSubmissionRequest?: {
+    quoteResponse: QuoteResponse<string | TxData> & QuoteMetadata;
+    isBridgeTx: boolean;
```
infiniteflower: Can we not just reconstruct if it's a bridge tx from the quote response itself? I.e. check the source and dest chain ids? A single source of truth would be better, instead of extra fields to keep track of.

packages/bridge-status-controller/src/types.ts
```diff
@@ -215,6 +217,10 @@ export type SourceChainTxMetaId = string;
 
 export type BridgeStatusControllerState = {
   txHistory: Record<SourceChainTxMetaId, BridgeHistoryItem>;
+  currentSubmissionRequest?: {
+    quoteResponse: QuoteResponse<string | TxData> & QuoteMetadata;
```
infiniteflower: Just wondering if it's possible to just reference a `txHistory` item (by `txMetaId`) instead of dumping the entire `quoteResponse` into there for your purposes in the client app.

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -82,6 +82,10 @@ const metadata: StateMetadata<BridgeStatusControllerState> = {
     persist: true,
     anonymous: false,
   },
+  currentSubmissionRequest: {
```
infiniteflower: Would this get wiped when the user restarts their app and cause bugs? Like let's say the app closes for whatever reason while the tx is in submission.

packages/bridge-status-controller/src/types.ts
rarquevaux: I can try to do that. Just wondering what you mean by "reconstruct if it's a bridge"? Likely a dummy question but is it possible that we handle also regular swaps here?

packages/bridge-status-controller/src/types.ts
micaelae: +1

dan437: I tried that at first, but a `txHistory` item was not present when we needed it, only later, which was too late. That's why this new state param was added.

packages/bridge-status-controller/src/utils/transaction.ts
```diff
@@ -347,6 +347,7 @@ export const getAddTransactionBatchParams = async ({
     TransactionController['addTransactionBatch']
   >[0] = {
     disable7702,
+    isGasFeeIncluded: Boolean(gasIncluded7702),
```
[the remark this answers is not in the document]
dan437: This is intentional.

packages/bridge-controller/src/types.ts
```diff
@@ -222,7 +222,7 @@ export type QuoteRequest<
   /**
    * Whether to request quotes that use EIP-7702 delegated gasless execution
    */
-  gasless7702: boolean;
+  gasIncluded7702: boolean;
```
infiniteflower, line 225: Isn't this a breaking change? If so can we properly indicate it in the CHANGELOG?

pedronfigueiredo (approved):

micaelae (approved):

infiniteflower (approved):

PR #6366, opened 2025-08-22
Release/510.0.0

bergarces:
## Explanation

- Bumps `@metamask/assets-controllers` to 74.1.0.
- Bumps devDependency to the same version for bridge-controller and updates changelog.

## Checklist

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [74.1.0]
+
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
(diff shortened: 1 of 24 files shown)


Prithpal-Sooriya (dismissed):

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [74.1.0]
```
mcmire, line 10: Why is this a minor release? Is there any new functionality being added?

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Bump `@metamask/assets-controllers` from `^74.0.0` to `^74.1.0` ([#6366](https://github.com/MetaMask/core/pull/6366))
+
```
mcmire: This seems to be a dev dependency. It doesn't need to be noted in the changelog.

``` suggestion
```

packages/assets-controllers/CHANGELOG.md
mcmire, line 1: "Uses the correct internal account type for the asset" — What does this mean? Are there any more details that could be provided here? Specifically, what part of the API is being changed and how?

bergarces: Each asset `type` will now return the values used by `InternalAccount` `type` field.

Which one could argue is a breaking change, but we are seeing it as a bug fix, given that the selectors and the type were added in 74.0.0 and we have realised we need the internal account types for BTC different account types to be correctly identified.

packages/assets-controllers/CHANGELOG.md
bergarces: A new field is being added on top of the bug fixes.

bergarces:
I don't want to block any releases and I won't be back until Tuesday, so I'm going to close this and delete the branch until I can redo the release with all the comments addressed.

packages/assets-controllers/CHANGELOG.md
mcmire: Okay. Should this be in "Fixed" instead?

packages/assets-controllers/CHANGELOG.md
mcmire: I didn't see mention of a new field in the changelog, was that added in #6358?

mcmire:
@bergarces Sounds good. Ping me on Tuesday when you make the new release and I'll make sure to hop on it.

PR #6367, opened 2025-08-22
feat: implement enableAllPopularNetworks function with comprehensive …

salimtb:
…tests

## Explanation

This PR implements 

- the `enableAllPopularNetworks()` to enable all popular networks
- modifies the `enableNetwork()` function to use exclusive behavior, ensuring only one network is enabled per namespace at a time.

## Checklist

### DIFF
--- a/packages/network-enablement-controller/src/NetworkEnablementController.ts
+++ b/packages/network-enablement-controller/src/NetworkEnablementController.ts
@@ -231,23 +231,43 @@ export class NetworkEnablementController extends BaseController<
    * and Solana mainnet. Unlike the enableNetwork method which has exclusive behavior,
    * this method enables multiple networks across namespaces simultaneously.
    *
-   * Popular networks that don't exist in the current state will be skipped silently.
+   * Popular networks that don't exist in NetworkController or MultichainNetworkController configurations will be skipped silently.
    */
   enableAllPopularNetworks(): void {
     this.update((s) => {
-      // Enable all popular EVM networks
+      // Get current network configurations to check if networks exist
+      const networkControllerState = this.messagingSystem.call(
+        'NetworkController:getState',
+      );
+      const multichainState = this.messagingSystem.call(
+        'MultichainNetworkController:getState',
+      );
+
+      // Enable all popular EVM networks that exist in NetworkController configurations
       POPULAR_NETWORKS.forEach((chainId) => {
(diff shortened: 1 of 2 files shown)


packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -215,25 +214,45 @@ export class NetworkEnablementController extends BaseController<
...
+   * This method enables all networks defined in POPULAR_NETWORKS (EVM networks)
+   * and Solana mainnet. Unlike the enableNetwork method which has exclusive behavior,
+   * this method enables multiple networks across namespaces simultaneously.
+   *
+   * Popular networks that don't exist in the current state will be skipped silently.
+   */
+  enableAllPopularNetworks(): void {
+    this.update((s) => {
+      // Enable all popular EVM networks
+      POPULAR_NETWORKS.forEach((chainId) => {
```
Prithpal-Sooriya, line 316: Careful here, on mobile the popular networks are all added, but if we use this on extension, we do not have all popular networks added. (Likewise users could have deleted/removed a popular network on mobile?)

For safety, we could use "All current popular networks" by filtering against the NetworkController/MultichainNetworkController.

salimtb: correct yes , good catch i'll fix this now

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -215,25 +214,45 @@ export class NetworkEnablementController extends BaseController<
...
-        Object.keys(s.enabledNetworkMap[namespace]).forEach((key) => {
-          const { reference: keyReference } = deriveKeys(key as CaipChainId);
-          if (!isPopularNetwork(keyReference)) {
-            s.enabledNetworkMap[namespace][key as CaipChainId | Hex] = false;
-          }
-        });
-      }
+      // disable all networks in the same namespace
+      Object.keys(s.enabledNetworkMap[namespace]).forEach((key) => {
+        s.enabledNetworkMap[namespace][key as CaipChainId | Hex] = false;
```
Prithpal-Sooriya, line 219: Ooooh interesting, so the core/mobile controller actually uses true/false.
Lol the extension version just deletes the entries 😆

PR #6369, opened 2025-08-25
feat: add account api support

salimtb:
## Explanation

This PR implements functionality for AccountTrackerController to fetch native balances using the AccountsAPI when external services are enabled, with comprehensive test coverage to ensure proper behavior across different configurations.

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

PR #6372, opened 2025-08-25
Release/511.0.0

bergarces:
## Explanation

New minor release for `@metamask/assets-controllers`.

## Checklist

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [74.1.0]
+
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
@@ -1903,7 +1905,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use Ethers for AssetsContractController ([#845](https://github.com/MetaMask/core/pull/845))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@74.0.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@74.1.0...HEAD
+[74.1.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@74.0.0...@metamask/assets-controllers@74.1.0
 [74.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.3.0...@metamask/assets-controllers@74.0.0
 [73.3.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.2.0...@metamask/assets-controllers@73.3.0
 [73.2.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@73.1.0...@metamask/assets-controllers@73.2.0
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -12,16 +12,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
-- Uses the correct internal account type for the asset ([#6358](https://github.com/MetaMask/core/pull/6358)).
+
 - Enable `AccountTrackerController` to fetch native balances using AccountsAPI when `allowExternalServices` is enabled ([#6369](https://github.com/MetaMask/core/pull/6369))
+
   - Implement native balance fetching via AccountsAPI when `useAccountsAPI` and `allowExternalServices` are both true
   - Add fallback to RPC balance fetching when external services are disabled
   - Add comprehensive test coverage for both AccountsAPI and RPC balance fetching scenarios
 
+- Add new `accountId` field to the `Asset` type ([#6358](https://github.com/MetaMask/core/pull/6358))
+
 ### Fixed
 
-- Ensure that the evm addresses used for an internal mapping are always lowercase to avoid mismatches with client format ([#6358](https://github.com/MetaMask/core/pull/6358)).
-- Prevents mutation of memoized fields, which was causing issues ([#6358](https://github.com/MetaMask/core/pull/6358)).
+- Uses `InternalAccount['type']` for the `Asset['type']` property ([#6358](https://github.com/MetaMask/core/pull/6358))
+
+- Ensure that the evm addresses used to fetch balances from AccountTrackerController state is lowercase, in order to account for discrepancies between clients ([#6358](https://github.com/MetaMask/core/pull/6358))
+
+- Prevents mutation of memoized fields used inside selectors ([#6358](https://github.com/MetaMask/core/pull/6358))
 
 - Fix duplicate token balance entries caused by case-sensitive address comparison in `TokenBalancesController.updateBalances` ([#6354](https://github.com/MetaMask/core/pull/6354))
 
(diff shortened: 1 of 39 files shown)


packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,26 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   - Add fallback to RPC balance fetching when external services are disabled
   - Add comprehensive test coverage for both AccountsAPI and RPC balance fetching scenarios
 
+- Add new `accountId` field to the `Asset` type ([#6358](https://github.com/MetaMask/core/pull/6358))
+
 ### Fixed
 
-- Ensure that the evm addresses used for an internal mapping are always lowercase to avoid mismatches with client format ([#6358](https://github.com/MetaMask/core/pull/6358)).
-- Prevents mutation of memoized fields, which was causing issues ([#6358](https://github.com/MetaMask/core/pull/6358)).
+- Uses `InternalAccount['type']` for the `Asset['type']` property ([#6358](https://github.com/MetaMask/core/pull/6358))
```
mcmire, line 28: If either the extension or mobile app is already using this code are they able to upgrae to this change without needing any additional changes? This seems like a backward-incompatible change to me since we are changing the type of a property in `Asset` and we are changing what `selectAccountsToGroupIdMap` returns. Thoughts...?

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,26 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- Ensure that the evm addresses used to fetch balances from AccountTrackerController state is lowercase, in order to account for discrepancies between clients ([#6358](https://github.com/MetaMask/core/pull/6358))
```
mcmire, line 30: Similar, it also seems like you'd need to change either another controller or the client to adapt to this change, is this true or is this harmless?

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,26 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
-- Uses the correct internal account type for the asset ([#6358](https://github.com/MetaMask/core/pull/6358)).
+
 - Enable `AccountTrackerController` to fetch native balances using AccountsAPI when `allowExternalServices` is enabled ([#6369](https://github.com/MetaMask/core/pull/6369))
+
   - Implement native balance fetching via AccountsAPI when `useAccountsAPI` and `allowExternalServices` are both true
   - Add fallback to RPC balance fetching when external services are disabled
   - Add comprehensive test coverage for both AccountsAPI and RPC balance fetching scenarios
 
+- Add new `accountId` field to the `Asset` type ([#6358](https://github.com/MetaMask/core/pull/6358))
```
mcmire, line 24: It seems like we might be missing a note for `selectAccountsToGroupIdMap` here?

``` suggestion
- Add new `accountId` field to the `Asset` type ([#6358](https://github.com/MetaMask/core/pull/6358))

- Update `selectAccountsToGroupIdMap` so the `type` of each returned asset is `InternalAccount['type']` rather than a string ([#6358](https://github.com/MetaMask/core/pull/6358))
```

(which, as noted below, is possibly a breaking change?)

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,26 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+## [74.1.0]
+
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.1.0` to `^8.2.0` ([#6355](https://github.com/MetaMask/core/pull/6355))
-- Uses the correct internal account type for the asset ([#6358](https://github.com/MetaMask/core/pull/6358)).
+
 - Enable `AccountTrackerController` to fetch native balances using AccountsAPI when `allowExternalServices` is enabled ([#6369](https://github.com/MetaMask/core/pull/6369))
```
mcmire, line 14: Should this go in "Added"? This seems like a new feature.

mcmire: `allowExternalServices` is also an option that wasn't there before, and that isn't clear from this entry.

packages/assets-controllers/CHANGELOG.md
mcmire, line 1: Does [this change](https://github.com/MetaMask/core/pull/6369/files#diff-09448a98d6ea70ebb5f28c2d45687d07bfb5beef1db71d326bf151e6246e410dR368) need to be noted somewhere? What is the purpose of this?

packages/assets-controllers/CHANGELOG.md
```diff
@@ -1903,7 +1912,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use Ethers for AssetsContractController ([#845](https://github.com/MetaMask/core/pull/845))
```
mcmire, line 1915: What is the effect of this change? Does this change the implementation or do we generally have the same behavior? Curious if we should mention this at all or if it's a developer-only thing.

``` suggestion
```

mcmire (commented):
Had some comments. Mainly curious about possible breaking changes.

packages/assets-controllers/CHANGELOG.md
bergarces: Neither of them are using the code yet. I was in the process to implement it and a new discussion about how we need to differentiate between different BTC accounts raised, prompting me to do this change along with other needed ones (such as adding accountId).

packages/assets-controllers/CHANGELOG.md
bergarces: It's harmless.

The issue here is that the accounts are checksumed in the mobile AccountTrackerController state only, which meant those balances were not being fetched in mobile. This alllows me to fetch them.

packages/assets-controllers/CHANGELOG.md
bergarces: This is not a breaking change, it is a new field that did not exist before.

That selector is not being exported.

packages/assets-controllers/CHANGELOG.md
bergarces: Not my code, maybe @salimtb can comment.

packages/assets-controllers/CHANGELOG.md
bergarces: @salimtb Can you clarify here?

packages/assets-controllers/CHANGELOG.md
bergarces: This is a very old entry. Not relevant to this release.

packages/assets-controllers/CHANGELOG.md
darkwing: Echo's Bernardo's comment; this is not used yet.

packages/assets-controllers/CHANGELOG.md
mcmire: Ah okay, gotcha, thanks for confirming.

packages/assets-controllers/CHANGELOG.md
mcmire: Ah you are right. Sorry 😬

mcmire (dismissed):
One comment left, but it's minor and non-blocking.

packages/assets-controllers/CHANGELOG.md
salimtb: regarding `allowExternalServices` , it's linked to the basic functionality , if it's off we should not use accountAPI to fetch balances.
hmm new feature but still live inside the same function ( refersh )  and the same controller ( accountTracker ) , we're just adding a new fetcher which is accountAPI in addition of rpc.
i'll move it to added , maybe it's more accurate

packages/assets-controllers/CHANGELOG.md
salimtb: The goal is to normalize the staked balance format to a hex string. Since getShares returns a decimal value, we’ll convert it to hex. I’ll add a changelog entry for this.

salimtb: fixed [3a459f6](https://github.com/MetaMask/core/pull/6372/commits/3a459f64e94e04fb93e03ffddc59c1162cdcf2d9)

packages/assets-controllers/CHANGELOG.md
salimtb: fixed [3a459f6](https://github.com/MetaMask/core/pull/6372/commits/3a459f64e94e04fb93e03ffddc59c1162cdcf2d9)

amitabh94 (approved):

PR #6033, opened 2025-06-25
feat: add gator permissions controller

V00D00-child:
#### Current State and Why Change is Needed

MetaMask clients currently lack a dedicated system for managing gator permissions that are stored to profile sync via the `@metamask/gator-permissions-snap`.

#### Solution and How It Works

This change introduces a new `@metamask/gator-permissions-controller` package that provides a comprehensive solution for managing gator permissions in MetaMask clients with gator-snap integration.

#### Changes That Might Not Be Obvious

- Serialization Strategy: The controller `state` uses JSON serialization for storing permission data fetched from `@metamask/gator-permissions-snap`, which allows for efficient storage and retrieval while maintaining data integrity. The deserialize permission data is represented as a list of gator permissions filtered by permission type and chainId.
- Default Permission Structure: The controller initializes with an empty structure for all three permission types, ensuring consistent state even when no permissions are configured
 
####  Package Dependencies and Integration

The new package depends on `@metamask/snaps-controllers` as a peer dependency, ensuring it can leverage sending RPC requests to an installed Metamask Snap. This integration allows the `GatorPermissionsController` to forward requests to `@metamask/gator-permissions-snap` to fetch users' Gator permissions that have been stored in the MetaMask Profile Sync service. 

The `@metamask/gator-permissions-snap` will take on the responsibility of authenticating with MetaMask Profile Sync service using an`SRP` identifier via integration with `@metamask/message-signing-snap`.

#### No Dependency Upgrades Required

This is a new package that introduces new functionality without requiring changes to existing dependencies. The package uses the current stable versions of `@metamask/base-controller`, `@metamask/utils `,`@metamask/snaps-sdk`, and `@metamask/snaps-utils` following the established patterns in the MetaMask codebase.

## References

Related to(MM snap-7715-permissions): [Persisting Granted Permissions with MM Profile Sync](https://github.com/MetaMask/snap-7715-permissions/pull/84)
Requires(MM snap-7715-permissions):[Add new permissionsProvider_getGrantedPermissions RPC](https://github.com/MetaMask/snap-7715-permissions/pull/108)
Required by(MM Extension): https://github.com/MetaMask/metamask-extension/pull/33996

### Gator Permissions Data Flow
```mermaid
graph TD
    %% dApp flow for storing permissions
    A[dApp<br/>client side RPC] -->|RPC| GPS[gator-permissions-snap]
    C -->|WRITE| D[(permissions stored<br/>across all sites)]
    
    %% User flow for reading permissions
[...]

PR #6379, opened 2025-08-25
Integrate erc7715 types from @metamask/7715-permission-types package

V00D00-child:
## Explanation

Integrate [ERC-7715](https://eip.tools/eip/7715) types from [@metamask/7715-permission-types](https://github.com/MetaMask/delegation-toolkit/tree/main/packages/7715-permission-types) package to avoid type duplication across multiple repos. MM clients will use the  ERC-7715 types exposed via the `@metamask/gator-permissions-controller` to manage gator permissions.

## References

- Related to https://github.com/MetaMask/metamask-extension/pull/35219

## Checklist

### DIFF
--- a/packages/gator-permissions-controller/src/types.ts
+++ b/packages/gator-permissions-controller/src/types.ts
@@ -1,3 +1,13 @@
+import type {
+  PermissionTypes,
+  Signer,
+  BasePermission,
+  NativeTokenStreamPermission,
+  NativeTokenPeriodicPermission,
+  Erc20TokenStreamPermission,
+  Erc20TokenPeriodicPermission,
+  Rule,
+} from '@metamask/7715-permission-types';
 import type { Hex } from '@metamask/utils';
 
 /**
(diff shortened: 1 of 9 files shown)


packages/gator-permissions-controller/src/types.ts
```diff
@@ -20,102 +30,25 @@ export enum GatorPermissionsSnapRpcMethod {
...
-   */
-  data: Record<string, unknown>;
-
-  /**
-   * set of restrictions or conditions that a signer must abide by when redeeming a Permission.
-   */
-  rules?: Record<string, unknown>;
-};
-
 export type MetaMaskBasePermissionData = {
```
MoMannn: We already have this defined in the params: https://github.com/MetaMask/delegation-toolkit/blob/a22ecb557e3173fb38bed3f3d1344b57539febaf/packages/7715-permission-types/src/types.ts#L104 

But I see its not exported. Might be better to first update the types to export this as well so not have duplication.

packages/gator-permissions-controller/src/types.ts
```diff
@@ -20,102 +30,25 @@ export enum GatorPermissionsSnapRpcMethod {
...
- * Represents an ERC-7715 wallet signer type.
- *
- */
-export type WalletSigner = {
-  type: 'wallet';
-  data: Record<string, unknown>;
-};
-
-export type SignerParam = AccountSigner | WalletSigner;
+export type PermissionTypesWithCustom = PermissionTypes | CustomPermission;
```
MoMannn, line 45: I think the idea is to not allow CustomPermission for the first version and to limit to the permissions that we know how to handle in the snap. That is why we didn't define it in the types for now.

MoMannn (changes requested):

packages/gator-permissions-controller/src/types.ts
V00D00-child: There is already logic in place in the `GatorPermissionContoller` to handle custom permission type: https://github.com/MetaMask/core/blob/main/packages/gator-permissions-controller/src/GatorPermissionsController.ts#L377.

I intentionally left out support for custom permission type, but there was push back on that idea in a previous PR, so I added it:(see link to comment) https://github.com/MetaMask/core/pull/6033#discussion_r2215333192

MoMannn: Then it might be a good to add the custom on to the permission types so that we are consistent? @jeffsmale90 what do you think?

packages/gator-permissions-controller/src/types.ts
V00D00-child: Opened a PR to expose that type. Once that is merged and released, I can update the package version to reduce the duplication: https://github.com/MetaMask/delegation-toolkit/pull/65

V00D00-child: Updated: we are now using `@metamask/7715-permission-types v0.3.0`

MoMannn (approved):

PR #6394, opened 2025-08-27
fix(multichain-account-service): fix account event subscription + clear state before re-init service

ccharly:
## Explanation

The event subscription has to be moved in the constructor otherwise if we call `init` multiple times, we will subscribe to those same events multiple times.

Also, moving those subscriptions on the constructor ensure that the service will pick up newly added accounts after the service is being created even if `init` has not been called yet, which seems to be another bug on the extension given that the service is not initialized properly after the first on-boarding...

## References

N/A

## Checklist

### DIFF
--- a/packages/multichain-account-service/src/MultichainAccountService.ts
+++ b/packages/multichain-account-service/src/MultichainAccountService.ts
@@ -113,6 +113,13 @@ export class MultichainAccountService {
       'MultichainAccountService:alignWallet',
       (...args) => this.alignWallet(...args),
     );
+
+    this.#messenger.subscribe('AccountsController:accountAdded', (account) =>
+      this.#handleOnAccountAdded(account),
+    );
+    this.#messenger.subscribe('AccountsController:accountRemoved', (id) =>
+      this.#handleOnAccountRemoved(id),
+    );
   }
 
   /**
@@ -120,6 +127,9 @@ export class MultichainAccountService {
    * multichain accounts and wallets.
    */
   init(): void {
+    this.#wallets.clear();
+    this.#accountIdToContext.clear();
+
     // Create initial wallets.
     const { keyrings } = this.#messenger.call('KeyringController:getState');
     for (const keyring of keyrings) {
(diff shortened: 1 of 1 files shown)


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -113,13 +113,23 @@ export class MultichainAccountService {
...
+      this.#handleOnAccountRemoved(id),
+    );
   }
 
   /**
    * Initialize the service and constructs the internal reprensentation of
    * multichain accounts and wallets.
    */
   init(): void {
+    this.#wallets.clear();
```
fabiobozzo, line 141: Just for me, correct me if I'm wrong please: MultichainAccountService transient state (wallets, convenience maps etc.) are built from keyrings, inside `init`. So if an `handleOnAccountAdded` runs first, then you start `init` with some wallets already. 

Then we clear `wallets` and.... rebuild it from scratch, just in case. Correct?

ccharly: Yes, that's correct.

I was wondering if `init` should actually become `reset` (which IMO emphasis a bit more what's going on, we're going to reset the state entirely, so we clear everything and re-build from scratch).

And `init` would just become the default initialization of the controller (assuming everything is "blank" already).

Given the current logic on the clients, I honestly wonder if `init` should actually be called inside the constructor of the service. And we would call the new `reset` only when we know the other state dependencies have been changed/reset.

Any opinion?

Basically code-wise, that'd be something something like:
```ts
constructor() {
  ...
  
  this.#init();
}

reset() {
  this.#wallets.clear();
  ...
    
  this.#init();
}
```

fabiobozzo: Ok so: 

constructor → "Set up service and make it ready"
reset() → "need to rebuild everything"
#init() → "build state from current keyring accounts"

why doesn't constructor just calls `reset` then? 

In any case, I like the current approach as it is: is there another place where we need to "reset" other than MultichainAccountService is spawned up?

ccharly: We do need to reset the state of the account-tree and service during the "forget password" flow.

But honestly, we should not really need this anyway since we're reacting to events, that should be enough, or at least, that's what we should aim for!

I'll keep things simple and keep the `init` as it is for now. Thanks 😃

fabiobozzo (approved):

PR #6395, opened 2025-08-27
fix: Keep delegated handlers when unregistering actions

FrederikBolding:
## Explanation

Keep delegated handlers around even when unregistering actions. This helps with the resiliency of the messenger system in case actions are unregistered and registered elsewhere in the chain of delegations. This is also very useful for testing.

Additionally adjusts the error message thrown when a delegated action handler cannot be found to match an undelegated one.

### DIFF
--- a/packages/messenger/CHANGELOG.md
+++ b/packages/messenger/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
+### Changed
 
 - Keep delegated handlers when unregistering actions ([#6395](https://github.com/MetaMask/core/pull/6395))
 
--- a/packages/messenger/CHANGELOG.md
+++ b/packages/messenger/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Keep delegated handlers when unregistering actions ([#6395](https://github.com/MetaMask/core/pull/6395))
+
 ## [0.1.0]
 
 ### Added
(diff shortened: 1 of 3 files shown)


packages/messenger/src/Messenger.test.ts
```diff
@@ -1264,31 +1264,6 @@ describe('Messenger', () => {
         actions: ['Source:getLength'],
       });
 
-      expect(() => delegatedMessenger.call('Source:getLength', 'test')).toThrow(
-        `Cannot call 'Source:getLength', action not registered.`,
-      );
-    });
-
-    it('unregisters delegated action handlers when action is unregistered', () => {
```
Gudahtt, line 1272: Nit: It would be nice to test these cases:

* `throws an error when delegated action is called after handler is unregistered`
* `allows calling delegated action that is unregistered then registered again`

packages/messenger/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Keep delegated handlers when unregistering actions ([#6395](https://github.com/MetaMask/core/pull/6395))
```
Gudahtt, line 12: Nit: Should this be changed instead? It was not a bug, it was done intentionally.

Gudahtt (dismissed):
Looks great! Two minor suggestions

packages/messenger/CHANGELOG.md
FrederikBolding: Good point 👍

packages/messenger/src/Messenger.test.ts
Gudahtt: I'll add these in a follow-up PR

Gudahtt: Done here: https://github.com/MetaMask/core/pull/6419

PR #6397, opened 2025-08-27
feat: add account discovery to `MultichainAccountWallet` and providers

hmalik88:
### EvmAccountProvider
1. `discoverAndCreateAccounts` method added to perform discovery and return the created account. The method will add an account to the keyring and remove it (except for `groupIndex` 0) if it is found there is no transaction history.
2. `getEvmProvider` method was added to grab the provider from the `NetworkController`.

### SolAccountProvider
1. `discoverAndCreateAccounts` method added to perform discovery and return the created account. This method does not create an account if discovery yields no accounts.
2. Added a `#getKeyringClientFromSnapId` method that initializes the keyring client into a newly added `#client` property.

### MultichainAccountWallet
1. `discoverAndCreateAccounts` method added to act as an orchestrator for provider discovery. This method "fast-forwards" lagging providers to the highest `groupIndex` being processed by a provider. At the end of discovery, it will sync the wallet and then align all created groups. The method returns only the discovered accounts (for consumption of the clients' metrics).

## References

N/A

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "534.0.0",
+  "version": "535.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/bridge-controller/CHANGELOG.md
+++ b/packages/bridge-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [42.0.0]
+
 ### Added
 
 - Add `gas_included_7702` field to metrics tracking for EIP-7702 gasless transactions ([#6363](https://github.com/MetaMask/core/pull/6363))
@@ -554,7 +556,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Initial release ([#5317](https://github.com/MetaMask/core/pull/5317))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@41.4.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@42.0.0...HEAD
+[42.0.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@41.4.0...@metamask/bridge-controller@42.0.0
 [41.4.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@41.3.0...@metamask/bridge-controller@41.4.0
 [41.3.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@41.2.0...@metamask/bridge-controller@41.3.0
 [41.2.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-controller@41.1.0...@metamask/bridge-controller@41.2.0
--- a/packages/bridge-controller/package.json
+++ b/packages/bridge-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/bridge-controller",
-  "version": "41.4.0",
+  "version": "42.0.0",
   "description": "Manages bridge-related quote fetching functionality for MetaMask",
   "keywords": [
     "MetaMask",
--- a/packages/bridge-status-controller/CHANGELOG.md
+++ b/packages/bridge-status-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [42.0.0]
+
 ### Added
 
 - Add `getBridgeHistoryItemByTxMetaId` method available via messaging system for external access to bridge history items ([#6363](https://github.com/MetaMask/core/pull/6363))
@@ -14,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Bump `@metamask/bridge-controller` peer dependency from `^41.0.0` to `^42.0.0` ([#6476](https://github.com/MetaMask/core/pull/6476))
 - Bump `@metamask/base-controller` from `^8.2.0` to `^8.3.0` ([#6465](https://github.com/MetaMask/core/pull/6465))
 - Pass the `isGasFeeIncluded` parameter through transaction utilities ([#6363](https://github.com/MetaMask/core/pull/6363))
 
@@ -530,7 +533,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Initial release ([#5317](https://github.com/MetaMask/core/pull/5317))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@41.0.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@42.0.0...HEAD
+[42.0.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@41.0.0...@metamask/bridge-status-controller@42.0.0
 [41.0.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@40.2.0...@metamask/bridge-status-controller@41.0.0
 [40.2.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@40.1.0...@metamask/bridge-status-controller@40.2.0
 [40.1.0]: https://github.com/MetaMask/core/compare/@metamask/bridge-status-controller@40.0.0...@metamask/bridge-status-controller@40.1.0
--- a/packages/bridge-status-controller/package.json
+++ b/packages/bridge-status-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/bridge-status-controller",
-  "version": "41.0.0",
+  "version": "42.0.0",
   "description": "Manages bridge-related status fetching functionality for MetaMask",
   "keywords": [
     "MetaMask",
@@ -59,7 +59,7 @@
   "devDependencies": {
     "@metamask/accounts-controller": "^33.0.0",
     "@metamask/auto-changelog": "^3.4.4",
-    "@metamask/bridge-controller": "^41.4.0",
+    "@metamask/bridge-controller": "^42.0.0",
     "@metamask/gas-fee-controller": "^24.0.0",
     "@metamask/network-controller": "^24.1.0",
     "@metamask/snaps-controllers": "^14.0.1",
@@ -77,7 +77,7 @@
   },
   "peerDependencies": {
     "@metamask/accounts-controller": "^33.0.0",
-    "@metamask/bridge-controller": "^41.0.0",
+    "@metamask/bridge-controller": "^42.0.0",
     "@metamask/gas-fee-controller": "^24.0.0",
     "@metamask/network-controller": "^24.0.0",
     "@metamask/snaps-controllers": "^14.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2727,7 +2727,7 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@metamask/bridge-controller@npm:^41.4.0, @metamask/bridge-controller@workspace:packages/bridge-controller":
+"@metamask/bridge-controller@npm:^42.0.0, @metamask/bridge-controller@workspace:packages/bridge-controller":
   version: 0.0.0-use.local
   resolution: "@metamask/bridge-controller@workspace:packages/bridge-controller"
   dependencies:
@@ -2783,7 +2783,7 @@ __metadata:
     "@metamask/accounts-controller": "npm:^33.0.0"
     "@metamask/auto-changelog": "npm:^3.4.4"
     "@metamask/base-controller": "npm:^8.3.0"
-    "@metamask/bridge-controller": "npm:^41.4.0"
+    "@metamask/bridge-controller": "npm:^42.0.0"
     "@metamask/controller-utils": "npm:^11.12.0"
     "@metamask/gas-fee-controller": "npm:^24.0.0"
     "@metamask/keyring-api": "npm:^20.1.0"
@@ -2807,7 +2807,7 @@ __metadata:
     uuid: "npm:^8.3.2"
   peerDependencies:
     "@metamask/accounts-controller": ^33.0.0
-    "@metamask/bridge-controller": ^41.0.0
+    "@metamask/bridge-controller": ^42.0.0
     "@metamask/gas-fee-controller": ^24.0.0
     "@metamask/network-controller": ^24.0.0
     "@metamask/snaps-controllers": ^14.0.0


packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +100,62 @@ export class EvmAccountProvider extends BaseAccountProvider {
...
+   *
+   * NOTE: This method should only be called on a newly created wallet.
+   * There should be already one existing account on this associated entropy source.
+   *
+   * @param opts - The options for the discovery and creation of accounts.
+   * @param opts.entropySource - The entropy source to use for the discovery and creation of accounts.
+   * @param opts.groupIndex - The index of the group to create the accounts for.
+   * @returns The accounts for the Evm provider.
+   */
+  async discoverAndCreateAccounts(opts: {
```
fabiobozzo, line 128: Question ❓ 💡 

How does this discovery NOT conflict with the extension? 
https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/metamask-controller.js#L5543

Or will you disable the extension's too?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -28,13 +31,34 @@ function assertInternalAccountExists(
...
 
+  /**
+   * Get the Evm provider.
+   *
+   * @returns The Evm provider.
+   */
+  getEvmProvider(): Provider {
+    const networkClientId = this.messenger.call(
+      'NetworkController:findNetworkClientIdByChainId',
+      '0x1',
```
fabiobozzo: What about other evm chains? 🤔

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
```diff
@@ -42,8 +42,14 @@ export abstract class BaseAccountProvider
 {
   protected readonly messenger: MultichainAccountServiceMessenger;
 
-  constructor(messenger: MultichainAccountServiceMessenger) {
+  readonly providerType: AccountProviderType;
```
fabiobozzo: Dummy question: 🙃 what do you use that prop for? Is it just used as a group key for the output of `MultichainAccountWallet.discoverAndCreateAccounts` ?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -326,4 +326,101 @@ export class MultichainAccountWallet<
...
+   * NOTE: This method should only be called on a newly created wallet.
+   *
+   * @param opts - The options for the discovery and creation of accounts.
+   * @param opts.skippedProviders - The providers to skip.
+   * @returns The accounts for each provider.
+   */
+  async discoverAndCreateAccounts({
+    skippedProviders = [],
+  }: {
+    skippedProviders: AccountProviderType[];
```
ccharly: I feel like having this forces us to have an `AccountProviderType`, which I believe we don't need it because:
- It prevents custom account providers (since they cannot extend the `enum`)
- We don't really need to identify any of the provider for now in the system

That being said, that could be useful for debugging purposes, but I'd rather prefer having a `string` identifier which could resolve to anything that makes sense for the provider (`snapId` for Snap-based providers, the `keyring.type` for keyring-based providers for example).

WDYT?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -326,4 +326,101 @@ export class MultichainAccountWallet<
...
+      (provider) => !skippedProviders.includes(provider.providerType),
+    );
+
+    const results: Record<AccountProviderType, Bip44Account<KeyringAccount>[]> =
+      {};
+    const nextIndex = new Map<Bip44Provider, number>(
+      providers.map((p) => [p, 0]),
+    );
+    const stopped = new Set<Bip44Provider>();
+    const inFlight = new Map<Bip44Provider, Promise<void>>();
```
ccharly: WDYT about using a single `Map` for this with something like:
```ts
const contexts = new Map<Bip44Provider, {
  stopped: boolean;
  running: Promise<void>;
  groupIndex: number;
}>();
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -326,4 +326,101 @@ export class MultichainAccountWallet<
...
+
+    let high = 0;
```
ccharly: I'd go with `maxGroupIndex` maybe?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -326,4 +326,101 @@ export class MultichainAccountWallet<
...
+      const run = (async () => {
+        try {
+          const accounts = await p.discoverAndCreateAccounts({
+            entropySource: this.#entropySource,
+            groupIndex: index,
+          });
+
+          inFlight.delete(p);
+
+          if (accounts.length > 0) {
```
ccharly: It should be `!accounts.length` here no? We stop the discovery if we don't find any accounts for a given index I'd say?

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: We could even add the `accounts` in there too, so we can extract `results` from it afterward.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -326,4 +326,101 @@ export class MultichainAccountWallet<
...
+      inFlight.set(p, run);
+    };
+
+    for (const p of providers) {
+      schedule(p, 0);
+    }
+
+    while (inFlight.size > 0) {
+      await Promise.race(inFlight.values());
+    }
```
ccharly: I wonder if this is truly concurrent-safe here 🤔 Since we're using `.values()` we will be using iterators, and the first "scheduled job" might alter the `inFlight` map. So maybe we should use `Array.from(inFlight.values())` just in case?

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Also, using an object might remove every `?? 0` we have 🤔

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: We will either feature flag this or entirely remove the client code 😄

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: Good question. There's two considerations here:

1. EVM addresses are the same across the namespace so that's not an issue.
2. There could be activity on one network and not the other (e.g. no activity on Ethereum mainnet, but activity on Polygon). Currently, IIRC, we just check for Ethereum mainnet activity? Or actually, I think at least in the extension the provider that gets grabbed is the currently selected network. I'm not sure what the means with the removal of the GNS. I feel like it would be a lot more complicated to cycle through the existing added networks and run discovery on them. Wondering if we should just use EVM mainnet activity as the rule here?

cc: @ccharly

packages/multichain-account-service/src/providers/BaseAccountProvider.ts
hmalik88: yup!

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Keeping the enum is not a hill I'm willing to die on, so I'm ok with removing the type...WDYT about removing the param as well then? I primarily had it added in because I noticed the clients were skipping SOL discovery at some points, but if we'll be forcing (for lack of a better word) our vision of how discovery implementation should be at the various points in import SRP process and just doing discovery for all active providers...then maybe we don't need it?

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Good point, I'll change.

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Ah yes, accident 🤓

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: Yes, I initially checked the extension's code and they indeed get the currently selected network, but.. IDK what would become this "selected network" with the removal of GNS either 😅

I think, we can probably mimic what's done on the extension for now and re-visit it later?

But I agree that looping through all networks might just kill the performance here.

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Yes, I would just remove it entirely, definitely.

If we really need to disable some providers during the discovery, we'll think of something different probably, but honestly, I don't think we would want to do this (our new solution should solve some older issues we had with discovery 🤞)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: Opting to keep as is because of the following reasons:

1. We know that selected network will not be a thing in the near future, so it's best to move away from having that assumption.
2. We can't cycle through every EVM network for activity (performance wise that's an unacceptable criteria) and let that be the guideline to adding an account, so it's best to default to the OG and check Ethereum mainnet. At the very least, there is some protection there for a "first" account since this discovery starts for the EVM provider at index 1 (there's already an account existing as the keyring is created).

packages/multichain-account-service/src/types.ts
```diff
@@ -122,3 +130,7 @@ export type MultichainAccountServiceMessenger = RestrictedMessenger<
   AllowedActions['type'],
   AllowedEvents['type']
 >;
+
```
hmalik88: Added the newly added type alias from the accounts-api just for testing purposes, will swap out once I cut a release.

packages/multichain-account-service/src/types.ts
```diff
@@ -122,3 +138,31 @@ export type MultichainAccountServiceMessenger = RestrictedMessenger<
...
+export type Bip44AccountProvider = AccountProvider<
+  Bip44Account<KeyringAccount>
+>;
+
+/**
+ * Locally widen the withKeyring action so the messenger accepts the optional options
+ * parameter (createIfMissing/createWithData) in addition to selector and operation.
+ */
+export type KeyringControllerWithKeyringWithOptionsAction = {
```
hmalik88: This relaxed action type was created to circumvent the function signature that typescript defaults to which is without options.

packages/multichain-account-service/src/providers/SolAccountProvider.test.ts
```diff
@@ -239,17 +240,54 @@ describe('SolAccountProvider', () => {
...
+
+    const discovered = await provider.discoverAndCreateAccounts({
+      entropySource: MOCK_HD_KEYRING_1.metadata.id,
+      groupIndex: 0,
+    });
+
+    expect(discovered).toStrictEqual([MOCK_SOL_ACCOUNT_1]);
+  });
+
+  it('should not return any accounts if no account is discovered', async () => {
```
ccharly:
```suggestion
  it('does not return any accounts if no account is discovered', async () => {
```

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `discoverAndCreateAccounts` methods for Evm and Solana providers ([#6397](https://github.com/MetaMask/core/pull/6397))
```
ccharly: Nit: I think EVM is sightly more correct than Evm (given that it's an acronym), WDYT?

```suggestion
- Add `discoverAndCreateAccounts` methods for EVM and Solana providers ([#6397](https://github.com/MetaMask/core/pull/6397))
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,58 @@ export class MultichainAccountService {
     return Array.from(this.#wallets.values());
   }
 
+  /**
+   * Creates a new multichain account wallet with the given mnemonic.
+   *
+   * NOTE: This method should only be called in client code where a mutex lock is acquired.
```
ccharly: You meant when the mutex is NOT acquired? (if you're referring to the `KeyringController` global mutex).

I guess to avoid dead-locks with a double-lock on that mutex, right?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,58 @@ export class MultichainAccountService {
...
+      },
+      { createIfMissing: true, createWithData: { mnemonic } },
+    )) as [Hex[], KeyringMetadata];
+
+    const [accounts, metadata] = result;
+
+    // Make sure the keyring has no accounts after creating it.
+    if (accounts.length > 0) {
+      throw new Error('Expected keyring with no accounts');
+    }
```
ccharly: I think this logic will never fail (unless we mock `withKeyring` like in your tests).

We might want to check if any HD keyrings already uses those mnemonics. That's the logic we use in the extension:
- https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/metamask-controller.js#L5220-L5230

This would probably require to pull some more code (like the one to convert to actual english words), but that sounds like a much safer option to me.

Given that it could scope creep a bit the "discover PR" here, we might want to do it it another PR with that new logic? Also, this `withKeyring` variant is marked as deprecated, so we should use `addNewKeyring` instead (after having checked that this SRP was not in-use already OFC).

WDYT?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+
+    const discoveredAccounts: Record<string, number> = {};
+
+    for (const [p, ctx] of providerContexts) {
+      /* istanbul ignore next
+       * In production, Snap providers exist and hit the `isSnapAccountProvider`
+       * branch. Unit tests here mock generic providers only, so that path is
+       * not exercised.
+       */
+      const groupKey = isSnapAccountProvider(p) ? p.snapId : 'Evm';
```
ccharly: To avoid having this, we could also add a new type `NamedAccountProvider<Account>` (only defined in the service) which would extends `AccountProvider<Account>` and add a new method `getName()` so we can define this name in the providers themselves?

We would any kind of conditional here, and it could also be useful in the service/wallets/groups if need to have logging with attached with the provider name.

WDYT?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+    }
+
+    let maxGroupIndex = 0;
+
+    const schedule = (p: Bip44AccountProvider, index: number) => {
+      // Ok to cast here because we know that the ctx will always be set.
+      const providerCtx = providerContexts.get(p) as {
+        stopped: boolean;
+        running?: Promise<void>;
+        index: number;
```
ccharly: I'd just go with `groupIndex` instead of `index`. I was mixing both terms in the past, and I think it can confuse people sometimes. Better be explicit IMO.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+        count: number;
+      };
+
+      providerCtx.index = index;
+
+      providerCtx.running = (async () => {
```
ccharly: While the current logic looks good here, WDYT about future-proofing this with:
```ts
if (providerCtx.running) {
  // Wait for current scheduled job before starting a new one.
  await providerCtx.running;
  providerCtx.running = undefined;
}
```

I don't think this should break anything, but it makes it a little safer if we add new `schedule` entrypoint without checking for `running` first.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+        try {
+          const accounts = await p.discoverAndCreateAccounts({
+            entropySource: this.#entropySource,
+            groupIndex: index,
+          });
+
+          if (!accounts.length) {
```
ccharly: Just to add slightly more context. (I tend to be overly explicit, but maybe you don't want that 😅 LMK what you think!)

```suggestion
          // This provider found no accounts, future discovery won't yield
          // any new accounts, so we marked it as "stopped".
          if (!accounts.length) {
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+            providerCtx.stopped = true;
+            return;
+          }
+
+          providerCtx.count += accounts.length;
+
+          const next = index + 1;
```
ccharly: Same here, `nextGroupIndex` would be pretty cool and similar to `maxGroupIndex`.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+            const target = Math.max(maxGroupIndex, providerCtx.index);
+            schedule(p, target);
+          }
+        }
+      })();
+    };
+
+    for (const p of providers) {
+      schedule(p, 0);
+    }
```
ccharly: Small suggestion here:

```suggestion
    // Start discovery for every providers.
    for (const p of providers) {
      schedule(p, 0);
    }
```

Also, I wonder if we should make the discovery aware of the current highest index from the wallet? Like, it would be awesome to be able to re-run a discovery where we left it off (like if we shutdown MM in the middle of the discovery, we start again from index N-something).

Lastly, we could re-use the discovery to try to discover new accounts that we don't know about when we "unlock" the extension for example? (Imagine you have created account on another wallet, MM won't know about those accounts yet, but we could discovery them starting from our highest known index).

WDYT? If yes, then I think it's just a matter of using `getNextGroupIndex` here instead.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +357,122 @@ export class MultichainAccountWallet<
...
+
+    while ([...providerContexts.values()].some((ctx) => Boolean(ctx.running))) {
+      const racers = [...providerContexts.values()]
+        .map((c) => c.running)
+        .filter(Boolean) as Promise<void>[];
+      await Promise.race(racers);
+    }
```
ccharly: Maybe we can turn this into a `do while` instead? 🤔

```suggestion
    do {
      const racers = [...providerContexts.values()]
        .map((c) => c.running)
        .filter(Boolean) as Promise<void>[];
      
      if (racers.length) {
        await Promise.race(racers);
      }
    } while (racers.some(Boolean));
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,58 @@ export class MultichainAccountService {
...
+
+    const wallet = new MultichainAccountWallet({
```
fabiobozzo: Just to help me understand the flow: 
- new hd keyring is created
- we expect `accounts` to have one and only one evm account in it 
- we create a new wallet w/ that entropy source
- will the alignment run and create a new solana account, when the wallet is constructed?

fabiobozzo (approved):

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: No, I meant that this method should be called inside of a block of code where the `createVault` mutex has been acquired: https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/metamask-controller.js#L5218, not sure what the equivalent is in mobile.

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: The client code checks if the keyring has already been created, so I was expecting this method to be called after that point. BUT, I was also just considering moving that part of the code into this, I suppose it'll make it even cleaner. WDYT?

`addNewKeyring` is not exposed to the messaging system, so I had to use this workaround. An alternative is to just pass that function to this method from the client, but I don't feel too good about doing that either.

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Yeah, I can do that.

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Not necessarily true that discovery won't yield accounts past this point, but I'll write a comment that this is a stopping condition in our discovery.

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Yeah, that's valid. I'll change it.

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Yup, makes sense.

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: The steps are as follows:

1. New HD keyring is created with **no** accounts.
2. We create a wallet with that entropy source id.
3. We return that wallet and associated entropy source id. We expect that client to run `wallet.discoverAndCreateAccounts` on that said wallet. The wallet will start discovery from `groupIndex` 0 for all providers and then run alignment on said wallet. The client can consume the returned entropy source id for other things (I believe the seedless onboarding code uses the entropy source id).

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: oh.. I was not aware we had this mutex in place 🤔 seems similar to the `withKeyring` one.. but ok, I guess we can leave the comment for now)

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Moving this piece of code there sounds cleaner to me yes 👍

As for the `addNewKeyring` then I guess we might not need it if we check for existing SRPs beforehand. Though, `withKeyring` and `createWithData` are considered deprecated.

So I wonder if we should expose `addNewKeyring` through the messaging system 🤔 (we can check with @mikesposito if there's any other alternative to this).

mikesposito: > So I wonder if we should expose `addNewKeyring` through the messaging system 🤔 (we can check with @mikesposito if there's any other alternative to this). 

Exposing `addNewKeyring` through a new messenger action sounds like a good idea to me, and I can see it being useful for other use cases as well

ccharly: Here's the PR to add it:
- https://github.com/MetaMask/core/pull/6439

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -15,8 +19,34 @@ import { SnapAccountProvider } from './SnapAccountProvider';
...
+
   constructor(messenger: MultichainAccountServiceMessenger) {
     super(SolAccountProvider.SOLANA_SNAP_ID, messenger);
+    this.#client = this.#getKeyringClientFromSnapId(
+      SolAccountProvider.SOLANA_SNAP_ID,
+    );
+  }
+
+  getName(): string {
+    return 'Solana';
```
ccharly, line 31: We could re-use the Snap ID too?

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: > We expect that client to run wallet.discoverAndCreateAccounts

where should that be triggered? similar places to alignWallets?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,58 @@ export class MultichainAccountService {
...
+   */
+  async createMultichainAccountWallet({
+    mnemonic,
+  }: {
+    mnemonic: string;
+  }): Promise<
+    [MultichainAccountWallet<Bip44Account<KeyringAccount>>, EntropySourceId]
+  > {
+    // create a new wallet with the given mnemonic
+    const result = (await this.#messenger.call(
```
fabiobozzo: To be replaced with `addNewKeyring` from https://github.com/MetaMask/core/pull/6439 ?

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Various parts where importing an SRP is expected, I think @ccharly will be handling this part. The entry points are defined in the notion doc that I created.

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Yup!

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Done: [0d8fd74](https://github.com/MetaMask/core/pull/6397/commits/0d8fd74e98e31b245b63746c826385cd5ac29b1d)

packages/multichain-account-service/src/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const convertMnemonicToWordlistIndices = (
+  mnemonic: string,
+): Uint8Array => {
+  const indices = mnemonic
+    .toString()
+    .split(' ')
+    .map((word) => wordlist.indexOf(word));
+  return new Uint8Array(new Uint16Array(indices).buffer);
+};
```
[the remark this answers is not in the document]
hmalik88: Not an issue: encoding and decoding happens in a JS environment which is little endian. Odd length array is also a non-issue since importing an SRP is constrained to 12 or 24 word mnemonics.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,57 @@ export class MultichainAccountService {
...
+    ) as (EthKeyring & { mnemonic: Uint8Array })[];
+
+    const alreadyHasImportedSrp = existingKeyrings.some((keyring) => {
+      return (
+        Buffer.from(
+          convertEnglishWordlistIndicesToCodepoints(keyring.mnemonic),
+        ).toString('utf8') === mnemonic
+      );
+    });
+
```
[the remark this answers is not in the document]
hmalik88: It's not an unsafe assumption 😑, HD keyrings do have `mnemonic: Uint8Array`. Fair point about the double `Buffer.from`, you're only slightly annoying, Cursor.

packages/multichain-account-service/src/utils.ts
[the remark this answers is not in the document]
hmalik88: Same as above, not an issue.

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+### Added
+
+- Add `discoverAndCreateAccounts` methods for EVM and Solana providers ([#6397](https://github.com/MetaMask/core/pull/6397))
+- Add `discoverAndCreateAccounts` method to `MultichainAccountWallet` to orchestrate provider discovery ([#6397](https://github.com/MetaMask/core/pull/6397))
+- Add `createMultichainAccountWallet` method to create a new multichain account wallet from a mnemonic ([#6397](https://github.com/MetaMask/core/pull/6397))
+  - An action handler was also registered for this method so that it can be called from the clients.
+
 ### Changed
```
ccharly, line 17: We need to add a `**BREAKING:**` entry because we now need to register `NetworkController` and `KeyringController` actions to the service's messenger.

packages/multichain-account-service/src/types.ts
```diff
@@ -122,3 +138,13 @@ export type MultichainAccountServiceMessenger = RestrictedMessenger<
...
+
+/**
+ * The context for a provider discovery.
+ */
+export type ProviderDiscoveryContext = {
+  stopped: boolean;
+  running?: Promise<void>;
+  groupIndex: number;
+  count: number;
+};
```
ccharly: We can remove it now, we moved it to the wallet file instead.
```suggestion
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -35,6 +37,27 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+
+  /**
+   * Get the Evm provider.
+   *
+   * @returns The Evm provider.
+   */
+  getEvmProvider(): Provider {
+    const networkClientId = this.messenger.call(
+      'NetworkController:findNetworkClientIdByChainId',
+      '0x1',
```
ccharly: Maybe we could make a constant out of this one? 🤔 Unless we have them somewhere in some other packages.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -35,6 +37,27 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
   }
 
+  getName(): string {
+    return 'EVM';
+  }
+
+  /**
+   * Get the Evm provider.
+   *
+   * @returns The Evm provider.
```
ccharly:
```suggestion
   * Get the EVM provider.
   *
   * @returns The EVM provider.
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,57 @@ export class MultichainAccountService {
...
+   * @param options.mnemonic - The mnemonic to use to create the new wallet.
+   * @throws If the mnemonic has already been imported.
+   * @returns The a tuple of the new multichain account wallet and the entropy source id.
+   */
+  async createMultichainAccountWallet({
+    mnemonic,
+  }: {
+    mnemonic: string;
+  }): Promise<
+    [MultichainAccountWallet<Bip44Account<KeyringAccount>>, EntropySourceId]
```
ccharly: I think returning the wallet is enough, since we can get the entropy source from it afterward (with `wallet.entropySource`).

packages/multichain-account-service/src/utils.ts
ccharly: Nit: But I think naming this file `mnemonic.ts` would have been cool! WDYT?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +99,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
     return accountsArray;
   }
 
-  async discoverAndCreateAccounts(_: {
+  /**
+   * Discover and create accounts for the Evm provider.
```
ccharly:
```suggestion
   * Discover and create accounts for the EVM provider.
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +99,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+   *
+   * NOTE: This method should only be called on a newly created wallet.
+   *
+   * @param opts - The options for the discovery and creation of accounts.
+   * @param opts.entropySource - The entropy source to use for the discovery and creation of accounts.
+   * @param opts.groupIndex - The index of the group to create the accounts for.
+   * @returns The accounts for the Evm provider.
```
ccharly:
```suggestion
   * @returns The accounts for the EVM provider.
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +365,89 @@ export class MultichainAccountWallet<
       this.#isAlignmentInProgress = false;
     }
   }
+
+  /**
+   * Discover and create accounts for all providers.
+   *
+   * NOTE: This method should only be called on a newly created wallet.
+   *
```
ccharly: I think this is no longer true? We can run discovery on any wallets now since we're starting from its highest known group index?
```suggestion
```

packages/multichain-account-service/src/providers/SolAccountProvider.ts
hmalik88: My intention was to keep this as a human readable name so it can just be used right out of the returned discovery count in the clients

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: You mean the '0x1'?

packages/multichain-account-service/src/utils.ts
hmalik88: I can change it

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: ah true, i forgot the getter is there

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: yup, will remove

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: Yes, I meant this one. I feel like it reads better if we use `ETH_MAINNET` or something like that?

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -832,6 +851,27 @@ describe('MultichainAccountService', () => {
...
+      }));
+
+      const wallet = await messenger.call(
+        'MultichainAccountService:createMultichainAccountWallet',
+        { mnemonic: MOCK_MNEMONIC },
+      );
+
+      expect(wallet).toBeDefined();
+      expect(wallet.entropySource).toBe('abc');
+    });
```
ccharly: I think this is a duplicate (we have it in the `describe('createMultichainAccountWallet', ...)` block):

```suggestion
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,55 @@ export class MultichainAccountService {
...
+   */
+  async createMultichainAccountWallet({
+    mnemonic,
+  }: {
+    mnemonic: string;
+  }): Promise<MultichainAccountWallet<Bip44Account<KeyringAccount>>> {
+    const existingKeyrings = this.#messenger.call(
+      'KeyringController:getKeyringsByType',
+      KeyringTypes.hd,
+    ) as (EthKeyring & { mnemonic: Uint8Array })[];
```
ccharly: Type-wise, this field could be `null` or `undefined`, [see here](https://github.com/MetaMask/accounts/blob/main/packages/keyring-eth-hd/src/hd-keyring.ts#L107).

In practice, I assume it's always defined, but maybe we want to align the type here too? We could also have `eth-hd-keyring` as a dev dep and use the `HdKeyring` type here too.

Which would force to handle the "null case".

WDYT?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,55 @@ export class MultichainAccountService {
...
+  /**
+   * Creates a new multichain account wallet with the given mnemonic.
+   *
+   * NOTE: This method should only be called in client code where a mutex lock is acquired.
+   * `discoverAndCreateAccounts` should be called after this method to discover and create accounts.
+   *
+   * @param options - Options.
+   * @param options.mnemonic - The mnemonic to use to create the new wallet.
+   * @throws If the mnemonic has already been imported.
+   * @returns The a tuple of the new multichain account wallet and the entropy source id.
```
ccharly:
```suggestion
   * @returns The new multichain account wallet.
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +99,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
     return accountsArray;
   }
 
-  async discoverAndCreateAccounts(_: {
+  /**
+   * Discover and create accounts for the EVM provider.
+   *
+   * NOTE: This method should only be called on a newly created wallet.
+   *
```
ccharly:
```suggestion
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +99,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+    // We don't want to remove the account if it's the first one.
+    if (count === 0 && didCreate && groupIndex !== 0) {
+      await this.withKeyring<EthKeyring>(
+        { id: entropySource },
+        async ({ keyring }) => {
+          keyring.removeAccount?.(address);
+        },
+      );
+      return [];
+    }
```
ccharly: Should we wrap this in a `try finally`? So, if anything goes wrong with the provider request, we would still `removeAccount` the created account?

hmalik88:
@metamaskbot publish-preview

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +101,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+      }
+      const [added] = await keyring.addAccounts(1);
+      return [added, true];
+    });
+
+    const countHex = (await provider.request({
+      method: 'eth_getTransactionCount',
+      params: [address, 'latest'],
+    })) as Hex;
+    const count = parseInt(countHex, 16);
```
ccharly: Nit: We could have moved that into a separate method, like `#checkAddressActivity`?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +101,58 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+      EthKeyring,
+      [Hex, boolean]
+    >({ id: entropySource }, async ({ keyring }) => {
+      const existing = await keyring.getAccounts();
+      if (groupIndex < existing.length) {
+        return [existing[groupIndex], false];
+      }
+      const [added] = await keyring.addAccounts(1);
+      return [added, true];
+    });
```
ccharly: Since this is very similar to `createAccounts`, WDYT about adding a `#createAccount` method (which would exactly this code).

So we can re-use it here AND in `createAccounts` (we would have no use of the 2nd element of the tuple for `createAccounts`, but that's not a problem IMO).

WDYT?

packages/multichain-account-service/src/MultichainAccountService.test.ts
hmalik88: oh it's needed to cover the action i believe

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +368,87 @@ export class MultichainAccountWallet<
       this.#isAlignmentInProgress = false;
     }
   }
+
+  /**
+   * Discover and create accounts for all providers.
+   *
+   * @returns The discovered accounts for each provider.
+   */
+  async discoverAndCreateAccounts(): Promise<Record<string, number>> {
```
ccharly: Actually, I was thinking of something, instead of using a raw `Record` here, WDYT about using a more "complex" object.

I thought we could add a `DiscoveryStats` object, like:
```ts
type DiscoveryStats = {
  providers: {
    [providerName: string]: {
      discoveredAndCreated: number;
      groupIndexes: number[],
      error: boolean,
    },
  },
  total: {
    discoveredAndCreated: number;
    groupIndexes: number[],
  },
}
```

Or something like that, where we would track those stats during the discovery, thus, giving us a lot of information to return after the discovery ended.

This way we can use those stats for our metrics and ideally compute any kind of useful stat/metric we need from that?

That sounds a bit more complex than what we have, but it would makes more sense to me.

WDYT?

hmalik88: I think it would be overkill, because I'm not sure how meaningful the particular group indices that were discovered would be to our metrics beyond the total count we're already calculating for each provider. We also are logging each step as well which I think IS useful for debugging purposes, but again not sure how that is useful to our metrics that say a network request failed.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: Yeah, makes sense. I'll add it.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: hmm I think it's fine to keep as is since I don't see that method being used elsewhere

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: yup, makes sense!

packages/multichain-account-service/src/MultichainAccountService.test.ts
hmalik88: Changed the description, you're right that the description was a duplicate.

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: aha, yes true 😄 just wanted to share just in case 😉

Though, I wonder if we should wrap the `Record` in another type-alias to give it a bit more context around it 🤔

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -223,17 +273,74 @@ describe('EvmAccountProvider', () => {
...
+      scopes: [EthScope.Eoa],
+      type: EthAccountType.Eoa,
+      metadata: {
+        name: 'Account 1',
+        keyring: { type: KeyringTypes.hd },
+        importTime: 0,
+        lastSelected: 0,
+        nameLastUpdatedAt: 0,
+      },
+    };
```
ccharly: Nit: We could have used the `MOCK_HD_ACCOUNT_1` for this maybe?

Like:
```ts
const expectedAccount = {
  ...MockAccountBuilder.from(MOCK_HD_ACCOUNT_1).withAddressSuffix(0).get(),
  id: expect.any(String),
}
```

hmalik88: True

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: yeah, I can do that.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -60,10 +90,24 @@ export class SolAccountProvider extends SnapAccountProvider {
...
     groupIndex: number;
   }): Promise<Bip44Account<KeyringAccount>[]> {
-    return []; // TODO: Implement account discovery.
+    const discoveredAccounts = await this.#client.discoverAccounts(
+      [SolScope.Mainnet],
+      entropySource,
+      groupIndex,
+    );
+
+    // Solana Snap always returns one discovered account.
```
danroc: Although it's currently true, I think we shouldn't make this assumption.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -35,33 +39,71 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+    return result;
+  }
+
+  async createAccounts({
+    entropySource,
+    groupIndex,
+  }: {
+    entropySource: EntropySourceId;
+    groupIndex: number;
+  }) {
```
danroc: Nit: Some return types are missing (I thought it was enforced by the linter).

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +302,58 @@ export class MultichainAccountService {
...
+   * @param options - Options.
+   * @param options.mnemonic - The mnemonic to use to create the new wallet.
+   * @throws If the mnemonic has already been imported.
+   * @returns The new multichain account wallet.
+   */
+  async createMultichainAccountWallet({
+    mnemonic,
+  }: {
+    mnemonic: string;
+  }): Promise<MultichainAccountWallet<Bip44Account<KeyringAccount>>> {
```
ccharly: We might want to re-consider adding this method for now.

I just saw that the mobile app was not up-to-date (missing 1 major) with the `KeyringController`'s version, and we do need the next version of this controller for the new `:addNewKeyring` action.

I think we can just split up this PR in 2:
- Discovery (which has been heavily reviewed and should be almost good to go)
- New wallet method
  * This could be a separate PR, cause I know @danroc had some suggestion for this implementation too.

Do you think we can do that @hmalik88?

packages/multichain-account-service/src/mnemonic.ts
danroc, line 1: I would consider moving these functions to a different package (like `key-tree` or `utils`). I'll let @ccharly share his opinion on this.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
hmalik88: What would you prefer?

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Why not just bring mobile UTD with the latest version?

packages/multichain-account-service/src/mnemonic.ts
hmalik88: Yeah, thought of that but was keeping them here for now for the sake of moving things along faster. We can always move it later.

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: We could do this too, but it's not a "simple major" and also require QA validation. So I cannot say how long this would take.

- https://github.com/MetaMask/metamask-mobile/pull/17157

IMO, it's worth to split it, so we can have discussion about mnemonic and wallet other creation topics on the other PR.

I don't think it's super mandatory to have this action (even though, I think it's definitely better to have it on the service and use it on the client).

However, the discovery is really high-prio for now 😅

packages/multichain-account-service/src/mnemonic.ts
ccharly: Would make sense to move it somewhere else indeed. `keyring-utils` could also be a good candidate actually 🤔 (it's also pretty fast to review/deploy)

packages/multichain-account-service/src/providers/SolAccountProvider.ts
danroc: We should create all accounts that were discovered, regardless of the number.

hmalik88: How would it ever be more than one?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -11,13 +11,30 @@ import type {
...
 } from '@metamask/keyring-api';
+import { createProjectLogger } from '@metamask/utils';
 
 import { MultichainAccountGroup } from './MultichainAccountGroup';
+import type { NamedAccountProvider } from './providers';
+
+/**
+ * The context for a provider discovery.
+ */
+export type AccountProviderDiscoveryContext = {
```
ccharly: Do we need to export this one? 🤔

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -11,13 +11,30 @@ import type {
...
+  provider: NamedAccountProvider;
+  stopped: boolean;
+  groupIndex: number;
+  count: number;
+};
+
+export type AccountDiscoveryMetrics = {
```
ccharly, line 36:
```suggestion
/**
 * Account discovery metrics result.
 */
export type AccountDiscoveryMetrics = {
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -11,13 +11,30 @@ import type {
...
+  [providerName: string]: number;
+};
+
+const log = createProjectLogger('multichain-account-service');
```
ccharly, line 40: Note for later: we will move this into a `logger.ts` file, so we can use it in other part of the service too.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +372,87 @@ export class MultichainAccountWallet<
...
+    await this.alignGroups();
+
+    const discoveredAccounts: Record<string, number> = {};
+    for (const context of providerContexts) {
+      const providerName = context.provider.getName();
+      discoveredAccounts[providerName] = context.count;
+    }
+
+    return discoveredAccounts;
+  }
```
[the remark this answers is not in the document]
ccharly: Not sure about the race condition. This looks to me 🤔

So, cursor is right about the group index we're logging here. 🙈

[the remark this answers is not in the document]
hmalik88: There doesn't seem to be an issue in practice with incorrect fast-forwarding. Fixed the incorrect logging, though.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -355,4 +372,87 @@ export class MultichainAccountWallet<
...
+    const runProviderDiscovery = async (
+      context: AccountProviderDiscoveryContext,
+    ) => {
+      const message = (stepName: string) =>
+        `[${context.provider.getName()}] Discovery ${stepName} (groupIndex=${context.groupIndex})`;
+
+      while (!context.stopped) {
+        // Fast‑forward to current high‑water mark
+        const targetGroupIndex = Math.max(context.groupIndex, maxGroupIndex);
+
```
ccharly: This could maybe [fix cursor report](https://github.com/MetaMask/core/pull/6397/files#r2325096911):

```suggestion
      while (!context.stopped) {
        // Fast‑forward to current high‑water mark
        const targetGroupIndex = Math.max(context.groupIndex, maxGroupIndex);
      
        const message = (stepName: string) =>
          `[${context.provider.getName()}] Discovery ${stepName} (groupIndex=${targetGroupIndex})`;

```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -35,33 +39,71 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
-        if (groupIndex < accounts.length) {
-          // Nothing new to create, we just re-use the existing accounts here,
-          return [accounts[groupIndex]];
+        const existing = await keyring.getAccounts();
+        if (groupIndex < existing.length) {
+          return [existing[groupIndex], false];
         }
 
         // For now, we don't allow for gap, so if we need to create a new
         // account, this has to be the next one.
```
ccharly: I think this comment could be updated now with the `throwOnGap`.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -35,33 +39,71 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
-        if (groupIndex !== accounts.length) {
+        if (groupIndex !== existing.length && throwOnGap) {
```
ccharly: Very nit: But checking for `throwOnGap` before checking the condition sounds slightly better to me, WDYT?

```suggestion
        if (throwOnGap && groupIndex !== existing.length) {
```

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -76,10 +118,64 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+      // remove it to avoid leaving a dangling account.
+      if (shouldCleanup) {
+        await this.withKeyring<EthKeyring>(
+          { id: entropySource },
+          async ({ keyring }) => {
+            keyring.removeAccount?.(address);
+          },
+        );
+      }
+      return [];
```
ccharly: I wonder if we should re-throw here actually... We already handle errors in the discovery orchestrator. This way we would log the error.

If we return `[]` this will be seen as a "nothing to discover" basically.

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: You already resolved that one actually, closing.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
ccharly: It could be more than 1 if the Solana Snap decide to support non-standard derivation path for Solana.

Let's say another wallet (software) uses a non-standard derivation path, and we want to be able to "discover" those accounts, then the Solana Snap could return more than one account.

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: See:
- https://github.com/MetaMask/core/pull/6478

packages/multichain-account-service/src/providers/SolAccountProvider.ts
hmalik88: Ok gotcha, i'll change.

PR #6398, opened 2025-08-27
feat: add-raw-balance-to-selector

bergarces:
## Explanation

Adds `rawBalance` to new asset selector as requested by @MetaMask/confirmations .

## Checklist

### DIFF
--- a/packages/assets-controllers/src/selectors/stringify-balance.ts
+++ b/packages/assets-controllers/src/selectors/stringify-balance.ts
@@ -1,6 +1,8 @@
 // From https://github.com/MetaMask/eth-token-tracker/blob/main/lib/util.js
 // Ensures backwards compatibility with display formatting.
 
+import { bigIntToHex, type Hex } from '@metamask/utils';
+
 /**
  * @param balance - The balance to stringify as a decimal string
  * @param decimals - The number of decimals of the balance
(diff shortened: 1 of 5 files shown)


packages/assets-controllers/src/selectors/stringify-balance.ts
```diff
@@ -46,3 +48,46 @@ export function stringifyBalanceWithDecimals(
...
+ * This is the inverse operation of stringifyBalanceWithDecimals.
+ *
+ * @param balanceString - The decimal string representation (e.g., "123.456")
+ * @param decimals - The number of decimals to apply (shifts decimal point right)
+ * @returns The balance as a bigint in the smallest unit
+ *
+ * @example
+ * parseBalanceWithDecimals("123.456", 18) // Returns 123456000000000000000n
+ * parseBalanceWithDecimals("0.001", 18)   // Returns 1000000000000000n
+ * parseBalanceWithDecimals("123", 18)     // Returns 123000000000000000000n
```
Prithpal-Sooriya: Is the example up-to-date? It looks like it now returns a Hex val?

bergarces: Ah, thanks for raising. I started it returning a bigint, but it was giving me a lot of hassle with tests (you cannot run `toStrictEqual` with bigints because they cannot be serialised and deserialised to/from json), so I went for the less complicated Hex.

Will correct this.

PR #6400, opened 2025-08-27
feat: emit events on tree updates and group selection

fabiobozzo:
## Explanation

Adds event emissions to `AccountTreeController` to enable other controllers to react to state changes. Previously, said controller had no way to notify consumers when the account tree structure changed or when the selected account group was updated.

**Solution:**
- Adds `AccountTreeController:accountTreeUpdated` event - emitted when nodes (wallets, groups, or accounts) are added or removed
- Adds `AccountTreeController:selectedAccountGroupUpdated` event - emitted when the selected account group changes  
- Events are only emitted for structural/selection changes, not metadata updates (name, pinning, hiding)

This enables other controllers to listen for tree changes and react accordingly, supporting the upcoming Account Syncing V2 functionality.

## References

[MUL-523](https://consensyssoftware.atlassian.net/browse/MUL-523)

## Checklist

[MUL-523]: https://consensyssoftware.atlassian.net/browse/MUL-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/shield-controller/CHANGELOG.md
+++ b/packages/shield-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Added missing exports and improved documentation ([#6412](https://github.com/MetaMask/core/pull/6412))
+
 ## [0.1.0]
 
 ### Added
--- a/packages/shield-controller/src/ShieldController.ts
+++ b/packages/shield-controller/src/ShieldController.ts
@@ -15,15 +15,24 @@ import type { CoverageResult, ShieldBackend } from './types';
 const log = createModuleLogger(projectLogger, 'ShieldController');
 
 export type CoverageResultRecordEntry = {
-  results: CoverageResult[]; // history of coverage results, latest first
+  /**
+   * History of coverage results, latest first.
+   */
+  results: CoverageResult[];
 };
 
 export type ShieldControllerState = {
+  /**
+   * Coverage results by transaction ID.
+   */
   coverageResults: Record<
     string, // txId
     CoverageResultRecordEntry
   >;
-  orderedTransactionHistory: string[]; // List of txIds ordered by time, latest first
+  /**
+   * List of txIds ordered by time, latest first.
+   */
+  orderedTransactionHistory: string[];
 };
 
 /**
--- a/packages/shield-controller/src/index.ts
+++ b/packages/shield-controller/src/index.ts
@@ -12,3 +12,4 @@ export {
   ShieldController,
   getDefaultShieldControllerState,
 } from './ShieldController';
+export { ShieldRemoteBackend } from './backend';


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -462,6 +462,8 @@ export class AccountTreeController extends BaseController<
         }
       }
     });
+
+    this.messagingSystem.publish(`${controllerName}:accountTreeUpdated`);
```
ccharly: WDYT about also publishing `state.accountTree` here? That would be very similar to `:stateChange`?

fabiobozzo: It'd be handy, I agree! I added it to the payload. Please check it out

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -724,6 +734,14 @@ export class AccountTreeController extends BaseController<
       state.accountTree.selectedAccountGroup = groupId;
     });
 
+    this.messagingSystem.publish(
+      `${controllerName}:selectedAccountGroupUpdated`,
```
ccharly: WDYT about using `:selectedAccountGroupChange` instead? This would match the `:selectedAccountChange` we already have on the accounts-controller?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -501,6 +506,11 @@ export class AccountTreeController extends BaseController<
         }
       });
 
+      this.messagingSystem.publish(
+        `${controllerName}:accountTreeUpdated`,
```
ccharly: I also wonder if we should go for `:accountTreeChange` to re-use the same naming convention of `:stateChange` event 🤔 We already have some `*Updated` event in some other controllers, but they might not always be related to their state, while the `accountTree` here is one of the main element of this controller's state 😅 WDYT?

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2451,4 +2451,204 @@ describe('AccountTreeController', () => {
       expect(spy).toHaveBeenCalledWith(groupId, hidden);
     });
   });
+
+  describe('Event Emissions', () => {
+    it('should NOT emit accountTreeUpdated when tree is initialized', () => {
```
ccharly: Prefer the imperative form for tests:

```suggestion
    it('does NOT emit accountTreeUpdated when tree is initialized', () => {
```

NOTE: If you could apply all this to every tests as well, this would be awesome!

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Good point! I'll rename to `selectedAccountGroupChange` to match the `selectedAccountChange` convention from accounts-controller.

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Makes sense! I'll rename to accountTreeChange to align with the existing stateChange event naming pattern. Consistency is better.

packages/account-tree-controller/src/AccountTreeController.test.ts
fabiobozzo: Ooops sorry 🤦

fabiobozzo:
@metamaskbot publish-preview

packages/account-tree-controller/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/account-tree-controller",
-  "version": "0.11.0",
+  "version": "0.10.0",
```
ccharly:
```suggestion
  "version": "0.11.0",
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add missing export for `AccountTreeControllerGetAccountsFromSelectedAccountGroupAction` ([#6404](https://github.com/MetaMask/core/pull/6404))
-- Add `AccountTreeController:setAccount{WalletName,GroupName,GroupPinned,GroupHidden}` actions ([#6404](https://github.com/MetaMask/core/pull/6404))
+- Add `AccountTreeController:accountTreeChange` event emitted when nodes (wallets, groups, or accounts) are added or removed ([#6400](https://github.com/MetaMask/core/pull/6400))
+- Add `AccountTreeController:selectedAccountGroupChange` event emitted when the selected account group changes ([#6400](https://github.com/MetaMask/core/pull/6400))
```
ccharly: This will have to be moved under the `## Unrealeased` section

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "520.0.0",
+  "version": "516.0.0",
```
ccharly:
```suggestion
  "version": "520.0.0",
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2451,4 +2451,223 @@ describe('AccountTreeController', () => {
...
+
+      controller.init();
+
+      const currentSelectedGroup =
+        controller.state.accountTree.selectedAccountGroup;
+
+      jest.clearAllMocks();
+
+      // Only test if there's a selected group (not empty string)
+      if (currentSelectedGroup) {
```
ccharly: To avoid conditionals, maybe we can just type-cast to avoid having `| ''` in the type resolution?

We control the initial state, so we should be able to have a deterministic "flow"

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2451,4 +2451,223 @@ describe('AccountTreeController', () => {
...
+      messenger.publish(
+        'AccountsController:selectedAccountChange',
+        MOCK_HD_ACCOUNT_2,
+      );
+
+      const newSelectedGroup =
+        controller.state.accountTree.selectedAccountGroup;
+
+      // Only check if the group actually changed
+      if (newSelectedGroup !== initialSelectedGroup) {
```
ccharly: Same here, we should be able to control the initial state of the test, thus we should not need any conditionals here.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -782,6 +797,13 @@ export class AccountTreeController extends BaseController<
     this.update((state) => {
       state.accountTree.selectedAccountGroup = groupId;
     });
+    this.messagingSystem.publish(
+      `${controllerName}:selectedAccountGroupChange`,
+      {
+        selectedAccountGroup: groupId,
+        previousSelectedAccountGroup: currentSelectedGroup,
+      },
+    );
```
ccharly: WDYT about using 2 parameters instead? We could do this instead:

```suggestion
    this.messagingSystem.publish(
      `${controllerName}:selectedAccountGroupChange`,
      groupId,
      previousGroupId,
    );
```

This way, people can use it like:
```ts
this.messenger.subscribe('AccountTreeController:selectedAccountGroupChange', (groupId) => {
  ...
});

this.messenger.subscribe('AccountTreeController:selectedAccountGroupChange', (groupId, previousGroupId) => {
  ... // Do something with the previous group ID if we need it!
});
```

You can see this as an "optional" parameter basically, similar to the `Patch[]` for the `:stateChange` [here](https://github.com/MetaMask/core/blob/main/packages/base-controller/src/BaseController.ts#L148).

fabiobozzo: applied! (TIL)

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -723,6 +746,11 @@ export class AccountTreeController extends BaseController<
     this.update((state) => {
       state.accountTree.selectedAccountGroup = groupId;
     });
+    this.messagingSystem.publish(
+      `${controllerName}:selectedAccountGroupChange`,
+      groupId,
+      currentSelectedGroup,
```
ccharly: Would you be ok to rename `currentSelectedGroup` to `previousSelectedGroup` for consistency with other methods?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -782,6 +810,11 @@ export class AccountTreeController extends BaseController<
...
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -500,6 +504,10 @@ export class AccountTreeController extends BaseController<
           }
         }
       });
+      this.messagingSystem.publish(
+        `${controllerName}:accountTreeChange`,
+        this.state.accountTree,
+      );
```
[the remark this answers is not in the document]
ccharly: This seems wrong, right @fabiobozzo?

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2451,4 +2451,250 @@ describe('AccountTreeController', () => {
...
+      );
+
+      controller.init();
+
+      // Set selected group to be the group we're about to empty
+      const snapWalletId = toMultichainAccountWalletId(
+        MOCK_HD_KEYRING_2.metadata.id,
+      );
+      const snapGroupId = toMultichainAccountGroupId(snapWalletId, 1);
+      controller.setSelectedAccountGroup(snapGroupId);
```
ccharly: Not sure why we use the `snap*` prefix here?

fabiobozzo: Rogue copy-pasted from the other tests I took inspiration from!

ccharly (approved):

PR #6402, opened 2025-08-27
feat: BIP-44 support for EarnController

ameliejyc:
## Explanation

To remove our dependency on AccountsController and use the new BIP-44 enabled AccountTreeController, this PR contains two main changes:

- Swaps out the listener `AccountsController:selectedAccountChange` for `AccountTreeController:selectedAccountGroupChange`
- Swaps out the action `AccountsController:getSelectedAccount` for `AccountTreeController:getAccountsFromSelectedAccountGroup` from which we derive the EVM-compatible account

This relies on the latest 0.12.1 release of the account-tree-controller package.

Small things of note:
- The account listener no longer needs the account payload to update the account address as the race condition of concern [has been fixed](https://github.com/MetaMask/core/pull/5555) (this is relevant to mention still because despite using the new AccountTreeController event, it still uses AccountsController under the hood)
- In the new `getSelectedEvmAccount()` function I initially handled the case of no account being found (which shouldn't be possible) by failing early with a 'No EVM-compatible account address found' error. But on seeing how this case is handled by consuming functions (sometimes returning early, sometimes returning empty objects, sometimes throwing error) I opted to keep it consistent with the current code

## References

Fixes the core side of this issue: https://consensyssoftware.atlassian.net/browse/TAT-1315?atlOrigin=eyJpIjoiMmQ2MDY1ZjQ4MDU5NDdiYmJhMjRhYzNiMThhMjEwYzIiLCJwIjoiaiJ9

And should be tested alongside the accompanying mobile repo update: https://github.com/MetaMask/metamask-mobile/pull/19160

## Checklist


ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

ameliejyc:
@metamaskbot publish-preview

nickewansmith (approved):
LGTM on mobile side

packages/earn-controller/package.json
```diff
@@ -51,10 +51,12 @@
     "@ethersproject/providers": "^5.7.0",
     "@metamask/base-controller": "^8.2.0",
     "@metamask/controller-utils": "^11.12.0",
+    "@metamask/keyring-api": "^20.1.0",
     "@metamask/stake-sdk": "^3.2.1",
     "reselect": "^5.1.1"
   },
   "devDependencies": {
+    "@metamask/account-tree-controller": "^0.12.1",
     "@metamask/accounts-controller": "^33.0.0",
```
cryptodev-2s: Nit: should we remove AccountsController from dev, peer dependencies and tsconfig ?

ameliejyc: Yes, I meant to! Thanks

packages/earn-controller/package.json
```diff
@@ -68,7 +69,6 @@
     "typescript": "~5.2.2"
   },
   "peerDependencies": {
-    "@metamask/accounts-controller": "^33.0.0",
```
mcmire, line 71: Worth noting that we removed this peer dependency in the changelog?

packages/earn-controller/package.json
```diff
@@ -51,11 +51,12 @@
     "@ethersproject/providers": "^5.7.0",
     "@metamask/base-controller": "^8.2.0",
     "@metamask/controller-utils": "^11.12.0",
+    "@metamask/keyring-api": "^20.1.0",
```
mcmire, line 54: Worth noting that we added a new dependency in the changelog?

packages/earn-controller/package.json
cryptodev-2s, line 71: Nit: as now EarnController relies on AccountTreeController actions/events should we add it as a peer dependency?

packages/earn-controller/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Migrated from `AccountsController:getSelectedAccount` to `AccountTreeController:getAccountsFromSelectedAccountGroup` for retrieving BIP-44 selected account information ([#6402](https://github.com/MetaMask/core/pull/6402))
+- **BREAKING:** Migrated from `AccountsController:selectedAccountChange` to `AccountTreeController:selectedAccountGroupChange` for BIP-44 account change management ([#6402](https://github.com/MetaMask/core/pull/6402))
```
mcmire: As a reminder, changelogs are for noting not changes to implementation but rather changes in behavior.

What effect does using AccountTreeController instead of AccountsController have on how this controller will be used in the clients? Is the state different? Specifically, why are these changes breaking — what has to be changed in the clients after an upgrade?

packages/earn-controller/src/EarnController.ts
```diff
@@ -904,7 +914,11 @@ export class EarnController extends BaseController<
       typeof TransactionController.prototype.addTransaction
     >[1];
   }) {
-    const address = this.#getCurrentAccount()?.address;
+    const address = this.#getSelectedEvmAccountAddress();
+
+    if (!address) {
+      throw new Error('No EVM-compatible account address found');
```
mcmire, line 920: Worth noting in the changelog that `executeLendingDeposit` now throws errors in certain cases?

packages/earn-controller/src/EarnController.test.ts
```diff
@@ -2361,6 +2417,35 @@ describe('EarnController', () => {
         ).toHaveBeenCalledWith(mockAccount1Address);
         expect(result).toBe(mockAllowance);
       });
+
+      it('doesn`t call underlyingTokenAllowance if no account address found', async () => {
```
mcmire, line 2421: Worth noting this in the changelog, or was this pre-existing behavior?

packages/earn-controller/CHANGELOG.md
ameliejyc: The usage of the new `AccountTreeController` doesn't affect the client's usage other than that the messenger must update the allowed events and actions, which is my third point. So I guess if that one is already classified as breaking, then I think you're saying that these two wouldn't have to be? Or maybe given they don't change behaviour they don't have to be mentioned at all?

mcmire: Yes, if they don't change behavior and they are just implementation details then I think we can omit them.

ameliejyc: Ok thanks for the clarity

packages/earn-controller/package.json
```diff
@@ -68,7 +69,7 @@
     "typescript": "~5.2.2"
   },
   "peerDependencies": {
-    "@metamask/accounts-controller": "^33.0.0",
+    "@metamask/account-tree-controller": "^0.12.1",
```
cryptodev-2s, line 72: Nit: Could we require only up to the latest major
```suggestion
    "@metamask/account-tree-controller": "^0.12.0",

File: packages/earn-controller/CHANGELOG.md [hunk 8]
```diff
@@ -7,8 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- Added `@metamask/keyring-api` as a dependency ([#6402](https://github.com/MetaMask/core/pull/6402))
+- Added `@metamask/account-tree-controller` as a dev and peer dependency ([#6402](https://github.com/MetaMask/core/pull/6402))
+
 ### Changed
 
+- **BREAKING:** `EarnController` messenger must now allow `AccountTreeController:selectedAccountGroupChange` and `AccountTreeController:getAccountsFromSelectedAccountGroup` for BIP-44 compatibility and must not allow `AccountsController:selectedAccountChange` and `AccountsController:getSelectedAccount` ([#6402](https://github.com/MetaMask/core/pull/6402))
+- `executeLendingDeposit`, `executeLendingWithdraw` and `executeLendingTokenApprove` now throw errors if no selected address is found ([#6402](https://github.com/MetaMask/core/pull/6402))
+- `getLendingTokenAllowance`, `getLendingTokenMaxWithdraw` and `getLendingTokenMaxDeposit` now return `undefined` is no selected address is found ([#6402](https://github.com/MetaMask/core/pull/6402))
+- Removed `@metamask/accounts-controller` as a dev and peer dependency ([#6402](https://github.com/MetaMask/core/pull/6402))
```
cryptodev-2s: Nit: could we put this under Removed Section

packages/earn-controller/package.json
ameliejyc: Our changes rely on a fix made in 0.12.1 so I think we should keep it

packages/earn-controller/CHANGELOG.md
ameliejyc: I forgot this was an option, thanks!

PR #6410, opened 2025-08-28
feat: Add 'gasUsed' to transaction metadata

pedronfigueiredo:
## Explanation

The property is returned by the transaction simulation results.

## References

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5677

## Checklist

### DIFF
--- a/packages/transaction-controller/src/TransactionController.test.ts
+++ b/packages/transaction-controller/src/TransactionController.test.ts
@@ -484,7 +484,6 @@ const SIMULATION_DATA_RESULT_MOCK: SimulationData = {
       isDecrease: false,
     },
   ],
-  // gasUsed: '0x5208',
 };
 
 const GAS_FEE_TOKEN_MOCK: GasFeeToken = {
--- a/packages/transaction-controller/src/TransactionController.test.ts
+++ b/packages/transaction-controller/src/TransactionController.test.ts
@@ -516,6 +516,10 @@ const METHOD_DATA_MOCK: MethodData = {
 };
 
 describe('TransactionController', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
   const uuidModuleMock = jest.mocked(uuidModule);
   const EthQueryMock = jest.mocked(EthQuery);
   const updateGasMock = jest.mocked(updateGas);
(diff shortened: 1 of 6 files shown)


packages/transaction-controller/src/types.ts
```diff
@@ -239,6 +239,11 @@ export type TransactionMeta = {
    */
   gasLimitNoBuffer?: string;
 
+  /**
+   * The gas used by the transaction in the simulation.
```
matthewwalsh0:
```suggestion
   * The estimated gas used by the transaction, after any refunds. Generated from transaction simulation.
```

packages/transaction-controller/src/types.ts
```diff
@@ -1480,6 +1485,9 @@ export type SimulationData = {
 
   /** Data concerning a change to the user's token balances. */
   tokenBalanceChanges: SimulationTokenBalanceChange[];
+
+  /** The gas used by the transaction in the simulation. */
+  gasUsed?: Hex;
```
matthewwalsh0: Since we're storing this in a dedicated property, can we avoid conflating it with the simulation data which is specific to balance changes and so we don't persist it twice?

packages/transaction-controller/src/utils/balance-changes.ts
```diff
@@ -162,6 +164,7 @@ export async function getBalanceChanges(
 
     return {
       tokenBalanceChanges: [],
+      gasUsed: undefined,
```
matthewwalsh0: Can we return this as a separate property and put the others into `simulationData` so it's not confused as part of the `SimulationData`?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -4260,6 +4260,7 @@ export class TransactionController extends BaseController<
       (txMeta) => {
         txMeta.gasFeeTokens = gasFeeTokens;
         txMeta.isGasFeeSponsored = isGasFeeSponsored;
+        txMeta.gasUsed = simulationData.gasUsed;
```
matthewwalsh0: Can we default this above with a `let` as we do the other properties?

packages/transaction-controller/src/TransactionController.test.ts
```diff
@@ -988,6 +989,7 @@ describe('TransactionController', () => {
 
     signMock = jest.fn().mockImplementation(async (transaction) => transaction);
     isEIP7702GasFeeTokensEnabledMock = jest.fn().mockResolvedValue(false);
+    getBalanceChangesMock.mockResolvedValue(SIMULATION_DATA_RESULT_MOCK);
```
matthewwalsh0: Can we have a dedicated unit test to verify the new property is updated?

matthewwalsh0 (changes requested):

matthewwalsh0 (approved):

PR #6411, opened 2025-08-28
fix: token balance and account tracker fixes

bergarces:
## Checklist

### DIFF
--- a/packages/assets-controllers/src/AccountTrackerController.ts
+++ b/packages/assets-controllers/src/AccountTrackerController.ts
@@ -694,12 +694,17 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
 
       aggregated.forEach(({ success, value, account, token, chainId }) => {
         if (success && value !== undefined) {
+          const checksumAddress = toChecksumHexAddress(account);
           const hexValue = `0x${value.toString(16)}`;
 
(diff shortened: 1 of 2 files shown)


packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -361,7 +362,7 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
         const balanceUpdates = nativeBalances.map((balance) => ({
           address: balance.account,
           chainId: balance.chainId,
-          balance: balance.value?.toString() ?? '0x0',
+          balance: balance.value ? BNToHex(balance.value) : '0x0',
```
bergarces, line 365: We always want a Hex value here.

packages/assets-controllers/src/multicall.ts
bergarces, line 1: decodeFunctionResult type is any, but it's really a BigNumber, not a BN. It needs conversion.

packages/assets-controllers/src/AccountTrackerController.ts
bergarces, line 1: Ensure we always checksum the indexed address when it comes from external sources.

Prithpal-Sooriya (approved):

PR #6416, opened 2025-08-29
feat: check against urls with path

imblue-dabadee:
## Checklist


eslint-warning-thresholds.json
AugmentedMode, line 1: We shouldn't be modifying this file should we ?

packages/phishing-controller/src/utils.ts
```diff
@@ -262,6 +530,71 @@ export const matchPartsAgainstList = (source: string[], list: string[][]) => {
...
+ * @param url - the url to check.
+ * @param urlPaths - the paths to check.
+ * @returns true if the hostname and path exists in the path list, false otherwise.
+ */
+export const doesURLPathExist = (
+  url: string,
+  urlPaths: Record<string, Record<string, Record<string, string[]>>>,
+) => {
+  const urlWithProtocol = url.startsWith('http') ? url : `https://${url}`;
+  const { hostname, pathname } = new URL(urlWithProtocol);
```
AugmentedMode: `new URL()` throws on bad input. Should we wrap this with a try/catch?

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -132,6 +136,7 @@ export type PhishingStalelist = {
 export type PhishingListState = {
   allowlist: string[];
   blocklist: string[];
+  blocklistPaths: Record<string, Record<string, Record<string, string[]>>>;
```
AugmentedMode: What do you think about a type like this instead for better readability 

```
type PathBlockNode = {
  isBlocked: boolean;          
  blockAllSubpaths: boolean;    
  children: Record<string, PathBlockNode>;  
};

type BlocklistPaths = Record<string, PathBlockNode>;  

```

eslint-warning-thresholds.json
mcmire: Hey there! This file gets automatically updated when you run `yarn lint`, it's okay to modify and commit this. The goal of this file is to track how many lint warnings there are so we can slowly remove them (it's a tech debt kind of thing). Only thing I would say is that the numbers in this file should go down.

packages/phishing-controller/src/utils.ts
```diff
@@ -48,6 +49,24 @@ const splitStringByPeriod = <Start extends string, End extends string>(
   ];
 };
 
+export const getHostnameAndPathComponents = (
```
AugmentedMode, line 52: Do we need to decode path urls ? 

for example lots of path urls can look like this `%70%61`. Would decoding this so that we can do the check better solve this or is this a non issue?

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -139,6 +145,9 @@ export type PhishingStalelist = {
  * type defining the persisted list state. This is the persisted state that is updated frequently with `this.maybeUpdateState()`.
  * @property allowlist - List of approved origins (legacy naming "whitelist")
  * @property blocklist - List of unapproved origins (legacy naming "blacklist")
+ * @property blocklistPaths - List of unapproved origins with paths (hostname + path, no query params).
+ * The first key is hostname+first path segment. The second key is the second path segment.
+ * The value of the second key is an array of blocked third path segments. We only store up to three path segments deep.
```
AugmentedMode: In this doc string you mentioned we can only store up to three paths, however in this function `insertToTrie` I do not see a limit of three path segments. Is this intended?

packages/phishing-controller/src/PhishingDetector.ts
```diff
@@ -158,6 +160,22 @@ export class PhishingDetector {
...
     const source = domainToParts(fqdn);
 
+    for (const { blocklistPaths, name, version } of this.#configs) {
+      if (!blocklistPaths || Object.keys(blocklistPaths).length === 0) {
+        continue;
+      }
+      const pathMatch = isTerminalPath(url, blocklistPaths);
+      if (pathMatch) {
+        return {
+          match: domainPartsToDomain(source), // TODO: revisit this. do we want to return the path?
```
AugmentedMode: TODO - what should be done here ?

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -116,18 +119,21 @@ export type C2DomainBlocklistResponse = {
...
- * @property version - Stalelist data structure iteration.
+ * allowlist - List of approved origins.
+ * blocklist - List of unapproved origins (hostname-only entries).
+ * blocklistPaths - List of unapproved origins with paths (hostname + path entries).
+ * fuzzylist - List of fuzzy-matched unapproved origins.
+ * tolerance - Fuzzy match tolerance level
+ * lastUpdated - Timestamp of last update.
+ * version - Stalelist data structure iteration.
  */
 export type PhishingStalelist = {
```
AugmentedMode, line 137: Why are we making this change ?

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add path-based blocking [#6416](https://github.com/MetaMask/core/pull/6416)
```
AugmentedMode, line 12: Should we include the new tree structure and whitelistPaths in this changelog?

packages/phishing-controller/src/PathTrie.ts
```diff
@@ -0,0 +1,169 @@
...
+};
+
+/**
+ * Check if a URL is a terminal path i.e. the last path component is a terminal node.
+ *
+ * @param url - The URL to check.
+ * @param pathTrie - The trie to check the URL in.
+ * @returns True if the URL is a terminal path, false otherwise.
+ */
+export const isTerminalPath = (url: string, pathTrie: PathTrie): boolean => {
```
AugmentedMode: say `example.com/path` is added to the tree. 

if `example.com/path/hello` is being checked here will it return true? meaning it is blocked ?

packages/phishing-controller/src/PhishingController.ts
imblue-dabadee: Oop, good catch, this is outdated

packages/phishing-controller/src/PhishingController.ts
imblue-dabadee: Reduce lint warnings

packages/phishing-controller/CHANGELOG.md
imblue-dabadee: Ok, should add:
- `whitelistPaths` since that is going to be exported now in the `PhishingControllerState`
- `blocklistPaths` to `PhishingListState` (which is also included within `PhishingControllerState`.

AugmentedMode (approved):
LGTM thanks for fixing all comments!

PR #6420, opened 2025-08-29
feat: Allow disabling namespace checks

Gudahtt:
## Explanation

Messenger namespace checks can now be disabled by using the `DISABLE_NAMESPACE` namespace constructor parameter, and setting the `Namespace` type parameter to `string`. This makes it much easier to mock messengers in unit tests.

## References

N/A

## Checklist

### DIFF
--- a/packages/messenger/src/Messenger.ts
+++ b/packages/messenger/src/Messenger.ts
@@ -96,7 +96,7 @@ export type MessengerEvents<
 
 /**
  * Messenger namespace checks can be disabled by using this as the `namespace` constructor
- * parameter, and using `string` as the Namespace type parameter.
+ * parameter, and using `MockAnyNamespace` as the Namespace type parameter.
  *
  * This is useful for mocking a variety of different actions/events in unit tests. Please do not
  * use this in production code.


packages/messenger/src/Messenger.ts
```diff
@@ -278,7 +287,6 @@ export class Messenger<
   registerActionHandler<
     ActionType extends Action['type'] & NamespacedName<Namespace>,
   >(actionType: ActionType, handler: ActionHandler<Action, ActionType>) {
-    /* istanbul ignore if */ // Branch unreachable with valid types
```
Gudahtt, line 281: It's still technically true that this is unreachable with valid types, but, I figured I may as well add tests given I was making some functional changes in this same area.

packages/messenger/src/Messenger.ts
```diff
@@ -94,6 +94,15 @@ export type MessengerEvents<
...
     : never;
 
+/**
+ * Messenger namespace checks can be disabled by using this as the `namespace` constructor
+ * parameter, and using `string` as the Namespace type parameter.
+ *
+ * This is useful for mocking a variety of different actions/events in unit tests. Please do not
+ * use this in production code.
+ */
+export const DISABLE_NAMESPACE = 'DISABLE_NAMESPACE';
```
mcmire: Technically this name makes sense, but if I'm on another team I might feel like this is doing some magic I don't understand.

What do you think about using the name `ANY_NAMESPACE` or `ALL_NAMESPACES` instead?

packages/messenger/src/Messenger.test.ts
```diff
@@ -27,6 +27,24 @@ describe('Messenger', () => {
...
     });
 
+    it('allows registering and calling an action handler for a different namespace using DISABLE_NAMESPACE', () => {
+      type CountAction = {
+        type: 'Fixture:count';
+        handler: (increment: number) => void;
+      };
+      const messenger = new Messenger<string, CountAction, never>({
+        namespace: DISABLE_NAMESPACE,
+      });
```
mcmire: Continuing with another comment, what do you think about adding a type alias for `string`, `AnyNamespace`? Then combined with the renaming of the constant, this would look like:

``` suggestion
      const messenger = new Messenger<AnyNamespace, CountAction, never>({
        namespace: ANY_NAMESPACE,
      });
```

Alternatively:

``` suggestion
      const messenger = new Messenger<AllNamespaces, CountAction, never>({
        namespace: ALL_NAMESPACES,
      });
```

packages/messenger/src/Messenger.ts
mcmire: I guess we don't want to make this sound like something that could be used in production, so I can see why you gave it the name you did.

Maybe using `__`, i.e., `__ANY_NAMESPACE__` or `__ALL_NAMESPACES__`, would still inform people that it's a development-only variable while still conveying while still hiding the technical details?

Gudahtt: I was hesitant to use `__` because this _is_ something we want people to use. Just not in production code. I considered using the `@deprecated` directive as well, and decided against it for the same reason.

Gudahtt: I see what you mean about "disable namespace" being a bad name for this though. 

"All" and "any" are good options. We could also go with `MOCK_NAMESPACE`, if we want to make it more clear it's for tests only.

matthiasgeihs: I think the attention here might have been to choose a name that is a bit more descriptive of the thing that it does. In the sense of "this messenger can be used with any namespace / all namespaces".

matthiasgeihs: I like `ANY_NAMESPACE` (or `ALL_NAMESPACES`), because it conveys the underlying logic of what this does.
`DISABLE_NAMESPACE` does the same thing, but sounds more like an instruction instead of a namespace name.
`MOCK_NAMESPACE` focuses more on the intended usage, but doesn't clearly convey the effect of using it.

Gudahtt: I'm tempted to go with `MOCK_ANY_NAMESPACE` because it conveys both the purpose and intended usage

Gudahtt: Name updated here: 7f56a960356bcceabd3ec027c7ab44aaacc34c56

Does this seem better?

packages/messenger/src/Messenger.test.ts
Gudahtt: I've added a `MockAnyNamespace` type here: 6e633510c68ad69ea66f7e70120f6a7520e372e9

packages/messenger/src/Messenger.ts
mcmire: I think it looks good.

packages/messenger/src/Messenger.ts
```diff
@@ -94,6 +94,23 @@ export type MessengerEvents<
     ? Event
     : never;
 
+/**
+ * Messenger namespace checks can be disabled by using this as the `namespace` constructor
+ * parameter, and using `string` as the Namespace type parameter.
```
FrederikBolding:
```suggestion
 * parameter, and using `MockAnyNamespace` as the Namespace type parameter.
```

Gudahtt: Thanks, good catch

PR #5698, opened 2025-04-23
Feat: remove current chain id dependency asset contract controllers

salimtb:
## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

PR #6422, opened 2025-08-29
feat: migrate eip-5792 & capabilities middleware handlers into monorepo

ffmcgee725:
## Explanation

We identified duplication in JSON RPC middleware logic between the mobile and extension clients, particularly around the 5792 middleware stack and capability handling. During recent Wallet API and Perps integration discussions, the team agreed this presents an opportunity to abstract and centralize this logic into a core monorepo module.

The motivation here is to:

* Reduce code duplication across mobile and extension clients
* Create a client-agnostic foundation for capabilities (including auxiliary funds, capability advertising, and required assets)
* Support the upcoming Metamask Pay (MM Pay) initiative by establishing a standardized middleware layer
* Facilitate future multi-chain and auxiliary funds work without creating divergent patterns

So the work done on this PR aims to:

1. Create a new core package `eip-5792-middleware` to host shared middleware logic.
2. Extract the 5792 middleware logic and capability handlers from both extension and mobile clients.
   a. [Extension](https://github.com/MetaMask/metamask-extension/pull/35541)
   b. [Mobile](https://github.com/MetaMask/metamask-mobile/pull/19064)
4. Refactor them to be client-agnostic and consumable by both clients.
5. Maintain feature parity with existing implementations while improving modularity and testability.
6. Include initial unit tests and integration hooks.

## References

* Fixes [#5698](https://github.com/MetaMask/MetaMask-planning/issues/5698)

## Checklist


ffmcgee725:
@metamaskbot publish-preview

ffmcgee725:
@metamaskbot publish-preview

ffmcgee725:
@metamaskbot publish-preview

packages/eip-5792-middleware/src/index.ts
matthewwalsh0, line 1: This is still a really great and valuable abstraction, but am I correct this package isn't technically a middleware itself, but instead just the hooks required for the existing wallet middleware?

packages/eip-5792-middleware/src/processSendCalls.ts
```diff
@@ -0,0 +1,437 @@
...
+  EIP5792ErrorCode,
+  KEYRING_TYPES_SUPPORTING_7702,
+  MessageType,
+  VERSION,
+} from './constants';
+import type { EIP5792Messenger } from './types';
+import { getAccountKeyringType } from './utils';
+
+type ProcessSendCallsHooks = {
+  addTransactionBatch: TransactionController['addTransactionBatch'];
```
matthewwalsh0: The confirmations team has a backlog item that I can prioritise, to add messenger actions for these core `TransactionController` methods.

Which would (in future) result in simpler hooks as they would rely primarily on the messenger.

packages/eip-5792-middleware/src/processSendCalls.ts
```diff
@@ -0,0 +1,437 @@
...
+  ) => Promise<void>;
+};
+
+type ProcessSendCallsRequest = JsonRpcRequest & {
+  networkClientId: string;
+  origin?: string;
+};
+
+/**
+ * Processes a sendCalls request for EIP-5792 atomic transactions.
```
matthewwalsh0: Minor, not always `atomic`, that's defined per request.

packages/eip-5792-middleware/src/processSendCalls.ts
```diff
@@ -0,0 +1,437 @@
...
+ *
+ * @param hooks - Object containing required controller hooks and utilities.
```
matthewwalsh0: Minor, I believe our best practice is to define the JSDoc on the type itself (`ProcessSendCallsRequest`) which we could export?

packages/eip-5792-middleware/src/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+  SendTransaction = 'eth_sendTransaction',
+}
+
+export enum AtomicCapabilityStatus {
+  Supported = 'supported',
+  Ready = 'ready',
+  Unsupported = 'unsupported',
+}
+
+// To be moved to @metamask/rpc-errors in future.
```
matthewwalsh0, line 14: Thank you for including the comment, was going to recommend the same 😄 

These were added quickly originally due to time constraints.

packages/eip-5792-middleware/src/getCapabilities.ts
```diff
@@ -0,0 +1,191 @@
...
+  isRelaySupported: (chainId: Hex) => Promise<boolean>;
+  getSendBundleSupportedChains: (
+    chainIds: Hex[],
+  ) => Promise<Record<string, boolean>>;
+};
+
+/**
+ * Retrieves the capabilities for atomic transactions on specified chains.
+ *
+ * @param hooks - Object containing required controller hooks and utilities.
```
matthewwalsh0: Same here, should we export these hook types for the benefit of the clients, and move the JSDoc to them?

packages/eip-5792-middleware/src/getCapabilities.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @param chainIds - Array of chain IDs to check for alternate gas fees support.
+ * @param batchSupport - Information about atomic batch support for each chain.
+ * @param getIsSmartTransaction - Function to check if a chain supports smart transactions.
+ * @param isRelaySupported - Function to check if relay is supported on a chain.
+ * @param getSendBundleSupportedChains - Function to get chains that support send bundle.
+ * @param messenger - Messenger instance for controller communication.
+ * @returns Promise resolving to GetCapabilitiesResult with alternate gas fees information.
+ */
+async function getAlternateGasFeesCapability(
```
matthewwalsh0: I appreciate that this initial version is using as much of the original logic as possible, I think that's wise for minimum risk.

For audit sake, it would be great in future if we could create some abstraction around each capability so they can be modularised into separate files, tested directly, and combined with a generic mechanism.

packages/eip-5792-middleware/src/getCapabilities.ts
matthewwalsh0, line 1: Minor, is it worth moving these core functions into a directory such as `methods` or `hooks`?

packages/eip-5792-middleware/src/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+export const KEYRING_TYPES_SUPPORTING_7702 = [
+  KeyringTypes.hd,
+  KeyringTypes.simple,
+];
+
+export enum MessageType {
+  SendTransaction = 'eth_sendTransaction',
+}
+
+export enum AtomicCapabilityStatus {
```
matthewwalsh0: This is technically part of the core specification, but the definition was done in the client (now here) so the validation in `@metamask/eth-json-rpc-middleware` can be entirely generic.

packages/eip-5792-middleware/README.md
```diff
@@ -0,0 +1,15 @@
+# `@metamask/eip-5792-middleware`
+
+Implements the JSON-RPC methods for sending multiple calls from the user's wallet, and checking their status, as referenced in [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792).
```
matthewwalsh0: Minor, we're not technically implementing handlers for the RPC methods themselves, but the hooks required by the wallet middleware in `eth-json-rpc-middleware`.

packages/eip-5792-middleware/src/index.test.ts
matthewwalsh0, line 1: Minor, is this just testing the language / defining the same data in two places?

packages/eip-5792-middleware/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Retrieves the keyring type for a given account address.
+ *
+ * @param accountAddress - The account address to look up.
+ * @param messenger - Messenger instance for controller communication.
+ * @returns The keyring type associated with the account.
+ * @throws JsonRpcError if the account type is unknown or not found.
+ */
+export function getAccountKeyringType(
```
matthewwalsh0, line 16: Minor, is there value in writing unit tests for this so we can mock it in other test files?

jiexi (dismissed):

matthewwalsh0 (changes requested):

packages/eip-5792-middleware/src/index.ts
jiexi: yeah good catch. I was just looking at eth-json-rpc-middleware. It's odd to me personally that these packages are split this way. In server architecture, this would be the equivalent of grouping the controllers and processors into separate packages rather than grouping controllers and processors in the same domain together

ffmcgee725: Any suggestion on the name change, if any then ?
I do agree with the points being made, but I am following the decision / consensus reached on the [original thread discussion](https://consensys.slack.com/archives/C09CA95JU1G/p1756394965595529), as well as the patterns on existing packages.

packages/eip-5792-middleware/src/processSendCalls.ts
ffmcgee725: Nice! Do you want me to add this piece of information as documentation here, linking to the existing backlog item (share said item if so) ?

packages/eip-5792-middleware/src/index.ts
matthewwalsh0: I was mostly just confirming my understanding, but I think the name is fine as anything else might introduce more ambiguity.

To @jiexi's point, we could even move out the middleware itself from `eth-json-rpc-middleware` to here in future and make this a true middleware, and the `wallet-middleware` less of a monolith.

packages/eip-5792-middleware/src/processSendCalls.ts
matthewwalsh0: No need to commit it, just wanted to mention so others don't question why we have many callbacks in the hooks.

packages/eip-5792-middleware/src/constants.ts
ffmcgee725: If you have an existing issue that should be linked here, do feel free to share and I can update this.

packages/eip-5792-middleware/src/index.test.ts
ffmcgee725: Following the pattern of what exists on other packages, making sure only the expected functions are being exported, like [here](https://github.com/MetaMask/core/blob/main/packages/messenger/src/index.test.ts) or [here](https://github.com/MetaMask/core/blob/main/packages/controller-utils/src/index.test.ts).

packages/eip-5792-middleware/src/utils.ts
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/README.md
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/src/constants.ts
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/src/getCapabilities.ts
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/src/getCapabilities.ts
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/src/processSendCalls.ts
ffmcgee725: Originally I created these types to export and use them in [mobile creation of the 5792 middleware](https://github.com/MetaMask/metamask-mobile/blob/44178c949235788f18682e0fa067c8e81ecd83f9/app/core/RPCMethods/createEip5792Middleware.ts#L18) (this is from draft mobile PR branch of this refactor work), but noticed there is already an exported type coming from the eth-json-rpc-middleware package [here](https://github.com/MetaMask/eth-json-rpc-middleware/blob/45ed9985e5eca5766ae3d20aea71da9fb45e2b87/src/methods/wallet-send-calls.ts#L61), so decided to not export these at all for now since there is no need at the moment.

But perhaps exporting them for the future if necessary is the proper approach.

Also followed on JSDoc to the type itself here https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

packages/eip-5792-middleware/src/processSendCalls.ts
ffmcgee725: https://github.com/MetaMask/core/pull/6422/commits/e94ec4904ab988069fe1260d4568f14249686439

ffmcgee725:
@metamaskbot publish-preview

FrederikBolding (approved):
CODEOWNERS change LGTM. Have not reviewed the rest in detail.

matthewwalsh0 (approved):

PR #6423, opened 2025-08-29
Reword circuit break error message

mcmire:
## Explanation

The error that is thrown when an RPC endpoint responds with too many 5xx errors in a row — causing the underlying circuit to break — is too cryptic for end users. This commit rewords it to be more friendly and recommend that users switch to another endpoint. It also logs the original Cockatiel error so that users can inspect DevTools to learn more.

- Old error: "Execution prevented because the circuit breaker is open"
- New error: "RPC endpoint returned too many errors. Please try a different endpoint"

## Checklist


packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -566,6 +579,12 @@ export class RpcService implements AbstractRpcService {
         throw rpcErrors.parse({
           message: 'RPC endpoint did not return JSON.',
         });
+      } else if (error instanceof BrokenCircuitError) {
+        this.#logger?.info(error);
```
mcmire: This doesn't seem to show anything in the extension background DevTools. Not sure if we need to use `debug` or if logging just doesn't work like we expect.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -566,6 +579,12 @@ export class RpcService implements AbstractRpcService {
...
+        throw rpcErrors.resourceUnavailable({
+          message:
+            'RPC endpoint returned too many errors. Please try a different endpoint.',
```
Gudahtt: Maybe we could say something like "RPC endpoint returned too many errors, retrying in X minutes. Consider using a different RPC endpoint."

Switching to a different RPC endpoint would not be the correct solution if this was caused by a network issue (i.e. a proxy that blocked this traffic). This message keeps the "switch" recommendation a bit more open-ended, and sets expectations about timing in case the user does just need to wait (i.e. if they turned off the VPN responsible for switched WiFi networks or whatever)

mcmire: I've made this change here. The time remaining before retrying is updated dynamically: [fd6307e](https://github.com/MetaMask/core/pull/6423/commits/fd6307e66ca88dba27bb9774ffbb4125f38631b5)

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -566,6 +580,20 @@ export class RpcService implements AbstractRpcService {
...
+        const formattedRemainingCircuitOpenDuration = Intl.NumberFormat(
+          undefined,
+          { maximumFractionDigits: 2 },
+        ).format(
+          (remainingCircuitOpenDuration ?? this.#policy.circuitBreakDuration) /
+            Duration.Minute,
+        );
+        throw rpcErrors.resourceUnavailable({
+          message: `RPC endpoint returned too many errors, retrying in ${formattedRemainingCircuitOpenDuration} minutes. Consider using a different RPC endpoint.`,
+        });
```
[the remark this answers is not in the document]
mcmire: This should never happen; the next time a request is made, if the circuit break duration  has passed, Cockatiel will set the state of the circuit to half open, and then back to open if the endpoint continues to fail. Because the circuit breaker changes state, we will reset the `openedAt` property accordingly. What this means is that every time the code that Cursor is commenting on is run, `openedAt` should always be in the future.

mcmire:
This is now ready for review.

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: I wasn't passing the logger down correctly so that's why this didn't work. After fixing that I was able to successfully verify that this works. I also ended up changing this to `warn` to ensure that a stack trace is show in the logs.

PR #5065, opened 2024-12-12
Revert "chore: Remove `useRequestQueue` toggle to turn on/off Per Dapp Selected Network Feature (#4941)"

adonesky1:
This reverts commit 4814cf11670c76f7d3015206dbe3c78b383ddc3a.

## Explanation

We are not yet ready to release per-dapp selected network functionality on the mobile client and with this change there is no clean way to [update the version of SelectedNetworkController](https://github.com/MetaMask/metamask-mobile/issues/12434#issuecomment-2537358920) in the mobile client without having the Domains state starting to populate and possibly become corrupt since its not being consumed by/updated by the frontend in the expected way and may need to be migrated away when its time to actually start using the controller. 

Without this revert the @MetaMask/wallet-framework team is blocked from completing their goal to get both clients up to the latest versions of all controllers.

### `@metamask/queued-request-controller`

- **ADDED**: **BREAKING:** `createQueuedRequestMiddleware` now expects a
`useRequestQueue` property in its param options

### `@metamask/selected-network-controller`

- **ADDED**: **BREAKING:**  `SelectedNetworkController` constructor now expects
both a `useRequestQueuePreference` and a `onPreferencesStateChange` param

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've highlighted breaking changes using the "BREAKING" category above as appropriate
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

PR #4941, opened 2024-11-15
remove queuedRequestPreference

adonesky1:
### `@metamask/package-a`

- **<CATEGORY>**: Your change here
- **<CATEGORY>**: Your change here

### `@metamask/package-b`

- **<CATEGORY>**: Your change here
- **<CATEGORY>**: Your change here

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've highlighted breaking changes using the "BREAKING" category above as appropriate
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

PR #6430, opened 2025-09-02
chore: remove legacy queueing integration params from SelectedNetworkController

wenfix:
## Explanation

This PR is a re-do of https://github.com/MetaMask/core/pull/4941, which got reverted in https://github.com/MetaMask/core/pull/5065. 

It removes the togglability of per dapp selected network from the SelectedNetworkController. [Full explanation in the original PR here](https://github.com/MetaMask/core/pull/4941)

## References

Fixes https://github.com/MetaMask/MetaMask-planning/issues/5683

## Checklist

### DIFF
--- a/packages/controller-utils/src/util.ts
+++ b/packages/controller-utils/src/util.ts
@@ -7,7 +7,7 @@ import {
   isHexString,
   remove0x,
   getChecksumAddress,
-  isHexChecksumAddress,
+  isValidHexAddress as isValidHexAddressUtil,
 } from '@metamask/utils';
 import type { BigNumber } from 'bignumber.js';
 import BN from 'bn.js';
@@ -366,7 +366,7 @@ function isValidHexAddressUnmemoized(
   // We used to rely on `isValidAddress` from `@ethereumjs/util` which allows
   // for upper-case characters too. So we preserve this behavior and use our
   // faster and memoized validation function instead.
-  return isHexChecksumAddress(addressToCheck);
+  return isValidHexAddressUtil(addressToCheck);
 }
 
 /**
--- a/packages/selected-network-controller/CHANGELOG.md
+++ b/packages/selected-network-controller/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** The `SelectedNetworkController` constructor no longer expects a `useRequestPreference` boolean nor an `onPreferencesStateChange` listener. Removal of these parameters means that `domains` state will always be added for sites that are granted permissions. ([]())
+
 - Bump `@metamask/utils` from `^11.2.0` to `^11.4.2` ([#6054](https://github.com/MetaMask/core/pull/6054))
 - Bump `@metamask/base-controller` from `^8.0.1` to `^8.2.0` ([#6284](https://github.com/MetaMask/core/pull/6284), [#6355](https://github.com/MetaMask/core/pull/6355))
 
--- a/packages/selected-network-controller/src/SelectedNetworkController.ts
+++ b/packages/selected-network-controller/src/SelectedNetworkController.ts
@@ -96,10 +96,6 @@ export type SelectedNetworkControllerMessenger = RestrictedMessenger<
 export type SelectedNetworkControllerOptions = {
   state?: SelectedNetworkControllerState;
   messenger: SelectedNetworkControllerMessenger;
-  useRequestQueuePreference: boolean;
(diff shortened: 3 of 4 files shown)


packages/selected-network-controller/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** The `SelectedNetworkController` constructor no longer expects a `useRequestPreference` boolean nor an `onPreferencesStateChange` listener. Removal of these parameters means that `domains` state will always be added for sites that are granted permissions. ([]())
+
```
ffmcgee725:
```suggestion
- **BREAKING:** The `SelectedNetworkController` constructor no longer expects a `useRequestPreference` boolean nor an `onPreferencesStateChange` listener. Removal of these parameters means that `domains` state will always be added for sites that are granted permissions. ([#6430](https://github.com/MetaMask/core/pull/6430))
```

packages/controller-utils/src/util.ts
```diff
@@ -7,7 +7,7 @@ import {
   isHexString,
   remove0x,
   getChecksumAddress,
-  isHexChecksumAddress,
+  isValidHexAddress as isValidHexAddressUtil,
```
adonesky1: What is this change for?

wenfix: `isHexChecksumAddress` is no longer exported from the package, i've replaced w/ a similar function that has the same behavior under the hood.

adonesky1: 🤔 I guess what I'm trying to understand is why this change is happening now? We aren't bumping the version of `@metamask/utils`, and the change this PR is concerned with is in `SelectedNetworkController` not `controller-utils`, so I'm just trying to understand why this change is necessary?

wenfix: Think you're right. Might've done this change w/o updating dependencies which was causing a lint/build error because of that function. Reverted here: 2fc367917320ce3daba1012c33225ec716310325

packages/selected-network-controller/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** The `SelectedNetworkController` constructor no longer expects a `useRequestPreference` boolean nor an `onPreferencesStateChange` listener. Removal of these parameters means that `domains` state will always be added for sites that are granted permissions. ([#6430](https://github.com/MetaMask/core/pull/6430))
```
jiexi: I'm not sure this is actually breaking. If we bumped this package in extension, I don't think we'd have to make any changes to code and everything would just work as expected

wenfix: We'd get compilation errors as the parameters wouldn't be defined in the controller but yes, it wouldn't break javascript code. Fixed here: f576ca533d171937cdb8ba55510a3ddc6f470b38

packages/selected-network-controller/src/SelectedNetworkController.ts
```diff
@@ -395,10 +350,7 @@ export class SelectedNetworkController extends BaseController<
     let networkProxy = this.#domainProxyMap.get(domain);
     if (networkProxy === undefined) {
       let networkClient;
-      if (
-        this.#useRequestQueuePreference &&
-        this.#domainHasPermissions(domain)
-      ) {
+      if (this.#domainHasPermissions(domain)) {
```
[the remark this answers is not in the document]
adonesky1: @wenfix this isn't really correct any more since Snaps do get their own selected networks, but this seems to be coming from the fact that [there are still tests in place here](https://github.com/MetaMask/core/blob/6f5d545c5f8ed6926efba2d0bcd6bf5b34a94638/packages/selected-network-controller/tests/SelectedNetworkController.test.ts#L665) which imply that snaps should only ever get the globally selected network...

We should audit these tests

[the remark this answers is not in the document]
adonesky1: actually re-reading that test I don't think its making an incorrect assertion. I'm not clear on what cursor thinks is the issue here. Removing `useRequestQueuePreference` shouldn't impact snaps from what I can tell.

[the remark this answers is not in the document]
adonesky1: I've validated on [the extension PR](https://github.com/MetaMask/metamask-extension/pull/35626/files) that Snaps permissions are still working as expected

PR #6439, opened 2025-09-03
feat(keyring-controller): add new `KeyringController:addNewKeyring` action

ccharly:
## Explanation

Adding new `:addNewKeyring` action so we can use it in the `MultichainAccountService` when creating/importing new wallets.

This action could also be useful in other context, since it apply to all keyring types we already support and not only HD keyrings.

## References

N/A

## Checklist


packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -4151,6 +4152,43 @@ describe('KeyringController', () => {
...
+          name: 'mock-keyring',
+        };
+        jest
+          .spyOn(KeyringController.prototype, 'addNewKeyring')
+          .mockImplementationOnce(async () => mockKeyringMetadata);
+
+        await withController(
+          { keyringBuilders: [keyringBuilderFactory(MockKeyring)] },
+          async ({ controller, messenger }) => {
+            const mockKeyringType = KeyringTypes.hd;
```
mikesposito: nit: as we are have `MockKeyring` and we are also adding it to builders, perhaps we can use `MockKeyring.type` instead of `KeyringTypes.hd` for this test?

mikesposito (approved):
Looks good, just a small nitpicker comment

PR #6441, opened 2025-09-03
feat: add multichain account group events

fabiobozzo:
## Explanation

Add multichain account group lifecycle events to support the Back and Sync feature (state 2). Events are emitted at the group level for precise, targeted event emission without performance overhead.

**Current state**: MultichainAccountService creates and manages multichain account groups but doesn't notify other controllers when these operations occur.

**Solution**: Implemented two new events following the established messenger pattern, with selective emission to prevent performance degradation:
- `multichainAccountGroupCreated` - Emitted from wallet when new groups are created via `createNextMultichainAccountGroup()` or `createMultichainAccountGroup()`  
- `multichainAccountGroupUpdated` - Emitted from group level when specific groups are synchronized, ensuring only actually updated groups trigger events

**Architecture**: 
- Events are emitted at the appropriate granular level (group sync → group events, group creation → wallet events)
- Initialization syncs don't emit events to prevent noise during startup
- Selective emission prevents performance degradation when only specific groups/accounts are updated

## References

[MUL-520](https://consensyssoftware.atlassian.net/browse/MUL-520)

## Checklist

[MUL-520]: https://consensyssoftware.atlassian.net/browse/MUL-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -215,6 +224,10 @@ export class MultichainAccountService {
     if (group) {
       if (sync) {
         group.sync();
+        this.#messenger.publish(
+          'MultichainAccountService:multichainAccountGroupUpdated',
+          group,
+        );
```
ccharly: I think all those events (like every new ones I mean) have to be moved at wallet-level no?

This way, everytime we use the wallet, this would trigger the appropriate event. WDYT?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -100,6 +106,14 @@ export class MultichainAccountWallet<
     ] of this.#accountGroups.entries()) {
       multichainAccount.sync();
 
+      // Emit update event for synced groups
+      this.#messenger.publish(
+        'MultichainAccountService:multichainAccountGroupUpdated',
+        multichainAccount as unknown as MultichainAccountGroup<
```
fabiobozzo: @ccharly qq: I don't really like this double cast, albeit safe in this context.  
Should I change the event's payload type to `MultichainAccountGroup<Account> ` instead of `MultichainAccountGroup<Bip44Account<KeyringAccount>>` ?

packages/multichain-account-service/src/MultichainAccountService.ts
fabiobozzo: It makes sense to me, thanks 😃 
so... any wallet operations (not just those triggered through the service) would emit the new events. got it

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Not sure why you even need to cast at all? 🤔

fabiobozzo: My bad, just fixed types:
```
import type { MultichainAccountGroup } from './MultichainAccountGroup';
```
!=
```
import type { MultichainAccountGroup } from '@metamask/account-api';
```

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add multichain account group lifecycle events ([#6441](https://github.com/MetaMask/core/pull/6441))
+  - Add `multichainAccountGroupCreated` event emitted from wallet level when new groups are created
+  - Add `multichainAccountGroupUpdated` event emitted from wallet level when groups are synchronized
```
```suggestion
  - Add `multichainAccountGroupCreated` event emitted from wallet level when new groups are created.
  - Add `multichainAccountGroupUpdated` event emitted from wallet level when groups are synchronized.
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -288,12 +298,18 @@ export class MultichainAccountWallet<
...
       });
     }
 
     // Register the account to our internal map.
     this.#accountGroups.set(groupIndex, group); // `group` cannot be undefined here.
 
+    this.#messenger.publish(
+      'MultichainAccountService:multichainAccountGroupCreated',
+      group,
+    );
```
ccharly: Should be guarded if `emitEvents` no?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -288,25 +300,37 @@ export class MultichainAccountWallet<
...
+   * @param emitEvents - Whether to emit creation event (defaults to true).
    * @throws If any of the account providers fails to create their accounts.
    * @returns The multichain account group for the next group index available.
    */
-  async createNextMultichainAccountGroup(): Promise<
-    MultichainAccountGroup<Account>
-  > {
-    return this.createMultichainAccountGroup(this.getNextGroupIndex());
+  async createNextMultichainAccountGroup(
+    emitEvents = true,
```
ccharly: Not sure this should be exposed to the caller. We do not emit events during the initial sync (but that's internal cooking and that's fine), but we should let the caller decide to emit events or not IMO.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -208,11 +218,13 @@ export class MultichainAccountWallet<
    * Creates a multichain account group for a given group index.
    *
    * @param groupIndex - The group index to use.
+   * @param emitEvents - Whether to emit creation event (defaults to true).
    * @throws If any of the account providers fails to create their accounts.
    * @returns The multichain account group for this group index.
    */
   async createMultichainAccountGroup(
     groupIndex: number,
+    emitEvents = true,
```
ccharly: Same here.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -35,31 +36,39 @@ export class MultichainAccountWallet<
...
    * Force wallet synchronization.
    *
    * This can be used if account providers got new accounts that the wallet
    * doesn't know about.
+   *
+   * @param options - Sync options.
+   * @param options.emitEvents - Whether to emit update events. Defaults to true.
    */
-  sync(): void {
+  sync({ emitEvents = true }: { emitEvents?: boolean } = {}): void {
```
ccharly: WDYT about having a `#sync` with this `emitEvents` as parameter, so the "public `sync`" does not expose this flag to the caller?

And all existing `sync` calls would become `#sync` instead.

gantunesr: ➕

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo: Ok so: no more `emitEvents` param in the public method, and a new private method `#createMultichainAccountGroup`

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo: Ok so, for the wallet sync, I added a `#sync` as you suggested. For cross-class communication, I added a `_syncWithEvents` method to the group and marked it @internal. Wdyt?

ccharly: So.. Here's what I think, I wonder if we really need that flag, at all. Maybe we could just get away with a `#initialized` flag and we only fire if it's `true`?

Ultimately, we could also have a `group.status` and `wallet.status`, it would be `Uninitialized` at first, and then would become `Ready` (or another status, like `AlignmentInProgress`, etc..). But IMO, we can tackle that status later.

Simple flag could work for now. WDYT @fabiobozzo?

fabiobozzo: Good point, more readable than my initial approach, at least 💁‍♂️

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -213,6 +228,12 @@ export class MultichainAccountWallet<
    */
   async createMultichainAccountGroup(
     groupIndex: number,
+  ): Promise<MultichainAccountGroup<Account>> {
+    return this.#createMultichainAccountGroup(groupIndex);
+  }
+
+  async #createMultichainAccountGroup(
+    groupIndex: number,
```
ccharly: I think we can remove the use of the private variante now?

```suggestion
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -306,7 +335,7 @@ export class MultichainAccountWallet<
   async createNextMultichainAccountGroup(): Promise<
     MultichainAccountGroup<Account>
   > {
-    return this.createMultichainAccountGroup(this.getNextGroupIndex());
+    return this.#createMultichainAccountGroup(this.getNextGroupIndex());
```
ccharly: Same here.

```suggestion
    return this.createMultichainAccountGroup(this.getNextGroupIndex());
```

packages/multichain-account-service/src/MultichainAccountGroup.ts
```diff
@@ -62,6 +72,10 @@ export class MultichainAccountGroup<
    * account doesn't know about.
    */
   sync(): void {
+    this.#sync();
+  }
+
+  #sync(): void {
```
ccharly: Actually same goes for this one now 😅 
```suggestion
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -80,6 +90,10 @@ export class MultichainAccountWallet<
    * doesn't know about.
    */
   sync(): void {
+    this.#sync();
+  }
+
+  #sync(): void {
```
ccharly: Here too:

```suggestion
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -55,22 +56,31 @@ export class MultichainAccountWallet<
...
     this.#id = toMultichainAccountWalletId(entropySource);
     this.#providers = providers;
     this.#entropySource = entropySource;
+    this.#messenger = messenger;
     this.#accountGroups = new Map();
 
-    // Initial synchronization.
-    this.sync();
+    // Initial synchronization (don't emit events during initialization).
+    this.#sync();
```
ccharly:
```suggestion
    this.sync();
```

packages/multichain-account-service/src/MultichainAccountGroup.ts
```diff
@@ -36,23 +37,32 @@ export class MultichainAccountGroup<
...
     this.#id = toMultichainAccountGroupId(wallet.id, groupIndex);
     this.#groupIndex = groupIndex;
     this.#wallet = wallet;
     this.#providers = providers;
+    this.#messenger = messenger;
     this.#providerToAccounts = new Map();
     this.#accountToProvider = new Map();
 
-    this.sync();
+    this.#sync();
```
ccharly:
```suggestion
    this.sync();
```

ccharly (approved):
LGTM!

One small note for later, since we're only firing those events in our own methods, if for some reason we discover new groups during a `sync`, we won't be sending the "created event" for those.

But that's fine, because we don't want to track this kind of account creation for BIP-44/multichain accounts.

PR #6449, opened 2025-09-03
`@metamask/eip-5792-middleware` init version to 0.0.0

adonesky1:
`@metamask/eip-5792-middleware` was incorrectly initialized with a version of `1.0.0`. We need to initialize packages with a version of 0.0.0 before publication.


mcmire (approved):
LGTM!

PR #6454, opened 2025-09-03
SWAPS-2839 update bridge controllers for bitcoin

ghgoodreau:
## Explanation

This PR extends the bridge controller to support Bitcoin transactions, building on the existing Solana support to create a more generic non-EVM chain handling system.

  Current state: The bridge controller currently only supports EVM chains and Solana for cross-chain transactions.

  Solution: This PR:
  - Renames Solana-specific functions and types to be generic for all non-EVM chains (NonEvmFees
  instead of SolanaFees, handleNonEvmTx instead of handleSolanaTx)
  - Adds support for Bitcoin transactions using PSBT (Partially Signed Bitcoin Transaction) format
  - Updates the Snap interface to use the new unified computeFee and signAndSendTransaction methods
  that work across all non-EVM chains
  - Maintains backward compatibility while deprecating Solana-specific naming

  Key changes:
  - The nonEvmFeesInNative field now stores fees in the smallest units for each chain
  - Transaction handling now detects Bitcoin PSBT format alongside string trade data
  
## Checklist


packages/bridge-controller/src/utils/caip-formatters.ts
```diff
@@ -147,6 +154,9 @@ export const formatAddressToAssetId = (
   if (chainId === SolScope.Mainnet) {
     return CaipAssetTypeStruct.create(`${chainId}/token:${addressOrAssetId}`);
   }
+  if (chainId === 'tron:0x2b6653dc') {
```
aganglada: `TrxScope` should be available by now from the `keyring-api`

packages/bridge-controller/src/utils/snaps.ts
```diff
@@ -22,19 +23,70 @@ export const getMinimumBalanceForRentExemptionRequest = (snapId: string) => {
...
   return {
     snapId: snapId as never,
     origin: 'metamask',
-    handler: 'onRpcRequest' as never,
+    handler: 'onClientRequest' as never,
     request: {
-      method: 'getFeeForTransaction',
+      id: uuid(),
+      jsonrpc: '2.0',
+      method: 'ClientRequest:computeFee',
```
aganglada:
```suggestion
      method: 'computeFee',
```

aganglada: The `ClientRequest:` on the spec is only to outline it's an `onClientRequest`

packages/bridge-controller/src/utils/snaps.ts
```diff
@@ -22,19 +23,70 @@ export const getMinimumBalanceForRentExemptionRequest = (snapId: string) => {
...
+  options?: Record<string, unknown>,
+) => {
+  return {
+    snapId: snapId as never,
+    origin: 'metamask',
+    handler: 'onClientRequest' as never,
+    request: {
+      id: uuid(),
+      jsonrpc: '2.0',
+      method: 'ClientRequest:signAndSendTransaction',
```
aganglada: same here, only `signAndSendTransaction`

packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -588,6 +589,26 @@ describe('BridgeController', function () {
                   resolve('5000');
                 }, 200);
               }
+              if (
+                (params as { handler: string })?.handler ===
+                  'onClientRequest' &&
+                (params as { request?: { method: string } })?.request
+                  ?.method === 'ClientRequest:computeFee'
```
aganglada:
```suggestion
                  ?.method === 'computeFee'
```

packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -1678,6 +1699,28 @@ describe('BridgeController', function () {
                   resolve(expectedMinBalance);
                 }, 200);
               }
+              if (
+                (params as { handler: string })?.handler ===
+                  'onClientRequest' &&
+                (params as { request?: { method: string } })?.request
+                  ?.method === 'ClientRequest:computeFee'
```
aganglada:
```suggestion
                  ?.method === 'computeFee'
```

packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -1780,6 +1823,121 @@ describe('BridgeController', function () {
...
+            },
+          } as never;
+        }
+
+        if (actionType === 'SnapController:handleRequest') {
+          return new Promise((resolve) => {
+            if (
+              (params as { handler: string })?.handler === 'onClientRequest' &&
+              (params as { request?: { method: string } })?.request?.method ===
+                'ClientRequest:computeFee'
```
aganglada:
```suggestion
                'computeFee'
```

aganglada (commented):
Sorry for the confusion on the spec. The `ClientRequest:` bit was only to outline it it's an `onClientRequest`

packages/bridge-controller/src/utils/snaps.ts
```diff
@@ -22,19 +23,70 @@ export const getMinimumBalanceForRentExemptionRequest = (snapId: string) => {
...
+ * Works for Solana, BTC, Tron, and other non-EVM networks
+ *
+ * @param snapId - The snap ID to send the request to
+ * @param transaction - The base64 encoded transaction string
+ * @param scope - The CAIP-2 chain scope
+ * @param accountId - The account ID
+ * @param options - Optional network-specific options
+ * @returns The snap request object
+ */
+export const signAndSendTransactionRequest = (
```
micaelae: I'd move this util to bridge-status-controller since it's used there

packages/bridge-controller/src/utils/caip-formatters.ts
```diff
@@ -147,6 +154,9 @@ export const formatAddressToAssetId = (
...
+    return CaipAssetTypeStruct.create(`${chainId}/token:${addressOrAssetId}`);
```
micaelae: Is there a constant we can use for this?

packages/bridge-controller/src/utils/bridge.ts
```diff
@@ -190,6 +190,23 @@ export const isBitcoinChainId = (
   return chainId.toString() === ChainId.BTC.toString();
 };
 
+export const isTronChainId = (chainId: Hex | number | CaipChainId | string) => {
+  if (isCaipChainId(chainId)) {
+    return chainId === 'tron:0x2b6653dc'; // Tron mainnet CAIP-2 identifier
+  }
```
micaelae: constant can also be used here

packages/bridge-controller/src/utils/bridge.ts
```diff
@@ -190,6 +190,23 @@ export const isBitcoinChainId = (
...
+  return chainId.toString() === ChainId.TRON.toString();
+};
+
+export const isNonEvmChainId = (
+  chainId: GenericQuoteRequest['srcChainId'],
```
micaelae, line 201: Seems like this would be useful in clients as well. Can you add JSDOC to indicate that this only checks networks that are supported by swaps?

packages/bridge-status-controller/src/utils/transaction.ts
```diff
@@ -10,7 +11,6 @@ import {
   type QuoteResponse,
 } from '@metamask/bridge-controller';
 import { toHex } from '@metamask/controller-utils';
-import { SolScope } from '@metamask/keyring-api';
 import type {
```
micaelae, line 14: nice! I believe you can remove keyring-api from the package.json as well

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Export `isNonEvmChainId` utility function to check for non-EVM chains (Solana, Bitcoin, Tron) ([#6454](https://github.com/MetaMask/core/pull/6454))
+
 ### Changed
 
+- Update Snap methods to use new unified interface for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `getFeeForTransaction` with `computeFee` method (handled via onClientRequest)
+  - Update fee format to return native units (e.g., SOL) instead of smallest units (e.g., Lamports)
```
micaelae: Can you mark this as a breaking change?

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Add support for Tron chain alongside existing Bitcoin and Solana support
+  - Export `signAndSendTransactionRequest` utility function
+  - Fix precision issues in Solana fee conversion using BigInt math
```
micaelae: This can be removed too

packages/bridge-controller/CHANGELOG.md
micaelae: Actually wait - is it a breaking change?

packages/bridge-controller/src/bridge-controller.ts
```diff
@@ -727,51 +728,88 @@ export class BridgeController extends StaticIntervalPollingController<BridgePoll
...
+          let feeInNative = '0';
+
+          if (baseFee?.asset?.amount) {
+            if (isSolanaChainId(quote.srcChainId)) {
+              // Convert SOL to Lamports (1 SOL = 10^9 Lamports)
+              // Use string manipulation to avoid floating point precision issues
+              const parts = baseFee.asset.amount.split('.');
+              const wholePart = parts[0] || '0';
+              const decimalPart = (parts[1] || '').padEnd(9, '0').slice(0, 9);
+              feeInNative = BigInt(wholePart + decimalPart).toString();
```
micaelae: We typically use the `bignumber.js` lib to do this math (already included in this package). Was there a reason to use BigInt instead?

packages/bridge-controller/CHANGELOG.md
ghgoodreau: @micaelae it actually doesn't require client to update!

packages/bridge-controller/src/bridge-controller.ts
```diff
@@ -727,51 +728,88 @@ export class BridgeController extends StaticIntervalPollingController<BridgePoll
...
+            } else {
+              // For other chains (BTC, Tron), use the fee as-is
+              feeInNative = baseFee.asset.amount;
```
micaelae: I think we should make this generic across chains so the metadata calculation can work for any non-EVM chain. If you search the package for `solanaFeesInLamports` you can see how it's used. We could deprecate that and add a new metadata field for `feeInNative`

Looks like network fees for Tron and BTC are not getting included in the metadata (see `getBridgeQuotes` selector)

packages/bridge-controller/src/selectors.ts
```diff
@@ -286,10 +287,12 @@ const selectBridgeQuotesWithMetadata = createBridgeSelector(
         relayerFee,
         gasFee: QuoteMetadata['gasFee'];
 
-      if (isSolanaChainId(quote.quote.srcChainId)) {
-        totalEstimatedNetworkFee = calcSolanaTotalNetworkFee(
+      if (isNonEvmChainId(quote.quote.srcChainId)) {
+        // Use the new generic function for all non-EVM chains
+        totalEstimatedNetworkFee = calcNonEvmTotalNetworkFee(
           quote,
```
micaelae, line 295: Can you also update `getExchangeRateByChainIdAndAddress` to use isNonEvmChainId? BTC and tron exchange rates will be read from the same state keys as solana

packages/bridge-controller/src/utils/quote.ts
```diff
@@ -105,6 +117,30 @@ export const calcSolanaTotalNetworkFee = (
   };
 };
 
+export const calcNonEvmTotalNetworkFee = (
+  bridgeQuote: QuoteResponse & NonEvmFees,
```
micaelae, line 98: Minor: delete `calcSolanaTotalNetworkFee` and repurpose its tests for the new evm fee calculation

packages/bridge-controller/src/utils/quote.ts
```diff
@@ -105,6 +117,30 @@ export const calcSolanaTotalNetworkFee = (
...
+  { exchangeRate, usdExchangeRate }: ExchangeRate,
+  chainId: string | number,
+) => {
+  const { nonEvmFeesInNative } = bridgeQuote;
```
micaelae, line 101: For 1 SOL, is `nonEvmFeesInNative` equal to `1000000000` or `1`?

ghgoodreau: For 1 SOL, nonEvmFeesInNative equals 1000000000, and the calcNonEvmTotalNetworkFee function converts this to 1 SOL by dividing by 10^9

packages/bridge-controller/src/types.ts
```diff
@@ -78,7 +78,11 @@ export type L1GasFees = {
 };
 
 export type SolanaFees = {
-  solanaFeesInLamports?: string; // solana fees in lamports, appended by BridgeController.#appendSolanaFees
+  solanaFeesInLamports?: string; // @deprecated Use nonEvmFeesInNative instead
+};
```
micaelae: I think we should remove this completely and just mark it as a breaking change. afaik it's not used outside the bridge-controller tho

packages/bridge-controller/src/types.ts
```diff
@@ -77,8 +77,8 @@ export type L1GasFees = {
   l1GasFeesInHexWei?: string; // l1 fees for approval and trade in hex wei, appended by BridgeController.#appendL1GasFees
 };
 
-export type SolanaFees = {
-  solanaFeesInLamports?: string; // solana fees in lamports, appended by BridgeController.#appendSolanaFees
+export type NonEvmFees = {
+  nonEvmFeesInNative?: string; // Non-EVM chain fees in smallest units (lamports for Solana, satoshis for BTC, sun for Tron)
```
infiniteflower: Personally I've been calling them `atomic` units

packages/bridge-controller/src/constants/bridge.ts
```diff
@@ -5,6 +5,9 @@ import type { Hex } from '@metamask/utils';
 import { CHAIN_IDS } from './chains';
 import type { BridgeControllerState } from '../types';
 
+// Tron mainnet CAIP-2 identifier
+export const TRON_MAINNET_SCOPE = 'tron:0x2b6653dc' as const;
```
infiniteflower: Do we need this for a BTC related PR?

packages/bridge-controller/src/utils/bridge.ts
```diff
@@ -190,6 +191,30 @@ export const isBitcoinChainId = (
...
+};
+
+/**
+ * Checks if a chain ID represents a non-EVM blockchain supported by swaps
+ * Currently supports Solana, Bitcoin, and Tron
+ *
+ * @param chainId - The chain ID to check
+ * @returns True if the chain is a supported non-EVM chain, false otherwise
+ */
+export const isNonEvmChainId = (
```
infiniteflower, line 200: I wonder if there's an easier way, otherwise we need to keep updating this for every single new non EVM network

How about:

1. Convert chainId to CAIP
2. Check CAIP namespace for `eip155`
3. If `eip155`, then it's EVM

What do you think?

packages/bridge-controller/src/utils/quote.ts
```diff
@@ -31,12 +31,18 @@ export const isValidQuoteRequest = (
   if (requireAmount) {
     stringFields.push('srcTokenAmount');
   }
-  // If bridging and one of the chains is solana, require the dest wallet address
+  // If bridging between different chain types or different non-EVM chains, require dest wallet address
+  // Cases that need destWalletAddress:
+  // 1. EVM -> non-EVM
+  // 2. non-EVM -> EVM
+  // 3. non-EVM -> different non-EVM (e.g., SOL -> BTC)
+  // Only same-chain swaps don't need destWalletAddress
```
infiniteflower, line 39: Thank you for the detailed comments 🙇

packages/bridge-status-controller/src/utils/snaps.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Works for Solana, BTC, Tron, and other non-EVM networks
+ *
+ * @param snapId - The snap ID to send the request to
+ * @param transaction - The base64 encoded transaction string
+ * @param scope - The CAIP-2 chain scope
+ * @param accountId - The account ID
+ * @param options - Optional network-specific options
+ * @returns The snap request object
+ */
+export const signAndSendTransactionRequest = (
```
infiniteflower: Can we call this something else? Because it doesn't actually sign and send anything, looks like it just returns the request object

packages/bridge-controller/src/utils/bridge.ts
ghgoodreau: I like that idea! It does add some testing scope whereas the current functionality is already tested and working. Is this blocking? If not, I'd prefer to get these changes released to unblock the client PR, and then can update the function next week. Let me know what you think and I can add a // TODO comment @infiniteflower

packages/bridge-controller/src/constants/bridge.ts
ghgoodreau: I've removed these!

packages/bridge-controller/src/utils/quote.ts
```diff
@@ -88,20 +94,31 @@ const calcTokenAmount = (value: string | BigNumber, decimals: number) => {
...
-  const { solanaFeesInLamports } = bridgeQuote;
-  const solanaFeeInNative = calcTokenAmount(solanaFeesInLamports ?? '0', 9);
+  const { nonEvmFeesInNative } = bridgeQuote;
+
+  let feeInNative: BigNumber;
+
+  if (isSolanaChainId(chainId)) {
+    // Solana fees are stored in lamports (smallest units), need to convert to SOL
+    const decimals = 9;
+    feeInNative = calcTokenAmount(nonEvmFeesInNative ?? '0', decimals);
```
micaelae: Why do we convert the SOL amount to lamports in bridge-controller.ts if we need to convert it back to SOL here?

packages/bridge-controller/src/utils/bridge.ts
infiniteflower: Yea this is non-blocking, fine with adding a TODO and addressing it in a follow up PR

micaelae (approved):

packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -17,13 +17,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Update `getClientRequest` to create proper requests for all non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Use `formatChainIdToCaip` to get proper scope for each chain
+  - Extract transaction data from either string or PSBT object format
+- Remove dependency on `@metamask/keyring-api` ([#6454](https://github.com/MetaMask/core/pull/6454))
 - Bump `@metamask/controller-utils` from `^11.12.0` to `^11.14.0` ([#6620](https://github.com/MetaMask/core/pull/6620), [#6629](https://github.com/MetaMask/core/pull/6629))
 - Bump `@metamask/base-controller` from `^8.3.0` to `^8.4.0` ([#6632](https://github.com/MetaMask/core/pull/6632))
 
+### Removed
+
+- Remove direct dependency on `@metamask/keyring-api` - no longer needed with unified Snap interface ([#6454](https://github.com/MetaMask/core/pull/6454))
```
Gudahtt, line 53: The dependency was not removed here, this is unaccurate

Gudahtt: Welp, I see it was actually listed twice. It's also in the changed section.

packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -17,13 +17,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Add new controller metadata properties to `BridgeStatusController` ([#6589](https://github.com/MetaMask/core/pull/6589))
 
 ### Changed
 
+- Update transaction submission to use new unified Snap interface for all non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `signAndSendTransactionWithoutConfirmation` with `ClientRequest:signAndSendTransaction` method
+  - Update response handling to support new `transactionId` format from unified interface
+  - Support multiple response formats: string, `{ transactionId }`, `{ result: { signature } }`, and `{ signature }`
+  - Maintain backward compatibility with legacy response formats
+- Rename transaction handling functions for clarity ([#6454](https://github.com/MetaMask/core/pull/6454))
```
Gudahtt, line 34: These changelogs are meant for users of the package, not internal contributors. Please don't document internal-only changes such as internal function renames. It makes it much harder to tell what has actually changed for anyone using this package.

packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -17,13 +17,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Handle Bitcoin PSBT (Partially Signed Bitcoin Transaction) format in trade data
+  - Support Bitcoin transaction submission through unified Snap interface
+  - Add Bitcoin-specific transaction handling in `#handleNonEvmTx` method
+  - Support extraction of `unsignedPsbtBase64` from trade data for Bitcoin transactions
 - Add new controller metadata properties to `BridgeStatusController` ([#6589](https://github.com/MetaMask/core/pull/6589))
 
 ### Changed
 
+- Update transaction submission to use new unified Snap interface for all non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `signAndSendTransactionWithoutConfirmation` with `ClientRequest:signAndSendTransaction` method
```
Gudahtt, line 30: This is a breaking change.

I'm told in practice it probably doesn't require client changes because our snaps support `ClientRequest:signAndSendTransaction` already. But we should still be documenting it properly here so that it's clearly understood that there is a compatibility change here.

packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -17,13 +17,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Rename `handleSolanaTxResponse` to `handleNonEvmTxResponse` to reflect support for all non-EVM chains
+  - Rename `#handleSolanaTx` to `#handleNonEvmTx` in BridgeStatusController
+  - Export `handleSolanaTxResponse` as an alias for backward compatibility (deprecated)
+- Update transaction detection logic to identify non-EVM transactions ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Check for Bitcoin PSBT format (`unsignedPsbtBase64` in trade object) alongside string trade data
+  - Use `isNonEvmChainId` for determining non-EVM transaction handling
+- Update chain ID handling for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Add fallback chain ID (`0x0`) when CAIP format can't be converted to hex for source chains
+  - Add fallback chain ID (`0x1`) for non-EVM destination chains
+- Update `getClientRequest` to create proper requests for all non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
```
Gudahtt, line 44: This seems like an internal change as well (a subset of "Add support for Bitcoin bridge transactions"). No need to document internal changes like this.

packages/bridge-controller/CHANGELOG.md
```diff
@@ -38,9 +43,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Rename fee handling for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `SolanaFees` type with `NonEvmFees` type
+  - Replace `solanaFeesInLamports` field with `nonEvmFeesInNative` field
```
Gudahtt, line 48: It's unclear _what_ is being renamed here, is this a property on a return value? The context made it seem like it was a package export or a method or something. More context would be appreciated here

packages/bridge-controller/CHANGELOG.md
```diff
@@ -38,9 +43,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Update `#appendSolanaFees` to `#appendNonEvmFees` to support all non-EVM chains
```
Gudahtt, line 49: This line seems like an internal-only change

packages/bridge-controller/CHANGELOG.md
```diff
@@ -38,9 +43,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - The `nonEvmFeesInNative` field stores fees in the smallest units for each chain (lamports for Solana, satoshis for Bitcoin)
+- Update Snap methods to use new unified interface for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `getFeeForTransaction` with `computeFee` method that returns fees in native token units
+  - Update fee calculation to handle different unit conversions per chain
+  - Support fee computation for Bitcoin and Solana chains
+- Update quote validation to support Bitcoin-specific trade data format ([#6454](https://github.com/MetaMask/core/pull/6454))
```
Gudahtt, line 55: Nit: This is fine, but maybe better to move these two entries ("Update quote validation" and "Update selectors and utilities") as sub-bullets of "Add support for Bitcoin bridge transactions". It seems like a detail of that addition, I don't think it's as helpful to frame it as a standalone change.

packages/bridge-controller/CHANGELOG.md
```diff
@@ -38,9 +43,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Rename fee handling for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `SolanaFees` type with `NonEvmFees` type
+  - Replace `solanaFeesInLamports` field with `nonEvmFeesInNative` field
+  - Update `#appendSolanaFees` to `#appendNonEvmFees` to support all non-EVM chains
+  - The `nonEvmFeesInNative` field stores fees in the smallest units for each chain (lamports for Solana, satoshis for Bitcoin)
+- Update Snap methods to use new unified interface for non-EVM chains ([#6454](https://github.com/MetaMask/core/pull/6454))
+  - Replace `getFeeForTransaction` with `computeFee` method that returns fees in native token units
```
Gudahtt, line 52: This seems like a breaking change as well. This is a change in the expectations we have of snaps this controller inter-operates with.

packages/bridge-controller/src/utils/snaps.ts
```diff
@@ -22,19 +23,34 @@ export const getMinimumBalanceForRentExemptionRequest = (snapId: string) => {
...
   snapId: string,
   transaction: string,
+  accountId: string,
+  scope: CaipChainId,
 ) => {
   return {
     snapId: snapId as never,
     origin: 'metamask',
-    handler: 'onRpcRequest' as never,
+    handler: 'onClientRequest' as never,
```
Gudahtt, line 45: Why cast to `never` 🤔 As the name might imply, `never` should _never_ be resolved on any of our types. It's meant to alert when a conditional type is broken.

PR #6456, opened 2025-09-04
Feat/shield create subscription crypto

tuna1207:
## Explanation

- Add `getCryptoApproveTransactionParams` which handle validating and prepare params to create crypto approve transaction for user to sign then submit to subscription service
- Add `startSubscriptionWithCrypto` which register submitted crypto transaction with subscription service to start subscription

## Checklist


packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,118 @@ export class SubscriptionController extends BaseController<
...
+  /**
+   * Calculate total subscription price amount from price info
+   * e.g: $8 per month * 12 months min billing cycles = $96
+   *
+   * @param price - The price info
+   * @returns The price amount
+   */
+  #getSubscriptionPriceAmount(price: ProductPrice) {
+    const amount =
+      (BigInt(price.unitAmount) / BigInt(10) ** BigInt(price.unitDecimals)) *
```
chaitanyapotti: we don't need BigInt here. we can simply use number. We know that max unitDecimals are always 2.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,118 @@ export class SubscriptionController extends BaseController<
...
+    if (!conversionRate) {
+      throw new Error('Conversion rate not found');
+    }
+    // conversion rate is a float string e.g: "1.0"
+    // We need to handle float conversion rates with integer math for BigInt.
+    // We'll scale the conversion rate to an integer by multiplying by 10^18 (or another large factor).
+    // This allows us to avoid floating point math and keep precision.
+    const CONVERSION_RATE_SCALE = 10n ** 18n;
+    const conversionRateScaled = BigInt(
+      Math.round(Number(conversionRate) * Number(CONVERSION_RATE_SCALE)),
```
chaitanyapotti: This can result in precision loss.
We want to ensure the scale is lower than Number.MAX_SAFE_INTEGER.
So, we should only scale with the digits we care about.
conversionRate is in usd decimal. In most currencies, we only care about 2 decimals (cents)
So, scale must be max of 10 ** 4 (most exchanges trade with max 4 decimals of usd)

tuna1207: Updated 👍

packages/subscription-controller/src/SubscriptionController.ts

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,120 @@ export class SubscriptionController extends BaseController<
...
+   *
+   * @param price - The price info
+   * @returns The price amount
+   */
+  #getSubscriptionPriceAmount(price: ProductPrice) {
+    const amount =
+      (BigInt(price.unitAmount) / BigInt(10) ** BigInt(price.unitDecimals)) *
+      BigInt(price.minBillingCycles);
+    return amount;
+  }
```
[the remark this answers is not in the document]
tuna1207: updated to scale

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -81,13 +91,7 @@ export class SubscriptionService implements ISubscriptionService {
...
       });
 
-      const responseBody = await response.json();
-      if (!response.ok) {
-        const { message, error } = responseBody as ErrorMessage;
-        throw new Error(`HTTP error message: ${message}, error: ${error}`);
-      }
-
-      return responseBody as Result;
+      return response;
```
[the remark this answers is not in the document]
tuna1207: It's been updated to use handleFetch from @metamask/utils

chaitanyapotti (approved):

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,124 @@ export class SubscriptionController extends BaseController<
...
+
+  /**
+   * Calculate total subscription price amount from price info
+   * e.g: $8 per month * 12 months min billing cycles = $96
+   *
+   * @param price - The price info
+   * @returns The price amount
+   */
+  #getSubscriptionPriceAmount(price: ProductPrice) {
+    // no need to use BigInt since max unitDecimals are always 2 for price
```
himanshuchawla009, line 331: If we are assuming it will be always 2 , it should be a constant.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,124 @@ export class SubscriptionController extends BaseController<
...
+    const CONVERSION_RATE_SCALE = 10n ** 4n;
+    const conversionRateScaled = BigInt(
+      Number(conversionRate) * Number(CONVERSION_RATE_SCALE),
+    );
+    const priceAmount = this.#getSubscriptionPriceAmount(price);
+    const priceAmountScaled = BigInt(
+      priceAmount * Number(CONVERSION_RATE_SCALE),
+    );
+
+    const amount =
```
himanshuchawla009: it looks like we are redundantly multiply and dividing with CONVERSION_RATE_SCALE,  above code can be simplified as:

```
const SCALE = 10n ** 4n;
const conversionRateScaled = BigInt(Number(conversionRate) * Number(SCALE));
const priceAmount = this.#getSubscriptionPriceAmount(price);
const priceAmountScaled = BigInt(priceAmount * Number(SCALE));
const tokenDecimals = BigInt(tokenPaymentInfo.decimals);

const amount = (priceAmountScaled * (10n ** tokenDecimals)) / conversionRateScaled;

```

packages/subscription-controller/src/types.ts
```diff
@@ -57,7 +50,16 @@ export type Subscription = {
...
   interval: RecurringInterval;
-  paymentMethod: PaymentMethod;
+  paymentMethod: SubscriptionPaymentMethod;
+};
+
+export type SubscriptionPaymentMethod = {
+  type: PaymentType;
+  crypto?: {
+    payerAddress: string;
+    chainId: string;
```
himanshuchawla009: is chainId in hex ?  in that case it would be safe to use `Hex` as type
 
 same for `payerAddress`

packages/subscription-controller/src/types.ts
```diff
@@ -76,28 +78,58 @@ export type StartSubscriptionResponse = {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+) => Promise<any>;
+
 export type ProductPrice = {
-  interval: string; // "month" | "year"
-  unitAmount: string; // amount in the smallest unit of the currency, e.g., cents
+  interval: RecurringInterval;
+  unitAmount: number; // amount in the smallest unit of the currency, e.g., cents
   unitDecimals: number; // number of decimals for the smallest unit of the currency
   currency: string; // "usd"
```
himanshuchawla009: would be cleaner to define type of supported currencies.

packages/subscription-controller/src/types.ts
```diff
@@ -119,11 +151,38 @@ export type PricingResponse = {
...
 
+export type GetCryptoApproveTransactionRequest = {
+  /**
+   * payment chain ID
+   */
+  chainId: string;
+  /**
+   * Payment token address
+   */
+  paymentTokenAddress: string;
```
himanshuchawla009: all on chain addresses should be of type hex.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,124 @@ export class SubscriptionController extends BaseController<
...
+    }
+
+    const tokenApproveAmount = this.#getTokenApproveAmount(
+      price,
+      tokenPaymentInfo,
+    );
+
+    return {
+      approveAmount: tokenApproveAmount.toString(),
+      spenderAddress: chainPaymentInfo.paymentAddress,
```
himanshuchawla009: would be cleaner to rename `paymentAddress` to `spenderAddress` for consistency.

himanshuchawla009 (changes requested):

packages/subscription-controller/src/SubscriptionController.ts
chaitanyapotti: It can be 4 for some currencies. Max is 4

packages/subscription-controller/src/SubscriptionController.ts
chaitanyapotti: This is not to lose precision when doing bigint math

packages/subscription-controller/CHANGELOG.md
```diff
@@ -14,5 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `cancelSubscription`: Cancel user active subscription.
 - `startShieldSubscriptionWithCard`: start shield subscription via card (with trial option) ([#6300](https://github.com/MetaMask/core/pull/6300))
 - Add `getPricing` method ([#6356](https://github.com/MetaMask/core/pull/6356))
+- Add `startSubscriptionWithCrypto` method ([#6456](https://github.com/MetaMask/core/pull/6456))
+- Add `getCryptoApproveTransactionParams` method ([#6456](https://github.com/MetaMask/core/pull/6456))
```
matthiasgeihs: nit: Could put all changes related to one PR in one entry
```suggestion
- Add methods `startSubscriptionWithCrypto` and `getCryptoApproveTransactionParams` method ([#6456](https://github.com/MetaMask/core/pull/6456))
```

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -63,6 +61,18 @@ export class SubscriptionService implements ISubscriptionService {
...
+    request: StartCryptoSubscriptionRequest,
+  ): Promise<StartCryptoSubscriptionResponse> {
+    const path = 'subscriptions/crypto';
+    return await this.#makeRequest(path, 'POST', request);
+  }
+
+  async getPricing(): Promise<PricingResponse> {
+    const path = 'pricing';
+    return await this.#makeRequest<PricingResponse>(path);
+  }
```
matthiasgeihs: Please don't move other functions around, if they are not in context of this PR. It can create merge conflicts and all kinds of other problems. (Unless you have a good reason for that, in which case it would help if you add a comment in the PR on GitHub.)

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -81,13 +91,7 @@ export class SubscriptionService implements ISubscriptionService {
         body: body ? JSON.stringify(body) : undefined,
       });
 
-      const responseBody = await response.json();
-      if (!response.ok) {
-        const { message, error } = responseBody as ErrorMessage;
-        throw new Error(`HTTP error message: ${message}, error: ${error}`);
-      }
-
-      return responseBody as Result;
```
matthiasgeihs, line 90: We can't do this. This means you rely on a very specific implementation of `fetch`. It's not clear at all that this is a requirement to the consumers of the library. What happens if you just pass in `fetch` as before?

If you wanna use `handleFetch`, then we should remove the injected fetch. But that's out of the scope of this PR. Please revert here. You can create a separate PR for this, if you think it's a good change.

packages/subscription-controller/src/constants.test.ts
```diff
@@ -5,23 +5,21 @@ describe('constants', () => {
...
       });
     });
 
     it('should return correct URLs for prd environment', () => {
       const result = getEnvUrls(Env.PRD);
       expect(result).toStrictEqual({
-        subscriptionApiUrl: 'https://subscription-service.api.cx.metamask.io',
+        subscriptionApiUrl: 'https://subscription.api.cx.metamask.io',
       });
     });
```
matthiasgeihs: We can simply remove these tests!?

packages/subscription-controller/src/types.ts
```diff
@@ -57,7 +50,16 @@ export type Subscription = {
   currentPeriodEnd: string; // ISO 8601
   status: SubscriptionStatus;
   interval: RecurringInterval;
-  paymentMethod: PaymentMethod;
+  paymentMethod: SubscriptionPaymentMethod;
```
matthiasgeihs, line 56: What's the reason for the type rename? We can't just rename types without a good reason. Would be helpful if you add a comment here on the PR why this is renamed.

packages/subscription-controller/src/types.ts
```diff
@@ -76,28 +78,58 @@ export type StartSubscriptionResponse = {
...
+};
+
 export type AuthUtils = {
   getAccessToken: () => Promise<string>;
 };
 
+export type FetchFunction = (
+  input: RequestInfo | URL,
+  init?: RequestInit,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
matthiasgeihs: Can't do this.
Can't use any without a good reason.
The reason should be specified next to the ignore statement.

packages/subscription-controller/src/types.ts
```diff
@@ -76,28 +78,58 @@ export type StartSubscriptionResponse = {
...
+) => Promise<any>;
+
 export type ProductPrice = {
-  interval: string; // "month" | "year"
-  unitAmount: string; // amount in the smallest unit of the currency, e.g., cents
+  interval: RecurringInterval;
+  unitAmount: number; // amount in the smallest unit of the currency, e.g., cents
```
matthiasgeihs, line 109: Why the heck are you changing the type here? You can't just change the type. This comes from the backend. I've previously discussed with Tai that Stripe is using a decimal string here and we would be following the same.

Note that string also is used here for precision reasons. `number` has only very limited precision. Using `string` allows us to encode also big ints with much larger precision.

I think you can assume that `unitAmount` is always an integer.

packages/subscription-controller/src/types.ts
```diff
@@ -119,11 +151,38 @@ export type PricingResponse = {
...
+  approveAmount: string;
+  spenderAddress: string;
+  paymentTokenAddress: string;
+  chainId: string;
+};
+
 export type ISubscriptionService = {
   getSubscriptions(): Promise<GetSubscriptionsResponse>;
   cancelSubscription(request: { subscriptionId: string }): Promise<void>;
+  getPricing(): Promise<PricingResponse>;
```
matthiasgeihs: Please don't move other functions around arbitrarily.
You can simply insert `startSubscriptionWithCrypto` below `startSubscriptionWithCard` and that's fine.
If everyone starts reordering the functions according to their liking with each PR, this becomes a mess.
If we want to to have defined ordering, we can go with alphabetical, which is what for example TransactionController does.

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -29,8 +36,8 @@ const MOCK_SUBSCRIPTION: Subscription = {
     {
       name: ProductType.SHIELD,
       id: 'prod_shield_basic',
-      currency: 'USD',
-      amount: 9.99,
+      currency: 'usd',
+      amount: 900,
```
matthiasgeihs, line 40: Should be string

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -42,6 +49,43 @@ const MOCK_SUBSCRIPTION: Subscription = {
   },
 };
 
+const MOCK_PRODUCT_PRICE: ProductPricing = {
+  name: ProductType.SHIELD,
+  prices: [
+    {
+      interval: RecurringInterval.month,
+      currency: 'usd',
+      unitAmount: 900,
```
matthiasgeihs, line 58: Should be string

packages/subscription-controller/src/SubscriptionController.test.ts
```diff
@@ -546,4 +633,179 @@ describe('SubscriptionController', () => {
       });
     });
   });
+
+  describe('getCryptoApproveTransactionParams', () => {
+    it('returns transaction params for crypto approve transaction', async () => {
+      await withController(async ({ controller, mockService }) => {
+        // Provide product pricing and crypto payment info with unitDecimals small to avoid integer div to 0
```
matthiasgeihs, line 641: I don't get this comment. Can you elaborate?

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: `chainPaymentInfo` comes from the backend. From the viewpoint of the backend `paymentAddress` makes sense. `spenderAddress` is specific wording to the approval transaction. So I think it's fine that the fields are called differently.

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs, line 331: I think this assumption is quite dangerous. We are dealing with price information so correctness is important. If we are assuming a fixed value here, then we don't need this field, it could be a constant, as Himanshu says. If we have this field, we should deal with it correctly. I.e., use big ints to calculate.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,124 @@ export class SubscriptionController extends BaseController<
...
+  /**
+   * Calculate total subscription price amount from price info
+   * e.g: $8 per month * 12 months min billing cycles = $96
+   *
+   * @param price - The price info
+   * @returns The price amount
+   */
+  #getSubscriptionPriceAmount(price: ProductPrice) {
+    // no need to use BigInt since max unitDecimals are always 2 for price
+    const amount =
```
matthiasgeihs, line 332: Should use big int for the calculation. Alternatively, assert that the input values are within the expected range.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -222,6 +249,124 @@ export class SubscriptionController extends BaseController<
...
+    const priceAmount = this.#getSubscriptionPriceAmount(price);
+    const priceAmountScaled = BigInt(
+      priceAmount * Number(CONVERSION_RATE_SCALE),
+    );
+
+    const amount =
+      ((priceAmountScaled / CONVERSION_RATE_SCALE) *
+        BigInt(10) ** BigInt(tokenPaymentInfo.decimals) *
+        CONVERSION_RATE_SCALE) /
+      conversionRateScaled;
```
matthiasgeihs: I think we should be able to do that computation without relying on floating point arithmetic.

Does the following work?
```
tokenApproveAmount = priceInUSD * tokenPerUSD
= (priceAmount / priceDecimals) * (tokenAmount / tokenDecimals)
= (priceAmount * tokenAmount) / (priceDecimals * tokenDecimals)
```

packages/subscription-controller/src/SubscriptionController.ts
tuna1207: that `paymentAddress` type is from server response, we can update `sepnderAddress` to `paymentAddress` for consistency

packages/subscription-controller/src/types.ts
tuna1207: updated

packages/subscription-controller/src/SubscriptionController.ts
matthiasgeihs: I'm seeing
```
priceAmountScaled = BigInt(priceAmount * Number(CONVERSION_RATE_SCALE));
```
and then in then immediately in the next statement
```
... (priceAmountScaled / CONVERSION_RATE_SCALE) * ...
```
This just reverts the scaling!?

packages/subscription-controller/src/types.ts
tuna1207: updated

packages/subscription-controller/src/types.ts
tuna1207: updated

packages/subscription-controller/src/SubscriptionController.ts

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: @matthiasgeihs oh this happen when i test this in extension locally, fetch is only allowed to be used in the extension context, not injected into other library, the same pattern is being used with `bridgeController` and `bridgeStatusController`

packages/subscription-controller/src/types.ts
tuna1207: oh this one was updated when i was checking backend latest code and saw that `unitAmount` was returning number here, if this was decided to be updated, this shouldn't be number anymore

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: i think those move was because of my conflict resolve, will update those

packages/subscription-controller/src/types.ts
tuna1207: Confirmed to be `number` type

packages/subscription-controller/src/types.ts
tuna1207: Yes, this one was tho differentiate it with `PricingPaymentMethod` type, will include the explaination in the PR next time

packages/subscription-controller/src/types.ts
tuna1207: As discussed, will use `handleFetch` internally, no need to inject fetch

packages/subscription-controller/src/constants.test.ts
tuna1207: updated

packages/subscription-controller/src/types.ts
tuna1207: updated

packages/subscription-controller/src/SubscriptionController.ts
chaitanyapotti: Currencies always have decimals less than MAX_SAFE_INTEGER in javascript

packages/subscription-controller/src/SubscriptionService.test.ts
```diff
@@ -1,5 +1,4 @@
 import { handleFetch } from '@metamask/controller-utils';
-import nock, { cleanAll, isDone } from 'nock';
```
matthiasgeihs, line 1: Why did you remove the usage of `nock` here?
Remember that we should keep changes minimal within the context of the PR, if possible.

packages/subscription-controller/src/SubscriptionService.ts
```diff
@@ -1,3 +1,5 @@
+import { handleFetch } from '@metamask/controller-utils';
```
matthiasgeihs, line 1: need to move `@metamask/controller-utils` to main dependencies now

matthiasgeihs (dismissed):

matthiasgeihs (approved):

PR #6460, opened 2025-09-04
refactor: migrate `EnsController` to `@metamask/messenger`

mikesposito:
## Checklist

### DIFF
--- a/packages/ens-controller/src/EnsController.ts
+++ b/packages/ens-controller/src/EnsController.ts
@@ -1,6 +1,9 @@
 import { Web3Provider } from '@ethersproject/providers';
-import type { RestrictedMessenger } from '@metamask/base-controller';
-import { BaseController } from '@metamask/base-controller';
+import {
+  BaseController,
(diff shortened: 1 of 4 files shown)


packages/ens-controller/src/EnsController.ts
```diff
@@ -69,16 +73,24 @@ export type EnsControllerState = {
   ensResolutionsByAddress: { [key: string]: string };
 };
 
+export type EnsControllerActions = ControllerGetStateAction<
```
mcmire, line 77: It looks like `ens-controller` is using a barrel export in `index.ts`, so this would be a new export in the package. Should we note this in the changelog? (Same for `EnsControllerEvents`.)

mcmire:
Need to resolve conflicts :(

packages/ens-controller/src/EnsController.ts
Gudahtt: Good catch, done here: e37bf9a93

PR #6468, opened 2025-09-04
chore(multichain-account-service): refactor and improvements around discovery logic 

ccharly:
## Explanation

Improvements.

## References

N/A

## Checklist


packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -364,32 +364,28 @@ export class MultichainAccountWallet<
...
+    // from there).
+    let maxGroupIndex = this.getNextGroupIndex();
+
+    const providerContexts: {
+      provider: NamedAccountProvider;
+      stopped: boolean;
+      running?: Promise<void>;
+      groupIndex: number;
+      count: number;
+    }[] = this.#providers.map((provider) => ({
```
ccharly: Actually, instead of using a `Map`, we can just have the `provider` being part of the "discovery context". So everytime we `schedule` we pass-in the entire context (which holds a ref to its `provider`).

ccharly: This also avoid having those `.get() as ...`, since we know the context is defined as parameter already.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -364,32 +364,28 @@ export class MultichainAccountWallet<
...
+      provider,
+      stopped: false,
+      running: undefined,
+      groupIndex: maxGroupIndex,
+      count: 0,
+    }));
+
+    const schedule = async (
+      context: (typeof providerContexts)[number],
```
ccharly: Not sure about this though.. I thought of adding a local `type DiscoveryContext` (local to this function), WDYT?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -398,37 +394,39 @@ export class MultichainAccountWallet<
...
+        context.running = undefined;
       }
 
-      providerCtx.groupIndex = index;
+      // Keep track of the current group index being discovery by this provider.
+      context.groupIndex = groupIndex;
 
-      providerCtx.running = (async () => {
+      // Pending promise of the provider's discovery for this group index.
+      return (async (): Promise<void> => {
```
ccharly: We now return the un-awaited promise, since our pattern is `context.running = schedule(...)`, this make sure the `Promise.race` will actually `await` this non-awaited promise.

> [!CAUTION]
> This was the reason why the discovery was hanging initially.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -364,32 +364,28 @@ export class MultichainAccountWallet<
...
-        stopped: boolean;
-        running?: Promise<void>;
-        groupIndex: number;
-        count: number;
-      }
-    >();
-
-    for (const p of providers) {
-      providerContexts.set(p, { stopped: false, groupIndex: 0, count: 0 });
-    }
```
ccharly, line 376: We don't need this anymore since we create the initial contexts states with the `.map` above.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -438,44 +436,43 @@ export class MultichainAccountWallet<
...
+      context.running = schedule(context, maxGroupIndex);
     }
 
     let racers: Promise<void>[] = [];
     do {
       racers = [...providerContexts.values()]
-        .map((c) => c.running)
-        .filter(Boolean) as Promise<void>[];
+        .filter((ctx) => ctx.running !== undefined)
+        .map((ctx) => ctx.running) as Promise<void>[];
```
ccharly: I thought doing the `filter` before would reduce the number of iterations, instead of `map` (N) + `filter` (N -> M) we do `filter (N -> M)` + `map` (M).

Obviously, this is extremely nitpick given the number of providers we have 🙈

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -365,122 +374,76 @@ export class MultichainAccountWallet<
...
 
-      providerCtx.groupIndex = index;
+      while (!context.stopped) {
+        // Fast‑forward to current high‑water mark
+        const targetGroupIndex = Math.max(context.groupIndex, maxGroupIndex);
 
-      providerCtx.running = (async () => {
+        console.log(step('STARTED'));
+
+        let accounts: undefined | Bip44Account<KeyringAccount>[];
```
hmalik88: Can we do this so we don't need the optional chaining below.
```suggestion
        let accounts: Bip44Account<KeyringAccount>[] = [];
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -365,122 +374,76 @@ export class MultichainAccountWallet<
...
-       */
-      if (providerCtx.running) {
-        // Wait for current scheduled job before starting a new one.
-        await providerCtx.running.catch((error) => {
-          console.error(error);
-        });
-        providerCtx.running = undefined;
-      }
+    // One serialized loop per provider; all run concurrently
+    const run = async (context: AccountProviderDiscoveryContext) => {
```
hmalik88:
```suggestion
    const runProviderDiscovery = async (context: AccountProviderDiscoveryContext) => {
```

ccharly: Applied the suggestion manually since it required to be linted too!

hmalik88 (approved):

Issue #6443, opened 2025-09-03
Add new `includeInStateLogs` and `usedInUi` controller metadata properties


Gudahtt:
Add new `includeInStateLogs` and `usedInUi` controller metadata properties, as described in this ADR: https://github.com/MetaMask/decisions/blob/main/decisions/core/0014-Expand-Controller-Metadata.md

These properties are necessary for the clients to know what to include in state logs, and which data can be excluded from UI state (to improve UI state sync performance).

Gudahtt:
This has been completed

PR #6473, opened 2025-09-04
feat: Add new metadata to confirmation controllers

Gudahtt:
## Explanation

The new metadata properties `includeInStateLogs` and `usedInUi` have been added to all controllers maintained by the confirmations team.

## References

Relates to #6443

## Checklist

### DIFF
--- a/packages/user-operation-controller/src/UserOperationController.test.ts
+++ b/packages/user-operation-controller/src/UserOperationController.test.ts
@@ -1,3 +1,4 @@
+import { deriveStateFromMetadata } from '@metamask/base-controller';
 import { ApprovalType } from '@metamask/controller-utils';
 import { errorCodes } from '@metamask/rpc-errors';
 import {
@@ -34,7 +35,6 @@ import {
   validateSignUserOperationResponse,
   validateUpdateUserOperationResponse,
 } from './utils/validation';
-import { deriveStateFromMetadata } from '@metamask/base-controller';
 
 jest.mock('@metamask/transaction-controller');
 jest.mock('./utils/gas');


packages/address-book-controller/src/AddressBookController.ts
```diff
@@ -147,7 +147,12 @@ export type AddressBookControllerEvents =
   | AddressBookControllerContactDeletedEvent;
 
 const addressBookControllerMetadata = {
-  addressBook: { persist: true, anonymous: false },
+  addressBook: {
+    includeInStateLogs: true,
```
matthewwalsh0, line 151: Minor, very pedantic, but how are deciding the order of these properties if we're defining them in every controller? Alphabetical? Importance?

packages/address-book-controller/src/AddressBookController.ts
```diff
@@ -147,7 +147,12 @@ export type AddressBookControllerEvents =
...
+    persist: true,
+    anonymous: false,
+    usedInUi: true,
```
matthewwalsh0, line 154: Minor, will this be used to filter the background RPC patches sent to the UI in extension?

Will it exclude properties from the Redux store in mobile?

packages/address-book-controller/src/AddressBookController.ts
```diff
@@ -147,7 +147,12 @@ export type AddressBookControllerEvents =
   | AddressBookControllerContactDeletedEvent;
 
 const addressBookControllerMetadata = {
-  addressBook: { persist: true, anonymous: false },
+  addressBook: {
+    includeInStateLogs: true,
+    persist: true,
+    anonymous: false,
```
matthewwalsh0, line 153: Did we plan to rename this to `includeInDebugLogs`, is that a future PR?

Gudahtt: Yes exactly, it will happen as part of the migration to `base-controller/next`, tracked here: https://github.com/MetaMask/core/issues/5626

packages/address-book-controller/src/AddressBookController.ts
Gudahtt: I have been putting them in alphabetical order, leaving the order of pre-existing properties alone to minimize churn. I intended to move `anonymous` to be alphabetical as part of the rename

packages/address-book-controller/src/AddressBookController.ts
Gudahtt: Yep, that's the idea! We won't start using it to filter in practice without another round of review though, so we'll have an opportunity to catch mistakes still.

Initially the plan is just to filter the UI state in extension. The mobile Redux store is also used for persistence, so we can't really do it there. But there is work underway to change that. After the storage changes on mobile, I hope we can filter the UI Redux state using this value on mobile as well.

PR #6477, opened 2025-09-05
refactor: fully extract middleware into EIP-5792 Middleware package

ffmcgee725:
## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-691

## Checklist


ffmcgee725:
@metamaskbot publish-preview

packages/eip-5792-middleware/src/constants.ts
```diff
@@ -18,3 +30,47 @@ export enum EIP5792ErrorCode {
...
+      object({
+        to: optional(HexChecksumAddressStruct),
+        data: optional(StrictHexStruct),
+        value: optional(StrictHexStruct),
+        capabilities: optional(CapabilitiesStruct),
+      }),
+    ),
+    capabilities: optional(CapabilitiesStruct),
+  }),
+]);
```
jiexi: i know some of these aren't actually typescript types, but they're effectively runtime types and probably belong in `types.ts`. What are your thoughts here?

packages/eip-5792-middleware/src/methods/wallet_sendCalls.ts
```diff
@@ -0,0 +1,47 @@
...
+        getAccounts,
+      })
+    : undefined;
+
+  const sendCalls: SendCallsPayload = {
+    ...params,
+    from,
+  };
+
+  res.result = await processSendCalls(sendCalls, req);
```
jiexi, line 50: oh hmm, interesting. I was originally thinking that the hooks got inlined into the method/middleware. It's an interesting pattern to keep them separated if there are cases where we want to process sendCalls (or get capabilities, etc) outside of a rpc call. The awkward part though is that this hook still operates off a request object. 

I still think the hook should get inlined here, but what do you think about this?

jiexi: ahh okay i see. The hook passed into this middleware doesn't directly call the hooks in the hooks folder, but instead unwraps the args provided here as well as injecting any dependencies (i.e. hooks for the hook 🙃). I don't think this is the right pattern

adonesky1: Yeah I'm also not a huge fan of this pattern and found it confusing to look at when I first observed it in [the original implementation](https://github.com/MetaMask/eth-json-rpc-middleware/blob/main/src/methods/wallet-send-calls.ts#L61-L96). As you can see (from that link) however this is just copying over the pattern as it was in `@metamask/eth-json-rpc-middleware`. I think we should refatctor it to be less nested, more similar to our handlers/hooks pattern that we have for all the handlers in [eip1193-permission-middleware](https://github.com/MetaMask/core/tree/main/packages/eip1193-permission-middleware/src) or [multichain-api-middleware](https://github.com/MetaMask/core/tree/main/packages/multichain-api-middleware) for example, but I think its fair game to use this pattern for the initial pass since it preserves the existing pattern

packages/eip-5792-middleware/src/constants.ts
ffmcgee725: Makes sense to me. done here https://github.com/MetaMask/core/pull/6477/commits/30f2cdfd5d58f926ffb8077094185f15cec94426

packages/eip-5792-middleware/src/methods/wallet_sendCalls.ts
ffmcgee725: Issue created to track this: https://consensyssoftware.atlassian.net/browse/WAPI-699

packages/eip-5792-middleware/src/utils.ts
```diff
@@ -36,3 +38,86 @@ export function getAccountKeyringType(
...
   return keyringType as KeyringTypes;
 }
+
+/**
+ *
+ * @param address a
+ * @param req a
+ * @param param2 s
+ * @param param2.getAccounts s
+ * @returns s
```
jiexi: 🙏

jiexi: same for the others in this file. Thank you, sir! 🫡

ffmcgee725: wops, my bad forgot to 🧹 that section

https://github.com/MetaMask/core/pull/6477/commits/82a725cb24fb588cc74e61416db6132c3ae399eb

packages/eip-5792-middleware/src/utils.ts
```diff
@@ -36,3 +38,93 @@ export function getAccountKeyringType(
...
+
+/**
+ * Validates and normalizes a keyholder address for EIP-5792 operations.
+ *
+ * @param address - The Ethereum address to validate and normalize.
+ * @param req - The JSON-RPC request object for permission checking.
+ * @param options - Configuration object containing the getAccounts function.
+ * @param options.getAccounts - Function to retrieve accounts for the requester.
+ * @returns A normalized (lowercase) hex address if valid and authorized.
+ * @throws JsonRpcError with unauthorized error if the requester doesn't have      permission to access the address.
```
jiexi, line 50:
```suggestion
 * @throws JsonRpcError with unauthorized error if the requester doesn't have permission to access the address.
```

jiexi (approved):

PR #6478, opened 2025-09-05
feat: add `createMultichainAccountWallet` to `MultichainAccountService`

hmalik88:
### MultichainAccountService
1. `createMultichainAccountWallet` method that takes a mnemonic and creates a new keyring from that mnemonic. It ultimately returns the newly created wallet from that entropy source. This method should be called from the clients, the wallet is intentionally empty and ripe for calling of the `discoverAndCreateAccounts` method on it.

## Checklist


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -293,6 +301,58 @@ export class MultichainAccountService {
...
+
+    const alreadyHasImportedSrp = existingKeyrings.some((keyring) => {
+      if (!keyring.mnemonic) {
+        return false;
+      }
+      return (
+        convertEnglishWordlistIndicesToCodepoints(keyring.mnemonic).toString(
+          'utf8',
+        ) === mnemonic
+      );
```
ccharly: Actually, after some digging and internal discussion, I think we can do it the other way around.

Instead of comparing `string`s we would compare bytes instead. Thus, we only need a conversion function from mnemonic to bytes, so we can use this one from `@metamask/key-tree`:
- https://github.com/MetaMask/key-tree/blob/e89af7d409e4fabe085179dde0edf9724374fffe/src/utils.ts#L403-L423

This way, we never have to convert `keyring.mnemonic` back to a `string`.

Though, we won't be able to use `===` here, and would have to do a comparison of each bytes (but we might have something already somewhere for this).

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -290,6 +301,57 @@ export class MultichainAccountService {
...
+
+    if (alreadyHasImportedSrp) {
+      throw new Error('This Secret Recovery Phrase has already been imported.');
+    }
+
+    const result = (await this.#messenger.call(
+      'KeyringController:addNewKeyring',
+      KeyringTypes.hd,
+      { mnemonic },
+    )) as KeyringMetadata;
```
ccharly: Do we really need that typecast? 🤔 I can this the return type is well-defined on the `KeyringController` method already?

hmalik88: Nope, I could have sworn it complained before 🤔

ccharly (approved):

PR #6483, opened 2025-09-07
feat: Add bulk token scanning and introduce shared cache manager #6617

AugmentedMode:
## Explanation

This PR extends the **PhishingController** to support bulk token screening with caching, similar to the existing URL scanning functionality.  This PR also replaces the `urlScanCache` with a reusable `cacheManager` class that is shared between both `urlScanCache` and now `tokenScanCache`

**Changes introduced**  
- Added a new `bulkScanTokens` method that can screen up to 100 tokens per batch
- Implemented a new `cacheManager` class that handles caching logic with a 15-minute TTL.  
  - Replaced `urlScanCache` with `cacheManager`.  
  - Introduced `tokenScanCache` using the same `cacheManager` for consistency and reuse.  
- Defined cache key format as `${chainId}:${tokenAddress}` for token scans.  
- Registered `bulkScanTokens` as a controller action under `PhishingControllerBulkScanTokensAction`.  
- Exported new types for extension consumption and exposed the scanning functionality via the messenger.  

Screenshot showcasing the `tokenScanCache` is implemented as expected 

This is grabbed via the extension console using await stateHooks.getPersistedState()  after triggering a swap in uniswap. You will notice i was doing a swap for USDC which is the only token address stored in the cache.

Screenshot 2025-09-15 at 10 28 15 AM

## Checklist


packages/phishing-controller/src/CacheManager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    it('should return true for existing keys', () => {
+      cache.set('key1', { value: 'value1' });
+      expect(cache.has('key1')).toBe(true);
+    });
+
+    it('should return true for expired keys', () => {
+      cache.set('key1', { value: 'value1' });
+      clock.tick(301 * 1000);
+      expect(cache.has('key1')).toBe(true);
+    });
```
cryptotavares: Is this itended? If a key is considered expired why are we returning true here? I would expect the logic here to follow the `get` logic as well. Also do we even need a `has` method? It does not seem to be used anywhere.

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -861,6 +909,154 @@ export class PhishingController extends BaseController<
...
+    '0x74c': 'soneium',
+    '0x79a': 'soneium-minato',
+    '0x14a34': 'base-sepolia',
+    '0xab5': 'abstract',
+    '0x849ea': 'zero-network',
+    '0x138de': 'berachain',
+    '0x82': 'unichain',
+    '0x7e4': 'ronin',
+    '0x127': 'hedera',
+  };
```
cryptotavares: For better type safing we could use `as const` here. You can find more details about the [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions), as well as how it improves type inferrence in our [contributor docs](https://github.com/MetaMask/contributor-docs/blob/main/docs/typescript.md#type-inference)

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -861,6 +909,154 @@ export class PhishingController extends BaseController<
...
+
+  /**
+   * Scan multiple tokens for malicious activity in bulk.
+   *
+   * @param request - The bulk scan request containing chainId and tokens.
+   * @param request.chainId - The chain ID in hex format (e.g., '0x1' for Ethereum).
+   * @param request.tokens - Array of token addresses to scan.
+   * @returns A mapping of lowercase token addresses to their scan results. Tokens that fail to scan are omitted.
+   */
+  bulkScanTokens = async (
```
cryptotavares, line 986: This method should be broken down into smaller methods. It currently seems to be doing too much: identify which tokens are on cache and which ones we need to fetch, handles all the fetch logic, populates the cache from the fetch ones. It can still be the parent method, but if we move some of its logic into its own methods, we will reduce cyclomatic complexity and improve its readability

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -861,6 +909,154 @@ export class PhishingController extends BaseController<
...
+    request: BulkTokenScanRequest,
+  ): Promise<BulkTokenScanResponse> => {
+    const { chainId, tokens } = request;
+
+    if (!tokens || tokens.length === 0) {
+      return {};
+    }
+
+    // Limit to 100 tokens per request
```
cryptotavares: No need for this code comment... the const is pretty readable by itself

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -861,6 +909,154 @@ export class PhishingController extends BaseController<
...
+
+    if (!chain) {
+      console.warn(`Unknown chain ID: ${chainId}`);
+      return {};
+    }
+
+    const results: Record<string, TokenScanResult> = {};
+    const tokensToFetch: string[] = [];
+
+    // Check cache for each token
```
cryptotavares: btw this comments can easily be removed if we create smaller methods to handle each unit of logic lets say.

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -861,6 +909,154 @@ export class PhishingController extends BaseController<
...
+              console.warn(
+                `Token bulk screening API error: ${res.status} ${res.statusText}`,
+              );
+              return null;
+            }
+
+            return await res.json();
+          },
+          true,
+          8000, // 8 second timeout
```
cryptotavares: should this be a const, instead of adding a "magic number" and then a comment to make it readable? (calling it magic number because of the [`no-magic-numbers` eslint rule](https://eslint.org/docs/latest/rules/no-magic-numbers))

packages/phishing-controller/src/CacheManager.test.ts
AugmentedMode: I went ahead and removed the `has` method! But this was intentional where the `get` uses the ttl logic while the `has` ignores it. But I agree with you that its best to probaly just remove it !

PR #6490, opened 2025-09-08
feat(backend-platform): Add Websocket and AccountActivity Service

Kriys94:
## Checklist


Kriys94:
Note: Add healthcheck and jitter

Kriys94:
@metamaskbot publish-preview

packages/assets-controllers/package.json
```diff
@@ -54,6 +54,7 @@
     "@ethersproject/contracts": "^5.7.0",
     "@ethersproject/providers": "^5.7.0",
     "@metamask/abi-utils": "^2.0.3",
+    "@metamask/backend-platform": "workspace:^",
```
salimtb: i guess we will replace this with a published version ?

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
+import type { RestrictedMessenger } from '@metamask/base-controller';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mcmire: Why are we not using this variable? Should we use it instead of ignoring the fact that we are not using it?

packages/backend-platform/src/AccountActivityService.ts
```diff
@@ -0,0 +1,652 @@
...
+/**
+ * Configuration options for the account activity service
+ */
+export type AccountActivityServiceOptions = {
+  /** Custom subscription namespace (default: 'account-activity.v1') */
+  subscriptionNamespace?: string;
+};
+
+// Action types for the messaging system
+export type AccountActivityServiceSubscribeAccountsAction = {
```
mcmire: There is a pattern that we introduced a little while ago that DRYs up the code around "method actions", or actions that merely point to methods. 

The way that this works is:

- You define a `MESSENGER_EXPOSED_METHOD` constant that lists each method you'd like to expose through the messenger
- You run `yarn generate-method-action-types` to generate a types file
- You import `AccountActivityServiceMethodActions` from the generated file into this file, and you add this type to the `AccountActivityServiceActions` union.
- Instead of registering each action manually in the constructor, you add:
    ```
    this.#messenger.registerMethodActionHandlers(
      this,
      MESSENGER_EXPOSED_METHODS,
    );
    ```
- Now you can remove `AccountActivityServiceSubscribeAccountsAction`, `AccountActivityServiceUnsubscribeAccountsAction`, etc.

What do you think of using this pattern? You can see an example in `SampleGasPricesService` here: https://github.com/MetaMask/core/blob/c45132e2c5549064f611e7ab30dd3448f9950df8/packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts#L25

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+export type WebSocketConnectionInfo = {
+  state: WebSocketState;
+  url: string;
+  reconnectAttempts: number;
+  lastError?: string;
+  connectedAt?: number;
+};
+
+// Action types for the messaging system
+export type WebSocketServiceInitAction = {
```
mcmire: See note in `AccountActivityService` for DRYing up these types.

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+  // Value: ChannelCallback configuration
+  readonly #channelCallbacks = new Map<string, ChannelCallback>();
+
+  /**
+   * Creates a new WebSocket service instance
+   *
+   * @param options - Configuration options including messenger
+   */
+  constructor(
+    options: WebSocketServiceOptions & { messenger: WebSocketServiceMessenger },
```
mcmire: Nit: `messenger` is an option of this service, so instead of adding it here, should it be in `WebSocketServiceOptions`, and then you can use `Omit<WebSocketServiceOptions, 'messenger'>` as the type for the `#options` property?

Alternatively, should each property that is currently in `#options` be its own property?

packages/core-backend/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  AccountActivityServiceAccountUnsubscribedEvent,
+  AccountActivityServiceTransactionUpdatedEvent,
+  AccountActivityServiceBalanceUpdatedEvent,
+  AccountActivityServiceSubscriptionErrorEvent,
+  AccountActivityServiceStatusChangedEvent,
+  AccountActivityServiceEvents,
+  AccountActivityServiceMessenger,
+} from './AccountActivityService';
+export {
+  ACCOUNT_ACTIVITY_SERVICE_ALLOWED_ACTIONS,
```
mcmire, line 49: Is there a reason why these constants are being exported? We don't typically do this in packages. If we need to use an action or event elsewhere we repeat it.

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -119,7 +117,9 @@ export type AllowedEvents =
   | TokensControllerStateChangeEvent
   | PreferencesControllerStateChangeEvent
   | NetworkControllerStateChangeEvent
-  | KeyringControllerAccountRemovedEvent;
+  | KeyringControllerAccountRemovedEvent
+  | AccountActivityServiceBalanceUpdatedEvent
```
mcmire, line 121: Heads up that updates to `AllowedEvents` in a package are breaking changes (because the allowlist needs to be updated on the client side). Can you mention this in the changelog?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -78,20 +87,9 @@ export type TokenBalancesControllerGetStateAction = ControllerGetStateAction<
   TokenBalancesControllerState
 >;
 
-export type TokenBalancesControllerUpdateChainPollingConfigsAction = {
```
mcmire, line 81: It looks like we are removing these actions, should we mention this in the changelog?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -464,6 +475,16 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     );
   }
 
+  /**
+   * Get the default polling interval for this controller
+   * This is the base interval used for chains without specific configuration
+   * 
+   * @returns The default polling interval in milliseconds
+   */
+  getDefaultPollingInterval(): number {
```
mcmire, line 512: It looks like we are adding a new method to the controller, should we mention this in the changelog?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -792,6 +813,171 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     });
   };
 
+  /**
+   * Handle real-time balance updates from AccountActivityService
+   * Processes balance updates and updates the token balance state
+   * If any balance update has an error, triggers fallback polling for the chain
+   */
+  readonly #onAccountActivityBalanceUpdate = async ({
```
mcmire, line 858: Anything to mention in the changelog about these changes? Any differences in behavior?

mcmire (commented):
Hey, I noticed that a new package was being added to the monorepo. I haven't dived deeply into the implementation yet but I did leave some comments. 

It looks like we are introducing a new package and then trying to use it right away. To make the changes easier to merge and queue up, what are your thoughts on extracting the changes to `assets-controllers` to a new PR?

packages/backend-platform/src/AccountActivityService.ts
```diff
@@ -0,0 +1,652 @@
...
+ * System notification data for chain status updates
+ */
+export type SystemNotificationData = {
+  /** Array of chain IDs affected (e.g., ['eip155:137', 'eip155:1']) */
+  chainIds: string[];
+  /** Status of the chains: 'down' or 'up' */
+  status: 'down' | 'up';
+};
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mcmire: Why are we not using this variable? Should we use it?

packages/backend-platform/src/AccountActivityService.ts
```diff
@@ -0,0 +1,652 @@
...
+  typeof SERVICE_NAME,
+  AccountActivityServiceActions | AccountActivityServiceAllowedActions,
+  AccountActivityServiceEvents | AccountActivityServiceAllowedEvents,
+  AccountActivityServiceAllowedActions['type'],
+  AccountActivityServiceAllowedEvents['type']
+>;
+
+/**
+ * Account Activity Service
+ *
```
mcmire: Nit: Do we need this part? We already know the name of the symbol being documented.

``` suggestion
```

Otherwise, great JSDocs! We definitely need more.

packages/backend-platform/src/AccountActivityService.ts
```diff
@@ -0,0 +1,652 @@
...
+   * @param options - Configuration options including messenger and WebSocket service
+   */
+  constructor(
+    options: AccountActivityServiceOptions & {
+      messenger: AccountActivityServiceMessenger;
+      webSocketService: WebSocketService;
+    },
+  ) {
+    this.#messenger = options.messenger;
+    this.#webSocketService = options.webSocketService;
```
mcmire: Instead of taking the WebSocketService as an argument, what are your thoughts on having it use it through the messenger?

packages/backend-platform/src/AccountActivityService.ts
```diff
@@ -0,0 +1,652 @@
...
+  // =============================================================================
+  // Account Subscription Methods
+  // =============================================================================
+
+  /**
+   * Get the currently subscribed account address
+   *
+   * @returns The CAIP-10 formatted address of the currently subscribed account, or null if none
+   */
+  getCurrentSubscribedAccount(): string | null {
```
mcmire: Why not simply expose the property?

packages/assets-controllers/package.json
cryptodev-2s: We can already just reference the local version instead of workspace

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+ * 1. Calling disconnect() when app goes to background
+ * 2. Calling connect() when app returns to foreground
+ * 3. Calling destroy() on app termination
+ */
+export class WebSocketService {
+  readonly #messenger: WebSocketServiceMessenger;
+
+  readonly #options: Required<WebSocketServiceOptions>;
+
+  #ws!: WebSocket;
```
mcmire: Why can we assume this is set?

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+      `BackendWebSocketService:removeChannelCallback`,
+      this.removeChannelCallback.bind(this),
+    );
+
+    this.#messenger.registerActionHandler(
+      `BackendWebSocketService:getChannelCallbacks`,
+      this.getChannelCallbacks.bind(this),
+    );
+
+    this.init().catch((error) => {
```
mcmire: It looks like we're doing an asynchronous operation in the controller. We don't have a current guideline on this but that's something that we should really avoid if possible, as it leads to unpredictable behavior.

An asynchronous `init` method is something that's become an unofficial pattern in various places, and fortunately it seems that you've implemented this. What are your thoughts on asking consumers to await the `init` method after initializing the WebSocketService instead of automatically calling it in the constructor?

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+   * Initializes and connects the WebSocket service
+   *
+   * @returns Promise that resolves when initialization is complete
+   */
+  async init(): Promise<void> {
+    try {
+      await this.connect();
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : 'Unknown initialization error';
```
mcmire: We have a utility in `@metamask/utils` to convert a throwable to a string, what do you think about using that? 

``` suggestion
      const errorMessage = getErrorMessage(error);
```

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+      return;
+    }
+
+    // If already connecting, wait for the existing connection attempt to complete
+    if (this.#state === WebSocketState.CONNECTING && this.#connectionPromise) {
+      await this.#connectionPromise;
+      return;
+    }
+
+    console.log(`🔄 Starting connection attempt to ${this.#options.url}`);
```
mcmire: It looks like you are using various log levels to print messages, so what are your thoughts on using `loglevel` instead of `console.*`? This is a dependency we use in other `core` packages, and it allows log messages to be controlled at the build level (so we can show `console.log` errors in dev but only `console.info` errors in prod, for instance).

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+
+    // Create and store the connection promise
+    this.#connectionPromise = this.#doConnect();
+
+    try {
+      await this.#connectionPromise;
+      console.log(`✅ Connection attempt succeeded`);
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : 'Unknown connection error';
```
mcmire: Same as above, what do you think about:

``` suggestion
      const errorMessage = getErrorMessage(error);

File: packages/core-backend/src/WebsocketService.ts [hunk 20]
```diff
@@ -0,0 +1,1348 @@
...
+        this.#handleRequestTimeout();
+
+        reject(
+          new Error(`Request timeout after ${this.#options.requestTimeout}ms`),
+        );
+      }, this.#options.requestTimeout);
+
+      // Store in pending requests for response correlation
+      this.#pendingRequests.set(requestId, {
+        resolve: resolve as (value: unknown) => void,
```
mcmire, line 531: Instead of storing `resolve` and `reject` like this, we have a utility method in `@metamask/utils` that lets you create a deferred promise and get direct access to `resolve` and `reject` that you can call later: https://metamask.github.io/utils/latest/functions/createDeferredPromise.html

So you can say:

``` typescript
const { resolve, reject } = createDeferredPromise();
```

and then carry on with:

``` typescript
const timeout = setTimeout(() => {
  this.#pendingRequests.delete(requestId);
  console.warn(
    `🔴 Request timeout after ${this.#options.requestTimeout}ms - triggering reconnection`,
  );

  // Trigger reconnection on request timeout as it may indicate stale connection
  this.#handleRequestTimeout();

  reject(
    new Error(`Request timeout after ${this.#options.requestTimeout}ms`),
  );
}, this.#options.requestTimeout);

// ...
```

packages/core-backend/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+   * Optimized message parsing for mobile (reduces JSON.parse overhead)
+   *
+   * @param data - The raw message data to parse
+   * @returns Parsed message or null if parsing fails
+   */
+  #parseMessage(data: string): WebSocketMessage | null {
+    try {
+      return JSON.parse(data);
+    } catch {
+      // Fail fast on parse errors (mobile optimization)
```
mcmire, line 1125: Maybe we want to print a message here saying that JSON parsing failed?

packages/core-backend/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+      ) {
+        console.log(
+          `🔴 WebSocket disconnection detected - state: ${oldState} → ${newState}`,
+        );
+      }
+
+      // Publish connection state change event
+      try {
+        this.#messenger.publish(
+          'BackendWebSocketService:connectionStateChanged',
```
mcmire, line 1313: Should we use `SERVICE_NAME` here?

``` suggestion
          `${SERVICE_NAME}:connectionStateChanged`,

File: packages/backend-platform/src/WebsocketService.ts [hunk 23]
```diff
@@ -0,0 +1,1348 @@
...
+    }
+  }
+
+  /**
+   * Generates a unique message ID
+   *
+   * @returns Unique message identifier
+   */
+  #generateMessageId(): string {
+    return `${Date.now()}-${Math.random().toString(36).substring(2, 11)}`;
```
mcmire: What are your thoughts on using UUID v4 instead of generating a random string? We use this for network client IDs already, and perhaps other places too.

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+      this.#connectionPromise = null;
+    }
+  }
+
+  /**
+   * Internal method to perform the actual connection
+   *
+   * @returns Promise that resolves when connection is established
+   */
+  async #doConnect(): Promise<void> {
```
mcmire: If `#doConnect` calls `#establishConnection` internally, is this method necessary?

packages/backend-platform/src/WebsocketService.ts
```diff
@@ -0,0 +1,1348 @@
...
+  async sendMessage(message: ClientRequestMessage): Promise<void> {
+    if (this.#state !== WebSocketState.CONNECTED) {
+      throw new Error(`Cannot send message: WebSocket is ${this.#state}`);
+    }
+
+    try {
+      this.#ws.send(JSON.stringify(message));
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : 'Failed to send message';
```
mcmire: Similar as elsewhere

``` suggestion
      const errorMessage = getErrorMessage(error);

mcmire (commented):
Some more minor suggestions. I haven't done a full pass of the implementation so I may come back and make some more comments later but this is what I noticed for now.

Kriys94:
Note: Move Debouncer in the pooling directly

Kriys94:
For new Token we need to call "addToken" from TokenController. First check if this is part of allToken (imported tokens). As a second PR.

packages/backend-platform/src/AccountActivityService.ts
Kriys94: Thanks, good tip. I made the change. I wonder why this does not exist for events?

packages/assets-controllers/package.json
Kriys94: Yeah, I will create another PR for assets-controller changes

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

packages/backend-platform/src/AccountActivityService.ts
Kriys94: Done

packages/backend-platform/src/AccountActivityService.ts
Kriys94: Done

packages/backend-platform/src/AccountActivityService.ts
Kriys94: Done

packages/backend-platform/src/AccountActivityService.ts
mcmire: Typically events do not correlate with methods on controllers or services, so they need to be listed by hand anyway.

packages/backend-platform/src/AccountActivityService.ts
Kriys94: This is more a habit from someone not used to code in MM codebase, I find this more straightforward to inject WebSocketService as the dependency is easier to understand for IDEs (like understanding where functions are implemented). Here I know that the AccountActivityService logic fully depend on WebsocketService from the BackendPlatform package.

But I can switch to messenger if necessary

packages/backend-platform/src/WebsocketService.ts
Kriys94: Because #ws is set only when we successfully established the connection. In case of error #ws is undefined.

packages/core-backend/src/WebsocketService.ts
Kriys94: I've heard that `${SERVICE_NAME}:connectionStateChanged` is not recommended because it makes the search of the event or action more difficult to find

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

packages/backend-platform/src/WebsocketService.ts
Kriys94: Done

PR #6492, opened 2025-09-08
feat(account-tree-controller): add account group name uniqueness validation

PatrykLucka:
## Explanation

Currently there is no method for ensuring account group uniqeuness. This PR adds function that does that.

- Implemented a new private method `#assertAccountGroupNameIsUnique` to ensure account group names are unique across different groups.
- Updated `setAccountGroupName` to validate uniqueness before setting a name.
- Added tests to verify that the same name can be set for the same group and that duplicate names across different groups are prevented.

## Checklist

### DIFF
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -7,10 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add unique name validation for account groups to prevent duplicate group names ([#6492](https://github.com/MetaMask/core/pull/6492))
+  - `setAccountGroupName` now validates that group names are unique across all groups
+  - Added `isAccountGroupNameUnique` utility function in `group.ts` for checking name uniqueness
+  - Names are trimmed of leading/trailing whitespace before comparison to prevent accidental duplicates
+
 ### Changed
 
 - **BREAKING:** Remove support for `AccountsController:accountRenamed` event handling ([#6438](https://github.com/MetaMask/core/pull/6438))
 - Bump `@metamask/base-controller` from `^8.2.0` to `^8.3.0` ([#6465](https://github.com/MetaMask/core/pull/6465))
+- 
 
 ## [0.12.1]
 
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -10,15 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add unique name validation for account groups to prevent duplicate group names ([#6492](https://github.com/MetaMask/core/pull/6492))
-  - `setAccountGroupName` now validates that group names are unique across all groups
-  - Added `isAccountGroupNameUnique` utility function in `group.ts` for checking name uniqueness
-  - Names are trimmed of leading/trailing whitespace before comparison to prevent accidental duplicates
+  - `setAccountGroupName` now validates that group names are unique across all groups.
+  - Added `isAccountGroupNameUnique` utility function to check name uniqueness.
+  - Names are trimmed of leading/trailing whitespace before comparison to prevent accidental duplicates.
 
 ### Changed
 
 - **BREAKING:** Remove support for `AccountsController:accountRenamed` event handling ([#6438](https://github.com/MetaMask/core/pull/6438))
 - Bump `@metamask/base-controller` from `^8.2.0` to `^8.3.0` ([#6465](https://github.com/MetaMask/core/pull/6465))
-- 
 
 ## [0.12.1]
 
--- a/packages/account-tree-controller/src/AccountTreeController.test.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.test.ts
@@ -1949,6 +1949,82 @@ describe('AccountTreeController', () => {
         controller.state.accountWalletsMetadata[nonExistentWalletId],
       ).toBeUndefined();
     });
+
+    it('should allow setting the same name for the same group', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+
+      const wallets = controller.getAccountWalletObjects();
+      const groups = Object.values(wallets[0].groups);
+      const groupId = groups[0].id;
+
+      const customName = 'My Custom Group';
+
+      // Set the name first time - should succeed
+      controller.setAccountGroupName(groupId, customName);
+
+      // Set the same name again for the same group - should succeed
+      expect(() => {
+        controller.setAccountGroupName(groupId, customName);
+      }).not.toThrow();
+    });
+
+    it('should prevent setting duplicate names across different groups', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1, MOCK_HD_ACCOUNT_2],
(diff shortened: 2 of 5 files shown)


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -693,6 +693,23 @@ export class AccountTreeController extends BaseController<
...
+   */
+  #assertAccountGroupNameIsUnique(groupId: AccountGroupId, name: string): void {
+    for (const wallet of Object.values(this.state.accountTree.wallets)) {
+      for (const [currentGroupId, group] of Object.entries(wallet.groups)) {
+        if (currentGroupId !== groupId && group.metadata.name === name) {
+          throw new Error('Account group name already exists');
+        }
+      }
+    }
+  }
```
ccharly: Can you move this in `group.ts` and make it a function that takes the `AccountTreeControllerState` as argument?

The rational being that we could import it on the client side and apply this from the UI directly.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -693,6 +693,23 @@ export class AccountTreeController extends BaseController<
...
+   * Asserts that an account group name is unique across all groups.
+   *
+   * @param groupId - The account group ID to exclude from the check.
+   * @param name - The name to validate for uniqueness.
+   * @throws Error if the name already exists in another group.
+   */
+  #assertAccountGroupNameIsUnique(groupId: AccountGroupId, name: string): void {
+    for (const wallet of Object.values(this.state.accountTree.wallets)) {
+      for (const [currentGroupId, group] of Object.entries(wallet.groups)) {
+        if (currentGroupId !== groupId && group.metadata.name === name) {
```
ccharly: We can do this too, slightly "shorter" IMO:

```suggestion
      for (const group of Object.values(wallet.groups)) {
        if (group.id !== groupId && group.metadata.name === name) {
```

ccharly: Let's `strip` both `group.metadata.name` and `name` to avoid having leading/trailing whitespaces in the name (we don't want to "consider" them either when comparing each names IMO).

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1949,6 +1949,115 @@ describe('AccountTreeController', () => {
         controller.state.accountWalletsMetadata[nonExistentWalletId],
       ).toBeUndefined();
     });
+
+    it('should allow setting the same name for the same group', () => {
```
ccharly:
```suggestion
    it('allows setting the same name for the same group', () => {
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1949,6 +1949,115 @@ describe('AccountTreeController', () => {
...
+      // Set the name first time - should succeed
+      controller.setAccountGroupName(groupId, customName);
+
+      // Set the same name again for the same group - should succeed
+      expect(() => {
+        controller.setAccountGroupName(groupId, customName);
+      }).not.toThrow();
+    });
+
+    it('should prevent setting duplicate names across different groups', () => {
```
ccharly:
```suggestion
    it('prevents setting duplicate names across different groups', () => {
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1949,6 +1949,115 @@ describe('AccountTreeController', () => {
...
+      // Set name for first group - should succeed
+      controller.setAccountGroupName(groupId1, duplicateName);
+
+      // Try to set the same name for second group - should throw
+      expect(() => {
+        controller.setAccountGroupName(groupId2, duplicateName);
+      }).toThrow('Account group name already exists');
+    });
+
+    it('should ensure unique names when generating default names', () => {
```
ccharly:
```suggestion
    it('ensures unique names when generating default names', () => {
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1949,6 +1949,115 @@ describe('AccountTreeController', () => {
...
+
+      // All groups should have unique names by default
+      const names = groups.map((group) => group.metadata.name);
+      const uniqueNames = new Set(names);
+
+      expect(uniqueNames.size).toBe(names.length);
+      expect(names.every((name) => name.length > 0)).toBe(true);
+    });
+
+    it('should prevent duplicate names when comparing trimmed names', () => {
```
ccharly:
```suggestion
    it('prevents duplicate names when comparing trimmed names', () => {
```

packages/account-tree-controller/src/group.ts
```diff
@@ -84,3 +85,27 @@ export type AccountGroupObjectOf<GroupType extends AccountGroupType> = Extract<
...
+
+/**
+ * Asserts that an account group name is unique across all groups.
+ *
+ * @param state - The account tree controller state.
+ * @param groupId - The account group ID to exclude from the check.
+ * @param name - The name to validate for uniqueness.
+ * @throws Error if the name already exists in another group.
+ */
+export function assertAccountGroupNameIsUnique(
```
ccharly: WDYT about using boolean logic for this function actually? I think it would make it easier to use externally like:
```ts
if (isAccountGroupNameUnique(...)) {
  // Display error message on the UI.
}
```

Instead of having to deal with `try catch` to do this. WDYT?


Though, I would follow the `#assert...` pattern inside the controller. So adding a `#assertAccountGroupNameIsUnique` and throw in that method directly.

WDYT?

PatrykLucka: Sounds good, feels more consistent with how we do things in other places!

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,10 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add unique name validation for account groups to prevent duplicate group names ([#6492](https://github.com/MetaMask/core/pull/6492))
+  - `setAccountGroupName` now validates that group names are unique across all groups
+  - Added `isAccountGroupNameUnique` utility function in `group.ts` for checking name uniqueness
+  - Names are trimmed of leading/trailing whitespace before comparison to prevent accidental duplicates
```
```suggestion
- Add unique name validation for account groups to prevent duplicate group names ([#6492](https://github.com/MetaMask/core/pull/6492))
  - `setAccountGroupName` now validates that group names are unique across all groups.
  - Added `isAccountGroupNameUnique` utility function to check name uniqueness.
  - Names are trimmed of leading/trailing whitespace before comparison to prevent accidental duplicates.
```

packages/account-tree-controller/src/group.ts
```diff
@@ -84,3 +85,29 @@ export type AccountGroupObjectOf<GroupType extends AccountGroupType> = Extract<
...
+
+/**
+ * Checks if an account group name is unique across all groups.
+ *
+ * @param state - The account tree controller state.
+ * @param groupId - The account group ID to exclude from the check.
+ * @param name - The name to validate for uniqueness.
+ * @returns True if the name is unique, false otherwise.
+ */
+export function isAccountGroupNameUnique(
```
ccharly, line 97: We need to export it in the `index.ts` too

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,10 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 
 - **BREAKING:** Remove support for `AccountsController:accountRenamed` event handling ([#6438](https://github.com/MetaMask/core/pull/6438))
 - Bump `@metamask/base-controller` from `^8.2.0` to `^8.3.0` ([#6465](https://github.com/MetaMask/core/pull/6465))
+-
```
```suggestion
```

PR #6493, opened 2025-09-08
feat: added rewards controller and rewards data service messenger

satyajeetkolhapure:
## Explanation

This pull request adds Rewards Controller and messenger feature, which is used for silent login in the MetaMask Mobile app. The implementation enables automatic authentication for users when they switch accounts or unlock their wallet.

The RewardsController should:

Automatically authenticates a user through a silent signature from KeyringController if their EVM address owns a rewards profile
Skips auth for Solana and hardware wallets for now
Skips auth for 10 minutes of grace period

Note: This controller was part of metamask-mobile and moved to core as it can be used for metamask-extension too.
## Checklist


packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+const AUTH_GRACE_PERIOD_MS = 1000 * 60 * 10; // 10 minutes
+
+// Perps discount refresh threshold
+const PERPS_DISCOUNT_CACHE_THRESHOLD_MS = 1000 * 60 * 5; // 5 minutes
+
+// Season status cache threshold
+const SEASON_STATUS_CACHE_THRESHOLD_MS = 1000 * 60 * 1; // 1 minute
+
+// Referral details cache threshold
+const REFERRAL_DETAILS_CACHE_THRESHOLD_MS = 1000 * 60 * 10; // 10 minutes
```
GuillaumeRx: move to `constants.ts` ?

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+        total: seasonStatus.balance.total,
+        refereePortion: seasonStatus.balance.refereePortion,
+        updatedAt: seasonStatus.balance.updatedAt?.getTime(),
+      },
+      tier: tierState,
+      lastFetched: Date.now(),
+    };
+  }
+
+  constructor({
```
GuillaumeRx: Can we move the constructor before any method declaration ?

jbblanc (approved):

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+    try {
+      const selectedAccount = this.messagingSystem.call(
+        'AccountsController:getSelectedMultichainAccount',
+      );
+      await this.#performSilentAuth(selectedAccount);
+    } catch (error) {
+      // Silent failure - don't throw errors for background authentication
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+      if (errorMessage && !errorMessage?.includes('Engine does not exist')) {
```
GuillaumeRx: `!errorMessage?.includes('Engine does not exist')`  is a mobile only error, I'm not sure it belongs here

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+        ) {
+          const errorMessage = (signError as Error).message;
+          if (errorMessage.includes('controller is locked')) {
+            log('RewardsController: Keyring is locked, skipping silent auth');
+            return; // Exit silently when keyring is locked
+          }
+        }
+
+        throw signError;
+      }
```
GuillaumeRx: Can we avoid the nested `try...catch` blocks ?

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+      log('RewardsController: Silent auth successful');
+    } catch (error: unknown) {
+      // Handle 401 (not opted in) or other errors silently
+      if (error instanceof Error && error.message.includes('401')) {
+        // Not opted in
+      } else {
+        // Unknown error
+        subscription = null;
+        authUnexpectedError = true;
+      }
```
GuillaumeRx: Can we reverse this if statement to avoid the empty if ?

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+    );
+
+    // Try different encoding approaches to handle potential character issues
+    let hexMessage;
+    try {
+      // First try: direct toHex conversion
+      hexMessage = toHex(challengeResponse.message);
+    } catch (error) {
+      // Fallback: use Buffer to convert to hex if toHex fails
+      hexMessage = `0x${Buffer.from(challengeResponse.message, 'utf8').toString('hex')}`;
```
GuillaumeRx: Why do we need a fallback  ?

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+          'RewardsController: Failed to store subscription token:',
+          tokenResponse?.error || 'Unknown error',
+        );
+      }
+    }
+
+    // Update state with opt-in response data
+    // Update state with opt-in response data
+    this.update((state) => {
+      const caipAccount: CaipAccountId | null =
```
GuillaumeRx, line 1016: There's no need to define the type of the constant here if the functions returns that type

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+      if (!tokenResponse.success) {
+        log(
+          'RewardsController: Failed to store subscription token:',
+          tokenResponse?.error || 'Unknown error',
+        );
+      }
+    }
+
+    // Update state with opt-in response data
+    // Update state with opt-in response data
```
GuillaumeRx:
```suggestion
    // Update state with opt-in response data
```

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1150 @@
...
+      return false;
+    }
+
+    if (!code.trim()) {
+      return false;
+    }
+
+    if (code.length !== 6) {
+      return false;
+    }
```
GuillaumeRx: Can we do one statement here ?

packages/rewards-controller/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const controllerName = 'RewardsController'; // Default timeout for all API requests (10 seconds)
+export const DEFAULT_REQUEST_TIMEOUT_MS = 10000;
```
GuillaumeRx:
```suggestion
export const controllerName = 'RewardsController';

// Default timeout for all API requests (10 seconds)
export const DEFAULT_REQUEST_TIMEOUT_MS = 10000;
```

packages/rewards-controller/src/feature-flags.ts
GuillaumeRx, line 1: This could live in a private function inside the controller rather than defining it inside a separate file and having to create types for the messenger call function

packages/rewards-controller/src/messenger/RewardsControllerMessenger.ts
```diff
@@ -0,0 +1,97 @@
...
+  | AccountsControllerSelectedAccountChangeEvent
+  | KeyringControllerUnlockEvent;
+
+export type RewardsControllerMessenger = RestrictedMessenger<
+  typeof name,
+  RewardsControllerActions | AllowedActions,
+  RewardsControllerEvents | AllowedEvents,
+  AllowedActions['type'],
+  AllowedEvents['type']
+>;
```
GuillaumeRx: This should go in the controller file

packages/rewards-controller/src/messenger/RewardsControllerMessenger.ts
```diff
@@ -0,0 +1,97 @@
...
+      'RewardsDataService:logout',
+      'RewardsDataService:fetchGeoLocation',
+      'RewardsDataService:validateReferralCode',
+    ],
+    allowedEvents: [
+      'AccountsController:selectedAccountChange',
+      'KeyringController:unlock',
+    ],
+  });
+}
```
GuillaumeRx: This shouldn't live in the core repo but in the client repo

packages/rewards-controller/src/services/rewards-data-service.ts
```diff
@@ -0,0 +1,577 @@
...
+
+  readonly #environment: EnvironmentType;
+
+  readonly #getSubscriptionToken: GetSubscriptionToken;
+
+  constructor({
+    messenger,
+    fetch: fetchFunction,
+    version,
+    appType = 'mobile',
```
GuillaumeRx: Why do we default to mobile ?

packages/rewards-controller/src/services/rewards-data-service.ts
```diff
@@ -0,0 +1,577 @@
...
+    locale = 'en-US',
```
GuillaumeRx: Should we really have a default when this will be always set in the client ?

packages/rewards-controller/src/services/rewards-data-service.ts
```diff
@@ -0,0 +1,577 @@
...
+
+  readonly #rewardsApiUrl: string;
+
+  readonly #environment: EnvironmentType;
+
+  readonly #getSubscriptionToken: GetSubscriptionToken;
+
+  constructor({
+    messenger,
+    fetch: fetchFunction,
```
GuillaumeRx, line 143: Why do we have to pass fetch in the constructor ?

GuillaumeRx (changes requested):
In addition to the requested changes:

- please review the spacing inside the functions, it's a bit out of place

- There's a significant amount of logging, especially in `catch` blocks. I wonder if we can reduce the logging and throw regular errors instead.

packages/rewards-controller/src/feature-flags.ts
satyajeetkolhapure: Only issue I experienced when moving it inside the controller is mocking becomes difficult as there are many combinations of messenger mocking. I don't have strong opinion on keeping it separate file though. Test setup will become complex.

packages/rewards-controller/src/services/rewards-data-service.ts
satyajeetkolhapure: Actually the fetch is just a function for handing the fetch process and url is formed in a function makeRequest.
As in rewards case the fetch handling is similar for all calls we are sending a fetch implementation in the constructor.

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed (removed)

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
satyajeetkolhapure: fixed and added trim to code.length too

packages/rewards-controller/src/constants.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/messenger/RewardsControllerMessenger.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/messenger/RewardsControllerMessenger.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/services/rewards-data-service.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/services/rewards-data-service.ts
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1185 @@
...
+      );
+      return response.valid;
+    } catch (error) {
+      log(
+        'RewardsController: Failed to validate referral code:',
+        error instanceof Error ? error.message : String(error),
+      );
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
satyajeetkolhapure: fixed

packages/rewards-controller/src/RewardsController.ts
```diff
@@ -0,0 +1,1184 @@
...
+   * @param isHardwareAccount - Whether the account is a hardware wallet
+   * @returns boolean - True if silent auth should be skipped, false otherwise
+   */
+  #shouldSkipSilentAuth(
+    account: CaipAccountId,
+    address: string,
+    isHardwareAccount: boolean,
+  ): boolean {
+    // Skip for hardware and Solana accounts
+    if (isHardwareAccount || isSolanaAddress(address)) {
```
GuillaumeRx, line 464: There's no need to bring a package (`@solana/addresses`) to validate an address with `isSolanaAddress` as the CAIP-10 ID contains the chain ID that could be use to determine if it's a solana account.

packages/rewards-controller/src/feature-flags.ts
GuillaumeRx: You shouldn't be completely mocking the messenger but rather creating a real restricted messenger with mocked actions inside it

packages/rewards-controller/src/RewardsController.test.ts
```diff
@@ -0,0 +1,3745 @@
...
+    mockMessenger = {
+      subscribe: jest.fn(),
+      call: jest.fn(),
+      registerActionHandler: jest.fn(),
+      unregisterActionHandler: jest.fn(),
+      publish: jest.fn(),
+      clearEventSubscriptions: jest.fn(),
+      registerInitialEventPayload: jest.fn(),
+      unsubscribe: jest.fn(),
+    } as unknown as jest.Mocked<RewardsControllerMessenger>;
```
GuillaumeRx, line 143: You shouldn't be mocking completely the messenger but rather creating a real messenger with mocked actions inside it. You can look at other controllers to see how it's done.

packages/rewards-controller/src/RewardsController.test.ts
```diff
@@ -0,0 +1,3745 @@
...
+
+    // Reset feature flag to enabled by default
+    mockGetRewardsFeatureFlag.mockReturnValue(true);
```
GuillaumeRx, line 146: We should get rid of this mock and instead do a partial implementation of the feature flag action in the test messenger.

packages/rewards-controller/src/messenger/RewardsControllerMessenger.ts
GuillaumeRx, line 1: This can be deleted now that it exists in the controller file

packages/rewards-controller/src/services/rewards-data-service.test.ts
```diff
@@ -0,0 +1,1653 @@
...
+    createModuleLogger: jest.fn(() => logSpy),
+    __logSpy: logSpy,
+  };
+});
+
+const { __logSpy: logSpy } = jest.requireMock('../logger') as {
+  __logSpy: jest.Mock;
+};
+
+let mockMessenger: jest.Mocked<RewardsDataServiceMessenger>;
```
GuillaumeRx, line 37: We shouldn't mock the entire messenger.

packages/rewards-controller/src/services/rewards-data-service.test.ts
```diff
@@ -0,0 +1,1653 @@
...
+  });
+}
+
+// Helper to build service with injectable deps
+const buildService = ({
+  fetchImpl = jest
+    .fn()
+    .mockResolvedValue(
+      okJsonResponse({ sessionId: 'sess', subscription: { id: 'sub' } }),
+    ),
```
GuillaumeRx, line 82: We don't manually mock `fetch`, we rely on `nock` to provide fetch mocks.

packages/rewards-controller/src/services/rewards-data-service.test.ts
```diff
@@ -0,0 +1,1653 @@
...
+  rewardsApiUrl: string;
+}> = {}) => {
+  mockGetSubscriptionToken.mockResolvedValue({
+    success: true,
+    token: 'test-bearer-token',
+  });
+  mockMessenger = {
+    registerActionHandler: jest.fn(),
+    call: jest.fn(),
+  } as unknown as jest.Mocked<RewardsDataServiceMessenger>;
```
GuillaumeRx, line 105: We don't directly mock the messenger

packages/rewards-controller/src/services/rewards-data-service.ts
GuillaumeRx: There's no need to pass fetch as it's already present as a global in the versions of node.js we use. I don't fully get your comment, it doesn't answer my question.

packages/rewards-controller/src/services/rewards-data-service.ts
```diff
@@ -0,0 +1,577 @@
...
+
+  readonly #appType: 'mobile' | 'extension';
+
+  readonly #locale: string;
+
+  readonly #rewardsApiUrl: string;
+
+  readonly #environment: EnvironmentType;
+
+  readonly #getSubscriptionToken: GetSubscriptionToken;
```
GuillaumeRx, line 139: Where does this function comes from ? If it's available through a messenger action please use the messenger rather than passing it as a parameter.

packages/rewards-controller/src/RewardsController.ts
GuillaumeRx: Not fixed

packages/rewards-controller/src/RewardsController.ts
GuillaumeRx: not fixed

packages/rewards-controller/src/types.ts
```diff
@@ -0,0 +1,607 @@
...
+  account: CaipAccountId;
+  hasOptedIn?: boolean;
+  subscriptionId: string | null;
+  lastCheckedAuth: number;
+  lastCheckedAuthError: boolean;
+  perpsFeeDiscount: number | null;
+  lastPerpsDiscountRateFetched: number | null;
+};
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
Gudahtt, line 413: This comment seems to be unnecessary, there is no lint error here

packages/rewards-controller/src/types.ts
```diff
@@ -0,0 +1,607 @@
...
+export type RewardsControllerState = {
```
Gudahtt, line 414: Could you add descriptions to these properties? It's hard to tell what they are for

PR #6501, opened 2025-09-09
feat: shield-controller: check signature coverage

matthiasgeihs:
## Explanation

This PR adds support for validating signature coverage to the ShieldController.

## Checklist


packages/shield-controller/src/ShieldController.ts
```diff
@@ -169,13 +183,54 @@ export class ShieldController extends BaseController<
       this.#transactionControllerStateChangeHandler,
       (state) => state.transactions,
     );
+
+    this.messagingSystem.subscribe(
+      'SignatureController:stateChange',
+      this.#signatureControllerStateChangeHandler,
```
lwin-kyaw, line 189: Just wondering. This, `#signatureControllerStateChangeHandler` will be triggered again if the state changes in case of user signing/approving the request, right?

We might not wanna re-request the coverage in such cases?

packages/shield-controller/src/backend.ts
```diff
@@ -76,22 +83,46 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
+
+    const reqBody: InitSignatureCoverageCheckRequest = {
+      chainId: signatureRequest.chainId,
+      data: signatureRequest.messageParams.data,
+      from: signatureRequest.messageParams.from,
+      method: signatureRequest.type,
+      origin: signatureRequest.messageParams.origin,
+    };
+
+    const { coverageId } = await this.#initCoverageCheck(
```
lwin-kyaw, line 109: Please note that we can't always except `coverageId` to be in the response, with the `200` status.

Please check [the swagger](https://ruleset-engine.dev-api.cx.metamask.io/docs#/Signature%20Coverage%20Controller/SignatureCoverageController_initiateSignatureCoverage_v1), as it has updated recently. (please select `schema` tab under the `200` response, not the example value)

Screenshot 2025-09-11 at 6 56 50 PM

lwin-kyaw: The sample cases of 200 responses without `coverageId` ~
- where origin is localhost or empty string (requests directly from metamask wallet)
- when providing invalid chain id values

In such cases, we don't proceed to process the rulesets, we straight away return the `Unknown` result from the server.

lwin-kyaw: This case also applies to `transaction-coverage`

packages/shield-controller/src/ShieldController.ts
matthiasgeihs: The coverage is only requested when the SignatureRequest is added. You can check the handler code.

packages/shield-controller/src/backend.ts
matthiasgeihs: Why don't we throw an error in that case?
So basically now the init endpoint can also return a result? Kind of confusing. This creates unnecessary edge cases I believe.

matthiasgeihs: Can we please change the API backend. I do see several issues with this change. The expectation so far is that if we have a coverage result, either for signatures, or for transactions, it always has the same format. If we don't have that, a lot of complexity is added to the whole system. (Currently we can just treat the results within the same data structure. Then we might need separate data structure. Also, missing the `coverageId` might be problematic because it means we can't look up the corresponding backend calls.)

PR #6527, opened 2025-09-09
feat(multichain-account-service): add `:walletStatusChange` event + mutex for concurrent mutable operations

ccharly:
## Explanation

Wallet operations are not behind a mutex. This ensure that only one mutable operation can be executed at a time.

Also publish new event, so we know in which state a wallet is.

> [!NOTE]
> A follow-up PR will be using those events to add a new status in the `account-tree-controller`.

## References

N/A

## Checklist


packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -335,24 +417,20 @@ export class MultichainAccountWallet<
...
   }
 
   /**
    * Align all multichain account groups.
    */
   async alignGroups(): Promise<void> {
-    if (this.#isAlignmentInProgress) {
-      return; // Prevent concurrent alignments
-    }
-
```
ccharly, line 372: Actually, removed the old logic mainly cause this could be misleading.

If an alignement was in progress, calling `alignGroups` would have returned right away, so you needed to check `getIsAlignmentInProgress` to know if something was still running.

With the new `#withLock` mechanisms. All those operations will naturally be "queued" until the lock got released. So, we won't be running those concurrently, but more sequentially instead.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -335,24 +417,20 @@ export class MultichainAccountWallet<
    * @returns True if alignment is in progress, false otherwise.
    */
   getIsAlignmentInProgress(): boolean {
-    return this.#isAlignmentInProgress;
+    return this.#status === MultichainAccountWalletStatus.AlignmentInProgress;
   }
```
ccharly: I really wonder if we need this method anymore?

Any opinion @fabiobozzo?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -39,6 +41,40 @@ export type AccountDiscoveryMetrics = {
...
+   * Alignment is in progress for this wallet. Account groups will be
+   * automatically updated based on the active account providers.
+   */
+  AlignmentInProgress = 'alignment-in-progress',
+  /**
+   * An on-going operation (creating/deleting) is in progress for this
+   * wallet. Account groups will either be created or deleted during
+   * this operation.
+   */
+  OperationInProgress = 'operation-in-progress',
```
ccharly: I struggle to find a good name for this one... Maybe we should also split create/delete into 2 different status 🤔

(deleting is not possible for now though)

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -39,6 +41,40 @@ export type AccountDiscoveryMetrics = {
...
+ *
+ * Those status are used to report in which "state" the wallet is currently
+ * in. All of those operations cannot run concurrently, thus, the wallet
+ * cannot have multiple status at once.
+ */
+export enum MultichainAccountWalletStatus {
+  /**
+   * The wallet is not initialized yet.
+   */
+  Uninitialized = 'uninitalized',
```
fabiobozzo:
```suggestion
  Uninitialized = 'uninitialized',
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -154,6 +199,38 @@ export class MultichainAccountWallet<
...
+    try {
+      this.#status = status;
+      this.#messenger.publish(
+        'MultichainAccountService:walletStatusChange',
+        this.id,
+        this.#status,
+      );
+      return await operation();
+    } finally {
+      this.#status = MultichainAccountWalletStatus.Ready;
```
fabiobozzo: Don't we need a `Error` state?

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo: Indeed it must be removed. Along with `MultichainAccountServiceGetIsAlignmentInProgressAction`. I guess it was a good idea initially, then you build a more solid/extensible "status"? No more polling needed but event-based if I understand correctly @ccharly

fabiobozzo: Note for the changelog: it's a breaking change, isn't it? Although nobody is using that yet? 🤔

fabiobozzo (dismissed):

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Yes, I'll mark it as breaking too, but since we're in <1.0.0, that's fine!

Thanks for your input 💪

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Good question 🤔 the main problem I can see with this is... How to get back into the `Ready` state if that happens? 😅

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo: "CrudOperation"? "Mutation" ? No clue either 😅

ccharly: I actually went with `in-progress:create-accounts`, mainly because we don't have "delete operations".

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: After thinking about it, I don't think we need any error state for the moment.

packages/multichain-account-service/src/index.ts
```diff
@@ -8,10 +8,8 @@ export type {
   MultichainAccountServiceGetMultichainAccountGroupsAction,
   MultichainAccountServiceCreateMultichainAccountGroupAction,
   MultichainAccountServiceCreateNextMultichainAccountGroupAction,
-  MultichainAccountServiceGetIsAlignmentInProgressAction,
   MultichainAccountServiceSetBasicFunctionalityAction,
-  MultichainAccountServiceMultichainAccountGroupCreatedEvent,
-  MultichainAccountServiceMultichainAccountGroupUpdatedEvent,
+  MultichainAccountServiceWalletStatusChangeEvent,
```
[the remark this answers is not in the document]
ccharly: That's an error indeed, I'll fix this.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -162,12 +167,44 @@ export class MultichainAccountWallet<
...
+   * @returns The operation's result.
+   * @throws {Error} If the wallet is already running a mutable operation.
+   */
+  async #withLock<Return>(
+    status: MultichainAccountWalletStatus,
+    operation: () => Promise<Return>,
+  ) {
+    const release = await this.#lock.acquire();
+    try {
+      this.#status = status;
```
fabiobozzo, line 192: Don't we need to validate it's a valid status string, at runtime?

PR #6547, opened 2025-09-10
feat: intent-based swaps support

dawnseeker8:
## References
Relates to https://consensyssoftware.atlassian.net/browse/BETR-5

## Checklist

---

> [!NOTE]
> Introduces intent-based swap/bridge flow end-to-end.
> 
> - **Types/validators**: Add `Intent`/`IntentOrder` schemas and optional `quote.intent`; extend `StatusTypes` with `SUBMITTED`
> - **BridgeStatusController**: New `submitIntent` action; create synthetic TC tx, write history (`originalTransactionId`, `srcTxHashes`), and poll intent order status via new `IntentApiImpl`; map intent statuses to bridge/TC statuses; start/stop polling and track events (skip when `featureId`)
> - **API**: Implement `IntentApiImpl` with `submitIntent`/`getOrderStatus` and status mapping
> - **Tests**: Comprehensive unit tests for intent submission, polling, and validators; snapshot updates; jest coverage thresholds adjusted
> - **Misc**: Export intent types; changelogs updated; .gitignore adds Emacs patterns
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87e2d3acc6a6078b8e5f2b7abd77c20e58e1c9b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `AccountTrackerController` now triggers an immediate balance refresh when the keyring is unlocked, instead of waiting for the next polling cycle ([#7585](https://github.com/MetaMask/core/pull/7585))
+
 ## [95.0.0]
 
 ### Added
--- a/packages/assets-controllers/src/AccountTrackerController.test.ts
+++ b/packages/assets-controllers/src/AccountTrackerController.test.ts
@@ -427,6 +427,29 @@ describe('AccountTrackerController', () => {
     );
   });
 
+  it('should refresh balances when keyring is unlocked', async () => {
+    await withController(
+      {
+        selectedAccount: ACCOUNT_1,
+        listAccounts: [ACCOUNT_1],
+      },
+      async ({ controller, messenger }) => {
+        const refreshSpy = jest.spyOn(controller, 'refresh');
+
+        // Lock the keyring first
+        messenger.publish('KeyringController:lock');
+
+        // Clear any previous calls
+        refreshSpy.mockClear();
+
+        // Unlock the keyring - should trigger refresh
+        messenger.publish('KeyringController:unlock');
+
+        expect(refreshSpy).toHaveBeenCalled();
+      },
+    );
+  });
+
   describe('refresh', () => {
     it('does not refresh when fetching is disabled', async () => {
       const expectedState = {
--- a/packages/assets-controllers/src/AccountTrackerController.ts
+++ b/packages/assets-controllers/src/AccountTrackerController.ts
@@ -372,6 +372,10 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
 
     this.messenger.subscribe('KeyringController:unlock', () => {
       this.#isLocked = false;
+      const networkClientIds = this.#getNetworkClientIds();
+      this.refresh(networkClientIds).catch((error) => {
+        console.error('Error refreshing balances after keyring unlock:', error);
+      });
     });
 
     this.messenger.subscribe('KeyringController:lock', () => {


tsconfig.packages.json
```diff
@@ -14,6 +14,7 @@
      */
     "paths": {
       "@metamask/base-controller/next": ["../base-controller/src/next"],
+      "@metamask/intent-manager": ["../Intent-manager/dist/index"],
```
dawnseeker8: without this line, the `yarn build` in root directory will failed, which i don't know why, may be relative to @metamask/intent-manager didnt published to npm registry and still locall version.

Akaryatrh:
@metamaskbot publish-preview

packages/bridge-controller/src/utils/validators.ts
```diff
@@ -192,6 +192,42 @@ export const StepSchema = type({
...
+  protocol: IntentProtocolSchema,
+  order: IntentOrderSchema,
+  // Optional metadata to aid submission/routing
+  settlementContract: optional(HexAddressSchema),
+  relayer: optional(HexAddressSchema),
+});
+
+// Export types for use in other modules
+export type IntentOrder = Infer<typeof IntentOrderSchema>;
+export type Intent = Infer<typeof IntentSchema>;
```
micaelae: Can you remove these? they are already being exported by the `types.ts` file

packages/bridge-controller/src/utils/validators.ts
```diff
@@ -192,6 +192,42 @@ export const StepSchema = type({
...
+  buyAmount: optional(TruthyDigitStringSchema),
+  // Optional owner/from for convenience when building domain/message
+  from: optional(HexAddressSchema),
+});
+
+export const IntentSchema = type({
+  protocol: IntentProtocolSchema,
+  order: IntentOrderSchema,
+  // Optional metadata to aid submission/routing
+  settlementContract: optional(HexAddressSchema),
```
micaelae: Minor: use JSDoc in the schemas so the Intellisense can show the descriptions

packages/bridge-controller/src/utils/validators.ts
```diff
@@ -192,6 +192,42 @@ export const StepSchema = type({
 
 const RefuelDataSchema = StepSchema;
 
+// Allow digit strings for amounts/validTo for flexibility across providers
+const DigitStringOrNumberSchema = union([TruthyDigitStringSchema, number()]);
+
+// Intent support (e.g., CoW Swap EIP-712 order signing)
+const IntentProtocolSchema = enums(['cowswap']);
+
```
micaelae: I'd make this just a generic string so it's flexible when you add other providers

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -667,12 +702,215 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+
+      const intentApi = new IntentApiImpl(
+        this.#config.customBridgeApiBaseUrl,
+        this.#fetchFn,
+      );
+      const intentOrder = (await intentApi.getOrderStatus(
+        orderId,
+        providerName,
+        srcChainId.toString(),
+      )) as IntentOrder;
```
micaelae: The pattern we currently follow is to validate the response after fetching (see `fetchBridgeTxStatus`). This ensures that the data passed to and persisted by the clients are valid.

Suggestions for the intent-api:
1. add type schemas for IntentOrder and status to the validators.ts module
2. the intents module should validate responses after submission and after fetching status
3. write IntentApi as util functions to stay consistent with existing api calls

packages/bridge-status-controller/src/utils/intent-api.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.#baseUrl = baseUrl;
+    this.#fetchFn = fetchFn;
+  }
+
+  async submitIntent(params: IntentSubmissionParams): Promise<unknown> {
+    const endpoint = `${this.#baseUrl}/submitOrder`;
+    try {
+      const response = await this.#fetchFn(endpoint, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
```
micaelae: please add the client Id headers to submitOrder and getOrderstatus so the backend can have granular metrics (see `getClientIdHeader`)

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1586,298 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+      // Update transaction with proper initial status based on intent order
+      const statusUpdatedTxMeta = {
+        ...intentTxMeta,
+        status: initialTransactionStatus,
+      };
+
+      this.#updateTransactionFn(
+        statusUpdatedTxMeta,
+        `BridgeStatusController - Initial intent order status: ${intentOrder.status}`,
+      );
```
micaelae: Did the transactions team approve of this approach? Updating transactions is not recommended (we have a backlog item to remove our current usage) and can potentially get the tx in an invalid state

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -204,6 +213,16 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       'TransactionController:transactionFailed',
       ({ transactionMeta }) => {
         const { type, status, id } = transactionMeta;
+
+        // Skip intent transactions - they have their own tracking via CoW API
+        // Skip intent transactions - they have their own tracking via CoW API
+        if (
+          (transactionMeta as { swapMetaData?: { isIntentTx?: boolean } })
+            .swapMetaData?.isIntentTx
+        ) {
```
micaelae: This block of code's purpose is to stop status polling, update the `txHistory` controller state and to publish UnifiedSwap metrics. I'd recommend not skipping it unless there is separate metrics tracking for intentd. Moreover, we don't use the legacy `swapMetadata` field anymore since most data we need is in `txHistory`

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1586,298 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+    signature: string;
+    accountAddress: string;
+  }): Promise<Pick<TransactionMeta, 'id' | 'chainId' | 'type' | 'status'>> => {
+    const { quoteResponse, signature, accountAddress } = params;
+
+    // Build pre-confirmation properties for error tracking parity with submitTx
+    const account = this.messenger.call(
+      'AccountsController:getAccountByAddress',
+      accountAddress,
+    );
```
micaelae:
```suggestion
    const account = this.#getMultichainSelectedAccount(accountAddress);
```

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -667,12 +702,215 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+
+    // Map intent order status to bridge status using enum values
+    let statusType: StatusTypes;
+    const isComplete = [
+      IntentOrderStatus.CONFIRMED,
+      IntentOrderStatus.COMPLETED,
+    ].includes(intentOrder.status);
+    const isFailed = [
+      IntentOrderStatus.FAILED,
+      IntentOrderStatus.EXPIRED,
```
micaelae, line 766: Do you need to show a custom UI when the status is EXPIRED? If the status is transformed to to FAILED, the UI treats it the same way as a failed order

oscarwroche: No, we show it as failed

packages/bridge-status-controller/src/bridge-status-controller.ts
oscarwroche: I think we are tracking these transactions via CowSwap

packages/bridge-status-controller/src/bridge-status-controller.ts
oscarwroche: Got rid of it

packages/bridge-status-controller/src/bridge-status-controller.ts
a-breslin: @oscarwroche @Akaryatrh - we should publish UnifiedSwap events from this integration to ensure we have the same degree of visibility into user behaviours that we have for all other transactions.

micaelae: +1 to Andrew's point

@oscarwroche the tracking I'm referring to is for client-side metrics, which are separate from backend and CowSwap's data

packages/transaction-controller/src/helpers/PendingTransactionTracker.ts
```diff
@@ -667,7 +667,9 @@ export class PendingTransactionTracker {
       (tx) =>
         tx.status === TransactionStatus.submitted &&
         !tx.verifiedOnBlockchain &&
-        !tx.isUserOperation,
+        !tx.isUserOperation &&
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        !(tx as any).swapMetaData?.isIntentTx, // Exclude intent transactions from pending tracking
```
micaelae: Wouldn't excluding intent txs make the signed transaction appear stuck in "submitted" state in the clients?

The `orderStatus` api tracks the intent's overall progress (including off-chain processes), while the TransactionController tracks each signed transaction's confirmation status. Moreover, the bridge-status-controller waits for `TransactionController:transactionConfirmed` to be published in order to start polling for `orderStatus`; I think this change prevents the tx tracker from getting to that point

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,291 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+        from: accountAddress,
+        to:
+          intent.settlementContract ??
+          '0x9008D19f58AAbd9eD0D60971565AA8510560ab41', // Default settlement contract
+        data: `0x${orderUid.slice(-8)}`, // Use last 8 chars of orderUid to make each transaction unique
+        value: '0x0',
+        gas: '0x5208', // Minimal gas for display purposes
+        gasPrice: '0x3b9aca00', // 1 Gwei - will be converted to EIP-1559 fees if network supports it
+        skipInitialGasEstimate: true,
+      };
```
micaelae, line 1699: What does this transaction do? Is it a noop transaction that costs gas? If I'm understanding the usage correctly, it's purpose is to add a tx to history + TransactionController, which the frontend can use to show the intent order's status in the activity list

oscarwroche: This is the transaction's purpose indeed. I'll add a comment to clarify what it's for.

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1239,17 +1239,15 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
-    const maxGasLimit = toHex(transactionParams.gas ?? 0);
-
     // If txFee is provided (gasIncluded case), use the quote's gas fees
     // Convert to hex since txFee values from the quote are decimal strings
     if (txFee) {
       return {
         maxFeePerGas: toHex(txFee.maxFeePerGas ?? 0),
         maxPriorityFeePerGas: toHex(txFee.maxPriorityFeePerGas ?? 0),
-        gas: maxGasLimit,
+        gas: transactionParams.gas ? toHex(transactionParams.gas) : undefined,
```
Akaryatrh: @oscarwroche Can you store it back in a `const`?
```
const maxGasLimit = transactionParams.gas ? toHex(transactionParams.gas) : undefined;
```

Also can you explain why is it fixing the gas issue we had?

oscarwroche: Fixed

oscarwroche: It's because `maxGasLimit` was sometimes null which became zero along the way

packages/bridge-status-controller/src/utils/intent-api.ts~
```diff
@@ -0,0 +1,71 @@
...
+export type IntentSubmissionParams = {
+  srcChainId: string;
+  quoteId: string;
+  signature: string;
+  order: unknown;
+  userAddress: string;
+  aggregatorId: string;
+};
+
+export type IntentApi = {
```
micaelae: Can you remove this? Looks like a duplicate file

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+            await this.#waitForTxConfirmation(approvalTxId, {
+              timeoutMs: 30_000, // 30 seconds instead of 5 minutes
+              pollMs: 3_000, // Poll less frequently to avoid rate limits
+            });
+          } catch (error) {
+            // Log but don't throw - continue with intent order submission
+            console.warn(
+              'Approval confirmation failed, continuing with intent submission:',
+              error,
+            );
```
micaelae: If the approval fails, can the intent succeed?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -823,6 +1061,54 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+        'TransactionController:getState',
+      );
+      const meta = transactions.find((tx: TransactionMeta) => tx.id === txId);
+
+      if (meta) {
+        // Treat both 'confirmed' and 'finalized' as success to match TC lifecycle
+
+        if (
+          meta.status === TransactionStatus.confirmed ||
+          // Some environments move directly to finalized
```
micaelae: Can you link the code that sets status to `finalized`? Can't seem to find it so this seems unecessary

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+   * @param params.signature - Hex signature produced by eth_signTypedData_v4
+   * @param params.accountAddress - The EOA submitting the order
+   * @returns A lightweight TransactionMeta-like object for history linking
+   */
+  submitIntent = async (params: {
+    quoteResponse: QuoteResponse<TxData | string> & QuoteMetadata;
+    signature: string;
+    accountAddress: string;
+  }): Promise<Pick<TransactionMeta, 'id' | 'chainId' | 'type' | 'status'>> => {
+    const { quoteResponse, signature, accountAddress } = params;
```
micaelae, line 1601: Please add a `stopPollingForQuotes` call here to prevent the UI from updating during submission

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+
+      const chainId = quoteResponse.quote.srcChainId;
+
+      const submissionParams = {
+        srcChainId: chainId.toString(),
+        quoteId: intentQuote.id,
+        signature,
+        order: intentQuote.metadata.order,
+        userAddress: accountAddress,
+        aggregatorId: 'cowswap',
```
micaelae: Can this value be derived from the quote instead of hardcoding?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+    validUntil: number;
+    metadata: {
+      order: unknown;
+      settlementContract: string;
+      chainId: ChainId;
+      bridgeQuote: QuoteResponse<TxData | string> & QuoteMetadata;
+    };
+  } {
+    return {
+      id: `bridge-${Date.now()}`,
```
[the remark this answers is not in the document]
micaelae: According to bridge-api docs, this should be the requestId: https://bridge.dev-api.cx.metamask.io/docs#:~:text=SubmitIntentOrderRequestDto

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+      );
+
+      const chainId = quoteResponse.quote.srcChainId;
+
+      const submissionParams = {
+        srcChainId: chainId.toString(),
+        quoteId: intentQuote.id,
+        signature,
+        order: intentQuote.metadata.order,
+        userAddress: accountAddress,
```
micaelae, line 1658: I think this would be more readable if the params are directly plugged in specially bc Most fields returned by `convertBridgeQuoteToIntentQuote` are unused, based on https://bridge.dev-api.cx.metamask.io/docs#:~:text=SubmitIntentOrderRequestDto


```suggestion
      const { srcChainId, intent, requestId } = quoteResponse.quote;
      const submissionParams = {
        srcChainId,
        quoteId: requestId,
        signature,
        order: intent.order,
        userAddress: accountAddress,
```

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+
+      // This is a synthetic transaction whose purpose is to be able
+      // to track the order status via the history
+      const intentTransactionParams = {
+        chainId: formatChainIdToHex(chainId),
+        from: accountAddress,
+        to:
+          intent.settlementContract ??
+          '0x9008D19f58AAbd9eD0D60971565AA8510560ab41', // Default settlement contract
+        data: `0x${orderUid.slice(-8)}`, // Use last 8 chars of orderUid to make each transaction unique
```
micaelae, line 1695: Can QuoteResponse.trade.data be used as the synthetic tx? Instead of rebuilding the params

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1585,294 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+      case IntentOrderStatus.CONFIRMED:
+      case IntentOrderStatus.COMPLETED:
+        return TransactionStatus.confirmed;
+      case IntentOrderStatus.FAILED:
+      case IntentOrderStatus.EXPIRED:
+        return TransactionStatus.failed;
+      default:
+        return TransactionStatus.submitted;
+    }
+  }
```
micaelae: Please move this to the intent-api module to reduce the line count of the controller

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -667,12 +695,222 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+    } else {
+      statusType = StatusTypes.UNKNOWN;
+    }
+
+    // Extract transaction hashes from intent order
+    const txHash = intentOrder.txHash ?? '';
+    // Check metadata for additional transaction hashes
+    const metadataTxHashes = Array.isArray(intentOrder.metadata.txHashes)
+      ? intentOrder.metadata.txHashes
+      : [];
```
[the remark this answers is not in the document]
oscarwroche: No, it's handled with the `txHash` variable

packages/bridge-status-controller/src/bridge-status-controller.ts
oscarwroche: The comment above explains it a bit, basically the intent submission can be made even if the allowance tx isn't mined yet. We are trying to reach out to @khanti42 to better understand this.

oscarwroche: Ok so we're removing this :)

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1399,6 +1398,21 @@ export class TransactionController extends BaseController<
             traceContext: context,
           }),
       );
+    } else if (
```
matthewwalsh0: Converting `gasPrice` to the EIP-1559 properties should already be done in `gas-fees.ts`.

We also intentionally encapsulate all that complex gas fee logic in the util to minimise coupling in this file.

packages/transaction-controller/src/helpers/PendingTransactionTracker.ts
```diff
@@ -695,7 +695,9 @@ export class PendingTransactionTracker {
...
```
matthewwalsh0: I wasn't aware we still used this metadata since unified bridging and creating all swaps and bridges from the `BridgeStatusController`?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1284,6 +1583,226 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+      const { transactionMeta: txMetaPromise } = await this.#addTransactionFn(
+        intentTransactionParams,
+        {
+          origin: 'metamask',
+          actionId: generateActionId(),
+          requireApproval: false,
+          networkClientId,
+          type: transactionType,
+          skipInitialGasEstimate: true,
+          swaps: {
```
matthewwalsh0: This is a legacy pattern we want to avoid, as we can instead reference this controller state from the client, rather than coupling it to the `TransactionController`.

packages/transaction-controller/src/helpers/PendingTransactionTracker.ts
matthewwalsh0: Rather than relying on metadata, would these new transactions not warrant an alternate `transactionType` that we could exclude here?

Plus if these are EVM transactions submitted by the user's wallet, do we not want to check if they are confirmed here for consistency?

Do we currently rely on an API to confirm their success? Could we do both?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3113,6 +3127,31 @@ export class TransactionController extends BaseController<
           }
         }
 
+        // For intent-based transactions (e.g., CoW intents) that are not meant to be
```
matthewwalsh0: We already have a mechanism to skip publish through the `beforePublish` hook.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3113,6 +3127,31 @@ export class TransactionController extends BaseController<
...
+        // published on-chain by the TransactionController, skip the approve/publish flow.
+        // These are tracked externally and should not be signed or sent.
```
matthewwalsh0: If they are entirely external, then is the only benefit to display them in the activity view?

If they share none of the transaction lifecycle, perhaps it's cleaner just for the clients to read data from the `BridgeStatusController` directly?

My understanding from the code is these are instead quotes that are signed via `eth_signTypedData_v4` then processed entirely externally.

packages/transaction-controller/src/helpers/PendingTransactionTracker.ts
micaelae: This is the only place we use `swapMetaData` and we plan to remove this tech debt after v1. Unified swaps and bridges don't use the metadata field anymore

> Do we currently rely on an API to confirm their success? Could we do both?

The added transaction is basically a dummy/noop that is not meant to be submitted on-chain. The new `submitIntent` flow adds it as a placeholder for showing the intent order's status in the UI. Eventually when the bridge-api returns the order's final status we update the txMeta with the tx hash

I think the emulated updateTransaction and addTransaction handlers would be a better fit for this use-case (similar to userOperation). However I just found out about those handlers this week so we haven't tried it yet

packages/transaction-controller/src/TransactionController.ts
micaelae: That's correct @matthewwalsh0 . When the intent order is fulfilled it also shows up as an incoming transaction in the activity log

I was considering replacing this logic with emulated add/update transaction calls. See https://github.com/MetaMask/core/pull/6547#discussion_r2673185233. Would you recommend that or having clients read data from `BridgeStatusController` directly?

packages/transaction-controller/src/helpers/PendingTransactionTracker.ts
OGPoyraz: We don't need `any` here

```suggestion
        !tx.swapMetaData?.isIntentTx, // Exclude intent transactions from pending tracking
```

PR #6553, opened 2025-09-10
fix: Strip `accessToken` from `AuthenticationController` state logs

Gudahtt:
## Explanation

The metadata property was recently added to the
`AuthenticationController` in #6470, but we forgot to strip out the `accessToken`. This is currently being stripped from state logs in both clients.

The metadata was updated to strip out this token. Additionally, some improvements were added to the metadata snapshot tests to ensure that all properties are represented in the fixture state.

The `@metamask/utils` package was added as a dependency because the new metadata state deriver uses `Json` in its signature, so that type is now used in the exported types for this package.

## References

This is an amendment to #6470

## Checklist


packages/profile-sync-controller/CHANGELOG.md
Gudahtt, line 1: It didn't seem relevant to mention stripping out `accessToken` here, since it's not really a change from the previous release (the state log metadata isn't released yet).

I did mention the new dependency though

owencraston (approved):

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -49,7 +50,27 @@ const metadata: StateMetadata<AuthenticationControllerState> = {
...
   },
   srpSessionData: {
-    includeInStateLogs: true,
+    // Remove access token from state logs
+    includeInStateLogs: (srpSessionData) => {
+      // Using non-null assertion here to assert that it's not undefined, because if it was, this
+      // wouldn't get called.
+      // The type includes `| undefined` only because we don't yet use `strictOptionalTypes`
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      return Object.entries(srpSessionData!).reduce<Record<string, Json>>(
```
fabiobozzo:
```suggestion
      if (!srpSessionData) return {};
      
      return Object.entries(srpSessionData).reduce<Record<string, Json>>(
```
safer?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -49,7 +50,27 @@ const metadata: StateMetadata<AuthenticationControllerState> = {
...
+        (sanitizedSrpSessionData, [key, value]) => {
+          const token: Partial<(typeof value)['token']> = {
+            ...value.token,
+          };
+          delete token.accessToken;
```
fabiobozzo: Isn't that the same as 
```
const { accessToken, ...token } = value.token;
```
?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.test.ts
```diff
@@ -543,6 +552,8 @@ describe('metadata', () => {
     const controller = new AuthenticationController({
       messenger: createMockAuthenticationMessenger().messenger,
       metametrics: createMockAuthMetaMetrics(),
+      // Set `expiresIn` to an arbitrary number so that it stays consistent between test runs
+      state: mockSignedInState({ expiresIn: 1_000 }),
```
fabiobozzo, line 556: Nit, consider extracting `MOCK_EXPIRES_IN = 1_000`

Gudahtt: I almost did this, but in the end I figured it was better to have the actual mock value directly in the test, because the value itself is what we're looking for in the inline snapshot.

Usually in tests with common mocks, the mock values are less relevant. But it was useful for readability in this case.

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
Gudahtt: Hmmm. Yes, it is. Admittedly I coped this directly from the pre-existing implementation ([here](https://github.com/MetaMask/metamask-extension/blob/2417268f7904e830307b3a26aa812c29f2e09cf3/app/scripts/lib/state-utils.ts#L113)) and made minimal changes to address type errors and lint/formatting.

I do like your suggestion here much better, but I'm tempted to leave this as-is just because it's approved already.

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
Gudahtt: Yeah, I almost did this as well. But then we'd have another untested branch in the code that can't be reached. It would be an error if this ever returned an empty object, despite it matching the type (i.e. it's still never supposed to happen).

Here's the version I almost used:

```
if (!srpSessionData) {
  return srpSessionData;
}
```
This version was "safer" because if we ever updated this property to allow `null`, it would still work. Whereas my version would fail.

But.... this didn't work, because then it gave a type error about returning `undefined` which isn't valid JSON.

We do plan on enabling `strictOptionalTypes`, and that would eliminate the need for a non-null assertion here.

Gudahtt: Actually... now that I'm saying that outloud, I think something like this would be best:

```
if (!srpSessionData === null || srpSessionData === undefined) {
  return null;
}
```

This branch eliminates the type error, and it still works if we later update this property to allow `null`. It's still an unreachable branch, but that seems OK. I'll go with this option.

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
Gudahtt: I took the suggestion since I was making another change that would dismiss reviews anyway. Updated here: 65d23cbb4

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
Gudahtt: Done here: f29d7d16b

packages/profile-sync-controller/CHANGELOG.md
Gudahtt: 🤦 That quick, it was released. I'll have to add something now.

Gudahtt: OK, updated again

PR #6555, opened 2025-09-10
feat: Add new metadata to `EarnController`

Gudahtt:
## Explanation

The new metadata properties `includeInStateLogs` and `usedInUi` have been added to the `EarnController`.

## References

Fixes https://github.com/MetaMask/core/issues/6517

## Checklist

### DIFF
--- a/packages/earn-controller/src/EarnController.test.ts
+++ b/packages/earn-controller/src/EarnController.test.ts
@@ -2587,13 +2587,35 @@ describe('EarnController', () => {
     it('includes expected state in state logs', async () => {
       const { controller } = await setupController();
 
-      expect(
-        deriveStateFromMetadata(
-          controller.state,
-          controller.metadata,
-          'includeInStateLogs',
-        ),
-      ).toMatchInlineSnapshot(`
+      const derivedState = deriveStateFromMetadata(
+        controller.state,
+        controller.metadata,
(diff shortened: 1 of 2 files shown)


packages/earn-controller/src/EarnController.test.ts
```diff
@@ -2566,4 +2566,271 @@ describe('EarnController', () => {
...
+        controller.metadata,
+        'persist',
+      );
+
+      // Compare `pooled_staking` separately to minimize size of snapshot
+      const {
+        pooled_staking: derivedPooledStaking,
+        ...derivedStateWithoutPooledStaking
+      } = derivedState;
+      expect(derivedPooledStaking).toStrictEqual({
```
Gudahtt, line 2692: This is a bit.... janky. It's not as easy to read as I want it to be.

 I've hard-coded the expected values here that get setup in `setupController` from the mocks, because it was _way_ too large for an inline snapshot. And since those mocks are setup in the root `describe` block, there isn't an easy way to customize them here.

Ideally the constructor wouldn't do this initialization at all, leaving it to an `init` function. And ideally we'd be able to customize it in `setupController` rather than it being set for all tests. But either of those changes would be way too large in scope for this PR.

Please feel free to suggest alternatives, if you can think of any that would be more readable.

PR #6556, opened 2025-09-11
Add method to decode permission to GatorPermissionsController

jeffsmale90:
## Explanation

- **Why change?**
When signing a permission, MetaMask must present the permission in a way that is easily comprehensible to the user. Based on the design [shared here](https://www.notion.so/metamask-consensys/SignTypedData-with-Metadata-Specification-22bf86d67d688023be67e2ee06e3a56a#22bf86d67d688023be67e2ee06e3a56a), the permission must be decoded from the `eth_signTypedData_v4` request.

This PR adds that functionality to the `GatorPermissionsController`.
 
- **What’s the solution?**

Introduce `decodePermissionFromPermissionContextForOrigin` to `GatorPermissionsController` and register a new action handler. It:
  - Validates the caller origin matches `permissionsProviderSnapId` (throws `OriginNotAllowedError` otherwise).
  - Resolves enforcer contract addresses per chain via `@metamask/delegation-deployments`. Throws if contracts are missing for the chain.
  - Identifies the permission type from caveats via `identifyPermissionByEnforcers`, extracts expiry and permission-specific data via `getPermissionDataAndExpiry`, and builds a `DecodedPermission` via `reconstructDecodedPermission`.

- **Non-obvious pieces:**
  - Added `@metamask/delegation-core` for terms helpers like `createTimestampTerms` and `createNativeTokenStreamingTerms`, and `ROOT_AUTHORITY`.
  - Added `@metamask/delegation-deployments` to source enforcer addresses and `CHAIN_ID`.
  - Moved the decoding logic into a subfolder, to make it easier to export internal functionality from `utils.ts` to facilitate low level unit testing, without polluting the interface that is exposed to the `GatorPermissionsController` with these functions.
  
## References

Architecture design https://www.notion.so/metamask-consensys/SignTypedData-with-Metadata-Specification-22bf86d67d688023be67e2ee06e3a56a#22bf86d67d688023be67e2ee06e3a56a

## Checklist

- [ ] I've communicated my changes to consumers by updating changelogs for packages I've changed, highlighting breaking changes as necessary


packages/gator-permissions-controller/package.json
```diff
@@ -49,6 +49,8 @@
   "dependencies": {
     "@metamask/7715-permission-types": "^0.3.0",
     "@metamask/base-controller": "^8.3.0",
+    "@metamask/delegation-core": "^0.2.0-rc.1",
```
jeffsmale90: 0.2.0 stable release is just around the corner, so we can probably update this before merging.

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -490,4 +520,82 @@ export default class GatorPermissionsController extends BaseController<
...
+   * @param args.delegation - delegation representing the permission.
+   * @param args.metadata - metadata included in the request.
+   * @param args.metadata.justification - the justification as specified in the request metadata.
+   * @param args.metadata.origin - the origin as specified in the request metadata.
+   *
+   * @returns A decoded permission object suitable for UI consumption and follow-up actions.
+   * @throws If the origin is not allowed, the context cannot be decoded into exactly one delegation,
+   * or the enforcers/terms do not match a supported permission type.
+   */
+  public async decodePermissionFromPermissionContextForOrigin({
```
V00D00-child, line 544: I don't see why `decodePermissionFromPermissionContextForOrigin()` should be exposed on the `GatorPermissionController` when it does not mutate the controller's state. It does read `permissionsProviderSnapId`, but that value is also passed to the MM client environment. 

We should consider decoupling and exposing it as a until function from the `@metamask/gator-permissions-controller` package, similar to how we expose `serializeGatorPermissionsMap` and `deserializeGatorPermissionsMap`.

MM clients can call the `decodePermissionFromPermissionContextForOrigin` function without submitting a request to the background and import the function directly from `@metamask/gator-permissions-controller`.

jeffsmale90: This is actually being called from the `SignatureController` rather than the front end. It's a critical aspect of this design that a signature request for a delegation is rejected if a permission is unable to be decoded. Therefore, we've added the decoding step into the signature request preparation pipeline.

Therefore in order to ensure the `SignatureController` can call this method without a tight coupling to the `GatorPermissionsController` exposing it via the `Messenger` is the right approach.

packages/gator-permissions-controller/package.json
jeffsmale90: Done 🎉

PR #6557, opened 2025-09-11
chore: add whitelist to utils to ignore warnings for specific networks

Julink-eth:
## Explanation

Add a list of networks that will be treated as whitelisted not to display the warning messages in the clients (Extension/Mobile) when adding them as custom networks.
This list is starting with HyperEVM network.

## References

Related to https://github.com/MetaMask/metamask-mobile/pull/19167
Related to https://github.com/MetaMask/metamask-extension/pull/35609

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "550.0.0",
+  "version": "551.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add `queryAllAccounts` parameter support to `AccountTrackerController.refresh()`, `AccountTrackerController._executePoll()`, and `TokenBalancesController.updateBalances()` for flexible account selection during balance updates ([#6600](https://github.com/MetaMask/core/pull/6600))
 - Bump `@metamask/utils` from `^11.4.2` to `^11.8.0` ([#6588](https://github.com/MetaMask/core/pull/6588))
 
 ## [75.0.0]
--- a/packages/assets-controllers/src/AccountTrackerController.test.ts
+++ b/packages/assets-controllers/src/AccountTrackerController.test.ts
@@ -166,7 +166,7 @@ describe('AccountTrackerController', () => {
             listAccounts: [ACCOUNT_1, ACCOUNT_2],
           },
           async ({ controller, refresh }) => {
-            await refresh(clock, ['mainnet']);
+            await refresh(clock, ['mainnet'], true);
 
             expect(controller.state).toStrictEqual({
               accountsByChainId: {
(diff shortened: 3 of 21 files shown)


packages/controller-utils/src/constants.ts
```diff
@@ -139,6 +139,19 @@ export const BUILT_IN_NETWORKS = {
...
+ * This list is used not to display the warnings for those networks.
+ */
+export const WHITELISTED_NETWORKS = {
+  // HyperEVM
+  '0x3e7': {
+    name: 'HyperEVM',
+    symbol: 'HYPE',
+    rpcUrl: 'https://rpc.hyperliquid.xyz',
+  },
+};
```
mcmire: A few thoughts: 

- Is there a more descriptive name for this constant we can use, one that alludes to the intended purpose? My thought is that "whitelist" can refer to a lot of different things.
- Is there a more inclusive term we can use than "whitelist"?
- Given that this constant supports some logic in the clients, should we update the JSDoc to explain how we intend on using this constant?
- Should we use `as const` and assign a type to this (i.e., make this a "quasi-enum")?

Here is a suggestion based on these notes:

``` suggestion
/**
 * Both Extension and Mobile have logic to dissuade users from adding
 * networks that share chain IDs with other networks; if that happens
 * then those users will see warnings. Sometimes, however, we want to
 * get around that. This mapping contains networks that should bypass
 * validation.
 */
export const ALLOWED_NETWORKS_WITH_DUPLICATE_CHAINS = {
  // HyperEVM uses the same chain ID as Wanchain
  '0x3e7': {
    name: 'HyperEVM',
    symbol: 'HYPE',
    rpcUrl: 'https://rpc.hyperliquid.xyz',
  },
};
```

packages/controller-utils/src/constants.ts
```diff
@@ -139,6 +139,22 @@ export const BUILT_IN_NETWORKS = {
...
+ * validation.
+ */
+export const ALLOWED_NETWORKS_WITH_DUPLICATE_CHAINS = {
+  // HyperEVM uses the same chain ID as Wanchain
+  '0x3e7': {
+    name: 'HyperEVM',
+    symbol: 'HYPE',
+    rpcUrl: 'https://rpc.hyperliquid.xyz',
+  },
+};
```
mcmire: Had another thought. Perhaps we want to be more generic with the name in case the validation changes later?

``` suggestion
/**
 * When a user adds a custom network to MetaMask, we perform some basic
 * validations on the network. For instance, usually a network cannot share the
 * same chain as another. In some cases, however, we want to allow networks that
 * would normally be invalid. This mapping contains networks that should bypass
 * validation.
 */
export const NETWORKS_BYPASSING_VALIDATION = {
  // HyperEVM uses the same chain ID as Wanchain
  '0x3e7': {
    name: 'HyperEVM',
    symbol: 'HYPE',
    rpcUrl: 'https://rpc.hyperliquid.xyz',
  },
};
```

packages/controller-utils/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add a whitelist `WHITELISTED_NETWORKS` to allow clients to ignore warning messages for specific networks. ([#6557](https://github.com/MetaMask/core/pull/6557))
```
mcmire: Should we update this as well?

``` suggestion
- Add constant `NETWORKS_BYPASSING_VALIDATION` to allow clients to ignore warning messages for specific networks. ([#6557](https://github.com/MetaMask/core/pull/6557))
```

PR #6560, opened 2025-09-11
chore: bump multichain accounts deps

ccharly:
## Explanation

Bumping accounts dependencies and adapting new `account-api` and `keyring-api` breaking changes.

## References

N/A

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "542.0.0",
+  "version": "544.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/assets-controllers/package.json
+++ b/packages/assets-controllers/package.json
@@ -95,7 +95,7 @@
     "@metamask/preferences-controller": "^19.0.0",
     "@metamask/providers": "^22.1.0",
     "@metamask/snaps-controllers": "^14.0.1",
-    "@metamask/transaction-controller": "^60.2.0",
+    "@metamask/transaction-controller": "^60.3.0",
     "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.191",
     "@types/node": "^16.18.54",
--- a/packages/bridge-controller/package.json
+++ b/packages/bridge-controller/package.json
@@ -73,7 +73,7 @@
     "@metamask/remote-feature-flag-controller": "^1.7.0",
     "@metamask/snaps-controllers": "^14.0.1",
     "@metamask/superstruct": "^3.1.0",
-    "@metamask/transaction-controller": "^60.2.0",
+    "@metamask/transaction-controller": "^60.3.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
--- a/packages/bridge-status-controller/package.json
+++ b/packages/bridge-status-controller/package.json
@@ -63,7 +63,7 @@
     "@metamask/gas-fee-controller": "^24.0.0",
     "@metamask/network-controller": "^24.1.0",
     "@metamask/snaps-controllers": "^14.0.1",
-    "@metamask/transaction-controller": "^60.2.0",
+    "@metamask/transaction-controller": "^60.3.0",
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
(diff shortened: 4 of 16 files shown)


packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -467,14 +472,8 @@ export class MultichainAccountWallet<
...
-
-    const discoveredAccounts: Record<string, number> = {};
-    for (const context of providerContexts) {
-      const providerName = context.provider.getName();
-      discoveredAccounts[providerName] = context.count;
-    }
+    await this.alignAccounts();
 
-    return discoveredAccounts;
+    return providerContexts.flatMap((context) => context.accounts);
```
ccharly, line 477: We now only return discovered accounts. If we need to extract metrics from this, we can still rely on the `account.type` to know if it's a Solana/Bitcoin/EVM accounts that got discovered.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -360,7 +365,7 @@ export class MultichainAccountWallet<
   /**
    * Align all multichain account groups.
    */
-  async alignGroups(): Promise<void> {
+  async alignAccounts(): Promise<void> {
```
ccharly, line 368: Renamed to follow the `<operation>Accounts` pattern. Also, this function does re-align accounts (create missing one if needed)

But it WILL NOT create missing **groups** if there's any.

Thus, it's more accounts related.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -400,14 +405,14 @@ export class MultichainAccountWallet<
    *
    * @returns The discovered accounts for each provider.
    */
-  async discoverAndCreateAccounts(): Promise<AccountDiscoveryMetrics> {
+  async discoverAccounts(): Promise<Account[]> {
```
ccharly, line 408: Same here. Just following the `<operation>Accounts` pattern.

Also the return type got changed, [see this for more details](https://github.com/MetaMask/core/pull/6560/files#r2339866837).

packages/assets-controllers/src/MultichainBalancesController/MultichainBalancesController.test.ts
```diff
@@ -44,7 +44,7 @@ const mockBtcAccount = {
   },
   scopes: [BtcScope.Testnet],
   options: {},
-  methods: [BtcMethod.SendBitcoin],
```
ccharly, line 47: This method no longer exists. So we just use all methods from the `enum` now.

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -360,7 +365,7 @@ export class MultichainAccountWallet<
   /**
    * Align all multichain account groups.
```
mathieuartu: Should we modify this description then?

packages/multichain-account-service/src/providers/SolAccountProvider.test.ts
```diff
@@ -256,7 +256,7 @@ describe('SolAccountProvider', () => {
     // Simulate one discovered account at the requested index.
     mocks.handleRequest.mockReturnValue([MOCK_SOL_DISCOVERED_ACCOUNT_1]);
 
-    const created = await provider.discoverAndCreateAccounts({
+    const created = await provider.discoverAccounts({
```
mathieuartu: Nit: should we change `created` to `discovered` for consistency?

packages/multichain-account-service/package.json
```diff
@@ -77,7 +77,7 @@
     "webextension-polyfill": "^0.12.0"
   },
   "peerDependencies": {
-    "@metamask/account-api": "^0.9.0",
+    "@metamask/account-api": "^0.12.0",
```
cryptodev-2s, line 80: Nit: could we add a changelog for this peer dependency bump ?

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Add `multichainAccountGroupCreated` event emitted from wallet level when new groups are created.
   - Add `multichainAccountGroupUpdated` event emitted from wallet level when groups are synchronized.
 
+### Changed
+
+- **BREAKING:** Rename `MultichainAccountWallet.discoverAndCreateAccounts` to `MultichainAccountWallet.discoverAccounts` ([#6560](https://github.com/MetaMask/core/pull/6560))
```
cryptodev-2s: Nit: as `AccountProviderWrapper`, `BaseBip44AccountProvider` and `SnapAccountProvider` are also exported at the index level. Should we add a changelog entry for the same renaming ?

ccharly: That makes total sense too yes! I'll add those too. Thanks!

PR #6561, opened 2025-09-11
Release 543.0.0

matthewwalsh0:
Minor release of `@metamask/transaction-controller`.

### DIFF
--- a/packages/eip-5792-middleware/CHANGELOG.md
+++ b/packages/eip-5792-middleware/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Bump `@metamask/transaction-controller` from `^60.2.0` to `^60.3.0` ([#6561](https://github.com/MetaMask/core/pull/6561))
+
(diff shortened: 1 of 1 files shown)


packages/eip-5792-middleware/package.json
```diff
@@ -49,7 +49,7 @@
   "dependencies": {
     "@metamask/eth-json-rpc-middleware": "^17.0.1",
     "@metamask/superstruct": "^3.1.0",
-    "@metamask/transaction-controller": "^60.2.0",
+    "@metamask/transaction-controller": "^60.3.0",
```
cryptodev-2s, line 52: Nit: Could you add a changelog entry for this bump

matthewwalsh0: Specifically in this package, or all 8? 😅 

We haven't done this in the past? If it's needed, it should surely be automated by `create-release-branch`?

cryptodev-2s: Yes, only this package since it’s a dependency bump, all the others are just dev dependency bumps.

matthewwalsh0: Good point, but could still be automated I assume, to save some cumulative time in future.

PR #6567, opened 2025-09-11
fix(accounts-controller): fire `:accountAdded` before `:selectedAccountChange` on `KeyringController:stateChange`

ccharly:
## Explanation

We were firing `:selectedAccountChange` before firing `:accountAdded` which does have undesired side-effect in the `AccountTreeController`.

Mainly because the tree will try to get the associated selected account group, but this account group does not exist yet (since it's using `:accountAdded` to create the account groups).

This is more like a temporary solution, and `#update` could be slightly refactored to a more robust solution later on.

## References

N/A

## Checklist

### DIFF
--- a/packages/accounts-controller/CHANGELOG.md
+++ b/packages/accounts-controller/CHANGELOG.md
@@ -18,6 +18,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `@metamask/keyring-internal-api` from `^8.1.0` to `^9.0.0` ([#6560](https://github.com/MetaMask/core/pull/6560))
 - Bump `@metamask/eth-snap-keyring` from `^16.1.0` to `^17.0.0` ([#6560](https://github.com/MetaMask/core/pull/6560))
 
+### Fixed
+
+- Now publish `:accountAdded` before `:selectedAccountChange` on `KeyringController:stateChange` ([#6567](https://github.com/MetaMask/core/pull/6567))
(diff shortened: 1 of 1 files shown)


packages/accounts-controller/src/AccountsController.ts
```diff
@@ -833,63 +833,71 @@ export class AccountsController extends BaseController<
...
-    for (const account of diff.added) {
-      this.messagingSystem.publish('AccountsController:accountAdded', account);
-    }
+        for (const account of diff.added) {
+          this.messagingSystem.publish(
+            'AccountsController:accountAdded',
+            account,
+          );
+        }
+      },
```
ccharly, line 899: This new callback will get executed prior to the account selection. This solves the issue during onboarding when the keyring creates its initial account (which triggers a `:stateChange`).

With this change, we will be firing `:accountAdded` before `:selectedAccount`, thus allowing the account-tree-controller to first create its group before trying to mirror the selected account group.

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -833,63 +833,71 @@ export class AccountsController extends BaseController<
       added: [] as InternalAccount[],
     };
 
-    this.#update((state) => {
-      const { internalAccounts } = state;
+    this.#update(
```
ccharly, line 836: Watchout, this call now takes 2 callbacks:
1. The actual update to run
2. The pre-step before selecting the new account that can result from the state update

PR #6571, opened 2025-09-11
feat(account-tree-controller): add `wallet.status` support

ccharly:
## Explanation

We now listen to `:walletStatusChange` and update the tree accordingly.

This state can be used to track which operation is running on a given wallet.

## References

N/A

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3811,33 +3811,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@metamask/multichain-account-service@npm:^0.8.0":
-  version: 0.8.0
-  resolution: "@metamask/multichain-account-service@npm:0.8.0"
-  dependencies:
-    "@metamask/base-controller": "npm:^8.3.0"
-    "@metamask/eth-snap-keyring": "npm:^17.0.0"
-    "@metamask/keyring-api": "npm:^21.0.0"
-    "@metamask/keyring-internal-api": "npm:^9.0.0"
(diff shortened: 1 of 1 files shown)


packages/account-tree-controller/src/wallet.ts
```diff
@@ -39,6 +41,7 @@ type IsAccountWalletObject<
   Type extends {
     type: AccountWalletType;
     id: AccountWalletId;
+    status: string; // Has to be refined by the type extending this base type.
```
gantunesr, line 44: why not use an `enum` for the status values?

gantunesr: Ah ok, I see this is defined in `AccountWalletStatus` so the comment you left is to adjust the type here. Should it be a `TODO:`?

ccharly: So.. We wanted to experiment a bit with string literal unions with @danroc. Because `MultichainAccountWalletStatus` would normally "extend" `AccountWalletStatus`, but we cannot do this with standard `enum`. However, a union is "extensible", and the "common type" used here for this is `string`.

Each of the typed account wallets we have (`AccountWallet{Entropy,Snap,Keyring}Object`) defines their own set of status (the `AccountWalletEntropyObject` uses a `MultichainAccountWalletStatus` which has more possible values while the others use `AccountWalletStatus`).

Those are the "refined" status I'm referring to.

Maybe I should have added a bit more comment here too sorry 😅

PR #6575, opened 2025-09-11
Allow skipping lookupNetwork call in initializeProvider

mcmire:
## Explanation

Currently, the extension patches `network-controller` so that `initializeProvider` does not call `lookupNetwork`. Because NetworkController gets initialized in the background process before the UI process is initialized and because `lookupNetwork` makes a request to the globally selected network, it was found that for a user installing MetaMask and going through onboarding, we could share their IP address with third parties before they had a chance to complete. That was rightly flagged as a privacy violation, so the call to `lookupNetwork` was disabled.

Now, long-term, `initializeProvider` shouldn't even exist (we can do half of what it does in the constructor, and clients can already call `lookupNetwork` separately). So at the moment the API for NetworkController is not very ideal. However, we don't want to keep having to patch `network-controller`.

For the time being, then, this commit adds a `lookupNetwork` option to `initializeProvider`. This allows clients to disable the initial network request more straightforwardly.

## References

- See https://github.com/MetaMask/metamask-extension/blob/e2cc47ce5d5c5cb8c1bd1e9c958bd4bb0576d8e5/.yarn/patches/%40metamask-network-controller-npm-24.0.1-d913bfdf67.patch for the existing patch.
- See https://github.com/MetaMask/metamask-extension/pull/23005 for the original motivation behind the patch (and related issues).

## Checklist


packages/network-controller/src/NetworkController.ts
```diff
@@ -1554,13 +1554,33 @@ export class NetworkController extends BaseController<
...
 
   /**
-   * Ensures that network clients for Infura and custom RPC endpoints have been
-   * created. Then, consulting state, initializes and establishes the currently
-   * selected network client.
-   */
-  async initializeProvider() {
+   * Creates proxies for accessing the global network client and its block
+   * tracker. You must call this method in order to use
+   * `getProviderAndBlockTracker` (or its replacement,
```
cryptodev-2s, line 1559: Nit: Should we also remove this deprecated method? It has been marked as deprecated for a while, and we could include its removal in the upcoming breaking changes (In case we can create an issue for that or juste make the removal in a follow up PR)

Mrtenz: It's still used in the extension, I'm not sure how easy it is to replace it there.

mcmire: True. And besides, I didn't know how many breaking changes we wanted to include in the next release — if we wanted to just restrict it to messenger upgrades or include more. Also, `getSelectedNetworkClient` is a bit silly. I think it was intended to replace `getProviderAndBlockTracker`, but it sounds like it ought to return a network client object but it actually returns proxies 🤦🏻  So I think there's some more API cleanup we ought to do as well. Maybe we don't worry about it for now.

PR #6596, opened 2025-09-13
Feat/subscription pricing calculation expose

tuna1207:
## Explanation

- Expose `getTokenApproveAmount` method
- Change enum to constant for client import

## Checklist


packages/subscription-controller/src/types.ts
matthiasgeihs, line 1: Can you add a note in the issue description why you changed enum to const? This is helpful for others to understand why this change was done.

packages/subscription-controller/src/types.ts
```diff
@@ -1,49 +1,55 @@
...
+export const PAYMENT_TYPES = {
+  byCard: 'card',
+  byCrypto: 'crypto',
+} as const;
 
-export enum PaymentType {
-  byCard = 'card',
-  byCrypto = 'crypto',
-}
+export type PaymentType = (typeof PAYMENT_TYPES)[keyof typeof PAYMENT_TYPES];
```
matthiasgeihs, line 14: You can simplify this to:
`export type PaymentType = 'card' | 'crypto';`
Same for the other enum style type definitions.

Or is there a benefit of writing it the way you did that I don't see?

tuna1207: @matthiasgeihs oh it's so that the type follow the value of `PAYMENT_TYPES` correctly without needed to duplicate each of them
and be can use the const value instead of typing string

chaitanyapotti (dismissed):

PR #6601, opened 2025-09-15
fix(account-tree-controller): address group names consistency bug

fabiobozzo:
## Explanation

**CRITICAL FIX**: Resolves a severe multi-wallet account naming bug that caused duplicate, inconsistent, and confusing account group names across different seed phrases (SRPs).

### Current Problem
Users importing multiple seed phrases experienced catastrophic naming issues:
- **Duplicate names**: Multiple "Account 2", "Account 3" within the same wallet
- **Missing "Account 1"**: Some wallets would start at "Account 2" 
- **Cross-wallet confusion**: Second SRP incorrectly showing "Account 3" when it should be "Account 1"
- **Inconsistent after restart**: Names would change unpredictably on app restart

### Root Cause
The account group naming logic had critical flaws:

1. **Broken alphabetical sorting**: Used `Object.keys(wallet.groups).sort()` which caused conflicts and duplicates in multi-wallet scenarios
2. **Unsafe entropy access**: Could throw TypeError when accessing `group.metadata.entropy.groupIndex` without null checks
3. **Negative index bug**: Empty wallets would get `-1` index, causing invalid "Account 0" names

### Solution
Implemented a robust fix with comprehensive improvements:

1. **Per-wallet sequential numbering**: Each wallet maintains independent "Account 1", "Account 2", etc. numbering with dynamic pattern detection that works with any naming convention ("Account N", "Imported Account N", etc.)
2. **Smart conflict resolution**: Universal conflict detection with `while` loop that finds next available unique name, handles user renames, and ensures unique sequential names
3. **Entropy-aware indexing**: For multichain groups, prioritizes actual `groupIndex` from entropy metadata when available with safe null checks
4. **Starting point optimization**: Begins name checking at `wallet.groups.length` to minimize iterations and improve efficiency
5. **AutoHandleConflict integration**: Added optional `autoHandleConflict` parameter to `setAccountGroupName()` for automated conflict resolution with suffix generation (`"Account 1 (2)"`, `"Account 1 (3)"`, etc.) - ready for Backup & Sync integration

## References

Fixes: 
- [MUL-835](https://consensyssoftware.atlassian.net/browse/MUL-835)
- [MUL-826](https://consensyssoftware.atlassian.net/browse/MUL-826)

## Checklist

[MUL-835]: https://consensyssoftware.atlassian.net/browse/MUL-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/account-tree-controller/src/wallet.ts
```diff
@@ -23,13 +23,17 @@ import type { UpdatableField, ExtractFieldValues } from './type-utils.js';
 export type AccountTreeWalletPersistedMetadata = {
   /** Custom name set by user, overrides default naming logic */
   name?: UpdatableField<string>;
+  /** Next account number to use for default naming (Account 1, Account 2, etc.) */
+  nextAccountNumber?: number;
```
ccharly: We probably want to compute that from the number of groups we have inside the wallet?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,47 +394,99 @@ export class AccountTreeController extends BaseController<
...
+        );
       }
 
-      // Use computed name first, then fallback to default naming if empty
-      group.metadata.name =
-        rule.getComputedAccountGroupName(typedGroup) ||
-        rule.getDefaultAccountGroupName(groupIndex);
+      // Persist the assigned name to ensure consistency across restarts
+      // This prevents the name from changing on subsequent inits
+      this.update((state) => {
```
ccharly: We probably want to call `setAccountGroupName` here instead? This way, if we need to handle name conflict logic (like adding a suffix), this would be done directly by this function.

Or at least, have some common implementation between the existing `setAccountGroupName` and maybe a new method that would take care of "automated names". WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,47 +394,99 @@ export class AccountTreeController extends BaseController<
...
+   * @param wallet - The wallet containing the group.
+   * @param group - The group to generate a name for.
+   * @param rule - The rule for this wallet type.
+   * @returns The generated default name.
+   */
+  #generateDefaultGroupName(
+    wallet: AccountWalletObject,
+    group: AccountGroupObject,
+    rule: Rule<AccountWalletType, AccountGroupType>,
+  ): string {
```
ccharly: I think, we should enforce the `N` from `Account N` to be associated with the group index of this group.

If it's already taken for whatever reason, then we add a suffix ` (2)` maybe.

We can check this internally, but we had a proposal for this already IIRC.

packages/account-tree-controller/src/wallet.ts
fabiobozzo: Now using `group.metadata.entropy.groupIndex` directly, which is more reliable isn't it?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Removed. This made the extension crash in fact. I simplified it to avoid state mutations during tree building

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: My fix now is supposed to ensure `Account N` is always associated with the actual `groupIndex` - no more alphabetical sorting confusion, wdyt?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,30 +397,39 @@ export class AccountTreeController extends BaseController<
...
     // Apply persisted name if available (including empty strings)
-    if (persistedMetadata?.name !== undefined) {
-      group.metadata.name = persistedMetadata.name.value;
+    if (persistedGroupMetadata?.name !== undefined) {
+      group.metadata.name = persistedGroupMetadata.name.value;
     } else if (!group.metadata.name) {
       // Get the appropriate rule for this wallet type
       const rule = this.#getRuleForWallet(wallet);
       const typedWallet = wallet as AccountWalletObjectOf<typeof wallet.type>;
       const typedGroup = typedWallet.groups[group.id] as AccountGroupObject;
```
ccharly: Can you use this instead for `typedGroup`:

```ts
      const typedGroup = typedWallet.groups[group.id] as AccountGroupObjectOf<
        typeof group.type
      >;
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,30 +397,39 @@ export class AccountTreeController extends BaseController<
...
-        groupIndex = 0;
+      // Calculate group index for default naming
+      let groupIndex: number;
+
+      // For entropy-based multichain groups, use the actual groupIndex from metadata
+      // instead of calculating from alphabetical sort position
+      if (
+        group.type === AccountGroupType.MultichainAccount &&
+        'entropy' in group.metadata
+      ) {
```
ccharly: With my previous suggestion, we should not need this anymore:

```suggestion
      if (group.type === AccountGroupType.MultichainAccount) {
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,30 +397,39 @@ export class AccountTreeController extends BaseController<
...
+        const sortedGroupIds = Object.keys(wallet.groups).sort();
+        groupIndex = sortedGroupIds.indexOf(group.id);
+
+        // Defensive fallback: if group.id is not found in sortedGroupIds (should never happen
+        // in normal operation since we iterate over wallet.groups), use index 0 to prevent
+        // passing -1 to getDefaultAccountGroupName which would result in "Account 0"
+        /* istanbul ignore next */
+        if (groupIndex === -1) {
+          groupIndex = 0;
+        }
```
ccharly: I still think the `sort` could failed with other kind of wallets, like:
- Create account with address: `0xbbb` -> `Account 1`
- Create a 2nd account with address: `0xaaa` -> `Account 1` (since the `indexOf` will now return `0` for this new account after sorting)

No?

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **CRITICAL**: Fix multi-wallet account naming bug causing duplicate and inconsistent names across wallets ([#6601](https://github.com/MetaMask/core/pull/6601))
+  - Each wallet now has independent sequential numbering (Account 1, Account 2, etc.)
+  - Names are immediately persisted to ensure consistency across app restarts
+  - Entropy groups use their actual groupIndex instead of broken alphabetical sorting
+  - Prevents issues like "Account 2, Account 2, Account 3..." and missing "Account 1"
```
```suggestion
  - Each wallet now has independent sequential numbering (Account 1, Account 2, etc.).
  - Names are immediately persisted to ensure consistency across app restarts.
  - Entropy groups use their actual group index instead of alphabetical sorting.
  - Prevents issues like "Account 2, Account 2, Account 3..." and missing "Account 1".
```

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: @ccharly Actually, I did fix this! Look at lines 420-431 - I replaced the broken alphabetical sorting with per-wallet sequential numbering logic that:

1. parses existing names -> finds all "Account X" names in the wallet
2. finds next available number -> taking the first available sequential number
3. no more alphabetical sorting -> completely avoids the 0xbbb/0xaaa duplicate situation

I think your example would now result in:
- Create 0xbbb → Account 1 
- Create 0xaaa → Account 2 ✅ (not Account 1!)

We should be good with both entropy AND non-entropy wallets. Do you agree?

fabiobozzo: I added a UT for your specific scenario, too.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,30 +397,38 @@ export class AccountTreeController extends BaseController<
...
+      // For entropy-based multichain groups, use the actual groupIndex from metadata
+      if (group.type === AccountGroupType.MultichainAccount) {
+        groupIndex = group.metadata.entropy.groupIndex;
+      } else {
+        // For other wallet types (snap/keyring), use per-wallet sequential numbering
+        // to avoid the alphabetical sorting issue that can cause duplicate names
+        const existingGroupNames = Object.values(wallet.groups)
+          .map((g) => g.metadata.name)
+          .filter((name) => /^Account \d+$/u.test(name))
+          .map((name) => parseInt(name.replace('Account ', ''), 10) - 1); // Convert to 0-based
```
ccharly: Not sure this would scale well with the actual naming of other wallets... I think for the moment we're always defaulting to `Account N`, but normally private key wallet would use `Imported Account N` (I don't think we're applying this logic today though 😅)

But that might be a non problem for now, and we can always re-adapt that later I guess.

That was the logic we were using for the `accounts-controller`, and this was also using the number of accounts as a fallback: https://github.com/MetaMask/core/blob/main/packages/accounts-controller/src/AccountsController.ts#L1063-L1100

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,30 +397,38 @@ export class AccountTreeController extends BaseController<
...
+
+        // Find the first available index
         groupIndex = 0;
+        while (existingGroupNames.includes(groupIndex)) {
+          groupIndex += 1;
+        }
       }
 
       // Use computed name first, then fallback to default naming if empty
```
ccharly: Below this part, we could still have issue with multichain account names IMO:
- You have your 1st multichain account -> "Account 1"
- You rename it to "Account 2"
- You create your 2nd multichain account -> "Account 2"

We now have 2 accounts named "Account 2".

We could maybe run a small conflict logic that if your default/computed name already exists, then we add a prefix (until we find a unique name).

WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: @ccharly Different wallet types might use different patterns indeed. Gotcha. 

I made the logic dynamic by detecting the naming pattern from `rule.getDefaultAccountGroupName(0)` and building a regex to extract numbers from ANY pattern, now. This should future-proof the logic so it works with "Account N", "Imported Account N", "Private Key N", or any pattern without hard-coding assumptions.

I added a test case to prove that the pattern detection and number extraction works for any format.
You're right it's not a problem today since we only use "Account N", but I prefer the logic to be ready for future patterns, already.

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: @ccharly That's exactly the bug I was missing - entropy groups not checking for conflicts with renamed accounts. 😮‍💨 

Please take a look now: conflict detection should be more universal, for ALL wallet types (entropy + non-entropy) using a `while` **loop that increments until it finds a unique name**.

Your scenario (Account 1 → rename to "Account 2" → new account should get "Account 3") is a case that would cause duplicate names:

```typescript
do {
  proposedName = rule.getDefaultAccountGroupName(groupIndex);
  nameExists = Object.values(wallet.groups).some(
    (g) => g.id !== group.id && g.metadata.name === proposedName
  );
  if (nameExists) groupIndex += 1;
} while (nameExists);
```

Again, a new test case proves that conflict avoidance works.
Your scenario should now be impossible.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +397,63 @@ export class AccountTreeController extends BaseController<
...
+        // Generate default name and ensure it's unique within the wallet
+        let proposedName = '';
+        let groupIndex: number;
+
+        // For entropy-based multichain groups, start with the actual groupIndex
+        if (group.type === AccountGroupType.MultichainAccount) {
+          groupIndex = group.metadata.entropy.groupIndex;
+        } else {
+          // For other wallet types, start with sequential numbering
+          groupIndex = 0;
```
ccharly: I wonder if we should use the max number of accounts of this wallet as a starting point 🤔 This way we avoid the weird case where:
- Using a private key wallet:
  * Import private key -> Account 1
  * Rename Account 1 -> My Account
  * Import another private key -> Account 1
    - If we use the `.length`, it would become Account 2, which I think, makes more sense here, WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +397,63 @@ export class AccountTreeController extends BaseController<
...
+        // Find a unique name by checking for conflicts and incrementing if needed
+        let nameExists = true;
+        while (nameExists) {
+          proposedName = rule.getDefaultAccountGroupName(groupIndex);
+
+          // Check if this name already exists in the wallet (excluding current group)
+          const currentProposedName = proposedName; // Capture for closure
+          nameExists = Object.values(wallet.groups).some(
+            (g) => g.id !== group.id && g.metadata.name === currentProposedName,
+          );
```
ccharly: Maybe it would make sense to re-use the same logic than `isAccountGroupNameUnique` and add a new variant that takes a wallet as input, so we can use here + in `isAccountGroupNameUnique`.

Something like `isAccountGroupNameUniqueFromWallet` maybe?

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: And maybe we could apply the same logic for multichain accounts too... I need to double-check internally, but that could make sense too (despite losing some "meaning" since we would not really used the `groupIndex` anymore, but that's maybe fine)

Normal case scenario would just be sequential and would work 100% of the time.

fabiobozzo: Actually, this scenario is already handled by my current conflict resolution logic. That `while` loop would detect that "Account 1" conflicts with the renamed account and automatically increment to "Account 2". 

But you make a good point about efficiency, imo: starting with the wallet length is more efficient than always starting at 0 and incrementing. No more unnecessary iterations I suppose. I'll update the logic to start with `Object.keys(wallet.groups).length` as the initial groupIndex for non-entropy groups.

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Done `isAccountGroupNameUniqueFromWallet` ✔️ Though I'm not reusing `isAccountGroupNameUnique` internally 👁️👁️

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +397,65 @@ export class AccountTreeController extends BaseController<
...
+        let groupIndex: number;
+
+        // For entropy-based multichain groups, start with the actual groupIndex
+        if (group.type === AccountGroupType.MultichainAccount) {
+          groupIndex = group.metadata.entropy.groupIndex;
+        } else {
+          // For other wallet types, start with the number of existing groups
+          // This avoids unnecessary conflict checks in most cases
+          groupIndex = Object.keys(wallet.groups).length - 1;
+        }
```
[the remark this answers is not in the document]
fabiobozzo: Added safety check: `group.metadata.entropy && group.metadata.entropy.groupIndex`. 

`Object.keys(wallet.groups).length - 1` would have returned -1 for empty wallets. ⛔ 
`Object.keys(wallet.groups).length` returns 0 for empty wallets ✔️

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +397,68 @@ export class AccountTreeController extends BaseController<
...
+          nameExists = !this.#isAccountGroupNameUniqueFromWallet(
+            wallet,
+            group.id,
+            proposedName,
+          );
+
+          if (nameExists) {
+            groupIndex += 1; // Try next number
+          }
+        }
```
ccharly: Nit: That looks like a `do while` to me 🤔 WDYT about:

```suggestion
        do {
          proposedName = rule.getDefaultAccountGroupName(groupIndex);

          // Check if this name already exists in the wallet (excluding current group)
          const nameExists = !this.#isAccountGroupNameUniqueFromWallet(
            wallet,
            group.id,
            proposedName,
          );

          if (nameExists) {
            groupIndex += 1; // Try next number
          }
        } while (nameExists);
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -758,6 +785,30 @@ export class AccountTreeController extends BaseController<
...
+    const trimmedName = name.trim();
+
+    // Check for duplicates within this wallet
+    for (const group of Object.values(wallet.groups)) {
+      if (group.id !== groupId && group.metadata.name.trim() === trimmedName) {
+        return false;
+      }
+    }
+    return true;
+  }
```
ccharly: Actually, let's move this as a free function (similar to our `isAccountGroupNameUnique`) so we can re-use it in that function and use it here too. WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -983,35 +1034,73 @@ export class AccountTreeController extends BaseController<
...
+   */
+  resolveNameConflict(groupId: AccountGroupId, name: string): string {
+    let suffix = 2;
+    let candidateName = `${name} (${suffix})`;
+
+    // Keep incrementing suffix until we find a unique name
+    while (!isAccountGroupNameUnique(this.state, groupId, candidateName)) {
+      suffix += 1;
+      candidateName = `${name} (${suffix})`;
+    }
```
ccharly: I wonder if we could just re-use your new function `isAccountGroupNameUniqueFromWallet`? This way we won't "find" the wallet associated everytime there's a conflict? We just get it once, and re-use it until we find a unique name?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -983,35 +1034,73 @@ export class AccountTreeController extends BaseController<
...
+    let finalName = name;
+
+    // Handle name conflicts based on the autoHandleConflict flag
+    /* istanbul ignore next */
+    if (
+      autoHandleConflict &&
+      !isAccountGroupNameUnique(this.state, groupId, name)
+    ) {
+      /* istanbul ignore next */
+      finalName = this.resolveNameConflict(groupId, name);
```
ccharly: Following-up my previous comment, maybe we should first "get the wallet" then we use `isAccountGroupNameUniqueFromWallet` and pass it down to `resolveNameConflict` too? This way we avoid trying to find the wallet multiple times. WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +397,68 @@ export class AccountTreeController extends BaseController<
...
+        if (
+          group.type === AccountGroupType.MultichainAccount &&
+          group.metadata.entropy
+        ) {
+          groupIndex = group.metadata.entropy.groupIndex;
+        } else {
+          // For other wallet types, start with the number of existing groups
+          // This gives us the next logical sequential number
+          groupIndex = Object.keys(wallet.groups).length;
+        }
```
ccharly: So, from out internal discussions, I think we agreed on parsing the highest account index being used (similarly to what we do in the `accounts-controller` today).

So we go through all accounts from a group, and then we parse each of them to find the highest index.

We should also use `const nameIndex = Math.max(highestAccountNameIndex, Object.keys(wallet.groups).length));` so we make sure to always use an index that "at least" aligned with the number of accounts a group holds.

fabiobozzo: Ok here I implemented the parsing logic as you suggested, now! We parse all existing group names to find the highest index and use `Math.max(highestAccountNameIndex, groupIndex)` to ensure proper sequential numbering. It respects now both existing names and entropy metadata. Please check that out 😉

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: I moved that private method to `group.ts` as isAccountGroupNameUniqueFromWallet, is that what you meant?
isAccountGroupNameUnique uses now the new function for consistency, internally.

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: More appropriate since we always need to generate at least one name. ✔️

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: I agree. Now `resolveNameConflict` gets the wallet once and reuses `isAccountGroupNameUniqueFromWallet` call, it avoids repeated wallet lookups while generating suffixes.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +400,89 @@ export class AccountTreeController extends BaseController<
...
+
+        // Use the higher of the two: highest parsed index or computed index
+        groupIndex = Math.max(highestAccountNameIndex, groupIndex);
```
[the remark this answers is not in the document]
fabiobozzo: This behavior is intentional and approved by the team. The Math.max logic ensures that new accounts don't conflict with existing renamed accounts, even if they create gaps. User input validation prevents "Account N" patterns in manual renames, making this edge case acceptable for our use case.

Discussed with @ccharly @Tlees-MMI  and team - keeping current implementation.

fabiobozzo:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -983,44 +1035,86 @@ export class AccountTreeController extends BaseController<
...
     this.#assertAccountGroupExists(groupId);
 
-    // Validate that the name is unique
-    this.#assertAccountGroupNameIsUnique(groupId, name);
-
     const walletId = this.#groupIdToWalletId.get(groupId);
+    /* istanbul ignore next */
+    if (!walletId) {
+      throw new Error(`Account group with ID "${groupId}" not found in tree`);
+    }
```
ccharly: Actually, I wonder if we should start using `assert` a bit more for those invariants 🤔

This would be naturally covered + they look nice IMO

Could you try that and see if the the `ignore next` can be removed?

ccharly: by `assert` I mean our function here: https://github.com/MetaMask/utils/blob/main/src/assert.ts#L69-L94

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +400,90 @@ export class AccountTreeController extends BaseController<
...
+        // Parse the highest account index being used (similar to accounts-controller)
+        let highestAccountNameIndex = 0;
+        for (const existingGroup of Object.values(wallet.groups)) {
+          // Skip the current group being processed
+          if (existingGroup.id === group.id) {
+            continue;
+          }
+          // Parse the existing group name to extract the numeric index
+          const nameMatch =
+            existingGroup.metadata.name.match(/Account (\d+)$/u);
```
ccharly: Side note: this regex might be a "too wide/open", if someone uses `My Account N` then this would be matched by this regex, and that's slightly different from our original implementation on the `AccountsController`.

But IMO, that's ok to keep it as-is for now.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +400,90 @@ export class AccountTreeController extends BaseController<
...
-        rule.getComputedAccountGroupName(typedGroup) ||
-        rule.getDefaultAccountGroupName(groupIndex);
+      const computedName = rule.getComputedAccountGroupName(typedGroup);
+
+      if (computedName) {
+        group.metadata.name = computedName;
+      } else {
+        // Generate default name and ensure it's unique within the wallet
+        let proposedName = '';
+        let groupIndex: number;
```
ccharly: I think `groupIndex` is misleading here. I would rename it to `proposedNameIndex` maybe?

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -13,9 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Coming from the old account model, a non-EVM account could have been selected and the lastly selected EVM account might not be using the same group index.
 - Bump `@metamask/utils` from `^11.4.2` to `^11.8.0` ([#6588](https://github.com/MetaMask/core/pull/6588))
 
-### Removed
+### Fixed
 
-- Remove use of `:getSelectedAccount` action ([#6608](https://github.com/MetaMask/core/pull/6608))
+- **CRITICAL**: Fix multi-wallet account naming bug causing duplicate and inconsistent names across wallets ([#6601](https://github.com/MetaMask/core/pull/6601))
```
```suggestion
- Fix multi-wallet account naming bug causing duplicate and inconsistent names across wallets ([#6601](https://github.com/MetaMask/core/pull/6601))
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +400,90 @@ export class AccountTreeController extends BaseController<
...
+            proposedName,
+          );
+
+          /* istanbul ignore next */
+          if (nameExists) {
+            groupIndex += 1; // Try next number
+          }
+        } while (nameExists);
+
+        group.metadata.name = proposedName;
```
ccharly: We should persist the account name here no?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -394,44 +400,99 @@ export class AccountTreeController extends BaseController<
...
+        ) {
+          proposedNameIndex = group.metadata.entropy.groupIndex;
+        } else {
+          // For other wallet types, start with the number of existing groups
+          // This gives us the next logical sequential number
+          proposedNameIndex = Object.keys(wallet.groups).length;
+        }
+
+        // Use the higher of the two: highest parsed index or computed index
+        proposedNameIndex = Math.max(highestAccountNameIndex, proposedNameIndex);
```
ccharly: Actually, shouldn't we add `+ 1` to `highestAccountNameIndex`? 🤔 Since this name should already exists, we cannot re-use it anyway

Also, we want to get the name of the "next account", so the `+ 1` seems aligned with this?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: True! we're calculating the index for the proposed name, not necessarily the group's actual index. 😮‍💨

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: mmm 🤔 since this is a string assertion rather than a struct validation, I used the traditional if/throw pattern for clarity, do you think it's worth to `assert` instead ?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Right... what would be the cons of adding a **^** to the regex ? 💭 

`/^Account (\d+)$/u` ?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: (instanbul ignore was removed) ✔️ now it's time to go back at arranging that 100% coverage 💤

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: You're absolutely right! I was missing the this.update call to persist the generated name. 🤦🤦🤦 Added it now - so that we persist the generated names during initialization.. to `accountGroupsMetadata`, across restarts, if I'm correct. Right?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Yep, I was initially fooled by a cursorbot remark, about this. Now I implemented it in a way so that it only adds +1 when there are actually existing parsed names (highestAccountNameIndex > 0), preserving the original behavior for new wallets while fixing the gap issue in the other case. Does it make sense?

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: NVM! Your assert suggestion actually helped me improve the tests while keeping the coverage. THanks!

fabiobozzo:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
fabiobozzo: Added explanatory TODO there ✔️

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -30,9 +40,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-### Removed
+### Fixed
 
-- Remove use of `:getSelectedAccount` action ([#6608](https://github.com/MetaMask/core/pull/6608))
+- Fix multi-wallet account naming bug causing duplicate and inconsistent names across wallets ([#6601](https://github.com/MetaMask/core/pull/6601))
+  - Each wallet now has independent sequential numbering (Account 1, Account 2, etc.).
+  - Names are immediately persisted to ensure consistency across app restarts.
+  - Entropy groups use their actual group index instead of alphabetical sorting.
+  - Prevents issues like "Account 2, Account 2, Account 3..." and missing "Account 1".
```
ccharly: We should revert this back

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -387,51 +399,111 @@ export class AccountTreeController extends BaseController<
...
+          // For other wallet types, start with the number of existing groups
+          // This gives us the next logical sequential number
+          proposedNameIndex = Object.keys(wallet.groups).length;
+        }
+
+        // Use the higher of the two: highest parsed index or computed index
+        // Always add +1 to get the next available index for the new group
+        proposedNameIndex =
+          Math.max(highestAccountNameIndex, proposedNameIndex) + 1;
+
```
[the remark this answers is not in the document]
fabiobozzo: @ccharly ☝️ reverting our +1 change

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -387,51 +399,112 @@ export class AccountTreeController extends BaseController<
...
+          // This gives us the next logical sequential number
+          proposedNameIndex = Object.keys(wallet.groups).length;
+        }
+
+        // Use the higher of the two: highest parsed index or computed index
+        proposedNameIndex = Math.max(
+          highestAccountNameIndex,
+          proposedNameIndex,
+        );
+
```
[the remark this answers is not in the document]
fabiobozzo: Again, that's by design.

fabiobozzo:
@metamaskbot publish-preview

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,12 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 
 - Single group sync events will not get enqueued anymore if a full sync is in progress ([#6651](https://github.com/MetaMask/core/pull/6651))
   - This prevents too many unnecessary storage fetches (which would prevent being rate limited).
   - This could rarely lead to inconsistencies until the next single updates or next full sync.
 
+### Fixed
+
+- Fix critical multi-wallet account group naming inconsistencies and duplicates ([#6601](https://github.com/MetaMask/core/pull/6601))
```
```suggestion
- Fix multi-wallet account group naming inconsistencies and duplicates ([#6601](https://github.com/MetaMask/core/pull/6601))
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -267,6 +272,13 @@ export class AccountTreeController extends BaseController<
     this.update((state) => {
       state.accountTree.wallets = wallets;
 
+      // Apply group metadata within the state update
+      for (const wallet of Object.values(wallets)) {
```
ccharly: As discussed internally, we should be using `state` here, otherwise the parameters we pass down to `#applyAccountGroupMetadata` will reference the `wallets` ref, which won't mutate the `state.accountTree` names.

Though, the persisted names are good 👍

```suggestion
      for (const wallet of Object.values(state.accountTree.wallets)) {
```

ccharly: Actually.. now that we pass the `state` to `#applyAccountGroupMetadata` we could maybe re-use it that directly?

So the caller is not expected to pass a reference to the `immer` draft directly. That sounds better to me, WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -387,51 +399,112 @@ export class AccountTreeController extends BaseController<
...
+            proposedName,
+          );
+
+          /* istanbul ignore next */
+          if (nameExists) {
+            proposedNameIndex += 1; // Try next number
+          }
+        } while (nameExists);
+
+        group.metadata.name = proposedName;
```
ccharly: Related to [my previous comment](https://github.com/MetaMask/core/pull/6601/files#r2362275934), we should do this instead:
```suggestion
        state.accountTree.wallets[wallet.id].groups[group.id].metadata.name = proposedName;
```

ccharly: Actually, your change looks better 💪

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -267,6 +272,19 @@ export class AccountTreeController extends BaseController<
...
+        state.accountTree.wallets,
+      )) {
+        for (const groupId of Object.keys(wallet.groups)) {
+          this.#applyAccountGroupMetadata(
+            state,
+            walletId as AccountWalletId,
+            groupId as AccountGroupId,
+          );
+        }
+      }
```
ccharly: We can probably re-use the original pattern, which will avoid those type-casts

```suggestion
      for (const wallet of Object.values(wallets)) {
        for (const group of Object.values(wallet.groups)) {
          this.#applyAccountGroupMetadata(state, wallet.id, group.id);
        }
      }
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `autoHandleConflict` parameter to `setAccountGroupName` method for automatic conflict resolution with suffix generation ([#6601](https://github.com/MetaMask/core/pull/6601))
+
+### Fixed
```
ccharly, line 19:
```suggestion
### Changed

- Computed names (inherited from previous existing accounts) is disabled temporarily ([#6601](https://github.com/MetaMask/core/pull/6601))
  - They do interfere with the naming mechanism, so we disable them temporarily in favor of the new per-wallet sequential naming.

### Fixed
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1072,7 +1174,7 @@ export class AccountTreeController extends BaseController<
 
     this.update((state) => {
       // Update persistent metadata
-      state.accountGroupsMetadata[groupId] ??= {};
+      state.accountGroupsMetadata[groupId] = {};
```
ccharly: Wait, why are we changing this? We could overwrite other metadata because of this no?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1110,7 +1212,7 @@ export class AccountTreeController extends BaseController<
...
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -983,44 +1049,87 @@ export class AccountTreeController extends BaseController<
...
+    } else {
+      // Validate that the name is unique
+      this.#assertAccountGroupNameIsUnique(groupId, finalName);
+    }
 
     this.update((state) => {
+      /* istanbul ignore next */
+      if (!state.accountGroupsMetadata[groupId]) {
+        state.accountGroupsMetadata[groupId] = {};
+      }
```
ccharly, line 1118: This is ok for now, we can refactor that later.

PR #6607, opened 2025-09-15
chore: Synchronously initialise network controller provider if `lookupNetwork` is `false`

Mrtenz:
## Explanation

This updates the `initializeProvider` function of the network controller to be synchronous if `lookupNetwork` is `false`, i.e., when no async operations need to be performed. Previously the function was `async` meaning it would always return a promise, regardless of whether any async operations are performed. With some function overloads and by removing async (returning a promise instead when necessary), we can make it fully synchronous when `lookupNetwork` is `false`.

Doing this will allow for safe initialisation of the network controller state in places where awaiting may be difficult, like in the client's modular controller initialisation.

## Checklist

### DIFF
--- a/packages/network-controller/src/NetworkController.ts
+++ b/packages/network-controller/src/NetworkController.ts
@@ -1571,7 +1571,50 @@ export class NetworkController extends BaseController<
    * You can pass `false` if you'd like to disable this request and call
    * `lookupNetwork` yourself.
    */
-  async initializeProvider({
+  initializeProvider(options: { lookupNetwork: false }): void;
+
+  /**
+   * Creates proxies for accessing the global network client and its block
+   * tracker. You must call this method in order to use
+   * `getProviderAndBlockTracker` (or its replacement,
+   * `getSelectedNetworkClient`).
+   *
+   * @param options - Optional arguments.
+   * @param options.lookupNetwork - Usually, metadata for the global network
+   * will be populated via a call to `lookupNetwork` after creating the provider
+   * and block tracker proxies. This allows for responding to the status of the
+   * global network after initializing this controller; however, it requires
+   * making a request to the network to do so. In the clients, where controllers
+   * are initialized before the UI is shown, this may be undesirable, as it
+   * means that if the user has just installed MetaMask, their IP address may be
+   * shared with a third party before they have a chance to finish onboarding.
+   * You can pass `false` if you'd like to disable this request and call
+   * `lookupNetwork` yourself.
+   * @returns A promise that resolves when the network lookup completes.
+   */
+  initializeProvider(options: { lookupNetwork?: true }): Promise<void>;
+
+  /**
+   * Creates proxies for accessing the global network client and its block
+   * tracker. You must call this method in order to use
+   * `getProviderAndBlockTracker` (or its replacement,
+   * `getSelectedNetworkClient`).
+   *
+   * @param options - Optional arguments.
+   * @param options.lookupNetwork - Usually, metadata for the global network
+   * will be populated via a call to `lookupNetwork` after creating the provider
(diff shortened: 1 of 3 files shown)


packages/network-controller/src/NetworkController.ts
```diff
@@ -1571,16 +1571,61 @@ export class NetworkController extends BaseController<
...
+   * global network after initializing this controller; however, it requires
+   * making a request to the network to do so. In the clients, where controllers
+   * are initialized before the UI is shown, this may be undesirable, as it
+   * means that if the user has just installed MetaMask, their IP address may be
+   * shared with a third party before they have a chance to finish onboarding.
+   * You can pass `false` if you'd like to disable this request and call
+   * `lookupNetwork` yourself.
+   * @returns A promise if `lookupNetwork` is true or undefined, otherwise
+   * returns nothing.
+   */
```
mcmire: Nit: Do you need this JSDoc block? As far as I am aware, it won't be recognized by the TypeScript server or TypeDoc.

``` suggestion
```

packages/network-controller/src/NetworkController.ts
```diff
@@ -1571,16 +1571,61 @@ export class NetworkController extends BaseController<
...
+   * global network after initializing this controller; however, it requires
+   * making a request to the network to do so. In the clients, where controllers
+   * are initialized before the UI is shown, this may be undesirable, as it
+   * means that if the user has just installed MetaMask, their IP address may be
+   * shared with a third party before they have a chance to finish onboarding.
+   * You can pass `false` if you'd like to disable this request and call
+   * `lookupNetwork` yourself.
+   * @returns A promise that resolves when the network lookup completes.
+   */
+  initializeProvider(options?: { lookupNetwork?: boolean }): Promise<void>;
```
mcmire, line 1595: For accuracy, should this be `true`?

``` suggestion
  initializeProvider(options?: { lookupNetwork?: true }): Promise<void>;
```

mcmire (dismissed):
LGTM!

packages/network-controller/src/NetworkController.ts
Mrtenz: I tried this, but that results in type errors elsewhere because `boolean` isn't assignable to `true`.

packages/network-controller/src/NetworkController.ts
Mrtenz: I assumed TypeScript would recognise it based on the overload that was used 🤔

packages/network-controller/src/NetworkController.ts
mcmire: Ah, okay, no problem.

packages/network-controller/src/NetworkController.ts
mcmire: It does, but when TypeScript looks up the JSDoc for an overload it looks for it to exist  above the overload itself. I believe it effectively ignores the JSDoc above the implementation in that case.

Mrtenz: Oh, I see what you mean now. Applied your suggestion!

mcmire (approved):

packages/network-controller/src/NetworkController.ts
```diff
@@ -1571,16 +1571,41 @@ export class NetworkController extends BaseController<
...
     lookupNetwork?: boolean;
   } = {}) {
     this.#applyNetworkSelection(this.state.selectedNetworkClientId);
 
     if (lookupNetwork) {
-      await this.lookupNetwork();
+      return this.lookupNetwork();
     }
+
+    return undefined;
```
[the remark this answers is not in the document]
mcmire: Not sure this is true.

PR #6617, opened 2025-09-16
Feat/add token screening tx controller

AugmentedMode:
## Checklist


cryptotavares (changes requested):
There are two solutions that would not require coupling our code into the transactions controller. We can:
* just listen for transaction controller state changes within the phishing controller, and whenever there are state changes in a transaction simulation data, we fire the bulk scanning method. This should be seamless as all controllers already emit state change events.
* add the logic to the actual (afterSimulate)[https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/controller-init/confirmations/transaction-controller-init.ts#L131] hook on the transaction controller initialization. The goal of the hooks architecture is for us to be able to add non transaction related logic to be run in different steps of the transaction lifecycle but at the same time keep the codebase decoupled and the domain ownership easily identified.

I would personaly go with option 1 that I highlighted above, as it allows us to adopt the solution on mobile without any changes (despite the UI ones to display the trust signals).

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -498,6 +534,82 @@ export class PhishingController extends BaseController<
...
+            this.#scanTokensFromSimulation(transaction).catch((error) =>
+              console.error('Error scanning tokens from simulation:', error),
+            );
+          }
+        }
+      }
+    } catch (error) {
+      console.error('Error processing transaction state change:', error);
+    }
+  }
```
cryptotavares, line 577: Instead of adding the selector to the stateChanges listener, and then having to iterate through the entire transaction object, why are we not just getting the patches produced by immer and processing those? It should be faster and would prevent a lot of unnecessary processing (the transactions data is updated very frequently.. for example to update gas values).

If we rely on immer patches we could do something like this (I am sure that the js code can be polished):
```js
    const addresses: string[] = [];
    const previousTransactions = patches.reduce((acc, patch) => {
      if (patch.op === 'remove') {
        return acc;
      }

      const { path } = patch;
      if (
        path.includes('simulationData') &&
        path.includes('tokenBalanceChanges') &&
        path.includes('address')
      ) {
        acc.push(patch.value);
      }
      return acc;
    }, addresses);
```

AugmentedMode: Hey Joao! I pushed an update to use patches to extract token addresses from the transaction data changes. I had to implement it slightly differently than your suggested approach above, as I discovered that the entire transaction object gets replaced when simulation data is added /changed, rather than individual properties being updated.

It looks like the patches occur at higher levels like `['transactions', 0]  rather than the deeper property paths from my findings. 

This approach i pushed should now handle transaction level patches, then extract token addresses from the current transaction state and group them by chain for bulk scanning.

Let me know your feedback 🙏

I held off on updating test for now but I did manually QA that this is working as expected !

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -48,8 +53,7 @@ export const PHISHING_DETECTION_BASE_URL =
 export const PHISHING_DETECTION_SCAN_ENDPOINT = 'v2/scan';
 export const PHISHING_DETECTION_BULK_SCAN_ENDPOINT = 'bulk-scan';
 
-export const SECURITY_ALERTS_BASE_URL =
-  'https://security-alerts.api.cx.metamask.io';
+export const SECURITY_ALERTS_BASE_URL = 'http://localhost:3000';
```
cryptotavares: Typo.. I don't think that we want to hit localhost 😅

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -498,6 +527,105 @@ export class PhishingController extends BaseController<
...
+    try {
+      const tokensByChain = new Map<string, Set<string>>();
+
+      for (const patch of patches) {
+        if (patch.op === 'remove') {
+          continue;
+        }
+
+        // Handle transaction-level patches (includes simulation data updates)
+        if (this.#isTransactionPatch(patch) && patch.path.length === 2) {
```
cryptotavares: We can remove the `path.length === 2` condition, and add it to the isTransactionPatch (which is already validating `path.length >= 2`)

PR #6618, opened 2025-09-16
feat: add `Monad Mainnet` support to `@metamask/assets-controllers`

stanleyyconsensys:
## Explanation
This PR adds support for Monad Mainnet to the MetaMask assets controllers by integrating the new blockchain network (chain ID 0x8f) into the token detection and pricing systems.

### Notes
Monad Mainnet has not launch yet, there is no RPC but only infura to support this network 

Accounts API is not yet merged, so even we add the chain id, it wont discover the balance via API, but the RPC

## Checklist

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Shared fiat currency and token formatters ([#6577](https://github.com/MetaMask/core/pull/6577))
 - Add `Monad Mainnet` support ([#6618](https://github.com/MetaMask/core/pull/6618))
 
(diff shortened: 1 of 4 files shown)


packages/assets-controllers/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Shared fiat currency and token formatters ([#6577](https://github.com/MetaMask/core/pull/6577))
+- Add `Monad Mainnet` support ([#6618](https://github.com/MetaMask/core/pull/6618))
```
Prithpal-Sooriya, line 12: nit/comment - should we breakdown the places that have added this support?
- added in AssetContractControlller
- added in assetsUtil
- added in token prices service

PR #6624, opened 2025-09-16
feat(multichain-account-service): add account providers config + add Solana discovery timeout

ccharly:
## Explanation

Allow to pass some custom configurations to each built-in account providers.

This should allow to easily customize them from the clients directly.

## References

N/A

## Checklist


packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -165,13 +173,48 @@ function setup({
...
+      });
+
+      expect(mocks.EvmAccountProvider.constructor).toHaveBeenCalledWith(
+        messenger,
+        providersConfigs[EvmAccountProvider.NAME],
+      );
+      expect(mocks.SolAccountProvider.constructor).toHaveBeenCalledWith(
+        messenger,
+        providersConfigs[SolAccountProvider.NAME],
+      );
```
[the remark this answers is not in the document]
ccharly: That's correct, and IDK why `jest` is considering `messenger` to be valid (cause this test is currently successful...

Nevertheless, I'll change that.

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
...
   });
   service.init();
 
-  return { service, messenger, mocks };
+  return { service, serviceMessenger, messenger, mocks };
 }
 
 describe('MultichainAccountService', () => {
+  describe('constructor', () => {
+    it('forwards conflicts to each providers', () => {
```
hmalik88:
```suggestion
    it('forwards configs to each provider', () => {
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -28,9 +32,13 @@ export const serviceName = 'MultichainAccountService';
 /**
  * The options that {@link MultichainAccountService} takes.
  */
-type MultichainAccountServiceOptions = {
+export type MultichainAccountServiceOptions = {
   messenger: MultichainAccountServiceMessenger;
   providers?: NamedAccountProvider[];
+  providersConfigs?: {
```
hmalik88:
```suggestion
  providerConfigs?: {
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -87,15 +89,18 @@ function setup({
   messenger = getRootMessenger(),
   keyrings = [MOCK_HD_KEYRING_1, MOCK_HD_KEYRING_2],
   accounts,
+  providersConfigs,
```
hmalik88:
```suggestion
  providerConfigs,
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -87,15 +89,18 @@ function setup({
...
 }: {
   messenger?: Messenger<
     MultichainAccountServiceActions | AllowedActions,
     MultichainAccountServiceEvents | AllowedEvents
   >;
   keyrings?: KeyringObject[];
   accounts?: KeyringAccount[];
+  providersConfigs?: MultichainAccountServiceOptions['providersConfigs'];
```
hmalik88:
```suggestion
  providerConfigs?: MultichainAccountServiceOptions['providerConfigs'];
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
     );
   }
 
+  const serviceMessenger = getMultichainAccountServiceMessenger(messenger);
   const service = new MultichainAccountService({
-    messenger: getMultichainAccountServiceMessenger(messenger),
+    messenger: serviceMessenger,
+    providersConfigs,
```
hmalik88:
```suggestion
    providerConfigs,
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
...
   service.init();
 
-  return { service, messenger, mocks };
+  return { service, serviceMessenger, messenger, mocks };
 }
 
 describe('MultichainAccountService', () => {
+  describe('constructor', () => {
+    it('forwards conflicts to each providers', () => {
+      const providersConfigs: MultichainAccountServiceOptions['providersConfigs'] =
```
hmalik88:
```suggestion
      const providerConfigs: MultichainAccountServiceOptions['providerConfigs'] =
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
...
+          [SolAccountProvider.NAME]: {
+            discovery: {
+              timeoutMs: 1000,
+            },
+          },
+        };
+
+      const { mocks, serviceMessenger } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1, MOCK_SOL_ACCOUNT_1],
+        providersConfigs,
```
hmalik88:
```suggestion
        providerConfigs,
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
...
+      });
+
+      expect(mocks.EvmAccountProvider.constructor).toHaveBeenCalledWith(
+        serviceMessenger,
+        providersConfigs[EvmAccountProvider.NAME],
```
hmalik88:
```suggestion
        providerConfigs[EvmAccountProvider.NAME],
```

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -163,15 +172,51 @@ function setup({
...
+      );
+      expect(mocks.SolAccountProvider.constructor).toHaveBeenCalledWith(
+        serviceMessenger,
+        providersConfigs[SolAccountProvider.NAME],
```
hmalik88:
```suggestion
        providerConfigs[SolAccountProvider.NAME],
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -69,19 +77,30 @@ export class MultichainAccountService {
    * @param options.messenger - The messenger suited to this
    * MultichainAccountService.
    * @param options.providers - Optional list of account
+   * @param options.providersConfigs - Optional providers configs
```
hmalik88:
```suggestion
   * @param options.providerConfigs - Optional provider configs
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -69,19 +77,30 @@ export class MultichainAccountService {
...
    * providers.
    */
-  constructor({ messenger, providers = [] }: MultichainAccountServiceOptions) {
+  constructor({
+    messenger,
+    providers = [],
+    providersConfigs,
```
hmalik88:
```suggestion
    providerConfigs,
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -69,19 +77,30 @@ export class MultichainAccountService {
...
     this.#messenger = messenger;
     this.#wallets = new Map();
     this.#accountIdToContext = new Map();
 
     // TODO: Rely on keyring capabilities once the keyring API is used by all keyrings.
     this.#providers = [
-      new EvmAccountProvider(this.#messenger),
+      new EvmAccountProvider(
+        this.#messenger,
+        providersConfigs?.[EvmAccountProvider.NAME],
```
hmalik88:
```suggestion
        providerConfigs?.[EvmAccountProvider.NAME],
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -69,19 +77,30 @@ export class MultichainAccountService {
...
+      ),
       new AccountProviderWrapper(
         this.#messenger,
-        new SolAccountProvider(this.#messenger),
+        new SolAccountProvider(
+          this.#messenger,
+          providersConfigs?.[SolAccountProvider.NAME],
```
hmalik88:
```suggestion
          providerConfigs?.[SolAccountProvider.NAME],
```

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -14,21 +14,43 @@ import type { Json, JsonRpcRequest } from '@metamask/utils';
...
   readonly #client: KeyringClient;
 
-  constructor(messenger: MultichainAccountServiceMessenger) {
+  readonly #config: SolAccountProviderConfig;
+
+  constructor(
+    messenger: MultichainAccountServiceMessenger,
+    config: SolAccountProviderConfig = {
+      discovery: {
+        // Use a pretty high timeout (12s) for each Solana discovery for now.
```
hmalik88: 12s seems pretty high for each discovery call, that's 24x over EVM (imagine importing 10+ accounts, it adds up quickly). If there's a legitimate failure/slowdown with the discovery on the solana snap side, keeping the timeout this high can mask that problem. Perhaps, the metrics that we spoke about discovery can actually be useful for this scenario. IIRC we were going to add an errors field, reporting these timeout errors in production can make it easier to diagnose if there's an actual problem in the snap; be it some laggy APIs or the actual snap code. I would opt to keep this closer to the EVM timeout, maybe 1.5-2x max that number.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -102,10 +124,13 @@ export class SolAccountProvider extends SnapAccountProvider {
     entropySource: EntropySourceId;
     groupIndex: number;
   }): Promise<Bip44Account<KeyringAccount>[]> {
-    const discoveredAccounts = await this.#client.discoverAccounts(
-      [SolScope.Mainnet],
-      entropySource,
-      groupIndex,
+    const discoveredAccounts = await withTimeout(
```
hmalik88: Do we know that the solana snap does not have a timeout mechanism in place already for discovery? I'm also thinking that we should have the timeout & retry stuff mentioned in the architecture documentation as well just so snap providers aren't implementing this stuff within the snap itself. Albeit small, doing this on the service side actually reduces some code duplication (if they are doing some retry/time-out mechanism).

packages/multichain-account-service/src/MultichainAccountService.test.ts
ccharly: God... How did I end up writing this 🙃 thanks!

packages/multichain-account-service/src/MultichainAccountService.test.ts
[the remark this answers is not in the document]
ccharly: I did change it now. Resolving.

packages/multichain-account-service/src/providers/SolAccountProvider.ts
ccharly: I haven't check on the Solana Snap, but I 100% agree with you that we have to implement this at service level.

Mainly cause we don't want this to start failing if the Snap is buggy or anything like that.

Regarding the retry, we could also add it there too, yes, this way we would have a similar way (for each built-in providers) to run the discovery 👍

I'll add this!

packages/multichain-account-service/src/providers/SolAccountProvider.ts
ccharly: Yeah, I honestly don't know how to deal with that one, but I saw some very long-running discovery (but that could be because the Snap was actually busy doing other things 🤔)

Anyway, I guess that if we add some retries + adapt the back-off delay, we could end-up getting closer to those ~10s of long-running discovery, so that might be better to go with this (smaller timeouts + more retries + slightly higher back-off delay than EVM)

I'll change this too. Thanks

PR #6635, opened 2025-09-17
Fix: TokenBalancesController selective session stopping to prevent polling interference

salimtb:
## 🐛 Problem

The `TokenBalancesController` had a critical issue where old polling sessions would interfere with new ones when chain configurations changed. This occurred because:

1. When `_startPolling` was called with new chain IDs, it would start new custom interval timers
2. The base class would later call `_stopPollingByPollingTokenSetId` for the old session 
3. The custom `_stopPollingByPollingTokenSetId` implementation would indiscriminately stop ALL polling, including the newly started sessions
4. This resulted in polling stopping unexpectedly when users switched networks or when chain configurations were updated

## 🔧 Solution

Implemented **selective session stopping** in `TokenBalancesController`:

1. **Enhanced `_stopPollingByPollingTokenSetId`**: Now parses the `tokenSetId` to extract chain IDs and only stops polling if they match the currently active session
2. **Added session validation**: Compares incoming stop requests against `#requestedChainIds` to prevent interference 
3. **Graceful error handling**: Handles malformed `tokenSetId` by falling back to stopping all polling
4. **Centralized timer management**: Introduced `#stopAllPolling()` method for consistent cleanup

## Checklist

### DIFF
--- a/packages/assets-controllers/src/TokenBalancesController.ts
+++ b/packages/assets-controllers/src/TokenBalancesController.ts
@@ -47,7 +47,6 @@ import type {
   TokensControllerState,
   TokensControllerStateChangeEvent,
 } from './TokensController';
-import type { PollingTokenSetId } from '../../polling-controller/src/types';
 
 export type ChainIdHex = Hex;
 export type ChecksumAddress = Hex;
@@ -446,7 +445,7 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
    *
    * @param tokenSetId - The token set ID to stop polling for
    */
-  override _stopPollingByPollingTokenSetId(tokenSetId: PollingTokenSetId) {
+  override _stopPollingByPollingTokenSetId(tokenSetId: string) {
     let parsedTokenSetId;
     let chainsToStop: ChainIdHex[] = [];
 


packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -442,12 +443,28 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
...
+   *
+   * @param tokenSetId - The token set ID to stop polling for
    */
-  override _stopPollingByPollingTokenSetId() {
-    this.#isControllerPollingActive = false;
-    this.#requestedChainIds = []; // Clear original intent when stopping
-    this.#intervalPollingTimers.forEach((timer) => clearInterval(timer));
-    this.#intervalPollingTimers.clear();
+  override _stopPollingByPollingTokenSetId(tokenSetId: PollingTokenSetId) {
+    const parsedTokenSetId = JSON.parse(tokenSetId);
```
Prithpal-Sooriya: Hmm okay, so we are passing in a string and parsing the contents?
Does this need to be a string or can we pass in the data we need to stop polling?

salimtb: this is a polling token , so yes we need to parse it , however we need to handle the errors , i'll do it now

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -442,12 +443,28 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
...
+    const chainsToStop = parsedTokenSetId.chainIds || [];
+
+    // Compare with current chains - only stop if it matches our current session
+    const currentChainsSet = new Set(this.#requestedChainIds);
+    const stopChainsSet = new Set(chainsToStop);
+
+    // Check if this stop request is for our current session
+    const isCurrentSession =
+      currentChainsSet.size === stopChainsSet.size &&
+      [...currentChainsSet].every((chain) => stopChainsSet.has(chain));
```
Prithpal-Sooriya, line 472: Okay I think I get it.

However what is a user want to only turn off some chains but not all?

Prithpal-Sooriya (approved):

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -47,6 +47,7 @@ import type {
   TokensControllerState,
   TokensControllerStateChangeEvent,
 } from './TokensController';
+import type { PollingTokenSetId } from '../../polling-controller/src/types';
```
cryptodev-2s: Nit: Should we instead export this type from `polling-controller` and use it here 
```suggestion
import type { PollingTokenSetId } from '@metamask/polling-controller';

salimtb: there's no export member for it lol

cryptodev-2s: Yes, we’ll need to add it to the exports in the polling-controller index.

cryptodev-2s: Because I don’t think we handle cross-package imports in this way.

salimtb: using a string instead now

Prithpal-Sooriya (approved):

packages/assets-controllers/src/TokenBalancesController.ts
Prithpal-Sooriya: This was confusion on my side, resolving

Prithpal-Sooriya: Dev notes:

This _stopPollingByPollingTokenSetId was stopping polling for all chains, now we only stop when the chains to stop are identical.

When can this happen? Pretty much _never_, any time you change chains, this is always different. We are only keeping this logic as a fallback for the current websocket work.

Code flow:
1. We start polling (add to our polling list)
    - Internally this handles stop polling for chains we're are not listening to anymore.
2. This override is called.

PR #6642, opened 2025-09-17
fix(account-tree-controller): swallow group creation errors

mathieuartu:
## Explanation

This PR swallows all group creation errors during backup and sync. This was already the case for multichain syncs but we forgot to do it for legacy syncs. It's now swallowed in every flow.

## Checklist

### DIFF
--- a/packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
+++ b/packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
@@ -58,12 +58,18 @@ export const createMultichainAccountGroup = async (
       });
     }
   } catch (error) {
+    // This can happen if the Snap Keyring is not ready yet when invoking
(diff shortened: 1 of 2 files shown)


packages/account-tree-controller/src/backup-and-sync/syncing/group.ts
```diff
@@ -58,12 +58,18 @@ export const createMultichainAccountGroup = async (
       });
     }
   } catch (error) {
+    // This can happen if the Snap Keyring is not ready yet when invoking
```
danroc, line 61: I suppose that most of the time this will be true, but what would happen if both EVM and Snap accounts are not created? Do we still recover later?

mathieuartu: Yes, this could be a problem. But any subsequent full sync will attempt the exact same thing until it finally works

PR #6645, opened 2025-09-17
feat: add bip44 default pairs to bridge config validation

infiniteflower:
## Explanation

This PR adds validation for `bip44DefaultPairs` from LaunchDarkly for the `BridgeController`

## Checklist

### DIFF
--- a/packages/bridge-controller/src/utils/validators.ts
+++ b/packages/bridge-controller/src/utils/validators.ts
@@ -103,6 +103,11 @@ const GenericQuoteRequestSchema = type({
 
 const FeatureIdSchema = enums(Object.values(FeatureId));
 
+const Bip44DefaultPairSchema = type({
+  destAsset: string(),
+  sourceAsset: string(),
+});
+
 /**
  * This is the schema for the feature flags response from the RemoteFeatureFlagController
  */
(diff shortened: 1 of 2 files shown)


packages/bridge-controller/src/utils/validators.ts
```diff
@@ -90,6 +103,7 @@ export const ChainConfigurationSchema = type({
   isSingleSwapBridgeButtonEnabled: optional(boolean()),
   isGaslessSwapEnabled: optional(boolean()),
   noFeeAssets: optional(array(string())),
+  bip44DefaultPairs: optional(DefaultPairSchema),
```
micaelae: based on the LD configs in use, doesn't this need to be `defaultPairs`? And line 139 should be `bip44DefaultPairs`

infiniteflower: I called the top level one `bip44DefaultPairs` since it's not chain specific, only CAIP chain namespace specific and requires us to be on BIP44.

The chain specific config is just `defaultPairs` since it doesn't really matter if we're on BIP44 or not.

micaelae: this is the chain-specific config though? and line 139 is the top level config

infiniteflower: Oh good catch! Thank you!

PR #6646, opened 2025-09-17
feat(account-tree-controller): use `multichain-account-service` to build multichain account wallet/group nodes

ccharly:
## Explanation

The `multichain-account-service` is responsible of holding/grouping all BIP-44 accounts together (in wallets and groups).

The `account-tree-controller` was re-using a similar matching logic to build it's tree and match those wallets and groups. But... In reality, this is the responsibility of the service and thus, the controller should "consume" the wallets/groups from the service directly.

Architecture wise, this is more correct, and other kind of wallets could follow the same pattern (having 1 wallet for each keyring types and 1 wallet for each other kind of account management Snaps).

This also fixes an issue with backup & sync which was listening to `:accountAdded` even for BIP-44 accounts. The problem was because the `EvmAccountProvider` is sometimes temporarily creating an account in order to get its associated address (during discovery) and then remove it afterward. Since backup & sync does not handle account deletion, this temporary account was triggering an account group creation on backup & sync, thus, "persisting" this account on the user remote profile.

## References

N/A

## Checklist


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1037,8 +1086,131 @@ export class AccountTreeController extends BaseController<
...
+      // Trigger atomic sync for new group.
+      this.#backupAndSyncService.enqueueSingleGroupSync(groupId);
+    }
+
+    // Map group ID to its containing wallet ID for efficient direct access
+    this.#groupIdToWalletId.set(groupId, walletId);
+
+    // Update the reverse mapping for all accounts account.
+    for (const account of accounts) {
+      this.#accountIdToContext.set(account.id, {
```
ccharly, line 1199: TODO: We need to clear accounts that might not be on the group (after the update) anymore

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1037,8 +1086,131 @@ export class AccountTreeController extends BaseController<
...
+        walletId: wallet.id,
+        groupId: group.id,
+        sortOrder: ACCOUNT_TYPE_TO_SORT_ORDER[account.type],
+      });
+    }
+
+    // We need to do this at every insertion because race conditions can happen
+    // during the account creation process where one provider completes before the other.
+    // The discovery process in the service can also lead to some accounts being created "out of order".
+    this.#sortAccountsOfGroup(group);
```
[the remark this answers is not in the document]
ccharly: Spotted that initially too:
- https://github.com/MetaMask/core/pull/6646#discussion_r2466551963

Should be solved by:
- https://github.com/MetaMask/core/pull/6646/commits/e679792e7b16cff930386c1a3012e611633584fd

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -277,6 +305,17 @@ export class AccountTreeController extends BaseController<
     const previousSelectedAccountGroup =
       this.state.accountTree.selectedAccountGroup;
 
+    // Insert all multichain account wallet/groups.
+    for (const wallet of this.#getMultichainAccountWallets()) {
```
fabiobozzo, line 311: No need to wrap this call with try-catch? e.g. in case `MultichainAccountService` is not available during init for some reason?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -240,6 +253,20 @@ export class AccountTreeController extends BaseController<
       },
     );
 
+    this.messenger.subscribe(
```
fabiobozzo, line 258: What if these 2 new handlers fire before the initialization completes? 

Should we add a guard before each like `if (!this.#initialized) return;` ?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -945,12 +1022,17 @@ export class AccountTreeController extends BaseController<
    * @param wallets - Account tree.
    * @param account - The account to be inserted.
    */
-  #insert(
+  #insertAccount(
     wallets: AccountTreeControllerState['accountTree']['wallets'],
     account: InternalAccount,
   ) {
+    if (isBip44Account(account)) {
```
fabiobozzo: `handleAccountAdded` already does this. Should we remove that check from there, or from here, and update init (the other caller of `insertAccount`)? Just to _DRY_

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -971,64 +1053,35 @@ export class AccountTreeController extends BaseController<
...
-          ...{ pinned: false, hidden: false }, // Default UI states
-          ...result.group.metadata, // Allow rules to override defaults
-        },
-        // We do need to type-cast since we're not narrowing `result` with
-        // the union tag `result.group.type`.
-      } as AccountGroupObject;
-      group = wallet.groups[groupId];
+    assert(
+      result.group.type === AccountGroupType.SingleAccount,
+      `Single account insertion should always result in a '${AccountGroupType.SingleAccount}' group type`,
```
fabiobozzo: Nit 🤏🏻 
```
`insertAccount expects SingleAccount groups, got ${result.group.type}. BIP-44 accounts should use ${AccountGroupType.SingleAccount}`
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1037,8 +1090,138 @@ export class AccountTreeController extends BaseController<
...
+   */
+  #insertOrUpdateMultichainAccountWalletAndGroup(
+    wallets: AccountTreeControllerState['accountTree']['wallets'],
+    multichainAccountWallet: MultichainAccountWallet<
+      Bip44Account<KeyringAccount>
+    >,
+    multichainAccountGroup: MultichainAccountGroup<
+      Bip44Account<KeyringAccount>
+    >,
+  ) {
```
fabiobozzo, line 1119: Maybe too defensive, but do we need to null check for ...
```
if (!multichainAccountWallet || !multichainAccountGroup) {
  throw new Error('Invalid multichain account wallet or group');
}
```
or do those 2 args' types imply they're not optional, already? 🤔

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -1037,8 +1090,138 @@ export class AccountTreeController extends BaseController<
...
+          hidden: false,
+        },
+      };
+    } else {
+      // We clear existing contexts for previous accounts of that group because:
+      // - Accounts might have been removed
+      // - Accounts context mapping will get updated below
+      for (const accountId of group.accounts) {
+        this.#accountIdToContext.delete(accountId);
+      }
```
fabiobozzo: another nitpick
```suggestion
      group.accounts.forEach(accountId => this.#accountIdToContext.delete(accountId));
```

packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -1,22 +1,12 @@
...
 import type { AccountGroupObjectOf } from '../group';
-import { BaseRule, type Rule, type RuleResult } from '../rule';
+import { BaseRule } from '../rule';
 import type { AccountWalletObjectOf } from '../wallet';
 
-export class EntropyRule
-  extends BaseRule
-  implements Rule<AccountWalletType.Entropy, AccountGroupType.MultichainAccount>
-{
+export class EntropyRule extends BaseRule {
```
fabiobozzo: Naming uncertainty here 🙃 the class no longer implements `Rule` interface so.. should we rename it to smth like `EntropyRuleHelper` ? Or if the messenger is not needed anymore we could even remove `extends BaseRule` ? 💭

packages/account-tree-controller/src/rules/entropy.ts
```diff
@@ -1,22 +1,12 @@
...
   readonly walletType = AccountWalletType.Entropy;
```
fabiobozzo:
```suggestion
  /** @deprecated - kept for naming compatibility, no longer used for account matching */
  readonly walletType = AccountWalletType.Entropy;
```

packages/account-tree-controller/tests/mockMessenger.ts
```diff
@@ -63,6 +65,7 @@ export function getAccountTreeControllerMessenger(
       'UserStorageController:performBatchSetStorage',
       'AuthenticationController:getSessionProfile',
       'MultichainAccountService:createMultichainAccountGroup',
+      'MultichainAccountService:getMultichainAccountWallets',
```
fabiobozzo, line 68: Don't we need to mock that action handler too in the test setup? 
```
messenger.registerActionHandler(
  'MultichainAccountService:getMultichainAccountWallets',
  jest.fn().mockReturnValue([]), 
);
```

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -176,6 +194,22 @@ const MOCK_SNAP_ACCOUNT_1: Bip44Account<InternalAccount> = {
   },
 };
 
+const MOCK_SNAP_ACCOUNT_1: InternalAccount = {
```
fabiobozzo, line 197: That's NOT bip44, is it? I'm saying that bc it must match the filtering logic

fabiobozzo (commented):
LGTM ✨ 

Finally, great separation of concerns, as you anticipated me long ago! Much cleaner than it was

packages/account-tree-controller/src/rules/entropy.ts
ccharly: Ok so I've just rewrote the entire naming logic for entropy-based wallets/groups, those are just free functions now!

packages/account-tree-controller/src/rules/entropy.ts
ccharly: I just go read of that entirely!

packages/account-tree-controller/src/AccountTreeController.test.ts
ccharly: That's not nope 👍 I renamed the original `MOCK_SNAP_ACCOUNT_1` to `MOCK_SOL_ACCOUNT_1` for better consistency.

This new one is just a "normal EVM account" backed by a Snap.

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: That's a very good point, and IMO that make sense to add this `!this.initialized` check! I'll do this, thanks

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: IMO no, that would be considered a bug if we cannot read the existing wallets, so I would just keep it as-is.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -851,50 +896,58 @@ export class AccountTreeController extends BaseController<
...
-        this.state.accountTree.selectedAccountGroup;
-      let selectedAccountGroupChanged = false;
+      // If it's in the context, then `group` should be defined.
+      const group = this.state.accountTree.wallets[walletId]?.groups[groupId];
 
-      this.update((state) => {
-        const accounts =
-          state.accountTree.wallets[walletId]?.groups[groupId]?.accounts;
+      // We're only considering non-BIP-44 accounts for single account events.
+      if (group.type !== AccountGroupType.MultichainAccount) {
```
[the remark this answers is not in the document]
ccharly: I've added an `assert` right above, to prevent this behavior (even though, this should really never happen)

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: It's more like a guard to prevent any future call to this with a BIP-44 account.

So I wonder if we should turn this into an `assert` instead 🤔 WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Actually prefer my message 😁 Just went with "Account insertion" rather than "Single account insertion" though

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Went for the `assert`!

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Too defensive indeed. The events associated with this method do not allow any nullish values for the group (and the wallet used in the one associated with the group).

So they both are well defined!

PR #6677, opened 2025-09-22
fix: update start index for proposed account name for pk/hardware imports

hmalik88:
## Explanation

Currently, PK/Hardware imports were being started at `Account 2` as opposed to `Account 1`. When we apply group metadata we already have that group in state, the logic in `#applyGroupMetadata` doesn't account for the fact that the group we're applying metadata shouldn't be counted for in the following line: https://github.com/MetaMask/core/blob/main/packages/account-tree-controller/src/AccountTreeController.ts#L461. Due to the way we apply group metadata (after we've added the groups into state), we need to take the min between `proposedNameIndex` and `highestAccountNameIndex` NOT the max. This change in logic accounts for group metadata being applied in the scenario when the controller is being rehydrated through `init` and when a new pk/hardware account is added in an active session (and handled through `accountAdded`).

## Before

https://github.com/user-attachments/assets/8ab1b3c6-574f-45bc-bf3a-3a3de7b45836

## After 

https://github.com/user-attachments/assets/66e53362-e587-45d3-b7ae-43e09da17a09

## Checklist

### DIFF
--- a/packages/account-tree-controller/src/AccountTreeController.test.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.test.ts
@@ -2902,7 +2902,7 @@ describe('AccountTreeController', () => {
       // Critical assertion: should have 2 unique names (no duplicates)
       expect(uniqueNames.size).toBe(2);
 
-      // Due to optimization, names start at wallet.length, so we get "Account 3" and "Account 4"
+      // Due to optimization, names start at wallet.length, so we get "Account 1" and "Account 2"
       expect(allNames).toContain('Account 1');
       expect(allNames).toContain('Account 2');
 


packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -2903,8 +2903,8 @@ describe('AccountTreeController', () => {
       expect(uniqueNames.size).toBe(2);
 
       // Due to optimization, names start at wallet.length, so we get "Account 3" and "Account 4"
```
montelaidev:
```suggestion
      // Due to optimization, names start at wallet.length, so we get "Account 1" and "Account 2"
```

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -458,11 +458,11 @@ export class AccountTreeController extends BaseController<
       } else {
         // For other wallet types, start with the number of existing groups
         // This gives us the next logical sequential number
-        proposedNameIndex = Object.keys(wallet.groups).length;
+        proposedNameIndex = Object.keys(wallet.groups).length - 1;
       }
 
       // Use the higher of the two: highest parsed index or computed index
```
fabiobozzo, line 464: Nit: Use the **lowest** of the two: highest parsed index or computed index

PR #6679, opened 2025-09-22
fix(account-tree-controller): fix account name prefixes

ccharly:
## Explanation

Re-introduce proper prefixes based on the keyring types for non-HD keyrings.

## References

N/A

## Checklist


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -443,44 +440,42 @@ export class AccountTreeController extends BaseController<
         const nameMatch = existingGroup.metadata.name.match(/Account (\d+)$/u);
         if (nameMatch) {
           const nameIndex = parseInt(nameMatch[1], 10);
-          if (nameIndex > highestAccountNameIndex) {
-            highestAccountNameIndex = nameIndex;
+          if (nameIndex > highestNameIndex) {
+            highestNameIndex = nameIndex;
```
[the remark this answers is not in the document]
ccharly: Yes, but for now, we want to use a "global pattern" instead of restricting it to the prefix only.

Though, I think we should re-add back the old `Math.max(.length, ...)` so if the regex does not match any accounts, we still end up using the next available "natural" index (like ["Account A", "Account B"], then next "natural index" is "Account 3")

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -437,50 +452,53 @@ export class AccountTreeController extends BaseController<
...
+        nextNaturalNameIndex ?? Object.keys(wallet.groups).length,
+      );
 
       // Find a unique name by checking for conflicts and incrementing if needed
-      let nameExists: boolean;
+      let proposedNameExists: boolean;
+      let proposedName = '';
       do {
-        proposedName = rule.getDefaultAccountGroupName(proposedNameIndex);
+        proposedName = `${namePrefix} ${proposedNameIndex}`;
```
fabiobozzo, line 488: In case namePrefix is empty... wouldn't you end up w/ ` Account 2` (with a leading space) ?

packages/account-tree-controller/src/rules/keyring.ts
```diff
@@ -49,6 +49,47 @@ export function getAccountWalletNameFromKeyringType(type: KeyringTypes) {
   }
 }
 
+/**
+ * Get group name prefix from a keyring type.
+ *
+ * @param type - Keyring's type.
+ * @returns Wallet name.
+ */
+export function getAccountGroupPrefixFromKeyringType(type: KeyringTypes) {
```
fabiobozzo, line 58: Nit: perhaps we could refactor some shared inner function for `getAccountWalletNameFromKeyringType` to reduce duplication, but not so important..

packages/account-tree-controller/src/rules/keyring.ts
```diff
@@ -49,6 +49,47 @@ export function getAccountWalletNameFromKeyringType(type: KeyringTypes) {
...
+  switch (type) {
+    case KeyringTypes.simple: {
+      return 'Imported Account';
+    }
+    case KeyringTypes.trezor: {
+      return 'Trezor Account';
```
fabiobozzo, line 64: I kinda remember reading `Trezor Acc` somewhere. Was that an 'old' behavior maybe? I guess for mobile UI space constraints.

packages/account-tree-controller/src/rules/keyring.ts
ccharly: Well actually, we could 🤔 given that all those `switch` could effectively be just mappings, we could merge both of them (while still keeping those functions to handle the "unknown" case, though, I'm not even sure we need one TBH)

I can try something for this 👍

packages/account-tree-controller/src/AccountTreeController.ts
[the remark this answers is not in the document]
fabiobozzo: So that regex stays... only for HD account groups? and hw and snap's have their own prefixes. 
Is it worth to make it `^Account (\d+)$` instead? (**starting with** "Account")

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Yup, not sure how to handle that though, I think we could use `assert(namePrefix.length > 0, "Account name prefix cannot be empty")` maybe?

packages/account-tree-controller/src/rules/keyring.ts
ccharly: I checked internally, and we wanna go with "<Keyring> Account" now, and IMO that's a better pattern 👍

packages/account-tree-controller/src/AccountTreeController.ts
mathieuartu: @fabiobozzo We would actually end up with having only the index here, no?
So I think I agree @ccharly , by using an assertion we make sure this actually never happens.

packages/account-tree-controller/src/rules/keyring.ts
ccharly: I'll make a follow-up PR with all those small improvements

For now, I want the big fix to be live today hopefully 🤞

PR #6683, opened 2025-09-23
fix(account-tree-controller): add account ordering logic with account groups

hmalik88:
## Explanation

It's been periodically observed that the accounts displayed in an account group can be "out of order". This can result from two scenarios:

1. **Account Creation Race via `createNextMultichainAccountGroup`** - Account creation takes longer than expected for a certain provider and thus the faster provider gets to the account tree controller state first which can lead to a scenario where e.g. you have Solana added before an EVM account.
2. **Account Discovery** - Discovery for a certain provider might not yield any results on a certain group index and an "out of order" provider creates an account within the group first, with the "in order" provider creating its account through the alignment process.

In order to always maintain order and avoid adding complicated logic around trying to prevent a race condition or force account creation in a certain order during discovery, I added the `sortOrder` property to an account's context object. This `sortOrder` property is used in sorting a group's accounts at the insertion level.

### Added

- Types: `AccountTypeKey`.
- Map: `ACCOUNT_TYPE_TO_SORT_ORDER`. 
- Property: Added `sortOrder` to an account's context object to track its sort order.

### Pros

1. No need for complicated logic where we try to control the account creation process to eliminate out of ordered accounts in the above two scenarios at the `MultichainAccountService` level.
2. The `MultichainAccountService` doesn't care about the order of accounts and thus this should be dealt with at the `AccountTreeController` level; the controller that is responsible for UI state.
3. Easy to update the ordering we want to show in the UI through the `ACCOUNT_TYPE_TO_SORT_ORDER` map.

### Cons

1. The changes will have a slight increase in memory footprint with a new `sortOrder` property in an account's context mapping.
2. The `AccountTreeController` will have to re-calculate at every insertion of an account, which is necessary considering an account can be added in any order (cost is at best $O(n)$ and at worst $O(nlogn)$ where $n$ is the number of providers in a type of group).

## Checklist

### DIFF
--- a/packages/account-tree-controller/src/group.ts
+++ b/packages/account-tree-controller/src/group.ts
@@ -37,11 +37,11 @@ export const MAX_SORT_ORDER = 9999;
 export const ACCOUNT_TYPE_TO_SORT_ORDER: Record<KeyringAccountType, number> = {
   [EthAccountType.Eoa]: 0,
   [EthAccountType.Erc4337]: 1,
-  [BtcAccountType.P2pkh]: 2,
-  [BtcAccountType.P2sh]: 3,
-  [BtcAccountType.P2wpkh]: 4,
-  [BtcAccountType.P2tr]: 5,
-  [SolAccountType.DataAccount]: 6,
+  [SolAccountType.DataAccount]: 2,
+  [BtcAccountType.P2pkh]: 3,
+  [BtcAccountType.P2sh]: 4,
+  [BtcAccountType.P2wpkh]: 5,
+  [BtcAccountType.P2tr]: 6,
   [TrxAccountType.Eoa]: 7,
   [AnyAccountType.Account]: MAX_SORT_ORDER,
 };


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -757,6 +772,11 @@ export class AccountTreeController extends BaseController<
 
     const groupId = result.group.id;
     let group = wallet.groups[groupId];
+    const type = account.type as AccountTypeKey;
+    const accountOrder: AccountOrderTuple = [
+      AccountTypeOrder[type],
+      account.id,
+    ];
```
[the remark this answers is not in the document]
hmalik88: We will never have an account being added that doesn't have a mapping.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -643,6 +649,15 @@ export class AccountTreeController extends BaseController<
...
             accounts.splice(index, 1);
+            const accountOrder =
+              state.accountTree.wallets[walletId]?.groups[groupId].metadata
+                .accountOrder;
+            const accountOrderIndex = accountOrder.findIndex(
+              (order) => order[1] === accountId,
+            );
+            if (accountOrderIndex !== -1) {
+              accountOrder.splice(accountOrderIndex, 1);
+            }
```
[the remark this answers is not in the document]
hmalik88: We will never have an account being added that doesn't have a mapping.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -776,11 +782,15 @@ export class AccountTreeController extends BaseController<
 
     const groupId = result.group.id;
     let group = wallet.groups[groupId];
+    const type = account.type as AccountTypeKey;
+    const { id } = account;
+    const sortOrder = AccountTypeOrder[type];
```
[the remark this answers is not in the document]
hmalik88: This is a safe assumption, every account will have a type defined in `AccountTypeOrder`

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -799,13 +809,32 @@ export class AccountTreeController extends BaseController<
...
+         * and return attribution vary across engines; final ordering is covered
+         * by behavior tests. Ignoring the entire comparator avoids flaky line
+         * coverage without reducing scenario coverage.
+         */
+        (a, b) => {
+          const aSortOrder =
+            a === id ? sortOrder : this.#accountIdToContext.get(a)?.sortOrder;
+          const bSortOrder =
+            b === id ? sortOrder : this.#accountIdToContext.get(b)?.sortOrder;
+          return (aSortOrder ?? 0) - (bSortOrder ?? 0);
```
ccharly: Do we want to default to `0` here? 🤔 That means that "unknown account type" would come first, right?

packages/account-tree-controller/src/group.ts
```diff
@@ -21,6 +29,22 @@ export type AccountTreeGroupPersistedMetadata = {
   hidden?: UpdatableField<boolean>;
 };
 
+type NonGenericAccountType<T extends KeyringAccountType = KeyringAccountType> =
+  T extends `${AnyAccountType.Account}` ? never : T;
```
ccharly: Any reason why we want to exclude them? We could also have them in the map, thus avoiding some type casts (the `type as AccountTypeKey`)

We could just use a very height sort-order for the generic accounts to always have them sorted last?

I think having a `MAX_SORT_ORDER` or `LAST_SORT_ORDER` constant could be great anyway, so we can use that as fallback in case we miss a mapping.

WDYT?

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -799,13 +809,32 @@ export class AccountTreeController extends BaseController<
...
+        /* istanbul ignore next: Comparator branch execution (a===id vs b===id)
+         * and return attribution vary across engines; final ordering is covered
+         * by behavior tests. Ignoring the entire comparator avoids flaky line
+         * coverage without reducing scenario coverage.
+         */
+        (a, b) => {
+          const aSortOrder =
+            a === id ? sortOrder : this.#accountIdToContext.get(a)?.sortOrder;
+          const bSortOrder =
+            b === id ? sortOrder : this.#accountIdToContext.get(b)?.sortOrder;
```
ccharly, line 827: We have that extra `===` to avoid having to get from the map? Actually no, that's because it's not yet part of the map, right? (just wanted to double-check 😄)

hmalik88: Yes, because it's not part of the map yet. Otherwise I'd have to do the map setting in both the if and else block.

packages/account-tree-controller/src/group.ts
hmalik88: I think it's more accurate to have this type because `AnyAccountType.Account` isn't an actual account type that is used in an `InternalAccount`. `AnyAccountType.Account` bleeding into the `InternalAccount`'s `type` type is because of the `KeyringAccountStruct` being spread into the `InternalAccountStruct`.  We actually don't use it in the `InternalAccountTypes` either: https://github.com/MetaMask/accounts/blob/main/packages/keyring-internal-api/src/types.ts#L133-L141.

packages/account-tree-controller/src/AccountTreeController.ts
hmalik88: In practice, we wouldn't ever have the nullish coalescing happen since we don't have unknown account types reaching the `AccountsController`. This was more so to appease TS haha.

packages/account-tree-controller/src/group.ts
ccharly: Yes, but we could end up having those account types in a dev environment (we introduced this one to avoid having to declare the new account type when integrating new "coin"). Once the integration is good enough, we can then add a "true" `AccountType`.

So, we might actually have `AnyAccountType.Account` accounts within the tree 😅

Lastly, we're trying to move away from `InternalAccount`, since we moved most of the relevant metadata into the tree now (`InternalAccount` is basically a `KeyringAccount` + some metadata (name, import time, etc..)), and `KeyringAccount` do use `AnyAccountType.Account` already:
- https://github.com/MetaMask/accounts/blob/main/packages/keyring-api/src/api/account.ts#L87

So I guess we could add it there and give the "least priority" possible?

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: Related to my other comment regarding `AnyAccountType`, we could definitely have an `undefined` here I think 😅
- https://github.com/MetaMask/core/pull/6683#discussion_r2375351308

packages/account-tree-controller/src/group.ts
hmalik88: Ah I see ok, I'll update.

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -799,13 +808,32 @@ export class AccountTreeController extends BaseController<
...
+          return (aSortOrder ?? 0) - (bSortOrder ?? 0);
```
ccharly: I think we should default to `MAX_SORT_ORDER` instead of 0 (even thouhg it should never really happen).

```suggestion
          return (aSortOrder ?? MAX_SORT_ORDER) - (bSortOrder ?? MAX_SORT_ORDER);
```

packages/account-tree-controller/src/group.ts
```diff
@@ -21,6 +29,25 @@ export type AccountTreeGroupPersistedMetadata = {
   hidden?: UpdatableField<boolean>;
 };
 
+const MAX_SORT_ORDER = 9999;
+
+/**
+ * Order of account types.
+ */
+export const AccountTypeOrder: Record<KeyringAccountType, number> = {
```
ccharly: Given it's a constant, maybe we should name it `ACCOUNT_TYPE_TO_SORT_ORDER` instead?

packages/account-tree-controller/src/group.ts
```diff
@@ -21,6 +29,25 @@ export type AccountTreeGroupPersistedMetadata = {
...
+  [BtcAccountType.P2pkh]: 2,
+  [BtcAccountType.P2sh]: 3,
+  [BtcAccountType.P2wpkh]: 4,
+  [BtcAccountType.P2tr]: 5,
+  [SolAccountType.DataAccount]: 6,
+  [TrxAccountType.Eoa]: 7,
+  [AnyAccountType.Account]: MAX_SORT_ORDER,
+};
+
+export type AccountTypeKey = keyof typeof AccountTypeOrder;
```
ccharly: I don't know if we want to keep this type 🤔 but if we keep it, I'd name it `AccountTypeOrderKey` instead?

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1118,6 +1153,55 @@ describe('AccountTreeController', () => {
     });
   });
 
+  describe('account ordering by type', () => {
+    it('orders accounts in group according to AccountTypeOrder regardless of insertion order', () => {
+      const evmAccount = MOCK_HD_ACCOUNT_1;
+
+      const solAccount = {
+        ...MOCK_SNAP_ACCOUNT_1,
```
ccharly, line 1161: Just to make the test a bit more clear, I'd add `id: 'mock-sol-id-1'` (instead of just `mock-snap-id-1`)

ccharly (approved):
LGTM, thanks for all those changes!

PR #6687, opened 2025-09-23
feat: add gasSponsored flag to bridgeController

Battambang:
- Today, quotes can indicate when gas is embedded into the transaction itself via `gasIncluded` and when EIP‑7702 delegated execution can be used via `gasIncluded7702`, but there’s no explicit way to signal that a third party will sponsor the gas cost.
- This prevents the UI/clients from differentiating between “quote includes gas fees” and “quote is gasless to the user because gas is sponsored,” which affects UX (badging, prioritization) and any downstream decision logic.

- Added an optional boolean `quote.gasSponsored` to `QuoteSchema` in `@validators.ts` to allow the bridge API to flag gas‑sponsored quotes.
- Ensured end‑to‑end preservation:
  - Validation now accepts payloads with or without `gasSponsored`.
  - `BridgeController.fetchQuotes` and the fee‑augmentation path preserve the field unchanged so the UI and metrics layers can consume it.
- Tests:
  - Added a test in `bridge-controller.test.ts` verifying that `fetchQuotes` returns quotes where `quote.gasSponsored` is preserved when present, and remains undefined when omitted.
- Backward compatibility:
  - The field is optional; existing payloads and flows continue to work unchanged.
  - No sorting or fee calculations are altered by this addition.

- Distinction between flags:
  - `gasIncluded`: Quote includes gas fees in the trade economics, but the user still submits and pays gas.
  - `gasIncluded7702`: Quote can use delegated execution under EIP‑7702.
  - `gasSponsored`: A third party sponsors the gas so the user doesn’t pay it; this is orthogonal to whether gas is “included” in pricing or delegated via 7702.

## References

This PR is related to this BridgeAPI PR https://github.com/consensys-vertical-apps/va-mmcx-bridge-api/pull/527.
PR related for MM Extension: https://github.com/MetaMask/metamask-extension/pull/36227.
PR related for MM Mobile: https://github.com/MetaMask/metamask-mobile/pull/20878.

## Checklist

---

> [!NOTE]
> Introduce optional `quote.gasSponsored` and ensure `fetchQuotes` returns it unchanged; update tests and changelog.
> 
> - **Schema/Validation**:
>   - Add optional `quote.gasSponsored` to `QuoteSchema` in `src/utils/validators.ts`.
> - **Controller Tests**:
>   - Add helper `makeQuoteRequest` and a test ensuring `fetchQuotes` preserves `quote.gasSponsored` when present in responses (`bridge-controller.test.ts`).
> - **Changelog**:
>   - Note that quotes returned by `fetchQuotes` now include `gasSponsored`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5d9aafb20417aa8e103e3eed5a50d2e332089aa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-controller/src/bridge-controller.test.ts
+++ b/packages/bridge-controller/src/bridge-controller.test.ts
@@ -2872,10 +2872,7 @@ describe('BridgeController', function () {
       } as QuoteResponse;
       const secondQuote: QuoteResponse =
         mockBridgeQuotesNativeErc20Eth[1] as QuoteResponse;
-      const quotesWithFlag: QuoteResponse[] = [
-        firstQuoteWithFlag,
-        secondQuote,
-      ];
+      const quotesWithFlag: QuoteResponse[] = [firstQuoteWithFlag, secondQuote];
 
       const fetchBridgeQuotesSpy = jest
         .spyOn(fetchUtils, 'fetchBridgeQuotes')


packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -2727,6 +2727,189 @@ describe('BridgeController', function () {
...
+
+      const quotes = await bridgeController.fetchQuotes({
+        srcChainId: 1,
+        destChainId: 10,
+        srcTokenAddress: '0x0000000000000000000000000000000000000000',
+        destTokenAddress: '0x0000000000000000000000000000000000000000',
+        srcTokenAmount: '1000',
+        walletAddress: '0x123',
+        slippage: 0.5,
+      } as never);
```
mcmire: Nit: I guess this is an existing issue, but why are we type-casting this input? Is there something we are leaving out intentionally? As it is we are not really testing this in a more real-world way, we are completely bypassing TypeScript.

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `gasSponsored` field to QuoteSchema to support gas fees sponsored ([#6687](https://github.com/MetaMask/core/pull/6687))
```
mcmire: Nit: It looks like `QuoteSchema` isn't exported. Should we describe this in a way that makes sense from the perspective of the public API? Maybe we can say that quotes returned by `fetchQuotes` now include a `gasFeeSponsored` property?

``` suggestion
- Quotes as returned by `fetchQuotes` now include a `gasSponsored` property ([#6687](https://github.com/MetaMask/core/pull/6687))

**Review by mcmire — DISMISSED** [review]

File: packages/bridge-controller/src/utils/validators.ts [hunk 3]
```diff
@@ -228,6 +228,7 @@ export const QuoteSchema = type({
       totalFeeAmountUsd: optional(string()),
     }),
   ),
+  gasSponsored: optional(boolean()),
```
stanleyyconsensys, line 250: nit

please update the comment for gasIncluded7702, as gas sponser will also set this field to true

stanleyyconsensys: please double confirm 

if gasIncluded7702 is true and gasSponsored is true

what happen to the extension when it is not ready support that from UI?

packages/bridge-controller/src/bridge-controller.test.ts
Battambang: Unit test refactored to simplify by reusing an existing quote request and by removing type casting.

packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -2727,6 +2743,36 @@ describe('BridgeController', function () {
...
+          validationFailures: [],
+        });
+
+      const quotes = await bridgeController.fetchQuotes(makeQuoteRequest());
+
+      expect(fetchBridgeQuotesSpy).toHaveBeenCalledTimes(1);
+      expect(quotes).toHaveLength(2);
+      expect(quotes[0].quote.gasSponsored).toBe(true);
+      expect(quotes[1].quote.gasSponsored).toBeUndefined();
+    });
```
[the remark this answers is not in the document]
Battambang: This is just a comment user information, no enforcement is applied on the field at the controller level, the logic of setting the fields is done on the BridgeAPI side.

PR #6693, opened 2025-09-23
fix: fix added popular network

salimtb:
## Explanation

The current `#onAddNetwork` behavior in `NetworkEnablementController` was inconsistent and didn't provide a good user experience:
- All newly added networks were automatically enabled, regardless of context
- The logic for determining "popular networks mode" required ALL popular networks to be enabled, which was impractical
- Users experienced unexpected network switching when adding popular networks in multi-network scenarios

This PR implements a more intuitive network addition behavior:

1. **Improved Popular Networks Detection**: Changed from requiring "all popular networks enabled" to ">2 popular networks enabled", making the detection much more practical and user-friendly.

2. **Context-Aware Network Addition**:
   - **If in popular networks mode (>2 popular networks enabled) AND adding a popular network** → Keep current selection (add but don't enable the new network)
   - **Otherwise** → Switch to the newly added network (disable all others, enable the new one)

3. **Key Benefits**:
   - When users have multiple popular networks enabled and add another popular network, their current workflow isn't disrupted
   - When users add non-popular networks or add networks when not in "popular mode", they get the expected behavior of switching to the new network
   - The popular networks threshold (>2) is much more realistic than requiring all popular networks

**Are there any changes whose purpose might not obvious to those unfamiliar with the domain?**

- The method rename from `#areAllPopularNetworksEnabled()` to `#isInPopularNetworksMode()` reflects the conceptual shift from checking all networks to checking if the user is in a "popular networks" usage pattern
- The ">2" threshold was chosen because it indicates the user is actively using multiple popular networks (suggesting they prefer the multi-network mode) vs. having just 1-2 networks enabled

## References

This change improves the user experience for network management in MetaMask by implementing more intuitive network addition behavior based on user context.

## Checklist

- [ ] I've prepared draft p

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "574.0.0",
+  "version": "575.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -7,10 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add new group naming for non-HD keyring accounts ([#6679](https://github.com/MetaMask/core/pull/6679))
+  - Hardware-wallet account groups are now named: "Ledger|Trezor|QR|Lattice|OneKey Account N".
+  - Private key account groups are now named: "Imported Account N".
+  - Snap account groups are now named: "Snap Account N".
+- Account group names now use natural indexing as a fallback ([#6677](https://github.com/MetaMask/core/pull/6677)), ([#6679](https://github.com/MetaMask/core/pull/6679))
+  - If a user names his accounts without any indexes, we would just use the number of accounts to compute the next available index.
+
 ### Fixed
 
-- Fix group naming for PK/Hardware accounts ([#6677](https://github.com/MetaMask/core/pull/6677))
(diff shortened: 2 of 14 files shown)


packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -446,6 +446,42 @@ export class NetworkEnablementController extends BaseController<
...
+
+        const { namespace, storageKey } = deriveKeys(chainId as Hex);
+        const isEnabled = this.state.enabledNetworkMap[namespace]?.[storageKey];
+        return isEnabled ? count + 1 : count;
+      },
+      0,
+    );
+
+    // Return true if more than 2 popular networks are enabled
+    return enabledPopularNetworksCount > 2;
```
Prithpal-Sooriya: I think we can clean this up by using a .filter.length.

```ts
const enabledPopularNetworks = POPULAR_NETWORKS.filter(chainId => {
  // check network controller state that network exists
  if (!networkControllerState.networkConfigurationsByChainId[chainId as Hex]) {
    return false;
  }
  
  // check network enable map
  const { namespace, storageKey } = deriveKeys(chainId as Hex);
  return Boolean(this.state.enabledNetworkMap[namespace]?.[storageKey])
})

return enabledPopularNetworks.length > 2
```

Prithpal-Sooriya: This `> 2 ` is a little odd, why is this the case? Is this because non-EVM?
Would it be safer to say > 1?

salimtb: yes it's because of non evm indeed

PR #6696, opened 2025-09-23
feat(account-tree-controller): allow more pattern when extracting highest index from account group names

ccharly:
## Explanation

Now allows a bit more pattern when extracting account group name index.

## References

N/A

## Checklist

### DIFF
--- a/packages/network-enablement-controller/CHANGELOG.md
+++ b/packages/network-enablement-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -452,7 +452,8 @@ export class AccountTreeController extends BaseController<
           continue;
         }
         // Parse the existing group name to extract the numeric index
-        const nameMatch = existingGroup.metadata.name.match(/Account (\d+)$/u);
+        const nameMatch =
+          existingGroup.metadata.name.match(/account\s+(\d+)$/iu);
```
mathieuartu, line 456: Is there a reason we're relaxing the case sensitivity here?

ccharly: I just checked internally and if I understood correctly, we're ok to relax this so we can match "more" patterns.

Any concern with this?

ccharly: As discussed internally, we allow those patterns now and we no longer reserve "Account N" names 👍

PR #6703, opened 2025-09-23
feat: remove `isMultichainOrigin` guard from `wallet_invokeMethod` handler

jiexi:
## Explanation

Currently it is possible for the window.ethereum provider to grant solana accounts and scopes in its permission requests. Whether or not that should be allowed UX is up for debate. Regardless, it causes an issue where if Solana accounts and scopes are granted AND a dapp also uses our solana provider, the dapp will be unable to use the solana provider to make requests without first re-requesting solana accounts from the solana provider (not the EVM provider). This is because the permission granted via window.ethereum has `isMultichainOrigin: false` where as the solana provider's granted permissions go through the multichain api which will have them set `true` which then gets caught in this `isMultichainOrigin` guard in `wallet_invokeMethod` handler.

The original purpose of this guard was to make the multichain api granted permissions equivalent to a window.ethereum set of permissions, but not the other way around. Trying to encourage multichain api usage over window.ethereum usage in this manner doesn't really make sense / is not worth this hassle anymore.

## Checklist

### DIFF
--- a/packages/multichain-api-middleware/CHANGELOG.md
+++ b/packages/multichain-api-middleware/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `wallet_invokeMethod` requests no longer fail with unauthorized error if the `isMultichainOrigin` property is false on the requesting origin's CAIP-25 Permission.
 
-1
 ## [1.1.0]
 
 ### Changed


packages/multichain-api-middleware/src/handlers/wallet-invokeMethod.ts
```diff
@@ -80,7 +80,7 @@ async function walletInvokeMethodHandler(
   } catch {
     // noop
   }
-  if (!caveat?.value?.isMultichainOrigin) {
+  if (!caveat) {
```
adonesky1, line 83: 🤔 This seems like a strange conditional to leave dangling. Why wouldn't we just remove it altogether?

jiexi: it's because we still to make sure the permission/caveat itself actually exists still. Previously the `!caveat?.value?.isMultichainOrigin` was pulling double duty

adonesky1: hmmm probably not worth it here but we throw the same error below in the next block which does checks against the caveat so could consolidate:
```
 const scopeObject = getSessionScopes(caveat.value, {
    getNonEvmSupportedMethods: hooks.getNonEvmSupportedMethods,
  })[scope];

  if (!scopeObject?.methods?.includes(wrappedRequest.method)) {
    return end(providerErrors.unauthorized());
  }
 ```
 
 At this point this check is mostly a type check

Comment (jiexi) [reply packages/multichain-api-middleware/src/handlers/wallet-invokeMethod.ts:83]: this might blow up on `getSessionScopes(undefined)`?

Comment (adonesky1) [reply packages/multichain-api-middleware/src/handlers/wallet-invokeMethod.ts:83]: yeah exactly we aren't handling it gracefully downstream. 

I suppose it just feels silly that this is just a typecheck now... but whatever its fine.

Maybe?
```suggestion
  if (!caveat?.value) {
```

adonesky1: or below we could add the optional chaining:
```
const scopeObject = getSessionScopes(caveat?.value,
```

PR #6709, opened 2025-09-24
feat(account-tree-controller): force-init before any account insertion/removal

ccharly:
## Explanation

We now enforce that `init` got called before making any change to the tree to avoid having out-of-order account groups.

I stumbled upon this bug when running E2E on mobile. Basically, the way the fixture are setup on mobile were triggering `:accountAdded` before calling `AccountTreeController.init`, which resulted in having "Account 2" inserted before "Account 1".

But, since "Account 1" was already part of the initial `internalAccounts` it should have been inserted before any new accounts.

Thus, forcing `init` to be called implicitly fixes this order/sequence.

## References

E2E were failing on this PR, because of this bug:
- https://github.com/MetaMask/metamask-mobile/pull/20255

## Checklist

### DIFF
--- a/packages/account-tree-controller/src/AccountTreeController.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.ts
@@ -339,7 +339,7 @@ export class AccountTreeController extends BaseController<
    * Re-initialize the controller's state.
    *
    * This is done in one single (atomic) `update` block to avoid having a temporary
-   * cleared state.
+   * cleared state. Use this when you need to force a full re-init even if already initialized.
    */
   reinit() {
     this.#initialized = false;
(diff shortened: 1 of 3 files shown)


owencraston (approved):

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -320,6 +331,28 @@ export class AccountTreeController extends BaseController<
...
     }
+
+    this.#initialized = true;
+  }
+
+  /**
+   * Re-initialize the controller's state.
+   *
+   * This is done in one single (atomic) `update` block to avoid having a temporary
+   * cleared state.
```
fabiobozzo:
```suggestion
   * cleared state. Use this when you need to force a full re-init even if already initialized.
```

fabiobozzo (approved):

PR #6717, opened 2025-09-24
fix(account-tree-controller): prevent `:account{Added,Removed}` if `init` has not been called yet

ccharly:
## Explanation

I initially thought calling `init` implicitly would solve some issues, but it actually adds one when it comes to automatically migrating accounts from v1 state.

In the clients we usually use the tree like so:
```ts
this.accountsController.updateAccounts();
this.accountTreeController.init();
```

The `updateAccounts` part re-creates the list of internal accounts and potentially migrate them if needed (in the case of BIP-44, this is needed to inject the `options.entropy` for those accounts).

If a `KeyingController:stateChange` happens before `updateAccounts`, this might trigger some `:accountAdded` event, but some accounts might not have been automatically migrated yet (since `updateAccounts` has not been called at this point). Thus, this can lead the `AccountTreeController` to misuse those accounts and put them in wrong wallet (like older HD accounts had not `options.entropy` objects, thus they are not considered BIP-44 accounts yet).

To avoid such issues (which seems to mainly happen for E2E), we wait for `init` to have been called explicitly. 

In anyway, calling `updateAccounts` right before `init` would make sure that the list of accounts is up-to-date AND migrated.

## Checklist


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -664,33 +655,32 @@ export class AccountTreeController extends BaseController<
...
 
-      this.messagingSystem.publish(
-        `${controllerName}:accountTreeChange`,
-        this.state.accountTree,
-      );
-    }
+    this.messagingSystem.publish(
+      `${controllerName}:accountTreeChange`,
+      this.state.accountTree,
+    );
```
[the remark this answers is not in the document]
ccharly: We did not have this check at first, but I think we can keep it in case `:accountAdded` got sent twice for whatever reasons.

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -1215,6 +1215,26 @@ describe('AccountTreeController', () => {
         },
       } as AccountTreeControllerState);
     });
+
+    it('does not remove account if init has not been called', () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mathieuartu: Is this necessary? Should we not include `controller` in the destructured properties?

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Prevent `:account{Added,Removed}` to used if `init` has not been called yet ([#6717](https://github.com/MetaMask/core/pull/6717))
```
mathieuartu:
```suggestion
- Prevent `:account{Added,Removed}` to be used if `init` has not been called yet ([#6717](https://github.com/MetaMask/core/pull/6717))
```

packages/account-tree-controller/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - We now wait for internal accounts to be fully ready (before `init` gets called).
```
mathieuartu: Is this line relevant for this `core` change? Seems more client-side related. Maybe something like this:

 ```suggestion
   - We now wait for `init` to have been called at least once. Clients will need to ensure internal accounts are fully ready before calling `init`.
```

packages/account-tree-controller/CHANGELOG.md
mathieuartu: I'm writing this comment last, and writing the other comments before made me realize that this might be a breaking change?

ccharly: This could be seen as a breaking change in terms of behavior, but I actually think the current behavior is not valid either 😅

Also, "we" are the consumer the `AccountsController` and clients don't really have control over those `:account{Added,Removed}` event strictly speaking. I think it's implied that `init` has to be called before being able to use the tree anyway.

Lastly, making this a "true breaking change" would quickly bubble up to other controllers, while the public interface remains the same.

But I get your point and I'm on a fence here 😅 for simplicity, I'd would just treat it as a fix (given it does not require any change as of now from any consumers).

Would that be ok?

packages/account-tree-controller/src/AccountTreeController.test.ts
ccharly: Lemme test without it! Always fear that it would be garbage collected right away if no one holds a ref onto it 😅

But that's probably fine without the ref?

ccharly: Decided to change it for: https://github.com/MetaMask/core/pull/6717/commits/5c3c45117e0f40dc29bc608ab141abf30e13b931

packages/account-tree-controller/CHANGELOG.md
mathieuartu: I agree with all your points! Thanks for elaborating ❤️ Yep that'd be alright with me!

mathieuartu (approved):

PR #6724, opened 2025-09-25
add: error field from patch

bergarces:
## Explanation

- Adds the error field from the existing patch.
- Applies comments from previous PR to fix field descriptions

## References

Related to https://consensyssoftware.atlassian.net/browse/ASSETS-1308

## Checklist

### DIFF
--- a/packages/assets-controllers/src/NftController.ts
+++ b/packages/assets-controllers/src/NftController.ts
@@ -79,40 +79,18 @@ type SuggestedNftMeta = {
   origin: string;
 };
 
-/**
- * Nft
- *
- * NFT representation
- *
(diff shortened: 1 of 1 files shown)


packages/assets-controllers/src/NftController.ts
```diff
@@ -211,30 +208,29 @@ export type NftMetadata = {
   rarityRank?: string;
   topBid?: TopBid;
   chainId?: number;
+  error?: string;
```
bergarces, line 211: This was in the patch but missing in the original PR.

I thought it wasn't needed as there was no error running with this version, but it's used in a component.

PR #6729, opened 2025-09-25
feat(multichain-account-service): add logs

ccharly:
## Explanation

Adding more debug logs to the service so we can diagnose potential issues better (since it's not always easy to guess what happened only with states)

## References

N/A

## Checklist

### DIFF
--- a/packages/multichain-account-service/src/MultichainAccountGroup.ts
+++ b/packages/multichain-account-service/src/MultichainAccountGroup.ts
@@ -239,7 +239,7 @@ export class MultichainAccountGroup<
         const accounts = this.#providerToAccounts.get(provider);
         if (!accounts || accounts.length === 0) {
           this.#log(
-            `Found missing accounts for account provider "${provider.getName()}", createing them now...`,
+            `Found missing accounts for account provider "${provider.getName()}", creating them now...`,
           );
           const created = await provider.createAccounts({
             entropySource: this.wallet.entropySource,


packages/multichain-account-service/src/tests/providers.ts
```diff
@@ -58,6 +58,7 @@ export function setupNamedAccountProvider({
       // Assuming this never fails.
       getAccounts().find((account) => account.id === id),
   );
+  mocks.createAccounts.mockResolvedValue([]);
```
ccharly, line 61: This was required for the coverage, now that we log the number of accounts that got created.

packages/multichain-account-service/src/MultichainAccountGroup.test.ts
```diff
@@ -166,7 +166,7 @@ describe('MultichainAccount', () => {
     });
   });
 
-  describe('align', () => {
+  describe('alignAccounts', () => {
```
ccharly, line 169: Spotted this one when checking for tests.

packages/multichain-account-service/src/logger.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable jsdoc/require-jsdoc */
+
+import { createProjectLogger, createModuleLogger } from '@metamask/utils';
+
+export const projectLogger = createProjectLogger('multichain-account-service');
+
+export { createModuleLogger };
+
+export function warn(message: string) {
```
mathieuartu: I find this name a bit confusing, especially since you import `projectLogger as log`. One could imagine that `warn` would achieve the same result as `log`. WDYT of renaming this to `addWarningPrefix`?

ccharly: That's a good point 🤔 Lemme try to find something else!

ccharly: Actually, replaced that by a `WARNING_PREFIX` constant

PR #6734, opened 2025-09-26
chore: Adds Tron network to network enablement controller

zone-live:
## Explanation

Adds Tron network to network enablement controller.
Makes use of the non-evm scopes via the `@metamask/keyring-api`.

## Checklist


packages/network-enablement-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add Tron network support ([#6734](https://github.com/MetaMask/core/pull/6734))
```
Prithpal-Sooriya, line 12: If poss, can you expand on the dev changes.
- Adding the Tron namespace to the enabledNetworkMap state.
- Also adding the keyring API to reuse types?

packages/network-enablement-controller/package.json
```diff
@@ -63,6 +63,7 @@
   "dependencies": {
     "@metamask/base-controller": "^8.4.0",
     "@metamask/controller-utils": "^11.14.0",
+    "@metamask/keyring-api": "^21.0.0",
```
Prithpal-Sooriya, line 66: Q - are we using this library just for the types? If this is the case we could move this into a dev-dependency?

packages/network-enablement-controller/CHANGELOG.md
zone-live: Yap, will do!

packages/network-enablement-controller/package.json
zone-live: No, not just as types, but as values as well in the ctrl.

PR #6736, opened 2025-09-26
feat(user-storage-controller): use deferred promises cache for KDF operations

mathieuartu:
## Explanation

```md
- Use deferred promises for encryption/decryption KDF operations ([#6736](https://github.com/MetaMask/core/pull/6736))
  - That will prevent duplicate KDF operations from being computed if one with the same options is already in progress.
  - For operations that already completed, we use the already existing cache.
```

This improves startup performance on mobile by ~70%; going from an average of ~6s to an average of ~2s for the time it takes to derive the key.

## References

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-1100

## Checklist

---

> [!NOTE]
> Deduplicates concurrent scrypt KDF operations via a bounded promise cache, adds concurrency/eviction tests, and updates changelog.
> 
> - **Encryption (`src/shared/encryption`)**:
>   - `encryption.ts`: Add deferred KDF promise cache to avoid duplicate scrypt derivations; cap size using `MAX_KDF_PROMISE_CACHE_SIZE`; add cache-key generation and `#performKdfOperation` helper; minor refactors to use `salt` consistently.
>   - `constants.ts`: Add `MAX_KDF_PROMISE_CACHE_SIZE`.
> - **Tests**:
>   - `encryption.test.ts`: Add concurrent encrypt/decrypt tests (same/different passwords), load tests, and cache size eviction test.
> - **Changelog**:
>   - Note use of deferred promises for KDF to prevent duplicate work.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b179d180bf179911c649bd916329ac028917023. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -250,33 +258,85 @@ class EncryptorDecryptor {
...
+
+  #createKdfCacheKey(
+    hashedPassword: string,
+    o: EncryptedPayload['o'],
+    salt: Uint8Array,
+    nativeScryptCrypto?: NativeScrypt,
+  ): string {
+    const saltStr = byteArrayToBase64(salt);
+    const hasNative = Boolean(nativeScryptCrypto);
+    return `${hashedPassword}:${o.N}:${o.r}:${o.p}:${o.dkLen}:${saltStr}:${hasNative}`;
```
fabiobozzo, line 315: Maybe weird thought: can `hashedPassword` contain `:` ? would that cause key collisions? 🤔

packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -250,33 +258,85 @@ class EncryptorDecryptor {
...
+      o,
+      newSalt,
+      nativeScryptCrypto,
+    );
+
+    // Check if there's already an ongoing KDF operation with the same parameters
+    const existingPromise = this.#kdfPromiseCache.get(cacheKey);
+    if (existingPromise) {
+      return existingPromise;
+    }
```
fabiobozzo, line 275:
```suggestion
    if (existingPromise) {
      return existingPromise.catch(() => {
        // If shared promise failed, retry individually
        this.#kdfPromiseCache.delete(cacheKey);
        return this.#performKdfOperation(password, o, newSalt, hashedPassword, nativeScryptCrypto);
  });
    }
```
What happens if the existing promise rejects? Would all concurrent operations fail? Do we need a fallback mechanism?

packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -250,33 +258,85 @@ class EncryptorDecryptor {
...
+    const kdfPromise = this.#performKdfOperation(
+      password,
+      o,
+      newSalt,
+      hashedPassword,
+      nativeScryptCrypto,
+    );
+
+    // Cache the promise and set up cleanup
+    this.#kdfPromiseCache.set(cacheKey, kdfPromise);
```
fabiobozzo, line 296: Although it's unlikely if promises never resolve or reject the map grows indefinitely, doesn't it? 

Do we need to manage overflow like
```suggestion
  if (this.#kdfPromiseCache.size >= this.#MAX_CACHE_SIZE) {
    this.#kdfPromiseCache.clear();
  }

    this.#kdfPromiseCache.set(cacheKey, kdfPromise);
```
or does `set` already take care of it internally or smth like that?

packages/profile-sync-controller/src/shared/encryption/encryption.ts
gantunesr: This is a sha256 hash which should be a hex string therefore it'll be a (big) error if it contains `:`

gantunesr (commented):
Looks good, will approve after the comments are resolved

packages/profile-sync-controller/src/shared/encryption/encryption.ts
mathieuartu: Encryption errors are usually silenced in the user storage SDK so that the entries are either not returned or not sent. Nothing changes in terms of outcome here, it's the same as now: If there's an error, this will be tried again on the next encryption / decryption attempt.

packages/profile-sync-controller/src/shared/encryption/encryption.ts
mathieuartu: Good point, I'll make a change in that direction :) Thanks!

mathieuartu: Done with [309edae](https://github.com/MetaMask/core/pull/6736/commits/309edae4a4b7dd9de542ff85bc1c58dd46f9afb2)

packages/profile-sync-controller/src/shared/encryption/encryption.test.ts
```diff
@@ -112,4 +112,132 @@ describe('encryption tests', () => {
...
+        encryption.decryptString(encrypted, password).then((decrypted) => {
+          expect(decrypted).toBe(`${baseData}-${i}`);
+          return decrypted;
+        }),
+      );
+
+      await Promise.all(decryptPromises);
+
+      // Log performance for manual verification (deferred promises should make this faster)
+      console.log(`Concurrent operations completed in ${duration}ms`);
```
ccharly: Super nit: But it could have been nice to have the old timing (with this test) and the new ones, so we know what kind of speedup factor we can expect with the new deferred promises. WDYT?

packages/profile-sync-controller/src/shared/encryption/encryption.test.ts
```diff
@@ -112,4 +112,132 @@ describe('encryption tests', () => {
...
+    });
+  });
```
ccharly, line 268: Nit: I wonder if we should have added a test that use operations that would trigger the cache eviction, and see that it still behaves "normally"?

cmd-ob (approved):

packages/profile-sync-controller/src/shared/encryption/encryption.ts
```diff
@@ -250,33 +261,94 @@ class EncryptorDecryptor {
...
+    }
+
+    // Limit cache size to prevent unbounded growth
+    if (this.#kdfPromiseCache.size >= this.#MAX_KDF_PROMISE_CACHE_SIZE) {
+      // Remove the oldest entry (first inserted)
+      const firstKey = this.#kdfPromiseCache.keys().next().value;
+      if (firstKey) {
+        this.#kdfPromiseCache.delete(firstKey);
+      }
+    }
```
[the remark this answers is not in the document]
ccharly: 1. Regarding the cache eviction logic, yes that's true, we could also use `lodash.memoize` for this to cache the key once ready, but that's beyond the scope of this PR IMO
2. Even if we remove the promise from the cache, consumers already `await`ing it, will still hold a ref on this promise, thus, it won't invalidate their calls, so I don't think this is a problem!

packages/profile-sync-controller/src/shared/encryption/encryption.test.ts
mathieuartu: I removed this timer as this was a bit irrelevant

packages/profile-sync-controller/src/sdk/user-storage.test.ts
```diff
@@ -23,6 +21,7 @@ import encryption, { createSHA256Hash } from '../shared/encryption';
 import { SHARED_SALT } from '../shared/encryption/constants';
 import { Env } from '../shared/env';
 import { USER_STORAGE_FEATURE_NAMES } from '../shared/storage-schema';
+import type { UserStorageGenericFeatureKey } from 'src/shared/storage-schema';
```
ccharly: Shouldn't this be `../shared/storage-schema`? I guess `src/...` probably works too

packages/profile-sync-controller/src/shared/encryption/encryption.test.ts
```diff
@@ -112,4 +113,157 @@ describe('encryption tests', () => {
...
+        // Use different passwords to create unique cache keys
+        const uniquePassword = `cache-test-${i}`;
+        promises.push(encryption.encryptString('test-data', uniquePassword));
+      }
+
+      // All operations should complete successfully despite cache limit
+      const results = await Promise.all(promises);
+      expect(results).toHaveLength(numOperations);
+
+      // Verify a sampling of results can be decrypted (testing all 25 would be slow)
```
ccharly, line 253: Nit:

```suggestion
      // Verify a sampling of results can be decrypted (testing 25 would be slow)
```

ccharly (approved):

ccharly (approved):

PR #6743, opened 2025-09-28
fix: improve balance fetch and not reset to 0

salimtb:
### Current State and Problem
Previously, the `TokenBalancesController` would update token balances, native balances, and staked balances regardless of whether the values had actually changed. This led to several inefficiencies:

1. **Unnecessary state updates**: The controller would trigger state changes even when balance values remained the same
2. **Performance overhead**: Unchanged balances would still go through the entire update pipeline, causing unnecessary re-renders and processing
3. **Reset behavior**: Balances could be reset to zero and then immediately updated with the same value, creating unnecessary intermediate states

### Solution
This PR implements value comparison logic to skip updates when balances haven't actually changed:

1. **Token Balance Updates**: Added comparison between `currentBalance` and `newBalance` before updating the state. Only updates are applied when values differ, preventing unnecessary state mutations.

2. **Native Balance Updates**: Enhanced the filter logic to compare existing balances from `AccountTrackerController` state with new balance values. Only balances that have actually changed are included in the update batch sent to `AccountTrackerController:updateNativeBalances`.

3. **Staked Balance Updates**: Similar optimization for staked balances, comparing current staked balance values with new ones before including them in the update batch sent to `AccountTrackerController:updateStakedBalances`.

### Key Improvements
- **Reduced State Mutations**: Prevents unnecessary state updates when balance values are identical
- **Better Performance**: Eliminates redundant processing and re-renders caused by unchanged balance updates  
- **Cleaner State Management**: Avoids intermediate states where balances are reset to zero before being set to the same value
- **Batch Optimization**: Only sends actual changes to the AccountTracker, reducing the payload size and processing overhead

### Technical Details
The implementation uses `isEqual` comparison for the main state update check and direct value comparison (`currentBalance !== newBalance`) for individual balance updates. This ensures that both the overall state update and individual balance updates are optimized to only occur when necessary.

## Checklist

---

> [!NOTE]
> Avoids unnecessary token/native/staked balance updates by comparing existing values and only updating when changed, reducing state mutations and calls.
> 
> - **Fixed: Balance update churn**
>   - `TokenBalancesController`
>     - Initialize missing entries without overwriting existing balances; only set balances when values change.
>     - Filter `updateNativeBalances`/`updateStakedBalances` calls by diffing against `AccountTrackerController:getState` and only send changed entries.
>     - Add `AccountTrackerController:getState` to allowed actions.
>   - `AccountTrackerController`
>     - Optimize `updateNativeBalances` and `updateStakedBalances` to clone state, compare current values, and call `update` only if changes exist.
>   - **Tests**: Permit and stub `AccountTrackerController:getState`; update messenger setups accordingly.
>   - **Changelog**: Note fix for unnecessary balance updates to improve performance.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 460183d171ac0f3f470cb8976d743cecdec2b681. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -702,26 +702,46 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
   updateNativeBalances(
     balances: { address: string; chainId: Hex; balance: Hex }[],
   ) {
-    this.update((state) => {
-      balances.forEach(({ address, chainId, balance }) => {
-        const checksumAddress = toChecksumHexAddress(address);
+    const nextAccountsByChainId = cloneDeep(this.state.accountsByChainId);
```
Prithpal-Sooriya, line 705: nit - logic is very similar in both areas, can we refactor to reuse some logic?

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -639,18 +653,34 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
...
-          address: balance.account,
-          chainId: balance.chainId,
-          balance: balance.value ? BNToHex(balance.value) : '0x0',
-        }));
-
-        this.messagingSystem.call(
-          'AccountTrackerController:updateNativeBalances',
-          balanceUpdates,
-        );
+        const balanceUpdates = nativeBalances
```
Prithpal-Sooriya, line 663: nit - same, this logic is very similar in both places. Is it worth refactoring this to reuse this logic?

PR #6758, opened 2025-09-30
fix(account-tree-controller): re-use computed names for groups

ccharly:
## Explanation

Re-enable computed names for account group names.

This allow to automatically migrate previous EVM names back to their account group. This won't apply for newly added accounts though.

Once account groups are named at least once, they won't be renamed again, so that should be enough to not rely on any migrations for this.

## References

N/A

## Checklist


ccharly:
@metamaskbot publish-preview

PR #6759, opened 2025-09-30
feat(multichain-account-service): add option to decide what to await for when creating group

mathieuartu:
## Explanation

```md
- Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()`
  - Introduces `options.waitForAllProvidersToFinishCreatingAccounts`, that will make `createMultichainAccountGroup` await either only the EVM provider or all the providers to have created their accounts depending on the value. Defaults to `false` (only awaits for EVM accounts creation by default).
```

## Checklist

---

> [!NOTE]
> Adds an `options.waitForAllProvidersToFinishCreatingAccounts` flag to `createMultichainAccountGroup`, defaulting to awaiting only EVM, and updates `createNextMultichainAccountGroup` to await all.
> 
> - **Multichain Account Service**:
>   - **`createMultichainAccountGroup`**:
>     - Add optional `options` with `waitForAllProvidersToFinishCreatingAccounts` (default `false`).
>     - If `true`: await all providers via `Promise.allSettled`; throw on any failure with aggregated warnings.
>     - If `false` (default): await EVM provider; start other providers in background, log but ignore their errors.
>     - Update JSDoc for parameters and error semantics.
>   - **`createNextMultichainAccountGroup`**: now calls `createMultichainAccountGroup` with `waitForAllProvidersToFinishCreatingAccounts: true`.
>   - **Tests**: add coverage for failure when waiting for all providers; adjust non-EVM background behavior test.
>   - **Changelog**: document new `options` parameter and behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 403d744d5c9890ea967898e50202dfb07dc7cca9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/multichain-account-service/CHANGELOG.md
+++ b/packages/multichain-account-service/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()` ([#]())
+- Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()` ([#6759](https://github.com/MetaMask/core/pull/6759))
   - Introduces `options.waitForAllProvidersToFinishCreatingAccounts`, that will make `createMultichainAccountGroup` await either only the EVM provider or all the providers to have created their accounts depending on the value. Defaults to `false` (only awaits for EVM accounts creation by default).
 - Only await for EVM account creation in `MultichainAccountWallet.createMultichainAccountGroup()` instead of all types of providers ([#6755](https://github.com/MetaMask/core/pull/6755))
   - Other type of providers will create accounts in the background and won't throw errors in case they fail to do so.
--- a/packages/multichain-account-service/CHANGELOG.md
+++ b/packages/multichain-account-service/CHANGELOG.md
@@ -7,12 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [1.4.0]
-
 ### Changed
 
 - Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()` ([#6759](https://github.com/MetaMask/core/pull/6759))
   - Introduces `options.waitForAllProvidersToFinishCreatingAccounts`, that will make `createMultichainAccountGroup` await either only the EVM provider or all the providers to have created their accounts depending on the value. Defaults to `false` (only awaits for EVM accounts creation by default).
+
+## [1.4.0]
+
+### Changed
+
 - Only await for EVM account creation in `MultichainAccountWallet.createMultichainAccountGroup()` instead of all types of providers ([#6755](https://github.com/MetaMask/core/pull/6755))
   - Other type of providers will create accounts in the background and won't throw errors in case they fail to do so.
   - Multichain account groups will now be "misaligned" for a short period of time, until each of the other providers finish creating their accounts.
--- a/packages/multichain-account-service/CHANGELOG.md
+++ b/packages/multichain-account-service/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Added
 
 - Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()` ([#6759](https://github.com/MetaMask/core/pull/6759))
   - Introduces `options.waitForAllProvidersToFinishCreatingAccounts`, that will make `createMultichainAccountGroup` await either only the EVM provider or all the providers to have created their accounts depending on the value. Defaults to `false` (only awaits for EVM accounts creation by default).
--- a/packages/multichain-account-service/CHANGELOG.md
+++ b/packages/multichain-account-service/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add an optional `options` parameter to `MultichainAccountWallet.createMultichainAccountGroup()` ([#]())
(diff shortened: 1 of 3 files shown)


packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
```suggestion
### Added
```

ccharly (approved):
LGTM

owencraston (approved):

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -324,41 +335,72 @@ export class MultichainAccountWallet<
...
+          this.#log(`${WARNING_PREFIX} ${message}`);
+          console.warn(message);
+
+          throw new Error(error);
+        }
+      } else {
+        // Extract the EVM provider from the list of providers.
+        // We will only await the EVM provider to create its accounts, while
+        // all other providers will be started in the background.
+        const [evmProvider, ...otherProviders] = this.#providers;
```
fabiobozzo, line 371: Does that assume EVM provider is always first? Can't the provider order change in theory?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -419,7 +461,9 @@ export class MultichainAccountWallet<
   async createNextMultichainAccountGroup(): Promise<
     MultichainAccountGroup<Account>
   > {
-    return this.createMultichainAccountGroup(this.getNextGroupIndex());
+    return this.createMultichainAccountGroup(this.getNextGroupIndex(), {
```
fabiobozzo, line 464: This is a breaking change isn't it? Do we have to point that out in changelog? 👀

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
```diff
@@ -347,9 +347,32 @@ describe('MultichainAccountWallet', () => {
...
+    it('fails to create an account group if any of the provider fails to create its account and waitForAllProvidersToFinishCreatingAccounts is true', async () => {
+      const groupIndex = 1;
+
+      const mockEvmAccount = MockAccountBuilder.from(MOCK_HD_ACCOUNT_1)
+        .withEntropySource(MOCK_HD_KEYRING_1.metadata.id)
+        .withGroupIndex(0)
+        .get();
+      const { wallet, providers } = setup({
+        accounts: [[mockEvmAccount]], // 1 provider
+      });
```
fabiobozzo, line 362: This case only uses 1 provider but claims to test "all providers" behavior, do I read that right? Should it use multiple providers to properly test instead?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -324,41 +335,72 @@ export class MultichainAccountWallet<
...
+        } catch (error) {
+          const errorMessage = `Unable to create multichain account group for index: ${groupIndex} with provider "${evmProvider.getName()}". Error: ${(error as Error).message}`;
+          this.#log(`${ERROR_PREFIX} ${errorMessage}:`, error);
+          throw new Error(errorMessage);
+        }
+
+        // Create account with other providers in the background
+        otherProviders.forEach((provider) => {
+          provider
+            .createAccounts({
```
fabiobozzo, line 393: Shouldn't we track background promises too? Or is it the meaning to ignore failures here and future alignments will fix it? or do I miss something? 🤔

packages/multichain-account-service/src/MultichainAccountWallet.ts
fabiobozzo:
```
const evmProvider = this.#providers.find(p => p instanceof EvmAccountProvider);
assert(evmProvider, 'EVM account provider not found');
const otherProviders = this.#providers.filter(p => p !== evmProvider);
```

Maybe overkill 😋

fabiobozzo (approved):
LGTM. @mathieuartu I left a couple comments as food for thoughts, no remarks

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: That's not needed IMO. We could have an extra `assert` after we build the list of providers to enforce this precondition (like EVM provider always comes first)

packages/multichain-account-service/src/MultichainAccountWallet.ts
mathieuartu: Yes exactly, we willingly ignore failures for providers creating accounts in the background

ccharly: Yes that's expected to not throw in that case, mostly because would not be any way to catch them truly

Though, we make sure the `EvmAccountProvider` is successful.

If any of the other providers fail, then yes we would need an alignment for this (unfortunately, we don't have any state to tell us a group is misaligned for now, and we might need that in the near future IMO)

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Last change we made around this method was also altering the behavior a bit, but we did not treat it as a breaking change either.

I think, it could be seen as such, but for now, the only guarantee that we have for the multichain account group is that we ALWAYS HAVE at least an EVM account, and that still holds true with the previous change we made AND this one IMO.

Lastly, for simplicity, I would like to avoid breaking change in those packages to avoid bubbling up to other dependents controllers (we need this fix to be merged ASAP).

If we change our mind, we'll make a new 2.0.0 and amend the changelog if really needed later!

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
ccharly: It does claim that "if any of the provider fails", so the test is ok.

Though, I think we could have used the scenario EVM + Solana provider where Solana fails, that could have been clearer indeed 👍

PR #6789, opened 2025-10-07
Feat/eip 7702 internal rpc middleware

MoMannn:
## Explanation

This PR adds RPC methods for upgrading Externally Owned Accounts (EOAs) to EIP-7702 delegator accounts. This functions are only callable by preinstalled snaps. It adds

wallet_upgradeAccount for upgrading the account
wallet_getAccountUpgradeStatus for checking if account is upgraded or not

## References

Needed for: https://github.com/MetaMask/metamask-extension/pull/36452

## Checklist

---

> [!NOTE]
> Introduce `@metamask/eip-7702-internal-rpc-middleware` providing `wallet_upgradeAccount` and `wallet_getAccountUpgradeStatus`, with validation utilities, tests, and monorepo wiring.
> 
> - **New Package**: `packages/eip-7702-internal-rpc-middleware`
>   - Implements internal JSON-RPC methods:
>     - `wallet_upgradeAccount` (upgrade EOA via EIP-7702)
>     - `wallet_getAccountUpgradeStatus` (query upgrade status)
>   - Hook-based design (`upgradeAccount`, `isEip7702Supported`, network/account hooks).
>   - Utils: `validateParams`, `validateAndNormalizeAddress`; constants and types.
>   - Tests: unit tests for methods and utils; Jest config with 100% coverage thresholds.
>   - Docs/config: README, CHANGELOG, LICENSE, TypeDoc, tsconfigs, build scripts.
> - **Monorepo Integration**:
>   - Added to `README.md` (package list and dependency graph).
>   - Added ownership: `.github/CODEOWNERS`, `teams.json`.
>   - Added to build references: `tsconfig.build.json`.
>   - Workspace entry in `yarn.lock`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9cdfcb8fc9dfc3145e4263b100b4a646139fafcd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-preview.yml
+++ b/.github/workflows/publish-preview.yml
@@ -49,7 +49,7 @@ jobs:
         env:
           YARN_NPM_AUTH_TOKEN: ${{ secrets.PUBLISH_PREVIEW_NPM_TOKEN }}
       - name: Generate preview build message
-        run: yarn ts-node scripts/generate-preview-build-message.ts
+        run: yarn tsx scripts/generate-preview-build-message.ts
       - name: Post build preview in comment
         run: gh pr comment "${PR_NUMBER}" --body-file preview-build-message.txt
         env:
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -15,3 +15,13 @@ nodeLinker: node-modules
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
     spec: "https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js"
+
+# Configure the NPM minimal age gate to 3 days, meaning packages must be at
+# least 3 days old to be installed.
+npmMinimalAgeGate: 4320 # 3 days (in minutes)
+
+# Override the minimal age gate, allowing certain packages to be installed
+# regardless of their publish age.
+npmPreapprovedPackages:
+  - "@metamask/*"
+  - "@lavamoat/*"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "625.0.0",
+  "version": "629.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
@@ -104,7 +104,7 @@
     "typescript-eslint": "^8.7.0",
     "yargs": "^17.7.2"
   },
-  "packageManager": "yarn@4.2.2",
+  "packageManager": "yarn@4.10.3",
   "engines": {
     "node": "^18.18 || >=20"
   },
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,10 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add multicall addresses in `MULTICALL_CONTRACT_BY_CHAINID` ([#6896](https://github.com/MetaMask/core/pull/6896))
+  - Add multicall address for Chains: `Injective`, `Hemi`, `Plasma`, `Nonmia`, `XRPL`, `Soneium`, `Genesys`, `EDU`, `Abstract`, `Berachain`, `MegaETH Testnet`, `Apechain`, `Matchain`, `Monad Testnet`, `Monad`, `Katana`, `Lens`, `Plume`, `XDC`
+
 ### Changed
 
-- Bump `@metamask/network-controller` from `^24.2.2` to `^24.3.0` ([#6883](https://github.com/MetaMask/core/pull/6883))
-- Bump `@metamask/transaction-controller` from `^60.7.0` to `^60.8.0` ([#6883](https://github.com/MetaMask/core/pull/6883))
+- Batch `OnAssetConversion` and `OnAssetsMarketData` requests to non-EVM account Snaps ([#6886](https://github.com/MetaMask/core/pull/6886))
+
+## [81.0.1]
 
 ### Fixed
 
@@ -2136,7 +2142,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use Ethers for AssetsContractController ([#845](https://github.com/MetaMask/core/pull/845))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@81.0.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@81.0.1...HEAD
+[81.0.1]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@81.0.0...@metamask/assets-controllers@81.0.1
 [81.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@80.0.0...@metamask/assets-controllers@81.0.0
 [80.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@79.0.1...@metamask/assets-controllers@80.0.0
 [79.0.1]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@79.0.0...@metamask/assets-controllers@79.0.1
--- a/packages/assets-controllers/package.json
+++ b/packages/assets-controllers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/assets-controllers",
-  "version": "81.0.0",
+  "version": "81.0.1",
   "description": "Controllers which manage interactions involving ERC-20, ERC-721, and ERC-1155 tokens (including NFTs)",
   "keywords": [
     "MetaMask",
@@ -96,7 +96,7 @@
     "@metamask/preferences-controller": "^20.0.2",
     "@metamask/providers": "^22.1.0",
     "@metamask/snaps-controllers": "^14.0.1",
-    "@metamask/transaction-controller": "^60.8.0",
+    "@metamask/transaction-controller": "^60.9.0",
     "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.191",
     "@types/node": "^16.18.54",
--- a/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.test.ts
+++ b/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.test.ts
@@ -71,6 +71,20 @@ const fakeEvmAccountWithoutMetadata: InternalAccount = {
   methods: [],
 };
 
+const fakeNonEvmAccount2: InternalAccount = {
+  id: 'account5',
+  type: 'solana:data-account',
+  address: '0x123',
+  metadata: {
+    name: 'Test Account',
+    // @ts-expect-error-next-line
+    snap: { id: 'test-snap-2', enabled: true },
+  },
+  scopes: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp'],
+  options: {},
+  methods: [],
+};
+
 const fakeMarketData = {
   price: 202.11,
   priceChange: 0,
@@ -127,6 +141,9 @@ const setupController = ({
         account1: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501'],
         account2: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501'],
         account3: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501'],
+        account5: [
+          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/token:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v',
+        ],
       },
       assetsMetadata: {
         'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501': {
@@ -136,6 +153,14 @@ const setupController = ({
           iconUrl: 'https://example.com/solana.png',
           units: [{ symbol: 'SOL', name: 'Solana', decimals: 9 }],
         },
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/token:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v':
+          {
+            name: 'USDC',
+            symbol: 'USDC',
+            fungible: true,
+            iconUrl: 'https://example.com/usdc.png',
+            units: [{ symbol: 'USDC', name: 'USDC', decimals: 2 }],
+          },
       },
     }),
   );
@@ -356,8 +381,8 @@ describe('MultichainAssetsRatesController', () => {
             type: KeyringTypes.snap,
           },
           snap: {
-            id: 'mock-sol-snap',
-            name: 'mock-sol-snap',
+            id: 'mock-sol-snap-1',
+            name: 'mock-sol-snap-1',
             enabled: true,
           },
           lastSelected: 0,
@@ -377,8 +402,8 @@ describe('MultichainAssetsRatesController', () => {
             type: KeyringTypes.snap,
           },
           snap: {
-            id: 'mock-sol-snap',
-            name: 'mock-sol-snap',
+            id: 'mock-sol-snap-2',
+            name: 'mock-sol-snap-2',
             enabled: true,
           },
           lastSelected: 0,
@@ -393,42 +418,49 @@ describe('MultichainAssetsRatesController', () => {
       accountsAssets: testAccounts,
     });
 
-    const snapSpy = jest
-      .fn()
-      .mockResolvedValueOnce({
-        conversionRates: {
-          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501': {
-            'swift:0/iso4217:USD': {
-              rate: '100',
-              conversionTime: 1738539923277,
+    const mockResponses = {
+      onAssetsConversion: [
+        {
+          conversionRates: {
+            'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501': {
+              'swift:0/iso4217:USD': {
+                rate: '100',
+                conversionTime: 1738539923277,
+              },
             },
           },
         },
-      })
-      .mockResolvedValueOnce({
-        marketData: {
-          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501': {
-            'swift:0/iso4217:USD': fakeMarketData,
-          },
-        },
-      })
-      .mockResolvedValueOnce({
-        conversionRates: {
-          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/token1:501': {
-            'swift:0/iso4217:USD': {
-              rate: '200',
-              conversionTime: 1738539923277,
+        {
+          conversionRates: {
+            'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/token1:501': {
+              'swift:0/iso4217:USD': {
+                rate: '200',
+                conversionTime: 1738539923277,
+              },
             },
           },
         },
-      })
-      .mockResolvedValueOnce({
-        marketData: {
-          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/token1:501': {
+      ],
+      onAssetsMarketData: [
+        {
+          marketData: {
             'swift:0/iso4217:USD': fakeMarketData,
           },
         },
-      });
+        {
+          marketData: {
+            'swift:0/iso4217:USD': fakeMarketData,
+          },
+        },
+      ],
+    };
+
+    const snapSpy = jest.fn().mockImplementation((args) => {
+      const { handler } = args;
+      return Promise.resolve(
+        mockResponses[handler as keyof typeof mockResponses].shift(),
+      );
+    });
     messenger.registerActionHandler('SnapController:handleRequest', snapSpy);
 
     messenger.publish('MultichainAssetsController:accountAssetListUpdated', {
@@ -443,6 +475,7 @@ describe('MultichainAssetsRatesController', () => {
         },
       },
     });
+
     // Wait for the asynchronous subscriber to run.
     await Promise.resolve();
     await advanceTime({ clock, duration: 10 });
@@ -625,7 +658,7 @@ describe('MultichainAssetsRatesController', () => {
 
     it('handles mixed success and failure scenarios', async () => {
       const { controller, messenger } = setupController({
-        accountsAssets: [fakeNonEvmAccount, fakeEvmAccount2],
+        accountsAssets: [fakeNonEvmAccount, fakeNonEvmAccount2],
       });
 
       const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation();
@@ -957,7 +990,7 @@ describe('MultichainAssetsRatesController', () => {
       const snapHandler = jest.fn().mockResolvedValue({
         conversionRates: {
           'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501': {
-            USD: {
+            'swift:0/iso4217:USD': {
               rate: '100.50',
               conversionTime: Date.now(),
             },
--- a/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts
+++ b/packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts
@@ -31,7 +31,6 @@ import type {
 import { HandlerType } from '@metamask/snaps-utils';
 import { Mutex } from 'async-mutex';
 import type { Draft } from 'immer';
-import { cloneDeep } from 'lodash';
 
 import { MAP_CAIP_CURRENCIES } from './constant';
 import type {
@@ -182,6 +181,15 @@ export type ConversionRatesWithMarketData = {
 };
 
 /**
+ * Arguments for a Snap request.
+ */
+type SnapRequestArgs<T> = {
+  snapId: SnapId;
+  handler: HandlerType;
+  params: T;
+};
+
+/**
  * Controller that manages multichain token conversion rates.
  *
  * This controller polls for token conversion rates and updates its state.
@@ -321,6 +329,42 @@ export class MultichainAssetsRatesController extends StaticIntervalPollingContro
   }
 
   /**
+   * Adds the assets to a map of Snap ID to assets.
+   *
+   * @param snapIdToAssets - The map of Snap ID to assets.
+   * @param account - The account to add the assets for.
+   * @param assets - The assets to add.
+   */
+  #addAssetsToSnapIdMap(
+    snapIdToAssets: Map<SnapId, Set<CaipAssetType>>,
+    account: InternalAccount,
+    assets: CaipAssetType[],
+  ): void {
+    // Prevent creating a new set if there are no assets to add.
+    if (assets.length === 0) {
+      return;
+    }
+
+    // FIXME: Instead of using the Snap ID from the account, we should
+    // select the Snap based on the supported scopes defined in the Snaps'
+    // manifest.
+    const snapId = account.metadata.snap?.id as SnapId | undefined;
+    if (!snapId) {
+      return;
+    }
+
+    let snapAssets = snapIdToAssets.get(snapId);
+    if (!snapAssets) {
+      snapAssets = new Set();
+      snapIdToAssets.set(snapId, snapAssets);
+    }
+
+    for (const asset of assets) {
+      snapAssets.add(asset);
+    }
+  }
+
+  /**
    * Updates token conversion rates for each non-EVM account.
    *
    * @returns A promise that resolves when the rates are updated.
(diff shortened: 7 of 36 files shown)


packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+    return { isUpgraded: false, upgradedAddress: null };
+  }
+
+  if (!code.startsWith(DELEGATION_INDICATOR_PREFIX)) {
+    return { isUpgraded: false, upgradedAddress: null };
+  }
+
+  const expectedLength = DELEGATION_INDICATOR_PREFIX.length + 40; // 0xef0100 + 40 hex chars
+  if (code.length !== expectedLength) {
+    return { isUpgraded: false, upgradedAddress: null };
```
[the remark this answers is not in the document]
MoMannn: that is file

packages/eip-7702-internal-rpc-middleware/src/wallet_upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const atomicBatchChainSupport = atomicBatchSupport.find(
+      (result) => result.chainId.toLowerCase() === hexChainId.toLowerCase(),
+    );
+
+    const isChainSupported =
+      atomicBatchChainSupport &&
+      (!atomicBatchChainSupport.delegationAddress ||
+        atomicBatchChainSupport.isSupported);
+
```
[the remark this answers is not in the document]
MoMannn: this is the correct logic

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,94 @@
+import { providerErrors, rpcErrors } from '@metamask/rpc-errors';
+import type { Struct, StructError } from '@metamask/superstruct';
+import { validate } from '@metamask/superstruct';
+import type { Hex, JsonRpcRequest } from '@metamask/utils';
+
+/**
+ * Validates and normalizes address for EIP-7702 operations.
```
jeffsmale90: nit: it's not clear to me from this description what validation and normalization is happening.

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param options.getAccounts - Function to get accounts for the requester.
+ * @returns Promise that resolves to the upgrade result containing transaction hash and delegation info.
+ */
+export async function upgradeAccount(
+  req: JsonRpcRequest<Json[]> & { origin: string },
+  _res: PendingJsonRpcResponse,
+  {
+    upgradeAccount: upgradeAccountFn,
+    getCurrentChainIdForDomain,
+    isAtomicBatchSupported,
```
jeffsmale90: `isAtomicBatchSupported` should really be named `isEip7702Supported` (or similar).

In extension we may infer that EIP7702 is supported via the configuration `isAtomicBatchSupported`, but at the middleware layer, this should be more specific to the usage rather than implementation.

jeffsmale90 (approved):
This looks great. A couple minor naming comments.

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.test.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@metamask/utils';
+
+import { walletGetAccountUpgradeStatus } from './wallet_getAccountUpgradeStatus';
+import { getAccountUpgradeStatus } from '../hooks/getAccountUpgradeStatus';
+import type { GetAccountUpgradeStatusHooks } from '../types';
+
+jest.mock('../hooks/getAccountUpgradeStatus');
+
+describe('walletGetAccountUpgradeStatus', () => {
+  it('calls getAccountUpgradeStatus hook with correct parameters and sets result', async () => {
```
mcmire: Instead of merely testing that the hook is called correctly should we also verify some part of the hook's behavior, perhaps the happy path? Otherwise we are testing that these pieces work independently but we aren't really testing that they work together.

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const req = {} as JsonRpcRequest<Json[]> & { origin: string };
```
mcmire: Can we make real requests and responses rather than partial ones? Otherwise we're not really testing that the types are correct, we're just assuming that they are.

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const res = { result: null } as PendingJsonRpcResponse;
+    const hooks = {
+      getCurrentChainIdForDomain: jest.fn(),
+      getCode: jest.fn(),
+      getNetworkConfigurationByChainId: jest.fn(),
+      getAccounts: jest.fn(),
+    };
+
+    const mockResult = { account: '0xabc', isUpgraded: true, chainId: 1 };
+    (getAccountUpgradeStatus as jest.Mock).mockResolvedValue(mockResult);
```
mcmire: If we add

``` typescript
const getAccountUpgradeStatusMock = jest.mocked(getAccountUpgradeStatus);
```

to the top of the file, after the imports, then we don't need to assert that this is a `jest.Mock`, we can use `getAccountUpgradeStatusMock` from now on.

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+} from '../types';
+import { GetAccountUpgradeStatusParamsStruct } from '../types';
+import { validateParams, validateAndNormalizeAddress } from '../utils';
+
+const isAccountUpgraded = async (
+  address: string,
+  networkClientId: string,
+  getCode: (address: string, networkClientId: string) => Promise<string | null>,
+): Promise<{ isUpgraded: boolean; upgradedAddress: Hex | null }> => {
+  // This is a mock implementation - in real usage this would come from @metamask/eip7702-utils
```
mcmire: Why is this a mock implementation? Is this a new package that we plan to introduce?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // Use current chain ID if not provided
+  let targetChainId: number;
+  if (chainId !== undefined) {
+    targetChainId = chainId;
+  } else {
+    const currentChainIdForDomain = getCurrentChainIdForDomain(origin);
+    if (!currentChainIdForDomain) {
+      throw rpcErrors.invalidParams({
+        message: `No network configuration found for origin: ${origin}`,
```
mcmire: Since the hook is called `getCurrentChainIdForDomain` should we say that we can't find a chain ID rather than a network configuration? This relies on implementation details which we don't have access to here.

``` suggestion
        message: `Could not determine current chain ID for origin: ${origin}`,
```

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+      });
+    }
+    targetChainId = parseInt(currentChainIdForDomain, 16);
+  }
+
+  try {
+    // Get the network configuration for the target chain
+    const hexChainId = toHex(targetChainId);
+    const networkConfiguration = getNetworkConfigurationByChainId(hexChainId);
```
mcmire: This is a pretty common operation across our stack and it seems like something we shouldn't have to do here. Should we expect that we have a hook like `getSelectedNetworkClientIdForChain` instead?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    return {
+      account: normalizedAccount,
+      isUpgraded,
+      upgradedAddress,
+      chainId: targetChainId,
+    };
+  } catch (error) {
+    // Re-throw RPC errors as-is
+    if (error && typeof error === 'object' && 'code' in error) {
```
mcmire: Should we check that this is an instance of RpcError (which we can get from `@metamask/rpc-errors`) instead of checking for `code`? Is that really what we are interested in here?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+      throw error as unknown as Error;
+    }
+    throw rpcErrors.internal({
+      message: `Failed to get account upgrade status: ${error instanceof Error ? error.message : String(error)}`,
```
mcmire: Tip: `@metamask/utils` has a `getErrorMessage` function that can be used to convert a throwable into a string:

``` suggestion
      message: `Failed to get account upgrade status: ${getErrorMessage(error)}`,

File: packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts [hunk 13]
```diff
@@ -0,0 +1,150 @@
...
+
+    if (!networkClientId) {
+      throw rpcErrors.invalidParams({
+        message: `Network client ID not found for chain ID ${targetChainId}`,
+      });
+    }
+
+    // Check if the account is upgraded using the EIP7702 utils
+    const { isUpgraded, upgradedAddress } = await isAccountUpgraded(
+      normalizedAccount as Hex,
```
mcmire: Why do we have to assert this is a `Hex`?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+    getNetworkConfigurationByChainId,
+    getAccounts,
+  }: GetAccountUpgradeStatusHooks,
+): Promise<GetAccountUpgradeStatusResult> {
+  const { params, origin } = req;
+
+  // Validate parameters using Superstruct
+  validateParams(params, tuple([GetAccountUpgradeStatusParamsStruct]));
+
+  const [statusParams] = params as [GetAccountUpgradeStatusParams];
```
mcmire: Instead of using a type assertion here can we let `validateParams` do the assertion for us (by adding `asserts params is GetAccountUpgradeStatusParams` to `validateParams`)? (EDIT: Oh, hmm, it looks like it already does do something like this. Maybe this type assertion isn't necessary?)

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.test.ts
```diff
@@ -0,0 +1,370 @@
...
+    response,
+    handler,
+  };
+};
+
+describe('upgradeAccount', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
```
mcmire: Jest is already configured to clear and restore mocks in this repo, we don't need to do this manually.

``` suggestion
```

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (chainId !== undefined) {
+    targetChainId = chainId;
+  } else {
+    const currentChainIdForDomain = getCurrentChainIdForDomain(origin);
+    if (!currentChainIdForDomain) {
+      throw rpcErrors.invalidParams({
+        message: `No network configuration found for origin: ${origin}`,
+      });
+    }
+    targetChainId = parseInt(currentChainIdForDomain, 16);
```
mcmire: `@metamask/controller-utils` has a `fromHex` function, should we use that instead?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (chainId !== undefined) {
+    targetChainId = chainId;
+  } else {
+    const currentChainIdForDomain = getCurrentChainIdForDomain(origin);
+    if (!currentChainIdForDomain) {
+      throw rpcErrors.invalidParams({
+        message: `No network configuration found for origin: ${origin}`,
+      });
+    }
+    targetChainId = parseInt(currentChainIdForDomain, 16);
```
mcmire: `@metamask/controller-utils` has a `fromHex` function, should we use that instead?

packages/eip-7702-internal-rpc-middleware/src/index.test.ts
mcmire, line 1: This is an old pattern that you might find in other packages, but we don't really need this anymore, as `index.ts` is now the source of truth. What do you think about removing this file?

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
```diff
@@ -0,0 +1,180 @@
...
+jest.mock('@metamask/superstruct', () => ({
+  ...jest.requireActual('@metamask/superstruct'),
+  validate: jest.fn(),
+}));
+
+describe('validateParams', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
```
mcmire: Jest does this automatically.

``` suggestion
```

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
```diff
@@ -0,0 +1,180 @@
...
+import {
+  validateParams,
+  validateAndNormalizeAddress,
+  resemblesAddress,
+} from './utils';
+
+jest.mock('@metamask/superstruct', () => ({
+  ...jest.requireActual('@metamask/superstruct'),
+  validate: jest.fn(),
```
mcmire: Why are we mocking Superstruct?

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  ) {
+    // Ensure that an "unauthorized" error is thrown if the requester
+    // does not have the `eth_accounts` permission.
+    const accounts = await getAccounts(req);
+
+    const normalizedAccounts: string[] = accounts.map((_address) =>
+      _address.toLowerCase(),
+    );
+
+    const normalizedAddress = address.toLowerCase() as Hex;
```
mcmire: Do we need to typecast this as `Hex` if `normalizedAccounts` is `string[]`?

mcmire (commented):
Did a brief pass, will do a round 2 later.

One question I had: Is it necessary to add `internal-rpc` within the package name? We didn't do that for `eip-5792-middleware` I notice.

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+  // Validate and normalize the account address with authorization check
+  const normalizedAccount = await validateAndNormalizeAddress(account, req, {
+    getAccounts,
+  });
+
+  // Use current chain ID if not provided
+  let targetChainId: number;
+  if (chainId !== undefined) {
+    targetChainId = chainId;
+  } else {
```
adonesky1:
```suggestion
  // Use current app selected chain ID if not passed as a param
  let targetChainId: number;
  if (chainId !== undefined) {
    targetChainId = chainId;
  } else {
```

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+  const [upgradeParams] = params as [UpgradeAccountParams];
+  const { account, chainId } = upgradeParams;
+
+  // Validate and normalize the account address with authorization check
+  const normalizedAccount = await validateAndNormalizeAddress(account, req, {
+    getAccounts,
+  });
+
+  // Use current chain ID if not provided
+  let targetChainId: number;
```
adonesky1: [nit]
```suggestion
  let decChainId: number;
```

packages/eip-7702-internal-rpc-middleware/src/wallet_upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+  // Validate and normalize the account address with authorization check
+  const normalizedAccount = await validateAndNormalizeAddress(account, req, {
+    getAccounts,
+  });
+
+  // Use current chain ID if not provided
+  let targetChainId: number;
+  if (chainId !== undefined) {
+    targetChainId = chainId;
```
adonesky1, line 55: I don't think this param should be decimal formatted, both because its not idiomatically correct in the ecosystem, and because of javascript integer overflow concerns

packages/eip-7702-internal-rpc-middleware/src/hooks/upgradeAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    const { upgradeContractAddress } = atomicBatchChainSupport;
+
+    // Perform the upgrade using existing EIP-7702 functionality
+    const result = await upgradeAccountFn(
+      normalizedAccount,
+      upgradeContractAddress,
+      targetChainId,
+    );
```
adonesky1: For [this reason](https://github.com/MetaMask/core/pull/6789/files#r2414770850) I think we ought to consider changing this function to expect hexadecimal values for the chainId instead

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param hooks - The hooks object containing status check functionality.
+ */
+export async function walletGetAccountUpgradeStatus(
+  req: JsonRpcRequest<Json[]> & { origin: string },
+  res: PendingJsonRpcResponse,
+  hooks: GetAccountUpgradeStatusHooks,
+): Promise<void> {
+  if (!hooks.getCode) {
+    throw rpcErrors.methodNotSupported();
+  }
```
adonesky1: 🤔 this seems strange to me we should be able to guarantee that the hook is passed in with types. And in any case the right error here would be `rpcErrors.internal()`

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_upgradeAccount.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param hooks - The hooks object containing upgrade functionality.
+ */
+export async function walletUpgradeAccount(
+  req: JsonRpcRequest<Json[]> & { origin: string },
+  res: PendingJsonRpcResponse,
+  hooks: UpgradeAccountHooks,
+): Promise<void> {
+  if (!hooks.upgradeAccount) {
+    throw rpcErrors.methodNotSupported();
+  }
```
adonesky1: Same thing as here: https://github.com/MetaMask/core/pull/6789/files#r2414777738

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  { getAccounts }: { getAccounts: (req: JsonRpcRequest) => Promise<string[]> },
+): Promise<Hex> {
+  if (
+    typeof address === 'string' &&
+    address.length > 0 &&
+    resemblesAddress(address)
+  ) {
+    // Ensure that an "unauthorized" error is thrown if the requester
+    // does not have the `eth_accounts` permission.
+    const accounts = await getAccounts(req);
```
adonesky1:
```suggestion
    const accounts = await getPermittedAccountsForOrigin(req.origin);
```

I think this hook should be renamed to indicate that it fetches the permitted accounts for given origin. We have some utility functions for extracting the permissioned accounts from the CAIP-25 permission that you'll want to use for example: https://github.com/MetaMask/core/blob/30107f464758c889cfe0050866cbbc17b8663f42/packages/chain-agnostic-permission/src/operators/caip-permission-operator-accounts.ts#L77

I would also recommend only passing the origin rather than the whole req object

packages/eip-7702-internal-rpc-middleware/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+**Example:**
+
+```json
+{
+  "method": "wallet_upgradeAccount",
+  "params": [
+    {
+      "account": "0x1234567890123456789012345678901234567890",
+      "chainId": 1
```
adonesky1: same as [here](https://github.com/MetaMask/core/pull/6789/files#r2414770850),

This param should be hexadecimal formatted, both because decimal is not idiomatically correct in the ecosystem, and because of javascript integer overflow concerns

packages/eip-7702-internal-rpc-middleware/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "baseUrl": "./",
+    "rootDir": "../.."
+  },
+  "references": [
+    { "path": "../network-controller" },
+    { "path": "../transaction-controller" },
+    { "path": "../base-controller" },
+    { "path": "../accounts-controller" },
+    { "path": "../preferences-controller" },
+    { "path": "../keyring-controller" }
```
jiexi: are all of these references needed?

packages/eip-7702-internal-rpc-middleware/tsconfig.build.json
```diff
@@ -0,0 +1,17 @@
...
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
+  "references": [
+    { "path": "../network-controller/tsconfig.build.json" },
+    { "path": "../transaction-controller/tsconfig.build.json" },
+    { "path": "../base-controller/tsconfig.build.json" },
+    { "path": "../accounts-controller/tsconfig.build.json" },
+    { "path": "../preferences-controller/tsconfig.build.json" },
+    { "path": "../keyring-controller/tsconfig.build.json" }
```
jiexi, line 14: are all of these references needed?

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.ts
jiexi: I think this is an artifact of the old pattern where many unrelated middleware were in `@metamask/eth-json-rpc-middleware`'s `createWalletMiddleware` fn. So hooks were overloaded to enable/disable specific middleware handlers.

I think we should fix that now by inlining the hooks into the handler themselves

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+    getAccounts,
+  }: GetAccountUpgradeStatusHooks,
+): Promise<GetAccountUpgradeStatusResult> {
+  const { params, origin } = req;
+
+  // Validate parameters using Superstruct
+  validateParams(params, tuple([GetAccountUpgradeStatusParamsStruct]));
+
+  const [statusParams] = params as [GetAccountUpgradeStatusParams];
+  const { account, chainId } = statusParams;
```
jiexi: did we want to enforce that the currently selected network for the request matches that of the request params?

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
MoMannn: Ah, comment that got left in from before. Removing.

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
MoMannn: Basically to achieve 100% test coverage, change to not mocking but still getting the tests.

packages/eip-7702-internal-rpc-middleware/src/methods/wallet_getAccountUpgradeStatus.ts
MoMannn: I followed the eip-5792-middleware pattern. Which I got as reference. Will change to inlining the hooks into the handler and move to a flat stucture.

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
MoMannn: Not sure if I completely understand the question. But the permission can be on any network as such we need to allow updating the account on any network. So independent on what the user has selected?

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // Extract the 20-byte address (40 hex characters after the prefix)
+  const upgradedAddress = `0x${code.slice(8, 48)}` as Hex;
```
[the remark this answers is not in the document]
MoMannn: It works correctly

packages/eip-7702-internal-rpc-middleware/src/utils.ts
MoMannn: If we dont: https://github.com/MetaMask/core/pull/6789#discussion_r2416194009

packages/eip-7702-internal-rpc-middleware/src/hooks/getAccountUpgradeStatus.ts
MoMannn: Sure

packages/eip-7702-internal-rpc-middleware/src/wallet_upgradeAccount.ts
```diff
@@ -0,0 +1,120 @@
...
+        message: `No network configuration found for origin: ${origin}`,
+      });
+    }
+    targetChainId = currentChainIdForDomain;
+  }
+
+  try {
+    // Get the EIP7702 network configuration for the target chain
+    const hexChainId = targetChainId;
+    const atomicBatchSupport = await hooks.isEip7702Supported({
```
MoMannn: TODO: This logic will probably need to be moved to extension because now we have the logic expecting `atomicBatchChainSupport` result from the hook not  isEip7702Supported. So will change that.

packages/eip-7702-internal-rpc-middleware/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@metamask/eip-7702-internal-rpc-middleware",
+  "version": "1.0.0",
```
jiexi: i think packages are supposed to initially be added as 0.0.0

packages/eip-7702-internal-rpc-middleware/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  chainId: optional(StrictHexStruct),
+});
+
+// Type definitions derived from schemas
+export type UpgradeAccountParams = Infer<typeof UpgradeAccountParamsStruct>;
+
+export type UpgradeAccountResult = {
+  transactionHash: string; // Hash of the EIP-7702 authorization transaction
+  upgradedAccount: string; // Address of the upgraded account (same as input)
+  delegatedTo: string; // Address of the contract delegated to (determined by wallet)
```
jiexi: should these be hex?

packages/eip-7702-internal-rpc-middleware/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+export type GetAccountUpgradeStatusParams = Infer<
+  typeof GetAccountUpgradeStatusParamsStruct
+>;
+
+export type GetAccountUpgradeStatusResult = {
+  account: string; // Address of the checked account
+  chainId: Hex; // Chain ID where the check was performed
+  isSupported: boolean; // Whether upgrade to smart account is supported on the chain
+  isUpgraded: boolean; // Whether the account is upgraded
+  upgradedAddress: string | null; // Address to which the account is upgraded
```
jiexi: should `account` and `upgradedAddress` be hex?

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  it('throws RPC error with formatted message for invalid parameters', () => {
+    const testStruct = object({
+      name: string(),
+      age: number(),
+    });
+    const invalidValue = { name: 123, age: 'invalid' };
+
+    expect(() => validateParams(invalidValue, testStruct)).toThrow(
+      'Invalid parameters',
```
jiexi: should this assert that it is also a rpcErrors.internal error?

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @returns A formatted error message string with validation failure details.
+ */
+function formatValidationError(error: StructError, message: string): string {
+  return `${message}\n\n${error
+    .failures()
+    .map(
+      (f) => `${f.path.join(' > ')}${f.path.length ? ' - ' : ''}${f.message}`,
+    )
+    .join('\n')}`;
+}
```
jiexi, line 82: I feel like I've seen this pattern in a few other parts of the codebase already,  or code shaped like this at least

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+      // we know that normalizedAddress is a valid Hex string because of isHexAddress
+      return normalizedAddress as Hex;
+    }
+
+    throw providerErrors.unauthorized();
+  }
+
+  throw rpcErrors.invalidParams({
+    message: `Invalid parameters: must provide an Ethereum address.`,
+  });
```
jiexi: nit, is it okay if we inverse the if statement at the start of this function and then throw for an early exit?

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    );
+  });
+
+  it('formats validation errors with empty path (root level errors)', () => {
+    // Test with a struct that expects a string but gets a non-object
+    const testStruct = string();
+    const invalidValue = 123;
+
+    expect(() => validateParams(invalidValue, testStruct)).toThrow(
+      'Invalid parameters',
```
jiexi: should there be specific field validation errors in the error message?

packages/eip-7702-internal-rpc-middleware/src/utils.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      .mockResolvedValue(['0x9999999999999999999999999999999999999999']);
+
+    await expect(
+      validateAndNormalizeAddress(
+        address,
+        mockOrigin,
+        getPermittedAccountsForOrigin,
+      ),
+    ).rejects.toThrow(
+      'The requested account and/or method has not been authorized by the user.',
```
jiexi:
```suggestion
      providerErrors.unauthorized()
```

packages/eip-7702-internal-rpc-middleware/src/utils.ts
jiexi: `must provide a valid Hex address` Maybe?

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.test.ts
```diff
@@ -0,0 +1,334 @@
...
+} from '@metamask/utils';
+
+import { walletGetAccountUpgradeStatus } from './wallet_getAccountUpgradeStatus';
+import type { WalletGetAccountUpgradeStatusHooks } from './wallet_getAccountUpgradeStatus';
+
+type TestHooks = {
+  [K in keyof WalletGetAccountUpgradeStatusHooks]: jest.MockedFunction<
+    WalletGetAccountUpgradeStatusHooks[K]
+  >;
+};
```
jiexi: is this needed? I feel like this would work with the inferred types already

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    const hooks = createTestHooks();
+    const req = createTestRequest();
+    const res = createTestResponse();
+
+    // Mock getCode to throw a non-RPC error
+    hooks.getCode.mockRejectedValue(new Error('Network error'));
+
+    await expect(
+      walletGetAccountUpgradeStatus(req, res, hooks),
+    ).rejects.toThrow('Failed to get account upgrade status: Network error');
```
jiexi: should this check for an instance of the internal error?

packages/eip-7702-internal-rpc-middleware/src/wallet_upgradeAccount.test.ts
```diff
@@ -0,0 +1,235 @@
...
+} from '@metamask/utils';
+
+import { walletUpgradeAccount } from './wallet_upgradeAccount';
+import type { WalletUpgradeAccountHooks } from './wallet_upgradeAccount';
+
+type TestHooks = {
+  [K in keyof WalletUpgradeAccountHooks]: jest.MockedFunction<
+    WalletUpgradeAccountHooks[K]
+  >;
+};
```
jiexi: same

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * The RPC method handler middleware for `wallet_getAccountUpgradeStatus`
+ *
+ * @param req - The JSON RPC request's end callback.
+ * @param res - The JSON RPC request's pending response object.
+ * @param hooks - The hooks required for account upgrade status checking.
+ */
+export async function walletGetAccountUpgradeStatus(
+  req: JsonRpcRequest<Json[]> & { origin: string },
```
jiexi: just wondering why we using array params. Not sure if this was discussed somewhere already or not

packages/eip-7702-internal-rpc-middleware/src/types.ts
MoMannn: yep

packages/eip-7702-internal-rpc-middleware/src/utils.ts
MoMannn: There is the same function in the `eip-5792-middleware` package this is copied from there. Found similar thing in `user-operation-controller` as well. But its not in any package/lib that is shared across packages. 

Should I extract this function to some kind of common utils? No sure to which package this would be most appropriate.

packages/eip-7702-internal-rpc-middleware/src/utils.ts
MoMannn: Sure, maybe EVM address? Hex is still more general and can be another chain type address.

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.ts
MoMannn: So its like kind of a standard pattern to always use an array for JSON RPC params even if its only 1 parameter. 

But with closer inspection of the code I see both ways are used O.o 

`wallet_requestPermissions` is using an array with a single object while `wallet_invokeSnap` is not using an array but just a single object. 

I don't know if there is a reasoning when to use which?

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  const normalizedAddress = address.toLowerCase();
+
+  if (!normalizedAccounts.includes(normalizedAddress)) {
+    throw providerErrors.unauthorized();
+  }
+
+  // we know that normalizedAddress is a valid Hex string because of isHexAddress
+  return normalizedAddress as Hex;
```
jiexi: the type cast shouldn't be needed since [isHexAddress](https://github.com/MetaMask/utils/blob/339faaef0f3ecb94bf506d5b101c07bce4ebc4d5/src/hex.ts#L59) includes type narrowing

packages/eip-7702-internal-rpc-middleware/src/utils.ts
jiexi: We probably should, but we don't have to do it as part of this PR

packages/eip-7702-internal-rpc-middleware/src/wallet_getAccountUpgradeStatus.ts
jiexi: I think it's because JSON-RPC 1.0 only supported an array as the param value, but 2.0 supports both array and object. So that pattern carried over here even though I don't think we ever supported 1.0 spec. 

IMO we should use an object as the root value as that's what we care about. @adonesky1

adonesky1: Yeah I don't love that we have it mixed and matched across our API but 🤷 this is the tricky thing about having public APIs where modifying would be breaking for tons of dispersed consumers. I agree w/ Jiexi that this should just use the newer cleaner pattern of using an object as the root value

packages/eip-7702-internal-rpc-middleware/src/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    !isHexAddress(address)
+  ) {
+    throw rpcErrors.invalidParams({
+      message: `Invalid parameters: must provide an EVM address.`,
+    });
+  }
+
+  // Ensure that an "unauthorized" error is thrown if the requester
+  // does not have the `eth_accounts` permission.
+  const accounts = await getPermittedAccountsForOrigin(origin);
```
jiexi, line 34: getPermittedAccounts does not throw when there is no permission in extension

https://github.com/MetaMask/metamask-extension/blob/86598128b74da543788f598828a885d17764113f/app/scripts/metamask-controller.js#L5449-L5451

MoMannn: This is either a wider problem or not a problem. 
Both 5792-middleware: https://github.com/MetaMask/core/blob/fd638ef67c7e6e9565cc4ec09618e4d3543f88c0/packages/eip-5792-middleware/src/utils.ts#L65

and eth-json-middleware: https://github.com/MetaMask/core/blob/fd638ef67c7e6e9565cc4ec09618e4d3543f88c0/packages/eth-json-rpc-middleware/src/utils/validation.ts#L18

Use the same kind of validation.

jiexi (approved):

packages/eip-7702-internal-rpc-middleware/src/utils.ts
jiexi: sorry about that. I can't read

PR #6791, opened 2025-10-07
chore: Add the optional "Client-Version" header in the Bridge Controller

dan437:
## Explanation
Adds the optional "Client-Version" header into API calls from the Bridge Controller. E.g. `"Client-Version": "13.5.0"`

## Checklist


packages/bridge-controller/src/utils/fetch.ts
```diff
@@ -23,22 +23,25 @@ import type {
 
 const CACHE_REFRESH_TEN_MINUTES = 10 * Duration.Minute;
 
-export const getClientIdHeader = (clientId: string) => ({
+export const getClientHeaders = (clientId: string, clientVersion?: string) => ({
   'X-Client-Id': clientId,
+  ...(clientVersion ? { 'Client-Version': clientVersion } : {}),
```
dan437, line 28: the “X-” prefix for custom HTTP headers (like X-Client-Id) was deprecated by IETF in RFC 6648 (2012). RFC 6648 states new parameters SHOULD NOT use the “X-” prefix, that's why I just simply call it "Client-Version" here.

micaelae (approved):

PR #6794, opened 2025-10-07
fix: fix check sum addresses on api fetcher

salimtb:
## Explanation

**Current State & Problem:**
The `AccountsApiBalanceFetcher` was creating duplicate native token entries in balance results - one with a checksummed address and another with the original mixed/lowercase address from the API response. This occurred because of inconsistent address casing handling when processing balance data from the multi-chain accounts API.

**Root Cause:**
On line 299 in `api-balance-fetcher.ts`, the account address extracted from the API response (`b.accountAddress?.split(':')[2]`) was not being checksummed before use. This led to:

1. **Storage**: Mixed-case addresses were used as keys in the `nativeBalancesFromAPI` Map (e.g., `0xabc123...`)
2. **Lookup**: Checksummed addresses from `addressChainMap` were used for lookups (e.g., `0xAbC123...`) 
3. **Result**: Key mismatch caused lookup failures, leading to duplicate balance entries for the same native token

**Solution:**
The fix ensures consistent address formatting by applying the `checksum()` function to the account address immediately after extraction from the API response. This guarantees that:

- All addresses stored in `nativeBalancesFromAPI` are consistently checksummed
- Subsequent lookups using checksummed addresses from `addressChainMap` will find existing entries
- Eliminates duplicate native token balance entries

**Technical Details:**
The change is minimal but critical - replacing `b.accountAddress?.split(':')[2] as ChecksumAddress` with `checksum(b.accountAddress?.split(':')[2] || '')` ensures the extracted address is properly checksummed before being used as a map key or in balance processing logic.

## Checklist

**Note:** This is a bug fix that maintains existing API behavior while eliminating duplicate entries. No breaking changes or consumer updates are required.

---

> [!NOTE]
> Checksums the API-provided account address in `AccountsApiBalanceFetcher` to avoid duplicate native token entries and updates the changelog.
> 
> - **Fix**
>   - Normalize `accountAddress` via `checksum()` in `packages/assets-controllers/src/multi-chain-accounts-service/api-balance-fetcher.ts` when processing API balances to ensure consistent keys and prevent duplicate native token entries.
> - **Changelog**
>   - Add entry under `Fixed` describing the duplicate native token fix in `packages/assets-controllers/CHANGELOG.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d23d9a6d58b2ca44a0a248cd48bb107f75459ef3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,13 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-
-- Fix duplicate native token entries in `AccountsApiBalanceFetcher` by ensuring consistent address checksumming ([#6794](https://github.com/MetaMask/core/pull/6794))
-
 ### Changed
 
 - Bump `@metamask/multichain-account-service` from `^1.5.0` to `^1.6.0` ([#6786](https://github.com/MetaMask/core/pull/6786))
+
+### Fixed
+
+- Fix duplicate native token entries in `AccountsApiBalanceFetcher` by ensuring consistent address checksumming ([#6794](https://github.com/MetaMask/core/pull/6794))
 
 ## [78.0.0]
 


packages/assets-controllers/src/multi-chain-accounts-service/api-balance-fetcher.ts
```diff
@@ -296,10 +296,11 @@ export class AccountsApiBalanceFetcher implements BalanceFetcher {
     // Process regular API balances
     if (balances) {
       const apiBalances = balances.flatMap((b) => {
-        const account = b.accountAddress?.split(':')[2] as ChecksumAddress;
-        if (!account) {
+        const addressPart = b.accountAddress?.split(':')[2];
```
Prithpal-Sooriya, line 299: Approving, but I wonder if there is a util we can use instead here?

Prithpal-Sooriya (approved):

packages/assets-controllers/src/multi-chain-accounts-service/api-balance-fetcher.ts
salimtb: we should create one for hex addresses on the util package

PR #6813, opened 2025-10-10
fix: prune wallet metadata on account removal

fabiobozzo:
## Explanation

Fix wallet metadata cleanup bug in AccountTreeController.  
When wallets were completely removed (all groups pruned), their custom metadata (names, etc.) was not being cleaned up from `accountWalletsMetadata`, causing state pollution.

The `#pruneEmptyGroupAndWallet` method now properly removes wallet metadata when wallets are deleted, ensuring complete cleanup during account removal operations.

## References

Related to Ledger naming issues mentioned during 2025-10-10 SU: this ensures wallet metadata doesn't persist after wallet removal.

## Checklist

---

> [!NOTE]
> Removes `accountWalletsMetadata` for a wallet when its last group is pruned, with a new test and changelog entry.
> 
> - **Account Tree Controller**:
>   - **Wallet pruning**: In `src/AccountTreeController.ts`, `#pruneEmptyGroupAndWallet` now deletes `state.accountWalletsMetadata[walletId]` when removing an empty wallet.
>   - **Tests**: Add test to verify custom wallet metadata is cleared upon wallet removal in `src/AccountTreeController.test.ts`.
> - **Docs**:
>   - Update `CHANGELOG.md` under Unreleased with the fix entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4a1cd94c430bf97bf233cdd419aa2a54fa763325. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/account-tree-controller/CHANGELOG.md
+++ b/packages/account-tree-controller/CHANGELOG.md
@@ -7,12 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix wallet metadata cleanup when wallets are completely removed ([#6813](https://github.com/MetaMask/core/pull/6813))
+
 ## [1.4.2]
 
 ### Fixed
 
-- Fix wallet metadata cleanup when wallets are completely removed ([#6813](https://github.com/MetaMask/core/pull/6813))
-  - Wallet metadata (custom names, etc.) is now properly cleaned up from `accountWalletsMetadata` when wallets are pruned
 - Ensure `isLegacyAccountSyncingDisabled` is always set in `UserStorageSyncedWallet` after one successful full sync ([#6805](https://github.com/MetaMask/core/pull/6805))
   - This was not set in some rare edge case scenarios, and created situations were legacy syncs would always be re-triggered during full syncs.
   - We now verify this field is correctly set, and also catch empty objects for `UserStorageSyncedWallet`.


packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -891,6 +891,8 @@ export class AccountTreeController extends BaseController<
 
     if (Object.keys(wallets[walletId].groups).length === 0) {
       delete wallets[walletId];
+      // Clean up metadata for the pruned wallet
+      delete state.accountWalletsMetadata[walletId];
```
mathieuartu, line 895: I think this needs to be wrapped with a `this.update((state) => { ... }` in order to be properly propagated.
Examples [here](https://github.com/MetaMask/core/blob/544be71c53d06a2455de324bc6b6cab26b457886/packages/assets-controllers/src/TokensController.ts#L342-L346), [here](https://github.com/MetaMask/core/blob/a5e9de68b95ff3e97c810aa5c4639040b8da876d/packages/keyring-controller/src/KeyringController.ts#L1197-L1202) and [here](https://github.com/MetaMask/core/blob/03a71ae848af22b503807ee9f5c504dd2c470b39/packages/preferences-controller/src/PreferencesController.ts#L467-L472).

Although it's not _always_ the case, it's most often. Should we do this? If yes, then we should also consider wrapping the whole method since we also delete state properties L886 and L890.

LMK what you think

fabiobozzo: The state parameter passed to `#pruneEmptyGroupAndWallet` is already the mutable state object from the update() callback, so direct mutations like my new delete... are correct, they should be properly propagated.

If I understand correctly, those 3 examples are public methods that need their own update() wrapper because they're entry points. pruneEmptyGroupAndWallet is a private helper method called from within an existing update context 
```
handleAccountRemoved(accountId: AccountId) {
  ... 
  if (context) {
    ...
    this.update((state) => {
```

mathieuartu: Thanks, didn't unfold enough to see that 😄 My bad!

PR #6819, opened 2025-10-10
feat(core-backend): update websocket behavior

Kriys94:
### Current State & Motivation

The `core-backend` package had several architectural issues that needed to be addressed:

1. **Inflexible Chain Management**: `AccountActivityService` was proactively fetching supported chains from the Accounts API (`/v2/supportedNetworks`) with a hardcoded fallback list. This approach was:
   - Out of sync with backend capabilities (client-side caching vs. real-time backend state)
   - Unnecessarily complex with cache expiration logic
   - Made assumptions about supported chains rather than reacting to actual backend status

2. **Incomplete Connection Management**: `BackendWebSocketService` only subscribed to `AuthenticationController:stateChange` events but didn't directly respond to wallet lock/unlock events from `KeyringController`. This created a dependency where connection management relied solely on authentication state changes.

3. **Reconnection Logic Based on Close Codes**: The service used WebSocket close codes to determine reconnection behavior, which was fragile and didn't properly distinguish between manual disconnects (user action) and unexpected disconnects (network issues).

4. **Type Mismatches**: The `Transaction` and `Asset` types didn't match the backend API contract, causing potential integration issues.

5. **No Performance Monitoring**: There was no way to trace WebSocket operations for performance analysis in production environments.

6. **Inconsistent Address Casing**: WebSocket-based token balance and detection updates had inconsistent address normalization, causing issues with state lookups and controller integration.

#### 1. System Notification-Driven Chain Tracking (`AccountActivityService`)

**Before**: Client proactively fetches and caches supported chains
```typescript
async getSupportedChains(): Promise<string[]> {
  // Fetch from API with cache + fallback to hardcoded list
}
```

**After**: Client reacts to backend system notifications
```typescript
#chainsUp: Set<string> = new Set();

// Backend automatically sends system notifications with chain status
// Service tracks chains dynamically based on real-time backend state
```

**Benefits**:
- Real-time chain status that matches backend capabilities
- No stale cache issues
- Simpler code (removed ~100 lines)
- Backend has single source of truth

**Flow**:
1. Client subscribes to account activity
2. Backend automatically sends system notification: `{chainIds: ['eip155:1', 'eip155:137'], status: 'up'}`
3. Service tracks these chains internally
4. On disconnect, service flushes all tracked chains as 'down'
5. On reconnect, backend sends fresh system notification with current status

#### 2. Simplified Reconnection Logic (`BackendWebSocketService`)

**Before**: Reconnection based on close codes
```typescript
#shouldReconnectOnClose(code: number): boolean {
  return code !== 1000; // Only don't reconnect on normal closure
}
```

**After**: Reconnection based on manual disconnect flag
```typescript
#manualDisconnect = false;

connect() {
  this.#manualDisconnect = false; // Reset flag on explicit connect
}

disconnect() {
  this.#manualDisconnect = true; // Set flag to prevent auto-reconnect
}

onclose() {
  if (!this.#manualDisconnect) {
    this.#scheduleReconnect(); // Auto-reconnect unless manually disconnected
  }
}
```

**Benefits**:
- Clear intent: manual disconnects stay disconnected, unexpected disconnects auto-reconnect
- No ambiguity with close codes
- Simpler logic

#### 3. KeyringController Event Integration

Added direct subscriptions to wallet lock state:

```typescript
#subscribeEvents() {
  // Sign in/out
  this.#messenger.subscribe('AuthenticationController:stateChange', ...);
  
  // Wallet lock/unlock (NEW)
  this.#messenger.subscribe('KeyringController:unlock', () => this.connect());
  this.#messenger.subscribe('KeyringController:lock', () => this.disconnect());
}
```

**Connection Requirements** (all must be true):
1. ✅ Feature enabled (`isEnabled()` callback)
2. ✅ User signed in (`AuthenticationController.isSignedIn`)
3. ✅ Wallet unlocked (`KeyringController` state)

#### 4. Type Alignment with Backend API

- `Transaction.hash` → `Transaction.id` (matches backend field name)
- `Asset` now requires `decimals: number` (needed for proper token formatting)

#### 5. Performance Tracing Integration

Added optional `traceFn` parameter to enable performance monitoring:

```typescript
new BackendWebSocketService({
  messenger,
  url: BACKEND_WS_URL,
  traceFn: trace as TraceCallback, // Inject platform-specific trace function
  // ...other options
});
```

**Benefits**:
- Platform-agnostic tracing support (enables Sentry integration in extension/mobile)
- Traces WebSocket operations (connect, disconnect methods)
- Enables performance monitoring in production
- Core package remains platform-agnostic with no-op default

**Integration**:
- **Extension**: Uses `trace` from `shared/lib/trace.ts` (Sentry-backed)
- **Mobile**: Uses `trace` from `app/util/trace.ts` (Sentry-backed)
- **Core**: Defaults to no-op if not provided, keeping the package platform-agnostic

**Example usage**:
```typescript
// Extension/Mobile init files inject their platform-specific trace function
const service = new BackendWebSocketService({
  messenger,
  url: BACKEND_WS_URL,
  traceFn: trace as TraceCallback,
  // When operations occur, they're automatically traced:
  // - "BackendWebSocketService:connect"
  // - "BackendWebSocketService:disconnect"
});
```

#### 6. Address Normalization for WebSocket Updates (`assets-controllers`)

Fixed inconsistent address casing in WebSocket-based token balance and detection updates to ensure proper state lookups and controller integration.

**TokenBalancesController** (`#onAccountActivityBalanceUpdate`):
- **Account addresses**: Normalized to lowercase for `tokenBalances` state lookups
- **Token addresses**: Normalized to checksum format for `tokenBalances` state storage
- **Native balance updates**: Uses checksummed account addresses for `AccountTrackerController:updateNativeBalances`
- **New tokens**: Uses checksummed token addresses for `TokensController:addTokens`

**TokenDetectionController** (`addDetectedTokensViaWs`):
- **Token addresses**: Explicitly normalized to checksum format before calling `TokensController:addTokens`
- **Token cache lookups**: Uses lowercase token addresses for `tokensChainsCache` lookups
- **Fresh cache**: Fetches `tokensChainsCache` at the start of the method to prevent stale data

**Benefits**:
- Consistent address format across different state stores
- Prevents state lookup failures due to casing mismatches
- Ensures compatibility between controllers with different address format expectations
- Robust handling without assumptions about input format

**Example**:
```typescript
// Before: Address casing was inconsistent
state.tokenBalances[address][tokenAddress] = balance; // Could fail if casing mismatched

// After: Explicit normalization
const normalizedAccount = address.toLowerCase();
const checksummedToken = toChecksumAddress(tokenAddress);
state.tokenBalances[normalizedAccount as Hex][checksummedToken] = balance;
```

### Breaking Changes Impact

**For `AccountActivityService` consumers**:
- ❌ **Removed**: `getSupportedChains()` method - consumers should subscribe to `AccountActivityService:statusChanged` events instead
- Chain status is now event-driven rather than request-response

**For type definitions**:
- `Transaction` objects must use `id` instead of `hash`
- `Asset` objects must include `decimals` field

### Testing Updates

- Removed `nock` dependency (no longer fetching from external API)
- Updated all tests to use system notification-driven flow
- Added tests for KeyringController event integration
- Added tests for manual disconnect flag behavior
- Increased test coverage for edge cases in reconnection logic
- Updated `TokenDetectionController` tests to expect checksummed addresses in `addTokens` calls

## References

- Part of ongoing core-backend stabilization effort
- Aligns with backend system notification architecture

## Checklist

---

> [!NOTE]
> Revamps core-backend WebSocket lifecycle (keyring lock/unlock, manual disconnect, tracing, new params/types, system notification–driven chains) and fixes assets controllers to normalize addresses for WebSocket token updates; docs/tests updated.
> 
> - **Core Backend (`@metamask/core-backend`)**:
>   - **WebSocket**: New connection model (idempotent connect, auto-reconnect unless manual disconnect), subscribes to `KeyringController:lock/unlock`; exposes richer `getConnectionInfo`.
>   - **API Changes (BREAKING)**:
>     - `BackendWebSocketService.subscribe` now requires `channelType`.
>     - Types: `Transaction.hash` → `id`; `Asset` adds required `decimals`; `ServerNotificationMessage` includes `timestamp`.
>     - Messenger/events expanded; add peer dep on `@metamask/keyring-controller`.
>   - **Tracing**: Optional `traceFn` to instrument connect/disconnect/notifications.
>   - **AccountActivityService**: Drops `getSupportedChains`; chains tracked via system notifications; publishes `statusChanged` with optional `timestamp`.
>   - **Docs**: README updated with new flow and “WebSocket Connection Management”.
> - **Assets Controllers**:
>   - **TokenBalancesController**: Normalizes account (lowercase) and token (checksum) for WS updates; fixes tracking checks; updates native balance propagation; adds decimals handling in tests.
>   - **TokenDetectionController**: Adds checksummed addresses and lowercase cache lookups in `addDetectedTokensViaWs`; metrics use checksummed addresses.
>   - **Changelog**: Notes fix for address casing in WS balance updates.
> - **Repo**:
>   - README: adds `@metamask/core-backend` to package list and dependency graph; minor graph edge additions.
>   - Build/config: add keyring-controller refs; adjust dev/peer deps; test updates to reflect new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16e8b323623f91ed9f7a08eb90a37b5e021de01d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/TokenBalancesController.ts
+++ b/packages/assets-controllers/src/TokenBalancesController.ts
@@ -844,12 +844,10 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     account: ChecksumAddress,
     chainId: ChainIdHex,
   ): boolean {
-    const normalizedAddress = tokenAddress.toLowerCase();
-
     // Check if token exists in allTokens
     if (
       this.#allTokens?.[chainId]?.[account.toLowerCase()]?.some(
-        (token) => token.address.toLowerCase() === normalizedAddress,
+        (token) => token.address === tokenAddress,
       )
     ) {
       return true;
@@ -858,7 +856,7 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     // Check if token exists in allIgnoredTokens
     if (
       this.#allIgnoredTokens?.[chainId]?.[account.toLowerCase()]?.some(
-        (addr) => addr.toLowerCase() === normalizedAddress,
+        (token) => token === tokenAddress,
       )
     ) {
       return true;
@@ -1124,23 +1122,26 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     updates: BalanceUpdate[];
   }) => {
     const chainId = caipChainIdToHex(chain);
-    const account = checksum(address);
+    const checksummedAccount = checksum(address);
 
     try {
       // Process all balance updates at once
       const { tokenBalances, newTokens, nativeBalanceUpdates } =
-        this.#prepareBalanceUpdates(updates, account, chainId);
+        this.#prepareBalanceUpdates(updates, checksummedAccount, chainId);
 
       // Update state once with all token balances
       if (tokenBalances.length > 0) {
         this.update((state) => {
-          // Initialize account and chain structure
-          state.tokenBalances[account] ??= {};
-          state.tokenBalances[account][chainId] ??= {};
+          // Temporary until ADR to normalize all keys - tokenBalances state requires: account in lowercase, token in checksum
+          const lowercaseAccount =
+            checksummedAccount.toLowerCase() as ChecksumAddress;
+          state.tokenBalances[lowercaseAccount] ??= {};
+          state.tokenBalances[lowercaseAccount][chainId] ??= {};
 
           // Apply all token balance updates
           for (const { tokenAddress, balance } of tokenBalances) {
-            state.tokenBalances[account][chainId][tokenAddress] = balance;
+            state.tokenBalances[lowercaseAccount][chainId][tokenAddress] =
+              balance;
           }
         });
       }
--- a/packages/core-backend/CHANGELOG.md
+++ b/packages/core-backend/CHANGELOG.md
@@ -7,34 +7,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Added
 
-- **BREAKING**: `BackendWebSocketService` - Simplified connection management and added KeyringController event integration ([#6819](https://github.com/MetaMask/core/pull/6819))
-  - Added `KeyringController:lock` and `KeyringController:unlock` event subscriptions to automatically manage WebSocket connections based on wallet lock state
-  - Renamed internal method `setupAuthentication()` to `subscribeEvents()` to reflect broader event handling responsibilities
-  - Simplified reconnection logic: auto-reconnect on any unexpected disconnect, stay disconnected on manual disconnects (tracked via `#manualDisconnect` flag)
-  - Updated `connect()` to reset manual disconnect flag, allowing reconnection after previous manual disconnects
-  - Updated `disconnect()` to set manual disconnect flag, preventing automatic reconnection
-  - Improved error handling in `connect()` to properly rethrow errors to callers
-- **BREAKING**: `AccountActivityService` - Replaced API-based chain support detection with system notification-driven chain tracking ([#6819](https://github.com/MetaMask/core/pull/6819))
-  - Added internal `#chainsUp` Set to track chains reported as 'up' via system notifications
-  - Updated system notification handler to dynamically track chain status (add to set when 'up', remove when 'down')
-  - Updated WebSocket state change handler to flush all tracked chains as 'down' on disconnect/error (instead of using hardcoded list)
-  - Chain status is now entirely driven by backend system notifications rather than proactive API calls
-- **BREAKING**: Updated `Transaction` type definition - renamed `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
-- **BREAKING**: Updated `Asset` type definition - added required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
-- `BackendWebSocketService` - Added optional `traceFn` parameter to constructor for performance tracing integration (e.g., Sentry)
+- **BREAKING:** Add required argument `channelType` to `BackendWebSocketService.subscribe` method ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Add `channelType` to argument of the `BackendWebSocketService:subscribe` messenger action
+  - Add `channelType` to `WebSocketSubscription` type
+- **BREAKING**: Update `Asset` type definition: add required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Add optional `traceFn` parameter to `BackendWebSocketService` constructor for performance tracing integration (e.g., Sentry) ([#6819](https://github.com/MetaMask/core/pull/6819))
   - Enables tracing of WebSocket operations including connect, disconnect methods
   - Trace function receives operation metadata and callback to wrap for performance monitoring
-- Updated documentation (README.md) to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
-  - Added "WebSocket Connection Management" section explaining connection requirements and behavior
-  - Updated sequence diagram to show system notification-driven chain status flow
-  - Updated key flow characteristics to reflect internal chain tracking mechanism
+- Add optional `timestamp` property to `ServerNotificationMessage` and `SystemNoticationData` types ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Add optional `timestamp` property to `AccountActivityService:statusChanged` event and corresponding event type ([#6819](https://github.com/MetaMask/core/pull/6819))
+
+### Changed
+
+- **BREAKING:** Update `BackendWebSocketService` to automatically manage WebSocket connections based on wallet lock state ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - `KeyringController:lock` and `KeyringController:unlock` are now required events in the `BackendWebSocketService` messenger
+- **BREAKING**: Update `Transaction` type definition: rename `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
+- **BREAKING:** Add peer dependency on `@metamask/keyring-controller` (^23.0.0) ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Update `BackendWebSocketService` to simplify reconnection logic: auto-reconnect on any unexpected disconnect (not just code 1000), stay disconnected when manually disconnecting via `disconnect` ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Improve error handling in `BackendWebSocketService.connect()` to properly rethrow errors to callers ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Update `AccountActivityService` to replace API-based chain support detection with system notification-driven chain tracking ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Instead of hardcoding a list of supported chains, assume that the backend has the list
+  - When receiving a system notification, capture the backend-tracked status of each chain instead of assuming it is up or down
+  - Flush all tracked chains as 'down' on disconnect/error (instead of using hardcoded list)
+- Update documentation in `README.md` to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Add "WebSocket Connection Management" section explaining connection requirements and behavior
+  - Update sequence diagram to show system notification-driven chain status flow
+  - Update key flow characteristics to reflect internal chain tracking mechanism
 
 ### Removed
 
-- **BREAKING**: Removed `getSupportedChains()` public method and all related API fetching logic
-- **BREAKING**: Removed hardcoded `DEFAULT_SUPPORTED_CHAINS` fallback list and cache expiration mechanism
+- **BREAKING**: Remove `getSupportedChains` method from `AccountActivityService` ([#6819](https://github.com/MetaMask/core/pull/6819))
 
 ## [1.0.1]
 
--- a/packages/core-backend/src/AccountActivityService.test.ts
+++ b/packages/core-backend/src/AccountActivityService.test.ts
@@ -321,8 +321,12 @@ describe('AccountActivityService', () => {
         expect(service.name).toBe('AccountActivityService');
 
         // Status changed event is only published when WebSocket connects
-        const publishSpy = jest.spyOn(messenger, 'publish');
-        expect(publishSpy).not.toHaveBeenCalled();
+        const statusChangedEventListener = jest.fn();
+        messenger.subscribe(
+          'AccountActivityService:statusChanged',
+          statusChangedEventListener,
+        );
+        expect(statusChangedEventListener).not.toHaveBeenCalled();
 
         // Verify system notification callback was registered
         expect(mocks.addChannelCallback).toHaveBeenCalledWith({
@@ -571,10 +575,12 @@ describe('AccountActivityService', () => {
 
       it('should track chains as up and down based on system notifications', async () => {
         await withService(async ({ messenger, mocks }) => {
-          const publishSpy = jest.spyOn(messenger, 'publish');
+          const statusChangedEventListener = jest.fn();
+          messenger.subscribe(
+            'AccountActivityService:statusChanged',
+            statusChangedEventListener,
+          );
           const systemCallback = getSystemNotificationCallback(mocks);
-
-          publishSpy.mockClear();
 
           // Simulate chains coming up
           const timestamp1 = 1760344704595;
@@ -588,16 +594,11 @@ describe('AccountActivityService', () => {
             timestamp: timestamp1,
           });
 
-          expect(publishSpy).toHaveBeenCalledWith(
-            'AccountActivityService:statusChanged',
-            {
-              chainIds: ['eip155:1', 'eip155:137'],
-              status: 'up',
-              timestamp: timestamp1,
-            },
-          );
-
-          publishSpy.mockClear();
+          expect(statusChangedEventListener).toHaveBeenCalledWith({
+            chainIds: ['eip155:1', 'eip155:137'],
+            status: 'up',
+            timestamp: timestamp1,
+          });
 
           // Simulate one chain going down
           const timestamp2 = 1760344704696;
@@ -611,14 +612,11 @@ describe('AccountActivityService', () => {
             timestamp: timestamp2,
           });
 
-          expect(publishSpy).toHaveBeenCalledWith(
-            'AccountActivityService:statusChanged',
-            {
-              chainIds: ['eip155:137'],
-              status: 'down',
-              timestamp: timestamp2,
-            },
-          );
+          expect(statusChangedEventListener).toHaveBeenCalledWith({
+            chainIds: ['eip155:137'],
+            status: 'down',
+            timestamp: timestamp2,
+          });
         });
       });
     });
@@ -626,12 +624,13 @@ describe('AccountActivityService', () => {
     describe('handleWebSocketStateChange', () => {
       it('should handle WebSocket ERROR state by publishing tracked chains as down', async () => {
         await withService(async ({ messenger, rootMessenger, mocks }) => {
-          const publishSpy = jest.spyOn(messenger, 'publish');
+          const statusChangedEventListener = jest.fn();
+          messenger.subscribe(
+            'AccountActivityService:statusChanged',
+            statusChangedEventListener,
+          );
 
           mocks.getSelectedAccount.mockReturnValue(null);
-
-          // Clear any publish calls from service initialization
-          publishSpy.mockClear();
 
           // First, simulate receiving a system notification with chains up
           const systemCallback = getSystemNotificationCallback(mocks);
@@ -645,8 +644,6 @@ describe('AccountActivityService', () => {
             timestamp: 1760344704595,
           });
 
-          publishSpy.mockClear();
-
           // Publish WebSocket ERROR state event - should flush tracked chains as down
           await rootMessenger.publish(
             'BackendWebSocketService:connectionStateChanged',
@@ -654,29 +651,32 @@ describe('AccountActivityService', () => {
               state: WebSocketState.ERROR,
               url: 'ws://test',
               reconnectAttempts: 2,
+              timeout: 10000,
+              reconnectDelay: 500,
+              maxReconnectDelay: 5000,
+              requestTimeout: 30000,
             },
           );
           await completeAsyncOperations(100);
 
           // Verify that the ERROR state triggered the status change for tracked chains
-          expect(publishSpy).toHaveBeenCalledWith(
-            'AccountActivityService:statusChanged',
-            {
-              chainIds: ['eip155:1', 'eip155:137', 'eip155:56'],
-              status: 'down',
-            },
-          );
+          expect(statusChangedEventListener).toHaveBeenCalledWith({
+            chainIds: ['eip155:1', 'eip155:137', 'eip155:56'],
+            status: 'down',
+            timestamp: expect.any(Number),
+          });
         });
       });
 
       it('should not publish status change on disconnect when no chains are tracked', async () => {
         await withService(async ({ messenger, rootMessenger, mocks }) => {
-          const publishSpy = jest.spyOn(messenger, 'publish');
+          const statusChangedEventListener = jest.fn();
+          messenger.subscribe(
+            'AccountActivityService:statusChanged',
+            statusChangedEventListener,
+          );
 
           mocks.getSelectedAccount.mockReturnValue(null);
-
-          // Clear any publish calls from service initialization
-          publishSpy.mockClear();
 
           // Publish WebSocket ERROR state event without any tracked chains
           await rootMessenger.publish(
@@ -685,15 +685,16 @@ describe('AccountActivityService', () => {
               state: WebSocketState.ERROR,
               url: 'ws://test',
               reconnectAttempts: 2,
+              timeout: 10000,
+              reconnectDelay: 500,
+              maxReconnectDelay: 5000,
+              requestTimeout: 30000,
             },
           );
           await completeAsyncOperations(100);
 
           // Verify that no status change was published since no chains were tracked
-          expect(publishSpy).not.toHaveBeenCalledWith(
-            'AccountActivityService:statusChanged',
-            expect.anything(),
-          );
+          expect(statusChangedEventListener).not.toHaveBeenCalled();
         });
       });
     });
@@ -787,6 +788,10 @@ describe('AccountActivityService', () => {
               state: WebSocketState.CONNECTED,
               url: 'ws://test',
               reconnectAttempts: 0,
+              timeout: 10000,
+              reconnectDelay: 500,
+              maxReconnectDelay: 5000,
+              requestTimeout: 30000,
             },
           );
           // Wait for async handler to complete
@@ -795,38 +800,6 @@ describe('AccountActivityService', () => {
           // Should attempt to get selected account even when none exists
           expect(mocks.getSelectedAccount).toHaveBeenCalledTimes(1);
           expect(mocks.getSelectedAccount).toHaveReturnedWith(null);
-        });
-      });
-
-      it('should handle system notification publish failures gracefully by throwing error when publish fails', async () => {
-        await withService(async ({ mocks, messenger }) => {
-          const systemCallback = getSystemNotificationCallback(mocks);
-
-          // Mock publish to throw error
-          jest.spyOn(messenger, 'publish').mockImplementation(() => {
-            throw new Error('Publish failed');
-          });
-
-          const systemNotification = {
-            event: 'system-notification',
-            channel: 'system-notifications.v1.account-activity.v1',
-            data: { chainIds: ['0x1', '0x2'], status: 'connected' },
-            timestamp: 1760344704595,
-          };
-
-          // Should throw error when publish fails
-          expect(() => systemCallback(systemNotification)).toThrow(
-            'Publish failed',
-          );
-
-          // Should have attempted to publish the notification
-          expect(messenger.publish).toHaveBeenCalledWith(
-            expect.any(String),
-            expect.objectContaining({
-              chainIds: ['0x1', '0x2'],
-              status: 'connected',
-            }),
-          );
         });
       });
 
@@ -935,6 +908,10 @@ describe('AccountActivityService', () => {
                 state: WebSocketState.CONNECTED,
                 url: 'ws://test',
                 reconnectAttempts: 0,
+                timeout: 10000,
+                reconnectDelay: 500,
+                maxReconnectDelay: 5000,
+                requestTimeout: 30000,
               },
             );
             await completeAsyncOperations();
--- a/packages/core-backend/src/BackendWebSocketService.test.ts
+++ b/packages/core-backend/src/BackendWebSocketService.test.ts
@@ -247,10 +247,6 @@ type TestSetup = {
   mocks: {
     getBearerToken: jest.Mock;
   };
-  spies: {
-    publish: jest.SpyInstance;
-    call: jest.SpyInstance;
-  };
   completeAsyncOperations: (advanceMs?: number) => Promise<void>;
   getMockWebSocket: () => MockWebSocket;
   cleanup: () => void;
@@ -268,10 +264,6 @@ type WithServiceCallback<ReturnValue> = (payload: {
   >;
   mocks: {
     getBearerToken: jest.Mock;
-  };
-  spies: {
-    publish: jest.SpyInstance;
-    call: jest.SpyInstance;
   };
   completeAsyncOperations: (advanceMs?: number) => Promise<void>;
   getMockWebSocket: () => MockWebSocket;
@@ -296,10 +288,6 @@ const setupBackendWebSocketService = ({
   // Create real messenger with registered actions
   const messengerSetup = getMessenger();
   const { rootMessenger, messenger, mocks } = messengerSetup;
-
-  // Create spies BEFORE service construction to capture constructor calls
-  const publishSpy = jest.spyOn(messenger, 'publish');
-  const callSpy = jest.spyOn(messenger, 'call');
 
   // Default test options (shorter timeouts for faster tests)
   const defaultOptions = {
@@ -344,16 +332,10 @@ const setupBackendWebSocketService = ({
     messenger,
     rootMessenger,
     mocks,
-    spies: {
-      publish: publishSpy,
-      call: callSpy,
-    },
     completeAsyncOperations,
     getMockWebSocket,
     cleanup: () => {
       service?.destroy();
-      publishSpy.mockRestore();
-      callSpy.mockRestore();
       jest.useRealTimers();
       jest.clearAllMocks();
     },
@@ -387,7 +369,6 @@ async function withService<ReturnValue>(
       messenger: setup.messenger,
       rootMessenger: setup.rootMessenger,
       mocks: setup.mocks,
-      spies: setup.spies,
       completeAsyncOperations: setup.completeAsyncOperations,
       getMockWebSocket: setup.getMockWebSocket,
     });
@@ -473,39 +454,25 @@ describe('BackendWebSocketService', () => {
     });
 
     it('should apply default values for options not provided', async () => {
-      const { messenger } = getMessenger();
-
-      // Create service with only url and messenger provided
-      // Should use defaults: timeout=10000, reconnectDelay=500, maxReconnectDelay=5000, requestTimeout=30000
-      const service = new BackendWebSocketService({
-        messenger,
-        url: 'ws://test.example.com',
-        // timeout, reconnectDelay, maxReconnectDelay, requestTimeout not provided - will use defaults
-      });
-
-      expect(service.getConnectionInfo().url).toBe('ws://test.example.com');
-      expect(service).toBeInstanceOf(BackendWebSocketService);
-
-      service.destroy();
-    });
-
-    it('should accept traceFn option for performance tracing', async () => {
-      const mockTraceFn = jest.fn((_request, fn) => {
-        // Mock trace function that just executes the callback
-        return fn?.();
-      });
-
-      const { messenger } = getMessenger();
-
-      const service = new BackendWebSocketService({
-        messenger,
-        url: 'ws://test.example.com',
-        traceFn: mockTraceFn,
-      });
-
-      expect(service).toBeInstanceOf(BackendWebSocketService);
-
-      service.destroy();
+      await withService(
+        {
+          options: {
+            url: 'ws://test.example.com',
+            timeout: undefined,
+            reconnectDelay: undefined,
+            maxReconnectDelay: undefined,
+            requestTimeout: undefined,
+          },
+        },
+        async ({ service }) => {
+          expect(service.getConnectionInfo().url).toBe('ws://test.example.com');
+          expect(service.getConnectionInfo().timeout).toBe(10000);
+          expect(service.getConnectionInfo().reconnectDelay).toBe(500);
+          expect(service.getConnectionInfo().maxReconnectDelay).toBe(5000);
+          expect(service.getConnectionInfo().requestTimeout).toBe(30000);
+          expect(service).toBeInstanceOf(BackendWebSocketService);
+        },
+      );
     });
   });
 
@@ -514,7 +481,13 @@ describe('BackendWebSocketService', () => {
   // =====================================================
   describe('connection lifecycle - connect / disconnect', () => {
     it('should establish WebSocket connection and set state to CONNECTED, publishing state change event', async () => {
-      await withService(async ({ service, spies }) => {
+      await withService(async ({ service, messenger }) => {
+        const connectionStateChangedListener = jest.fn();
+        messenger.subscribe(
+          'BackendWebSocketService:connectionStateChanged',
+          connectionStateChangedListener,
+        );
+
         await service.connect();
 
         const connectionInfo = service.getConnectionInfo();
@@ -522,8 +495,7 @@ describe('BackendWebSocketService', () => {
         expect(connectionInfo.reconnectAttempts).toBe(0);
         expect(connectionInfo.url).toBe('ws://localhost:8080');
 
-        expect(spies.publish).toHaveBeenCalledWith(
-          'BackendWebSocketService:connectionStateChanged',
+        expect(connectionStateChangedListener).toHaveBeenCalledWith(
           expect.objectContaining({
             state: WebSocketState.CONNECTED,
             reconnectAttempts: 0,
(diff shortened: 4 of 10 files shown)


packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -456,6 +479,9 @@ export class BackendWebSocketService {
       return;
     }
 
+    // Mark this as a manual disconnect to prevent automatic reconnection
+    this.#manualDisconnect = true;
+
```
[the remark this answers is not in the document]
Kriys94: Ack

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -926,24 +949,19 @@ export class BackendWebSocketService {
...
-        } else {
-          // Handle runtime errors
-          this.#handleError(new Error(`WebSocket error: ${event.type}`));
+        // Handle connection-phase errors
+        if (this.#connectionTimeout) {
+          clearTimeout(this.#connectionTimeout);
+          this.#connectionTimeout = null;
         }
+        const error = new Error(`WebSocket connection error to ${wsUrl}`);
+        reject(error);
```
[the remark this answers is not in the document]
Kriys94: Ack

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -217,6 +221,8 @@ export type BackendWebSocketServiceAllowedActions =
 
 export type BackendWebSocketServiceAllowedEvents =
   | AuthenticationController.AuthenticationControllerStateChangeEvent
+  | KeyringControllerLockEvent
+  | KeyringControllerUnlockEvent
   | BackendWebSocketServiceConnectionStateChangedEvent;
```
cryptodev-2s: Nit: This typically doesn’t belong here, since it’s an event owned by the service.
As it’s already defined in `BackendWebSocketServiceEvents`, we can safely remove it from this file.
```suggestion

File: packages/core-backend/src/BackendWebSocketService.ts
Comment (cryptodev-2s) [inline packages/core-backend/src/BackendWebSocketService.ts]: Nit: This typically doesn’t belong here, since it’s an action owned by the service.
As it’s already defined in `BackendWebSocketServiceActions`, we can safely remove it from this file.
```suggestion

File: packages/core-backend/src/BackendWebSocketService.ts [hunk 4]
```diff
@@ -217,6 +221,8 @@ export type BackendWebSocketServiceAllowedActions =
 
 export type BackendWebSocketServiceAllowedEvents =
   | AuthenticationController.AuthenticationControllerStateChangeEvent
+  | KeyringControllerLockEvent
```
cryptodev-2s, line 234: Should we add `@metamask/keyring-controller` as a `devDependency` and 
`peerDependency` + to `tsconfig.json` and `tsconfig.build.json`

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING**: Updated `Transaction` type definition - renamed `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
+- **BREAKING**: Updated `Asset` type definition - added required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Updated documentation (README.md) to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Added "WebSocket Connection Management" section explaining connection requirements and behavior
+  - Updated sequence diagram to show system notification-driven chain status flow
+  - Updated key flow characteristics to reflect internal chain tracking mechanism
+
+### Removed
+
+- Removed `nock` test dependency - no longer needed after removing API-based chain support fetching ([#6819](https://github.com/MetaMask/core/pull/6819))
```
cryptodev-2s: Nit: This is just a devDependency removal it can be omitted
```suggestion

File: packages/core-backend/CHANGELOG.md [hunk 6]
```diff
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Added `KeyringController:lock` and `KeyringController:unlock` event subscriptions to automatically manage WebSocket connections based on wallet lock state
+  - Renamed internal method `setupAuthentication()` to `subscribeEvents()` to reflect broader event handling responsibilities
+  - Simplified reconnection logic: auto-reconnect on any unexpected disconnect, stay disconnected on manual disconnects (tracked via `#manualDisconnect` flag)
+  - Removed `shouldReconnectOnClose()` method - reconnection decisions now based solely on manual disconnect flag rather than close codes
+  - Updated `connect()` to reset manual disconnect flag, allowing reconnection after previous manual disconnects
+  - Updated `disconnect()` to set manual disconnect flag, preventing automatic reconnection
+  - Improved error handling in `connect()` to properly rethrow errors to callers
+- **BREAKING**: `AccountActivityService` - Replaced API-based chain support detection with system notification-driven chain tracking ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Removed `getSupportedChains()` public method and all related API fetching logic
```
cryptodev-2s: Nit: should we put this under removed section instead

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING**: `BackendWebSocketService` - Simplified connection management and added KeyringController event integration ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Added `KeyringController:lock` and `KeyringController:unlock` event subscriptions to automatically manage WebSocket connections based on wallet lock state
+  - Renamed internal method `setupAuthentication()` to `subscribeEvents()` to reflect broader event handling responsibilities
+  - Simplified reconnection logic: auto-reconnect on any unexpected disconnect, stay disconnected on manual disconnects (tracked via `#manualDisconnect` flag)
+  - Removed `shouldReconnectOnClose()` method - reconnection decisions now based solely on manual disconnect flag rather than close codes
```
cryptodev-2s: Nit: should we put this under removed section instead

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Updated `connect()` to reset manual disconnect flag, allowing reconnection after previous manual disconnects
+  - Updated `disconnect()` to set manual disconnect flag, preventing automatic reconnection
+  - Improved error handling in `connect()` to properly rethrow errors to callers
+- **BREAKING**: `AccountActivityService` - Replaced API-based chain support detection with system notification-driven chain tracking ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Removed `getSupportedChains()` public method and all related API fetching logic
+  - Removed hardcoded `DEFAULT_SUPPORTED_CHAINS` fallback list and cache expiration mechanism
```
cryptodev-2s: Nit: should we put this under removed section instead

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -585,7 +622,7 @@ export class BackendWebSocketService {
       state: this.#state,
       url: this.#options.url,
       reconnectAttempts: this.#reconnectAttempts,
-      connectedAt: this.#connectedAt ?? undefined,
+      connectedAt: this.#connectedAt,
```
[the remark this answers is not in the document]
Kriys94: Ack

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `BackendWebSocketService` - Added optional `traceFn` parameter to constructor for performance tracing integration (e.g., Sentry)
```
cryptodev-2s: Nit: This is just a nice-to-have, feel free to disregard if you prefer.
We usually list Breaking Changes first, so if you could move this section right below those entries, that would be great.

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING**: Updated `Transaction` type definition - renamed `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
+- **BREAKING**: Updated `Asset` type definition - added required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
+- Updated documentation (README.md) to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Added "WebSocket Connection Management" section explaining connection requirements and behavior
+  - Updated sequence diagram to show system notification-driven chain status flow
+  - Updated key flow characteristics to reflect internal chain tracking mechanism
+
+### Removed
+
+- Removed `shouldReconnectOnClose()` method - reconnection decisions now based solely on manual disconnect flag rather than close codes
```
cryptodev-2s: Nit: Correct me if I’m wrong, but `#shouldReconnectOnClose` is a private method and doesn’t introduce any functional changes to the service’s behavior if removed.
If that’s the case, it can be omitted from the changelog entries.
```suggestion

File: packages/core-backend/CHANGELOG.md [hunk 12]
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Removed `getSupportedChains()` public method and all related API fetching logic
+- Removed hardcoded `DEFAULT_SUPPORTED_CHAINS` fallback list and cache expiration mechanism
```
cryptodev-2s: Nit: Should we mark these removal as breaking
```suggestion
- **BREAKING**: Removed `getSupportedChains()` public method and all related API fetching logic
- **BREAKING**: Removed hardcoded `DEFAULT_SUPPORTED_CHAINS` fallback list and cache expiration mechanism

File: packages/core-backend/CHANGELOG.md [hunk 13]
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Chain status is now entirely driven by backend system notifications rather than proactive API calls
+- **BREAKING**: Updated `Transaction` type definition - renamed `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
+- **BREAKING**: Updated `Asset` type definition - added required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
+- `BackendWebSocketService` - Added optional `traceFn` parameter to constructor for performance tracing integration (e.g., Sentry)
+  - Enables tracing of WebSocket operations including connect, disconnect methods
+  - Trace function receives operation metadata and callback to wrap for performance monitoring
+- Updated documentation (README.md) to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Added "WebSocket Connection Management" section explaining connection requirements and behavior
+  - Updated sequence diagram to show system notification-driven chain status flow
+  - Updated key flow characteristics to reflect internal chain tracking mechanism
```
mcmire: A few notes:

- Should we be more specific about which changes are breaking? For instance, simplifying the reconnection logic in itself doesn't seem to be breaking, at least to me, as is updating `connect`, `disconnect`, etc. (And to be clear, if a change is "breaking" it means that there is some change that needs to be made on the client side to adjust to the change or else an error at compile-time or runtime will occur.)
- For the breaking changes, can we explain what needs to change in consuming code to adapt to the changes if it is not clear?
- Is there anything that we don't need to document, e.g. changes to internal methods or developer-only changes?
- For changes that extend the API should we add them under Added?
- Can we ensure that all entries have a PR next to them?
- Can we use present tense instead of past tense for entries?

Putting this together, what if we do something like this:

``` suggestion
### Added

- **BREAKING:** Add required argument `channelType` to `BackendWebSocketService.subscribe` method ([#6819](https://github.com/MetaMask/core/pull/6819))
  - Add `channelType` to argument of the `BackendWebSocketService:subscribe` messenger action
  - Add `channelType` to `WebSocketSubscription` type
- **BREAKING**: Update `Asset` type definition: add required `decimals` field for proper token amount formatting ([#6819](https://github.com/MetaMask/core/pull/6819))
- Add optional `traceFn` parameter to `BackendWebSocketService` constructor for performance tracing integration (e.g., Sentry) ([#6819](https://github.com/MetaMask/core/pull/6819))
  - Enables tracing of WebSocket operations including connect, disconnect methods
  - Trace function receives operation metadata and callback to wrap for performance monitoring
- Add optional `timestamp` property to `ServerNotificationMessage` and `SystemNoticationData` types ([#6819](https://github.com/MetaMask/core/pull/6819))
- Add optional `timestamp` property to payload for callback to `BackendWebSocketService.addChannelCallback` and `BackendWebSocketService.subscribe`, as well as corresponding actions
- Add optional `timestamp` property to `AccountActivityService:statusChanged` event and corresponding event type ([#6819](https://github.com/MetaMask/core/pull/6819))

### Changed

- **BREAKING:** Update `BackendWebSocketService` to automatically manage WebSocket connections based on wallet lock state ([#6819](https://github.com/MetaMask/core/pull/6819))
  - `KeyringController:lock` and `KeyringController:unlock` are now required events in the `BackendWebSocketService` messenger
- **BREAKING**: Update `Transaction` type definition: rename `hash` field to `id` for consistency with backend API ([#6819](https://github.com/MetaMask/core/pull/6819))
- **BREAKING:** Add peer dependency on `@metamask/keyring-controller` (^23.0.0) ([#6819](https://github.com/MetaMask/core/pull/6819))
- Update `BackendWebSocketService` to simplify reconnection logic: auto-reconnect on any unexpected disconnect (not just code 1000), stay disconnected when manually disconnecting via `disconnect` ([#6819](https://github.com/MetaMask/core/pull/6819))
- Improve error handling in `BackendWebSocketService.connect()` to properly rethrow errors to callers ([#6819](https://github.com/MetaMask/core/pull/6819))
- Update `AccountActivityService` to replace API-based chain support detection with system notification-driven chain tracking ([#6819](https://github.com/MetaMask/core/pull/6819))
  - Instead of hardcoding a list of supported chains, assume that the backend has the list
  - When receiving a system notification, capture the backend-tracked status of each chain instead of assuming it is up or down
  - Flush all tracked chains as 'down' on disconnect/error (instead of using hardcoded list)
- Update documentation in `README.md` to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
  - Add "WebSocket Connection Management" section explaining connection requirements and behavior
  - Update sequence diagram to show system notification-driven chain status flow
  - Update key flow characteristics to reflect internal chain tracking mechanism
```

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -212,12 +223,12 @@ export type BackendWebSocketServiceActions =
   BackendWebSocketServiceMethodActions;
 
 export type BackendWebSocketServiceAllowedActions =
-  | AuthenticationController.AuthenticationControllerGetBearerToken
-  | BackendWebSocketServiceMethodActions;
```
mcmire, line 216: Good catch, this definitely shouldn't be here.

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -312,6 +334,9 @@ export class BackendWebSocketService {
   constructor(options: BackendWebSocketServiceOptions) {
     this.#messenger = options.messenger;
     this.#isEnabled = options.isEnabled;
+    // Default to no-op trace function to keep core platform-agnostic
+    this.#trace =
+      options.traceFn ?? (((_request, fn) => fn?.()) as TraceCallback);
```
mcmire: Why do we need a type assertion here? We should hardly ever need type assertions especially for defining default values.

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -332,41 +357,41 @@ export class BackendWebSocketService {
...
-        (state: AuthenticationController.AuthenticationControllerState) =>
-          state.isSignedIn,
-      );
-    } catch (error) {
-      throw new Error(`Authentication setup failed: ${getErrorMessage(error)}`);
-    }
+  #subscribeEvents(): void {
+    // Subscribe to authentication state changes (sign in/out)
+    this.#messenger.subscribe(
+      'AuthenticationController:stateChange',
```
mcmire, line 378: It looks like you took away the third argument. Thoughts on adding it back? As it is, this function will get called every time `AuthenticationController` changes state, so it will try to disconnect even when `isSignedIn` does not change.

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -332,41 +357,41 @@ export class BackendWebSocketService {
...
+      (state: AuthenticationController.AuthenticationControllerState) => {
+        if (state.isSignedIn) {
+          // eslint-disable-next-line no-void
+          void this.connect();
+        } else {
+          // eslint-disable-next-line no-void
+          void this.disconnect();
```
mcmire: I guess using `void` is technically valid in TypeScript, but there is a reason we have a lint rule against it: the syntax is arcane and not very self-descriptive.

There is probably another violation you are seeing here, which is `no-floating-promises`. This rule requires that all promises be awaited. I'm guessing you don't want to do this because you are in a handler so it doesn't matter anyway? If that's true, then what are your thoughts on: 1) ignoring this rule instead of `no-void`; 2) leaving a comment explaining the reason for the ignore? Something like:

``` suggestion
          // We are in a subscription handler, so we don't need to await the promise.
          // eslint-disable-next-line @typescript-eslint/no-floating-promises
          this.connect();
        } else {
          // We are in a subscription handler, so we don't need to await the promise.
          // eslint-disable-next-line @typescript-eslint/no-floating-promises
          this.disconnect();

File: packages/core-backend/src/BackendWebSocketService.ts [hunk 18]
```diff
@@ -332,41 +357,41 @@ export class BackendWebSocketService {
...
+    // Subscribe to wallet unlock event
+    this.#messenger.subscribe('KeyringController:unlock', () => {
+      // eslint-disable-next-line no-void
+      void this.connect();
+    });
+
+    // Subscribe to wallet lock event
+    this.#messenger.subscribe('KeyringController:lock', () => {
+      // eslint-disable-next-line no-void
+      void this.disconnect();
```
mcmire: Similar as above comment, what about?

``` suggestion
      // We are in a subscription handler, so we don't need to await the promise.
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      this.connect();
    });

    // Subscribe to wallet lock event
    this.#messenger.subscribe('KeyringController:lock', () => {
      // We are in a subscription handler, so we don't need to await the promise.
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      this.disconnect();

File: packages/core-backend/src/BackendWebSocketService.ts [hunk 19]
```diff
@@ -1135,24 +1241,42 @@ export class BackendWebSocketService {
...
+      },
+    );
 
-    if (shouldReconnect) {
-      log('Connection lost unexpectedly, will attempt reconnection', {
-        code: event.code,
-      });
-      this.#scheduleReconnect();
-    }
+    // For any unexpected disconnects, attempt reconnection
```
mcmire, line 1280: Were we already reconnecting upon unexpected disconnect? Is the only connection-related change in this PR the fact that `disconnect` will now disable automatic reconnect, or is there something I'm missing?

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Removed
+
+- **BREAKING**: Removed `getSupportedChains()` public method and all related API fetching logic
+- **BREAKING**: Removed hardcoded `DEFAULT_SUPPORTED_CHAINS` fallback list and cache expiration mechanism
```
mcmire: This isn't a public constant. Do we need to mention this in the changelog?

``` suggestion
```

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Enables tracing of WebSocket operations including connect, disconnect methods
+  - Trace function receives operation metadata and callback to wrap for performance monitoring
+- Updated documentation (README.md) to reflect new connection management model and chain tracking behavior ([#6819](https://github.com/MetaMask/core/pull/6819))
+  - Added "WebSocket Connection Management" section explaining connection requirements and behavior
+  - Updated sequence diagram to show system notification-driven chain status flow
+  - Updated key flow characteristics to reflect internal chain tracking mechanism
+
+### Removed
+
+- **BREAKING**: Removed `getSupportedChains()` public method and all related API fetching logic
```
mcmire: Should we specify where this was removed from? We are also missing a PR link. We also don't need to mention implementation details.

``` suggestion
- **BREAKING**: Remove `getSupportedChains` method from `AccountActivityService` ([#6819](https://github.com/MetaMask/core/pull/6819))

File: packages/core-backend/src/BackendWebSocketService.ts [hunk 22]
```diff
@@ -914,36 +955,50 @@ export class BackendWebSocketService {
...
-        this.#ws = ws;
-        this.#setState(WebSocketState.CONNECTED);
-        this.#connectedAt = Date.now();
 
-        // Reset reconnect attempts on successful connection
-        this.#reconnectAttempts = 0;
+        // Calculate connection latency
+        const connectionLatency = Date.now() - connectionStartTime;
+
+        // Trace successful connection with latency
```
mcmire, line 971: This pattern seems to be repeated three times in this file, is there a way to refactor it?

packages/core-backend/src/AccountActivityService.test.ts
```diff
@@ -612,27 +568,86 @@ describe('AccountActivityService', () => {
           );
         });
       });
+
+      it('should track chains as up and down based on system notifications', async () => {
+        await withService(async ({ messenger, mocks }) => {
+          const publishSpy = jest.spyOn(messenger, 'publish');
```
mcmire: Not sure if anyone has pointed this out yet, but I see that we are still mocking the messenger in these tests. Do we anticipate updating the tests in the future? As a reminder we can test that messenger events are published by subscribing to them ahead of time with a mock function and then asserting that the mock function is called. For example:

``` typescript
const statusChangedEventListener = jest.fn();
messenger.subscribe(
  'AccountActivityService:statusChanged',
  statusChangedEventListener,
);

// ... do the thing ...

expect(statusChangedEventListener).toHaveBeenCalledWith(...);
```

packages/core-backend/src/AccountActivityService.test.ts
```diff
@@ -612,27 +568,86 @@ describe('AccountActivityService', () => {
...
+          const systemCallback = getSystemNotificationCallback(mocks);
+
+          publishSpy.mockClear();
```
mcmire: Nit: FYI, we shouldn't need to clear mocks manually in tests. That is already being done by Jest thanks to the `resetMocks` and `restoreMocks` configuration options.

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -461,6 +471,42 @@ describe('BackendWebSocketService', () => {
...
+      const service = new BackendWebSocketService({
+        messenger,
+        url: 'ws://test.example.com',
+        // timeout, reconnectDelay, maxReconnectDelay, requestTimeout not provided - will use defaults
+      });
+
+      expect(service.getConnectionInfo().url).toBe('ws://test.example.com');
+      expect(service).toBeInstanceOf(BackendWebSocketService);
+
+      service.destroy();
```
mcmire: Can we use `withService` so that `destroy` gets called automatically? 

And I see that there a comment saying that the default should be X, Y, and Z but shouldn't we test that this is so?

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -461,6 +471,42 @@ describe('BackendWebSocketService', () => {
...
+    });
+
+    it('should accept traceFn option for performance tracing', async () => {
```
mcmire: What are we testing here?

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -486,53 +532,22 @@ describe('BackendWebSocketService', () => {
...
-          connectPromise.catch(() => {
-            // Expected rejection - no action needed
-          });
+          const connectPromise = service.connect().catch((error) => error);
 
           await completeAsyncOperations(TEST_CONSTANTS.TIMEOUT_MS + 50);
 
-          await expect(connectPromise).rejects.toThrow(
+          const error = await connectPromise;
+          expect(error.message).toBe(
```
mcmire, line 519: Is there a reason why we are not using `rejects.toThrow`? Seems like it'd be simpler.

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -620,232 +639,176 @@ describe('BackendWebSocketService', () => {
...
+            'WebSocket connection error',
           );
+          expect(service.getConnectionInfo().state).toBe(WebSocketState.ERROR);
         },
       );
     });
 
-    it('should handle WebSocket error events during runtime without immediate state change', async () => {
-      await withService(async ({ service, getMockWebSocket }) => {
+    it('should return immediately when already connected', async () => {
```
mcmire: What does "return immediately" mean? Isn't that what functions do? It seems like what we are really testing here is that the state gets updated after connecting. Unless there is an opposite case that this test is making an implicit assumption about.

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -620,232 +639,176 @@ describe('BackendWebSocketService', () => {
...
-        expect(service.name).toBe('BackendWebSocketService');
-
-        const connectionInfo = service.getConnectionInfo();
-        expect(connectionInfo.state).toBe(WebSocketState.DISCONNECTED);
-        expect(connectionInfo.url).toBe('ws://localhost:8080');
-      });
-    });
-
-    it('should disconnect WebSocket connection and set state to DISCONNECTED when connected', async () => {
+    it('should return immediately when disconnect called on disconnected service', async () => {
```
mcmire: Similar comment as above, this seems to misstate what is being tested.

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -620,232 +639,176 @@ describe('BackendWebSocketService', () => {
...
+          expect(service.getConnectionInfo().state).toBe(
+            WebSocketState.DISCONNECTED,
           );
-          expect(service.getConnectionInfo().state).toBe(WebSocketState.ERROR);
         },
       );
     });
 
-    it('should handle WebSocket close events during connection establishment by setting state to ERROR', async () => {
+    it('should calculate connection duration as 0 when close happens before connectedAt is set', async () => {
```
mcmire: Nit: The fact that duration is 0 seems to be an implementation detail, is there a way we can reword this test that describes the visible behavior?

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -957,22 +922,55 @@ describe('BackendWebSocketService', () => {
...
+            requestId: 'unsub-req-id',
+            successful: ['test-channel'],
+            failed: [],
+          },
+        });
+
+        await unsubscribePromise;
+
+        // Subscription should be cleaned up
+        expect(service.channelHasSubscription('test-channel')).toBe(false);
```
mcmire, line 924: Why are these lines here. Are they necessary to exercise the logic referred to in the description of the test?

``` suggestion
```

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -1273,167 +1187,172 @@ describe('BackendWebSocketService', () => {
...
+
+          // Add different channel callback
+          service.addChannelCallback({
+            channelName: 'different-channel',
+            callback: jest.fn(),
+          });
+
+          expect(service.getChannelCallbacks()).toHaveLength(2);
+
+          // Remove callback - should return true
```
mcmire, line 1172: Is this comment saying anything that the code isn't?

``` suggestion
```

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -1273,167 +1187,172 @@ describe('BackendWebSocketService', () => {
...
+          expect(service.removeChannelCallback('test-channel-duplicate')).toBe(
+            true,
+          );
+          expect(service.getChannelCallbacks()).toHaveLength(1);
+
+          // Try to remove non-existent callback - should return false
```
mcmire, line 1178:
``` suggestion
```

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -1475,92 +1394,26 @@ describe('BackendWebSocketService', () => {
...
-              messenger,
-              url: 'ws://test',
-            });
-          }).toThrow(
-            'Authentication setup failed: AuthenticationController not available',
-          );
-        },
-      );
+        expect(connectSpy).toHaveBeenCalled();
+        connectSpy.mockRestore();
```
mcmire: Stated elsewhere, but we don't need to unregister mock functions, as they will be automatically reset/restored after each test.

``` suggestion
```

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: If I put (((_request, fn) => fn?.()) as any) I got a linter issue

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: > Were we already reconnecting upon unexpected disconnect?
Yes we were but if the server disconnected with the event code 1000 then no. I realized it by having the WebSocket open for more than 1 hour. The server disconnected, but didn't reconnect. Which is undesirable because we want MetaMask to reconnect back. I have tested it and it works well now.

> disconnect will now disable automatic reconnect
yes that is the only case where we dont try to not re-connect, i.e. only when we close the app.

packages/core-backend/src/AccountActivityService.test.ts
Kriys94: By mocking publish, I'm kind of testing the output of handleSystemNotification. I wonder how I can test that? Your example is mocking the callback of a subscribe; here, I want to test the output.

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Remove test

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: I don't manage to make the test work with rejects.toThrow?

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: This is mainly testing and covering the fact that if you call .connect() multiple times while already CONNECTED, then nothing is executed, it covers L445 in BackendWebsocketService.ts

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Because below you have // Try to remove non-existent callback - should return false. WHich explain the 2 cases

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -858,7 +856,7 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     // Check if token exists in allIgnoredTokens
     if (
       this.#allIgnoredTokens?.[chainId]?.[account.toLowerCase()]?.some(
-        (addr) => addr.toLowerCase() === normalizedAddress,
+        (token) => token === tokenAddress,
```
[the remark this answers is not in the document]
Kriys94: Ack, we normalized in the caller of this function

packages/core-backend/CHANGELOG.md
Kriys94: Looks better. Thanks

packages/core-backend/CHANGELOG.md
Kriys94: DOne

packages/core-backend/CHANGELOG.md
Kriys94: Done

packages/core-backend/src/AccountActivityService.test.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done

Kriys94: I changed the description

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: I find this comment interesting to read

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: DOne

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: Wierd, I dont know it was not pushed. I put it back. THanks again

packages/assets-controllers/src/TokenBalancesController.ts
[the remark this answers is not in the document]
Kriys94: Ack. The input is in checksum and the key is checksum too

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: To me, that is quite straightforward, the callback is already abstracting many logic, I think it would be over-engineering to add another layer of abstraction

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -852,13 +897,15 @@ export class BackendWebSocketService {
     const subscription = {
       subscriptionId,
       channels: [...channels],
+      channelType,
```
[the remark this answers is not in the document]
Kriys94: Ack, this is already managed in handleServerResponse

packages/core-backend/src/AccountActivityService.test.ts
cryptodev-2s: You could try something like the suggestion below, it would let you test the output directly.
```typescript
      it('should track chains as up and down based on system notifications', async () => {
        await withService(async ({ messenger, mocks }) => {
          const statusChangedEventListener = jest.fn();
          messenger.subscribe(
            'AccountActivityService:statusChanged',
            statusChangedEventListener,
          );
          const systemCallback = getSystemNotificationCallback(mocks);

          // Simulate chains coming up
          const timestamp1 = 1760344704595;
          systemCallback({
            event: 'system-notification',
            channel: 'system-notifications.v1.account-activity.v1',
            data: {
              chainIds: ['eip155:1', 'eip155:137'],
              status: 'up',
            },
            timestamp: timestamp1,
          });

          expect(statusChangedEventListener).toHaveBeenCalledWith({
            chainIds: ['eip155:1', 'eip155:137'],
            status: 'up',
            timestamp: timestamp1,
          });

          // Simulate one chain going down
          const timestamp2 = 1760344704696;
          systemCallback({
            event: 'system-notification',
            channel: 'system-notifications.v1.account-activity.v1',
            data: {
              chainIds: ['eip155:137'],
              status: 'down',
            },
            timestamp: timestamp2,
          });

          expect(statusChangedEventListener).toHaveBeenCalledWith({
            chainIds: ['eip155:137'],
            status: 'down',
            timestamp: timestamp2,
          });
        });
      });
    });

File: packages/core-backend/src/BackendWebSocketService.test.ts [hunk 37]
```diff
@@ -486,53 +518,22 @@ describe('BackendWebSocketService', () => {
...
-        );
-        expect(spies.publish).toHaveBeenNthCalledWith(
-          2,
-          'BackendWebSocketService:connectionStateChanged',
-          expect.objectContaining({ state: WebSocketState.CONNECTED }),
-        );
-      });
-    });
-
     it('should handle connection timeout by rejecting with timeout error and setting state to DISCONNECTED', async () => {
```
cryptodev-2s: Nit: I think there’s a typo in the state name here.
```suggestion
    it('should handle connection timeout by rejecting with timeout error and setting state to ERROR', async () => {

packages/core-backend/src/AccountActivityService.test.ts
Kriys94: I see I updated it. Thanks

Kriys94: No more jest.spyOn(messenger now ;) Thanks guys

packages/core-backend/src/BackendWebSocketService.test.ts
mcmire: Oh, interesting, the promise returns an error, I see. Never mind :)

packages/core-backend/src/BackendWebSocketService.test.ts
mcmire: Oh I see, thanks for clarifying.

packages/core-backend/src/BackendWebSocketService.ts
mcmire: Okay. Just thought that if we wanted to change how this works we could change it one place rather than three, but if it's not a big deal no problem.

packages/core-backend/src/BackendWebSocketService.ts
mcmire: Well, that's an indication that it ought to be written a different way, no?

But not a big deal, we can address this in a future PR.

packages/core-backend/src/BackendWebSocketService.ts
mcmire: Nice, thanks for clarifying.

PR #6820, opened 2025-10-12
feat: transaction pay controller

matthewwalsh0:
## Explanation

Add the initial version of the `TransactionPayController`.

This intends to migrate the majority of the MetaMask Pay functionality from the mobile client, so it can be used without duplication in the extension.

For an architecture overview, see [ARCHITECTURE.md](https://github.com/MetaMask/core/blob/feat/transaction-pay-controller/packages/transaction-pay-controller/ARCHITECTURE.md).

Includes:

- Required token generation from ERC-20 token transfers and required gas fees.
- Alternate pay strategies using MetaMask bridge infrastructure or Relay.
- Publish hook for `TransactionController` to submit and wait for quotes.
- Token and balance generation using asset controllers.
- Payment token selection using `updatePaymentToken` action / method.
- Automatic quote and total generation on transaction creation and update.

## Checklist

---

> [!NOTE]
> Adds @metamask/transaction-pay-controller to power MetaMask Pay with bridge/relay strategies, automatic required-token detection, quote orchestration, and a TransactionController publish hook.
> 
> - **New Package: `@metamask/transaction-pay-controller`**
>   - Controller to fund transactions by supplying required ERC-20/native tokens across chains.
>   - Detects required tokens (ERC-20 transfers via `0xa9059cbb`, gas fees), selects payment token, computes source amounts, fetches quotes, and aggregates totals.
>   - **Strategies**:
>     - `BridgeStrategy`: fetches via `BridgeController`, submits via `BridgeStatusController`, supports batch txs and configurable refresh intervals.
>     - `RelayStrategy`: fetches from Relay API, submits txs directly; supports Hyperliquid/Polygon-native normalization and skip-original-tx flow.
>   - **Publish Hook**: `TransactionPayPublishHook` executes strategy quotes on publish and awaits completion.
>   - **Utilities**: required-token parsing, gas cost calc, quote refresh queue, totals calc, token/rate/balance helpers.
>   - Comprehensive tests, TypeDoc, Jest config, and exports.
> - **Monorepo Integration**:
>   - Update `README.md` (package list + dependency graph), `CODEOWNERS`, `teams.json`.
>   - Add tsconfig references, yarn.lock entries, and changelog/license/metadata files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 83d69770efd9ae804d73731535d40f7cee2a1dee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/src/TransactionPayController.ts
+++ b/packages/transaction-pay-controller/src/TransactionPayController.ts
@@ -1,7 +1,6 @@
 import type { StateMetadata } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
 import type { TransactionMeta } from '@metamask/transaction-controller';
-import type { Hex } from '@metamask/utils';
 import type { Draft } from 'immer';
 import { noop } from 'lodash';
 


OGPoyraz (dismissed):

FrederikBolding (dismissed):
CODEOWNERS LGTM

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -0,0 +1,143 @@
...
+    pollTransactionChanges(
+      messenger,
+      this.#updateTransactionData.bind(this),
+      this.#removeTransactionData.bind(this),
+    );
+
+    queueRefreshQuotes(messenger, this.#updateTransactionData.bind(this));
+  }
+
+  updatePaymentToken({
```
Mrtenz: Nit: Should this be exposed through the messenger?

matthewwalsh0: 😄 It should indeed, but we're currently using the controller instance directly in mobile, hence not yet required.

Mrtenz: We want to avoid direct references to controllers where possible. I'd prefer adding this to the messenger now instead of refactoring this later on 😅

OGPoyraz (dismissed):

packages/transaction-pay-controller/ARCHITECTURE.md
```diff
@@ -0,0 +1,70 @@
...
+Quotes are retrieved from the [Relay API](https://docs.relay.link/what-is-relay), then submitted via a transaction directly to the `TransactionController`.
+
+The resulting transaction deposits the necessary funds (on the source network), then a Relayer on the target chain immediately transfers the necessary funds and optionally executes any requested call data.
+
+## Lifecycle
+
+The high level interaction with the `TransactionPayController` is as follows:
+
+1. Client assigns the `TransactionPayPublishHook` as a publish hook in the `TransactionController` during initialisation.
+2. Controller subscribes to `TranasctionController` state changes during initialisation.
```
pedronfigueiredo: Typo `TransactionController`

packages/transaction-pay-controller/src/TransactionPayController.ts
matthewwalsh0: Done. I much prefer the messenger but there's still lots of references to the controller instances in mobile, we'll pivot to `Engine.controllerMessenger.call` if that's the new standard.

packages/transaction-pay-controller/ARCHITECTURE.md
matthewwalsh0: Done.

FrederikBolding (approved):

packages/transaction-pay-controller/src/TransactionPayController.ts
Mrtenz: In an ideal world we wouldn't need to `Engine.controllerMessenger` and all actions are called from other controllers, but I understand that may not be possible either. To be clear, not saying you need to refactor this right away, but at least now we have the option to. Saw this a lot when working on the modular init refactors, where we were calling controller methods through direct references because the method wasn't exposed through the messenger, resulting in cases like

```ts
new Controller({
  someMethod: () => otherController.someMethod(),
});
```

Just trying to avoid cases like this.

matthewwalsh0: Oh absolutely, we 100% want to avoid the legacy pattern of controller callbacks using other controller instances.

But I think logically there will alway have to be scenarios where the UIs will need to interact with the controllers "directly" (via the messenger), since we need to initiate processes and actions from user input, like adding transactions, creating Perps trades, deleting a network etc.

packages/transaction-pay-controller/src/utils/required-tokens.ts
```diff
@@ -0,0 +1,280 @@
...
+ * @param decimals - Number of decimals for the token.
+ * @param fiatRates - Fiat rates for the token.
+ * @returns Object containing amount in fiat, human-readable, raw, and USD formats.
+ */
+function calculateAmounts(
+  amountRawInput: BigNumber.Value,
+  decimals: number,
+  fiatRates: FiatRates,
+) {
+  const amountRawValue = new BigNumber(amountRawInput);
```
pedronfigueiredo, line 219: do we need to pass 16 here to denote a hexadecimal value?

packages/transaction-pay-controller/src/utils/source-amounts.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const { paymentToken, tokens } = transactionData;
+
+  if (!tokens.length || !paymentToken) {
+    return;
+  }
+
+  const sourceAmounts = tokens
+    .map((t) => calculateSourceAmount(paymentToken, t, messenger))
+    .filter(Boolean) as TransactionPaySourceAmount[];
```
pedronfigueiredo, line 42: This filter changes the length of the sourceAmounts array compared to the initial tokens array, this could cause issues on `quotes.ts` when relying on tokens[i] to select a certain token.

packages/transaction-pay-controller/src/utils/required-tokens.ts
matthewwalsh0: If the value starts with `0x`, it should be automatic.

packages/transaction-pay-controller/src/utils/source-amounts.ts
matthewwalsh0: Good spot, this is probably okay for now since it's the second gas fee token that is most likely to be removed if the user has balance, but safer to lookup by the token address on the source amount.

packages/transaction-pay-controller/src/TransactionPayController.ts
Gudahtt: We're working on a proposal now for setting new conventions for calling the messenger from the UI.

Using `Engine.controllerMessenger` in the meantime sounds ideal to me

PR #6829, opened 2025-10-14
fix: fix addTokens call with an already added tokens to state

sahar-fehri:
## Explanation

When we first call the function `addTokens` with an array of tokens to import, and then call it again with the exact same token, we notice that `state.allTokens` is updated with duplicate entries of the same token.

`state.allTokens` are stored in checksum format.
When initially state already has USDC token for a user address; and a call to `addTokens` is made with a lowercase USDC address. The newTokensMap in this line https://github.com/MetaMask/core/blob/9f480c8ac7c88bc434c6bcd80dc7f0c58d0c1698/packages/assets-controllers/src/TokensController.ts#L514 is computed without checking if the value already exists in state.

The fix converts the tokensToImport received in the call to the checksum format.

## Checklist

---

> [!NOTE]
> Normalize token addresses to checksum in `TokensController.addTokens` to dedupe and avoid duplicate entries when the same token is added multiple times.
> 
> - **Assets Controllers**:
>   - **TokensController**: Normalize `tokensToImport` addresses to checksum and use checksummed list for deduping/updates in `addTokens`, preventing duplicate entries in `state.allTokens`.
> - **Tests**:
>   - Add unit test ensuring duplicate tokens are not re-added when the same token is imported twice (case-insensitive).
> - **Changelog**:
>   - Add Unreleased fix entry documenting duplicate token prevention in `state.allTokens`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6df8521e7b169c6e4a09dbab011370d0ace9af95. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix duplicate token in state.allTokens after calling addTokens function with the same token twice ([#6829](https://github.com/MetaMask/core/pull/6829)
+
 ## [80.0.0]
 
 ### Added


packages/assets-controllers/src/TokensController.test.ts
```diff
@@ -1878,6 +1878,71 @@ describe('TokensController', () => {
         },
       );
     });
+
+    it('should not add duplicate tokens to state', async () => {
```
sahar-fehri, line 1882: This test fails on main and `state.allTokens` ends up with duplicate entries of the same token

PR #6847, opened 2025-10-15
fix: shield coverage result polling

lwin-kyaw:
## Explanation

This PR includes ~
- Updated internal coverage result polling and log logic.
   - Added cancellation logic to the polling.
   - Updated implementation of timeout.
   - Cancel any pending requests before starting new polling or logging.
- Updated TransactionMeta comparison in `TransactionController:stateChange` subscriber to avoid triggering multiple check coverage result unnecessarily.
- Removed `Personal Sign` check from the check signature coverage result.

## Checklist

---

> [!NOTE]
> Replaces manual shield coverage polling with a Cockatiel/controller-utils policy that supports retries and cancellation, updates error handling, tests, changelog, and adds required deps.
> 
> - **Backend**:
>   - **Polling refactor**: Replace manual timeout loop in `ShieldRemoteBackend.#getCoverageResult` with `PollingWithCockatielPolicy`, using per-request IDs (`txMeta.id`/`signatureRequest.id`).
>   - **Cancellation**: Abort pending polls before `logSignature`/`logTransaction` via `#pollingPolicy.abortPendingRequest`.
>   - **Retry/backoff**: Compute retries/backoff with `ConstantBackoff` and `DEFAULT_MAX_RETRIES` (`computePollingIntervalAndRetryCount`).
>   - **Error handling**: Throw `HttpError` with parsed body message for non-200 responses.
> - **New Utility**:
>   - Add `src/polling-with-policy.ts` implementing Cockatiel-based policy with retry filtering (retry 4xx except 400) and request tracking.
> - **Tests**:
>   - Update `backend.test.ts` expectations for new timeout/error messages.
>   - Add `polling-with-policy.test.ts` covering retry, max-retry, cancellation, and concurrent requests.
>   - Add `tests/utils.delay` helper.
> - **Dependencies**:
>   - Add `@metamask/controller-utils` and `cockatiel`.
> - **Docs**:
>   - Update `CHANGELOG.md` noting optimized coverage-result polling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e13e024d2d62203968a51dcb78bc10f0d683ad04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shield-controller/CHANGELOG.md
+++ b/packages/shield-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed and optimized shield-coverage-result polling with Cockatiel Policy from Controller-utils. ([#6847](https://github.com/MetaMask/core/pull/6847))
+
 ## [1.0.0]
 
 ### Added
@@ -21,10 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING:** Bump `@metamask/signature-controller` from `^34.0.0` to `^35.0.0` ([#6962](https://github.com/MetaMask/core/pull/6962))
 - **BREAKING:** Bump `@metamask/transaction-controller` from `^60.0.0` to `^61.0.0` ([#6962](https://github.com/MetaMask/core/pull/6962))
 - Bump `@metamask/base-controller` from `^8.4.2` to `^9.0.0` ([#6962](https://github.com/MetaMask/core/pull/6962))
-
-### Fixed
-
-- Fixed and optimized shield-coverage-result polling with Cockatiel Policy from Controller-utils. ([#6847](https://github.com/MetaMask/core/pull/6847))
 
 ## [0.4.0]
 


packages/shield-controller/src/ShieldController.ts
```diff
@@ -443,4 +447,61 @@ export class ShieldController extends BaseController<
...
+        previousSimulationData?.tokenBalanceChanges.length ||
+      simulationData?.tokenBalanceChanges.some(
+        (tokenBalanceChange, index) =>
+          tokenBalanceChange.difference !==
+          previousSimulationData?.tokenBalanceChanges[index].difference,
+      )
+    ) {
+      return true;
+    }
+
```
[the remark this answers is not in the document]
lwin-kyaw: `tokenBalanceChanges` is a required filed in `SimulationData`.

lwin-kyaw:
@metamaskbot publish-preview

lwin-kyaw:
@metamaskbot publish-preview

packages/shield-controller/src/ShieldController.ts
```diff
@@ -268,14 +268,18 @@ export class ShieldController extends BaseController<
     );
     for (const transaction of transactions) {
       const previousTransaction = previousTransactionsById.get(transaction.id);
+      const simulationDataChanged = this.#compareTransactionSimulationData(
+        previousTransaction?.simulationData,
+        transaction.simulationData,
+      );
```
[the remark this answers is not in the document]
matthiasgeihs: Bot must have had a bad day. That's not the case, as far as I can see.

packages/shield-controller/src/backend.ts
```diff
@@ -189,44 +197,51 @@ export class ShieldRemoteBackend implements ShieldBackend {
       pollInterval?: number;
     },
   ): Promise<GetCoverageResultResponse> {
+    if (this.#abortController && !this.#abortController.signal.aborted) {
```
matthiasgeihs: The way you integrated the abort functionality is problematic.

You use one abort controller for all requests.

There may be situations where you have result polling for different transactions and signatures at the same time. If one log request comes in, it cancels the polling from all other unrelated transactions and signatures. That's most definitely not what we want.

matthiasgeihs (changes requested):
Please note my comment on using a singular abort controller for unrelated result requests. I think it's highly problematic. You only want to cancel polling for that specific coverageId, not for all coverage requests.

packages/shield-controller/src/backend.ts
```diff
@@ -189,44 +197,51 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
+      throw new Error('unexpected error');
+    };
 
-      poll().then(resolve).catch(reject);
-    });
+    return await withTimeoutAndCancellation<GetCoverageResultResponse>(
+      poll,
+      timeout,
+      abortController,
+    );
```
matthiasgeihs: Instead of rewriting the entire logic, you could have just used the existing variable `timeoutReached` (maybe rename it) to cancel the polling.

packages/shield-controller/src/backend.ts
```diff
@@ -189,44 +197,51 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
-          const startTime = Date.now();
-          const res = await this.#fetch(configs.coverageResultUrl, {
-            method: 'POST',
-            headers,
-            body: JSON.stringify(reqBody),
-          });
-          if (res.status === 200) {
-            return (await res.json()) as GetCoverageResultResponse;
-          }
+    let shouldContinuePolling = true;
```
matthiasgeihs: This should be inside the `poll` function!?

packages/shield-controller/src/backend.ts
matthiasgeihs: We should avoid refactoring existing logic, if not necessary. I suspect the logic might have been unclear and that's why you have rewritten it. Instead you could also check with the author to clarify the existing code.

lwin-kyaw: Restore the existing logic and appended the cancellation logic in this commit, https://github.com/MetaMask/core/pull/6847/commits/9225267bd45d850213dc5bbca15e1ee1d1603fc0

packages/shield-controller/src/backend.ts
lwin-kyaw: I have updated the design with `Map<transactionId | signatureId, AbortController` to handle multiple requests.
https://github.com/MetaMask/core/pull/6847/commits/9225267bd45d850213dc5bbca15e1ee1d1603fc0

packages/shield-controller/src/backend.ts
```diff
@@ -193,15 +210,26 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
+        this.#removeAbortHandler(configs.requestId, abortHandler);
+        reject(new Error('Coverage result polling cancelled'));
+      };
+      abortController.signal.addEventListener('abort', abortHandler);
+
       setTimeout(() => {
         timeoutReached = true;
+        this.#removeAbortHandler(configs.requestId, abortHandler);
         reject(new Error('Timeout waiting for coverage result'));
       }, timeout);
```
matthiasgeihs: We could remove some redundancy here. Would it be possible to have the cleanup and rejection in just one place? Maybe you could simplify call `abortController.abort()` when the timeout triggers?
```ts
        timeoutReached = true;
        this.#removeAbortHandler(configs.requestId, abortHandler);
        reject(new Error(msg));
```

matthiasgeihs: Since you are already using an AbortSignal above, you could also use an `AbortSignal.timeout` here (instead of `setTimeout`).

packages/shield-controller/src/backend.ts
```diff
@@ -214,8 +242,10 @@ export class ShieldRemoteBackend implements ShieldBackend {
             method: 'POST',
             headers,
             body: JSON.stringify(reqBody),
+            signal: abortController.signal,
```
matthiasgeihs: This does not respect the timeout.
You can have a look at my draft, how the two can be combined: https://github.com/MetaMask/core/blob/40ff4f5c13ae6e117e714df8669e9f57f81f920a/packages/shield-controller/src/backend.ts#L224

Similarly, the `while` condition does not respect the abortController.signal.

packages/shield-controller/src/backend.ts
```diff
@@ -236,6 +266,30 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
+      abortController.abort();
+      this.#removeAbortHandler(id);
+    }
+  }
+
+  #removeAbortHandler(requestId: string, abortHandler?: () => void) {
+    const abortController = this.#abortControllerMap.get(requestId);
+    if (abortController) {
+      if (abortHandler) {
+        abortController.signal.removeEventListener('abort', abortHandler);
```
matthiasgeihs: I'm wondering: Is it necessary to remove the handler, if we drop the whole controller anyways? (Might simplify a few things.)

packages/shield-controller/tests/mocks/abortController.ts
matthiasgeihs, line 1: I don't see any use of mocking the AbortController. On the downside, it bloats the codebase and can also cause some issues (e.g., it currently doesn't work with `AbortSignal.any`). We should remove this mock.

packages/shield-controller/src/backend.test.ts
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable jest/no-conditional-in-test */
```
matthiasgeihs: If we add lint ignores, we should always also add a comment, why these are necessary.

In this case, it looks like this has been added because the `fetchMock` implementations use `if then else`. Would it be possible to rewrite these mocks without using conditionals?

packages/shield-controller/src/backend.test.ts
```diff
@@ -45,6 +48,25 @@ function setup({
...
+  afterAll(() => {
+    // Restore original AbortController
+    globalThis.AbortController = originalAbortController;
+  });
+
+  afterEach(() => {
+    // Clean up mocks after each test
+    jest.clearAllMocks();
+  });
+
```
matthiasgeihs: This should be removed. See my comment on `AbortControllerMock`.

packages/shield-controller/src/backend.test.ts
```diff
@@ -220,6 +242,69 @@ describe('ShieldRemoteBackend', () => {
...
+      const { backend, fetchMock } = setup({
+        getCoverageResultTimeout: 10000,
+        getCoverageResultPollInterval: 100,
+      });
+
+      const signatureRequest = generateMockSignatureRequest();
+      const coverageId = 'coverageId';
+
+      // Setup fetch to handle init, result polling, and log requests
+      fetchMock.mockImplementation(
```
matthiasgeihs: There are several similar mock implementations. We should be able to reduce some redundancy here. Could we merge these into one mock implementation (potentially in a separate file)?

packages/shield-controller/src/backend.test.ts
```diff
@@ -250,5 +335,295 @@ describe('ShieldRemoteBackend', () => {
...
+      const { backend, fetchMock } = setup({
+        getCoverageResultTimeout: 10000,
+        getCoverageResultPollInterval: 100,
+      });
+
+      const txMeta = generateMockTxMeta();
+      const coverageId = 'coverageId';
+
+      // Setup fetch to handle init, result polling, and log requests
+      fetchMock.mockImplementation(
```
matthiasgeihs: This mock implementation is in parts redundant to the other fetchMock implementations.

packages/shield-controller/src/backend.test.ts
```diff
@@ -250,5 +335,295 @@ describe('ShieldRemoteBackend', () => {
...
+      const { backend, fetchMock } = setup({
+        getCoverageResultTimeout: 10000, // Long timeout to avoid timeout during test
+        getCoverageResultPollInterval: 100,
+      });
+
+      const signatureRequest = generateMockSignatureRequest();
+      const coverageId = 'coverageId';
+
+      // Mock `/init` and `/result` responses.
+      fetchMock.mockImplementation(
```
matthiasgeihs: This mock implementation is in parts redundant to the other fetchMock implementations.

packages/shield-controller/src/backend.test.ts
```diff
@@ -250,5 +335,295 @@ describe('ShieldRemoteBackend', () => {
...
+      const { backend, fetchMock } = setup({
+        getCoverageResultTimeout: 200, // Short timeout to simulate the coverage result polling timeout
+        getCoverageResultPollInterval: 50,
+      });
+
+      const signatureRequest = generateMockSignatureRequest();
+      const coverageId = 'coverageId';
+
+      // Mock fetch responses
+      fetchMock.mockImplementation(
```
matthiasgeihs: This mock implementation is in parts redundant to the other fetchMock implementations.

packages/shield-controller/src/backend.test.ts
```diff
@@ -250,5 +335,295 @@ describe('ShieldRemoteBackend', () => {
...
+    it('should handle fetch request cancellation properly', async () => {
+      const { backend, fetchMock } = setup({
+        getCoverageResultTimeout: 200, // Short timeout
+        getCoverageResultPollInterval: 50,
+      });
+
+      const signatureRequest = generateMockSignatureRequest();
+      const coverageId = 'test';
+
+      fetchMock.mockImplementation(
```
matthiasgeihs: This mock implementation is in parts redundant to the other fetchMock implementations.

packages/shield-controller/src/backend.ts
```diff
@@ -236,6 +266,30 @@ export class ShieldRemoteBackend implements ShieldBackend {
       Authorization: `Bearer ${accessToken}`,
     };
   }
+
+  #cancelPreviousPendingRequests(id: string) {
```
matthiasgeihs:
```suggestion
  #cancelPendingRequests(id: string) {
```
If it's "pending" request, it's always a "previous" request. So we can omit `previous`.

packages/shield-controller/src/backend.ts
```diff
@@ -193,15 +210,26 @@ export class ShieldRemoteBackend implements ShieldBackend {
       coverageId,
     };
 
+    const abortController = this.#assignNewAbortController(configs.requestId);
```
matthiasgeihs: Is there a reason why you create this here and not within the Promise?

(We could probably move all things from the outer context into the Promise. Might be cleaner overall.)

packages/shield-controller/src/backend.ts
```diff
@@ -193,15 +210,26 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
     const pollInterval =
       configs?.pollInterval ?? this.#getCoverageResultPollInterval;
 
     const headers = await this.#createHeaders();
     return await new Promise((resolve, reject) => {
       let timeoutReached = false;
+
+      const abortHandler = () => {
+        timeoutReached = true;
+        this.#removeAbortHandler(configs.requestId, abortHandler);
```
matthiasgeihs: Unrelated to here, but could we change `configs` to `config`? Configuration is usually used in singular. Alternatively, you could use `opts` or `options`.

packages/shield-controller/src/backend.ts
lwin-kyaw: Please advise me if I'm wrong.

One problem with using `AbortSignal.timeout` is that, we can't manually cancel the timeout.
Let's say when another request comes in, we need to cancel the pending requests right?
Cancelling request needs two things here 
- cancel the abort controller (to stop polling)
- cancel the timeout 

For the timeout cancel, if we are using JS timers, we can simply cancel the timers. However, there's no way to cancel the `AbortSignal.timeout`, which will eventually throw which would be problematic, too.

packages/shield-controller/src/backend.ts
lwin-kyaw: Yes, this doesn't respect the timeout. But this cancel the ongoing network request, when the abortController is aborted (aka new request is triggered).

To be frank, this (AbortController with network request) is a very common pattern (especially in frontend/react). 
Ref: https://stackoverflow.com/a/61056569

packages/shield-controller/tests/mocks/abortController.ts
lwin-kyaw: My bad on not removing the MockAbortController when it's no longer needed anymore.
I will take note of that, and will take a closer and more careful look in the future. 🙏

packages/shield-controller/src/backend.ts
lwin-kyaw: > Similarly, the while condition does not respect the abortController.signal.

When `abortController.signal` is aborted, we set `timeoutReached=true` where break the `while` loop.

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,163 @@
...
+   *
+   * @param requestId - The ID of the request to handle the timeout for.
+   */
+  #handleRequestTimeout(requestId: string) {
+    const requestEntry = this.#cleanUpRequestEntryIfExists(requestId);
+    if (requestEntry) {
+      // Abort the request, this will also trigger the abort handler (hence, handleRequestAbort will be called)
+      requestEntry.abortController.abort();
+    }
+  }
```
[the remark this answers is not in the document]
lwin-kyaw: The removed entry is returned from `cleanUpRequestEntryIfExists` and the `AbortController` instance is still referenced.
This referenced instance is later aborted in line 123 and event listener is removed on aborted.

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.ts
lwin-kyaw: All these logics were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.ts
lwin-kyaw: All these logics were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.ts
lwin-kyaw: All these logics were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.test.ts
lwin-kyaw: All these new tests were refactored and removed here, https://github.com/MetaMask/core/pull/6847/commits/565c42c45d82d8021565d57a3c6dc99994795665

packages/shield-controller/src/backend.ts
lwin-kyaw: I think we should clean up (`removeEventListener`) whenever we do `addEventListener`.
Eventually, the event listeners will be clean up when the event target (AbortController) is garbage collected. But still memory leaks can if AbortController is not GC, due to some unknown reasons :)

packages/shield-controller/src/backend.ts
```diff
@@ -192,40 +208,33 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
-        // will already be rejected by now.
-        throw new Error('unexpected error');
+        throw new Error(`Failed to get coverage result: ${res.status}`);
       };
 
-      poll().then(resolve).catch(reject);
+      this.#pollingWithTimeout
+        .pollRequest(config.requestId, requestCoverageFn, pollingOptions)
+        .then(resolve)
+        .catch(reject);
```
matthiasgeihs: The refactor makes everything much more readable. 👍

Now you can even drop the wrapping `new Promise` and just write `return await this.#pollingWithTimeout.pollRequest(...);`.

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,181 @@
...
+        this.#cleanUpOnFinished(requestId);
+        return result;
+      } catch {
+        // polling failed due to timeout or cancelled,
+        // we need to clean up the request entry and throw the error
+        if (this.#isAbortedAndNotTimeoutReason(abortController.signal)) {
+          throw new Error(`${pollingOptions.fnName}: Request cancelled`);
+        }
+      }
+      await this.#delay(pollInterval);
```
matthiasgeihs: Note that this is not canceled by the abort signal. So if we have a long polling interval, then it can take some time until this returns.
Previously this was achieved by rejecting immediately when the timeout occurred.

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,181 @@
...
+    this.abortPendingRequests(requestId);
+
+    // insert the request entry for the next polling cycle
+    const { abortController } = this.#insertRequestEntry(requestId, timeout);
+
+    while (!abortController.signal.aborted) {
+      try {
+        const result = await requestFn(abortController.signal);
+        // polling success, we just need to clean up the request entry and return the result
+        this.#cleanUpOnFinished(requestId);
```
matthiasgeihs: Do you have to call this cleanup function also in the `catch` case?

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,181 @@
...
+    },
+  ) {
+    const timeout = pollingOptions.timeout ?? this.#timeout;
+    const pollInterval = pollingOptions.pollInterval ?? this.#pollInterval;
+
+    // clean up the request entry if it exists
+    this.abortPendingRequests(requestId);
+
+    // insert the request entry for the next polling cycle
+    const { abortController } = this.#insertRequestEntry(requestId, timeout);
```
matthiasgeihs: The whole functionality of this is a bit difficult to grasp as it's spread across the whole file and different handlers (especially the different cleanup handlers are confusing).

Would it be possible to streamline the logic a bit:
1. Create 1 abort controller. If this aborts, the polling aborts and the abort handling is cleaned up.
2. Attached the timeout event to the abort controller. (i.e., on timeout, call `controller.abort()`)

packages/shield-controller/src/polling-with-timeout-abort.ts
lwin-kyaw: Yes, the abort doesn't cancelled the delay between polling. 
But I don't think this is a big issue, a small delay won't hurt so much. After the delay, the polling will break.

packages/shield-controller/src/polling-with-timeout-abort.ts
lwin-kyaw: Not really, coz `cleanUpOnFinished` will be called in the `AbortHandler` and `Timeout` handle, where we only want to clean up. 
We don't want to clean in case of normal errors, we want the polling to continue.

packages/shield-controller/src/polling-with-timeout-abort.ts
lwin-kyaw: Addressed it here, https://github.com/MetaMask/core/pull/6847/commits/27a1acb1db69baa76b7425a24e473f9a58d9d668

packages/shield-controller/src/polling-with-timeout-abort.ts
lwin-kyaw: Added a cancellation to the delay here, https://github.com/MetaMask/core/pull/6847/commits/27a1acb1db69baa76b7425a24e473f9a58d9d668

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+      timeout?: number;
+      pollInterval?: number;
+      fnName?: string;
+    } = {},
+  ) {
+    const timeout = pollingOptions.timeout ?? this.#timeout;
+    const pollInterval = pollingOptions.pollInterval ?? this.#pollInterval;
+
+    // clean up the request entry if it exists
+    this.abortPendingRequest(requestId);
```
matthiasgeihs: Just curious, why did you rename this from `abortPendingRequests` to `abortPendingRequest`?

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+    // At this point, the polling loop has exited and abortController is aborted
+    const abortReason = abortController.signal.reason;
+    const errorMessage = pollingOptions.fnName
+      ? `${pollingOptions.fnName}: ${abortReason}`
+      : abortReason;
+    throw new Error(errorMessage);
+  }
+
+  /**
+   * Abort the pending requests.
```
matthiasgeihs: nit: 
```suggestion
   * Abort the pending request.
```
because you changed the function name

Just curious, why did you rename this function from `abortPendingRequests` to `abortPendingRequest`?

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+   * @param requestId - The ID of the request to abort.
+   */
+  abortPendingRequest(requestId: string) {
+    // firstly clean up the request entry if it exists
+    // note: this does not abort the request, it only cleans up the request entry for the next polling cycle
+    const existingEntry = this.#cleanUpRequestEntryIfExists(requestId);
+    // then abort the request if it exists
+    // note: this does abort the request, but it will not trigger the abort handler (hence, {@link cleanUpRequestEntryIfExists} will not be called)
+    // coz the AbortHandler event listener is already removed from the AbortSignal
+    existingEntry?.abortController.abort(this.ABORT_REASON_CANCELLED);
```
matthiasgeihs: Sorry, I still don't get it. Why don't you simply call `abortController.abort` here? Wouldn't that be much simpler?
```suggestion
    const requestEntry = this.#requestEntries.get(requestId);
    requestEntry?.abortController.abort(this.ABORT_REASON_CANCELLED);
```

Could there be a case where `existingEntry` is undefined here and the controller doesn't get aborted?

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+   * @param abortSignal - The abort signal to listen to.
+   * @returns A promise that resolves when the delay is complete.
+   */
+  async #delayWithAbortSignal(ms: number, abortSignal: AbortSignal) {
+    return new Promise((resolve) => {
+      let timer: NodeJS.Timeout | null = null;
+
+      const abortHandlerForDelay = () => {
+        // clear the timeout and resolve the promise
+        // Note: we don't reject the promise as this is only a dummy delay
```
matthiasgeihs: This is quite unusual. The convention is that if a signal gets aborted, then the function throws an error, afaik. Why did you decide to do it differently here? I usually suggest to stick to conventions because otherwise people will have a hard time understanding the code.

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+    return requestEntry;
+  }
+
+  /**
+   * Delay with an abort signal.
+   * This will delay the execution of the code until the abort signal is triggered.
+   *
+   * @param ms - The number of milliseconds to delay.
+   * @param abortSignal - The abort signal to listen to.
+   * @returns A promise that resolves when the delay is complete.
```
matthiasgeihs:
```suggestion
   * @returns A promise that resolves when the delay is complete or when the signal is aborted.
```

packages/shield-controller/src/polling-with-timeout-abort.ts
```diff
@@ -0,0 +1,200 @@
...
+   * @returns The request entry that was cleaned up, if it exists.
+   */
+  #cleanUpOnFinished(requestId: string): RequestEntry | undefined {
+    const requestEntry = this.#cleanUpRequestEntryIfExists(requestId);
+    if (requestEntry) {
+      requestEntry.abortController.signal.removeEventListener(
+        'abort',
+        requestEntry.abortHandler,
+      );
+    }
```
matthiasgeihs: Not sure if the bot is right here, but I have a related question:
Why do you have two cleanup functions, where one only cleans up part of it (except for the handler)?
Couldn't we use just one cleanup function in all cases?

matthiasgeihs (commented):
Sorry, I really don't want to drag this out any longer, but I still don't get it.

Why do you have separate cleanup functions? Could it happen that one is called without the other? Would that be a problem? (e.g., could it happen that the event handler doesn't get removed?) Would it be possible to merge these two cleanup functions and just always call the same one?

Let's have a call to clear things up.

packages/shield-controller/src/backend.ts
```diff
@@ -324,3 +324,27 @@ export function parseSignatureRequestMethod(
...
+
+  const maxRetries = isNaN(computedMaxRetries)
+    ? DEFAULT_MAX_RETRIES
+    : computedMaxRetries;
+
+  return {
+    backoff,
+    maxRetries,
+  };
+}
```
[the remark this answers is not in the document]
lwin-kyaw: Expected.

packages/shield-controller/src/polling-with-policy.test.ts
```diff
@@ -0,0 +1,146 @@
...
+          setTimeout(() => {
+            // eslint-disable-next-line jest/no-conditional-in-test
+            if (abortSignal.aborted) {
+              reject(new Error('test error'));
+            }
+            // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors -- HttpError is a valid error which extends Error class
+            reject(new HttpError(412, 'Results are not available yet'));
+          }, 100);
+        });
+      });
```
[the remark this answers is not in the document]
lwin-kyaw: This is to simulate the retry behavior in tests.

packages/shield-controller/src/polling-with-policy.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  abortPendingRequest(requestId: string) {
+    const abortController = this.#requestEntry.get(requestId);
+    abortController?.abort();
+    this.#cleanup(requestId);
+  }
+
+  #addNewRequestEntry(requestId: string) {
+    // abort the previous request if it exists
+    this.abortPendingRequest(requestId);
```
matthiasgeihs, line 57: Doesn't this have the same issue as before that the request is only aborted async and we could have two concurrent requests interfering with each other?

matthiasgeihs (approved):

packages/shield-controller/src/polling-with-policy.ts
lwin-kyaw: No, coz this only aborts the current AbortController in the `#requestEntry` and does not affect the new abort controller which is added after that.
We don't also attach listeners to the AbortController, so it's safe to abort.

packages/shield-controller/src/backend.ts
```diff
@@ -324,3 +324,28 @@ export function parseSignatureRequestMethod(
...
+ * @param timeout - The timeout in milliseconds.
+ * @param pollInterval - The poll interval in milliseconds.
+ * @returns The polling interval and retry count.
+ */
+function computePollingIntervalAndRetryCount(
+  timeout: number,
+  pollInterval: number,
+) {
+  const backoff = new ConstantBackoff(pollInterval);
+  const computedMaxRetries = Math.floor(timeout / pollInterval) + 1;
```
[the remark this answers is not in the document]
lwin-kyaw: Intended.

PR #6861, opened 2025-10-16
fix(core-backend): reconnection logic

Kriys94:
### Current State
The WebSocket service experienced a critical issue when the backend server accepted connections but immediately disconnected afterward. I was only able to reproduce this behavior locally, not in prod endpoint.

### The Problem
When the server accepts the WebSocket connection but then disconnects rapidly:
- The client would immediately attempt to reconnect (fast reconnection)
- This created a connection thrashing loop: connect → disconnect → reconnect → disconnect
- Multiple code paths could trigger reconnection attempts independently
- The `scheduleConnect` method could be called multiple times during rapid disconnect cycles
- Multiple overlapping reconnection attempts could be scheduled simultaneously
- All disconnected clients would reconnect at exactly the same time, creating a "thundering herd" problem
- This led to unpredictable behavior, duplicate connection attempts, and worsened the reconnection storm

### The Solution
This PR fixes the issue with three key improvements:

#### 1. **Centralize Reconnection via `scheduleConnect`**
All reconnection logic now flows through a single entry point:
- All code paths that need to trigger reconnection now use `scheduleConnect`
- Eliminates scattered reconnection logic across the codebase
- Provides a single point of control for reconnection behavior
- Makes the reconnection strategy consistent and predictable
- Easier to maintain, test, and debug reconnection logic

#### 2. **Idempotent `scheduleConnect`**
Makes `scheduleConnect` idempotent to ensure multiple calls don't create duplicate or overlapping connection attempts:
- Checks if a reconnection is already scheduled before creating a new one
- Ensures only one reconnection timer is active at any given time
- Eliminates race conditions from concurrent connection attempts
- Maintains clear, single-threaded reconnection state

#### 3. **Jitter on Reconnection**
Adds randomized jitter to reconnection timing to prevent thundering herd problems:
- Randomizes the reconnection delay to spread out reconnection attempts across clients
- When the server disconnects many clients simultaneously, they won't all reconnect at the exact same moment
- Reduces load spikes on the server during recovery from incidents
- Improves overall system stability and reliability

**Combined Effect:**
With all three fixes, when the server accepts connections but disconnects immediately:
- All reconnection attempts flow through a single, centralized method (consistency)
- Only one reconnection attempt will be scheduled per client (idempotency)
- Reconnection attempts are staggered across the client population (jitter)
- The reconnection logic follows proper backoff strategy with randomization
- No duplicate timers or overlapping connection attempts
- Predictable, controlled reconnection behavior that respects server health
- Server experiences gradual reconnection load rather than synchronized spikes

### Technical Details
The fix centralizes all reconnection logic through `scheduleConnect`, makes it idempotent by guarding against duplicate timer creation, and adds jitter to the reconnection delay calculation to randomize timing across clients. This prevents both the individual reconnection storm (from rapid connect/disconnect cycles) and the collective thundering herd problem (from synchronized reconnections), while ensuring consistent behavior across the entire service.

## References

- Related to `BackendWebSocketService` in `@metamask/core-backend`
- Fixes # [ADD ISSUE NUMBER HERE]

## Checklist

---

> [!NOTE]
> Introduces a controlled `forceReconnection()` flow, idempotent/jittered backoff scheduling, and stability timer to prevent reconnect thrashing; updates AccountActivityService to use it and expands tests.
> 
> - **BackendWebSocketService**
>   - **New API**: `forceReconnection()` messenger action and method for controlled disconnect-then-reconnect.
>   - **Reconnection**: Idempotent scheduler using `ExponentialBackoff` with jitter; `connect()` no-ops if a reconnect is scheduled; stable-connection timer (10s) resets attempts/backoff.
>   - **Behavior changes**: `disconnect()` now synchronous and resets attempts; improved error handling and timer cleanup; avoids duplicate timers and races.
>   - **Actions/Types**: Expose `forceReconnection`; update method action types union.
> - **AccountActivityService**
>   - Switch to `BackendWebSocketService:forceReconnection` for recovery; remove direct `disconnect`/`connect` sequence.
>   - Update allowed messenger actions accordingly.
> - **Tests**
>   - Add/adjust tests for new reconnection semantics, stable timer, idempotent scheduling, and force reconnection flows; minor test cleanup tweaks.
> - **Changelog**
>   - Document new API, reconnection improvements, fixes for race conditions/memory leaks, and breaking messenger action change for AccountActivityService.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 373a4f67a60b236d96bc7e591bf8ed742d4b1630. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shield-controller/CHANGELOG.md
+++ b/packages/shield-controller/CHANGELOG.md
@@ -7,10 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added optional constructor params, `normalizeSignatureRequest` function which normalize the requests for TypedSignature similar to the security-alerts API. ([#6906](https://github.com/MetaMask/core/pull/6906))
+- Added util function, `parseSignatureRequestMethod` to correctly parse the Json-Rpc method value for the signature request. ([#6906](https://github.com/MetaMask/core/pull/6906))
+
 ### Changed
 
 - Bump `@metamask/base-controller` from `^8.4.1` to `^8.4.2` ([#6917](https://github.com/MetaMask/core/pull/6917))
 - Bump `@metamask/transaction-controller` from `^60.7.0` to `^60.8.0` ([#6883](https://github.com/MetaMask/core/pull/6883))
+- Updated internal MessagingSystem subscriber for TransactionController and SignatureController `stateChange` events. ([#6906](https://github.com/MetaMask/core/pull/6906))
+  - Removed `personal_sign` check from the signature-coverage check. Now every signature requests will be sent to ruleset-engine.
+  - Updated `TransactionMeta.SimulationData` check conditional to shallow comparison instead of referential comparison, to avoid triggering unnecessary coverage-check requests.
+- Removed signature data validation from the internal `makeInitSignatureCoverageCheckBody` function. ([#6906](https://github.com/MetaMask/core/pull/6906))
+  - As signature data is not always `string` (e.g. `eth_signTypedData` uses Array of Object) and the data is already validated in the SignatureController before adding to the state.
 
 ## [0.3.2]
 
--- a/packages/shield-controller/package.json
+++ b/packages/shield-controller/package.json
@@ -61,6 +61,7 @@
     "@types/jest": "^27.4.1",
     "deepmerge": "^4.2.2",
     "jest": "^27.5.1",
+    "lodash": "^4.17.21",
     "ts-jest": "^27.1.4",
     "typedoc": "^0.24.8",
     "typedoc-plugin-missing-exports": "^2.0.0",
--- a/packages/shield-controller/src/ShieldController.test.ts
+++ b/packages/shield-controller/src/ShieldController.test.ts
@@ -11,6 +11,8 @@ import {
 } from '@metamask/transaction-controller';
 
 import { ShieldController } from './ShieldController';
+import type { NormalizeSignatureRequestFn } from './types';
+import { TX_META_SIMULATION_DATA_MOCKS } from '../tests/data';
 import { createMockBackend, MOCK_COVERAGE_ID } from '../tests/mocks/backend';
 import { createMockMessenger } from '../tests/mocks/messenger';
 import {
@@ -25,14 +27,17 @@ import {
  * @param options - The options for setup.
  * @param options.coverageHistoryLimit - The coverage history limit.
  * @param options.transactionHistoryLimit - The transaction history limit.
+ * @param options.normalizeSignatureRequest - The function to normalize the signature request.
  * @returns Objects that have been created for testing.
  */
 function setup({
   coverageHistoryLimit,
   transactionHistoryLimit,
+  normalizeSignatureRequest,
 }: {
   coverageHistoryLimit?: number;
   transactionHistoryLimit?: number;
+  normalizeSignatureRequest?: NormalizeSignatureRequestFn;
 } = {}) {
   const backend = createMockBackend();
   const { messenger, baseMessenger } = createMockMessenger();
@@ -42,6 +47,7 @@ function setup({
     coverageHistoryLimit,
     transactionHistoryLimit,
     messenger,
+    normalizeSignatureRequest,
   });
   controller.start();
   return {
@@ -169,6 +175,47 @@ describe('ShieldController', () => {
       });
     });
 
+    TX_META_SIMULATION_DATA_MOCKS.forEach(
+      ({ description, previousSimulationData, newSimulationData }) => {
+        it(`should check coverage when ${description}`, async () => {
+          const { baseMessenger, backend } = setup();
+          const previousTxMeta = {
+            ...generateMockTxMeta(),
+            simulationData: previousSimulationData,
+          };
+          const coverageResultReceived =
+            setupCoverageResultReceived(baseMessenger);
+
+          // Add transaction.
+          baseMessenger.publish(
+            'TransactionController:stateChange',
+            { transactions: [previousTxMeta] } as TransactionControllerState,
+            undefined as never,
+          );
+          expect(await coverageResultReceived).toBeUndefined();
+          expect(backend.checkCoverage).toHaveBeenCalledWith({
+            txMeta: previousTxMeta,
+          });
+
+          // Simulate transaction.
+          const txMeta2 = { ...previousTxMeta };
+          txMeta2.simulationData = newSimulationData;
+          const coverageResultReceived2 =
+            setupCoverageResultReceived(baseMessenger);
+          baseMessenger.publish(
+            'TransactionController:stateChange',
+            { transactions: [txMeta2] } as TransactionControllerState,
+            undefined as never,
+          );
+          expect(await coverageResultReceived2).toBeUndefined();
+          expect(backend.checkCoverage).toHaveBeenCalledWith({
+            coverageId: MOCK_COVERAGE_ID,
+            txMeta: txMeta2,
+          });
+        });
+      },
+    );
+
     it('throws an error when the coverage ID has changed', async () => {
       const { controller, backend } = setup();
       backend.checkCoverage.mockResolvedValueOnce({
@@ -186,9 +233,10 @@ describe('ShieldController', () => {
   });
 
   describe('checkSignatureCoverage', () => {
+    const MOCK_SIGNATURE_REQUEST = generateMockSignatureRequest();
+
     it('should check signature coverage', async () => {
       const { baseMessenger, backend } = setup();
-      const signatureRequest = generateMockSignatureRequest();
       const coverageResultReceived = new Promise<void>((resolve) => {
         baseMessenger.subscribe(
           'ShieldController:coverageResultReceived',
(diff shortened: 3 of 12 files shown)


packages/core-backend/src/AccountActivityService.ts
```diff
@@ -559,16 +559,11 @@ export class AccountActivityService {
...
+    log('Forcing WebSocket reconnection to clean up subscription state');
 
-      await this.#messenger.call('BackendWebSocketService:disconnect');
-      await this.#messenger.call('BackendWebSocketService:connect');
-    } catch (error) {
-      log('Failed to force WebSocket reconnection', { error });
-    }
+    // Use the dedicated forceReconnection method which performs a controlled
+    // disconnect-then-connect sequence to clean up subscription state
+    await this.#messenger.call('BackendWebSocketService:forceReconnection');
```
Kriys94, line 566: I did this because I want to fully control the reconnection via scheduleConnect

packages/core-backend/src/BackendWebSocketService-method-action-types.ts
```diff
@@ -25,6 +25,27 @@ export type BackendWebSocketServiceDisconnectAction = {
...
+ * - Recovering from subscription/unsubscription issues
+ * - Cleaning up orphaned subscriptions
+ * - Forcing a fresh subscription state
+ *
+ * @returns Promise that resolves when disconnection is complete (reconnection is scheduled)
+ */
+export type BackendWebSocketServiceForceReconnectionAction = {
+  type: `BackendWebSocketService:forceReconnection`;
+  handler: BackendWebSocketService['forceReconnection'];
+};
```
Kriys94, line 47: I did this because I want to fully control the reconnection via scheduleConnect

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -299,6 +300,8 @@ export class BackendWebSocketService {
 
   #connectionTimeout: NodeJS.Timeout | null = null;
 
+  #stableConnectionTimer: NodeJS.Timeout | null = null;
```
Kriys94, line 304: Even if the backend accepts a connection, we make sure that we dont re-initialise the the reconnectionAttempt (used for the exponential backoff). This makes sure that the exponential backoff is not bypassed.

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -442,6 +443,12 @@ export class BackendWebSocketService {
       return;
     }
 
+    // If a reconnect is already scheduled, defer to it to avoid bypassing exponential backoff
+    // This prevents rapid loops when server accepts then immediately closes connections
+    if (this.#reconnectTimer) {
+      return;
+    }
```
Kriys94, line 457: // If a reconnect is already scheduled, defer to it to avoid bypassing exponential backoff

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -452,15 +459,11 @@ export class BackendWebSocketService {
           'AuthenticationController:getBearerToken',
         );
         if (!token) {
-          this.#scheduleReconnect();
```
Kriys94, line 455: move this into the catch block L488

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -1300,14 +1349,48 @@ export class BackendWebSocketService {
...
+   *
+   * If a reconnect is already scheduled, this is a no-op to prevent:
+   * - Orphaned timers (memory leak)
+   * - Inflated reconnect attempts counter
+   * - Prematurely long delays
    */
-  #scheduleReconnect(): void {
+  #scheduleConnect(): void {
+    // If a reconnect is already scheduled, don't schedule another one
+    if (this.#reconnectTimer) {
```
Kriys94, line 1382: Important, before I was just replacing this.#reconnectTimer. Now I just bypass if there is already an reconnectTimer in flight

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `forceReconnection()` method to `BackendWebSocketService` for controlled subscription state cleanup ([#6861](https://github.com/MetaMask/core/pull/6861))
```
mcmire, line 12: Great changelog entries! It's really easy to understand the new behavior and the consequences of that behavior.

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -1300,14 +1349,48 @@ export class BackendWebSocketService {
...
+   *
+   * For user-initiated actions (sign in, unlock), call connect() directly instead.
+   *
+   * If a reconnect is already scheduled, this is a no-op to prevent:
+   * - Orphaned timers (memory leak)
+   * - Inflated reconnect attempts counter
+   * - Prematurely long delays
    */
-  #scheduleReconnect(): void {
+  #scheduleConnect(): void {
```
mcmire: Nit: It seems this is always called after connecting once or attempting to connect once, maybe `#scheduleReconnect` is a better name?

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -1561,7 +1765,8 @@ describe('BackendWebSocketService', () => {
           mockEnabledCallback.mockReturnValue(false);
 
           // Advance time to trigger reconnection check
-          await completeAsyncOperations(50);
+          // With jitter (±25%), delay with reconnectDelay=50ms is 37.5-62.5ms
+          await completeAsyncOperations(70);
```
mcmire, line 1754: Tip: Similar as above, you might try mocking `Math.random` if this test starts failing intermittently.

packages/core-backend/src/BackendWebSocketService.test.ts
```diff
@@ -703,13 +683,15 @@ describe('BackendWebSocketService', () => {
           );
 
           // Advance time to trigger reconnection attempt
-          await completeAsyncOperations(100);
+          // With jitter (±25%), first attempt delay is 375-625ms (base 500ms)
```
mcmire: Tip: If this test fails intermittently, you could try mocking `Math.random`. We use a similar "exponential backoff with jitter" algorithm when making requests to the network, you can see the tests for this here: https://github.com/MetaMask/core/blob/b0c9d5817fa27a9fc248fe5d9e5cb867fc1b2c03/packages/controller-utils/src/create-service-policy.test.ts#L233

packages/core-backend/src/BackendWebSocketService.ts
```diff
@@ -1300,14 +1349,48 @@ export class BackendWebSocketService {
...
+    // If a reconnect is already scheduled, don't schedule another one
+    if (this.#reconnectTimer) {
+      return;
+    }
+
+    // Increment attempts BEFORE calculating delay so backoff grows properly
     this.#reconnectAttempts += 1;
 
+    // Calculate delay with exponential backoff
```
mcmire: Nit: Probably not important now, but if we like we might want to reuse the same exponential backoff formula we are using for network requests. The Cockatiel library provides a way to generate a series of backoffs here: https://github.com/connor4312/cockatiel/blob/f87137b9bb55e0cdfb02e7d5c401e47c6f5fcc9f/src/backoff/ExponentialBackoff.ts ([this](https://github.com/connor4312/cockatiel/blob/f87137b9bb55e0cdfb02e7d5c401e47c6f5fcc9f/src/backoff/ExponentialBackoffGenerators.ts#L55) is the formula itself).

packages/core-backend/CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Performs a controlled disconnect-then-reconnect sequence with exponential backoff
+  - Useful for recovering from subscription/unsubscription issues and cleaning up orphaned subscriptions
+  - Add `BackendWebSocketService:forceReconnection` messenger action
+- Add stable connection timer to prevent rapid reconnection loops ([#6861](https://github.com/MetaMask/core/pull/6861))
+  - Connection must stay stable for 10 seconds before resetting reconnect attempts
+  - Prevents issues when server accepts connection then immediately closes it
+
+### Changed
+
+- Update `AccountActivityService` to use new `forceReconnection()` method instead of manually calling disconnect/connect ([#6861](https://github.com/MetaMask/core/pull/6861))
```
mcmire, line 23: It looks like the list of allowed actions for `AccountActivityService` was updated but this is not reflected here, should we add it?

``` suggestion
- **BREAKING:** Update allowed actions for `AccountActivityService` messenger: remove `BackendWebSocketService:disconnect`, add `BackendWebSocketService:forceConnect` ([#6861](https://github.com/MetaMask/core/pull/6861))
- Update `AccountActivityService` to use new `forceReconnection()` method instead of manually calling disconnect/connect ([#6861](https://github.com/MetaMask/core/pull/6861))

File: packages/core-backend/CHANGELOG.md [hunk 13]
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Improve reconnection scheduling in `BackendWebSocketService` to be idempotent ([#6861](https://github.com/MetaMask/core/pull/6861))
+  - Prevents duplicate reconnection timers and inflated attempt counters
+  - Scheduler checks if reconnect is already scheduled before creating new timer
+  - Renamed internal `#scheduleReconnect()` to `#scheduleConnect()` for clarity
```
mcmire: Nit: This is an implementation detail and can be left out of the changelog

``` suggestion
```

Kriys94: Done

packages/core-backend/CHANGELOG.md
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: THanks for this. I replaced with the Cockatiel library and it works well

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Now I mock Math.random, but I prefer to keep 70 because 50 is causing some test flakiness, given that in this test, `reconnectDelay: 50`, it's better to have some room to make the reconnection really happen.

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Done. I added the mock and put back `await completeAsyncOperations(100);`

packages/core-backend/src/BackendWebSocketService.ts
Kriys94: Done

packages/core-backend/src/BackendWebSocketService.test.ts
Kriys94: Now I use a mock of Math.random

PR #6892, opened 2025-10-17
feat: replace external SSE fetching lib with internal fetch-event-source util

micaelae:
## Explanation

Replaces external @microsoft/fetch-event-source dependency with a custom SSE fetch helper and updates quote streaming to use it. This removes the need to patch `fetch-event-source` on clients. Also updates test mocks for SSE fetching

## Checklist

---

> [!NOTE]
> Replaces `@microsoft/fetch-event-source` with a custom `fetchServerEvents`, updates quote streaming/error handling and QuoteResponse types (incl. Bitcoin), adjusts selectors/tests, and removes the dependency.
> 
> - **Bridge Controller**:
>   - **SSE streaming**: Implement `utils/fetch-server-events.ts` and replace `fetchEventSource` with `fetchServerEvents` in `utils/fetch.ts` (parses `event`/`data`, rethrows `error` events); improve abort error handling in `bridge-controller.ts`.
>   - **Types/Selectors**: Extend `QuoteResponse.trade` to accept `TxData | string | BitcoinTradeData`; add `isEvmQuoteResponse` and use it in `selectors.ts`; update fee/quote utils to use `TxData` casts.
>   - **Tests**: Rework SSE mocks to stream via `ReadableStream`; update SSE tests and snapshots; add server error propagation test.
>   - **Changelog/Deps**: Mark breaking validation change (BitcoinTradeData), note switch to `fetchServerEvents`, and remove `@microsoft/fetch-event-source` from dependencies.
> - **Bridge Status Controller**:
>   - Update type references/usages to new `QuoteResponse` shape in controller, metrics, transaction utils, and tests; add changelog entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cdc95afa3d342700ba6e21bf486de11c4f8ae5ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -31,6 +31,7 @@
 
 ## Delegation Team
 /packages/gator-permissions-controller @MetaMask/delegation
+/packages/eip-7702-internal-rpc-middleware @MetaMask/delegation @MetaMask/core-platform
 
 ## Earn Team
 /packages/earn-controller                   @MetaMask/earn
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Each package in this repository has its own README where you can find installati
 - [`@metamask/delegation-controller`](packages/delegation-controller)
 - [`@metamask/earn-controller`](packages/earn-controller)
 - [`@metamask/eip-5792-middleware`](packages/eip-5792-middleware)
+- [`@metamask/eip-7702-internal-rpc-middleware`](packages/eip-7702-internal-rpc-middleware)
 - [`@metamask/eip1193-permission-middleware`](packages/eip1193-permission-middleware)
 - [`@metamask/ens-controller`](packages/ens-controller)
 - [`@metamask/error-reporting-service`](packages/error-reporting-service)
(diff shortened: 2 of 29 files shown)


packages/bridge-controller/src/utils/fetch-server-events.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Streams server-sent events from the given URL
+ *
+ * @param url - The URL to stream events from
+ * @param options - The options for the SSE stream
+ * @param options.onMessage - The function to call when a message is received
+ * @param options.onError - The function to call when an error occurs
+ * @param options.onClose - The function to call when the stream finishes successfully
+ * @param options.fetchFn - The function to use to fetch the events. Consumers need to provide a fetch function that supports server-sent events.
+ */
+export const fetchServerEvents = async (
```
naugtur, line 11: This implements manually fetching SSE with no retries or continuation. 
I continue to recommend using the builtin EventSource in the browser and a polyfill in react native for a less brittle outcome.

SSE is a formal spec around what in early 21 century we used to call long polling. You make a HTTP request and start processing it without waiting for the body to finish. The server side is sending bytes on that request whenever an event happens. It's still a HTTP request though, so it will ultimately finish or time out. That's when you need to reestablish it. I don't see that done here, so the feature will work for the first few minutes and then require restarting metamask. 

I might be misunderstanding the goals here, because I didn't look into the feature, so if getting the initial response is key and all that is sent after is only a nice to have, ignore this comment.

micaelae: Thanks for the review. A few clarifications on why we picked this approach:

- We intentionally don’t retry or continue the stream since the UI request resets whenever the Swap experience is closed (i.e popup closes, navigation). We don’t want to preserve or reuse the connection in this case
- We also skip retries because the controller already polls on a schedule, we can revisit this in the future but the goal of this is just to get parity with the existing fetching behavior we use
- We avoided using EventSource to keep the controller platform-agnostic and avoid bringing in polyfills on mobile
- The current implementation still processes streamed bytes right away without waiting for the full body to complete

PR #6902, opened 2025-10-20
feat(json-rpc-engine): Type-safe context inference

rekmarks:
## Checklist

---

> [!NOTE]
> Introduce generic, type-safe MiddlewareContext and JsonRpcEngineV2.create() factory with improved type inference; update server, utils/exports, tests, and docs accordingly.
> 
> - **Engine V2**:
>   - Add `JsonRpcEngineV2.create()` static factory; make constructor private and enforce non-empty middleware.
>   - Enhance generics: middleware now typed with `Context`, `ResultConstraint`; improved request/notification typing and composition.
>   - Support context-typed `handle()`/`asMiddleware()`; validate incompatible contexts at type level; new internal invalid-engine typing.
> - **MiddlewareContext**:
>   - Make `MiddlewareContext` generic with strict key/value typing; add `assertGet`/`set` typings, `EmptyContext`, and `MergeContexts` util.
> - **Utils/Types**:
>   - Re-export `Json`, `JsonRpcParams`; add `UnionToIntersection` type.
>   - Extend `JsonRpcEngineError` with cross-version `isInstance()`.
> - **Server**:
>   - Construct engine via `create()` when passed middleware; clarify/adjust param validation and id handling.
> - **Compatibility utils**:
>   - Type `propagateToContext` to accept typed context; minor param-handling tweaks.
> - **Tests**:
>   - Update to use `create()` and new helper middlewares; add coverage for context typing, invalid compositions, and new behaviors.
> - **Docs**:
>   - Rewrite/expand README examples for `create()`, typed middleware/context, composition patterns, and error semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75f61e771138c3cf60063129e658e03039e4e829. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/json-rpc-engine/src/v2/JsonRpcEngineV2.ts
+++ b/packages/json-rpc-engine/src/v2/JsonRpcEngineV2.ts
@@ -74,9 +74,13 @@ type ConstructorOptions<
 };
 
 type RequestOf<Middleware> =
-  // Non-polluting `any` constraint.
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Middleware extends JsonRpcMiddleware<infer Request, any, any>
+  Middleware extends JsonRpcMiddleware<
+    infer Request,
+    ResultConstraint<infer Request>,
+    // Non-polluting `any` constraint.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    any
+  >
     ? Request
     : never;
 
--- a/packages/json-rpc-engine/src/v2/MiddlewareContext.ts
+++ b/packages/json-rpc-engine/src/v2/MiddlewareContext.ts
@@ -1,3 +1,5 @@
+import type { UnionToIntersection } from './utils';
+
 /**
  * An context object for middleware that attempts to protect against accidental
  * modifications. Its interface is frozen.
@@ -71,20 +73,6 @@ export class MiddlewareContext<
  * Infer the KeyValues type from a {@link MiddlewareContext}.
  */
 type InferKeyValues<T> = T extends MiddlewareContext<infer U> ? U : never;
-
-/**
- * An unholy incantation that converts a union of object types into an
- * intersection of object types.
- *
- * @example
- * type A = { a: string } | { b: number };
- * type B = UnionToIntersection<A>; // { a: string } & { b: number }
- */
-type UnionToIntersection<U> = (
-  U extends never ? never : (k: U) => void
-) extends (k: infer I) => void
-  ? I
-  : never;
 
 /**
  * Simplifies an object type by "merging" its properties.
--- a/packages/json-rpc-engine/src/v2/utils.ts
+++ b/packages/json-rpc-engine/src/v2/utils.ts
@@ -25,6 +25,20 @@ export const isNotification = <Params extends JsonRpcParams>(
 ): msg is JsonRpcNotification<Params> => !isRequest(msg);
 
 /**
+ * An unholy incantation that converts a union of object types into an
+ * intersection of object types.
+ *
+ * @example
+ * type A = { a: string } | { b: number };
+ * type B = UnionToIntersection<A>; // { a: string } & { b: number }
+ */
+export type UnionToIntersection<U> = (
+  U extends never ? never : (k: U) => void
+) extends (k: infer I) => void
+  ? I
+  : never;
+
+/**
  * JSON-stringifies a value.
  *
  * @param value - The value to stringify.
--- a/packages/json-rpc-engine/README.md
+++ b/packages/json-rpc-engine/README.md
@@ -513,6 +513,8 @@ engine in which they are used. To be used within the same engine, a set of middl
 compatible types. Specifically, all middleware must:
 
 - Handle either `JsonRpcRequest`, `JsonRpcNotification`, or both (i.e. `JsonRpcCall`)
+  - It is okay to mix `JsonRpcCall` middleware with either `JsonRpcRequest` or `JsonRpcNotification`
+    middleware, as long as the latter two are not mixed together.
 - Return valid results for the overall request type
 - Specify mutually inclusive context types
   - The context types may be the same, partially intersecting, or completely disjoint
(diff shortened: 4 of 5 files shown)


packages/json-rpc-engine/src/v2/JsonRpcEngineV2.ts
```diff
@@ -98,15 +132,65 @@ type HandleOptions = {
...
+      JsonRpcMiddleware<
+        InputRequest,
+        ResultConstraint<InputRequest>,
+        MergedContext
+      >
+    >;
+    return new JsonRpcEngineV2<InputRequest, MergedContext>({
+      middleware: mw,
+    }) as MergedContext extends never
+      ? never
```
mcmire: Nit: Instead of `never`, do you think it'd be worth it to return a type that contains some kind of error message? For instance, going off of [this TypeScript discussion](https://github.com/microsoft/TypeScript/issues/23689#issuecomment-2557999403) you could add this to the top:

``` typescript
const INVALID_ENGINE = Symbol('Invalid engine');

type InvalidEngine<T extends string> = { [INVALID_ENGINE]: T };
```

And then here we could say:

``` suggestion
      ? InvalidEngine<"Some middleware have incompatible contexts.">
```

When attempting to use an invalid engine, the developer would see:

```
Property 'handle' does not exist on type 'InvalidEngine<"Some middleware have incompatible contexts.">'. (tsserver 2339)
```

packages/json-rpc-engine/src/v2/MiddlewareContext.test.ts
```diff
@@ -21,28 +16,48 @@ describe('MiddlewareContext', () => {
     expect(Object.isFrozen(context)).toBe(true);
   });
 
+  it('type errors and returns undefined when getting unknown keys', () => {
+    const context = new MiddlewareContext<{ test: string }>();
+    // @ts-expect-error - foo is not a valid key
```
mcmire, line 21: Nit: Would it be worth adding `tsd` or `tstyche` tests at some point for some of these classes? This would allow us to write assertions at the type level.

packages/json-rpc-engine/src/v2/MiddlewareContext.ts
```diff
@@ -1,29 +1,144 @@
...
   }
 }
+
+/**
+ * Infer the KeyValues type from a {@link MiddlewareContext}.
+ */
+// Using `any` in this constraint does not pollute other types.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type InferKeyValues<T extends MiddlewareContext<any>> =
+  T extends MiddlewareContext<infer U> ? U : never;
```
mcmire: Nit: Is it possible to simplify this to:

``` suggestion
type InferKeyValues<T> = T extends MiddlewareContext<infer U> ? U : never;
```

packages/json-rpc-engine/src/v2/MiddlewareContext.ts
```diff
@@ -1,29 +1,144 @@
...
+ *
+ * - Expands intersections into a single object type.
+ * - Forces mapped/conditional results to resolve into a readable shape.
+ * - No runtime effect; purely a type-level normalization.
+ *
+ * @example
+ * type A = { a: string } & { b: number };
+ * type B = Simplify<A>; // { a: string; b: number }
+ */
+type Simplify<T> = T extends infer O ? { [K in keyof O]: O[K] } : never;
```
mcmire, line 88: I would have thought this assumed that `O` is an object type. But `Simplify<1234>` just works. Interesting 🤔

packages/json-rpc-engine/src/v2/MiddlewareContext.ts
```diff
@@ -1,29 +1,144 @@
...
+/**
+ * An unholy incantation that converts a union of object types into an
+ * intersection of object types.
+ *
+ * @example
+ * type A = { a: string } | { b: number };
+ * type B = UnionToIntersection<A>; // { a: string } & { b: number }
+ */
+type UnionToIntersection<U> = (
+  U extends unknown ? (k: U) => void : never
```
mcmire: Incantation is right! 😅 But this looks vaguely familiar based on my adventures with functions and contravariance. So I _think_ I see what TypeScript is doing.

Nit: Would this be _ever so slightly_ more readable if this were flipped?

``` suggestion
  U extends never ? never : (k: U) => void
```

packages/json-rpc-engine/src/v2/utils.ts
```diff
@@ -33,9 +34,29 @@ export function stringify(value: unknown): string {
...
   }
+
+  /**
+   * Check if a value is a {@link JsonRpcEngineError} instance.
+   * Works across different package versions in the same realm.
+   *
+   * @param value - The value to check.
+   * @returns Whether the value is a {@link JsonRpcEngineError} instance.
+   */
+  static isInstance<Value extends Error>(
```
mcmire, line 68: Interesting. Is there a case where you needed this?

mcmire (approved):
Had one question about representing an invalid engine better, as well as some other comments. None of this is blocking though. I think this works really well.

packages/json-rpc-engine/src/v2/MiddlewareContext.ts
rekmarks: Yeah, this one is whacky. Also called "Prettify" in some code bases. Why does it work? I have no idea. (I can see why it would but I doubt that I understand the mechanics end-to-end.)

packages/json-rpc-engine/src/v2/utils.ts
rekmarks: I have been in the case where I want to know if something is a `JsonRpcError` and `instanceof` has not availed me due to multiple different versions at play. I could see the same thing happening with the new `JsonRpcEngineError`, since we have multiple versions of `JsonRpcEngine` in production at the moment (e.g. [3 different major versions in the extension](https://github.com/MetaMask/metamask-extension/blob/9d9e625b37d10ba57615257cbe117eaf770c6710/yarn.lock#L6480-L6511)).

Edit: It's possible that those different versions do not occur in the same realm, but if they ever do, this ensures that there won't be a problem.

packages/json-rpc-engine/src/v2/JsonRpcEngineV2.ts
rekmarks: This is a great idea, I'll try it.

packages/json-rpc-engine/src/v2/utils.ts
mcmire: Oh yeah! Actually now that you mention it, I have run into this problem too. This solution makes sense.

packages/json-rpc-engine/src/v2/MiddlewareContext.test.ts
rekmarks: I think we're squeaking by with `@ts-expect-error` for now, although I'm not opposed to the idea if we think it becomes warranted.

PR #6932, opened 2025-10-23
feat: bitcoin fee calc for display and error handling

ghgoodreau:
## Explanation

Fix Bitcoin network fee computation in bridge quotes

Problem:
Bitcoin bridge quotes were not computing network fees, causing them to always show $0 for gas costs.

Root Cause:
The appendNonEvmFees function only called the snap's computeFee method when trade was a string. This worked for Solana (which has string trades), but Bitcoin trades are objects with structure { unsignedPsbtBase64: string, inputsToSign: null }.

Fix:
Remove the typeof trade === 'string' check
Extract unsignedPsbtBase64 from Bitcoin trade objects before passing to snap
Support both 'base' (Solana) and 'priority' (Bitcoin) fee types in snap response
Add error handling to return quotes with nonEvmFeesInNative: undefined if snap fails.
> snap fails on bitcoin when the account does not have enough for gas. we added this error handling to allow 'quote shopping' on bitcoin without a runtime error causing the quote to fail.

## References

Draft client PR: https://github.com/MetaMask/metamask-extension/pull/37137

## Checklist

---

> [!NOTE]
> Fixes Bitcoin fee calculation by normalizing trade data to PSBT base64, supporting 'priority' fee type, and adding robust error handling.
> 
> - **Fees (non‑EVM)**:
>   - Normalize `trade` to a string for snaps; extract `unsignedPsbtBase64` from Bitcoin `trade` objects.
>   - Support `'priority'` fee type in snap responses (fallback to `'base'`/first result).
>   - Skip fee computation when no snap account or no trade; wrap in try/catch and return quotes with `nonEvmFeesInNative: undefined` on failure.
> - **Types/Exports**:
>   - Export `BitcoinTradeData` in `src/index.ts` for external use.
> - **Changelog**:
>   - Add entry documenting Bitcoin fee computation fix and clean up formatting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cafd6d09e8802ac4a867a4fafdf5a2b9cc5e7197. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-controller/CHANGELOG.md
+++ b/packages/bridge-controller/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Clean up SSE stream reader after use ([#6965](https://github.com/MetaMask/core/pull/6965))
- 
+
 ### Fixed
 
 - Fix Bitcoin network fee computation by extracting `unsignedPsbtBase64` from Bitcoin trade objects and supporting `'priority'` fee type from Bitcoin snap ([#XXXX](https://github.com/MetaMask/core/pull/XXXX))


packages/bridge-controller/src/utils/quote-fees.ts
```diff
@@ -130,16 +144,29 @@ const appendNonEvmFees = async (
...
+        // Return quote with undefined fee if snap fails (e.g., insufficient UTXO funds)
+        // Client can render special UI or skip the quote card row for quotes with missing fee data
+        console.error(
+          `Failed to compute non-EVM fees for quote ${quote.requestId}:`,
+          error,
+        );
+        return {
+          ...quoteResponse,
+          nonEvmFeesInNative: undefined,
+        };
```
micaelae, line 168: This is inconsistent with how we handle missing fees for EVM. `appendL1GasFees` filters out quotes if the fee calculation throws an error because that makes the clients' validations for insufficient funds and ability to submit inaccurate

How common are these failures?

Also if the user has insufficient bitcoin, does the snap throw an error here?

SteP-n-s (approved):

packages/bridge-controller/src/utils/quote-fees.ts
ghgoodreau: @micaelae The failure is common as it occurs every time a Bitcoin quote is requested with an insufficient balance, as the snap fails to calculate due to insufficient UXTO. this was causing quotes to not be shown

micaelae (dismissed):

micaelae (dismissed):

PR #6938, opened 2025-10-24
feat: changes to  AccountTrackerController to enable migration in extension

bergarces:
## Explanation

Adds optional callback to disable fetching balances in AccountTrackerController.

Attempts to refresh balance when a network is added and when the keyring is unlocked.

## References

* Related to https://consensyssoftware.atlassian.net/browse/ASSETS-1368

## Checklist

---

> [!NOTE]
> Adds optional fetchingEnabled gate, refreshes on networkAdded/unlock, and updates tx handlers to refresh both from/to addresses via new refreshAddresses; updates tests and changelog with breaking notes.
> 
> - **AccountTrackerController**:
>   - Add optional `fetchingEnabled` constructor callback; `refresh` short-circuits when disabled.
>   - Subscribe to `NetworkController:networkAdded` and `KeyringController:unlock` to trigger balance refreshes.
>   - Update tx event handlers to refresh both `from` and `to` via new `refreshAddresses`; remove single-address refresh path.
>   - Adjust allowed actions/events and inline `PreferencesController:getState` typing; remove legacy `AccountsController:selectedAccountChange` reference.
> - **Tests**:
>   - Add cases for `networkAdded`, `unlock`, and disabled fetching; address checksum fixes.
> - **Changelog**:
>   - Document breaking event/action changes and the new `fetchingEnabled` option.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 878ed7e1b0254edf6d523de100f346aa372da94c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "651.0.0",
+  "version": "652.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/seedless-onboarding-controller/CHANGELOG.md
+++ b/packages/seedless-onboarding-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [6.0.0]
+
 ### Added
 
 - Added new public method, `checkIsSeedlessOnboardingUserAuthenticated` to validate the controller authenticate tokens state. ([#6998](https://github.com/MetaMask/core/pull/6998))
@@ -216,7 +218,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `checkIsPasswordOutdated`: Check if the password is current device is outdated, i.e. user changed password in another device.
     - `clearState`: Reset the state of the controller to the defaults.
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@5.0.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@6.0.0...HEAD
+[6.0.0]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@5.0.0...@metamask/seedless-onboarding-controller@6.0.0
 [5.0.0]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@4.1.1...@metamask/seedless-onboarding-controller@5.0.0
 [4.1.1]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@4.1.0...@metamask/seedless-onboarding-controller@4.1.1
 [4.1.0]: https://github.com/MetaMask/core/compare/@metamask/seedless-onboarding-controller@4.0.0...@metamask/seedless-onboarding-controller@4.1.0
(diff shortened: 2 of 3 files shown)


bergarces:
@metamaskbot publish-preview

packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -540,13 +565,28 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
     });
   }
 
-  async #refreshAddress(networkClientIds: NetworkClientId[], address: string) {
-    const checksumAddress = toChecksumHexAddress(address) as ChecksumAddress;
+  async refreshAddresses({
```
bergarces, line 570: Making this public the same as `refresh`. The reason being, that if we end up missing some edge case in which we need to update and there is no event available to listen to, we can then dispatch an action from the client to do it.

packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -540,13 +565,28 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
...
+        checksummedAddresses.includes(toChecksumHexAddress(account.address)),
+      );
+
     await this.#refreshAccounts({
       networkClientIds,
-      queryAllAccounts: false,
-      selectedAccount: checksumAddress,
-      allAccounts: [],
+      queryAllAccounts: true,
+      selectedAccount: '0x0',
```
bergarces, line 590: With `queryAllAccounts` set to `true`, this value is not being used.

We really shouldn't be delegating this logic down to the BalanceFetcher class, but that'd require a chunkier refactor.

packages/assets-controllers/src/AccountTrackerController.test.ts
```diff
@@ -167,7 +168,7 @@ describe('AccountTrackerController', () => {
         mockedGetTokenBalancesForMultipleAddresses.mockResolvedValueOnce({
           tokenBalances: {
             '0x0000000000000000000000000000000000000000': {
-              [CHECKSUM_ADDRESS_1]: new BN('123456', 16),
+              [ADDRESS_1]: new BN('123456', 16),
```
bergarces, line 171: @salimtb I'd like your input here if possible.

From what I can see, if we use multicall (`queryAllAccounts: true`), the resulting addresses will be lowercase, but when we query a single address (`queryAllAccounts: false`) the result is a checksummed address.

Therefore, since I have changed the these two events to query multiple accounts (both the `from` and the `to`), I had to change this return value from the test.

Does that sound right?

salimtb: this sounds right yes

salimtb (approved):

packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -174,7 +174,10 @@ export type AccountTrackerControllerActions =
  */
 export type AllowedActions =
   | AccountsControllerListAccountsAction
-  | PreferencesControllerGetStateAction
+  | {
+      type: 'PreferencesController:getState';
+      handler: () => { isMultiAccountBalancesEnabled: boolean };
+    }
```
bergarces, line 180: This fixes type issues in clients, as the extension PreferencesController state is not compatible with core. We only care about `isMultiAccountBalancesEnabled`.

bergarces:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #6990, opened 2025-10-29
refactor(json-rpc-engine): Genericize `JsonRpcServer` middleware / engine

rekmarks:
## Explanation

As it turns out, the `JsonRpcServer` needs its middleware / engine to be generic in order to support certain use cases. Without generics, its middleware type needs to be `JsonRpcMiddleware<JsonRpcCall>`, which prevents constructing engines that only support either requests or notifications.

Enabling this comes at the cost of some type safety, but the server will still handle any errors thrown by its engine due to the latter's lack of support of requests or notifications. The alternative would be to make the server aware of which type of requests it supports at runtime, which seems overkill.

## References

N/A

## Checklist

---

> [!NOTE]
> Genericizes JsonRpcServer over middleware/engine types, exports RequestOf/MergedContextOf, updates v2 index exports, adds tests for request/notification-only servers, and clarifies docs with warnings.
> 
> - **Server (v2/`JsonRpcServer`)**
>   - Genericize class and `Options` over `JsonRpcMiddleware`; type underlying `JsonRpcEngineV2` via `RequestOf`/`MergedContextOf`.
>   - Accept middleware directly with preserved types; add warnings in JSDoc about request vs. notification handling.
> - **Types/Exports**
>   - Export `RequestOf` and `MergedContextOf` from `v2/JsonRpcEngineV2`.
>   - Replace wildcard re-export in `v2/index.ts` with named exports and type re-exports (`JsonRpcMiddleware`, `MiddlewareParams`, `Next`, `ResultConstraint`).
> - **Tests**
>   - Add tests ensuring errors when a server supports only requests or only notifications and receives the wrong kind.
>   - Minor wording/import adjustments in `JsonRpcServer.test.ts`.
> - **Docs**
>   - README: refine warning about mixing `JsonRpcRequest`/`JsonRpcNotification` middleware; add warning about servers that only accept one kind.
>   - Changelog: add PR reference for `JsonRpcEngineV2` item.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e68ff036f23fcc0f1137a7c8f805393ecea3ad5b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/json-rpc-engine/src/v2/JsonRpcServer.test.ts
+++ b/packages/json-rpc-engine/src/v2/JsonRpcServer.test.ts
@@ -219,7 +219,7 @@ describe('JsonRpcServer', () => {
     });
   });
 
-  it('throws if passed a notification when only requests are supported', async () => {
+  it('errors if passed a notification when only requests are supported', async () => {
     const onError = jest.fn();
     const server = new JsonRpcServer<JsonRpcMiddleware<JsonRpcRequest>>({
       middleware: [() => null],
@@ -227,8 +227,9 @@ describe('JsonRpcServer', () => {
     });
 
     const notification = { jsonrpc, method: 'hello' };
-    await server.handle(notification);
+    const response = await server.handle(notification);
 
+    expect(response).toBeUndefined();
     expect(onError).toHaveBeenCalledTimes(1);
     expect(onError).toHaveBeenCalledWith(
       new JsonRpcEngineError(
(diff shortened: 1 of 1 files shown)


packages/json-rpc-engine/src/v2/JsonRpcServer.test.ts
```diff
@@ -218,6 +219,43 @@ describe('JsonRpcServer', () => {
...
+
+    expect(onError).toHaveBeenCalledTimes(1);
+    expect(onError).toHaveBeenCalledWith(
+      new JsonRpcEngineError(
+        `Result returned for notification: ${stringify(notification)}`,
+      ),
+    );
+  });
+
+  it('throws if passed a request when only notifications are supported', async () => {
```
FrederikBolding: These added tests are a bit misleading since it doesn't throw, it returns the JSON-RPC error. Can we assert the result of `server.handle` and modify the test names?

PR #7000, opened 2025-10-29
feat(multichain-account-service): add per-provider throttling for non-EVM account creation

mathieuartu:
## References

Related to [MUL-1222](https://consensyssoftware.atlassian.net/browse/MUL-1222), [MUL-1221](https://consensyssoftware.atlassian.net/browse/MUL-1221)

## Checklist

[MUL-1222]: https://consensyssoftware.atlassian.net/browse/MUL-1222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> Adds per-provider concurrency limits for non-EVM account creation (Solana defaults to 3), refactors creation flow to optionally await/aggregate errors, and updates providers/tests accordingly.
> 
> - **Core**
>   - Introduce `withMaxConcurrency` in `SnapAccountProvider` using a semaphore; add optional `maxConcurrency` to `SnapAccountProviderConfig`.
>   - Add `toRejectedErrorMessage` util and factor non‑EVM creation into `MultichainAccountWallet.#createNonEvmAccounts`, supporting await-all with aggregated errors and background mode logging.
>   - Forward optional `providerConfigs` safely (`?.`) and allow Solana-specific config (including `maxConcurrency`).
> - **Providers**
>   - Update `SolAccountProvider`, `BtcAccountProvider`, and `TrxAccountProvider` to use `SnapAccountProviderConfig` and throttle `createAccounts` via `withMaxConcurrency`.
>   - Default Solana `maxConcurrency: 3`; others unthrottled by default.
> - **Tests**
>   - Add concurrency tests for `SnapAccountProvider`.
>   - Update wallet tests to verify aggregated non‑EVM failures when awaiting all and provider config forwarding.
> - **Docs**
>   - Update `CHANGELOG.md` with new throttling behavior and defaults.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6e2148641ba44deabd1a26b79d466ce4efb6a391. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

[MUL-1221]: https://consensyssoftware.atlassian.net/browse/MUL-1221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mathieuartu:
@metamaskbot publish-preview

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -220,6 +248,83 @@ export class MultichainAccountWallet<
...
+
+        let message = `${error}:`;
+        for (const r of results) {
+          if (r.status === 'rejected') {
+            message += `\n- ${r.reason}`;
+          }
+        }
+        this.#log(`${WARNING_PREFIX} ${message}`);
+        console.warn(message);
+        throw new Error(error);
```
ccharly: WDYT about adding a new helper to "format" this kind of errors? Like:

```ts
// This could live in `logger.ts` for now maybe? 🤔
function toRejectedErrorMessage<Result>(prefix: string, results: PromiseSettledResult<Result>[]) {
  let errorMessage = `${prefix}:`;
  for (const r of results) {
    if (r.status === 'rejected') {
      errorMessage += `\n- ${r.reason}`;
    }
  }
  
  return errorMessage;
}
```

So we can use it like:

```suggestion
        const errorMessage = toRejectedErrorMessage(`Unable to create multichain account group for index: ${groupIndex}`, results);

        this.#log(`${WARNING_PREFIX} ${errorMessage}`);
        console.warn(errorMessage);
        throw new Error(errorMessage);
```

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -220,6 +248,83 @@ export class MultichainAccountWallet<
...
+      const queuedTasks = providers.map((provider) => {
+        const queue = MultichainAccountWallet.getProviderQueue(
+          provider.getName(),
+        );
+        const operation = async () =>
+          provider.createAccounts({
+            entropySource: this.#entropySource,
+            groupIndex,
+          });
+        return queue ? queue.runExclusive(operation) : operation();
```
ccharly: WDYT about having a `#runWithProvider` method, which would take a `provider` as 1st parameter and a callback as 2nd argument.

This method would basically just apply the same logic here, but it would "hide" the fact that we might run it with `queue.runExeclusive` or just calling the callback directly?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -220,6 +248,83 @@ export class MultichainAccountWallet<
...
+        provider.getName(),
+      );
+      const operation = async () => {
+        try {
+          await provider.createAccounts({
+            entropySource: this.#entropySource,
+            groupIndex,
+          });
+        } catch (error) {
+          const errorMessage = `Could not to create account with provider "${provider.getName()}" for multichain account group index: ${groupIndex}`;
```
ccharly: Both error messages (with `awaitAll=true/false`) are not consistent with each other, maybe we want to unify those 2? We could still have a small annotation for the async one, so we know this happened in the background?

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -47,6 +51,30 @@ export class MultichainAccountWallet<
...
 {
+  /**
+   * Per-provider semaphores to throttle background account creation for non-EVM providers.
+   * Only providers listed in PROVIDER_CONCURRENCY_MAP are throttled, others run unthrottled by default.
+   */
+  private static readonly PROVIDER_CONCURRENCY_MAP: Record<string, number> = {
+    [SOL_ACCOUNT_PROVIDER_NAME]: 3, // Limit Solana to 3 concurrent account creations.
+  };
+
+  private static readonly providerQueues: Map<string, Semaphore> = new Map();
```
ccharly: I wonder for this if we could do this instead:
- We rename `NamedAccountProvider` to `Bip44AccountProvider` (we can omit that step, but I think `Named*` is bit limited now 😅)
- We add an `execute` method to the interface
- This method `execute` would basically be "owned" by the provider, thus, allowing to configure it the way he wants
  * This allows use to move the `Semaphore` to the `SolAccountProvider` (we can scope it for this specific provider for now)
  * All other providers would just execute the callback right away, thus, preserving the original behavior for all others
  * This `execute` is basically just the same method than my other proposal with `runWithProvider` actually ([from here](https://github.com/MetaMask/core/pull/7000/files#r2480879305))
- We can then add a `maxConcurrency` field to the `SolAccountProviderConfig` (this does not need to be scoped to `createAccounts`, I guess we can move this "globally" for now, in case we need that concurrency option for other future keyring method calls, WDYT?)

packages/multichain-account-service/src/MultichainAccountWallet.ts
mathieuartu: I moved the whole logic one level higher to keep things DRY and simplify the logic. LMK what you think 😄

packages/multichain-account-service/src/MultichainAccountWallet.ts
mathieuartu: As mentioned in my previous response, I took a completely different route, using a mix of my ideas and yours, LMK what you think! 😄

packages/multichain-account-service/src/MultichainAccountWallet.test.ts
```diff
@@ -364,6 +364,51 @@ describe('MultichainAccountWallet', () => {
...
+      solProvider.createAccounts.mockRejectedValueOnce(
+        new Error('SOL create failed'),
+      );
+
+      await expect(
+        wallet.createMultichainAccountGroup(nextIndex, {
+          waitForAllProvidersToFinishCreatingAccounts: true,
+        }),
+      ).rejects.toThrow(
+        `Unable to create multichain account group for index: ${nextIndex}`,
```
ccharly: Shouldn't we check for the underlying error too? I think t hat `.toThrow` does not check the entire string here, IMO this should be:
```suggestion
        `Unable to create multichain account group for index: ${nextIndex}\n- SOL create failed`,
```

(We can also move the error message above (`'SOL create failed'`) in a variable so we can re-use it here too if you prefer)

packages/multichain-account-service/src/providers/SolAccountProvider.ts
```diff
@@ -11,21 +11,13 @@ import { KeyringClient } from '@metamask/keyring-snap-client';
...
-import type { MultichainAccountServiceMessenger } from 'src/types';
 
-import { SnapAccountProvider } from './SnapAccountProvider';
+import {
+  SnapAccountProvider,
+  type SnapAccountProviderConfig,
+} from './SnapAccountProvider';
 import { withRetry, withTimeout } from './utils';
-
-export type SolAccountProviderConfig = {
```
ccharly, line 19: IMO we could keep those as aliases? I think having them matching their actual provider class name is good (especially if one day we start having diverging config again)

So maybe this could just be `export type SolAccountProviderConfig = SnapAccountProviderConfig` for now?

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add per-provider throttling for non-EVM account creation to improve performance on low-end devices ([#7000](https://github.com/MetaMask/core/pull/7000))
+  - Solana provider is now limited to 3 concurrent account creations when creating multichain account groups.
```
ccharly: Just adding `(by default)` so that we know it can be updated externally

```suggestion
  - Solana provider is now limited to 3 concurrent account creations (by default) when creating multichain account groups.
```

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add per-provider throttling for non-EVM account creation to improve performance on low-end devices ([#7000](https://github.com/MetaMask/core/pull/7000))
```
ccharly, line 12: Not sure my phrasing is the best here, but that's breaking since the config now requires this extra non-optional field:

```suggestion
- **BREAKING:** Add per-provider throttling for non-EVM account creation to improve performance on low-end devices ([#7000](https://github.com/MetaMask/core/pull/7000))
  - Snap-based providers now requires a `maxConcurrency` field.
```

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -38,7 +40,9 @@ export type MultichainAccountServiceOptions = {
   providers?: NamedAccountProvider[];
   providerConfigs?: {
     [EvmAccountProvider.NAME]?: EvmAccountProviderConfig;
-    [SolAccountProvider.NAME]?: SolAccountProviderConfig;
+    [SolAccountProvider.NAME]?: SnapAccountProviderConfig;
+    [BtcAccountProvider.NAME]?: SnapAccountProviderConfig;
+    [TrxAccountProvider.NAME]?: SnapAccountProviderConfig;
```
ccharly: Similar comment about the aliases, but we could re-use them here too!

packages/multichain-account-service/CHANGELOG.md
mathieuartu: I've made changes to make this PR non-breaking, dismissing this feedback 😄

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -39,6 +49,8 @@ export type MultichainAccountServiceOptions = {
   providerConfigs?: {
     [EvmAccountProvider.NAME]?: EvmAccountProviderConfig;
     [SolAccountProvider.NAME]?: SolAccountProviderConfig;
+    [BtcAccountProvider.NAME]?: BtcAccountProviderConfig;
+    [TrxAccountProvider.NAME]?: TrxAccountProviderConfig;
```
ccharly: Actually, we can remove those, those 2 providers are not yet considered "stable" thus, we cannot forward those configs to their providers yet (this we have to be passed in the clients service initialization directly)
```suggestion
```

ccharly: That being said... It's only about types here, so we can probably leave it as-is, that's fine actually! My bad 😁

ccharly (dismissed):
Nice one 💪

PR #7001, opened 2025-10-29
refactor!(eth-json-rpc-provider): Migrate to `JsonRpcEngineV2`

rekmarks:
## Explanation

Per #6327, migrates `@metamask/eth-json-rpc-provider` to `JsonRpcEngineV2` following its recent introduction. Intended to be closely followed by #6976 and subsequently released.

The `InternalProvider` is updated to use `JsonRpcServer` under the hood. It can be constructed with a `JsonRpcServer` or legacy engine, but the latter will be wrapped by a `JsonRpcServer` in order to ensure consistent behavior across all internal providers. Meanwhile, `providerFromEngine()` is removed, it being a useless wrapper over the `InternalProvider` constructor.

Elsewhere in the monorepo, these changes revealed a discrepancy in behavior between the error handling of the legacy engine and `asV2Middleware()`, which the latter is amended to resolve.

In addition, the changes to the `InternalProvider` revealed that the legacy engine tolerates responses with `{ result: undefined }`. This is impossible to express using the V2 engine, which caused some `NetworkController` tests reliant on the legacy behavior to fail. Further investigation proved that these `undefined` results would error elsewhere in our JSON-RPC pipelines, so we simply remove these test cases. The upshot is that we no longer retry `undefined` results for "child requests" in the `retryOnEmpty` middleware. It is unclear if this was occurring in practice, and it ought to be treated as a breach of contract by the RPC endpoint if it does.

## References

- Progresses: #6327

## Checklist

  - Not intended to be shipped without #6976, but it appears we could if we wanted to:
    - https://github.com/MetaMask/metamask-extension/pull/37521
    - https://github.com/MetaMask/metamask-mobile/pull/22154

---

> [!NOTE]
> Migrates InternalProvider to JsonRpcEngineV2, adds providerFromMiddlewareV2, removes providerFromEngine, aligns error handling and retry semantics across packages, and updates consumers/tests and config.
> 
> - **@metamask/eth-json-rpc-provider**:
>   - **Migration**: `InternalProvider` now wraps `JsonRpcEngineV2` (accepts legacy engine or V2 server).
>   - **API**: Add `providerFromMiddlewareV2`; deprecate `providerFromMiddleware`; remove `providerFromEngine`.
>   - **IDs**: Switch from `uuid` to `nanoid` for request ids.
>   - **Exports/Tests**: Update exports and comprehensive tests for V1/V2 engines.
> - **@metamask/json-rpc-engine**:
>   - **V2 Enhancements**: Export additional V2 types; refine `asV2Middleware` to use `deserializeError` and ignore `undefined` errors from legacy middleware; add tests.
>   - **Server**: Type refinements for `JsonRpcServer` and `JsonRpcEngineV2`.
> - **@metamask/eth-json-rpc-middleware**:
>   - **Retry Behavior (BREAKING)**: `retryOnEmpty` no longer treats `undefined` as an empty value; now errors on advancing block tag with `undefined` results.
> - **@metamask/network-controller**:
>   - **Provider**: Replace `providerFromEngine` with direct `InternalProvider` construction in `create-network-client`.
>   - **Tests/Changelog**: Align tests with new empty-value semantics and provider changes.
> - **Tooling**:
>   - Map `@metamask/json-rpc-engine/v2` in Jest/TS paths.
>   - Update block-tracker tests to use `InternalProvider` directly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 684a40c07ab442de5458942b7d8e0de2f82594be. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "660.0.0",
+  "version": "662.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/account-tree-controller/jest.config.js
+++ b/packages/account-tree-controller/jest.config.js
@@ -17,7 +17,7 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 98.34,
+      branches: 100,
       functions: 100,
       lines: 100,
       statements: 100,
--- a/packages/account-tree-controller/src/AccountTreeController.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.ts
@@ -632,6 +632,9 @@ export class AccountTreeController extends BaseController<
     const group = wallet.groups[groupId];
     const persistedGroupMetadata = state.accountGroupsMetadata[groupId];
 
+    // Ensure metadata object exists once at the beginning
+    state.accountGroupsMetadata[groupId] ??= {};
+
     // Apply persisted name if available (including empty strings)
     if (persistedGroupMetadata?.name !== undefined) {
       state.accountTree.wallets[walletId].groups[groupId].metadata.name =
@@ -660,7 +663,6 @@ export class AccountTreeController extends BaseController<
       log(`[${group.id}] Set default name to: "${group.metadata.name}"`);
 
       // Persist the generated name to ensure consistency
-      state.accountGroupsMetadata[groupId] ??= {};
       state.accountGroupsMetadata[groupId].name = {
         value: proposedName,
         // The `lastUpdatedAt` field is used for backup and sync, when comparing local names
@@ -681,7 +683,6 @@ export class AccountTreeController extends BaseController<
           this.#accountOrderCallbacks?.isPinnedAccount?.(account),
         );
       }
-      state.accountGroupsMetadata[groupId] ??= {};
       state.accountGroupsMetadata[groupId].pinned = {
         value: isPinned,
         lastUpdatedAt: 0,
@@ -700,7 +701,6 @@ export class AccountTreeController extends BaseController<
           this.#accountOrderCallbacks?.isHiddenAccount?.(account),
         );
       }
-      state.accountGroupsMetadata[groupId] ??= {};
       state.accountGroupsMetadata[groupId].hidden = {
         value: isHidden,
         lastUpdatedAt: 0,
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [87.1.0]
+
+### Added
+
+- Added `getTrendingTokens` function to fetch trending tokens for specific chains ([#7054]) (https://github.com/MetaMask/core/pull/7054)
+- Added new types `SortTrendingBy` and `TrendingAsset` ([#7054]) (https://github.com/MetaMask/core/pull/7054)
+
 ## [87.0.0]
 
 ### Added
@@ -2239,7 +2246,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use Ethers for AssetsContractController ([#845](https://github.com/MetaMask/core/pull/845))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@87.0.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@87.1.0...HEAD
+[87.1.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@87.0.0...@metamask/assets-controllers@87.1.0
 [87.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@86.0.0...@metamask/assets-controllers@87.0.0
 [86.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@85.0.0...@metamask/assets-controllers@86.0.0
 [85.0.0]: https://github.com/MetaMask/core/compare/@metamask/assets-controllers@84.0.0...@metamask/assets-controllers@85.0.0
--- a/packages/assets-controllers/package.json
+++ b/packages/assets-controllers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/assets-controllers",
-  "version": "87.0.0",
+  "version": "87.1.0",
   "description": "Controllers which manage interactions involving ERC-20, ERC-721, and ERC-1155 tokens (including NFTs)",
   "keywords": [
     "MetaMask",
--- a/packages/assets-controllers/src/index.ts
+++ b/packages/assets-controllers/src/index.ts
@@ -142,7 +142,7 @@ export {
   SUPPORTED_CHAIN_IDS,
   getNativeTokenAddress,
 } from './token-prices-service';
-export { searchTokens } from './token-service';
+export { searchTokens, getTrendingTokens } from './token-service';
 export { RatesController, Cryptocurrency } from './RatesController';
 export type {
   RatesControllerState,
@@ -231,3 +231,4 @@ export type {
 } from './selectors/token-selectors';
 export { selectAssetsBySelectedAccountGroup } from './selectors/token-selectors';
 export { createFormatters } from './utils/formatters';
+export type { SortTrendingBy, TrendingAsset } from './token-service';
--- a/packages/assets-controllers/src/token-service.test.ts
+++ b/packages/assets-controllers/src/token-service.test.ts
@@ -2,9 +2,11 @@ import { toHex } from '@metamask/controller-utils';
 import type { CaipChainId } from '@metamask/utils';
 import nock from 'nock';
 
+import type { SortTrendingBy } from './token-service';
 import {
   fetchTokenListByChainId,
   fetchTokenMetadata,
+  getTrendingTokens,
   searchTokens,
   TOKEN_END_POINT_API,
   TOKEN_METADATA_NO_SUPPORT_ERROR,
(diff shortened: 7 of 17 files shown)


packages/eth-json-rpc-provider/src/internal-provider.ts
```diff
@@ -103,17 +105,17 @@ export class InternalProvider {
    *
    * @param eip1193Request - The request to send.
    * @param callback - A function that is called upon the success or failure of the request.
-   * @deprecated Please use `request` instead.
+   * @deprecated Use {@link request} instead.
```
mcmire: I forgot we still had these methods around. They have been deprecated for a while. If we are introducing breaking changes in this release anyway, how do you feel if we removed these methods (in another PR)? We wouldn't have to make sure they still work.

packages/eth-json-rpc-provider/src/internal-provider.ts
```diff
@@ -1,11 +1,16 @@
-import type { JsonRpcEngine } from '@metamask/json-rpc-engine';
+import { asV2Middleware, type JsonRpcEngine } from '@metamask/json-rpc-engine';
```
mcmire, line 1: Nit: Should we alias this to `LegacyJsonRpcEngine` within this file?

``` suggestion
import {
  asV2Middleware,
  type JsonRpcEngine as LegacyJsonRpcEngine,
} from '@metamask/json-rpc-engine';
```

packages/eth-json-rpc-provider/src/internal-provider.ts
```diff
@@ -1,11 +1,16 @@
...
+import type { JsonRpcMiddleware } from '@metamask/json-rpc-engine/v2';
```
mcmire: Similar, should we alias this to `JsonRpcMiddlewareV2`?

``` suggestion
import type { JsonRpcMiddleware as JsonRpcMiddlewareV2 } from '@metamask/json-rpc-engine/v2';
```

packages/eth-json-rpc-provider/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   - `SafeEventEmitterProvider` is for now still exported as a deprecated alias of `InternalProvider` for backwards compatibility.
+- **BREAKING:** Use `JsonRpcServer` instead of `JsonRpcEngine` ([#7001](https://github.com/MetaMask/core/pull/7001))
+  - Adds a new `server` constructor option to the `InternalProvider` class, mutually exclusive with the now deprecated `engine` option.
+  - Legacy `JsonRpcEngine` instances are wrapped in a `JsonRpcServer` internally wherever they appear.
+    This change should mostly be unobservable. However, due to differences in error handling, this may be breaking for consumers.
+
+### Removed
+
+- **BREAKING:** Remove `providerFromEngine` ([#7001](https://github.com/MetaMask/core/pull/7001))
+  - Use `InternalProvider` directly instead.
```
mcmire, line 32: Should we include a Deprecated section?

``` suggestion

### Deprecated

- Deprecate `providerFromMiddleware` ([#7001](https://github.com/MetaMask/core/pull/7001))
  - Use `providerFromMiddlewareV2` instead, which supports the new middleware from `@metamask/json-rpc-engine/v2`.

File: packages/eth-json-rpc-provider/CHANGELOG.md
Comment (mcmire) [inline packages/eth-json-rpc-provider/CHANGELOG.md:15]: Should we include that `providerFromMiddlewareV2` was added? Perhaps something like this (feel free to reword):

``` suggestion
### Added

- Add `providerFromMiddlewareV2` ([#7001](https://github.com/MetaMask/core/pull/7001))
  - This is designed to handle the new middleware from `@metamask/json-rpc-engine/v2`.

````

File: packages/eth-json-rpc-provider/src/internal-provider.ts [hunk 5]
```diff
@@ -20,48 +25,56 @@ type Eip1193Request<Params extends JsonRpcParams> = {
...
+
+type Options<Middleware extends InternalProviderMiddleware> =
+  | {
+      /**
+       * @deprecated Use `server` instead.
+       */
+      engine: JsonRpcEngine;
+    }
+  | {
+      server: JsonRpcServer<Middleware>;
```
mcmire: Just curious, what are your thoughts on having the internal use `JsonRpcEngine(V2)` rather than `JsonRpcServer`? One of the issues with the existing `JsonRpcEngine` is that it wraps non-JSON-RPC errors, which can complicate or even hide the stack trace in Sentry. If we don't allow this and just have these kinds of errors get thrown then it should make error handling easier. Or is that planned for another PR?

packages/json-rpc-engine/CHANGELOG.md
```diff
@@ -9,9 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `JsonRpcEngineV2` ([#6176](https://github.com/MetaMask/core/pull/6176), [#6971](https://github.com/MetaMask/core/pull/6971), [#6975](https://github.com/MetaMask/core/pull/6975), [#6990](https://github.com/MetaMask/core/pull/6990), [#6991](https://github.com/MetaMask/core/pull/6991), [#7032](https://github.com/MetaMask/core/pull/7032))
-  - This is a complete rewrite of `JsonRpcEngine`, intended to replace the original implementation.
-    See the readme for details.
+- `JsonRpcEngineV2` ([#6176](https://github.com/MetaMask/core/pull/6176), [#6971](https://github.com/MetaMask/core/pull/6971), [#6975](https://github.com/MetaMask/core/pull/6975), [#6990](https://github.com/MetaMask/core/pull/6990), [#6991](https://github.com/MetaMask/core/pull/6991), [#7001](https://github.com/MetaMask/core/pull/7001), [#7032](https://github.com/MetaMask/core/pull/7032))
```
mcmire: Nit: Perhaps 

``` suggestion
- Add `JsonRpcEngineV2` ([#6176](https://github.com/MetaMask/core/pull/6176), [#6971](https://github.com/MetaMask/core/pull/6971), [#6975](https://github.com/MetaMask/core/pull/6975), [#6990](https://github.com/MetaMask/core/pull/6990), [#6991](https://github.com/MetaMask/core/pull/6991), [#7001](https://github.com/MetaMask/core/pull/7001), [#7032](https://github.com/MetaMask/core/pull/7032))
```

jest.config.packages.js
```diff
@@ -80,6 +80,9 @@ module.exports = {
   // Here we ensure that Jest resolves `@metamask/*` imports to the uncompiled source code for packages that live in this repo.
   // NOTE: This must be synchronized with the `paths` option in `tsconfig.packages.json`.
   moduleNameMapper: {
+    '^@metamask/json-rpc-engine/v2$': [
```
mcmire, line 83: Ah yes makes sense 👍🏻

mcmire (commented):
Some suggestions but overall makes sense / looks good.

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: My only concern would be that we pass the internal provider to other libraries that rely on `send` or `sendAsync`. On review, this is in fact true of `eth-query`: https://github.com/MetaMask/eth-query/blob/fc00c5d18b8cd2ec24d0a927de67def65d2435c0/index.js#L111

I can update the deprecation notice to highlight that it's only for backwards compatibility with other libraries.

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: I don't think that's necessary; in IDEs that support JSDoc it shows up like ~`JsonRpcEngine`~, and we added the `V2` suffix to the new engine to handle this case without having to alias the legacy engine.

jest.config.packages.js
rekmarks: "Many Bothans died to give us this information."

packages/eth-json-rpc-provider/src/internal-provider.ts
mcmire: Well, we really ought to stop using that library and use the provider directly. But true, maybe it's a larger effort to outright remove them.

packages/eth-json-rpc-provider/src/internal-provider.ts
mcmire: Okay, that's fair.

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: Good callout. It's not planned for another PR.

If the concern is just to capture the raw error in Sentry, we can use the `onError` option of the `JsonRpcServer`. However, to your point I think that may not be very ergonomic considering how the `InternalProvider` is used...

On review, it would be trivial to swap out the server for a v2 engine. I'll do it.

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: Would an issue for this be best in `/core` or some other location?

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: [4685f4a](https://github.com/MetaMask/core/pull/7001/commits/4685f4a1a5ac1a5213481c9470dced40371ab833)

packages/eth-json-rpc-provider/src/internal-provider.ts
mcmire: Perhaps `core` but also other locations. I think we need to go through and audit where all we are using `eth-query` (and maybe `ethjs-query`).

mcmire: Needless to say it's out of scope for all of this work. I think that's something that we can take on later perhaps.

rekmarks: #7059

packages/eth-json-rpc-provider/package.json
```diff
@@ -56,7 +56,7 @@
     "@metamask/json-rpc-engine": "^10.1.1",
     "@metamask/rpc-errors": "^7.0.2",
     "@metamask/utils": "^11.8.1",
-    "uuid": "^8.3.2"
+    "nanoid": "^3.3.8"
```
mcmire, line 59: Admittedly when I added `request` to the provider I swapped out `nanoid` for `uuid`. My thinking was that `uuid` was more standard. I'm just curious, what are your thoughts on this? (I can be convinced the other way)

mcmire (commented):
This looks good, but I am going to do another pass tomorrow just to be sure.

packages/eth-json-rpc-provider/package.json
rekmarks: Haha, it comes full circle.

Unlike `nanoid`, `uuid` is implementing a standard, however not one that matters for purposes of the `InternalProvider`. Generally speaking, all we need are JSON-RPC ids that are unique within our JSON-RPC pipelines, i.e. per engine and array of middleware functions. This is because some of our middleware depend on the uniqueness of the IDs, e.g. in order to map responses to their promises on the dapp side. (If you had an engine with middleware that you are certain doesn't depend on unique ids you don't need unique ids, but I digress.)

So, given that the only property we care about is uniqueness per engine, `nanoid` is more suitable because its faster and its IDs are smaller. In fact, even `nanoid` is probably overkill, but that's out of scope for this issue.

packages/eth-json-rpc-provider/src/internal-provider.ts
```diff
@@ -137,6 +113,62 @@ export class InternalProvider {
...
       convertEip1193RequestToJsonRpcRequest(eip1193Request);
-    this.#engine.handle(jsonRpcRequest, callback);
+    this.#handleWithCallback(jsonRpcRequest, callback);
   };
+
+  readonly #handle = async <Result extends Json>(
+    jsonRpcRequest: JsonRpcRequest,
+  ): Promise<JsonRpcSuccess<Result>> => {
+    const { id, jsonrpc } = jsonRpcRequest;
+    // This typecast is technicaly unsafe, but we need it to preserve the provider's
```
mcmire, line 123: Ah, right. Different requests will have different results of different types, so should we allow `handle` on the engine to take a type parameter? Or maybe there is just not a way to make this type-safe in the end, anyway, since all we can safely say is that each request must return JSON and from there we need to typecast the result based on our own assumptions of the JSON-RPC method we are calling?

packages/eth-json-rpc-provider/src/provider-from-middleware.ts
```diff
@@ -1,22 +1,50 @@
...
   Params extends JsonRpcParams,
   Result extends Json,
->(middleware: JsonRpcMiddleware<Params, Result>): InternalProvider {
-  const engine: JsonRpcEngine = new JsonRpcEngine();
-  engine.push(middleware);
-  const provider: InternalProvider = providerFromEngine(engine);
-  return provider;
+>(middleware: LegacyJsonRpcMiddleware<Params, Result>): InternalProvider {
+  return providerFromMiddlewareV2(
+    asV2Middleware(middleware) as JsonRpcMiddleware<JsonRpcRequest>,
```
mcmire, line 31: I see that if we remove the type assertion we get a type error around params. Do we need to do some kind of transformation of params to remove `undefined` or something?

mcmire (commented):
Two more questions.

packages/eth-json-rpc-provider/src/internal-provider.ts
rekmarks: I experimented with narrowing the return type of the engine based on its middleware and found that the juice wasn't worth the squeeze. It significantly complicates the types, and the benefit is more "kinda nice" than "necessary". This is why we left the engine / server types at "returns `undefined` for notifications and `Json` / `JsonRpcResponse` otherwise".

Per-method parameters and return types are not supported at all by the engine types, and IMO make more sense at the level of something like a provider. Because those types would have no basis in the engine, they'd be more of a pinky-promise, which is I guess the behavior we're preserving here.

packages/eth-json-rpc-provider/src/provider-from-middleware.ts
rekmarks: It actually fails due to the context generics, specifically by tripping up the `InvalidEngine` type. I do not know why.

rekmarks: We put on our thinking caps and figured out why: [4a63eea](https://github.com/MetaMask/core/pull/7001/commits/4a63eeab480fd87648bff99b33aa894db713376f)

rekmarks:
For posterity: we labeled this `no-changelog` to skip adding a changelog entry in `packages/eth-block-tracker` due to changes in `packages/eth-block-tracker/tests`.

rekmarks:
@metamaskbot publish-preview

PR #7005, opened 2025-10-30
Chore: fetch price api supported chain ids

sahar-fehri:
## Explanation

Price api already exposes its supported chainIds here https://price.api.cx.metamask.io/v1/supportedNetworks.
Instead of relying on the old static array; this array pulls the supported chainIds from the API directly.

This would make it possible in future to add support for new tokens without making a core relase.

## References

* Related to Extension: https://github.com/MetaMask/metamask-extension/pull/37416

## Checklist

---

> [!NOTE]
> Replaces the static supported-chain list with a dynamic fetch from the Price API, adds a cached `supportedChainIds` state to `TokenRatesController`, and updates APIs/types to use it.
> 
> - **Token Prices Service (`codefi-v2`)**:
>   - Add `fetchSupportedChainIds()` (uses `GET /v1/supportedNetworks`), return hex chain IDs.
>   - Remove exported `SUPPORTED_CHAIN_IDS` and `validateChainIdSupported`.
>   - Generalize types to `Hex` chain IDs; keep `getNativeTokenAddress`.
> - **TokenRatesController**:
>   - Add `supportedChainIds` cache in state with `timestamp` and `DEFAULT_CACHE_REFRESH_THRESHOLD` (24h).
>   - Validate chains against cached `supportedChainIds`; auto-refresh cache when stale.
>   - Update default state/metadata and polling/update flows accordingly.
> - **Utils (`assetsUtil`)**:
>   - `fetchTokenContractExchangeRates` now requires `supportedChainIds` arg and uses it for chain validation.
> - **Exports**:
>   - Remove `SUPPORTED_CHAIN_IDS` from public exports.
> - **Tests/Docs**:
>   - Update tests across packages and changelog to reflect new API and behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8315c2ecc061b6211034509540d1c24aa9648db3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sahar-fehri:
@metamaskbot publish-preview

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -59,6 +59,8 @@ export type Token = {
 
 const DEFAULT_INTERVAL = 180000;
 
+export const DEFAULT_CACHE_REFRESH_THRESHOLD = 1000 * 60 * 60 * 24; // 24 hours
```
sahar-fehri, line 62: I'm caching this supported chains data because its not the kind of data that will change very often.

eslint-warning-thresholds.json
```diff
@@ -80,7 +80,7 @@
     "@typescript-eslint/prefer-promise-reject-errors": 2
   },
   "packages/assets-controllers/src/token-prices-service/codefi-v2.ts": {
-    "jsdoc/tag-lines": 2
+    "jsdoc/tag-lines": 1
```
Prithpal-Sooriya, line 83: Nice cleanup!
There's only 1 more warning left on this file... can we also clean it up? If it's difficult happy to leave it.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -208,6 +215,12 @@ const tokenRatesControllerMetadata: StateMetadata<TokenRatesControllerState> = {
     includeInDebugSnapshot: false,
     usedInUi: true,
   },
+  supportedChainIds: {
+    includeInStateLogs: false,
+    persist: true,
```
bergarces, line 220: What is the benefit of making it part of the state and persisted? The client seems to just need so that it can pass it back to fetch the token rates, which is a round-trip and a synchronisation issue that we could avoid.

Couldn't we just keep it as part of the token prices service state? It can be initialised empty, fetched on the first load and then reused afterwards. That would also mean we don't need to use a cache, it will just get fetched whenever we start the client and remain there until it's restarted.

What do you think?

sahar-fehri: Could you elaborate more on "Make it as part of the token prices service state" 🙏  Are you here referencing the `codefi-v2.ts` service?

bergarces: Ok, I've just seen that we actually need the list of chains because we are fetching historical prices directly from the client... Why don't we move that call to the price service then?

Seems like a bit of overkill having to orchestrate new state just for that reason.

sahar-fehri: We fetch historical prices and exchangeRates (in some cases) directly from clients.
Again, if we move those calls to the price service; how would it solve the issue? I think the services we have do not have state; usually the controller is what saves the state correct?

Also this was not done only for clients, the polling we have on `tokenRatesController` (every 3mins) would use the cached value of the supported chainIds instead of making a call to the api every time

bergarces: > Could you elaborate more on "Make it as part of the token prices service state" 🙏 Are you here referencing the codefi-v2.ts service?

Yes, instead of an exported constant `SUPPORTED_CHAIN_IDS`, make the `supportedChainIds` a property of `CodefiTokenPricesServiceV2`. Gets initialised as `undefined` or `[]` at constructor time then, when trying to access it, fetch it if empty and never fetch it again.

Whenever the wallet restarts and the service gets initialised again, it will fetch again only once.

Since it will be part of the the class state, you won't have to pass it to every single time to `fetchTokenContractExchangeRates` and `fetchTokenExchangeRates`.

Although on second look, it does look like we instantiate a new `CodefiTokenPricesServiceV2` every time we call any of those methods (which is crazy, but I don't know how big a refactor it would be to change it).

PR #7016, opened 2025-10-31
feat: support adding non-evm tokens

juanmigdr:
## Explanation
The feature to hide tokens is now available for NON-EVM assets but there is no way to bring the assets back. This task involves implementing a token import feature for non-EVM assets. This is crucial for improving user experience by allowing users to import. The implementation should be prioritized to align with upcoming Solana campaigns.

Also this PR solves an existing issue where ignored tokens cannot be added back through the "Custom" tokens flow:

https://github.com/user-attachments/assets/a7ff4650-c9e7-4aff-b662-52941e3de5f1

## References
https://consensyssoftware.atlassian.net/browse/ASSETS-1425

## Checklist

---

> [!NOTE]
> Adds `MultichainAssetsController.addAssets` for non‑EVM assets, ensures balances default to 0 for newly added assets without snap balances, and fixes case-insensitive ignored token re-add; updates tests and changelog.
> 
> - **MultichainAssetsController**:
>   - **New**: `addAssets` action/method to add multiple non‑EVM `CaipAssetType` assets; validates same-chain, refreshes metadata, updates `state.accountsAssets`, removes from `allIgnoredAssets`, and publishes `accountAssetListUpdated`.
>   - **Wiring**: Registers `MultichainAssetsController:addAssets` handler; updates action types.
>   - **Tests**: Covers single/multiple adds, duplicates, ignored removal, cross-chain error, empty input, and event emissions.
> - **MultichainBalancesController**:
>   - **Change**: On `accountAssetListUpdated`, sets balance to `{ amount: '0', unit: '' }` for newly added assets missing from snap response.
>   - **Tests**: Verifies zero-balance behavior and existing flows.
> - **TokensController**:
>   - **Fix**: Clear `allIgnoredTokens` when re‑adding tokens even if address case differs; normalize with `toChecksumHexAddress` for dedupe/filtering in `addTokens`.
>   - **Tests**: Ensure re‑add clears ignored list with differing case.
> - **ERC1155Standard tests**:
>   - Add scenarios for optional params, IPFS gateway conversion, and network error handling.
> - **Changelog**:
>   - Notes new `addAssets` API and the case-insensitive ignored token fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b72b0d809e6cee7a656aa91d3938dfc70d12415. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/Standards/NftStandards/ERC1155/ERC1155Standard.test.ts
+++ b/packages/assets-controllers/src/Standards/NftStandards/ERC1155/ERC1155Standard.test.ts
@@ -331,137 +331,94 @@ describe('ERC1155Standard', () => {
     });
   });
 
-  describe('getDetails error handling', () => {
-    it('should throw error for non-ERC1155 contract', async () => {
-      // Mock failed ERC1155 interface check
-      nock('https://mainnet.infura.io:443', { encodedQueryParams: true })
-        .post('/v3/341eacb578dd44a1a049cbc5f6fd4035')
-        .reply(200, {
-          jsonrpc: '2.0',
-          id: 1,
-          result:
-            '0x0000000000000000000000000000000000000000000000000000000000000000',
-        });
+  describe('getDetails optional parameters', () => {
+    it('should return details without tokenURI when no tokenId is provided', async () => {
+      // Mock successful ERC1155 interface check
+      jest
+        .spyOn(erc1155Standard, 'contractSupportsBase1155Interface')
+        .mockResolvedValue(true);
+      jest.spyOn(erc1155Standard, 'getAssetSymbol').mockResolvedValue('TEST');
+      jest
+        .spyOn(erc1155Standard, 'getAssetName')
+        .mockResolvedValue('Test Token');
 
-      await expect(
-        erc1155Standard.getDetails(ERC1155_ADDRESS, 'https://ipfs.gateway.com'),
-      ).rejects.toThrow("This isn't a valid ERC1155 contract");
+      const ipfsGateway = 'https://ipfs.gateway.com';
+      const details = await erc1155Standard.getDetails(
+        ERC1155_ADDRESS,
+        ipfsGateway,
+        // No tokenId parameter to test the optional parameter behavior
+      );
+
+      expect(details.standard).toBe('ERC1155');
+      expect(details.tokenURI).toBeUndefined(); // Should be undefined when no tokenId
+      expect(details.symbol).toBe('TEST');
+      expect(details.name).toBe('Test Token');
+
+      // Restore original methods
+      jest.restoreAllMocks();
     });
 
-    it('should handle JSON parsing errors gracefully', async () => {
+    it('should convert IPFS URIs to gateway URLs', async () => {
       // Mock successful ERC1155 interface check
-      nock('https://mainnet.infura.io:443', { encodedQueryParams: true })
-        .post('/v3/341eacb578dd44a1a049cbc5f6fd4035')
-        .reply(200, {
-          jsonrpc: '2.0',
-          id: 1,
-          result:
-            '0x0000000000000000000000000000000000000000000000000000000000000001',
-        })
-        .persist();
+      jest
+        .spyOn(erc1155Standard, 'contractSupportsBase1155Interface')
+        .mockResolvedValue(true);
+      jest.spyOn(erc1155Standard, 'getAssetSymbol').mockResolvedValue('TEST');
+      jest
+        .spyOn(erc1155Standard, 'getAssetName')
+        .mockResolvedValue('Test Token');
+      // Mock getTokenURI to return IPFS URI
+      jest
+        .spyOn(erc1155Standard, 'getTokenURI')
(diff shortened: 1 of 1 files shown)


packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
```diff
@@ -296,6 +307,59 @@ export class MultichainAssetsController extends BaseController<
...
+      this.messenger.publish(`${controllerName}:accountAssetListUpdated`, {
+        assets: {
+          [accountId]: {
+            added: [assetId],
+            removed: [],
+          },
+        },
+      });
+
+      return this.state.accountsAssets[accountId] || [];
```
bergarces, line 382: Will this asset show in the list?

The selectors will not show any non-evm asset that does not have metadata: https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/selectors/token-selectors.ts#L358

bergarces: To elaborate here, the reason for filtering assets without metadata is because, for non-evm, the number of decimals is stored in that part of the state, and it is linked to the asset balance by the symbol.

Without balance and decimals, we cannot show the asset in the selector, as we do not have a raw balance property that flows like send can use.

juanmigdr: It will show because I triggered this:
```
      this.messenger.publish(`${controllerName}:accountAssetListUpdated`, {
        assets: {
          [accountId]: {
            added: [assetId],
            removed: [],
          },
        },
      });
```

Which triggers this in MultichainBalancesController
```
state.balances[accountId][assetId] = { amount: '0', unit: '' };
```

But maybe i am not understanding the question

packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
```diff
@@ -296,6 +307,59 @@ export class MultichainAssetsController extends BaseController<
...
 
+  /**
+   * Adds a single asset to the stored asset list for a specific account.
+   *
+   * @param assetId - The CAIP asset ID to add.
+   * @param accountId - The account ID to add the asset to.
+   * @returns The updated asset list for the account.
+   */
+  async addAsset(
+    assetId: CaipAssetType,
```
salimtb: this should be an array of assetIds since we want to support selecting and adding multiple assets. I also think the function name should be addAssets.

packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
juanmigdr: Regarding `multichainAssetsMetadata` the asset is added to metadata on line 323:
```
      await this.#refreshAssetsMetadata([assetId]);
```

packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
juanmigdr: In the current implementation in mobile assets are added in a loop 1 by 1. I can change the implementation. Should I also rename the function to `addAssets`? Or should we have separate functions and if so in which case do we need addAsset (singular)

Prithpal-Sooriya: Good call.
If we can add multiple at the controller level, this _should_ reduce state updates and computations. (Update all at once instead of one at a time).

In favour of renaming this if not yet used on clients.

juanmigdr: Done

packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
```diff
@@ -108,7 +113,8 @@ export type MultichainAssetsControllerStateChangeEvent =
 export type MultichainAssetsControllerActions =
   | MultichainAssetsControllerGetStateAction
   | MultichainAssetsControllerGetAssetMetadataAction
-  | MultichainAssetsControllerIgnoreAssetsAction;
+  | MultichainAssetsControllerIgnoreAssetsAction
+  | MultichainAssetsControllerAddAssetsAction;
```
bergarces, line 117: I'm assuming that adding a new action for a controller does not require any changes to the messengers on the clients. If so, it's fine not to mark it as breaking.

bergarces (approved):

Prithpal-Sooriya (approved):

packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
juanmigdr: Yu, not breaking!

Prithpal-Sooriya (approved):

PR #7017, opened 2025-10-31
Feat(analytics): new analytics controller [Phase 1.1]

NicolasMassart:
[Main](https://github.com/MetaMask/core/actions/workflows/main.yml)

## Explanation

New `@metamask/analytics-controller` package for unified, platform-agnostic analytics tracking controller for MetaMask.
This controller is designed for both the extension and mobile.

### Changes

- new analytics-controller package
- initial setup of the package files
- initial unit tests
- initial documentation

> [!IMPORTANT]
> The adapter type is work in progress and will be updated as we
> integrate with the new analytics system on mobile.
> We have this draft type to help us iterate on the implementation.
> It will be updated with proper types as we create the mobile adapter
> And the controller package will be released only when this is completed.

### Mobile side tests

Implemented the controller as a local resolution with `"@metamask/analytics-controller@^1.0.0": "file:../core/packages/analytics-controller",
`
and all the init process.
It loads well.
image

Mobile not ready yet to PR as controller is not published.

## References

fixes https://github.com/MetaMask/metamask-mobile/issues/21854
Mobile PR https://github.com/MetaMask/metamask-mobile/pull/22076

## Checklist

---

> [!NOTE]
> Introduces `@metamask/analytics-controller` with event/page tracking, user identification, state/messenger integration, tests/docs, and repo wiring.
> 
> - **New Package: `@metamask/analytics-controller`**
>   - **Controller**: `AnalyticsController` with methods `trackEvent`, `identify`, `trackPage`, `enable/disable`, `optIn/optOut`; default state via `getDefaultAnalyticsControllerState()` (UUIDv4 `analyticsId`).
>   - **Types/Adapters**: `AnalyticsPlatformAdapter` and action types (`AnalyticsController-*-action-types`).
>   - **Messaging**: Registers method action handlers under `AnalyticsController:*`.
>   - **Tooling**: `package.json`, `jest.config.js` (100% coverage thresholds), `tsconfig`/build configs, `typedoc`, `CHANGELOG.md`, licenses, README with usage.
>   - **Tests**: Unit tests for state init, methods, and messenger actions.
> - **Monorepo Integration**:
>   - Add package to `README.md` (list + mermaid graph), `.github/CODEOWNERS`, `tsconfig.json`, `tsconfig.build.json`, and `yarn.lock`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a631fb02816d2a36713e4c4118e8f9ff7017467e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/analytics-controller/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@metamask/analytics-controller",
+  "version": "1.0.0",
```
mcmire: The release automation detects changes to package versions to know what to release. If you initialize this version to 1.0.0 now, then when it is time to release this package, there will be no changes to the version, and you will be unable to make a release.

To fix this, can you set this to 0.0.0? Then you can set it to 1.0.0 when you want to release the package.

``` suggestion
  "version": "0.0.0",
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,201 @@
...
+    this.#platformAdapter = platformAdapter;
+
+    // Register action handlers
+    this.#registerActionHandlers();
+  }
+
+  /**
+   * Register action handlers for messenger system
+   */
+  #registerActionHandlers(): void {
```
mcmire: A little while ago we added a way to automatically expose methods as actions. Here are some instructions:

- Add `MESSENGER_EXPOSED_METHODS` somewhere in your controller file, e.g.: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L91
- Run `yarn generate-method-action-types`. This will generate a new file, `AnalyticsController-method-action-types.ts`.
- Import `AnalyticsControllerMethodActions` from this file.
- Add this type to `AnalyticsControllerActions`, e.g.: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L106
- Call `this.messenger.registerMethodActionHandlers` in your controller constructor: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L256

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,201 @@
...
+      this.#platformAdapter.trackPage(pageName, properties);
+    }
+  }
+
+  /**
+   * Set the enabled state.
+   *
+   * @param enabled - Whether analytics tracking is enabled (default: true)
+   */
+  setEnabled(enabled: boolean = true): void {
```
mcmire: Since these implementations are so small, what are your thoughts on splitting these methods in to `enable`, `disable`, `optIn`, and `optOut`? The idea here is that methods should not be thought of as mere setters or getters but actions.

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,201 @@
...
+    properties?: AnalyticsEventProperties,
+  ): void {
+    if (!this.state.enabled) {
+      return;
+    }
+
+    // Delegate to platform adapter if supported
+    if (this.#platformAdapter.trackPage) {
+      this.#platformAdapter.trackPage(pageName, properties);
+    }
```
mcmire, line 273: Nit: 

``` suggestion
      this.#platformAdapter.trackPage?.(pageName, properties);
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    // Update state with analytics ID
+    this.update((state) => {
+      state.analyticsId = userId;
+    });
+
+    // Delegate to platform adapter if supported
+    if (this.#platformAdapter.identify) {
+      this.#platformAdapter.identify(userId, traits);
+    }
```
mcmire, line 256: Nit:

``` suggestion
      this.#platformAdapter.identify?.(userId, traits);
```

packages/analytics-controller/src/AnalyticsController.ts
mcmire, line 1: It seems like this is largely a wrapper around another interface. Does this need to be a controller? Does the enabled/disabled state really need to be persisted?

What does `optedIn` end up doing?

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  } = options;
+
+  // Create default mock adapter if not provided
+  const adapter =
+    platformAdapter ??
+    ({
+      trackEvent: jest.fn(),
+      identify: jest.fn(),
+      trackPage: jest.fn(),
+    } as PlatformAdapter);
```
mcmire: Why use a type assertion? This interface is so small, it should be easy to mock.

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  messenger: AnalyticsControllerMessenger;
+};
+
+/**
+ * Builds an AnalyticsController based on the given options.
+ *
+ * @param options - Controller options
+ * @returns The controller and messenger
+ */
+function withController(options: WithControllerOptions = {}): WithControllerReturn {
```
mcmire: The "with" in the name of this helper implies that this is a wrapper around a test. We do use `withController` in other controller tests, but the usual reason is that some controllers have a `destroy` method and the wrapper allows you to initialize the controller in a test and then automatically destroy it (to clean up timers, etc.). You can see an example of how this is often used in the `sample-controllers` test suite: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.test.ts#L58

However, that doesn't seem to be how you've written this helper. This could be confusing when working with this controller in conjunction with other controllers. What are your thoughts on either having the helper take a function, or renaming it to `setupController`?

mcmire (commented):
I know this is in draft but I couldn't help but review this now :)

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: it's for now, but that's because it's just the begining of the implem.
It will include more logic

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: Thanks, I will check that

packages/analytics-controller/LICENSE.MIT
```diff
@@ -0,0 +1,20 @@
+MIT License
+
+Copyright (c) 2024 MetaMask
```
Mrtenz:
```suggestion
Copyright (c) 2025 MetaMask
```

packages/analytics-controller/LICENSE.APACHE2
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2024 MetaMask
```
Mrtenz:
```suggestion
   Copyright 2025 MetaMask
```

.github/CODEOWNERS
```diff
@@ -49,6 +49,7 @@
 
 ## Mobile Platform Team
 /packages/app-metadata-controller @MetaMask/mobile-platform
+/packages/analytics-controller    @MetaMask/mobile-platform
```
GuillaumeRx: If this controller is supposed to be shared across clients shouldn't we also add extension platform ?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,320 @@
...
+    // Use analyticsId from state if provided, otherwise use the one from defaultState
+    const analyticsId: string =
+      state.analyticsId !== undefined
+        ? state.analyticsId
+        : defaultState.analyticsId;
+    const mergedState: AnalyticsControllerState = {
+      ...defaultState,
+      ...state,
+      analyticsId,
+    };
```
GuillaumeRx: Why do we have to extract the `analyticsId` ? I would expect the spread to take the id from `state` if it exists or  use the one from `defaultState` if it's undefined 🤔 

```suggestion
    const mergedState: AnalyticsControllerState = {
      ...defaultState,
      ...state,
    };
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,320 @@
...
+    eventName: string,
+    properties: AnalyticsEventProperties = {},
+  ): void {
+    // Don't track if analytics is disabled
+    if (!this.state.enabled) {
+      return;
+    }
+
+    // Delegate to platform adapter
+    this.#platformAdapter.trackEvent(eventName, properties);
```
GuillaumeRx, line 234: Can't we get closer to the segment adapter itself rather than having another layer of abstraction on top ?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,320 @@
...
+    this.messenger.registerMethodActionHandlers(
+      this,
+      MESSENGER_EXPOSED_METHODS,
+    );
+
+    projectLogger('AnalyticsController initialized and ready', {
+      enabled: this.state.enabled,
+      optedIn: this.state.optedIn,
+      analyticsId: this.state.analyticsId,
+    });
```
GuillaumeRx, line 213: We shouldn't be logging the state and especially the id

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: The goal is to have this controller usable on both mobile and extension, the adapter is platform dependent and has to be implemented on each platform

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: The enabled and optedIn status and id are not a private info. Logging it only happens in dev mode and will be the only way for devs to know their id and debut their events.

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: It allows to override the id on controller init. This is because we will have to migrate ids that are not yet handled by the controller so we have to pass the one we currently have in the mobile/extension state.

GuillaumeRx: That's what the spread overwrite will do by default I think

GuillaumeRx: But usually we do those kinds of operations with migrations on the client side :thinking:

packages/analytics-controller/src/AnalyticsController.ts
GuillaumeRx: Fair enough

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: fixed

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -0,0 +1,313 @@
...
+      ...getDefaultAnalyticsControllerState(),
+      ...state,
+    };
+
+    super({
+      name: controllerName,
+      metadata: analyticsControllerMetadata,
+      state: mergedState,
+      messenger,
+    });
```
GuillaumeRx: nit
```suggestion  
    super({
      name: controllerName,
      metadata: analyticsControllerMetadata,
      state: {
        ...getDefaultAnalyticsControllerState(),
        ...state,
      };,
      messenger,
    });
```

packages/analytics-controller/src/AnalyticsController.ts
GuillaumeRx: My thinking is that we shouldn't need three layers of abstraction around Segment in the clients.

Today we have custom classes created to interface with Segment:
- https://github.com/MetaMask/metamask-extension/blob/d4ef57f56992fac2579765a01c34b7f4dda5e2b0/app/scripts/lib/segment/analytics.js
- https://github.com/MetaMask/metamask-mobile/blob/1c7b545ec2a6c713129aae9ce1cebf9fbbeddef5/app/core/Analytics/MetaMetrics.ts#L591

I think it would be easier if we could remove the need of another layer of abstraction on top of these utils by just using the Segment classes directly, unifying the signature of them rather than adding another class to wrap them.

NicolasMassart: see https://consensyssoftware.atlassian.net/jira/polaris/projects/MWMR/ideas/view/8845679?selectedIssue=MWMR-23 for a global view of the refactoring project

PR #7040, opened 2025-11-03
fix: Fix build script not working because of missing `@ts-bridge/cli` dependency

### LINKED ISSUE #4960: Package-specific `build` scripts are broken
Each package has a `build` script, but this scripts currently fail with the error `command not found: ts-bridge`.

The top-level build command still works, but on a per-package level the `ts-bridge` binary is not present. We should fix this by adding `ts-bridge` to `devDependencies` for all packages. The package would still get hoisted so that we only have one instance, but this would ensure `ts-brigge` gets listed in `node_modules/.bin` for all packages so that it can be invoked by package npm scripts.

Mrtenz:
## Explanation

This fixes the `build` script for all packages by adding `@ts-bridge/cli` as `devDependency`, and includes an additional `build:all` script to build packages including their dependencies. This is a bit slower than building one package, so I didn't set it as default.

## References

Closes #4960.

## Checklist

---

> [!NOTE]
> Adds @ts-bridge/cli as a devDependency and introduces a consistent `build:all` script across packages, with constraints and formatting updates.
> 
> - **Build/Tooling**:
>   - Add `@ts-bridge/cli` as a `devDependency` across workspace packages.
>   - Introduce `scripts.build:all` (`ts-bridge ... --clean`) in all non-root packages and enforce via `yarn.config.cjs` constraints.
>   - Minor script tidy-ups (e.g., reorder or add `since-latest-release`, ensure `publish:preview` presence where required).
> - **Formatting**:
>   - Add Prettier plugin `prettier-plugin-packagejson` in `.prettierrc.js`.
> - **Dependencies**:
>   - Update `yarn.lock` to reflect new dev dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9c4a1c32d1e12f45b26e3a9f7d91be231e91b4fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,14 +1,12 @@
 /**
  * @type {import('prettier').Options}
  */
-// All of these are defaults except singleQuote, but we specify them
-// for explicitness
-const config = {
+module.exports = {
+  // All of these are defaults except singleQuote, but we specify them
+  // for explicitness
   quoteProps: 'as-needed',
   singleQuote: true,
   tabWidth: 2,
   trailingComma: 'all',
   plugins: ['prettier-plugin-packagejson'],
 };
-
-export default config;


.prettierrc.js
```diff
@@ -8,4 +8,5 @@ module.exports = {
   singleQuote: true,
   tabWidth: 2,
   trailingComma: 'all',
+  plugins: ['prettier-plugin-packagejson'],
```
Mrtenz, line 11: Not strictly required for this PR, but I noticed Prettier wasn't updating the order of scripts in `package.json`. Turns out we never added `prettier-plugin-packagejson` to the config after bumping to Prettier 3.

cryptodev-2s:
@Mrtenz I am wondering when are we supposed to use this new script ?

Mrtenz:
> @Mrtenz I am wondering when are we supposed to use this new script ?

@cryptodev-2s This is mainly for local development. `build` can be used to just build a single package, and `build:all` to build a package and only its dependencies. This is a bit quicker than building everything in the repository (especially as we're adding more packages).

.prettierrc.js
mcmire: Yikes, this is my fault, thanks for adding this.

PR #6713, opened 2025-09-24
New Action on the GatorPermissionsController to Allow the Consumer to SubmitRevocation

hanzel98:
## Explanation

**What is the current state of things and why does it need to change?**

The GatorPermissionsController currently provides functionality to fetch, enable, disable, and decode gator permissions, but it lacks the ability to revoke permissions. This missing capability prevents MetaMask clients from providing users with a complete permission management experience where they can not only grant permissions but also revoke them when needed.

**What is the solution your changes offer and how does it work?**

This PR adds a new `submitRevocation` action to the GatorPermissionsController that allows consumers (MetaMask clients) to submit permission revocations through the gator permissions provider snap. The implementation:

- Adds a new `RevocationParams` type that accepts a `delegationHash` (hex string) to identify the permission to revoke
- Implements a `submitRevocation` method that forwards the revocation request to the gator permissions provider snap using the `permissionsProvider_submitRevocation` RPC method
- Includes proper error handling using existing error types (`GatorPermissionsNotEnabledError` and `GatorPermissionsProviderError`)
- Added new filter in the function to request the permissions
- Follows the established patterns in the controller for action registration and snap communication

**Are there any changes whose purpose might not obvious to those unfamiliar with the domain?**

The `delegationHash` parameter is a unique identifier for ERC-7715 delegations/permissions that allows the snap to locate and revoke the specific permission. This hash-based approach is standard in the ERC-7715 delegation framework for permission management.

## References

[[Related to ]](https://app.zenhub.com/workspaces/readable-permissions-67982ce51eb4360029b2c1a1/issues/gh/metamask/delegator-readable-permissions/357)
This PR depends on this other PR of the permissions-controller https://github.com/MetaMask/core/pull/6713

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

**Note**: The changelog item should be checked off after updating the CHANGELOG.md file, and the last item can be checked as N/A since this is a new feature addition with no breaking changes.

---

> [!NOTE]
[...]

PR #7055, opened 2025-11-04
feat: Expose list of pending revocations in state

V00D00-child:
## Explanation

Expose the list of pending revocations in the state to allow MM clients to create custom selectors that filter on `pendingRevocations` to determine whether the revoke CTA should show the "pending" status.

- A revocation is added to `pendingRevocations` state on calls to `addPendingRevocation()`
- A revocation is removed from `pendingRevocations` state given a `transactionId` when the handler's `cleanup()` function is executed. 

## References

Forked from #6713

## Checklist

---

> [!NOTE]
> Adds `pendingRevocations` to controller state with lifecycle management via `addPendingRevocation` and `submitRevocation`, updating metadata, tests, and changelog.
> 
> - **GatorPermissionsController**:
>   - **State**: Add `pendingRevocations: { txId: string; permissionContext: Hex }[]` with default `[]` and UI exposure in metadata.
>   - **API**:
>     - New getter `pendingRevocations`.
>     - `addPendingRevocation` now appends to state immediately and cleans up on confirm/fail/drop/timeout by `txId`.
>     - `submitRevocation` submits to Snap and removes entry by `permissionContext` on success.
>   - **Internals**: Add private helpers to add/remove pending revocations; import `Hex` type.
> - **Tests**: Update snapshots, add coverage for pending revocations lifecycle and getter.
> - **Docs**: Update `CHANGELOG.md` to note exposed pending revocations and related actions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 77e676eb10a151ee9e920574ce85918148666246. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/gator-permissions-controller/src/GatorPermissionsController.ts
+++ b/packages/gator-permissions-controller/src/GatorPermissionsController.ts
@@ -841,8 +841,6 @@ export default class GatorPermissionsController extends BaseController<
           permissionContext,
         });
 
-        cleanup(transactionMeta.id);
-
         this.submitRevocation({ permissionContext }).catch((error) => {
           controllerLog(
             'Failed to submit revocation after transaction confirmed',
@@ -853,6 +851,8 @@ export default class GatorPermissionsController extends BaseController<
             },
           );
         });
+
+        cleanup(transactionMeta.id);
       }
     };
 


packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -104,6 +104,14 @@ export type GatorPermissionsControllerState = {
    * Default value is `@metamask/gator-permissions-snap`
    */
   gatorPermissionsProviderSnapId: SnapId;
+
+  /**
+   * List of gator permission pending a revocation transaction
+   */
+  pendingRevocations: {
+    txId: string;
```
hanzel98, line 112: We need to consider that some revocations won't have a transaction involved but we would still need to block a user from clicking multiple times. I don't know if that would affect this

V00D00-child: For these cases, we can lean on the `permissionsContext` that should always be defined if there is an empty txId

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -536,6 +568,15 @@ export default class GatorPermissionsController extends BaseController<
     });
   }
 
+  /**
```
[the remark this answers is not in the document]
V00D00-child: The reported memory leak assumes that disableGatorPermissions() should explicitly clean up all pending revocations, including their event listeners and timeouts.

However, the event handlers are defined within the scope of addPendingRevocation(), and we do not maintain external references to those listener functions. As a result, disableGatorPermissions() cannot directly unsubscribe them.

Additionally, addPendingRevocation() already uses a setTimeout as a safety net to ensure that any pending revocation is eventually cleaned up, even if permissions are disabled before the timeout expires. This mechanism prevents memory leaks by releasing resources once the timeout triggers.

It might still be worth reviewing whether we can improve the design by tracking listener references for more explicit cleanup, but as currently implemented, this behavior is intentional and not an actual leak.

PR #7067, opened 2025-11-06
chore: change `Monad Testnet` to infura network

stanleyyconsensys:
## Explanation
This PR migrates the Monad Testnet from a custom network to an Infura-supported network. The monad-testnet network configuration is moved from CustomNetworkType to InfuraNetworkType, with the custom entry deprecated.

- Moved monad-testnet from CustomNetworkType to InfuraNetworkType
- Deprecated the CustomNetworkType['monad-testnet'] entry with a JSDoc comment
- Removed monad-testnet from custom network configuration initialization
- Updated test expectations to reflect monad-testnet as an Infura network type

## Checklist

---

> [!NOTE]
> Migrates Monad Testnet from custom to Infura network type, updates default networks/changelogs, and adjusts tests and chain lists to include chainId `0x279f`.
> 
> - **Network & Types**:
>   - Add `monad-testnet` to `InfuraNetworkType`; mark `CustomNetworkType['monad-testnet']` as deprecated (`packages/controller-utils/src/types.ts`).
>   - Update default Infura networks to include Monad Testnet; expose Infura endpoint/config details in tests.
> - **Changelogs**:
>   - `controller-utils`: Added `monad-testnet` to `InfuraNetworkType`.
>   - `network-controller`: Changed defaults to include Monad Testnet.
> - **Tests**:
>   - `network-controller/tests/NetworkController.test.ts`: Add network config/client for chainId `0x279f` (Monad Testnet) with Infura endpoint.
>   - `assets-controllers/TokenDetectionController.test.ts`: Include `0x279f` in supported `chainIds` for detection flows.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05b5bc1aee5ccb0656464f3a9e5e1e099740e563. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "700.0.0",
+  "version": "701.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/subscription-controller/CHANGELOG.md
+++ b/packages/subscription-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [5.2.0]
+
 ### Added
 
 - Added `minBillingCyclesForBalance` property to `ProductPrice` type ([#7269](https://github.com/MetaMask/core/pull/7269))
@@ -226,7 +228,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `@metamask/controller-utils` from `^11.12.0` to `^11.14.0` ([#6620](https://github.com/MetaMask/core/pull/6620), [#6629](https://github.com/MetaMask/core/pull/6629))
 - Bump `@metamask/utils` from `^11.4.2` to `^11.8.0` ([#6588](https://github.com/MetaMask/core/pull/6588))
 
-[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@5.1.0...HEAD
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@5.2.0...HEAD
+[5.2.0]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@5.1.0...@metamask/subscription-controller@5.2.0
 [5.1.0]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@5.0.0...@metamask/subscription-controller@5.1.0
 [5.0.0]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@4.2.2...@metamask/subscription-controller@5.0.0
 [4.2.2]: https://github.com/MetaMask/core/compare/@metamask/subscription-controller@4.2.1...@metamask/subscription-controller@4.2.2
--- a/packages/subscription-controller/package.json
+++ b/packages/subscription-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/subscription-controller",
-  "version": "5.1.0",
+  "version": "5.2.0",
   "description": "Handle user subscription",
   "keywords": [
     "MetaMask",


packages/controller-utils/src/types.ts
```diff
@@ -24,6 +25,9 @@ export type InfuraNetworkType =
  */
 export const CustomNetworkType = {
   'megaeth-testnet': 'megaeth-testnet',
+  /**
```
stanleyyconsensys, line 28: we dont remove monad here to avoid breaking change

as controller utils is share everywhere

stanleyyconsensys:
@metamaskbot publish-preview

stanleyyconsensys:
@metamaskbot publish preview

packages/controller-utils/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Change Monad Testnet as infura network ([#7067](https://github.com/MetaMask/core/pull/7067))
+  - Add `monad-testnet` to `InfuraNetworkType`
```
mcmire: Nit: What are your thoughts on combining these two? They are essentially saying the same thing.

``` suggestion
- Add `monad-testnet` to `InfuraNetworkType` ([#7067](https://github.com/MetaMask/core/pull/7067))
```

Also what are your thoughts on adding this to Added rather than Changed, since this extends the existing API?

packages/network-controller/tests/NetworkController.test.ts
```diff
@@ -511,6 +511,21 @@ describe('NetworkController', () => {
                   },
                 ],
               },
+              "0x279f": Object {
```
mcmire, line 510: Since this affects the default set of networks for NetworkController, what are your thoughts on adding a changelog entry for this package? Perhaps we could add an entry to Changed such as:

``` markdown
### Changed

- Update the default set of Infura networks to include Monad Testnet ([#7067](https://github.com/MetaMask/core/pull/7067))
```

packages/assets-controllers/src/TokenDetectionController.test.ts
```diff
@@ -1192,6 +1192,7 @@ describe('TokenDetectionController', () => {
                 '0xa',
                 '0x89',
                 '0x531',
+                '0x279f',
```
mcmire, line 1196: It looks like changing `InfuraNetworkType` also changes the set of chains that TokenDetectionController changes, is this true? If so do you think we should update the changelog for this package similar to how we would want to update the NetworkController changelog?

``` markdown
### Changed

- Update the set of networks supported by TokenDetectionController to include Monad Testnet ([#7067](https://github.com/MetaMask/core/pull/7067))
```

packages/network-controller/tests/NetworkController.test.ts
stanleyyconsensys: i think we only need that when network controller bump up? otherwise it seem still remain the same?
shall we only do it when i do the bump up?

packages/assets-controllers/src/TokenDetectionController.test.ts
stanleyyconsensys: not really, it only impacted, becoz from the test it call "getDefaultNetworkControllerState" to get the network state to simulate all networks enabled, so the expected result has to inlcude monad-testnet

packages/network-controller/tests/NetworkController.test.ts
mcmire: We generally encourage updating changelogs as changes are made to packages, not at release time. As it stands your change affects NetworkController now, not when the release is created. (You can tell because the tests changed.) That's why I suggested updating the `network-controller` changelog in this PR rather than later.

packages/controller-utils/src/types.ts
mcmire: Yes, good call.

packages/assets-controllers/src/TokenDetectionController.test.ts
mcmire: Oh I gotcha. Makes sense.

packages/network-controller/tests/NetworkController.test.ts
cryptodev-2s: @stanleyyconsensys Could we add the changelog entry suggested above by @mcmire ?
These changes were made as part of the PR itself rather than through a release, and without updating the changelog now, we would later need to look back through past PRs to identify anything missing.

stanleyyconsensys: hi @cryptodev-2s just rebase and added

PR #7075, opened 2025-11-06
Ramps Controller

pkowalski:
### Current State:
Ramps currently depends on an sdk to handle all connections to the RAMPS api. We now have 2 different offerings which each use their own SDK. All of this routing logic is done on the mobile app. If this logic were to be extended to mobile or extension it would require rework on those codebases. 

### Why we want to change things:
Many other teams already use this controller design pattern. By moving towards this we can handle all business logic in 1 controller. It lets us expose Ramps data and apis to other experiences without those experiences requiring any knowledge of business logic. Lastly it would speed up any Ramps api integrations on extension and mobile in the future.

### Changes introduced:
This is the foundation for the Ramps Controller. It contains 1 function which updates its internal state which reflects eligibility for a ramps feature on a certain region. 

## References

* Related to [TRAM-2783](https://consensyssoftware.atlassian.net/browse/TRAM-2783)

## Checklist

[TRAM-2783]: https://consensyssoftware.atlassian.net/browse/TRAM-2783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> Introduces `@metamask/ramps-controller` with env-aware region eligibility fetching and wires it into the monorepo (configs, docs, ownership, build).
> 
> - **New package: `@metamask/ramps-controller`**
>   - Adds `RampsController` with state (`metamaskEnvironment`, `context`, `region`) and `getCountries` to fetch geolocation and region data via env-aware API URLs.
>   - Registers messenger action `RampsController:getCountries`; includes controller metadata.
>   - Includes full setup: tests, TypeScript build, Jest, Typedoc, README, CHANGELOG, LICENSE.
> - **Monorepo integration**
>   - Adds package to workspaces (`tsconfig.json`, `tsconfig.build.json`), README package list, and CODEOWNERS; assigns Ramps team.
>   - Updates README diagrams (adds `ramps_controller`; minor link additions).
>   - Updates LavaMoat allowScripts (adds `eslint-plugin-import-x>unrs-resolver`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cce1ac6a39f03a03f87f5f71aa5b56d9d0d50ac7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,8 @@
       "@keystonehq/bc-ur-registry-eth>hdkey>secp256k1": true,
       "babel-runtime>core-js": false,
       "simple-git-hooks": false,
-      "tsx>esbuild": false
+      "tsx>esbuild": false,
+      "eslint-plugin-import-x>unrs-resolver": false
     }
   }
 }


packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,215 @@
...
+};
+
+/**
+ * Ramps controller state
+ */
+export type RampsControllerState = {
+  // Environment tells us which API urls to us
+  metamaskEnvironment: string;
+  // Determines front end context (browser, mobile, etc)
+  context: string;
```
wachunei: We are using `Context.Browser` in default state, can we use `Context` type here or create our new `Context` type so we don't use  arbitrary`string`? Another options is to type the strings like `'staging' | 'production' | ...`

same for `metamaskEnvironment`, is there already a type for that?

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,215 @@
...
+      metadata: rampsControllerMetadata,
+      state: { ...defaultState, ...state } as RampsControllerState,
+    });
+
+    // Initialize the OnRampSDK
+    const environment = state?.metamaskEnvironment ?? SdkEnvironment.Staging;
+    const context = state?.context ?? Context.Browser;
+
+    // Initialize the Native Ramps SDK
+    const nativeEnv = getNativeSdkEnvironment(environment);
```
wachunei: passing `SdkEnvironment.Staging` to `getNativeSdkEnvironment` does not seem right, that case might be just working because of a coincidence.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,215 @@
...
+    super({
+      messenger,
+      name: controllerName,
+      metadata: rampsControllerMetadata,
+      state: { ...defaultState, ...state } as RampsControllerState,
+    });
+
+    // Initialize the OnRampSDK
+    const environment = state?.metamaskEnvironment ?? SdkEnvironment.Staging;
+    const context = state?.context ?? Context.Browser;
```
wachunei: As far as I'm seeing on the `RampsControllerState` there no reason to access state properties with `?.`, we should avoid that because when compiled might lead to unnecessary checks.

packages/ramps-controller/src/RampsController.ts
pkowalski: Fixed by removing mappings and using sdkEnvironment directly

packages/ramps-controller/src/RampsController.ts
pkowalski: Fixed

packages/ramps-controller/src/RampsController.ts
pkowalski: I set it to re-use Context instead. I think we can just reuse those enums and if need be in the future, re-declare them here. Just too many different enums for the same thing currently between rampsSDK and nativeSDK. Dont want to now add a third for the controller

packages/ramps-controller/LICENSE
```diff
@@ -0,0 +1,20 @@
...
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
```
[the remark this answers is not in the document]
pkowalski: Issue is with template, Not us. WIll update for release but don't want to block

PR #7080, opened 2025-11-06
feat(analytics-controller): update controller and adapter interface for mobile integration [Phase 1.4]

### LINKED ISSUE #7079: feat(analytics-controller): update controller and adapter interface for mobile integration [Phase 1.4]
Update analytics-controller and adapter interfaces to include changes required when implementing the mobile integration

## Technical Details

1. Remove `add(middleware: TPlugin): void` from `AnalyticsPlatformAdapter` interface
2. Add `onSetupCompleted(analyticsId: string): void` to `AnalyticsPlatformAdapter` interface  
3. Update `AnalyticsController` to call `onSetupCompleted` after initialization (only if `analyticsId` is set)
4. Add error handling (try-catch) - adapter failures shouldn't break controller
5. Update tests: verify hook called when `analyticsId` set, NOT called when not set

## Acceptance Criteria

- [ ] `add` method removed, `onSetupCompleted(analyticsId: string): void` added to `AnalyticsPlatformAdapter`
- [ ] `AnalyticsController` calls `onSetupCompleted` after init (only if `analyticsId` is set) with error handling
- [ ] Tests verify hook called when `analyticsId` set, NOT called when not set, error handling works
- [ ] Documentation and CHANGELOG updated

NicolasMassart:
> [!IMPORTANT]
> This is an unreleased package which will continue to be improved until ready for release.

### State & Methods

- Removed `enabled` field (computed via `selectEnabled` selector)
- Consolidated `optedInForRegularAccount` and `optedInForSocialAccount` into single `optedIn` property
- Consolidated four opt-in/opt-out methods into `optIn()` and `optOut()`
- Removed `enable()`/`disable()` methods
- Changed state persistence to `persist: false` (platform handles via `stateChange` event)

### API Changes

- `trackPage()` → `trackView()` (platform-agnostic)
- `trackEvent()` now accepts `AnalyticsTrackingEvent` object
- `identify()` uses `analyticsId` from state (removed `userId` parameter)
- Platform adapter: `trackEvent()` → `track()`, `trackPage()` → `view()`, `identify()` now required

### New Features

- Added `analyticsControllerSelectors` (`selectAnalyticsId`, `selectOptedIn`, `selectEnabled`)
- Added UUIDv4 validation for `analyticsId`
- Added `onSetupCompleted()` lifecycle hook for adapter setup
- Added `AnalyticsPlatformAdapterSetupError` for setup failures
- Removed `uuid` dependency (platform generates UUID)
- Added `anonymousEventsFeature` option for split event tracking that will soon be obsolete

### Types

- Added `AnalyticsTrackingEvent` and `AnalyticsUserTraits` types
- `getDefaultAnalyticsControllerState()` returns state without `analyticsId`

### Other

- Simplified validation error message
- Updated tests, action types, and README

## References

fixes #7079

## Checklist

---

> [!NOTE]
> Refactors analytics-controller to use platform-managed analyticsId with UUIDv4 validation, new adapter API and lifecycle hook, `trackView`, anonymous-events option, selectors, and updated tests/docs.
> 
> - **API/State (breaking)**:
>   - Remove `enabled`, `enable()`, `disable()`; tracking enabled via `selectEnabled` (derived from `optedIn`).
>   - Consolidate opt methods to `optIn()`/`optOut()` and state to `optedIn`.
>   - `analyticsId` must be provided by client platform; add UUIDv4 validation; controller no longer generates/persists it.
>   - `trackPage()` → `trackView(name, properties?)`.
>   - `trackEvent()` now takes `AnalyticsTrackingEvent` (regular + sensitive properties; supports anonymous split when enabled).
>   - `identify(traits?)` uses current `analyticsId` (no `userId` param).
> - **Adapter Interface (breaking)**:
>   - Rename `trackEvent()` → `track(eventName, properties?)`, `trackPage()` → `view(name, properties?)`.
>   - `identify(userId, traits?)` now required.
>   - Add `onSetupCompleted(analyticsId)` lifecycle hook; controller calls once on `init()` and swallows setup errors.
> - **State handling**:
>   - `getDefaultAnalyticsControllerState()` returns only opt-in defaults (no `analyticsId`).
>   - Controller persists only `optedIn`; `analyticsId` marked `persist: false` (platform-managed).
> - **New utilities**:
>   - Add `analyticsControllerSelectors` (`selectAnalyticsId`, `selectOptedIn`, `selectEnabled`).
>   - Add `AnalyticsPlatformAdapterSetupError` and state validator (`isValidUUIDv4`, `validateAnalyticsControllerState`).
>   - Remove `uuid` dependency.
> - **Docs/Tests**:
>   - Update README for platform-managed storage, lifecycle, anonymous-events feature.
>   - Overhaul tests to cover new API, validation, selectors, and lifecycle.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 02f6ddd6a8d27ea9965930e107dd426ec01ad79b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "720.0.0",
+  "version": "721.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/profile-metrics-controller/CHANGELOG.md
+++ b/packages/profile-metrics-controller/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING:** `ProfileMetricsController` now requires the `AccountsController:getState` action to be allowed ([#7471](https://github.com/MetaMask/core/pull/7471))
+  - The controller messenger does not require `AccountsController:listAccounts` action anymore.
+
+### Fixed
+
+- Collect EVM and non-EVM accounts during initial sync ([#7471](https://github.com/MetaMask/core/pull/7471))
+
 ## [1.1.0]
 
 ### Changed
--- a/packages/profile-metrics-controller/src/ProfileMetricsController.test.ts
+++ b/packages/profile-metrics-controller/src/ProfileMetricsController.test.ts
@@ -93,12 +93,19 @@ describe('ProfileMetricsController', () => {
               { options: { assertUserOptedIn: () => assertUserOptedIn } },
               async ({ controller, rootMessenger }) => {
                 rootMessenger.registerActionHandler(
-                  'AccountsController:listAccounts',
+                  'AccountsController:getState',
                   () => {
-                    return [
-                      createMockAccount('0xAccount1'),
-                      createMockAccount('0xAccount2', false),
-                    ];
+                    const account1 = createMockAccount('0xAccount1');
+                    const account2 = createMockAccount('0xAccount2', false);
+                    return {
+                      internalAccounts: {
+                        accounts: {
+                          [account1.id]: account1,
+                          [account2.id]: account2,
+                        },
+                        selectedAccount: account1.id,
+                      },
+                    };
                   },
                 );
 
@@ -132,12 +139,19 @@ describe('ProfileMetricsController', () => {
               },
               async ({ controller, rootMessenger }) => {
                 rootMessenger.registerActionHandler(
-                  'AccountsController:listAccounts',
+                  'AccountsController:getState',
                   () => {
-                    return [
-                      createMockAccount('0xAccount1'),
-                      createMockAccount('0xAccount2'),
-                    ];
+                    const account1 = createMockAccount('0xAccount1');
+                    const account2 = createMockAccount('0xAccount2');
+                    return {
+                      internalAccounts: {
+                        accounts: {
+                          [account1.id]: account1,
+                          [account2.id]: account2,
+                        },
+                        selectedAccount: account1.id,
+                      },
+                    };
                   },
                 );
 
@@ -543,7 +557,7 @@ function getMessenger(
   rootMessenger.delegate({
     messenger,
     actions: [
-      'AccountsController:listAccounts',
+      'AccountsController:getState',
       'ProfileMetricsService:submitMetrics',
     ],
     events: [
--- a/packages/profile-metrics-controller/src/ProfileMetricsController.ts
+++ b/packages/profile-metrics-controller/src/ProfileMetricsController.ts
@@ -1,7 +1,7 @@
 import type {
   AccountsControllerAccountAddedEvent,
-  AccountsControllerListAccountsAction,
   AccountsControllerAccountRemovedEvent,
+  AccountsControllerGetStateAction,
 } from '@metamask/accounts-controller';
 import type {
   ControllerGetStateAction,
@@ -100,7 +100,7 @@ export type ProfileMetricsControllerActions =
  */
 type AllowedActions =
   | ProfileMetricsServiceMethodActions
-  | AccountsControllerListAccountsAction;
+  | AccountsControllerGetStateAction;
 
 /**
  * Published when the state of {@link ProfileMetricsController} changes.
@@ -266,7 +266,10 @@ export class ProfileMetricsController extends StaticIntervalPollingController()<
         return;
       }
       const newGroupedAccounts = groupAccountsByEntropySourceId(
-        this.messenger.call('AccountsController:listAccounts'),
+        Object.values(
+          this.messenger.call('AccountsController:getState').internalAccounts
+            .accounts,
+        ),
       );
       this.update((state) => {
         for (const key of Object.keys(newGroupedAccounts)) {
--- a/packages/ramps-controller/CHANGELOG.md
+++ b/packages/ramps-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.0.0]
+
 ### Added
 
 - Initial release ([#7316](https://github.com/MetaMask/core/pull/7316))
@@ -14,4 +16,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Add `OnRampService` for interacting with the OnRamp API
   - Add geolocation detection via IP address lookup
 
-[Unreleased]: https://github.com/MetaMask/core/
+[Unreleased]: https://github.com/MetaMask/core/compare/@metamask/ramps-controller@1.0.0...HEAD
+[1.0.0]: https://github.com/MetaMask/core/releases/tag/@metamask/ramps-controller@1.0.0
--- a/packages/ramps-controller/package.json
+++ b/packages/ramps-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/ramps-controller",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "description": "A controller for managing cryptocurrency on/off ramps functionality",
   "keywords": [
     "MetaMask",


NicolasMassart:
@metamaskbot publish-preview

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -30,38 +37,38 @@ export const controllerName = 'AnalyticsController';
  */
 export type AnalyticsControllerState = {
   /**
-   * Whether analytics tracking is enabled
+   * User domain: Whether the user has opted in to analytics.
    */
-  enabled: boolean;
+  user_optedIn: boolean;
```
mcmire: I see that you've marked all of these properties as falling under the "user domain". Does this imply there will be more domains in the future?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
     }
 
-    // Delegate to platform adapter
-    this.#platformAdapter.trackEvent(eventName, properties);
+    // if event does not have properties, only send the non-anonymous empty event
+    // and return to prevent any additional processing
+    if (!event.hasProperties) {
+      this.#platformAdapter.track(event.name, {
+        anonymous: false,
+      } as AnalyticsEventProperties);
```
mcmire: Why are we using a type assertion here? Can we not expect that the platform adapter accepts the kinds of arguments that we want to pass in?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
     });
   }
 
   /**
-   * Opt in to analytics.
+   * Opt in to analytics through social account.
+   * This updates the user's social opt-in status.
    */
-  optIn(): void {
+  socialOptIn(): void {
```
mcmire: "Social opt out" sounds like a noun.

What do you think about `optInViaSocial`?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
+      state.user_socialOptedIn = false;
     });
   }
+
+  /**
+   * Get the analytics ID from the controller state.
+   *
+   * @returns The current analytics ID.
+   */
+  getAnalyticsId(): string {
```
mcmire: Instead of exposing state data through methods on the class (and/or messenger actions), what do you think about defining Redux selector functions instead? These selectors can then be used in the app.

For instance above the class or in a separate file, we could create these functions:

``` typescript
/**
 * @returns The analytics ID from the AnalyticsController state.
 */
export const selectAnalyticsId = (state: AnalyticsController) =>
  state.user_analyticsId;

/**
 * @returns Whether the user has opted in via a regular account.
 */
export const selectIsOptedIn = (state: AnalyticsController) =>
  state.user_optedIn;

/**
 * @returns Whether the user has opted in via a social account.
 */
export const selectIsOptedIn = (state: AnalyticsController) =>
  state.user_socialOptedIn;

/**
 * Analytics is enabled if the user has opted in through a regular account or a social account.
 *
 * @returns Whether analytics are being sent.
 */
export const selectIsEnabled = (state: AnalyticsController) =>
  getIsOpted(state) || getIsOptedInViaSocial(state);
```

Now these can be used in apps like so:

``` typescript
// === ui/selectors/analyticsController.ts ===

import * from analyticsControllerPackage from '@metamask/analytics-controller';

// on Mobile, this function would look slightly different
const selectFlatMetaMaskState = (state: RootReduxState) =>
  state.metamask;

const selectAnalyticsId = createSelector(
  selectFlatMetaMaskState,
  analyticsControllerPackage.selectAnalyticsId
);
  
// ... etc ...

// === Some component ===

import { selectAnalyticsId } from 'ui/selectors/analyticsController';

export function MyComponent() {
  const analyticsId = useSelector(selectAnalyticsId);
}

// === Outside of a component ===

import { selectAnalyticsId } from 'ui/selectors/analyticsController';

const analyticsId = selectAnalyticsId(analyticsController.state);

// ... etc ...
```

There is a section on using selectors in the controller guidelines here: https://github.com/MetaMask/core/blob/45bdd233a4fb1f318d02bcef85a5b50b4c1024b2/docs/controller-guidelines.md#expose-derived-state-using-selectors-instead-of-getters

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -1,42 +1,109 @@
 import type { Json } from '@metamask/utils';
 
+/**
+ * Represents values that can be passed as properties to the event tracking function.
+ * Similar to JsonValue from Segment SDK but decoupled for platform agnosticism.
+ */
+type AnalyticsJsonValue =
```
mcmire: Is it possible to simplify this type and use `Json`? I see that there are slight differences between this type and `Json`, namely that objects are allowed to have number keys. However if this data really gets serialized to JSON, wouldn't the keys get converted to strings anyway?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -30,38 +37,38 @@ export const controllerName = 'AnalyticsController';
...
    */
-  enabled: boolean;
+  user_optedIn: boolean;
 
   /**
-   * Whether the user has opted in to analytics
+   * User domain: Whether the user has opted in to analytics through social account.
    */
-  optedIn: boolean;
+  user_socialOptedIn: boolean;
```
mcmire: Nit: Should we call this `user_optedInViaSocialAccount`?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -1,42 +1,109 @@
...
+   * This is the same as page in Segment web SDK and screen in Segment mobile SDK.
+   * Each platform adapter should implement this method to track UI views
+   * using the appropriate method for the platform: "pages" for web, "screen" for mobile.
+   *
+   * @param name - The name of the UI item being viewed (pages for web, screen for mobile)
    * @param properties - Page properties
    */
-  trackPage?(pageName: string, properties?: AnalyticsEventProperties): void;
+  view(name: string, properties?: AnalyticsEventProperties): void;
+
```
mcmire, line 66: If we expect a certain string for `name`, should we enforce that in the type? e.g.

``` suggestion
  view(name: 'pages' | 'screen', properties?: AnalyticsEventProperties): void;
```

Or is that not what you meant?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -1,42 +1,109 @@
...
+/**
+ * Event properties structure with two distinct properties lists for anonymous and non-anonymous data.
+ * Similar to ITrackingEvent from legacy analytics but decoupled for platform agnosticism.
+ */
+export type AnalyticsTrackingEvent = {
+  readonly name: string;
+  properties: AnalyticsJsonMap;
+  sensitiveProperties: AnalyticsJsonMap;
+  saveDataRecording: boolean;
+  readonly isAnonymous: boolean;
```
mcmire: I'm not sure if it can be done in this PR — it seems that we are going for compatibility with the legacy system at the moment — but I'm curious if we should rename this to `hasPii` or `isSensitive` or something like that. There used to be a controller metadata property called `anonymous`, and it was always confusing for people. Or maybe we don't need it since if we have `sensitiveProperties`, presumably `isAnonymous` should be `false`? It's possible I misunderstand how this is used in practice.

packages/analytics-controller/src/AnalyticsController.ts
mcmire: Should we at least be more clear about this property to better distinguish it from the other "opted in" property? Perhaps `user_optedInViaRegularAccount` (going by the JSDoc in `computeEnabledState`)?

packages/analytics-controller/src/analyticsStateComputer.ts
```diff
@@ -0,0 +1,44 @@
...
+ * and all computation functions that need to consider them.
+ */
+
+/**
+ * Computes the `controller_enabled` state from controller state.
+ *
+ * @param state - The current controller state
+ * @returns `true` if analytics tracking should be enabled, `false` otherwise
+ */
+export function computeEnabledState(state: AnalyticsControllerState): boolean {
```
mcmire: See my note in `AnalyticsController.ts` for extracting selector functions to obtain pieces of state.

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
+      socialOptedIn: this.state.user_socialOptedIn,
+      analyticsId: this.state.user_analyticsId,
     });
+
+    // Call onSetupCompleted lifecycle hook after initialization
+    // Only call if analyticsId is set and is a valid UUIDv4 (this is the definition of "completed" setup)
+    if (
+      this.state.user_analyticsId &&
+      uuidValidate(this.state.user_analyticsId) &&
+      uuidVersion(this.state.user_analyticsId) === 4
```
mcmire: What do you think about calling these:

``` suggestion
      validateUuid(this.state.user_analyticsId) &&
      getUuidVersion(this.state.user_analyticsId) === 4

File: packages/analytics-controller/src/AnalyticsController.ts [hunk 11]
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
+    ) {
+      try {
+        this.#platformAdapter.onSetupCompleted(this.state.user_analyticsId);
+      } catch (error) {
+        // Log error but don't throw - adapter setup failure shouldn't break controller
+        projectLogger('Error calling platformAdapter.onSetupCompleted', error);
+      }
+    } else {
+      // analyticsId is undefined, null, empty string, or not a valid UUIDv4
+      throw new Error(
```
mcmire: Nit: What are your thoughts on extracting this check into a `#validateState` function/method, calling it right after you call `super` above? Then by the time we get here and try to call `onSetupCompleted` the state should already be valid. (You could also move it out of the controller, call it _before_ you call `super`, and pass it `initialState` — whichever makes sense to you.)

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,105 +221,177 @@ export class AnalyticsController extends BaseController<
...
+    this.#platformAdapter.view(name, properties);
   }
 
   /**
-   * Enable analytics tracking.
+   * Opt in to analytics.
+   * This updates the user's opt-in status.
    */
-  enable(): void {
+  optIn(): void {
```
mcmire, line 342: I see that we have a separate method for opting in via social account. Should we name this something different to distinguish how this might be used? `optInViaRegularAccount` or something? (Note I also made a similar suggestion for the state property.)

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
NicolasMassart: Here are the resons why I used this explicit type:

## Using Record<string, Json> (current):
✅ Explicit: must be an object whith named props of Json type value
✅ Type-safe: rejects arrays, primitives, null
✅ Consistent with BaseController.StateConstraint and ApprovalController

## Using Json directly:
✅ Simpler
✅ Matches runtime JSON serialization behavior
❌ Less type-safe: allows arrays, primitives, null
❌ Less explicit intent as naming the AnalyticsEventProperties and AnalyticsUserTraits types allows to clearly identify the usage in controller.

## My recommendation:
Keep Record<string, Json> for type safety and consistency. If you really prefers Json, we can switch, but it weakens compile-time guarantees.

NicolasMassart:
@metamaskbot publish-preview

packages/analytics-controller/src/analyticsStateValidator.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import type { AnalyticsControllerState } from './AnalyticsController';
+
+/**
+ * Validates that the analytics state has a valid UUIDv4 analyticsId.
+ *
+ * @param state - The analytics controller state to validate
+ * @throws {Error} If analyticsId is missing, invalid, or not a UUIDv4
+ */
+export function validateAnalyticsState(state: AnalyticsControllerState): void {
```
mcmire: Nit: Perhaps we should call this `validateAnalyticsControllerState`?

packages/analytics-controller/src/analyticsStateComputer.ts
```diff
@@ -0,0 +1,44 @@
...
```
mcmire: Nit: Since we have a selector do we need this file? My thought is that the selector can be used not only in Redux but also on its own.

packages/analytics-controller/src/analyticsStateValidator.ts
mcmire, line 1: Nit: Thoughts on calling this file `validateAnalyticsState.ts` to match the function being exported? The current name of the file makes it sounds like this exports a class or something.

packages/analytics-controller/src/analyticsStateComputer.ts
mcmire, line 1: Nit: If we want to keep this file, thoughts on renaming it `computeEnabledState` to match the function being exported? The current name makes it sound like this exports a class or something.

packages/analytics-controller/src/analyticsStateComputer.ts
```diff
@@ -0,0 +1,44 @@
+import type { AnalyticsControllerState } from './AnalyticsController';
+
+/**
```
mcmire: Nit: If we do keep this function it is necessary to have so much documentation for it? I'm all for explaining how things work, but the code seems pretty simple to me, and there is a lot of redundant information here that can be easily gleaned already. (For instance, the fact that we are accessing state implies that whenever state changes, the return value of this function may change. And the code already says `state.optedInForRegularAccount || state.optedInForSocialAccount`, we don't need to repeat that in the documentation. Finally, we should not need to list all of the possible return values for this function; maintainers should be able to look at tests to discover this.)

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
+      expect(controller.state.optedInForSocialAccount).toBe(false);
     });
   });
 
-  describe('messenger actions', () => {
-    it('handles trackEvent action', () => {
-      const mockAdapter = createMockAdapter();
-      const { messenger } = setupController({
-        platformAdapter: mockAdapter,
+  describe('selectAnalyticsId', () => {
```
mcmire: Should the tests for selectors be in this file? It seems like we already have tests for them in `selectors.test.ts`.

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
-    it('disables analytics tracking', () => {
+  describe('optOutForRegularAccount', () => {
+    it('sets optedInForRegularAccount to false', () => {
       const { controller } = setupController({
-        state: { enabled: true },
+        state: { optedInForRegularAccount: true },
       });
 
-      expect(controller.state.enabled).toBe(true);
+      expect(controller.state.optedInForRegularAccount).toBe(true);
```
mcmire: Nit: This assertion seems redundant to me. We already said that `optedInForRegularAccount` should be `true` when passing the initial state, we shouldn't need to double-check what the constructor is doing.

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
   });
 
-  describe('optIn', () => {
-    it('opts in to analytics', () => {
+  describe('optInForSocialAccount', () => {
+    it('sets optedInForSocialAccount to true', () => {
       const { controller } = setupController();
 
-      expect(controller.state.optedIn).toBe(false);
+      expect(controller.state.optedInForSocialAccount).toBe(false);
```
mcmire: Nit:

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
-    it('opts out of analytics', () => {
+  describe('optOutForSocialAccount', () => {
+    it('sets optedInForSocialAccount to false', () => {
       const { controller } = setupController({
-        state: { optedIn: true },
+        state: { optedInForSocialAccount: true },
       });
 
-      expect(controller.state.optedIn).toBe(true);
+      expect(controller.state.optedInForSocialAccount).toBe(true);
```
mcmire: Nit:

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
-    it('enables analytics tracking', () => {
-      const { controller } = setupController({
-        state: { enabled: false },
-      });
+  describe('optInForRegularAccount', () => {
+    it('sets optedInForRegularAccount to true', () => {
+      const { controller } = setupController();
 
-      expect(controller.state.enabled).toBe(false);
+      expect(controller.state.optedInForRegularAccount).toBe(false);
```
mcmire: Nit:

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
+      controller.identify(traits);
 
       expect(mockAdapter.identify).not.toHaveBeenCalled();
     });
   });
 
-  describe('trackPage', () => {
-    it('tracks page view', () => {
+  describe('trackView', () => {
+    it('tracks view', () => {
```
mcmire: Nit: The name of the method is called `trackView`, so I would expect it to... track the view :) Should we be more explicit about how the view gets tracked? What is really happening? I wonder if this is a case where it's okay to include some implementation details.

Also should clarify that this happens when enabled, as we do for other tests in this file?

``` suggestion
    it('calls platform adapter to track view when enabled via regular opt-in', () => {
```

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
+      controller.trackView('home', { referrer: 'test' });
 
-      expect(mockAdapter.trackPage).toHaveBeenCalledWith('home', {
+      expect(mockAdapter.view).toHaveBeenCalledWith('home', {
         referrer: 'test',
       });
     });
 
-    it('does not track page when disabled', () => {
+    it('does not track view when disabled', () => {
```
mcmire:
``` suggestion
    it('does not call platform adapter when disabled', () => {

File: packages/analytics-controller/src/AnalyticsController.test.ts [hunk 23]
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
+      controller.trackEvent(event);
 
-      expect(mockAdapter.trackEvent).toHaveBeenCalledWith('test_event', {});
+      expect(mockAdapter.track).not.toHaveBeenCalled();
     });
+  });
 
-    it('does not track event when disabled', () => {
+  describe('identify', () => {
+    it('identifies user with traits using current analytics ID', () => {
```
mcmire: Similar as for `trackView`, should we be more clear about how the user is identified? 

``` suggestion
    it('identifies user via platform adapter with traits using current analytics ID', () => {

File: packages/analytics-controller/src/AnalyticsController.test.ts [hunk 24]
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
-      expect(controller.state.enabled).toBe(false);
-      expect(controller.state.optedIn).toBe(true);
+      expect(controller).toBeDefined();
+      expect(mockAdapter.onSetupCompleted).toHaveBeenCalledTimes(1);
     });
+  });
 
-    it('uses default analyticsId when not provided in partial state', () => {
+  describe('trackEvent', () => {
+    it('tracks event when enabled via regular opt-in', () => {
```
mcmire:
``` suggestion
    it('calls platform adapter to track event when enabled via regular opt-in', () => {

File: packages/analytics-controller/src/AnalyticsController.test.ts [hunk 25]
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
+      controller.trackEvent(event);
+
+      expect(mockAdapter.track).toHaveBeenCalledWith('test_event', {
+        prop: 'value',
+      });
     });
 
-    it('uses provided analyticsId when passed in state at initialization', () => {
-      const customId = '550e8400-e29b-41d4-a716-446655440000';
+    it('tracks event when enabled via social opt-in', () => {
```
mcmire:
``` suggestion
    it('tracks event via platform adapter when enabled via social opt-in', () => {

File: packages/analytics-controller/src/AnalyticsController.test.ts [hunk 26]
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
       });
+      expect(mockAdapter.track).toHaveBeenNthCalledWith(2, 'test_event', {
+        isSensitive: true,
+        prop: 'value',
+        sensitive_prop: 'sensitive value',
+      });
     });
 
-    it('tracks event with default empty properties when properties not provided', () => {
+    it('does not track event when disabled', () => {
```
mcmire:
``` suggestion
    it('does not call platform adapter when disabled', () => {
```

packages/analytics-controller/src/AnalyticsController.test.ts
mcmire: Looks like we are missing a test `calls platform adapter to track view when enabled via social opt-in` or something equivalent?

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -117,340 +159,447 @@ describe('AnalyticsController', () => {
...
-      expect(restoredController.state.optedIn).toBe(
-        firstController.state.optedIn,
+      expect(mockAdapter.onSetupCompleted).toHaveBeenCalledTimes(1);
+      expect(mockAdapter.onSetupCompleted).toHaveBeenCalledWith(
+        controller.state.analyticsId,
       );
     });
 
-    it('initializes with custom enabled/optedIn', () => {
+    it('returns controller when onSetupCompleted throws error', () => {
```
mcmire: Nit: This is technically true, but what do you think about framing this in terms of not causing any issues?

``` suggestion
    it('discards errors thrown by onSetupCompleted', () => {

**Review by mcmire — APPROVED** [review]
More suggestions, but all of them are super minor, so I will approve.

File: packages/analytics-controller/src/AnalyticsController.test.ts (hunk 21 above)
Comment (NicolasMassart) [reply packages/analytics-controller/src/AnalyticsController.test.ts]: What a view is is explained in track view function, I think we should not overload tests with that.

**Comment by NicolasMassart** [comment]
@metamaskbot publish-preview

File: packages/analytics-controller/src/AnalyticsController.ts [hunk 28]
```diff
@@ -207,92 +221,96 @@ export class AnalyticsController extends BaseController<
...
     }
 
-    // Update state with analytics ID
-    this.update((state) => {
-      state.analyticsId = userId;
+    // Track regular properties (without isSensitive flag - it's the default)
+    // Note: Even if properties object is empty, we still send it to ensure
+    // an event with user ID is tracked. When only sensitiveProperties exist,
+    // this creates two events: one with empty props (user ID) and one with
+    // sensitive props (anonymous ID), which is the expected behavior.
```
mcmire: This last phrase is a bit confusing/redundant, maybe we don't need it? (I would expect the code I see to be the expected behavior. It would be strange if it weren't, no?)

``` suggestion
    // sensitive props (anonymous ID).

File: packages/analytics-controller/src/AnalyticsController.ts [hunk 29]
```diff
@@ -207,92 +221,96 @@ export class AnalyticsController extends BaseController<
...
-    this.update((state) => {
-      state.enabled = false;
-    });
+    // Delegate to platform adapter
+    this.#platformAdapter.view(name, properties);
   }
 
   /**
    * Opt in to analytics.
+   * This updates the user's opt-in status.
```
mcmire: Can this be simplified?

``` suggestion
   * Opt the user into analytics.
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -302,6 +320,7 @@ export class AnalyticsController extends BaseController<
 
   /**
    * Opt out of analytics.
+   * This updates the user's opt-in status.
```
mcmire:
``` suggestion
   * Opt the user out of analytics.
```

packages/analytics-controller/src/analyticsControllerStateValidator.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      [''],
+      ['not-a-uuid'],
+      ['12345'],
+      ['550e8400-e29b-41d4-a716'],
+      ['c232ab00-9414-11e8-8eb2-f2801f1b9fd1'],
+    ])('throws error for invalid input: %s', (analyticsId) => {
+      const state = {
+        optedIn: false,
+        analyticsId,
+      } as unknown as AnalyticsControllerState;
```
mcmire: Instead of typecasting what do you think about using `// @ts-expect-error`? This way if the input is somehow valid later, you will know.

``` suggestion
      // @ts-expect-error Invalid input.
      const state = {
        optedIn: false,
        analyticsId,
      };

File: packages/analytics-controller/src/analyticsControllerStateValidator.ts [hunk 32]
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * UUIDv4 format regex pattern.
+ * Format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
+ * - x is any hexadecimal digit [0-9a-f]
+ * - 4 indicates UUID version 4
+ * - y is the variant indicator [89ab]
+ */
+const UUID_V4_REGEX =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/iu;
```
mcmire, line 11: Is there a reason we aren't using `validate` from `uuid`? https://www.npmjs.com/package/uuid#uuidvalidatestr

packages/analytics-controller/src/selectors.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @returns Whether the user has opted in to analytics
+ */
+const selectOptedIn = (state: AnalyticsControllerState): boolean =>
+  state.optedIn;
+
+/**
+ * Selects whether analytics tracking is enabled.
+ * Use this selector to determine if tracking should occur (e.g., in controller methods).
+ * Currently returns the same value as selectOptedIn, but may diverge in the future
+ * if additional enablement logic is added (e.g., feature flags, platform restrictions).
```
mcmire: Nit: Maybe we don't need to speculate about the future, in case it changes?

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -30,69 +33,62 @@ export const controllerName = 'AnalyticsController';
...
+> {
+  return {
+    optedIn: false,
+  };
+}
+
 /**
  * The metadata for each property in {@link AnalyticsControllerState}.
+ *
+ * Note: `persist` is set to `false` for all fields because the platform
```
mcmire: Can you elaborate on this? How would the clients persist state for this controller in a different way than they normally would?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -30,69 +33,62 @@ export const controllerName = 'AnalyticsController';
...
+   * This is an identity (unique per user), not a preference.
+   * Must be provided by the platform - the controller does not generate it.
    */
   analyticsId: string;
 };
 
+/**
+ * Returns default values for AnalyticsController state.
+ *
+ * Note: analyticsId is NOT included - it's an identity that must be
```
mcmire, line 51: Interesting. Do we ever expect this to change? Maybe it should not be a state property?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -207,92 +221,96 @@ export class AnalyticsController extends BaseController<
...
     projectLogger('AnalyticsController initialized and ready', {
-      enabled: this.state.enabled,
+      enabled: analyticsControllerSelectors.selectEnabled(this.state),
       optedIn: this.state.optedIn,
       analyticsId: this.state.analyticsId,
     });
+
+    // Call onSetupCompleted lifecycle hook after initialization
+    // State is already validated, so analyticsId is guaranteed to be a valid UUIDv4
+    this.#platformAdapter
```
mcmire: Hmm. We don't usually allow asynchronous operations in constructors, because it could create unexpected behavior.

What are your thoughts on moving this to an asynchronous `init` method? You'd have to call this method on the client side, but then you could `await` this. (We follow this pattern in other controllers and services, such as AccountTreeController, MultichainAccountService, etc.)

packages/analytics-controller/README.md
```diff
@@ -24,184 +21,268 @@ or
...
+      plugin: new PrivacyPlugin(analyticsId),
+    });
   },
 };
 ```
 
-### 2. Initialize the Controller
+### 2. Load Analytics Settings from Storage
+
+The platform is responsible for loading and persisting analytics settings—the controller does not handle storage internally. This design allows:
```
mcmire: I'm wondering if we might be misusing the word "platform"? In conversations I've had, "platform" usually means the ecosystem that is shared between the extension and mobile app. For instance, the "Wallet Platform" covers commonalities across clients. Maybe instead of "platform" we really mean "client" instead?

packages/analytics-controller/README.md
```diff
@@ -24,184 +21,268 @@ or
...
+} from '@metamask/analytics-controller';
+
+async function loadAnalyticsSettings(): Promise<AnalyticsControllerState> {
+  // Load from platform storage (e.g., MMKV, AsyncStorage, localStorage)
+  const [savedAnalyticsId, savedOptedIn] = await Promise.all([
+    storage.getItem('analytics.id'),
+    storage.getItem('analytics.optedIn'),
+  ]);
+
+  const defaults = getDefaultAnalyticsControllerState();
```
mcmire: Where are these defaults being used? Maybe we don't need to use them if we still allow a partial representation of state to be provided (except for `analyticsId`)?

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -151,10 +147,17 @@ export type AnalyticsControllerMessenger = Messenger<
  * The options that AnalyticsController takes.
  */
 export type AnalyticsControllerOptions = {
-  state?: Partial<AnalyticsControllerState>;
+  /**
+   * Initial controller state. Must include a valid UUIDv4 `analyticsId`.
+   * The platform is responsible for generating and persisting the analyticsId.
+   */
+  state: AnalyticsControllerState;
```
mcmire, line 156: Should this still be a partial representation of state, but where `analyticsId` is required? Something like:

``` suggestion
  state: Omit<Partial<AnalyticsControllerState>, 'analyticsId'> & Pick<AnalyticsControllerState, 'analyticsId'>;
```

packages/analytics-controller/README.md
```diff
@@ -24,184 +21,268 @@ or
...
+  });
+
+  return controller;
+}
 ```
 
-**Important:** The `state` parameter is the single source of truth for initial values. Any properties you provide will override the defaults from `getDefaultAnalyticsControllerState()`.
+### 4. Access Analytics ID Before Controller Init
+
+One benefit of platform-managed storage is accessing the analytics ID before the controller is initialized:
```
mcmire: I'm a bit confused by "platform-managed storage". Assuming that you mean "client", clients are already responsible for persistence. Maybe you just mean "local storage"? But then, even if you have a controller and you're persisting that state _somehow_, you can already access that state ahead of time by accessing the persistence store directly. So I'm not sure why this is an advantage 🤔

packages/analytics-controller/README.md
```diff
@@ -24,184 +21,268 @@ or
...
+| Field         | Type      | Description                            |
+| ------------- | --------- | -------------------------------------- |
+| `analyticsId` | `string`  | UUIDv4 identifier (platform-generated) |
+| `optedIn`     | `boolean` | User opt-in status                     |
+
+### Why `analyticsId` Has No Default
+
+The `analyticsId` is an **identity** (unique per user), not a **preference** (static default). A default should return the same value each call (deterministic), but a UUID must be unique each time (non-deterministic). These are mutually exclusive.
+
+**Solution:** Platform generates the UUID once on first run, persists it, and provides it to the controller.
```
mcmire: I'm not sure this is totally accurate. We do not require that a `getDefault*State` function return the same values each time. But of course, a user should not receive a new UUID each time the client boots. What do you think about something like this:

``` suggestion
The `analyticsId` is used to uniquely identify the user. If the controller generated a new ID each time the client booted, the ID would be ineffective. Instead, the ID must be pre-generated or retrieved from storage and then passed into the controller.

File: packages/analytics-controller/README.md [hunk 42]
```diff
@@ -24,184 +21,268 @@ or
...
 
-To display analytics-controller logs in the mobile app, you can add the following to your `.js.env` file:
+Enable debug logging:
 
 ```bash
 export DEBUG="metamask:analytics-controller"
 ```
 
-This will enable debug logging for the analytics-controller, allowing you to see detailed logs of analytics events, state changes, and controller operations.
+## Development
```
mcmire: What are your thoughts on removing this? The Contributing doc already explains how to work with packages in the monorepo, and how to run tests. We shouldn't need to re-explain this for each package in the monorepo.

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: putting the optin back into state persistance. Only keeping analytics id in mmkv storage

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: Not expected to change after initial generation, kept in state for ease of use in the controller

packages/analytics-controller/src/AnalyticsController.ts
NicolasMassart: good idea!

packages/analytics-controller/README.md
NicolasMassart: local storage indeed, MMKV on mobile for now

packages/analytics-controller/src/analyticsControllerStateValidator.ts
NicolasMassart: We only verify une version here, the v4. As we do not generate UUID anymore in the controller, we only verify the v4 format and it seemed simple enough for a regex.
We can change in the future when we eventually put the UUIDv4 gerentation back in the controller.

packages/analytics-controller/README.md
NicolasMassart: I will get rid of it, yes.

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -151,12 +147,26 @@ export type AnalyticsControllerMessenger = Messenger<
...
+   * Platform adapter implementation for tracking events.
    */
   platformAdapter: AnalyticsPlatformAdapter;
+
+  /**
+   * Whether the anonymous events feature is enabled.
+   *
+   * @default false
+   */
+  anonymousEventsFeature?: boolean;
```
mcmire: Nit: Maybe we call this `isAnonymousEventsFeatureEnabled`? Or perhaps `honorSensitiveProperties` or `useSensitiveProperties`?

packages/analytics-controller/src/index.ts
```diff
@@ -1,16 +1,27 @@
-// Export controller class
-export { AnalyticsController } from './AnalyticsController';
+// Export controller class and state utilities
+export {
+  AnalyticsController,
+  getDefaultAnalyticsControllerState,
+  controllerName,
```
mcmire: I must have missed this before.

While it's common to export `controllerName` within a controller file, it's not common to export it from the package. Can we remove this one?

``` suggestion
```

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -176,119 +191,150 @@ export class AnalyticsController extends BaseController<
...
-  trackEvent(
-    eventName: string,
-    properties: AnalyticsEventProperties = {},
-  ): void {
-    // Don't track if analytics is disabled
-    if (!this.state.enabled) {
-      return;
+  init(): void {
+    if (this.#initialized) {
+      throw new Error('AnalyticsController already initialized.');
```
mcmire: Nit: Maybe we don't need to throw an error? Maybe we do nothing and use `console.warn` or `projectLogger` to print a warning instead?

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -61,149 +75,407 @@ function setupController(
...
       });
 
-      expect(controller.state.enabled).toBe(false);
-      expect(controller.state.optedIn).toBe(true);
+      expect(controller1.state.analyticsId).toBe(analyticsId1);
+      expect(controller2.state.analyticsId).toBe(analyticsId2);
     });
+  });
+
+  describe('onSetupCompleted lifecycle hook', () => {
```
mcmire: It seems that this is a duplicate test block. `setupController` calls `init` implicitly and so we are effectively doing the same thing that the `describe('init', () => {` block is doing. Thoughts on removing this block?

packages/analytics-controller/src/AnalyticsController.test.ts
```diff
@@ -214,239 +486,413 @@ describe('AnalyticsController', () => {
...
-
-      expect(mockAdapter.trackEvent).toHaveBeenCalledWith('test_event', {
-        prop: 'value',
-      });
+      expect(onSetupCompletedSpy).toHaveBeenCalledTimes(1);
+      expect(onSetupCompletedSpy).toHaveBeenCalledWith(analyticsId);
     });
 
-    it('tracks event with default empty properties when properties not provided', () => {
+    it('returns early if called multiple times', async () => {
```
mcmire: Nit: This seems like an implementation detail. Maybe this should say:

``` suggestion
    it('only calls onSetupCompleted one time even if called multiple times', async () => {

File: packages/analytics-controller/src/AnalyticsController.test.ts [hunk 48]
```diff
@@ -214,239 +486,413 @@ describe('AnalyticsController', () => {
...
-      messenger.call('AnalyticsController:optIn');
-
-      expect(controller.state.optedIn).toBe(true);
     });
+  });
 
-    it('handles optOut action', () => {
-      const { controller, messenger } = setupController({
-        state: { optedIn: true },
+  describe('stateChange event', () => {
```
mcmire: Is this test needed? If we are testing that the state changes when `optIn` is called then we are implicitly testing that the state change handler will be called. This is behavior that is common to all controllers, and is already tested in the tests for BaseController. So maybe we don't need this block?

packages/analytics-controller/README.md
```diff
@@ -24,184 +21,261 @@ or
...
-// Events are automatically filtered when analytics is disabled
-controller.disable();
-controller.trackEvent('some_event'); // This will not be tracked
+// Events are filtered when analytics is disabled (optedIn is false)
 ```
 
-### 4. Identify Users
+#### Anonymous Events Feature
+
+When `anonymousEventsFeature` is enabled in the constructor, events with sensitive properties are split into separate events:
```
mcmire: This option was renamed to `isAnonymousEventsFeatureEnabled`, should we update this section?

PR #7087, opened 2025-11-07
fix(multichain-account-service)!: add `resyncAccounts` support

ccharly:
## Explanation

Add a new `resyncAccounts` method/action to allow BIP-44 compatible Snaps to re-synchronize their accounts with MetaMask.

We have encountered this bug recently when the Solana Snap had less accounts than MetaMask itself. This causes issues where MetaMask uses those accounts (like to initiate a Send flow) which will fail once the Snap is invoked (since the Snap does not know this account).

Since some states are already missing thoses, accounts we're adding this recovery mechanism to re-synchronize both states, but the real root cause of this issue has still not be found yet.

## References

Mitigation for:
- https://github.com/MetaMask/metamask-extension/issues/37050

## Checklist

---

> [!NOTE]
> Adds provider/account re-sync via `resyncAccounts` and refactors provider typing, while making `init` async.
> 
> - **Core**
>   - Add `MultichainAccountService.resyncAccounts()` to best‑effort re-sync provider accounts with MetaMask (filters BIP‑44, reports errors without throwing).
>   - Change `init()` to `async`.
> - **Providers**
>   - Replace `NamedAccountProvider` with `Bip44AccountProvider` across service, wallet, group, and tests.
>   - Add `resyncAccounts()` to all providers:
>     - EVM: no‑op.
>     - Snap-based (`SnapAccountProvider`): implement re-sync via shared `KeyringClient`, detect de-sync, remove/recreate missing accounts, capture errors.
>   - `AccountProviderWrapper`: forward/skip `resyncAccounts` based on enablement.
>   - Consolidate Snap providers to use a shared `client` field; remove per-class clients in `{Sol,Btc,Trx}AccountProvider`.
> - **Tests**
>   - Update to `await service.init()`, add comprehensive re-sync tests for service and Snap providers, and adjust concurrency helpers.
> - **Changelog**
>   - Document new `resyncAccounts` APIs and breaking `async init`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d727e1dfebd2434606b9d1de12617723b1235ec4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/eslint-warning-thresholds.json
+++ b/eslint-warning-thresholds.json
@@ -2,6 +2,9 @@
   "packages/accounts-controller/src/AccountsController.test.ts": {
     "import-x/namespace": 1
   },
+  "packages/accounts-controller/src/utils.ts": {
+    "@typescript-eslint/no-unsafe-enum-comparison": 8
+  },
   "packages/approval-controller/src/ApprovalController.test.ts": {
     "import-x/order": 1,
     "jest/no-conditional-in-test": 16
@@ -14,12 +17,33 @@
     "n/prefer-global/text-decoder": 1,
     "no-shadow": 2
   },
+  "packages/assets-controllers/src/AccountTrackerController.ts": {
+    "@typescript-eslint/no-misused-promises": 4
+  },
   "packages/assets-controllers/src/CurrencyRateController.test.ts": {
     "jest/no-conditional-in-test": 1
+  },
+  "packages/assets-controllers/src/DeFiPositionsController/DeFiPositionsController.ts": {
+    "@typescript-eslint/no-misused-promises": 2
+  },
+  "packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts": {
+    "@typescript-eslint/no-misused-promises": 3
+  },
+  "packages/assets-controllers/src/MultichainAssetsController/utils.ts": {
+    "@typescript-eslint/no-unsafe-enum-comparison": 1
+  },
+  "packages/assets-controllers/src/MultichainAssetsRatesController/MultichainAssetsRatesController.ts": {
+    "@typescript-eslint/no-misused-promises": 2
+  },
+  "packages/assets-controllers/src/MultichainBalancesController/MultichainBalancesController.ts": {
+    "@typescript-eslint/no-misused-promises": 2
   },
   "packages/assets-controllers/src/NftController.test.ts": {
     "import-x/namespace": 9,
     "jest/no-conditional-in-test": 6
+  },
+  "packages/assets-controllers/src/NftController.ts": {
+    "@typescript-eslint/no-misused-promises": 2
   },
   "packages/assets-controllers/src/NftDetectionController.test.ts": {
     "import-x/namespace": 6,
@@ -49,6 +73,12 @@
   },
   "packages/assets-controllers/src/Standards/NftStandards/ERC721/ERC721Standard.ts": {
     "prettier/prettier": 1
+  },
+  "packages/assets-controllers/src/TokenBalancesController.ts": {
+    "@typescript-eslint/no-misused-promises": 1
+  },
+  "packages/assets-controllers/src/TokenDetectionController.ts": {
+    "@typescript-eslint/no-misused-promises": 5
   },
   "packages/assets-controllers/src/TokenListController.test.ts": {
     "import-x/namespace": 7,
@@ -84,6 +114,9 @@
   },
   "packages/base-controller/src/BaseController.test.ts": {
     "import-x/namespace": 13
+  },
+  "packages/bridge-status-controller/src/utils/transaction.ts": {
+    "@typescript-eslint/no-unsafe-enum-comparison": 2
   },
   "packages/build-utils/src/transforms/remove-fenced-code.test.ts": {
     "import-x/order": 1
@@ -130,7 +163,8 @@
     "@typescript-eslint/naming-convention": 1
   },
   "packages/eth-block-tracker/src/PollingBlockTracker.test.ts": {
-    "@typescript-eslint/naming-convention": 1
+    "@typescript-eslint/naming-convention": 1,
+    "@typescript-eslint/unbound-method": 4
   },
   "packages/eth-block-tracker/src/PollingBlockTracker.ts": {
     "@typescript-eslint/naming-convention": 1,
@@ -157,6 +191,9 @@
     "@typescript-eslint/no-explicit-any": 1,
     "jsdoc/require-jsdoc": 1,
     "no-restricted-syntax": 1
+  },
+  "packages/eth-json-rpc-middleware/src/block-ref-rewrite.test.ts": {
+    "@typescript-eslint/no-misused-promises": 3
   },
   "packages/eth-json-rpc-middleware/src/block-ref-rewrite.ts": {
     "jsdoc/match-description": 1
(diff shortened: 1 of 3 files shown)


ccharly:
@metamaskbot publish-preview

packages/multichain-account-service/src/providers/AccountProviderWrapper.ts
```diff
@@ -34,6 +35,21 @@ export class AccountProviderWrapper extends BaseBip44AccountProvider {
     this.isEnabled = enabled;
   }
 
+  /**
+   * Override getAccounts to return empty array when disabled.
```
danroc: WDYM?

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -162,7 +163,7 @@ export class MultichainAccountService {
    * Initialize the service and constructs the internal reprensentation of
    * multichain accounts and wallets.
    */
-  init(): void {
+  async init(): Promise<void> {
```
danroc, line 178: Did you make the `init` async only to future-proof it or does it need to be async?

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -11,9 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **BREAKING:** Added error reporting around account creation with the `ErrorReportingService` ([#7044](https://github.com/MetaMask/core/pull/7044))
   - The `@metamask/error-reporting-service` is now a peer dependency.
+- Add `MultichainAccountService.resyncAccounts` method and action ([#7087](https://github.com/MetaMask/core/pull/7087))
+- Add `SnapAccountProvider.resyncAccounts` method ([#7087](https://github.com/MetaMask/core/pull/7087))
```
danroc: It's actually to all providers, right?

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -92,6 +96,103 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
...
+      (account) =>
+        account.metadata.snap && account.metadata.snap.id === this.snapId,
+    );
+    const snapAccounts = new Set(
+      (await this.client.listAccounts()).map((account) => account.id),
+    );
+
+    // We want this part to be fast, so we only check for sizes, but we might need
+    // to make "diff" between the 2 states to not miss any de-sync.
+    if (localSnapAccounts.length > snapAccounts.size) {
```
danroc, line 158: It might be worth logging if `localSnapAccounts.length < snapAccounts.size`, which also shouldn't happen.

packages/multichain-account-service/src/providers/AccountProviderWrapper.ts
```diff
@@ -34,6 +35,21 @@ export class AccountProviderWrapper extends BaseBip44AccountProvider {
...
+   * @param accounts - List of local accounts.
+   * @returns Array of accounts, or empty array if disabled.
+   */
+  override async resyncAccounts(
+    accounts: Bip44Account<InternalAccount>[],
+  ): Promise<void> {
+    if (!this.isEnabled) {
+      return;
+    }
+    await this.provider.resyncAccounts(accounts);
```
gantunesr, line 49: why not have the inverse condition?
```suggestion
    if (this.isEnabled) {
      await this.provider.resyncAccounts(accounts);
    }
```

packages/multichain-account-service/src/providers/BtcAccountProvider.ts
```diff
@@ -23,8 +20,6 @@ export class BtcAccountProvider extends SnapAccountProvider {
 
   static BTC_SNAP_ID = 'npm:@metamask/bitcoin-wallet-snap' as SnapId;
 
-  readonly #client: KeyringClient;
```
gantunesr, line 29: why the client is not a private attr anymore?

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -92,6 +96,103 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
       });
   }
 
+  #getKeyringClientFromSnapId(snapId: string): KeyringClient {
```
gantunesr, line 114: nit: should this be `create...` since it's returning a new instance of `KeyringClient`?

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -92,6 +96,103 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
...
+              await this.#withSnapKeyring(
+                async ({ keyring }) =>
+                  await keyring.removeAccount(account.address),
+              );
+
+              // The Snap has no account in its state for this one, we re-create it.
+              await this.createAccounts({
+                entropySource,
+                groupIndex,
+              });
```
gantunesr, line 187: I wonder if we should do some changes at the `SnapKeyring` to avoid the execution of remove+create here

packages/multichain-account-service/CHANGELOG.md
ccharly: That's right, I'll change that (I initially marked that as "breaking" since it was a new method on this type, but since it's not `abstract` and is implemented there, it was not breaking after all 🙃

packages/multichain-account-service/src/MultichainAccountService.ts
ccharly: Mostly yes, since we already have some breaking changes for the upcoming release of this package, I wanted to ship it too to avoid multiple breaking changes in several releases.

Ultimately, I might decide to run `resyncAccounts` there too, but I'm still unsure if that's a good move or not!

packages/multichain-account-service/src/providers/AccountProviderWrapper.ts
ccharly: The usual bad copy/paste 🙊 I'll fix that

packages/multichain-account-service/src/providers/AccountProviderWrapper.ts
ccharly: I mostly followed the same pattern than other methods here, we check for `!isEnabled` or forward the call to the underlying provider

packages/multichain-account-service/src/providers/BtcAccountProvider.ts
ccharly: Moved it to `SnapAccountProvider` instead, since all concrete implementation of providers were just using the exact same logic for this AND I needed the client in `SnapAccountProvider` to call `keyring_listAccounts` as well

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: Yes that makes total sense! I'll add that

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: I just re-used what we had and I know we sometimes use `getSomething` to create other objects (but that's a bit ambiguous indeed)

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: I don't think so since `createAccount` imply that we know how to create account, here we abstract that through the providers.

However, I think that having a `resyncAccount` (singular) could be useful (if we have the other case where the Snaps has more accounts than MetaMask!

danroc (approved):

PR #7091, opened 2025-11-07
Ramps Controller Foundation

pkowalski:
### Current State:
Ramps currently depends on an sdk to handle all connections to the RAMPS api. We now have 2 different offerings which each use their own SDK. All of this routing logic is done on the mobile app. If this logic were to be extended to mobile or extension it would require rework on those codebases. 

### Why we want to change things:
Many other teams already use this controller design pattern. By moving towards this we can handle all business logic in 1 controller. It lets us expose Ramps data and apis to other experiences without those experiences requiring any knowledge of business logic. Lastly it would speed up any Ramps api integrations on extension and mobile in the future.

### Changes introduced:
This is the foundation for the Ramps Controller. It contains 1 function which updates its internal state which reflects eligibility for a ramps feature on a certain region. 

## References

* Related to [TRAM-2783](https://consensyssoftware.atlassian.net/browse/TRAM-2783)

## Checklist

[TRAM-2783]: https://consensyssoftware.atlassian.net/browse/TRAM-2783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> Introduce `@metamask/ramps-controller` with a `getCountries` action that resolves API base by environment, fetches geolocation and region eligibility, and updates controller state.
> 
> - **New Package**: `packages/ramps-controller`
>   - **Controller**: `RampsController` with state (`metamaskEnvironment`, `context`, `region`) and metadata exposure.
>   - **API Logic**: `getCountries` fetches `geolocation` then `regions/countries/<code>` using env-specific base URL; validates response and updates `state.region`; registers messenger action `RampsController:getCountries`.
>   - **Types/Exports**: Public types and enums (`SdkEnvironment`, `Context`) via `src/index.ts`.
>   - **Tests**: Comprehensive unit tests for env URL selection, success path, validation, error handling, state updates, and messenger action registration.
>   - **Build/Docs**: Jest config, TypeScript configs, Typedoc config, package metadata, README, LICENSE, CHANGELOG.
> - **Repo Integration**:
>   - Add `@metamask/ramps-controller` to `README.md` package list and dependency graph.
>   - Update `tsconfig.json` and `tsconfig.build.json` references.
>   - Update `.github/CODEOWNERS` to add Ramp team ownership and release file ownership entries.
>   - Add workspace entry in `yarn.lock`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9efa238b287b8fdd7e342543adfb566bb2963002. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12,154 +12,164 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@algolia/cache-browser-local-storage@npm:4.25.2":
-  version: 4.25.2
-  resolution: "@algolia/cache-browser-local-storage@npm:4.25.2"
+"@algolia/cache-browser-local-storage@npm:4.24.0":
+  version: 4.24.0
+  resolution: "@algolia/cache-browser-local-storage@npm:4.24.0"
   dependencies:
-    "@algolia/cache-common": "npm:4.25.2"
-  checksum: 10/ca0f39001e1d8d9b42adea349baadf1ff9f2ff43e87f4bc85928c0298cc533c1b061f393e1df40a87dbc365e4231dd85a6d8867fbdbd7e6de74c4a79745f307d
+    "@algolia/cache-common": "npm:4.24.0"
+  checksum: 10/f7f9bdb1fa37e788a5cb8c835e526caff2fa097f68736accd4c82ade5e5cb7f5bbd361cf8fc8c2a4628d979d81bd90597bdaed77ca72de8423593067b3d15040
   languageName: node
   linkType: hard
 
-"@algolia/cache-common@npm:4.25.2":
-  version: 4.25.2
-  resolution: "@algolia/cache-common@npm:4.25.2"
-  checksum: 10/706dca5d9c570490b4760646a94d5ed7b603cc846ae3b62c42fc5e7958cdd74b7bd37506be62bf4b97f4b6551526f7a64597c3240fe2c5e7477a8ece986c8d12
+"@algolia/cache-common@npm:4.24.0":
+  version: 4.24.0
+  resolution: "@algolia/cache-common@npm:4.24.0"
+  checksum: 10/bc1d0f8731713f7e6f10cd397b7d8f7464f14a2f4e1decc73a48e99ecbc0fe41bd4df1cc3eb0a4ecf286095e3eb3935b2ea40179de98e11676f8e7d78c622df8
   languageName: node
   linkType: hard
 
-"@algolia/cache-in-memory@npm:4.25.2":
-  version: 4.25.2
-  resolution: "@algolia/cache-in-memory@npm:4.25.2"
+"@algolia/cache-in-memory@npm:4.24.0":
+  version: 4.24.0
+  resolution: "@algolia/cache-in-memory@npm:4.24.0"
   dependencies:
-    "@algolia/cache-common": "npm:4.25.2"
-  checksum: 10/ffd66ff76cba41a1dd26e3575902881e7a5f685018c5d5b6ec2a972c682a56940b593ac3d33d59aac296db884328f0bd19a16e83281c69e3582ab121ed759d56
+    "@algolia/cache-common": "npm:4.24.0"
+  checksum: 10/0476f65f4b622b1b38f050a03b9bf02cf6cc77fc69ec785d16e244770eb2c5eea581b089a346d24bdbc3561be78d383f2a8b81179b801b2af72d9795bc48fee2
   languageName: node
   linkType: hard
 
-"@algolia/client-account@npm:4.25.2":
-  version: 4.25.2
-  resolution: "@algolia/client-account@npm:4.25.2"
+"@algolia/client-account@npm:4.24.0":
+  version: 4.24.0
+  resolution: "@algolia/client-account@npm:4.24.0"
   dependencies:
-    "@algolia/client-common": "npm:4.25.2"
-    "@algolia/client-search": "npm:4.25.2"
-    "@algolia/transporter": "npm:4.25.2"
-  checksum: 10/a39e1be468aef039093f604872afee2cdf8ba17b1b6ed46fe4454a199b147c54a6daccef4efa830d1536385a3f0f5999bb37c52c9b4e09b943117b82d9bd5b3a
+    "@algolia/client-common": "npm:4.24.0"
+    "@algolia/client-search": "npm:4.24.0"
+    "@algolia/transporter": "npm:4.24.0"
+  checksum: 10/059cf39f3e48b2e77a26435267284d2d15a7a3c4e904feb2b2ad2dd207a3ca2e2b3597847ec9f3b1141749b25fb2e6091e9933f53cb86ab278b5b93836c85aad
   languageName: node
   linkType: hard
 
-"@algolia/client-analytics@npm:4.25.2":
-  version: 4.25.2
-  resolution: "@algolia/client-analytics@npm:4.25.2"
+"@algolia/client-analytics@npm:4.24.0":
+  version: 4.24.0
+  resolution: "@algolia/client-analytics@npm:4.24.0"
   dependencies:
-    "@algolia/client-common": "npm:4.25.2"
-    "@algolia/client-search": "npm:4.25.2"
-    "@algolia/requester-common": "npm:4.25.2"
-    "@algolia/transporter": "npm:4.25.2"
-  checksum: 10/9954f32826d47cd99cb3bd34c7164d752ae5ff895e53ad5fddb4ac1949a98718ef0bb9c9e6bb0df1713438981d9bf038257137ab62df1a3bd72d6836f681b80e
+    "@algolia/client-common": "npm:4.24.0"
+    "@algolia/client-search": "npm:4.24.0"
+    "@algolia/requester-common": "npm:4.24.0"
+    "@algolia/transporter": "npm:4.24.0"
+  checksum: 10/eaa4be80636082a1fbeb0d099ef882ae6576fb0b6dc64988e9e6939533b4ddfffdbe16061cfd3f89b18bbf5aba21dff5a68af4f20b2719cf72d83a1f0774f6d5
   languageName: node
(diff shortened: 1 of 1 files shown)


packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+    state,
+  }: {
+    messenger: RampsControllerMessenger;
+    state: Partial<RampsControllerState>;
+  }) {
+    super({
+      messenger,
+      name: controllerName,
+      metadata: rampsControllerMetadata,
+      state: { ...defaultState, ...state } as RampsControllerState,
```
mcmire, line 130: Is there a reason why you are using a type assertion here? There should be no need for this, you should have a complete state object here.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+  MobileAndroid = 'mobile-android',
+  MobileIOS = 'mobile-ios',
+}
+
+/**
+ * Ramps controller state
+ */
+export type RampsControllerState = {
+  // Environment tells us which API urls to us
+  metamaskEnvironment: SdkEnvironment;
```
mcmire, line 34: Why is the `metamaskEnvironment` and `context` in state? Do we expect to be able to change these at runtime?

`context` also doesn't seem to be getting used anywhere.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+        data === null ||
+        typeof data.deposit !== 'boolean' ||
+        typeof data.aggregator !== 'boolean' ||
+        typeof data.global !== 'boolean'
+      ) {
+        throw new Error('Invalid country data format');
+      }
+
+      this.update((state) => {
+        state.region = {
```
mcmire, line 202: Can you explain how you intend to use this controller? Do you expect to call `getCountries()` somewhere and listen for changes to `region` somewhere else? Why not call it immediately and get the results?

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    const urlWithPath = new URL(service ?? '', url);
+    return urlWithPath.toString();
+  }
+
+  async #getGeolocation(): Promise<string> {
+    try {
+      const url = this.#getApiUrl();
+      const response = await fetch(new URL('geolocation', url).toString());
```
mcmire, line 157: Instead of making HTTP requests in your controller, we recommend extracting an API client from your controller that represents the "on ramp" API. It could serve as a direct layer on top of the layer and could be used independent of the controller anywhere in the stack. 

We have a pattern called "data services" that would be useful here. In this case perhaps you want to:

1. Create a separate file, perhaps `src/OnRampService.ts`.
2. Define an `OnRampService` class in it. The constructor should take two arguments: `messenger` and `policyOptions`. It might also be best to take `environment` and/or `context` options based on what you're already doing in the controller (the `environment` can be used to construct the base URL).
3. Add `getGeolocation` and `getCountries` methods to the class, similar to how you have here. Each method should make a `fetch` and that `fetch` should be rapped in a `policy.execute` call.
4. Add a `messenger` option to the constructor. Create a `OnRampServiceMessenger` type and expose those methods using the messenger (see other comment on how to do this in a more automatic way)
5. Add the service messenger actions to the RampsController's messenger, and then call the service methods through the messenger (e.g. `await messenger.call('OnRampService:getGeolocation', ...)`.

We have an example data service here: https://github.com/MetaMask/core/blob/e144cc58d7b82c0d75c685de85adc3de510a1b3e/packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts. And you can see how it's used in the controller here: https://github.com/MetaMask/core/blob/e144cc58d7b82c0d75c685de85adc3de510a1b3e/packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.ts

Let me know your thoughts on this!

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+  region: {
+    persist: true,
+    anonymous: true,
+    includeInDebugSnapshot: true,
+    includeInStateLogs: true,
+    usedInUi: true,
+  },
+};
+
+const defaultState: RampsControllerState = {
```
mcmire, line 92: Have you read our controller guidelines? This states that the default state should be wrapped in a function: https://github.com/MetaMask/core/blob/main/docs/controller-guidelines.md#provide-a-default-representation-of-state

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+  metamaskEnvironment: SdkEnvironment.Staging,
+  context: Context.Browser,
+  region: null,
+};
+
+enum ApiService {
```
mcmire, line 98: Do you need this enum if there is only one thing in it?

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+      });
+    } catch (error) {
+      // Handle or re-throw as needed; here we simply log
+      console.error('Error in getCountries:', error);
+      // Optionally, clear region on error:
+      this.update((state) => {
+        state.region = null;
+      });
+      // Or rethrow if you want the caller to handle it:
+      // throw error;
```
mcmire, line 217: Generally in controllers and services we don't recommend swallowing errors like this unless there is a good reason to do so. But it depends on the use case — can you provide an example of how this would be used in the client?

packages/ramps-controller/src/RampsController.ts
```diff
@@ -0,0 +1,233 @@
...
+    state: Partial<RampsControllerState>;
+  }) {
+    super({
+      messenger,
+      name: controllerName,
+      metadata: rampsControllerMetadata,
+      state: { ...defaultState, ...state } as RampsControllerState,
+    });
+
+    this.#registerMessageHandlers();
```
mcmire, line 133: A little while ago we added a way to automatically expose methods as actions. This way you don't need to think about and maintain the types so much whenever you add new methods in the future. Here are some instructions:

- Add `MESSENGER_EXPOSED_METHODS` somewhere in your controller file, e.g.: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L91
- Run `yarn generate-method-action-types`. This will generate a new file, `AnalyticsController-method-action-types.ts`.
- Import `AnalyticsControllerMethodActions` from this file.
- Add this type to `AnalyticsControllerActions`, e.g.: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L106
- Call `this.messenger.registerMethodActionHandlers` in your controller constructor: https://github.com/MetaMask/core/blob/e5897c7d5245c12c2dd66f347ec0fed291bb5ecd/packages/sample-controllers/src/sample-gas-prices-controller.ts#L256

packages/ramps-controller/src/RampsController.test.ts
```diff
@@ -0,0 +1,719 @@
...
+
+import { RampsController, SdkEnvironment, Context } from './RampsController';
+import type { RampsControllerMessenger } from './RampsController';
+
+describe('RampsController', () => {
+  let mockFetch: jest.MockedFunction<typeof fetch>;
+  let mockConsoleError: jest.SpyInstance;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
```
mcmire, line 18: Mocks get automatically reset between tests in this repo, so you don't need this:

``` suggestion
```

packages/ramps-controller/src/RampsController.test.ts
```diff
@@ -0,0 +1,719 @@
...
+    mockFetch = jest.spyOn(global, 'fetch') as jest.MockedFunction<
```
mcmire, line 19: We don't recommend mocking the `fetch` function directly. Nock is already set in this repo and that offers a more realistic way of mocking requests. You can see an example here: https://github.com/MetaMask/core/blob/e144cc58d7b82c0d75c685de85adc3de510a1b3e/packages/sample-controllers/src/sample-gas-prices-service/sample-gas-prices-service.test.ts#L29

packages/ramps-controller/src/RampsController.test.ts
```diff
@@ -0,0 +1,719 @@
...
+      typeof fetch
+    >;
+    mockConsoleError = jest.spyOn(console, 'error').mockImplementation();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
```
mcmire, line 28: Similar to comment above, this already happens by default.

``` suggestion
```

packages/ramps-controller/src/RampsController.ts
pkowalski: Good point. Will move context and environment out of state.

Context will be used for future calls. I will add it to the axios object as a default param.

packages/ramps-controller/src/RampsController.ts
pkowalski: True, we can do both. Return it but update the state. We will want to access it later without calling the function so keeping it in state was intentional.

pkowalski: I added a getter method which will init the state if not initialized. Let me know if this makes sense or if the patter is to access the state in the controller directly instead.

mcmire: Hmm, I don't see this, is it possible you forgot to push those changes?

PR #7106, opened 2025-11-10
fix: force timeout after 30s token detection

salimtb:

Currently, when `TokenDetectionController` uses the Accounts API to detect tokens, API calls can hang indefinitely if the API is slow, unresponsive, or experiencing network issues. This causes the entire token detection process to freeze without any fallback mechanism, resulting in a poor user experience where tokens are never detected.

This PR adds 30-second timeout protection for Accounts API calls in `TokenDetectionController`:

- Added `ACCOUNTS_API_TIMEOUT_MS` constant (30000ms) to define the timeout threshold

- Wrapped Accounts API calls with `Promise.race()` between the actual API call and a timeout promise

- When timeout occurs, the promise rejects and is caught, triggering an automatic fallback to RPC-based token detection

- Properly cleans up the timeout in a `finally` block to prevent memory leaks

- Includes error logging for debugging timeout and failure events

The timeout mechanism ensures that:

1. If the API responds within 30 seconds, detection proceeds normally via the API

2. If the API takes longer than 30 seconds, the timeout fires and RPC detection takes over

3. Users always get token detection results, either via API or RPC fallback

### Changes that might not be obvious

The timeout is implemented in the `#attemptAccountAPIDetection` private method, which wraps the existing `#addDetectedTokensViaAPI` call. This ensures the timeout protection applies to all Accounts API token detection flows without requiring changes to the core detection logic.

## References

- Improves reliability of token detection by preventing indefinite hangs

- Ensures users always get token detection results through automatic RPC fallback

- Related to ongoing work to improve Accounts API reliability and user experience

## Checklist

  - Added comprehensive timeout test using fake timers to simulate 30-second timeout scenario

  - Test verifies that API is called, timeout triggers, RPC fallback occurs, and tokens are successfully added

  - Added inline comments explaining timeout mechanism and cleanup

  - Updated `CHANGELOG.md` with timeout protection fix

  - Not applicable: This is an internal improvement with no breaking changes

### Key Files Changed

1. **`TokenDetectionController.ts`**
   - Added `ACCOUNTS_API_TIMEOUT_MS` constant (30000ms)
   - Modified `#attemptAccountAPIDetection` method to implement timeout protection
   - Uses `Promise.race()` to race between API call and timeout promise
   - Includes proper cleanup in `finally` block to prevent memory leaks

2. **`TokenDetectionController.test.ts`**
   - Added test case: `'should timeout and fallback to RPC when Accounts API call takes longer than 30 seconds'`
   - Uses `sinon.useFakeTimers()` to simulate time advancement
   - Verifies complete flow: API call → timeout → RPC fallback → token addition

### Code Flow

```
detectTokens()
  ↓
#attemptAccountAPIDetection()
  ↓
Promise.race([
  #addDetectedTokensViaAPI() ← actual API call
  timeout promise (30s)      ← timeout protection
])
  ↓
Success → continue with API results
  ↓
Timeout/Failure → fallback to RPC detection
  ↓
finally → cleanup timeout
```

### Testing Strategy

The test uses fake timers to simulate the 30-second timeout without actually waiting 30 seconds:

1. Mock API call to never resolve (simulates hanging request)
2. Start detection process
3. Advance fake timers by 30 seconds
4. Verify timeout triggered
5. Verify RPC fallback occurred
6. Verify tokens were successfully added

### User Experience

- **Before**: Token detection could hang indefinitely, leaving users without their tokens
- **After**: Token detection always completes within 30 seconds, with automatic RPC fallback

### Performance

- No performance impact on successful API calls (they complete normally)
- Failed/slow API calls are cut off at 30 seconds instead of hanging forever
- Memory leak prevention through proper timeout cleanup

### Reliability

- Significantly improves reliability of token detection
- Provides graceful degradation when Accounts API is experiencing issues
- Maintains existing functionality while adding safety net

---

> [!NOTE]
> Adds a 30s timeout to Accounts API token detection with RPC fallback, and refactors balance fetchers to return `unprocessedChainIds` so controllers can retry unsupported chains via RPC.
> 
> - **Token Detection (`TokenDetectionController`)**:
>   - Add `ACCOUNTS_API_TIMEOUT_MS` (30s) and wrap API calls with `safelyExecuteWithTimeout`; on timeout/failure, fall back to RPC.
>   - Process Accounts API `unprocessedNetworks` and add those chains to RPC detection.
> - **Balance Fetchers**:
>   - Change `fetch()` to return `{ balances, unprocessedChainIds }` in `AccountsApiBalanceFetcher` and `RpcBalanceFetcher`.
>   - Convert Accounts API `unprocessedNetworks` (decimal) to hex `unprocessedChainIds`; aggregate across batches.
>   - For `AccountTracker` RPC fetcher wrapper, filter out staked entries when `includeStakedAssets` is false while preserving `unprocessedChainIds`.
> - **Controllers**:
>   - Update `AccountTrackerController` and `TokenBalancesController` to consume new fetcher result shape and re-queue `unprocessedChainIds` for fallback fetchers.
> - **Tests & Misc**:
>   - Update tests across fetchers/controllers for new return type and timeout/fallback behavior; add coverage for missing currencies in `CurrencyRateController` responses.
>   - Update `CHANGELOG.md` to reflect timeout protection and unprocessed network handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a813ba4415446ce212cba0db7207cb2f052ff99. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -603,11 +606,38 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
...
-    return await this.#addDetectedTokensViaAPI({
-      chainIds: chainsToDetectUsingAccountAPI,
-      selectedAddress: addressToDetect,
-      supportedNetworks,
-    });
+    let timeoutId: NodeJS.Timeout | undefined;
+
+    try {
+      // Create a timeout promise that rejects after 30 seconds
+      const timeoutPromise = new Promise<{ result: 'failed' }>(
```
sahar-fehri: I have seen that there is a "not exported function" called `treatUnresolvedAfter`; that can be used like this
```
    resolutionValue = await Promise.race([
        promise,
        treatUnresolvedAfter(TIME_TO_WAIT_UNTIL_UNRESOLVED),
      ]);
```
but creating the timeout fct like this looks good too

packages/assets-controllers/src/AccountTrackerController.ts
```diff
@@ -630,21 +635,38 @@ export class AccountTrackerController extends StaticIntervalPollingController<Ac
...
+            );
             remainingChains = remainingChains.filter(
               (chain) => !processedChains.has(chain),
             );
           }
+
+          // Add unprocessed chains back to remainingChains for next fetcher
+          if (
+            result.unprocessedChainIds &&
+            result.unprocessedChainIds.length > 0
```
Prithpal-Sooriya, line 659: These chains are when the current fetcher fails to process? Nice!

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -603,11 +606,38 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
...
+      );
+
+      // Race between the API call and the timeout
+      const apiCallPromise = this.#addDetectedTokensViaAPI({
+        chainIds: chainsToDetectUsingAccountAPI,
+        selectedAddress: addressToDetect,
+        supportedNetworks,
+      });
+
+      return await Promise.race([apiCallPromise, timeoutPromise]);
```
Prithpal-Sooriya: Hmm can we reuse our existing utils? E.g.
```ts
  async #attemptAccountAPIDetection(
    chainsToDetectUsingAccountAPI: Hex[],
    addressToDetect: string,
    supportedNetworks: number[] | null,
  ) {
    // from controller-utils
    const result = await safelyExecuteWithTimeout(
      async () => {
        return this.#addDetectedTokensViaAPI({
          chainIds: chainsToDetectUsingAccountAPI,
          selectedAddress: addressToDetect,
          supportedNetworks,
        });
      },
      false,
      ACCOUNTS_API_TIMEOUT_MS,
    );

    if (!result) {
      return { result: 'failed' };
    }

    return result;
  }
```

salimtb: ooh true , comment addressed here [e5b43d0](https://github.com/MetaMask/core/pull/7106/commits/e5b43d00d7e845b4bce1c3b59b57bf08217c68fc)

salimtb:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #7114, opened 2025-11-11
ci: Check if any changes made ahead of PR affect a release to support merge queue

Mrtenz:
## Explanation

This adds the `merge_group` target and adds a check for releases, so we can safely enable the merge queue.

The release check will check for any differences on `main` or in the merge queue compared to the release branch. If any conflicts are detected, the release will be rejected from the merge queue, and will need to be updated with the latest changes first.

## Checklist

---

> [!NOTE]
> Add merge queue support and a release-conflict check action integrated into the main workflow for release PRs.
> 
> - **CI/Workflows**:
>   - Add `merge_group` trigger to `.github/workflows/main.yml`.
>   - Introduce `check-release` job to detect PR/merge-queue releases and run conflict checks.
>     - Uses `MetaMask/action-is-release` to detect release PRs and extracts PR number (supports merge queue refs).
>     - Wires results into overall job dependencies (`all-jobs-complete` now needs `check-release`).
> - **New Composite Action**:
>   - Add `.github/actions/check-release/action.yml` to detect release conflicts across packages.
>     - Identifies packages with version bumps vs `main`, computes changed files ahead of the PR/merge group target, and errors on conflicts.
>     - Supports `pull_request` and `merge_group` events; fetches PR branch via `gh` and compares against calculated target.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit babd67a7796e0f8638a4ac6ffaac673f601b4d85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/check-release/action.yml
+++ b/.github/actions/check-release/action.yml
@@ -25,7 +25,7 @@ runs:
         elif [ "$EVENT_NAME" = "merge_group" ]; then
           echo "TARGET=$(git rev-parse HEAD^)" >> "$GITHUB_OUTPUT"
         else
-          echo "::error::This action only supports `pull_request` and `merge_group` events."
+          echo "::error::This action only supports \`pull_request\` and \`merge_group\` events."
           exit 1
         fi
 


.github/workflows/main.yml
```diff
@@ -62,6 +63,25 @@ jobs:
     needs: check-workflows
     uses: ./.github/workflows/lint-build-test.yml
 
+  check-release:
+    name: Check release
+    needs: check-workflows
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if the commit or pull request is a release
```
Mrtenz, line 84: I considered reusing the `is-release` job below, but it adds a little bit of complexity around the conditionals, and we don't need the extra options below for merged commits.

.github/workflows/main.yml
```diff
@@ -62,6 +63,25 @@ jobs:
...
+        id: is-release
+        uses: MetaMask/action-is-release@d063725cd15ee145d7e795a2e77db31a3e3f2c92
+        with:
+          commit-starts-with: 'Release [version],Release v[version],Release/[version],Release/v[version],Release `[version]`'
+          commit-message: ${{ github.event.pull_request.title }}
```
Mrtenz, line 90: This is used so we can determine if an unmerged pull request is a release pull request. `github.event.pull_request.title` is not defined for other events (`push` and `merge_group`), so doesn't affect those.

.github/workflows/main.yml
```diff
@@ -62,6 +63,25 @@ jobs:
...
+          before: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha || github.event.before }}
```
Mrtenz: `github.event.before` is only defined on `push`, not on `pull_request` or `merge_group`, so I had to add this.

.github/workflows/main.yml
```diff
@@ -62,6 +63,25 @@ jobs:
     needs: check-workflows
     uses: ./.github/workflows/lint-build-test.yml
 
+  check-release:
```
Mrtenz, line 66: I didn't add a conditional here, but only to the step below. This way we can add `check-release` as dependency to `all-jobs-complete`. Otherwise, if `check-release` (the job) is skipped, `all-jobs-complete` would be skipped too.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,63 @@
...
+
+          if [ "$HEAD_VERSION" != "$MAIN_VERSION" ]; then
+            package_name=$(jq -r ".name" "$package")
+            echo "📦 Package \`$package_name\` is being released (version \`$MAIN_VERSION\` -> \`$HEAD_VERSION\`)"
+            RELEASED_PACKAGES+=("$package")
+          fi
+        done
+
+        # Get all files changed ahead of this PR.
+        git diff --name-only "$BEFORE..$(git merge-base HEAD refs/remotes/origin/main)" > changed-files.txt
```
Mrtenz: From my understanding, `git merge-base HEAD refs/remotes/origin/main` finds the common ancestor between `HEAD` (i.e., the current commit), and `origin/main`. We can then diff `$BEFORE` with that to find any differences on `main` not included in the current branch.

.github/actions/check-release/action.yml
Mrtenz, line 1: This is an action rather than a workflow to support the `uses` syntax in a step, rather than as a job. It's related to [this comment](https://github.com/MetaMask/core/pull/7114/files#r2518406779).

.github/workflows/main.yml
Gudahtt: `base.sha` doesn't seem correct here.

In the `action-is-release` action, we used `github.event.before` to get the diff of what changed in a given commit, to check whether the version number had changed. For the `push` event, the diff betweeen `github.event.before` and the current commit would only include changes from that push event.

Whereas `base.sha` refers to the commit on the base branch that the PR was originally branched from. It does not get updated when the branch is updated with a merge commit. That means if a core release goes out while a PR is open, `github.event.pull_request.base.sha..HEAD` will include a version bump from the unrelated core release on `main`.

Instead I think we need the merge-base here, between the base branch and the current commit

.github/workflows/main.yml
Gudahtt: Nit: We could maybe skip this for push events? Just because it wouldn't be relevant anymore for `push`, and this might make it a bit more clear why we have to similarly-named jobs

.github/workflows/main.yml
Gudahtt: Either that, or get the merge-base in `action-is-release` instead

.github/workflows/main.yml
Mrtenz: We could skip the individual steps on `push`, but not the job itself, since that would cause `all-jobs-complete` to be skipped.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,63 @@
...
+      run: |
+        set -euo pipefail
+        git fetch origin main
+
+        mapfile -t PACKAGES < <(find packages -maxdepth 2 -name "package.json" -not -path "*/node_modules/*")
+        RELEASED_PACKAGES=()
+
+        # Get all packages being released in this PR
+        for package in "${PACKAGES[@]}"; do
+          MAIN_VERSION=$(git show "refs/remotes/origin/main:$package" | jq -r .version)
```
Gudahtt: This comparison doesn't seem quite right. We're using `origin/main` as the "before the current commit" comparison point, but there might be more changes in-between `origin/main` and the current commit. e.g. if there is a release PR queued before the current one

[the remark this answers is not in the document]
Mrtenz: 2f0340c7f65e631ba3911da46e98c35f61979187

Mrtenz: 2f0340c7f65e631ba3911da46e98c35f61979187

.github/workflows/main.yml
```diff
@@ -62,6 +63,39 @@ jobs:
...
+        run: |
+          echo "MERGE_BASE=$(git merge-base HEAD "refs/remotes/origin/$BASE_REF")" >> "$GITHUB_OUTPUT"
+
+      - name: Check if the commit or pull request is a release
+        id: is-release
+        if: github.event_name != 'push'
+        uses: MetaMask/action-is-release@d063725cd15ee145d7e795a2e77db31a3e3f2c92
+        with:
+          commit-starts-with: 'Release [version],Release v[version],Release/[version],Release/v[version],Release `[version]`'
+          commit-message: ${{ github.event.pull_request.title }}
```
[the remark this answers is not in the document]
Mrtenz: This option is only needed for pull requests. Otherwise we can simply look at the commit message.

.github/workflows/main.yml
```diff
@@ -62,6 +63,39 @@ jobs:
...
+          before: ${{ steps.merge-base.outputs.MERGE_BASE }}
+
+      - name: Check release
+        if: github.event_name != 'push' && steps.is-release.outputs.IS_RELEASE == 'true'
+        uses: ./.github/actions/check-release
+        with:
+          before: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha }}
```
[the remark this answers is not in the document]
Gudahtt: Cursor didn't explain this super well, but, I think this comment is at least partially accurate.

If you didn't use any merge commits, `base.sha` is exactly what we want here I think (at least for the `pull_request` event, not 100% sure on `merge_group`), but merges break that assumption.

Essentially what we want to compare in `check-release` is the _previous_ merge base with the _current_ merge base. We want to compare _a_ merge base for both sides of the comparison.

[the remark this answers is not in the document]
Gudahtt: For the `merge_group` event, would `base.sha` refer to the previous commit in the queue itself? Even the merge-base of `base.sha` might not refer to the correct commit if that's the case.

We might need to consider something like.... grab the PR number from the merge-queue branch name, fetch the branch for that PR, then get the merge-base of that branch.

I found a discussion where someone talked about how to get the PR number from the merge queue branch name: https://github.com/orgs/community/discussions/120203
(in service of a different goal, of skipping the queue for PRs that were already up-to-date, which also would be pretty neat)

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,70 @@
...
+          fi
+        done
+
+        # Get all files changed ahead of this PR.
+        PULL_REQUEST_BRANCH=$(gh pr view "$PULL_REQUEST" --json headRefName --template "{{ .headRefName }}")
+
+        echo "🔍 Checking for release conflicts with files changed ahead of PR branch \`$PULL_REQUEST_BRANCH\`..."
+
+        git fetch origin "$PULL_REQUEST_BRANCH"
+        BEFORE=$(git merge-base "refs/remotes/origin/main" "$PULL_REQUEST_BRANCH")
```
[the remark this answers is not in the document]
Gudahtt: It's fetched on the previous line

.github/workflows/main.yml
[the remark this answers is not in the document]
Gudahtt: >Essentially what we want to compare in `check-release` is the _previous_ merge base with the _current_ merge base. We want to compare _a_ merge base for both sides of the comparison.

Sorry I made a mistake here.

The previous merge base is definitely the correct point of comparison, but the current merge base isn't. That would exclude other changes in the queue from the comparison. We need to be comparing the previous merge base to the "commit before whatever is in the queue", which I believe is just the previous commit given that we use squash & merge

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,70 @@
...
+        git diff --name-only "$BEFORE..refs/remotes/origin/main" > changed-files.txt
```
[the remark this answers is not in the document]
Gudahtt: Cursor is right here I believe. We had a discussion about this where I made a bad suggestion that is responsible for this problem I think, context: https://github.com/MetaMask/core/pull/7114#discussion_r2524370412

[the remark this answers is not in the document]
Gudahtt: I believe the `$BEFORE` commit is correct, but the commit we should be comparing it to here the change _ahead_ of this PR in the merge queue (or `main`, if this is first in the queue), which should be equivalent to `HEAD^` for repos like this where Squash & Merge is the only allowed merge type (if merge commits were allowed, we'd have to find the merge-base between HEAD and the merge queue branch itself instead, that's the more generic approach).

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,87 @@
...
+        fetch-depth: 0
+
+    - name: Get target reference
+      id: get-target
+      shell: bash
+      env:
+        EVENT_NAME: ${{ github.event_name }}
+      run: |
+        if [ "$EVENT_NAME" = "pull_request" ]; then
+          echo "TARGET=$(git merge-base HEAD refs/remotes/origin/main)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
Gudahtt: The merge-base of HEAD and `main` should nearly always be identical to `main`, except in the case where the a new commit was made to `main` since the workflow began. In that scenario, the merge base of HEAD is correct here.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,87 @@
...
+        elif [ "$EVENT_NAME" = "merge_group" ]; then
+          echo "TARGET=$(git rev-parse HEAD^)" >> "$GITHUB_OUTPUT"
+        else
+          echo "::error::This action only supports `pull_request` and `merge_group` events."
```
Gudahtt: When running this through shellcheck.net, it's suggesting that Bash is trying to execute what's in these backticks. Escaping it should ensure they don't get interpreted as a command.

Alternatively we could use a single quotes here instead

```suggestion
          echo "::error::This action only supports \`pull_request\` and \`merge_group\` events."
```

Gudahtt (commented):
Looks great! Just one minor suggestion

.github/actions/check-release/action.yml
Mrtenz: Good catch, fixed in https://github.com/MetaMask/core/pull/7114/commits/375f5ed6dce3c88158eab50eb7a2ca5e64a93469.

PR #7119, opened 2025-11-11
feat: price API v3 upgrade

bergarces:
## Explanation

Upgrade to Price API v3

Preview PR for extension: https://github.com/MetaMask/metamask-extension/pull/37741
Preview PR for mobile: https://github.com/MetaMask/metamask-mobile/pull/22876

## References

Fixes https://consensyssoftware.atlassian.net/browse/ASSETS-1746

## Checklist

---

> [!NOTE]
> Upgrades token price fetching to Price API v3, updates service/types and controllers to new assets-based API, and removes legacy polling paths.
> 
> - **Price API Integration (v3)**
>   - `CodefiTokenPricesServiceV2`: add v3 `/v3/spot-prices` with CAIP-19 `assetId`, fallback to v2 per-chain endpoint; expand `SUPPORTED_CURRENCIES`; introduce `SPOT_PRICES_SUPPORT_INFO`; handle native token addresses; circuit-breaker/degraded handling retained.
> - **Types/Interfaces**:
>   - BREAKING: `AbstractTokenPricesService.fetchTokenPrices` now accepts `assets: { chainId, tokenAddress }[]` and returns `EvmAssetWithMarketData[]` (replaces address->object map and adds `assetId`).
>   - Add `EvmAssetAddressWithChain`, `EvmAssetWithId`, `EvmAssetWithMarketData`.
> - **Controllers**:
>   - `TokenRatesController`: fetch prices grouped by native currency; support unsupported native currencies via USD conversion; include native token in queries; remove legacy polling/state/event deps; handle network deletions; enable/disable gating maintained.
>   - `CurrencyRateController`: on API failure, fallback to spot prices via `fetchTokenPrices` using native token asset; map multiple networks by native currency.
>   - `TokenSearchDiscoveryDataController`: adapt to new `fetchTokenPrices` response (array) and asset inputs.
> - **Utils**:
>   - `assetsUtil.fetchTokenContractExchangeRates`: switch to assets-based batching (includes native token) and map array response.
> - **Tests/Changelog**:
>   - Update tests to new arguments/returns and behaviors; changelog notes BREAKING changes and polling cleanup.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3062143315cfdbca1828f88723827d94e392aa4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


bergarces:
@metamaskbot publish-preview

bergarces:
@metamaskbot publish-preview

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -474,64 +492,68 @@ export class CodefiTokenPricesServiceV2
...
+              ? HEX_CHAIN_ID_TO_CAIP19_NATIVE_ASSET_MAP[asset.chainId]
+              : `${caipChainId}/erc20:${asset.tokenAddress.toLowerCase()}`,
+        };
+      },
+    );
+
+    const url = new URL(`${BASE_URL_V3}/spot-prices`);
     url.searchParams.append(
-      'tokenAddresses',
-      [getNativeTokenAddress(chainId), ...tokenAddresses].join(','),
```
bergarces, line 496: The service no longer gets a chain or list of chains, so it is not responsible of also adding native assets, those assets needs to be passed as part of the list.

packages/assets-controllers/src/token-prices-service/abstract-token-prices-service.ts
```diff
@@ -1,31 +1,7 @@
...
-  priceChange1d: number;
-  pricePercentChange1d: number;
-  pricePercentChange1h: number;
-  pricePercentChange1y: number;
-  pricePercentChange7d: number;
-  pricePercentChange14d: number;
-  pricePercentChange30d: number;
-  pricePercentChange200d: number;
-  totalVolume: number;
-};
```
bergarces, line 28: Unnecessary duplicate from TokenRatesController.

packages/assets-controllers/src/token-prices-service/abstract-token-prices-service.ts
```diff
@@ -55,43 +31,50 @@ export type ExchangeRatesByCurrency<Currency extends string> = {
...
+};
+
+export type EvmAssetWithId<ChainId extends Hex = Hex> =
+  EvmAssetAddressWithChain<ChainId> & {
+    assetId: CaipAssetType;
+  };
+
+export type EvmAssetWithMarketData<
+  ChainId extends Hex = Hex,
+  Currency extends string = string,
```
bergarces, line 36: I don't think there's any specific benefit on adding ChainId and Currency generics here, since we are unlikely to have properly defined types for the chains and currencies that we pass from the client (they'll be defined as Hex and string there anyway).

I just added them to the new types because these whole abstract class seems to be depending on them heavily and the previous types already had them (for no real benefit either IMO).

If anyone else thinks they are confusing and add no value, I'll be happy to remove them.

packages/assets-controllers/src/TokenSearchDiscoveryDataController/TokenSearchDiscoveryDataController.ts
```diff
@@ -172,22 +171,18 @@ export class TokenSearchDiscoveryDataController extends BaseController<
...
     try {
       const pricesData = await this.#tokenPricesService.fetchTokenPrices({
-        chainId,
-        tokenAddresses: [address as Hex],
+        assets: [{ chainId, tokenAddress: address as Hex }],
         currency: currentCurrency,
       });
 
-      return pricesData[address as Hex] ?? null;
+      return pricesData[0] ?? null;
```
bergarces, line 185: We are only sending one asset, so it's either in position 0 or not.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -320,25 +298,7 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-            return {
-              chainId: chainId as Hex,
-              nativeCurrency,
-            };
-          },
-        );
-
-        if (this.#pollState === PollState.Active) {
-          await this.updateExchangeRates(chainIdAndNativeCurrency);
-        }
```
bergarces, line 340: I removed all this code because this was never being called, as `#pollState` was always disabled.

Assets for specific chains are updated individually in the UI (at least in extension) when the network filter is changed or a new network is added.

So the only thing this handler has been doing was remove chain data when deleting chains.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -370,7 +330,13 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
     const tokenAddresses = getTokens(this.#allTokens[chainId]);
     const detectedTokenAddresses = getTokens(this.#allDetectedTokens[chainId]);
 
-    return [...new Set([...tokenAddresses, ...detectedTokenAddresses])].sort();
+    return [
+      ...new Set([
+        ...tokenAddresses,
+        ...detectedTokenAddresses,
+        getNativeTokenAddress(chainId),
```
bergarces, line 337: We add the native token here, as at this point we are still aware of what chains we are dealing with. After this, all the assets are split in batches and one batch might include multiple chains, making this operation more complicated.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-        } catch (error: unknown) {
-          updateFailed(error);
-          throw error;
-        } finally {
-          // Cleanup the tracking for this update.
-          delete this.#inProcessExchangeRateUpdates[updateKey];
-        }
-      },
-    );
+    await Promise.allSettled(promises);
```
bergarces, line 436: We were already doing `Promise.allSettled` before to fetch all chains at once. Now we have less calls than before as we put together multiple chains with the same native currency, so it shouldn't be an issue.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
         }
+      }
+    }
 
-        // Create a deferred promise to track this update.
-        const {
-          promise: inProgressUpdate,
-          resolve: updateSucceeded,
-          reject: updateFailed,
-        } = createDeferredPromise({ suppressUnhandledRejection: true });
```
bergarces, line 506: I have tried by different means in the UI triggering multiple identical fetches and I haven't been able to, so I don't think that storing the promises to reuse them is necessary.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-        return obj;
-      }, {});
-    }
+  async #fetchAndMapExchangeRatesForSupportedNativeCurrency(
+    assets: {
+      chainId: Hex;
+      tokenAddress: Hex;
+    }[],
+    currency: string,
+    marketData: Record<Hex, Record<Hex, MarketDataDetails>> = {},
```
bergarces, line 464: Controversial, but passing the global marketData object and mutate it directly saves us juggling with the returned results quite a bit.

Doing this is safe with parallel calls because the entry for the chain (first key) is created safely with `??=` and the token address (second key) will not be repeated in different batches.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
+    }[],
+    currency: string,
+    marketData: Record<Hex, Record<Hex, MarketDataDetails>>,
+  ) {
+    // Step -1: Then fetch all tracked tokens priced in USD
+    const marketDataInUSD =
+      await this.#fetchAndMapExchangeRatesForSupportedNativeCurrency(
+        assets,
+        'usd', // Fallback currency when the native currency is not supported
+      );
```
bergarces, line 505: This call recovers both tokens and native asset for the chain, so it saves us from doing two calls.

packages/assets-controllers/src/token-prices-service/abstract-token-prices-service.ts
```diff
@@ -38,60 +14,57 @@ export type ExchangeRate = {
...
 /**
  * An ideal token prices service. All implementations must confirm to this
  * interface.
  *
  * @template ChainId - A type union of valid arguments for the `chainId`
  * argument to `fetchTokenPrices`.
- * @template TokenAddress - A type union of all token addresses. The reason this
- * type parameter exists is so that we can guarantee that same addresses that
- * `fetchTokenPrices` receives are the same addresses that shown up in the
- * return value.
```
bergarces, line 67: The generic for `TokenAddress` is completely useless in this case because we do not have a range of addresses on our types, it can be any valid ethereum address. We always set it to `Hex`

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -144,6 +152,52 @@ export const SUPPORTED_CURRENCIES = [
...
+  'pepu',
+  // Polygon
+  'pol',
+  // Mantle
+  'mnt',
+  // Onomy
+  'nom',
+  // Avalanche
+  'avax',
 ] as const;
```
bergarces, line 201: Added all currencies that we were missing. Extracted from the API docs directly.

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -174,6 +228,58 @@ const chainIdToNativeTokenAddress: Record<Hex, Hex> = {
...
+
+// MISSING CHAINS WITH NO NATIVE ASSET PRICES IN V2
+// '0x42': 'eip155:66/slip44:996', // OKXChain Mainnet - Native symbol: OKT
+// '0x46': 'eip155:70/slip44:1170', // Hoo Smart Chain - Native symbol: HOO
+// '0x7a': 'eip155:122/slip44:XXX', // Fuse Mainnet - Native symbol: FUSE
+// '0x120': 'eip155:288/slip44:60', // Boba Network (Ethereum L2) - Native symbol: ETH
+// '0x150': 'eip155:336/slip44:809', // Shiden - Native symbol: SDN
+// '0x440': 'eip155:1088/slip44:60', // Metis Andromeda Mainnet (Ethereum L2) - Native symbol: ETH
+// '0x1388': 'eip155:5000/slip44:XXX', // Mantle - Native symbol: MNT
+// '0x28c58': 'eip155:167000/slip44:60', // Taiko Mainnet - Native symbol: ETH
```
bergarces, line 281: This is the list of chains that was in theory supported by price api v2, but no native prices were returned when passing the zero address, which is what we would have passed for v2.

I also couldn't find the equivalent native address, so in practice they were not supported and will continue not being supported.

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -174,6 +228,56 @@ const chainIdToNativeTokenAddress: Record<Hex, Hex> = {
...
+  '0xe708': 'eip155:59144/slip44:60', // Linea Mainnet - Native symbol: ETH
+  '0x13c31': null, // 'eip155:81457/slip44:60', // Blast Mainnet - Native symbol: ETH
+  '0x17dcd': null, // 'eip155:97741/slip44:XXX', // Pepe Unchained Mainnet - Native symbol: PEPU
+  '0x518af': null, // 'eip155:333999/slip44:1997', // Polis Mainnet - Native symbol: POLIS
+  '0x82750': 'eip155:534352/slip44:60', // Scroll Mainnet - Native symbol: ETH
+  '0x4e454152': 'eip155:60/slip44:60', // Aurora Mainnet (Ethereum L2 on NEAR) - Native symbol: ETH
+  '0x63564c40': 'eip155:1666600000/slip44:1023', // Harmony Mainnet Shard 0 - Native symbol: ONE
+  '0x279f': 'eip155:143/slip44:268435779', // Monad Testnet - Native symbol: MON
+  '0x3e7': 'eip155:999/slip44:2457', // HyperEVM - Native symbol: ETH
+} as const;
```
bergarces, line 273: Any entry here with null, means that the support for price-api v2 works but it doesn't for v3, as price-api does not recognise the native asset slip44 reference, which we need to fetch prices.

bergarces: After checking with platform-api. They have confirmed that if the entry worked with the zero address on v2, it should work by passing zero address as a token for v3. Not having the slip44 reference is not an issue then, as we are only using this internally.

There are, however, 2 chains for which that is not true: Meter and Polis.

There's also Monad, which does not seem to work at all, but that's probably because it's not on until next week.

I'm keeping the fallback to v2 for those chains, but as soon as they are sorted at backend level we can get rid of any v2 code.

bergarces:
@metamaskbot publish-preview

packages/assets-controllers/src/TokenRatesController.test.ts
Prithpal-Sooriya, line 1: TY for the test cleanup!

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-          resolve: updateSucceeded,
-          reject: updateFailed,
-        } = createDeferredPromise({ suppressUnhandledRejection: true });
-        this.#inProcessExchangeRateUpdates[updateKey] = inProgressUpdate;
-
-        try {
-          const contractInformations = await this.#fetchAndMapExchangeRates({
-            tokenAddresses,
-            chainId,
+    const promises = [
```
Prithpal-Sooriya, line 417: Whats the number of calls/promises we are generating here?
If this is large, thoughts on using `reduceInBatchesSerially` or one of the other batching mechanisms we have to process multiple promises in parallel?

packages/assets-controllers/src/TokenRatesController.ts
Prithpal-Sooriya: niiiiiice!!

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
Prithpal-Sooriya: Perfeeect

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -474,64 +437,149 @@ export class CodefiTokenPricesServiceV2
...
+    const assetsByChainId: Record<SupportedChainId, Hex[]> =
+      v2SupportedAssets.reduce(
+        (acc, { chainId, tokenAddress }) => {
+          (acc[chainId] ??= []).push(tokenAddress);
+          return acc;
+        },
+        {} as Record<SupportedChainId, Hex[]>,
+      );
+
+    const promises = Object.entries(assetsByChainId).map(
```
Prithpal-Sooriya, line 537: Same here. Are there many promises in parallel here?

If yes, thoughts on using the `reduceInBatchesSerially` or other batching utils so we process certain amount at a time?

Also unsure if we should add a timeout mechanism `safelyExecuteWithTimeout`, for if this API hangs for a long time?

Prithpal-Sooriya (approved):
Approved with 2 comments on adding some batching logic on our parallel promises

packages/assets-controllers/src/TokenRatesController.ts
bergarces: At most, one for every different currency. Up until now, we were generating one for every different chain and awaiting them as well with `Promise.allSettled`.

In a standard scenario with all the popular chains enabled, we'll be sending 5 promises in parallel (ETH, POL, BNB, SEI and AVAX).

I think we are still far from the number of chains that would require careful batching.

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
bergarces: Maximum of 3, assuming that the user has Monad, Meter and Polis networks enabled (the last two are rather obscure).

We should also be expecting to remove this fallback soon, because we will be able to test Monad for the correct caip19 asset when it becomes available next week. The other two chains I have already reported them to platform-api.

bergarces: They are called with `this.#policy.execute`, which has its own for retries, duration, etc.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -421,194 +367,195 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-        tokenAddresses,
-        chainId,
-        nativeCurrency,
-      });
-    }
+        for (const tokenPrice of batchMarketData) {
+          (partialMarketData[tokenPrice.chainId] ??= {})[
+            tokenPrice.tokenAddress
+          ] = tokenPrice;
+        }
```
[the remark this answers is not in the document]
bergarces: Does not matter, the return type includes it.

Issue #6010, opened 2025-06-19
[KeyringController] Drop uncached encryption support


mikesposito:
The KeyringController currently supports both cached and uncached encryption. This is a legacy feature that is no longer needed, as all clients have this enabled by default. 

Specifically, when `cacheEncryptionKey` is set to `false`, the KeyringController will not cache the encryption key, which is the default behavior in all current clients. Removing this feature simplifies the codebase and reduces complexity of critical parts of code responsible of keeping the vault secure.

PR #5963, opened 2025-06-11
refactor!: drop uncached encryption support

mikesposito:
## Explanation

Previously, `KeyringController` accepted a `cacheEncryptionKey` option that allowed the encryption key to be stored in memory and used during encryption/decryption directly as opposed to using a password. The `cacheEncryptionKey` option is being removed, and the encryption key is now always derived and cached when the password is provided.

This change allows to simplify `#unlockKeyrings` and `#updateVault` methods, and remove all the logic and tests related to `cacheEncryptionKey`. This also allows to remove `this.#password`, that has been replaced by `this.#encryptionKey`. 

The `this.#encryptionKey` assignment logic has been moved to two new internal methods with these specific responsibilities:
- `#deriveEncryptionKey(string)`: Derives the encryption key from the password, to be used during password login and password change.
- `#useEncryptionKey(string, string)`: Uses an existing encryption key to be used directly, to be used by `submitEncryptionKey` mainly.

With the upcoming changes in #5940, this allows to change the encryption key to use (i.e. by calling the aformentioned new internal methods) without having to deal with logic related to vault unlock/update, and code branches related to password-based encryption and key caching.

This PR can be tested on extension with the following: https://github.com/MetaMask/metamask-extension/pull/33613

## References

* Fixes #6010

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary
- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

---

> [!NOTE]
> Reworks encryption to always use a provided encryptor and key-based flow, removes `cacheEncryptionKey`, unifies encryptor types, adds generics, and aligns seedless onboarding types; updates tests and deps.
> 
> - **Keyring Controller**:
>   - **API/Types**: Require `encryptor` in constructor; merge `GenericEncryptor`/`ExportableKeyEncryptor` into unified `Encryptor`; add generics (`EncryptionKey`, `SupportedKeyDerivationOptions`, `EncryptionResult`).
[...]

PR #7127, opened 2025-11-12
refactor!: remove `cacheEncryptionKey` flag from `KeyringController`

mikesposito:
## Explanation

This PR includes part of the changes originally proposed in [this PR](https://github.com/MetaMask/core/pull/5963) that were inflating its size:
  - Remove the `cacheEncryptionKey` flag from `KeyringController` constructor options and instance variables
    - All code paths that used this flag have been removed (effectively, the controller behaves as if the flag was always `true`)
  - `GenericEncryptor` and `ExportableKeyEncryptor` have been merged into a single `Encryptor` type.
    - The `Encryptor` type has also been fixed to be compatible with `@metamask/browser-passworder` and `Encryptor` from mobile
  - As `Encryptor` now accepts type parameters to specify encryption key and derivation params types, the `KeyringController` has been updated to accept these type parameters and pass them to the `Encryptor` instance.

## References

* Related to #6010

## Checklist

---

> [!NOTE]
> Replaces Generic/Exportable encryptors with a single generic `Encryptor`, requires an `encryptor` in `KeyringController`, removes `cacheEncryptionKey`, and threads encryption generics through both controllers with tests and mocks updated.
> 
> - **Keyring Controller (BREAKING)**:
>   - Require `encryptor` in `KeyringController` constructor; remove `cacheEncryptionKey` option.
>   - Merge `GenericEncryptor` and `ExportableKeyEncryptor` into unified `Encryptor` with generics: `EncryptionKey`, `SupportedKeyDerivationOptions`, `EncryptionResult`.
>   - Update types and methods to use generic `Encryptor` (e.g., `decryptWithKey` now accepts encryption result object); remove exportability assertions and conditional branches tied to caching.
>   - Make `KeyringController` and `KeyringControllerOptions` generic over encryption types; consistently set/clear `encryptionKey`/`encryptionSalt`.
>   - Tests refactored to drop `cacheEncryptionKey` paths, add `createVault` helper, adjust mocks; tweak coverage thresholds.
> - **Seedless Onboarding Controller**:
>   - Add optional `SupportedKeyDerivationOptions` generic to controller and `VaultEncryptor`; update types, tests, and mocks accordingly.
> - **Changelogs**:
>   - Document breaking changes and new generics for both packages; note `decryptWithKey` type fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3ebcbb5e23ff72223b111716ba14be8d796899e7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/keyring-controller/tests/mocks/mockEncryptor.ts
```diff
@@ -41,15 +41,16 @@ export default class MockEncryptor implements ExportableKeyEncryptor {
   async decrypt(password: string, text: string): Promise<Json> {
     const { salt } = JSON.parse(text);
     const key = deriveKey(password, salt);
-    return await this.decryptWithKey(key, text);
+    const payload = JSON.parse(text);
```
FrederikBolding, line 42: We already parse above, perhaps we can combine the two?

mikesposito: Ah, good catch! done in [9dd2aae](https://github.com/MetaMask/core/pull/7127/commits/9dd2aaecb473198baa0e00b21d9dd03baf8effcd)

packages/keyring-controller/jest.config.js
```diff
@@ -17,10 +17,10 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 95.78,
+      branches: 94.03,
       functions: 100,
-      lines: 98.68,
-      statements: 98.69,
+      lines: 98.31,
+      statements: 98.33,
```
matthiasgeihs, line 23: Why decrease? 😭
Shouldn't this only go up? 😃

packages/keyring-controller/tests/mocks/mockEncryptor.ts
```diff
@@ -67,8 +67,9 @@ export default class MockEncryptor implements ExportableKeyEncryptor {
   ): Promise<DetailedDecryptResult> {
     const { salt } = JSON.parse(text);
     const key = deriveKey(password, salt);
+    const payload = JSON.parse(text);
     return {
-      vault: await this.decryptWithKey(key, text),
+      vault: await this.decryptWithKey(key, payload),
```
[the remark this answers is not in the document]
mikesposito: Done in [4bb2159](https://github.com/MetaMask/core/pull/7127/commits/4bb21596e07c893d3afa69edd5ee37290c0bfcf0)

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -735,13 +729,6 @@ export class KeyringController extends BaseController<
     this.#keyrings = [];
     this.#unsupportedKeyrings = [];
 
-    // This option allows the controller to cache an exported key
-    // for use in decrypting and encrypting data without password
-    this.#cacheEncryptionKey = Boolean(options.cacheEncryptionKey);
-    if (this.#cacheEncryptionKey) {
-      assertIsExportableKeyEncryptor(encryptor);
```
FrederikBolding, line 742: Is the idea that we would rely on types exclusively for this going forward?

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -335,11 +336,28 @@ type SessionState = {
...
- * serializable data with a password.
+ * An encryptor interface that supports encrypting and decrypting
+ * serializable data with a password, and exporting and importing keys.
  */
-export type GenericEncryptor = {
+export type Encryptor<
+  EncryptionKey = encryptorUtils.EncryptionKey | CryptoKey,
+  SupportedKeyDerivationParams = encryptorUtils.KeyDerivationOptions,
+  EncryptionResult extends
+    EncryptionResultConstraint<SupportedKeyDerivationParams> = DefaultEncryptionResult<SupportedKeyDerivationParams>,
```
FrederikBolding, line 359: Have you ensured that the types are compatible with the Encryptor on mobile?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1,5 +1,6 @@
 import { keccak256AndHexify } from '@metamask/auth-network-utils';
 import { BaseController, type StateMetadata } from '@metamask/base-controller';
+import type * as encryptionUtils from '@metamask/browser-passworder';
```
matthiasgeihs, line 3: Does that make `@metamask/browser-passworder` a dependency of `seedless-onboarding-controller`? Would it make sense to expose `encryptionUtils.KeyDerivationOptions` as part of `keyring-controller`?

packages/keyring-controller/src/KeyringController.ts
mikesposito: Yes, mobile can inject its own EncryptionKey and KeyMetadata type parameters: https://github.com/MetaMask/metamask-mobile/pull/22240/files#diff-80bb7ad3e69af643ad6f36ac52be4481bb9383293d764fa3de48d82fb7d9a2dc

packages/keyring-controller/src/KeyringController.ts
mikesposito: Yes. Types were broken and also used to depend on `cacheEncryptionKey`. Now that the Encryptor type makes more sense I felt like we can rely more on types than runtime checks (which, by the way, were only silencing type errors)

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -335,11 +336,28 @@ type SessionState = {
...
+  salt?: string;
+  keyMetadata?: SupportedKeyMetadata;
+};
+
+export type DefaultEncryptionResult<SupportedKeyMetadata> = {
+  data: string;
+  iv: string;
+  salt?: string;
+  keyMetadata?: SupportedKeyMetadata;
+};
```
matthiasgeihs, line 349: I'm a bit confused about the different EncryptionResult types here:
Why don't we have a unified `EncryptionResult` interface?
Can you give some context?

packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -2615,600 +2552,453 @@ describe('KeyringController', () => {
...
-                await controller.submitPassword(password);
-
-                expect(encryptSpy).toHaveBeenCalledTimes(1);
+                type: KeyringTypes.hd,
+                data: {},
               },
-            );
-          });
-
-        cacheEncryptionKey &&
```
FrederikBolding, line 3042: Yikes, glad we can remove this 😄

packages/keyring-controller/CHANGELOG.md
```diff
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING:** The `GenericEncryptor` and `ExportableKeyEncryptor` types have been merged into a single `Encryptor` type ([#7127](https://github.com/MetaMask/core/pull/7127))
+
+### Removed
+
+- **BREAKING:** The `cacheEncryptionKey` parameter has been removed from the `KeyringController` constructor options ([#7127](https://github.com/MetaMask/core/pull/7127))
+  - This parameter was previously used to enable encryption key in-memory caching, but it is no longer needed as the controller now always uses the latest encryption key.
+
+### Fixed
+
+- Fixed incorrect type for `decryptWithKey` method of `ExportableKeyEncryptor` ([#7127](https://github.com/MetaMask/core/pull/7127))
```
FrederikBolding: Which now no longer exists 😅 

How about something like this?
```suggestion
- Fixed incorrect type for `decryptWithKey` method of `ExportableKeyEncryptor` (now `Encryptor`) ([#7127](https://github.com/MetaMask/core/pull/7127))
```

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2061,50 +2046,36 @@ export class KeyringController extends BaseController<
...
-            key,
-            parsedEncryptedVault,
-          );
-
-          // This call is required on the first call because encryptionKey
-          // is not yet inside the memStore
-          updatedState.encryptionKey = encryptionKey;
-          updatedState.encryptionSalt = encryptionSalt;
+          encryptionSalt = parsedEncryptedVault.salt as string;
         }
```
matthiasgeihs, line 2066: nit

```ts
if (encryptionSalt && encryptionSalt !== parsedEncryptedVault.salt) {
  throw new Error(KeyringControllerError.ExpiredCredentials);
}

encryptionSalt = parsedEncryptedVault.salt as string;
```

packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -2615,600 +2552,453 @@ describe('KeyringController', () => {
...
+      await withController(async ({ controller }) => {
+        await controller.setLocked();
 
-              await expect(async () =>
-                controller.changePassword(''),
-              ).rejects.toThrow(KeyringControllerError.ControllerLocked);
-            },
-          );
-        });
+        await expect(async () => controller.changePassword('')).rejects.toThrow(
```
FrederikBolding, line 2571:
```suggestion
        await expect(controller.changePassword('')).rejects.toThrow(
```
Shouldn't this be enough?

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2061,50 +2046,36 @@ export class KeyringController extends BaseController<
...
-      } else {
-        if (typeof password !== 'string') {
+
+        if (typeof encryptionKey !== 'string') {
           throw new TypeError(KeyringControllerError.WrongPasswordType);
```
matthiasgeihs, line 2069: This throws `WrongPasswordType` when the encryptionKey is not a string?
(I realize it was the same before, but wondering if we should introduce a separate error for that.)

FrederikBolding (approved):

packages/keyring-controller/src/KeyringController.test.ts
mikesposito: I would expect it to be, yes, but apparently not?

```
 ● KeyringController › changePassword › should throw error if `isUnlocked` is false

    KeyringController - The operation cannot be completed while the controller is locked.
```

FrederikBolding: Ah because the function is not marked as async but just returns promises and this fails before then: https://github.com/MetaMask/core/blob/02372634bca89b82e630ad4cd8383066d4425a25/packages/keyring-controller/src/KeyringController.ts#L1452

packages/keyring-controller/src/KeyringController.ts
mikesposito: Extension (or `browser-passworder`) uses `data` as ciphertext property, while mobile uses `cipher`. The `EncryptionResultConstraint` represent the minimum object to be extended as encryption result type, while `DefaultEncryptionResult` represents the type coming from `browser-passworder` (which is also the default type parameter, hence its name).

Splitting them allows us to make the `Encryptor` type generic enough, `KeyringController` more type safe as it can always access `keyMetadata` and `salt`, and allows Mobile/Extension to define their EncryptionResult type

packages/keyring-controller/jest.config.js
mikesposito: The uncovered lines are the same, though this PR gets rid of a lot of code so the covered/uncovered ratio is different

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
mikesposito: That would be just a dev dependency (and it is in the `package.json` already). I'm unsure if it would make sense to export it from `KeyringController` as `KeyDerivationOptions` is now a type parameter, and browser-passworder's type is just the default

packages/keyring-controller/src/KeyringController.ts
mikesposito: That's a good suggestion, we can take the `encryptionSalt` assignment out of the if/else, though I'm trying to make as little change as possible to the current implementation, as the goal of this PR is just to remove `cacheEncryptionKey`

The cool stuff is in https://github.com/MetaMask/core/pull/7128

packages/keyring-controller/src/KeyringController.ts
mikesposito: This is being done in this follow-up PR: https://github.com/MetaMask/core/blob/5f8200dfbcda34a4c54ca93c371804ce264a1374/packages/keyring-controller/src/KeyringController.ts#L1986-L1986

packages/keyring-controller/src/KeyringController.ts
matthiasgeihs: Ideally, the situation would be more consistent. Having a multitude of interfaces for an otherwise simple return type feels a bit messy. Anyhow, nothing we can easily do about it now, I guess. Good at least to have the interfaces consistent now with the implementations at least.

packages/keyring-controller/src/KeyringController.ts
matthiasgeihs: ok

matthiasgeihs (approved):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
Gudahtt: Even just using it as a default type does result in it being added to the exported types.

Here's what I see locally after running `yarn build` for the types of this file:

```
import { BaseController } from "@metamask/base-controller";
import type * as encryptionUtils from "@metamask/browser-passworder";
import type { SEC1EncodedPublicKey } from "@metamask/toprf-secure-backup";
import { ToprfSecureBackup } from "@metamask/toprf-secure-backup";
import type { AuthConnection } from "./constants.mjs";
import { controllerName, SecretType } from "./constants.mjs";
import { SecretMetadata } from "./SecretMetadata.mjs";
import type { SeedlessOnboardingControllerMessenger, SeedlessOnboardingControllerOptions, SeedlessOnboardingControllerState, SocialBackupsMetadata } from "./types.mjs";
/**
 * Get the initial state for the Seedless Onboarding Controller with defaults.
 *
 * @param overrides - The overrides for the initial state.
 * @returns The initial state for the Seedless Onboarding Controller.
 */
export declare function getInitialSeedlessOnboardingControllerStateWithDefaults(overrides?: Partial<SeedlessOnboardingControllerState>): SeedlessOnboardingControllerState;
export declare class SeedlessOnboardingController<EncryptionKey, SupportedKeyDerivationOptions = encryptionUtils.KeyDerivationOptions> extends BaseController<typeof controllerName, SeedlessOnboardingControllerState, SeedlessOnboardingControllerMessenger> {
    #private;
```

It will definitely produce a type error if someone attempts to use this package with TypeScript without having `@metamask/browser-passworder` installed as well. Therefore it should be listed as a dependency. Unless we can get rid of this dependency somehow (e.g. inline the type, and validate that it matches `@metamask/browser-passworder` instead of using that type directly)

PR #7128, opened 2025-11-12
refactor!: `KeyringController` always derives encryption key

### LINKED ISSUE #6010: [KeyringController] Drop uncached encryption support

mikesposito:
## Explanation

This PR extracts the second and final set of changes from #5963, which refactors the `KeyringController` to always derive the encryption key from the password, rather than storing it in memory.

This change allows to simplify `#unlockKeyrings` and `#updateVault` methods, after the removeal of all the logic and tests related to `cacheEncryptionKey`. This also allows to remove `this.#password`, that has been replaced by `this.#encryptionKey`. 

The `this.#encryptionKey` assignment logic has been moved to two new internal methods with these specific responsibilities:
- `#deriveEncryptionKey(string)`: Derives the encryption key from the password, to be used during password login and password change.
- `#useEncryptionKey(string, string)`: Uses an existing encryption key to be used directly, to be used by `submitEncryptionKey` mainly.

Package previews from this branch are tested on clients with these PRs:
- https://github.com/MetaMask/metamask-extension/pull/37454 
- https://github.com/MetaMask/metamask-mobile/pull/22240

## References

* Fixes #6010

## Checklist

---

> [!NOTE]
> KeyringController now derives and uses an exportable encryption key (with salt) instead of storing the password, updates method flows and error handling, and requires encryptors to implement exportKey/keyFromPassword/generateSalt; dependencies and tests updated accordingly.
> 
> - **Keyring Controller**:
>   - Always derives and uses serialized `encryptionKey` + `salt`; removes in-memory password usage and persists vault via `encryptWithKey`.
>   - Refactors unlock/reencrypt flows: `submitPassword`, `submitEncryptionKey`, `changePassword`, and vault upgrades now use derived/exported keys.
>   - Adds internal helpers `#deriveAndSetEncryptionKey` and `#setEncryptionKey`; updates rollback/persist logic to track encryption credentials.
>   - Introduces `WrongEncryptionKeyType` error; improves validation and error paths (e.g., missing/expired credentials, vault shape).
>   - Updates `Encryptor` interface (BREAKING): requires `exportKey`, `keyFromPassword`, and `generateSalt`.
>   - Test suite overhauled for new flows; coverage thresholds tightened.
> - **Seedless Onboarding Controller**:
>   - Notes BREAKING requirement: provided `encryptor` must implement `exportKey`, `keyFromPassword`, `generateSalt`.
> - **Dependencies**:
>   - Bump `@metamask/browser-passworder` to `^6.0.0` in affected packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71fc1963434baf4e1248279dd3948a19e9e3347f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1874,13 +1906,97 @@ export class KeyringController<
...
+   *
+   * If `options.ignoreExistingVault` is set to `true`, the existing
+   * vault is completely ignored: the new key won't be able to decrypt
+   * the existing vault, and should be used to re-encrypt it.
+   *
+   * @param password - The password to use for decryption or derivation.
+   * @param options - Options for the key derivation.
+   * @param options.ignoreExistingVault - Whether to ignore the existing vault salt and key metadata
+   */
+  async #deriveEncryptionKey(
```
matthiasgeihs: nit
```suggestion
  async #deriveAndSetEncryptionKey(
```
(to emphasize that we also set the key here and `#setEncryptionKey` doesn't need to be called afterwards)

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2089,10 +2193,8 @@ export class KeyringController<
 
       this.update((state) => {
         state.keyrings = updatedKeyrings;
-        if (updatedState.encryptionKey || updatedState.encryptionSalt) {
-          state.encryptionKey = updatedState.encryptionKey;
-          state.encryptionSalt = updatedState.encryptionSalt;
-        }
+        state.encryptionKey = encryptionKey;
+        state.encryptionSalt = parsedEncryptedVault.salt;
```
matthiasgeihs: I think there is a potential bug here:
Above you set
```ts
        this.#setEncryptionKey(
          credentials.encryptionKey,
          credentials.encryptionSalt || parsedEncryptedVault.salt,
        );
```
but here you set
```ts
state.encryptionSalt = parsedEncryptedVault.salt;
```
Shouldn't this be:
```ts
state.encryptionSalt = this.#encryptionKey.salt;
```

mikesposito: As we are decrypting the vault here, the existing vault salt and `this.#encryptionKey.salt` should always be the same, even when the salt is undefined, right?

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2024,60 +2140,48 @@ export class KeyringController<
...
+          credentials.encryptionSalt || parsedEncryptedVault.salt,
+        );
+      }
 
-        // This call is required on the first call because encryptionKey
-        // is not yet inside the memStore
-        updatedState.encryptionKey = encryptionKey;
-        updatedState.encryptionSalt = encryptionSalt;
+      const encryptionKey = this.#encryptionKey?.serialized;
+      if (!encryptionKey) {
```
Gudahtt, line 2174: Nit: This branch appears to be unreachable. Maybe it was just included to convince TypeScript that the encryption key has been set?

If that's the case, we can use an `/* istanbul ignore next */` comment here to prevent it from being counted against us in the coverage stats, and a brief comment to explain why the branch is unreachable

packages/keyring-controller/src/KeyringController.ts
mikesposito: Good point! renamed in 9d1bb4d18a61

packages/keyring-controller/src/KeyringController.ts
Gudahtt: Yeah, we have a check to ensure they are the same in `#setEncryptionKey`. See where we throw the `KeyringControllerError.ExpiredCredentials` error.

Though... it is a bit confusing that the guarantee about these being the same is in a different function 🤔. Perhaps it would be simplest to use `this.#encryptionKey.salt;` just so the reader doesn't need to understand where it was validated?

Potentially a slight readability improvement here, but I don't think it's a bug

mikesposito: That makes sense. I was being conservative on what was there before, but I agree that using `this.#encryptionKey.salt` is clearer, and should be equivalent. Done in c59a9c1db603

packages/keyring-controller/src/KeyringController.ts
mikesposito: Yes indeed, just to make TypeScript happy. Done in b0da6a8d3464

lwin-kyaw (approved):

PR #7137, opened 2025-11-12
perf!: remove `getNextAvailableAccountName`

hmalik88:
## Explanation

`handleOnKeyringStateChange` is calling this function on every account which is taking ~ 200ms. It's been determined that we can remove this method. 

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking API removal and changes to how new accounts get `metadata.name` during `KeyringController:stateChange`, which could affect consumers relying on AccountsController-generated default names.
> 
> **Overview**
> Removes the `AccountsController.getNextAvailableAccountName` API (and messenger action/export) and stops computing default account names inside `#handleOnKeyringStateChange`, avoiding per-account name scans during keyring state updates.
> 
> As a result, newly added accounts via keyring state changes no longer receive an auto-generated `metadata.name` from `AccountsController` (tests are updated to expect empty names), and the now-dead name-generation/unit tests are deleted; `utils` tests add coverage for `isSimpleKeyringType`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75010f4183b67eb5650e7daea18c98f4d6424ff2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/accounts-controller/CHANGELOG.md
+++ b/packages/accounts-controller/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Remove method `getNextAvailableAccountName` and its usage ([#7137](https://github.com/MetaMask/core/pull/7137))
+- **BREAKING:** Remove method `getNextAvailableAccountName` and its usage ([#7137](https://github.com/MetaMask/core/pull/7137))
   - The function was removed since it was taking a lot of time in `handleOnKeyringStateChange`, we compute account names in the `AccountTreeController` regardless.
 
 ## [35.0.2]


packages/accounts-controller/src/AccountsController.ts
```diff
@@ -868,12 +868,6 @@ export class AccountsController extends BaseController<
                 internalAccounts.accounts,
               ) as InternalAccount[];
 
-              // Get next account name available for this given keyring.
-              const name = this.getNextAvailableAccountName(
```
gantunesr, line 886: should we also remove the action `AccountsControllerGetNextAvailableAccountNameAction`? And I see the method also being bind

hmalik88: I wasn't sure if I should remove that and if it was being used in the clients. Since @danroc mentioned that we should just remove the calls.

gantunesr: Thats a good point, but the call would fail making this a breaking change, right?

ccharly: So, I think we can categorize this PR as `perf:` improvement for now.

I think we can keep the action and make this just a minor/patch change (not sure which yet, but we can decide later 😄)

However, I'm not 100% sure of the side-effects of this change @hmalik88, so could you make a preview build + make test/integration PRs on both clients and see how that goes?

We just got some errors around naming with another fix today, and some e2e were failing because of that (because we did not migrate everything yet to the full BIP-44 integration unfortunately!)

hmalik88: @gantunesr I don't think the call would fail since we didn't remove the function itself, just the call to it in the state change handler. 

@ccharly yup, I can make a preview build and test.

hmalik88:
@metamaskbot publish-preview

packages/accounts-controller/src/AccountsController.ts
danroc: We should probably remove the action as well, since it shouldn’t be used anymore.

In my opinion, we should:
1. Remove the method and the action
2. Publish a preview
3. Create a PR on the clients with the updates and fixes
4. Merge and release the `accounts-controller`
5. Update and merge the client PRs

gantunesr: I also prefer to make these changes in one go

packages/accounts-controller/src/utils.test.ts
```diff
@@ -19,6 +23,21 @@ describe('utils', () => {
...
+
+    it('returns false for non-simple keyring type', () => {
+      expect(isSimpleKeyringType(KeyringTypes.hd)).toBe(false);
+      expect(isSimpleKeyringType(KeyringTypes.snap)).toBe(false);
+      expect(isSimpleKeyringType(KeyringTypes.trezor)).toBe(false);
+      expect(isSimpleKeyringType(KeyringTypes.oneKey)).toBe(false);
+      expect(isSimpleKeyringType(KeyringTypes.ledger)).toBe(false);
+      expect(isSimpleKeyringType(KeyringTypes.lattice)).toBe(false);
+    });
+  });
```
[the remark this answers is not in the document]
hmalik88: Might be used outside of this repo, so I will leave as is.

packages/accounts-controller/src/utils.test.ts
```diff
@@ -19,6 +23,21 @@ describe('utils', () => {
     });
   });
 
+  describe('isSimpleKeyringType', () => {
```
gantunesr, line 26: was this added to increase the coverage?

packages/accounts-controller/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Remove usage of method `getNextAvailableAccountName` and its usage ([#7137](https://github.com/MetaMask/core/pull/7137))
```
gantunesr: Should this be under `Removed`?

packages/accounts-controller/src/utils.test.ts
gantunesr: I see that the execution of `isSimpleKeyringType` was removed in the other file so this is no longer executed as part of the tests

hmalik88: yup added to increase coverage, CI fails otherwise

packages/accounts-controller/CHANGELOG.md
hmalik88: True, updated.

packages/accounts-controller/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/keyring-controller` from `^25.0.0` to `^25.1.0` ([#7713](https://github.com/MetaMask/core/pull/7713))
 
+### Removed
+
+- Remove method `getNextAvailableAccountName` and its usage ([#7137](https://github.com/MetaMask/core/pull/7137))
```
gantunesr:
```suggestion
- **BREAKING**: Remove method `getNextAvailableAccountName` and its usage ([#7137](https://github.com/MetaMask/core/pull/7137))
```

gantunesr (approved):

PR #7143, opened 2025-11-13
simulate with gas

chaitanyapotti:
## Explanation
Issue with Gas Field Missing in Re-Simulation (Blocker for Shield Release)

While testing the Shield approval transaction, we observed the following behavior:
- In the first simulation, the client sends the gas value along with other transaction parameters to Sentinel.
- In subsequent re-simulations, the client does not send the gas value.

This causes an issue for transactions that are eligible for gas sponsorship, since the gas value is a key factor in determining sponsorship eligibility to prevent malicious users from abusing and draining funds of STX.

This is currently a blocker for the Shield release.
Please help to address this issue as soon as possible.

## Test
We have tested the change and it worked.

## Checklist

---

> [!NOTE]
> Pass `txParams.gas` through to `simulateTransactions` in `getGasFeeTokens` to ensure accurate gas fee token/sponsorship calculations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0556dff72460609b3a35041a7dedc4fc762ba4bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/transaction-controller/src/utils/gas-fee-tokens.ts
+++ b/packages/transaction-controller/src/utils/gas-fee-tokens.ts
@@ -60,6 +60,7 @@ export async function getGasFeeTokens({
   const from = txParams.from as Hex;
   const to = txParams.to as Hex;
   const value = txParams.value as Hex;
+  const gas = txParams.gas as Hex;
 
   log('Request', { chainId, txParams });
 
@@ -95,6 +96,7 @@ export async function getGasFeeTokens({
           from,
           to,
           value,
+          gas,
         },
       ],
       suggestFees: {


tanguyenvn (approved):

packages/transaction-controller/src/utils/gas-fee-tokens.ts
```diff
@@ -95,6 +96,7 @@ export async function getGasFeeTokens({
           from,
           to,
           value,
+          gas,
```
matthewwalsh0, line 99: If this is the only route to retrieve the gas fee tokens, how is the `gas` available in the initial request?

PR #7164, opened 2025-11-14
Extend createServicePolicy to support live network status

mcmire:
## Explanation

In a future commit we will introduce changes to `network-controller` so that it will keep track of the status of each network as requests are made. These updates to `createServicePolicy` assist with that. See the changelog for more.

Besides this, the tests for `createServicePolicy` have been refactored slightly so that they are easier to maintain in the future.

## References

You can see how these changes will be used in the next PR: https://github.com/MetaMask/core/pull/7166

## Checklist

---

> [!NOTE]
> Adds `getCircuitState`, `onAvailable`, and `reset` to `ServicePolicy`, exports Cockatiel types, and updates logic/tests to support availability tracking and circuit state introspection.
> 
> - **controller-utils**:
>   - **ServicePolicy API**:
>     - Add `getCircuitState()` to expose underlying circuit state.
>     - Add `onAvailable` event for first success and post-recovery success.
>     - Add `reset()` to close the circuit and reset breaker counters.
>   - **Behavior/Internals**:
>     - Track availability status and emit `onAvailable`/`onDegraded` appropriately.
>     - Update `onBreak` to mark unavailable; wire `ConsecutiveBreaker` for reset.
>   - **Exports**:
>     - Export `CockatielEventEmitter` and `CockatielFailureReason`; re-export via `index`.
>   - **Tests**:
>     - Expand/refactor tests to cover `onAvailable`, `getCircuitState`, `reset`, and timing cases; update export snapshot.
>   - **Docs**:
>     - Update `CHANGELOG.md` with new methods and exports.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e597d0be1a4adc066b10cb7dc8a86c61408fd01d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/check-release/action.yml
+++ b/.github/actions/check-release/action.yml
@@ -1,21 +1,73 @@
 name: Check release
 description: Check for conflicts in packages being released in this PR.
 
-inputs:
-  pull-request:
-    description: 'The pull request number to get changed files from.'
-    required: true
-
 runs:
   using: composite
   steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
       with:
         fetch-depth: 0
 
+    - name: Get merge base
+      id: merge-base
+      shell: bash
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.ref || github.event.merge_group.base_ref }}
+      run: |
+        set -euo pipefail
+
+        # Strip invalid prefix from `github.event.merge_group.base_ref`
+        # It comes prefixed with `refs/heads/`, but the branch is not checked out in this context
+        # We need to express it as a remote branch
+        PREFIXED_REF_REGEX='refs/heads/(.+)'
+        if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+          BASE_REF="${BASH_REMATCH[1]}"
+        fi
+
+        MERGE_BASE=$(git merge-base HEAD "refs/remotes/origin/$BASE_REF")
+        echo "MERGE_BASE=$MERGE_BASE" >> "$GITHUB_OUTPUT"
+
+    - name: Check if the commit or pull request is a release
+      id: is-release
+      uses: MetaMask/action-is-release@v2
+      with:
+        commit-starts-with: 'Release [version],Release v[version],Release/[version],Release/v[version],Release `[version]`'
+        commit-message: ${{ github.event.pull_request.title }}
+        before: ${{ steps.merge-base.outputs.MERGE_BASE }}
+
+    # `action-is-release` checks out the repository with `fetch-depth: 2`, so we
+    # need to check it out again with full history.
+    - name: Checkout repository
+      uses: actions/checkout@v5
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
+      with:
+        fetch-depth: 0
+
+    - name: Get pull request number
+      id: pr-number
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
+      shell: bash
+      env:
+        EVENT_NAME: ${{ github.event_name }}
+        PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
+        MERGE_GROUP_HEAD_REF: ${{ github.event.merge_group.head_ref }}
+      run: |
+        if [ "$EVENT_NAME" = "pull_request" ]; then
+          echo "PR_NUMBER=$PULL_REQUEST_NUMBER" >> "$GITHUB_OUTPUT"
+        elif [ "$EVENT_NAME" = "merge_group" ]; then
+          PR_NUMBER_REGEX='/pr-([0-9]+)-'
+          if [[ "$MERGE_GROUP_HEAD_REF" =~ $PR_NUMBER_REGEX ]]; then
+            echo "PR_NUMBER=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+          else
+            echo "::error::Could not extract PR number from merge group head ref: $MERGE_GROUP_HEAD_REF."
+            exit 1
+          fi
+        fi
+
     - name: Get target reference
       id: get-target
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
       shell: bash
       env:
         EVENT_NAME: ${{ github.event_name }}
@@ -30,14 +82,15 @@ runs:
         fi
 
     - name: Check commits for changes in released packages
+      id: check-release
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
       shell: bash
       env:
         GH_TOKEN: ${{ github.token }}
-        PULL_REQUEST: ${{ inputs.pull-request }}
+        PULL_REQUEST: ${{ steps.pr-number.outputs.PR_NUMBER }}
         TARGET: ${{ steps.get-target.outputs.TARGET }}
       run: |
         set -euo pipefail
-        git fetch origin main
 
         mapfile -t PACKAGES < <(find packages -maxdepth 2 -name "package.json" -not -path "*/node_modules/*")
         RELEASED_PACKAGES=()
@@ -61,7 +114,7 @@ runs:
         git fetch origin "$PULL_REQUEST_BRANCH"
 
         # Get all files changed ahead of this PR.
-        BEFORE=$(git merge-base "refs/remotes/origin/main" "origin/$PULL_REQUEST_BRANCH")
+        BEFORE=$(git merge-base "refs/remotes/origin/main" "refs/remotes/origin/$PULL_REQUEST_BRANCH")
         git diff --name-only "$BEFORE..$TARGET" > changed-files.txt
 
         CONFLICTS=()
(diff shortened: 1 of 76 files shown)


packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -2731,76 +3179,382 @@ describe('createServicePolicy', () => {
...
+            optionsWithCircuitBreakDuration: {},
+          },
+          {
+            desc: 'a custom circuit break duration',
+            circuitBreakDuration: DEFAULT_CIRCUIT_BREAK_DURATION,
+            options: {
+              // This has to be high enough to exceed the exponential backoff
+              optionsWithCircuitBreakDuration: 5_000,
+            },
+          },
```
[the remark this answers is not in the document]
mcmire: This should be fixed now.

packages/controller-utils/src/create-service-policy.ts
```diff
@@ -276,27 +314,56 @@ export function createServicePolicy(
...
-      duration > degradedThreshold
-    ) {
-      onDegradedEventEmitter.emit();
+    if (circuitBreakerPolicy.state === CircuitState.Closed) {
+      if (duration > degradedThreshold) {
+        onDegradedEventEmitter.emit();
+      } else if (availabilityStatus !== AVAILABILITY_STATUSES.Available) {
+        availabilityStatus = AVAILABILITY_STATUSES.Available;
+        onAvailableEventEmitter.emit();
+      }
```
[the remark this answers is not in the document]
Gudahtt: Yeah, this seems wrong

Maybe this should be an `if` condition rather than `else if`

packages/controller-utils/src/create-service-policy.ts
```diff
@@ -276,27 +314,56 @@ export function createServicePolicy(
   circuitBreakerPolicy.onStateChange((state) => {
     internalCircuitState = getInternalCircuitState(state);
   });
+
+  circuitBreakerPolicy.onBreak(() => {
+    availabilityStatus = AVAILABILITY_STATUSES.Unavailable;
```
Gudahtt, line 320: Rather than updating the availability status separately in an `onBreak` handler and in the retry policy success handler, we could use `circuitBreakerPolicy.onStateChange` to consolidate the state updates in one place.

That would make this a bit easier to follow. Thoughts?

packages/controller-utils/src/create-service-policy.ts
```diff
@@ -306,14 +373,35 @@ export function createServicePolicy(
     return null;
   };
 
+  const getCircuitState = () => {
+    return circuitBreakerPolicy.state;
+  };
+
+  const reset = () => {
+    // Set the state of the policy to "isolated" regardless of its current state
```
Gudahtt, line 384: What purpose does this step serve? It looks like `consecutiveBreaker.success();` alone would be enough to reset the circuit breaker, but maybe I'm misunderstanding what that does

packages/controller-utils/src/create-service-policy.ts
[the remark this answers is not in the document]
mcmire: My intent here was that `onAvailable` would only get called if the circuit is still closed, and if the service responds successfully within the degraded threshold. And if these conditions were true for more than one request in a row, the callback would only get called once.

So emitting `onAvailable` if the service is slow doesn't seem right, because the status there should be "degraded".

Perhaps `availabilityStatus` is a bad name. The status this refers to only contains "unknown", "unavailable" and "available", but notably, it does not contain "degraded". Really, the only status that's valuable here is "available", the other two are there to represent reality better. But maybe we _should_ include degraded too, just for completeness? Or, maybe collapsing the status variable to a simple boolean (`isAvailable`) would make more sense?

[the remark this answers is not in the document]
Gudahtt: >So emitting onAvailable if the service is slow doesn't seem right

The point here is that we're _not_ emitting `onAvailable` when the service transitions from "unavailable" to "available", if the first successful request is slow enough to still be considered degraded.

I don't see anything here about emitting `onAvailable` when service is slow 🤔

packages/controller-utils/src/create-service-policy.ts
mcmire: The goal of this method is to prevent the following scenario from occurring:

1. A primary's circuit breaks, and traffic cuts over to its failover
3. The failover's circuit breaks
4. Later, the primary recovers, and traffic resumes on the primary
5. Still later, the primary's circuit breaks, and traffic again cuts over to the failover (which by now has recovered)
6. Requests to the failover instantly fail because the circuit is still open.

In this situation, I thought it best to reset the failover at step 3 so that at step 5 _(EDIT: corrected numbers here)_, a successful request can be made to the failover instead of having to wait for the circuit break duration to pass (if it has not already passed).

This change is introduced in the next PR: https://github.com/MetaMask/core/pull/7166/files#diff-47102459f0fdb56c6f8071aa3b2906bf32c060480ae02c2a4c8da7e187a034a1R453

`consecutiveBreaker.success()` would reset the breaker, but it would not reset the state of the circuit breaker policy, which not only keeps track of whether the circuit is open or closed but also when it was last opened so that it knows whether to throw BrokenCircuitError or transition to a half-open state.

packages/controller-utils/src/create-service-policy.ts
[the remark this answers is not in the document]
mcmire: Right. What I'm saying is that I don't think `onAvailable` should be emitted when the service is slow. The service may be "available" in the sense that it's not down, but it's not _fully_ available, which is really what I'm defining `available` to mean.

For more context, this `onAvailable` callback was intended to complement `onDegraded`, and I eventually plan on using it to know when to set the "available" status on NetworkController. I also plan on introducing a "degraded" status in between "unavailable", "blocked", and "available". The end goal is to tie the network status into the banner that we've added to the clients, so: 1) when a network transitions to "degraded", we can use it to show the "degraded" banner in the client, 2) when the network transitions to "unavailable", we can use it to show the "degraded" banner in the client, 3) when the network transitions to "available" or "unknown", then we know to hide the banner.

Does that plan make sense? Maybe you think of the network status in a different way?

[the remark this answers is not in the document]
Gudahtt: Ah OK, I hadn't realized that "degraded" was intended to be represented as "unavailable".

If that's the goal, then we're not correctly setting the status to "unavailable" when a degraded request occurs.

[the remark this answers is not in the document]
Gudahtt: Though I do think calling a degraded network "unavailable" is misleading. It is available, just slow.

If a "degraded" status was introduced, it would be clearer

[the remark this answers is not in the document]
mcmire: That's fair. I have made this change, does this make more sense? [b3909af](https://github.com/MetaMask/core/pull/7164/commits/b3909afaf3b1b03a6f83ddc5f9c9ae1afeb833aa)

packages/controller-utils/src/create-service-policy.ts
mcmire: Does this still make sense if we add a "degraded" status?

packages/controller-utils/src/create-service-policy.ts
mcmire: It is a bit weird that we have to call `isolate` and then call `dispose` on the result to reset the circuit breaker completely, but this was the only way I could find to do this. You can see what it does here: https://github.com/connor4312/cockatiel/blob/f87137b9bb55e0cdfb02e7d5c401e47c6f5fcc9f/src/CircuitBreakerPolicy.ts#L184

Gudahtt: Interesting, OK. Well, if it works 🤷

packages/controller-utils/src/create-service-policy.ts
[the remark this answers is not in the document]
Gudahtt: Yes, that looks better now, thanks!

packages/controller-utils/src/create-service-policy.ts
Gudahtt: Yes, this comment is no longer relevant with that status

Gudahtt (commented):
Looks great! I'll continue review tomorrow, the only thing I have left is the tests.

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -215,6 +229,29 @@ describe('createServicePolicy', () => {
...
+        const policy = createServicePolicy({
+          retryFilterPolicy: handleWhen(
+            (caughtError) => caughtError.message !== 'failure',
+          ),
+        });
+        policy.onAvailable(onAvailableListener);
+
+        const promise = policy.execute(mockService);
+        // It's safe not to await this promise; adding it to the promise queue
+        // is enough to prevent this test from running indefinitely.
```
Gudahtt, line 248: Nit: Not sure about this justification. Floating promises are problematic for reasons _other_ than keeping the test process alive / leaking memory. They also can cause test failures in the un-awaited code to appear in later tests.

I see this comment is pre-existing in many tests though, so best not to address this in this PR anyway.

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -215,6 +229,29 @@ describe('createServicePolicy', () => {
 
         expect(onDegradedListener).not.toHaveBeenCalled();
       });
+
+      it('does not call onAvailable listeners', async () => {
```
Gudahtt, line 233: Hmm. This behavior is a bit questionable actually. If a request 'fails" but was filtered out by the policy, it means that there is no outage, i.e. that the service is working as intended, it "failed successfully", and is likely available.

Gudahtt: Similarly, it does seem fair to call service degraded if it fails _really slowly_.

Gudahtt: I guess we can leave this change for a separate PR, if we want to make it, since it would be more of a departure from how it works today.

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -317,6 +355,26 @@ describe('createServicePolicy', () => {
             await ignoreRejection(promise);
 
             expect(onDegradedListener).toHaveBeenCalledTimes(1);
+            expect(onDegradedListener).toHaveBeenCalledWith({ error });
+          });
+
+          it('does not call onAvailable listeners', async () => {
```
Gudahtt, line 361: Nit: These tests seem unnecessarily exhaustive, and this test case is a good example of this.

It's great that we're testing that the default number of retries works as expected, and that a custom number of retries works as expected. But why do we need to test this both with default max consecutive failures *and* with custom max consecutive failures? And do we really need to test all four combinations of these conditions with a "always failing" policy and a "failing repeatedly until it finally succeeds" policy - do we really expect one to fail if the other succeeds?

This unit test suite seems to be taking the extreme approach of covering every possible combination of conditions. This is certainly the only way to be absolutely certain the code works as expected, but we rarely write tests this exhaustive for good reason: it's really expensive/time-consuming.

It is enough to test expected behaviour once. We don't need to test that behaviour is identical across scenarios that don't impact that behaviour.

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -2731,76 +3179,379 @@ describe('createServicePolicy', () => {
...
+          });
+        });
+      });
+    });
+  });
+
+  describe('wrapping a service that succeeds at first and then fails enough to break the circuit', () => {
+    describe.each([
+      {
+        desc: `the default max number of consecutive failures (${DEFAULT_MAX_CONSECUTIVE_FAILURES})`,
```
Gudahtt, line 3299: Nit: as mentioned in the other comment, I don't think we need to test the effectiveness of this option in every single test scenario

packages/controller-utils/src/create-service-policy.test.ts
mcmire: Sure, we can consider this for another PR.

packages/controller-utils/src/create-service-policy.test.ts
mcmire: Thank you for pointing this out. I agree that the exhaustive strategy I took with this file is hurting more than helping at this point. I think I just need to take a fresh approach and try to keep it simple. I'll think more about how I can do that.

packages/controller-utils/src/create-service-policy.test.ts
mcmire: Makes sense. This is an old pattern. In newer tests, I tend to use this pattern: 

``` typescript
policy.onRetry(() => {
  clock.next();
});
```

When I rewrite this file I'll go through and make sure to use this pattern instead.

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -2731,76 +3179,379 @@ describe('createServicePolicy', () => {
...
+        describe.each([
+          {
+            desc: `the default circuit break duration (${DEFAULT_CIRCUIT_BREAK_DURATION})`,
+            circuitBreakDuration: DEFAULT_CIRCUIT_BREAK_DURATION,
+            optionsWithCircuitBreakDuration: {},
+          },
+          {
+            desc: 'a custom circuit break duration',
+            circuitBreakDuration: DEFAULT_CIRCUIT_BREAK_DURATION,
+            optionsWithCircuitBreakDuration: {
```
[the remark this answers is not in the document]
mcmire: Ugh, I thought I fixed this already. This is true, but the test still works, so I'm inclined to go with it for now.

PR #7166, opened 2025-11-15
Revamp NetworkController RPC endpoint events

mcmire:
## Explanation

In a future commit we will introduce changes to `network-controller` so that it will keep track of the status of each network as requests are made. This commit paves the way for this to happen by redefining the existing RPC endpoint-related events that NetworkController produces.

Currently, when requests are made through the network clients that NetworkController exposes, three events are published:

- `NetworkController:rpcEndpointDegraded`
  - Published when enough successive retriable errors are encountered while making a request to an RPC endpoint that the maximum number of retries is reached.
- `NetworkController:rpcEndpointUnavailable`
  - Published when enough successive errors are encountered while making a request to an RPC endpoint that the underlying circuit breaks.
- `NetworkController:rpcEndpointRequestRetried`
  - Published when a request is retried (mainly used for testing).

It's important to note that in the context of the RPC failover feature, an "RPC endpoint" can actually encompass multiple URLs, so the above events actually fire for any URL.

While these events are useful for reporting metrics on RPC endpoints, in order to effectively be able to update the status of a network, we need events that are less granular and are guaranteed not to fire multiple times in a row. We also need a new event.

Now the list of events looks like this:

- `NetworkController:rpcEndpointDegraded`
  - The same as before.
- `NetworkController:rpcEndpointUnavailable`
  - The same as before.
- `NetworkController:rpcEndpointRetried`
  - Renamed from `NetworkController:rpcEndpointRequestRetried`.
- `NetworkController:rpcEndpointChainDegraded`
  - Similar to `NetworkController:rpcEndpointDegraded`, but won't be published again if the RPC endpoint is already in a degraded state.
- `NetworkController:rpcEndpointChainUnavailable`
  - Published when all of the circuits underlying all of the URLs for an RPC endpoint have broken (none of the URLs are available). Won't be published again if the RPC endpoint is already in an unavailable state.
- `NetworkController:rpcEndpointChainAvailable`
  - A new event. Published the first time a successful request is made to one of the URLs for an RPC endpoint, or following a degraded or unavailable status.

Going a bit deeper, in order to make the changes above, it was necessary to rewrite the core logic responsible for diverting traffic to failovers from `RpcService` to `RpcServiceChain`, which was a more natural fit, anyway. This also meant that we could simplify `RpcService`, as well as its tests.

## References

## Checklist

  - This will come in a later PR.

---

> [!NOTE]
> Introduce chain-level RPC endpoint events and a new RpcServiceChain, rename/update retry and payloads, and refactor failover logic with extensive test updates.
> 
> - **Events & Payloads (BREAKING)**:
>   - Add `NetworkController:rpcEndpointChainAvailable` and chain-level events `rpcEndpointChainDegraded`/`rpcEndpointChainUnavailable` with non-repeating semantics in `src/NetworkController.ts`.
>   - Rename `NetworkController:rpcEndpointRequestRetried` to `NetworkController:rpcEndpointRetried`.
>   - Update payloads: add `networkClientId`, add `primaryEndpointUrl` to per-endpoint events; chain-level events omit `primaryEndpointUrl`.
> - **Failover Architecture**:
>   - Introduce `RpcServiceChain` (`src/rpc-service/rpc-service-chain.ts`) to manage primary/failover endpoints, circuit states, and emit chain/service events.
>   - Simplify `RpcService` (`src/rpc-service/rpc-service.ts`): remove embedded failover, add `onAvailable`, `resetPolicy`, `getCircuitState`, improved error handling/logging.
>   - Wire into client creation: `create-network-client.ts` builds `RpcServiceChain` and publishes updated events; `create-auto-managed-network-client.ts` now passes `networkClientId` to `createNetworkClient`.
> - **Exports & Types**:
>   - Update exports in `src/index.ts`; adjust `RpcServiceRequestable` event listener types and remove `{ isolated: true }` from `onBreak` data.
> - **Tests**:
>   - Add comprehensive tests for new chain/service events and behaviors; update helpers/mocks to support dynamic responses.
> - **Changelog & Deps**:
>   - Document breaking changes and new events in `CHANGELOG.md`.
>   - Add `cockatiel` to devDependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 916a0e2ba9427e9b987bd6aaf86803424094eb14. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
+++ b/packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
@@ -907,7 +907,7 @@ describe('RpcServiceChain', () => {
     it('calls onServiceBreak each time the circuit of an RPC service in the chain breaks', async () => {
       const primaryEndpointUrl = 'https://first.endpoint';
       const secondaryEndpointUrl = 'https://second.endpoint';
-      const tertiaryEndpointUrl = 'https://second.endpoint';
+      const tertiaryEndpointUrl = 'https://third.endpoint';
       nock(primaryEndpointUrl)
         .post('/', {
           id: 1,


packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -137,31 +375,103 @@ export class RpcServiceChain implements RpcServiceRequestable {
     jsonRpcRequest: JsonRpcRequest<Params>,
     fetchOptions: FetchOptions = {},
   ): Promise<JsonRpcResponse<Result | null>> {
-    return this.#services[0].request(jsonRpcRequest, fetchOptions);
-  }
+    // Start with the primary (first) service and switch to failovers as the
```
mcmire, line 392: Prior to these changes, each `RpcService` object could have an optional `failoverService` property. This class, `RpcServiceChain`, would then build a chain (really, a linked list) of services. In order to make a request, `RpcServiceChain` would call `request` on the first service in the chain, and `RpcService` would decide whether it needed to call the next service in the chain, etc.

While this model is easy to understand, I needed access to certain data along the way, and it seemed easier to use a loop rather than a linked list. Anyway, I figured it really should be the responsibility of `RpcServiceChain` to manage how requests are sent across the chain.

packages/network-controller/src/rpc-service/rpc-service-requestable.ts
```diff
@@ -22,8 +28,8 @@ export type RpcServiceRequestable = {
    * @see {@link createServicePolicy}
    */
   onRetry(
-    listener: AddToCockatielEventData<
-      Parameters<ServicePolicy['onRetry']>[0],
+    listener: CockatielEventToEventListenerWithData<
```
mcmire, line 31: This should be doing the same thing as the previous code. I just thought that `AddToCockatielEventData<Parameters<...>[0]` was a bit ugly (and I added some more self-descriptive utility types, so this is one of them).

packages/network-controller/src/create-auto-managed-network-client.ts
```diff
@@ -81,13 +86,15 @@ const UNINITIALIZED_TARGET = { __UNINITIALIZED__: true };
 export function createAutoManagedNetworkClient<
   Configuration extends NetworkClientConfiguration,
 >({
+  networkClientId,
```
mcmire, line 89: We now expose the network client ID in the `rpcEndpoint*` messenger events, so we need to receive it and pass it down to `createNetworkClient`.

packages/network-controller/tests/network-client/rpc-failover.ts
```diff
@@ -120,188 +119,6 @@ export function testsForRpcFailoverBehavior({
       });
     });
 
-    it('publishes the NetworkController:rpcEndpointUnavailable event when the failover occurs', async () => {
```
mcmire, line 123: These tests got moved to `src/create-network-client-tests/rpc-endpoint-events.test.ts` as I realized they didn't belong here. It's true that this test file is concerned with the RPC failover feature (which is required to test the `rpcEndpoint*` events), but all of the tests in `tests/network-client` directory really exercise the middleware stack that `createNetworkClient` builds and in so doing loop over all of the RPC methods that our internal provider handles specially. We don't need to go to all that trouble to test the `rpcEndpoint*` events, we can just use an arbitrary RPC method. (I think eventually I will rename `tests/network-client` to `tests/internal-provider-api` or something like that, but that's a PR for another time.)

packages/network-controller/src/rpc-service/shared.ts
```diff
@@ -1,13 +1,69 @@
...
-    ? (data: Data extends void ? AdditionalData : Data & AdditionalData) => void
+export type CockatielEventToEventEmitter<Event> =
+  Event extends CockatielEvent<infer EventData>
+    ? CockatielEventEmitter<EventData>
     : never;
+
+/**
+ * Obtains the event data type from a Cockatiel event or event listener type.
+ */
+export type ExtractCockatielEventData<CockatielEventOrEventListener> =
```
mcmire, line 22: The Cockatiel types are a bit awkward to work with (especially since event listeners whose event payloads are empty are typed as `Event<void>` which is rather strange). These utilities just make them a bit easier to work with.

packages/network-controller/CHANGELOG.md
```diff
@@ -19,6 +24,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - In practice, this should happen rarely if ever.
 - **BREAKING:** Migrate `NetworkClient` to `JsonRpcEngineV2` ([#7065](https://github.com/MetaMask/core/pull/7065))
   - This ought to be unobservable, but we mark it as breaking out of an abundance of caution.
+- **BREAKING:** Split up and update payload data for `NetworkController:rpcEndpoint{Degraded,Unavailable}` ([#7166](https://github.com/MetaMask/core/pull/7166))
+  - The existing events are now called `NetworkController:rpcEndpointInstance{Degraded,Unavailable}` and retain their present behavior.
+  - `NetworkController:rpcEndpointInstance{Degraded,Unavailable}` do still exist, but they are now designed to represent the entire RPC endpoint and are guaranteed to not be published multiple times in a row. In particular, `NetworkController:rpcEndpointUnavailable` is published only after trying all of the designated URLs for a particular RPC endpoint and the underlying circuit for the last URL breaks, not as each primary's or failover's circuit breaks.
```
mcmire: > In particular, `NetworkController:rpcEndpointUnavailable` is published only after trying all of the designated URLs for a particular RPC endpoint and the underlying circuit for the last URL breaks, not as each primary's or failover's circuit breaks.

This change is I suppose a bit controversial, and the first version of this was slightly different before I landed on this approach. But I think it makes sense? Basically, if we can reach the network _somehow_, don't broadcast that it's unavailable until we're absolutely sure. That does mean that the `NetworkController:rpcEndpointUnavailable` may never be published for Infura networks if the failover does its job, but I think that's precisely the intent.

Gudahtt: This change entry seems to be using the term "RPC endpoint" to mean "network client". e.g. here:

>all of the designated URLs for a particular RPC endpoint 

I would think that "RPC endpoint" refers to a specific URL, so it doesn't really make sense for a single endpoint to have multiple URLs.

mcmire:
Adding `no-changelog` because there is a minor change to `createServicePolicy` which is non-user-facing.

packages/network-controller/CHANGELOG.md
mcmire: Yeah, I agree. I wanted to make sure that we use consistent terminology, but that's a bit of a mess now that we have the failover feature (the `RpcEndpoint` type has both `url` and `failoverUrls` properties, so at least from a data/type perspective "RPC endpoint" encompasses multiple URLs). Using "network client" here and perhaps elsewhere makes sense as a workaround.

Should I also change the event names, too, though? I admit I'm not in love with `rpcEndpointInstance*`.

mcmire: Maybe we start calling the combination of a primary and its failovers an "RPC endpoint group"? Or maybe _this_ is a "network" and that way "network client" still makes sense? I also might be thinking about this too hard 😵‍💫

mcmire: Okay, I've revisited this a bit and made some changes: https://github.com/MetaMask/core/compare/78608975022a114a75127d8a693d9d282cc8fabc...b16597acd3d63eba40fe9c03d7551e93dab82c5e?expand=1

I decided to (tentatively) go with these terms:

- "RPC endpoint" = a URL
- "RPC endpoint chain" = a collection of URLs

This means that the existing messenger events stay and we get three new ones:

- `:rpcEndpointChainUnavailable`
- `:rpcEndpointChainDegraded`
- `:rpcEndpointChainAvailable`

This also means that the `RpcEndpoint` type really should be called `RpcEndpointChain`, but we can always fix that later.

Does this help? Is this good/bad?

mcmire: Coming back to this after a night's sleep. I still think "RPC endpoint chain" makes sense. The only downside is the use of "chain" alone could be confusing as it could refer to "network chain". I've attempted to clarify this in test descriptions by using "chain of endpoints" in most cases. It's a bit awkward, but not terrible. That's the best I've got at the moment.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -528,19 +522,41 @@ export class RpcService implements AbstractRpcService {
...
+          'REQUEST SUCCESSFUL:',
+          this.endpointUrl.toString(),
+          response.status,
+        );
         return await response.json();
       });
     } catch (error) {
+      log('REQUEST ERROR:', this.endpointUrl.toString(), error);
+
+      this.#lastError = error;
```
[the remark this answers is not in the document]
mcmire: Good point. Addressed here: [cb498e3](https://github.com/MetaMask/core/pull/7166/commits/cb498e37ce7bc671cad857d65fe261b4442b6dce)

packages/network-controller/src/create-network-client-tests/rpc-endpoint-events.test.ts
```diff
@@ -0,0 +1,1228 @@
...
+                    // Hit the endpoint and exceed the max number of retries,
+                    // breaking the circuit; hit the failover
+                    await makeRpcCall(request);
+
+                    expect(networkAvailableEventHandler).toHaveBeenCalledWith({
+                      chainId,
+                      endpointUrl: rpcUrl,
+                      networkClientId: 'AAAA-AAAA-AAAA-AAAA',
+                      primaryEndpointUrl: rpcUrl,
+                    });
```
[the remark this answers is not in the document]
mcmire: True. I have fixed this test. I also noticed that the tests weren't realistic (if an endpoint responds with errors, the block number request will be problematic, not just subsequent requests) so I fixed that too: [cdd8491](https://github.com/MetaMask/core/pull/7166/commits/cdd849165657b69fc420686ce0b25cdaa471241a)

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -99,210 +490,42 @@ describe('RpcService', () => {
...
-
-        const promise = service.request({
-          id: 1,
-          jsonrpc: '2.0',
-          method: 'eth_chainId',
-          params: [],
-        });
-        await ignoreRejection(promise);
-        expect(onBreakListener).not.toHaveBeenCalled();
+      testsForNonRetriableErrors({
```
mcmire, line 423: I took the opportunity to do some refactoring, as there were some repeated checks, and it was getting a bit difficult to maintain. As a result, the diff after this point is not the best. However I have done my best to point out differences.

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1041,6 +987,103 @@ describe('RpcService', () => {
       });
 
       expect(onDegradedListener).toHaveBeenCalledTimes(1);
+      expect(onDegradedListener).toHaveBeenCalledWith({
+        endpointUrl: `${endpointUrl}/`,
+      });
+    });
+
+    it('calls the onAvailable callback the first time a successful request occurs', async () => {
```
mcmire, line 925: These next two tests are new.

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1062,526 +1105,394 @@ async function ignoreRejection<T>(
...
-    });
+      }),
+    );
+    expect(onDegradedListener).not.toHaveBeenCalled();
+  });
 
-    it('forwards the request to the failover service in addition to the primary endpoint while the circuit is broken, stopping when the primary endpoint recovers', async () => {
-      const clock = getClock();
-      const jsonRpcRequest = {
+  it('does not call onAvailable', async () => {
```
mcmire, line 1140: This test is new.

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1062,526 +1105,394 @@ async function ignoreRejection<T>(
...
+    await ignoreRejection(service.request(jsonRpcRequest));
+
+    expect(logger.warn).toHaveBeenCalledWith(
+      expect.objectContaining({
+        message: 'Execution prevented because the circuit breaker is open',
+      }),
+    );
+  });
+
+  it('calls the onAvailable callback if the endpoint becomes degraded via errors and then recovers', async () => {
```
mcmire, line 1384: This test is new.

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1062,526 +1105,394 @@ async function ignoreRejection<T>(
...
-          message: 'Execution prevented because the circuit breaker is open',
-        }),
-      );
-    });
+      }),
+    );
+    expect(onBreakListener).not.toHaveBeenCalled();
   });
 
-  describe('if a failover service is provided', () => {
```
mcmire, line 1261: All of the failover-related tests have been removed from this file, as RpcServiceChain now handles diverting traffic to failovers.

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1613,363 +1524,203 @@ function testsForRetriableResponses({
...
+    await ignoreRejection(service.request(jsonRpcRequest));
+
+    expect(onBreakListener).toHaveBeenCalledTimes(1);
+    expect(onBreakListener).toHaveBeenCalledWith({
+      error: expectedOnBreakError,
+      endpointUrl: `${endpointUrl}/`,
     });
   });
 
-  describe('if a failover service is provided', () => {
```
mcmire, line 1736: Same as above, this `describe` block has been removed.

packages/network-controller/CHANGELOG.md
Gudahtt: Ahhh, I hadn't realized this was referring to the primary and failure URLs as a group, that makes more sense.

Agreed about the new term, "chain" is much clearer. Great solution

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,154 @@ export type NetworkControllerNetworkRemovedEvent = {
 };
 
 /**
- * `rpcEndpointUnavailable` is published after an attempt to make a request to
- * an RPC endpoint fails too many times in a row (because of a connection error
- * or an unusable response).
+ * `NetworkController:rpcEndpointChainUnavailable` is published when the number
+ * of failed consecutive attempts to receive a 2xx response from the primary
```
Gudahtt: Nit: I think it's looking for non-500's rather than 200s
```suggestion
 * of failed consecutive attempts to receive a non-5xx response from the primary
```

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,154 @@ export type NetworkControllerNetworkRemovedEvent = {
...
+ * endpoint of a chain of endpoints reaches a maximum, causing further requests
+ * to be temporarily paused, and when subsequent traffic to a failover endpoint
+ * similarly fails.
+ *
+ * In other words, this event will not published if a primary is deemed to be
+ * unavailable but its failover is not.
```
Gudahtt: Nit: I read this wrong the first time, here is an attempt at making it harder to misinterpret
```suggestion
 * In other words, this event will not published if the failover is available,
 * even if the primary is not.
```

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,154 @@ export type NetworkControllerNetworkRemovedEvent = {
...
+ *
+ * Additionally, if this was the last `NetworkController:rpcEndpointChain*`
```
Gudahtt: Nit: Is this worth calling out here? It would never occur to me that this would get published repeatedly, since it represents the chain becoming unavailable. If it's already unavailable, how would it recur?

Gudahtt: Likewise for the same paragraph in the other type descriptions

packages/network-controller/src/NetworkController.ts
Gudahtt: Nit: the other type descriptions similarly mention 2xx responses, rather than non-5xx

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,154 @@ export type NetworkControllerNetworkRemovedEvent = {
...
+ * event to be published, the event will not be re-published (for instance, if
+ * both a primary and failover are deemed to be unavailable, or if more than one
+ * failover is deemed to be unavailable).
+ *
+ * @param payload - The event payload.
+ * @param payload.chainId - The chain ID of the target network.
+ * @param payload.endpointUrl - The URL of the endpoint among the chain of
```
Gudahtt: In the context of a _chain_ becoming unavailable, this is a confusing description. Wouldn't this nearly always be the failover chain? But technically it could be the primary if both circuits break, then the primary recovers then breaks again?

This seems more confusing for subscribers than helpful. I'm not really sure what the purpose of exposing the primary is either.

packages/network-controller/src/NetworkController.ts
mcmire: If we were listening to `onBreak` for each endpoint (RpcService) in the chain as we do for `NetworkController:rpcEndpointUnavailable`, then this would be published repeatedly as each endpoint's circuit broke. But that is not how this event works.

I guess this is an implementation detail, though. I can try to simplify this.

packages/network-controller/src/NetworkController.ts
mcmire: You are right. For this event, `endpointUrl` will always be the last failover URL, even if the primary recovers and then breaks again. So this description is not specific enough.

As for the payload data, you're also right that realistically, neither `endpointUrl` and `primaryEndpointUrl` are used. You might have noticed that each of the events has the same payload data. My thinking here was if we needed to hook this event up to metrics or the network connection banner in the future, we wouldn't need to make any changes to `network-controller`. That said, `endpointUrl` is not very useful for this event due to the behavior described above. So I think it would be safe to remove, if that helps clear the confusion?

mcmire: I have removed `endpointUrl` from this event. I have also tried to go through each event's JSDoc and clarify any confusing wording: [73017d1](https://github.com/MetaMask/core/pull/7166/commits/73017d115d0f55732c5b4239f92506279fd5e0b4)

packages/network-controller/src/NetworkController.ts
mcmire: I have removed this paragraph for this and other events here: [73017d1](https://github.com/MetaMask/core/pull/7166/commits/73017d115d0f55732c5b4239f92506279fd5e0b4)

packages/network-controller/src/NetworkController.ts
mcmire: Good point, I have applied this suggestion here: [73017d1](https://github.com/MetaMask/core/pull/7166/commits/73017d115d0f55732c5b4239f92506279fd5e0b4)

packages/network-controller/src/NetworkController.ts
mcmire: Yes, this is very confusing. I tried to go through each event's JSDoc and clarify any confusing wording here: [73017d1](https://github.com/MetaMask/core/pull/7166/commits/73017d115d0f55732c5b4239f92506279fd5e0b4)

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +165,148 @@ export function createNetworkClient({
...
+
+  rpcServiceChain.onServiceRetry(
+    ({ primaryEndpointUrl: _, endpointUrl, attempt }) => {
+      messenger.publish('NetworkController:rpcEndpointRetried', {
+        chainId: configuration.chainId,
+        networkClientId: id,
+        primaryEndpointUrl,
+        endpointUrl,
+        attempt,
+      });
```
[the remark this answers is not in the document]
mcmire: True. Fixed here: [026ce00](https://github.com/MetaMask/core/pull/7166/commits/026ce00f36c9a694f6af081578b140e850457f11)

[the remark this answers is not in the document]
mcmire: True. Fixed here: [026ce00](https://github.com/MetaMask/core/pull/7166/commits/026ce00f36c9a694f6af081578b140e850457f11)

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,119 @@ export type NetworkControllerNetworkRemovedEvent = {
...
+  type: 'NetworkController:rpcEndpointChainUnavailable';
+  payload: [
+    {
+      chainId: Hex;
+      error: unknown;
+      networkClientId: NetworkClientId;
+      primaryEndpointUrl: string;
+    },
+  ];
+};
```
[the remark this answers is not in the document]
mcmire: This is intentional, as per the discussion above.

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -139,31 +377,103 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+        }
+      }
+
+      return response;
+    }
+
+    // The only way we can end up here is if there are no services to loop over.
+    // That is not possible due to the types on the constructor, but TypeScript
+    // doesn't know this, so we have to appease it.
+    throw new Error('Nothing to return');
```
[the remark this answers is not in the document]
mcmire: It's a bit confusing; the `onAvailable` event actually occurs at the service policy level and is then propagated to RpcService and RpcServiceChain. So although there is some logic to track the status within `RpcServiceChain.request`, it only covers setting the status to unavailable.

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -27,20 +93,75 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+      service.onDegraded((data) => {
+        if (this.#status !== STATUSES.Degraded) {
+          log('Updating status to "degraded"', data);
+          this.#status = STATUSES.Degraded;
+          this.#onDegradedEventEmitter.emit({
+            ...data,
+            primaryEndpointUrl: this.#primaryService.endpointUrl.toString(),
+          });
+        }
+      });
```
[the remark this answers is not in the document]
mcmire: I don't think this is accurate. The chain should not transition from unavailable to degraded. If the circuit breaks, then the circuit break duration passes, then the endpoint fails again, onBreak will be emitted, not onDegraded. It is possible to transition from degraded to unavailable to available to degraded, but that is by design.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -315,7 +306,34 @@ export class RpcService implements AbstractRpcService {
...
+  }
+
+  /**
+   * @returns The last failure reason that the retry policy captured (or
+   * `undefined` if the last execution of the service was successful).
+   */
+  getLastInnerFailureReason(): { error: unknown } | undefined {
+    return this.#lastError === undefined
+      ? undefined
+      : { error: this.#lastError };
```
Gudahtt: Nit: why `{ error: this.#lastError }` rather than `this.#lastError`? It wasn't clear what the wrapper object was for.

packages/network-controller/src/rpc-service/rpc-service-chain.ts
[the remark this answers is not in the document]
mcmire: See above.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -325,39 +343,25 @@ export class RpcService implements AbstractRpcService {
...
+  onBreak(listener: Parameters<AbstractRpcService['onBreak']>[0]) {
     return this.#policy.onBreak((data) => {
-      listener({
-        ...data,
-        endpointUrl: this.endpointUrl.toString(),
-        failoverEndpointUrl: this.#failoverService
-          ? this.#failoverService.endpointUrl.toString()
-          : undefined,
-      });
+      if (!('isolated' in data)) {
```
Gudahtt, line 367: This is for the `Update `AbstractRpcService`/`RpcServiceRequestable` to remove `{ isolated: true }` from the `onBreak` event data type` change I guess?

Nit: If we're ignoring this because we _never_ call isolate, perhaps we should throw an error in case we do at some point in the future, so these events aren't totally suppressed? Or log something to the console at at least?

I was going to suggest a comment to explain why we're doing this, but an error or log would probably serve that purpose and highlight this mistake if it does come up.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -528,19 +522,43 @@ export class RpcService implements AbstractRpcService {
...
    */
-  async #processRequest<Result extends Json>(
+  async #executeAndProcessRequest<Result extends Json>(
     fetchOptions: FetchOptions,
   ): Promise<JsonRpcResponse<Result> | JsonRpcResponse<null>> {
     let response: Response | undefined;
     try {
-      return await this.#policy.execute(async () => {
+      log(
+        `[RpcService: ${this.endpointUrl}] Circuit state`,
```
Gudahtt: Nit: Maybe we can drop `RpcService` from the message, since the logger already has a namespace of `RpcService`. It'll be prefixed with that already

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -528,19 +522,43 @@ export class RpcService implements AbstractRpcService {
...
+      );
+      const jsonDecodedResponse = await this.#policy.execute(async (data) => {
+        log(
+          'REQUEST INITIATED:',
+          this.endpointUrl.toString(),
+          '::',
+          fetchOptions,
+          // @ts-expect-error This property _is_ here, the type of ServicePolicy
+          // is just wrong.
+          `(attempt ${data.attempt + 1})`,
```
Gudahtt: Interesting.

Nit: Is `data` the right term here, perhaps we should call this `context` like the type does?

Also, we should fix the `ServicePolicy` type at some point. It's definitely our mistake, the retry policy type in cockatiel has `attempt` in it. Maybe worth adding an issue for it, and/or a TODO comment?

packages/network-controller/src/rpc-service/rpc-service.test.ts
```diff
@@ -1062,526 +1105,394 @@ async function ignoreRejection<T>(
...
+      params: [],
+    };
+    // Cause the retry policy to give up
+    await ignoreRejection(service.request(jsonRpcRequest));
+    await service.request(jsonRpcRequest);
+
+    expect(onAvailableListener).toHaveBeenCalledTimes(1);
   });
+
+  /* eslint-enable jest/require-top-level-describe */
```
Gudahtt, line 1425: Nit: this is fine, but I wonder going forward if we should disable this rule? Or if we should move test helpers like this to separate files, and add them to the "test helpers" ESLint config section where this rule is disabled.

Either way we can deal with it later, in a separate PR

packages/network-controller/src/rpc-service/shared.ts
```diff
@@ -1,13 +1,69 @@
...
+ */
+export type CockatielEventToEventListenerWithData<Event, Data> = (
+  data: ExtendCockatielEventData<ExtractCockatielEventData<Event>, Data>,
+) => void;
+
+/**
+ * Converts a Cockatiel event type to an event listener type, but removing the
+ * requested keys from the data.
+ */
+export type CockatielEventToEventListenerWithoutData<
```
Gudahtt: Nit: This appears to be unused

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: When we emit `onBreak` in RpcServiceChain we have to pass not an error object but a Cockatiel `FailureReason` object: 

https://github.com/MetaMask/core/blob/0e13332c4f1ec77dba67126faecf30e4c8f1d4e9/packages/network-controller/src/rpc-service/rpc-service-chain.ts#L438

So here we prepare that here so we can pass it directly.

But you're right, maybe it's more clear for this to return an error instead. Actually in that case we don't even need a method, we can just have a `lastError` property. I've made that change here: [74d9d0f](https://github.com/MetaMask/core/pull/7166/commits/74d9d0f172dde4f817f82a54a878390e0527ed96)

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Hmm, okay, I guess the changelog entry isn't quite accurate. We _do_ call `.isolate` on the circuit breaker policy, but we're sort of abusing it. It's not so that we can hold the circuit open; we call it in `ServicePolicy.reset` so that we can reset the state of the circuit breaker.

The problem is that when we call `.isolate`, it will immediately call `onBreak` with `{ isolated: true }`: https://github.com/connor4312/cockatiel/blob/f87137b9bb55e0cdfb02e7d5c401e47c6f5fcc9f/src/CircuitBreakerPolicy.ts#L187. We don't want that to cause the `NetworkController:rpcEndpointUnavailable` or `NetworkController:rpcEndpointChainUnavailable` events to be published. So that's why we are ignoring the error here.

If that makes sense, I can add a comment and update the changelog entry.

packages/network-controller/src/rpc-service/rpc-service.test.ts
mcmire: Yeah, this isn't the first time I've had to do this, and it feels noisy to me. I do think the rule is valuable, so I'm hesitant to disable it. Maybe moving to a separate file is best. I'll think about it.

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Yes, good call. Removed here: [bbbfa9c](https://github.com/MetaMask/core/pull/7166/commits/bbbfa9cd62871061b415badb9e48c757ac00d504)

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Using `context` makes more sense. Updated here: [4f9d84b](https://github.com/MetaMask/core/pull/7166/commits/4f9d84b8a24d18adc9e73d337a9b9ee56b15ebce)

And yes, I agree we should fix `ServicePolicy`. I believe that would be a breaking change, so I was hesitant to do that given all of the other breaking changes being made. But maybe once I am back in town, I can address this.

packages/network-controller/src/rpc-service/shared.ts
mcmire: Good eye. Removed here: [1862024](https://github.com/MetaMask/core/pull/7166/commits/18620242152defd1e6cd532ad8f41e6998e29809)

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -139,31 +377,95 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+              primaryEndpointUrl: this.#primaryService.endpointUrl.toString(),
+              endpointUrl: service.endpointUrl.toString(),
+            });
+          }
+        }
+
+        // The service failed, and we throw whatever the error is. The calling
+        // code can try again if it so desires.
+        log(
+          `${isCircuitOpen ? "This service's circuit is open, but for some reason it wasn't handled above. " : "This service's circuit is closed. "}Re-throwing error.`,
```
Gudahtt: What do you mean here by "wasn't handled above"? It looks like the only case handled above is when there is a next service to attempt. If there is no next service, we would expect to reach this line even if the circuit is open.

mcmire: Hmm... I don't really understand this either. It might have been left over from a previous version. I don't think we need to say anything extra if the circuit is open, the previous log took care of that already. Fixed here: [2f29875](https://github.com/MetaMask/core/pull/7166/commits/2f29875e922e763607b1aac980a69c65c50be223)

packages/network-controller/src/rpc-service/rpc-service.ts
Gudahtt: Ah I see, makes sense

packages/network-controller/src/NetworkController.ts
Gudahtt: >My thinking here was if we needed to hook this event up to metrics or the network connection banner in the future, we wouldn't need to make any changes to network-controller.

We could derive that from the network client ID anyway though right? It still doesn't seem clear why this would be useful in the event. Maybe I'm missing something though.

Even if it's just included for convenience reasons, chain ID seems like the more valuable property (though that's also derivable from the network client ID).

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -139,31 +378,94 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+        for (const [i, service] of [...this.#services.entries()].slice(
+          availableServiceIndex + 1,
+        )) {
+          log(`Resetting policy for service #${i + 1}.`);
+          service.resetPolicy();
+        }
+      }
+
+      return response;
+    }
```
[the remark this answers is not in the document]
mcmire: That's okay, the primary's circuit does not need to be reset, it is already in the state we want it to be in.

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -7,18 +11,81 @@ import type {
...
+  CockatielEventToEventEmitterWithData,
+  CockatielEventToEventListenerWithData,
+  ExtendCockatielEventData,
+  ExtractCockatielEventData,
+  FetchOptions,
+} from './shared';
+import { projectLogger, createModuleLogger } from '../logger';
+
+// const log = createModuleLogger(projectLogger, 'RpcServiceChain');
+const log = console.log.bind(console);
```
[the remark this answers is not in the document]
mcmire: Fixed here: [72de1d6](https://github.com/MetaMask/core/pull/7166/commits/72de1d66f2ee7a392695866cc9729b97562248c5)

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Comment added here: [0f55384](https://github.com/MetaMask/core/pull/7166/commits/0f55384192283a290e98d99653e8a32d165669ee)

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -7,18 +11,80 @@ import type {
...
- * This class constructs a chain of RpcService objects which represent a
- * particular network. The first object in the chain is intended to be the
- * primary way of reaching the network and the remaining objects are used as
- * failovers.
+ * Statuses that the RPC service chain can be in.
  */
-export class RpcServiceChain implements RpcServiceRequestable {
+type Status = (typeof STATUSES)[keyof typeof STATUSES];
+
+type RpcServiceConfiguration = Omit<RpcServiceOptions, 'failoverService'>;
```
Gudahtt: This doesn't seem relevant anymore - there is no `failoverService` property to omit here

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +651,1446 @@ describe('RpcServiceChain', () => {
...
+
+      const jsonRpcRequest = {
+        id: 1,
+        jsonrpc: '2.0' as const,
+        method: 'eth_chainId',
+        params: [],
+      };
+      // Retry the first endpoint until its circuit breaks, then retry the
+      // second endpoint until *its* circuit breaks.
+      for (let i = 0; i < 5; i++) {
```
Gudahtt: Nit: This 5 is a little confusing, it was unclear why this was enough to break both circuits. Especially given that there was no assertion to confirm that it happened.

Maybe better to express as a constant that shows how it's calculated? e.g.

```
/**
 * The number of fetch requests made for a single request to an RPC service, using default max
 * retry attempts.
 */
const DEFAULT_REQUEST_ATTEMPTS = 1 + DEFAULT_MAX_RETRIES;
/**
 * Number of attempts required to break the circuit of an RPC service using default retry attempts
 * and max consecutive failures.
 *
 * Note: This calculation and later ones assume that there is no remainder.
 */
const DEFAULT_RPC_SERVICE_ATTEMPTS_UNTIL_BREAK =
  DEFAULT_MAX_CONSECUTIVE_FAILURES / DEFAULT_REQUEST_ATTEMPTS;

/**
 * Number of attempts required to break the circuit of an RPC service chain (with a single
 * failover) that uses default retry attempts and max consecutive failures.
 *
 * The value is one less than double the number of attempts needed to break a single circuit
 * because on failure of the primary, the request gets forwarded to the failover immediately.
 */
const DEFAULT_RPC_CHAIN_ATTEMPTS_UNTIL_BREAK =
  2 * DEFAULT_RPC_SERVICE_ATTEMPTS_UNTIL_BREAK - 1;
```

Then we can use this `DEFAULT_RPC_CHAIN_ATTEMPTS_UNTIL_BREAK` constant here instead of "5" (it does end up being 5)

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +651,1446 @@ describe('RpcServiceChain', () => {
...
+      // The circuit will break on the last time, and the third endpoint will
+      // be hit, until max retries is hit.
+      await expect(rpcServiceChain.request(jsonRpcRequest)).rejects.toThrow(
+        expectedError,
+      );
+      // Retry the third endpoint again, until max retries is hit.
+      await expect(rpcServiceChain.request(jsonRpcRequest)).rejects.toThrow(
+        expectedError,
+      );
+      // Retry the second endpoint for a third time, until max retries is hit.
```
Gudahtt: This is the third, not the second:
```suggestion
      // Retry the third endpoint for a third time, until max retries is hit.
```

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +651,1446 @@ describe('RpcServiceChain', () => {
...
+
+      expect(onDegradedListener).toHaveBeenCalledTimes(1);
+      expect(onDegradedListener).toHaveBeenCalledWith({
+        primaryEndpointUrl: `${endpointUrl}/`,
+        endpointUrl: `${endpointUrl}/`,
+        error: expectedDegradedError,
+      });
+    });
+
+    it("does not call onDegraded again when the primary service's circuit breaks and its failover responds successfully but slowly", async () => {
```
Gudahtt, line 1189: I find this test confusing. When this says "does not call onDegraded again", what does the _again_ mean?

It looks like the test triggers the `onDegraded` event once, but then doesn't do anything after that point. There is no behavior here that could potentially trigger it, this test isn't really guarding against anything. It also seems redundant alongside the previous test.

Gudahtt: Wait no, not redundant with the previous. This tests the failover responding slowly, the previous tests the primary responding slowly. Hmm....

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +651,1446 @@ describe('RpcServiceChain', () => {
...
+
+      expect(onDegradedListener).toHaveBeenCalledTimes(1);
+      expect(onDegradedListener).toHaveBeenCalledWith({
+        primaryEndpointUrl: `${primaryEndpointUrl}/`,
+        endpointUrl: `${primaryEndpointUrl}/`,
+        error: expectedDegradedError,
+      });
+    });
+
+    it("calls onDegraded again when a service's underlying circuit breaks, and then after waiting, the service responds successfully but slowly", async () => {
```
Gudahtt, line 1263: This description confused me. It seemed like it was saying that `onDegraded` _could_ get called twice if we pause long enough, but the "underlying circuit breaks" part meant the _failiover_ service (e.g. it was talking about if we went from degraded => unavailable => degraded).

This seems clearer:
```suggestion
    it("calls onDegraded again when the service chain becomes unavailable, and then after waiting, the failover service responds successfully but slowly", async () => {
```

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
Gudahtt: Nevermind, this is probably fine, I just misunderstood it.

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
Gudahtt: In general I find the term "a service's underlying circuit" confusing in the context of a service chain, because I don't know which it refers to (the primary or the failover). Better to avoid the term in these descriptions I think (or at least be clear about which service you mean)

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +165,172 @@ export function createNetworkClient({
...
+  const availableEndpointUrls: [string, ...string[]] = isRpcFailoverEnabled
+    ? [primaryEndpointUrl, ...(configuration.failoverRpcUrls ?? [])]
+    : [primaryEndpointUrl];
+  const buildRpcServiceConfiguration = (endpointUrl: string) => ({
+    ...getRpcServiceOptions(endpointUrl),
+    endpointUrl,
+    logger,
+  });
+
+  const getError = (value: object) => {
```
Gudahtt: What is this for 🤔 Where is this `value` parameter coming from, and why is the error sometimes under the `error` or `value` properties?

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +165,172 @@ export function createNetworkClient({
...
+  logger?: Logger;
+}) {
+  const availableEndpointUrls: [string, ...string[]] = isRpcFailoverEnabled
+    ? [primaryEndpointUrl, ...(configuration.failoverRpcUrls ?? [])]
+    : [primaryEndpointUrl];
+  const buildRpcServiceConfiguration = (endpointUrl: string) => ({
+    ...getRpcServiceOptions(endpointUrl),
+    endpointUrl,
+    logger,
+  });
```
Gudahtt: Nit: I don't think we need a builder function here, it's only called in one place. We can do something like

```suggestion
  const rpcServiceConfigurations = availableEndpointUrls.map((endpointUrl) => ({
    ...getRpcServiceOptions(endpointUrl),
    endpointUrl,
    logger,
  })),
```

Then reference `rpcServiceConfigurations[0]` and `...rpcServiceConfigurations.slice(1)` later in the constructor.

At least this seems simpler to me? At least slightly.

packages/network-controller/src/create-network-client.ts
Gudahtt: I see this was pre-existing. Hmm. The question stands but it's not a blocker since it was already present.

packages/network-controller/src/create-network-client-tests/rpc-endpoint-events.test.ts
```diff
@@ -0,0 +1,1290 @@
...
+                  networkClientId: 'AAAA-AAAA-AAAA-AAAA',
+                  primaryEndpointUrl: rpcUrl,
+                });
+              },
+            );
+          },
+        );
+      });
+
+      it('does not publish the NetworkController:rpcEndpointChainDegraded event again if the max number of retries is reached in making requests to a failover endpoint', async () => {
```
Gudahtt: Nit: Maybe it would be better to group these test cases (one group where failover is enabled, one where it isn't)?

At first it was hard to tell how this differed from previous tests, but it's because this has failover enabled. The previous tests did not.

packages/network-controller/src/create-network-client-tests/rpc-endpoint-events.test.ts
```diff
@@ -0,0 +1,1290 @@
...
+          method: 'eth_gasPrice',
+          params: [],
+        };
+        const expectedError = createResourceUnavailableError(503);
+        const expectedDegradedError = new HttpError(503);
+
+        await withMockedCommunications(
+          { providerType: networkClientType },
+          async (primaryComms) => {
+            await withMockedCommunications(
```
Gudahtt: Hmm. Seeing 3 nested `with___` blocks to setup the environment is not ideal.

Not something to address in this PR, it's already large enough. But it would be great if we could pass in a list of mocked communication objects to `withNetworkClient` to make this more concise (and reduce indentation)

packages/network-controller/CHANGELOG.md
```diff
@@ -7,8 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - The event payloads for all events have been changed as well: `failoverEndpointUrl` has been renamed to `endpointUrl`, and `endpointUrl` has been renamed to `primaryEndpointUrl`. In addition, `networkClientId` has been added to the payload.
+- **BREAKING:** Rename and update payload data for `NetworkController:rpcEndpointRequestRetried` ([#7166](https://github.com/MetaMask/core/pull/7166))
+  - This event is now called `NetworkController:rpcEndpointRetried`.
+  - The event payload has been changed as well: `failoverEndpointUrl` has been renamed to `endpointUrl`, and `endpointUrl` has been renamed to `primaryEndpointUrl`. In addition, `networkClientId` and `attempt` have been added to the payload.
+- **BREAKING:** Update `AbstractRpcService`/`RpcServiceRequestable` to remove `{ isolated: true }` from the `onBreak` event data type ([#7166](https://github.com/MetaMask/core/pull/7166))
+  - This represented the error produced when `isolate` is called on a Cockatiel circuit breaker policy, which we don't do (at least, not in the way it is designed to be called)
```
Gudahtt: Nit: This is a little confusing/mysterious. Here's an attempt at rewording it:

```suggestion
  - This represented the error produced when `isolate` is called on a Cockatiel circuit breaker policy. This never happens for our service (we use `isolate` internally, but this error is suppressed and cannot trigger `onBreak`)
```

packages/network-controller/CHANGELOG.md
```diff
@@ -7,8 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - New events are `NetworkController:rpcEndpointChainDegraded` and `NetworkController:rpcEndpointChainUnavailable`, and are designed to represent an entire chain of endpoints. They are also guaranteed to not be published multiple times in a row. In particular, `NetworkController:rpcEndpointUnavailable` is published only after trying all of the endpoints for a chain and when the underlying circuit for the last endpoint breaks, not as each primary's or failover's circuit breaks.
+  - The event payloads for all events have been changed as well: `failoverEndpointUrl` has been renamed to `endpointUrl`, and `endpointUrl` has been renamed to `primaryEndpointUrl`. In addition, `networkClientId` has been added to the payload.
+- **BREAKING:** Rename and update payload data for `NetworkController:rpcEndpointRequestRetried` ([#7166](https://github.com/MetaMask/core/pull/7166))
+  - This event is now called `NetworkController:rpcEndpointRetried`.
+  - The event payload has been changed as well: `failoverEndpointUrl` has been renamed to `endpointUrl`, and `endpointUrl` has been renamed to `primaryEndpointUrl`. In addition, `networkClientId` and `attempt` have been added to the payload.
```
Gudahtt: This seems to imply that `endpointUrl` is always the failover endpoint, but I don't think that's true.

Here's an attempt at rewording it:

```suggestion
  - The event payload has been changed as well: `failoverEndpointUrl` has been removed, and `primaryEndpointUrl` has been added. In addition, `networkClientId` and `attempt` have been added to the payload.
```

packages/network-controller/src/rpc-service/rpc-service-chain.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/src/create-network-client.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/CHANGELOG.md
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/CHANGELOG.md
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/067a807cae44b64c5ba8e645eb0c4bb974dc9d45

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
Gudahtt: .... I am very sorry, I must have been looking at a different test when I made this comment. Or maybe I misread it because the RPC mocks were setup with multiple call chains. But this test appears to have no failover, so my suggestion here was totally wrong. The old description is actually fine.

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/4b57f325f191b4a5a9ff78d708deac6da6fe1d08

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
cryptodev-2s: No worries, my bad haven't paid attention to that neither, fixed here https://github.com/MetaMask/core/pull/7166/commits/056703f217a6fa6c534de3a3ff4016b674adb9c7

packages/network-controller/src/create-network-client.ts
cryptodev-2s: The `getError` function extracts the error from Cockatiel's `FailureReason` type, which can have two shapes:

1. **`{ error: Error }`** - When the service throws an error (the common case for RPC failures)
2. **`{ value: T }`** - When the service returns a value that the retry filter policy considers a failure

The `value` parameter comes from the event data passed to the circuit breaker event listeners (`onBreak`, `onServiceBreak`, `onDegraded`, `onServiceDegraded`). These listeners receive data from Cockatiel's circuit breaker policy that includes the `FailureReason` object.

Gudahtt: Ahhh interesitng, OK, that makes sense.

Nit: Given that explanation, it seems that we really ought to have better types here. It seems like `value: FailureReason<unknown> | Record<never, never>` would work

Gudahtt: Nit: Also a brief TSDoc description would be helpful here

packages/network-controller/src/create-network-client-tests/rpc-endpoint-events.test.ts
cryptodev-2s: done [here](https://github.com/MetaMask/core/pull/7166/commits/10775285df7e26b02cc7b090b00f614b2e6bae9e?w=1)

cryptodev-2s: done [here](https://github.com/MetaMask/core/pull/7166/commits/321ff7c183ae056bd3e7ed984f1863a08442395f?w=1)

packages/network-controller/src/create-network-client.ts
cryptodev-2s: done [here](https://github.com/MetaMask/core/pull/7166/commits/7b9c7367f35bd720974c8007bc996b7cdc9962eb)

packages/network-controller/src/create-network-client-tests/rpc-endpoint-events.test.ts
cryptodev-2s: I will address this in a follow-up PR. A first [draft](https://github.com/MetaMask/core/pull/7239) is already prepared, and I will continue once this is merged.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7166/commits/2c356489f1e6f9ed425c44368a34180e772c8ef8

packages/network-controller/CHANGELOG.md
```diff
@@ -7,8 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 
+- **BREAKING:** Split up and update payload data for `NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointUnavailable` ([#7166](https://github.com/MetaMask/core/pull/7166))
+  - `NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointUnavailable` still exist and retain the same behavior as before.
+  - New events are `NetworkController:rpcEndpointChainDegraded` and `NetworkController:rpcEndpointChainUnavailable`, and are designed to represent an entire chain of endpoints. They are also guaranteed to not be published multiple times in a row. In particular, `NetworkController:rpcEndpointChainUnavailable` is published only after trying all of the endpoints for a chain and when the underlying circuit for the last endpoint breaks, not as each primary's or failover's circuit breaks.
+  - The event payloads have been changed:
+    - For individual endpoint events (`NetworkController:rpcEndpointUnavailable`, `NetworkController:rpcEndpointDegraded`): `failoverEndpointUrl` has been renamed to `endpointUrl`, and `endpointUrl` has been renamed to `primaryEndpointUrl`. In addition, `networkClientId` has been added to the payload.
```
Gudahtt: Nit: Same issue as the other line below that was modified already

```suggestion
    - For individual endpoint events (`NetworkController:rpcEndpointUnavailable`, `NetworkController:rpcEndpointDegraded`): `failoverEndpointUrl` has been removed, and `primaryEndpointUrl` has been added. In addition, `networkClientId` has been added to the payload.
```

cryptodev-2s: Thanks fixed here https://github.com/MetaMask/core/pull/7166/commits/30a94860c693ea18fac359bd2232a4be6346e828

packages/network-controller/src/create-network-client.ts
Gudahtt: Did you want to consider this as well?

```suggestion
  const getError = (value: FailureReason<unknown> | Record<never, never>) => {
```

It seemed to work locally when I tried it

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +165,184 @@ export function createNetworkClient({
...
+      const error = getError(rest);
+      messenger.publish('NetworkController:rpcEndpointUnavailable', {
+        chainId: configuration.chainId,
+        networkClientId: id,
+        primaryEndpointUrl: primaryEndpointUrlFromEvent,
+        endpointUrl,
+        error,
+      });
+    },
+  );
```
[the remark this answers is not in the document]
Gudahtt: Cursor seems correct here, we should take the same approach as with `onBreak` and throw an error here if the error is missing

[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7166/commits/4e7a37accc30a24a07d4e3ef538232559e42d448

packages/network-controller/src/create-network-client.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/61bdbbcc0864c14ddcc0e18d49fd019afe4e9442

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -139,31 +372,94 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+            log(
+              "This service's circuit is open. Proceeding to next service...",
+            );
+            continue;
+          }
+
+          if (
+            previousCircuitState !== CircuitState.Open &&
+            this.#status !== STATUSES.Unavailable &&
+            lastError !== undefined
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7166/commits/9fea617bcbeb36f7bfd211f444db37a6ecf1edb7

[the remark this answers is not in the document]
Gudahtt: I'm not sure I understand. it looks like the change you introduced to prevent this problem actually _introduces_ this problem 🤔 

The status change (and onBreak event) should be gated by the _current_ status, not the status from before the request was made. The status may have changed in the interim.

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +679,1446 @@ describe('RpcServiceChain', () => {
...
+          method: 'eth_chainId',
+          params: [],
+        })
+        .reply(200, () => {
+          return {
+            id: 1,
+            jsonrpc: '2.0',
+            result: '0x1',
+          };
+        });
```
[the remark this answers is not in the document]
Gudahtt: True, this test does look incorrect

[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7166/commits/c59898d7e58bfd9569aa1a6a11673561da4084ba

packages/network-controller/src/rpc-service/rpc-service-chain.ts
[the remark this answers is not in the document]
cryptodev-2s: Yes indeed reverted here https://github.com/MetaMask/core/pull/7166/commits/e4459ee849b67ac5d9145a2081508afee9018bfb

packages/network-controller/src/NetworkController.ts
```diff
@@ -443,25 +443,115 @@ export type NetworkControllerNetworkRemovedEvent = {
...
+export type NetworkControllerRpcEndpointChainUnavailableEvent = {
+  type: 'NetworkController:rpcEndpointChainUnavailable';
+  payload: [
+    {
+      chainId: Hex;
+      error: unknown;
+      networkClientId: NetworkClientId;
+    },
+  ];
+};
```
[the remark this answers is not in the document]
Gudahtt: Hmm. It is unclear to me why `endpointUrl` was only removed from the `Unavailable` event

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -139,31 +372,94 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+      // open (due to previous failures), we would receive a "circuit broken"
+      // error when we attempted to divert traffic to the failovers again.
+      //
+      if (availableServiceIndex !== undefined) {
+        for (const [i, service] of [...this.#services.entries()].slice(
+          availableServiceIndex + 1,
+        )) {
+          log(`Resetting policy for service #${i + 1}.`);
+          service.resetPolicy();
+        }
```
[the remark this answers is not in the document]
Gudahtt: This behavior is intended, it's not backwards.

Services _after_ the successful one may have been tried on a previous request, so their state could be anything at this point.

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +168,193 @@ export function createNetworkClient({
...
+
+  const rpcServiceChain = new RpcServiceChain([
+    rpcServiceConfigurations[0],
+    ...rpcServiceConfigurations.slice(1),
+  ]);
+
+  rpcServiceChain.onBreak(
+    ({
+      endpointUrl,
+      primaryEndpointUrl: primaryEndpointUrlFromEvent,
```
Gudahtt: Nit: Unused parameter

packages/network-controller/src/create-network-client.ts
```diff
@@ -194,6 +168,193 @@ export function createNetworkClient({
...
+        endpointUrl,
+        error,
+      });
+    },
+  );
+
+  rpcServiceChain.onDegraded(
+    ({
+      endpointUrl,
+      primaryEndpointUrl: primaryEndpointUrlFromEvent,
```
Gudahtt: Nit: unused parameter

packages/network-controller/src/rpc-service/rpc-service-chain.ts
```diff
@@ -51,15 +167,51 @@ export class RpcServiceChain implements RpcServiceRequestable {
...
+   * granularity.
+   *
+   * @param listener - The callback to be called.
+   * @returns An object with a `dispose` method which can be used to unregister
+   * the callback.
+   */
+  onBreak(
+    listener: CockatielEventToEventListenerWithData<
+      RpcService['onBreak'],
+      { primaryEndpointUrl: string }
```
Gudahtt: Nit: We can drop this from the type, and stop adding it to the listener payload

Gudahtt: Likewise for `onDegraded` and `onAvailable`

packages/network-controller/src/create-network-client.ts
cryptodev-2s: removed here https://github.com/MetaMask/core/pull/7166/commits/bcc840f449e3e129c41620d4606663e3a9dab784

packages/network-controller/src/create-network-client.ts
cryptodev-2s: removed here https://github.com/MetaMask/core/pull/7166/commits/bcc840f449e3e129c41620d4606663e3a9dab784

packages/network-controller/src/rpc-service/rpc-service-chain.ts
cryptodev-2s: removed here https://github.com/MetaMask/core/pull/7166/commits/bcc840f449e3e129c41620d4606663e3a9dab784

packages/network-controller/src/NetworkController.ts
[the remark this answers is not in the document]
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7166/commits/3918d08eaa077a4207566a60e4a25e5bc7fdb55b

packages/network-controller/src/rpc-service/rpc-service-chain.test.ts
```diff
@@ -537,159 +679,1445 @@ describe('RpcServiceChain', () => {
...
+      expect(onBreakListener).toHaveBeenNthCalledWith(2, {
+        endpointUrl: `${secondaryEndpointUrl}/`,
+        error: new Error("Fetch failed with status '503'"),
+      });
+    });
+
+    it('calls onServiceBreak each time the circuit of an RPC service in the chain breaks', async () => {
+      const primaryEndpointUrl = 'https://first.endpoint';
+      const secondaryEndpointUrl = 'https://second.endpoint';
+      const tertiaryEndpointUrl = 'https://second.endpoint';
```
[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7166/commits/916a0e2ba9427e9b987bd6aaf86803424094eb14

PR #7167, opened 2025-11-15
fix: remove crypto compare fallback

salimtb:
## Explanation

This PR removes the fallback to CryptoCompare API in both `CurrencyRatesController` and `TokenRatesController`.

### Current State
Previously, when the primary price data source failed, both controllers would fall back to the CryptoCompare API to fetch currency and token rates. This added complexity to the codebase and introduced a dependency on an additional external service.

### Solution
This change removes the CryptoCompare fallback logic from both controllers, simplifying the rate fetching flow and reducing external dependencies. The controllers will now rely solely on the primary price data source without attempting to fall back to CryptoCompare when the primary source fails.

### Rationale
- Reduces maintenance overhead by removing redundant API integration
- Simplifies error handling and rate fetching logic
- Aligns with the deprecation/removal of CryptoCompare as a fallback service
- The primary price data source has proven reliable enough that the fallback is no longer necessary

extension PR: https://github.com/MetaMask/metamask-extension/pull/37884
mobile PR: https://github.com/MetaMask/metamask-mobile/pull/22772

## References

- Extension PR: https://github.com/MetaMask/metamask-extension/pull/37884
- Related to removal of CryptoCompare API integration

## Checklist

---

> [!NOTE]
> Eliminates CryptoCompare fallbacks in currency and token rate controllers, replacing them with a token-prices service–based fallback and updating tests/behaviors accordingly.
> 
> - **BREAKING**:
>   - Remove CryptoCompare fallback from `CurrencyRateController` and `TokenRatesController`.
> - **CurrencyRateController**:
>   - Drop `fetchMultiExchangeRate`; primary source remains `fetchExchangeRates` from token-prices service.
>   - Add fallback using `fetchTokenPrices` mapped via `NetworkController:getState` to chain IDs and `getNativeTokenAddress`.
>   - On failures, update state with `null` values instead of throwing; skip empty/undefined native currencies.
>   - Allow `includeUsdRate`; maintain polling behavior; add `NetworkController:getState` to allowed actions.
> - **TokenRatesController**:
>   - Replace unsupported-native-currency path: fetch native+tokens in `usd`, then convert to native using native token USD price.
>   - Use `getNativeTokenAddress`; return empty when native USD price unavailable; keep batch fetching and consolidate state updates.
> - **Tests**:
>   - Remove `nock`/CryptoCompare stubs; update expectations to new fallback and failure semantics; add network state helpers.
> - **Docs/Changelog**:
>   - Update `CHANGELOG.md` noting breaking removal of CryptoCompare fallback; bump `@metamask/core-backend`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a4368861b67b4c2f124fcefa3c761cd2b5ef230. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/check-release/action.yml
+++ b/.github/actions/check-release/action.yml
@@ -1,21 +1,72 @@
 name: Check release
 description: Check for conflicts in packages being released in this PR.
 
-inputs:
-  pull-request:
-    description: 'The pull request number to get changed files from.'
-    required: true
-
 runs:
   using: composite
   steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
       with:
         fetch-depth: 0
 
+    - name: Get merge base
+      id: merge-base
+      shell: bash
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.ref || github.event.merge_group.base_ref }}
+      run: |
+        set -euo pipefail
+
+        # Strip invalid prefix from `github.event.merge_group.base_ref`
+        # It comes prefixed with `refs/heads/`, but the branch is not checked out in this context
+        # We need to express it as a remote branch
+        PREFIXED_REF_REGEX='refs/heads/(.+)'
+        if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+          BASE_REF="${BASH_REMATCH[1]}"
+        fi
+
+        MERGE_BASE=$(git merge-base HEAD "refs/remotes/origin/$BASE_REF")
+        echo "MERGE_BASE=$MERGE_BASE" >> "$GITHUB_OUTPUT"
+
+    - name: Check if the commit or pull request is a release
+      id: is-release
+      uses: MetaMask/action-is-release@v2
+      with:
+        commit-starts-with: 'Release [version],Release v[version],Release/[version],Release/v[version],Release `[version]`'
+        commit-message: ${{ github.event.pull_request.title }}
+        before: ${{ steps.merge-base.outputs.MERGE_BASE }}
+
+    # `action-is-release` checks out the repository with `fetch-depth: 2`, so we
+    # need to check it out again with full history.
+    - name: Checkout repository
+      uses: actions/checkout@v5
+      with:
+        fetch-depth: 0
+
+    - name: Get pull request number
+      id: pr-number
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
+      shell: bash
+      env:
+        EVENT_NAME: ${{ github.event_name }}
+        PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
+        MERGE_GROUP_HEAD_REF: ${{ github.event.merge_group.head_ref }}
+      run: |
+        if [ "$EVENT_NAME" = "pull_request" ]; then
+          echo "PR_NUMBER=$PULL_REQUEST_NUMBER" >> "$GITHUB_OUTPUT"
+        elif [ "$EVENT_NAME" = "merge_group" ]; then
+          PR_NUMBER_REGEX='/pr-([0-9]+)-'
+          if [[ "$MERGE_GROUP_HEAD_REF" =~ $PR_NUMBER_REGEX ]]; then
+            echo "PR_NUMBER=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+          else
+            echo "::error::Could not extract PR number from merge group head ref: $MERGE_GROUP_HEAD_REF."
+            exit 1
+          fi
+        fi
+
     - name: Get target reference
       id: get-target
+      if: steps.is-release.outputs.IS_RELEASE == 'true'
       shell: bash
       env:
         EVENT_NAME: ${{ github.event_name }}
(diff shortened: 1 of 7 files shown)


Prithpal-Sooriya:
Tested extension - confirmed that I do not see any crypto-compare API calls

packages/assets-controllers/src/CurrencyRateController.test.ts
```diff
@@ -1225,6 +1288,510 @@ describe('CurrencyRateController', () => {
...
+      await controller.updateExchangeRate(['ETH', 'POL']);
+
+      const conversionDate = getStubbedDate() / 1000;
+      expect(controller.state).toStrictEqual({
+        currentCurrency: 'usd',
+        currencyRates: {
+          ETH: {
+            conversionDate,
+            conversionRate: 2500.5,
+            usdConversionRate: null, // Line 283
```
Prithpal-Sooriya: Can you recheck this test file, make sure we don't include `// Line` comments

packages/assets-controllers/src/CurrencyRateController.test.ts
```diff
@@ -1225,6 +1288,510 @@ describe('CurrencyRateController', () => {
...
+
+      const controller = new CurrencyRateController({
+        messenger,
+        state: { currentCurrency: 'usd' },
+        tokenPricesService,
+      });
+
+      await controller.updateExchangeRate(['ETH', 'POL']);
+
+      // Should log error for POL (line 286-289)
```
Prithpal-Sooriya: Same, remove code comment lines

packages/assets-controllers/src/CurrencyRateController.test.ts
```diff
@@ -1225,6 +1288,510 @@ describe('CurrencyRateController', () => {
...
+      expect(controller.state).toStrictEqual({
+        currentCurrency: 'usd',
+        currencyRates: {
+          ETH: {
+            conversionDate,
+            conversionRate: 2500.5,
+            usdConversionRate: null,
+          },
+          POL: {
+            conversionDate: null, // Lines 292-294: error case
```
Prithpal-Sooriya: Same, remove code comment lines

Prithpal-Sooriya: just double check this file for other line comments 🙏🏾

packages/assets-controllers/src/CurrencyRateController.ts
```diff
@@ -215,7 +223,104 @@ export class CurrencyRateController extends StaticIntervalPollingController<Curr
...
+
+          return acc;
+        },
+        {} as Record<string, { fetchedCurrency: string; chainId: Hex }>,
+      );
+
+      // Step 3: Fetch token prices for each chainId
+      const ratesFromTokenPrices = await Promise.all(
+        Object.entries(currencyToChainIds).map(
+          async ([nativeCurrency, { chainId }]) => {
```
Prithpal-Sooriya: Can we change this to a batch? Worried of calling `Promise.all` on many chains at once.

packages/assets-controllers/src/TokenRatesController.ts
```diff
@@ -775,53 +736,64 @@ export class TokenRatesController extends StaticIntervalPollingController<TokenR
...
-      }),
-      getCurrencyConversionRate({
-        from: FALL_BACK_VS_CURRENCY,
-        to: nativeCurrency,
-      }),
-    ]);
-
-    if (fallbackCurrencyToNativeCurrencyConversionRate === null) {
+        nativeCurrency: 'usd',
+      });
```
bergarces, line 755: I don't think you need to make 2 calls.

The second one will automatically include the native currency in the result.

bergarces: Which by the way is something I will be removing soon, as that means fetching the native currency rate on every batch if there are too many tokens.

packages/assets-controllers/CHANGELOG.md
bergarces, line 1: We are missing a breaking change, since we need to add a new action to CurrencyRateController.

packages/assets-controllers/src/CurrencyRateController.ts
```diff
@@ -215,7 +223,110 @@ export class CurrencyRateController extends StaticIntervalPollingController<Curr
...
+        return ratesFromTokenPricesService;
+      }
+    } catch (error) {
+      console.error(
+        'Failed to fetch exchange rates from token prices service.',
+        error,
+      );
+    }
 
     try {
```
bergarces: Shouldn't we remove the crypto-compare fallback that comes after this completely?

packages/assets-controllers/src/CurrencyRateController.ts
```diff
@@ -215,7 +223,110 @@ export class CurrencyRateController extends StaticIntervalPollingController<Curr
...
+      const currencyToChainIds = Object.entries(nativeCurrenciesToFetch).reduce(
+        (acc, [nativeCurrency, fetchedCurrency]) => {
+          // Find chainIds that have this native currency
+          const chainIds = (
+            Object.entries(networkConfigurations) as [
+              Hex,
+              NetworkConfiguration,
+            ][]
+          )
+            .filter(
```
bergarces: Why not just use find? It'll return the first one or undefined.

packages/assets-controllers/src/TokenRatesController.ts
salimtb: is it ok to do it on your PR for now ?

bergarces: Yes. Let's do it that way. I'll be copying it over today.

PR #7169, opened 2025-11-17
ci: Refactor release conflicts action to contain all logic

Mrtenz:
## Explanation

This updates the release conflicts action to move all logic to within the action, making it easier to extract the action to a separate repository later.

## Checklist

---

> [!NOTE]
> Moves all release-conflict detection logic into `.github/actions/check-release` and updates the main workflow to call it directly for PRs and merge groups.
> 
> - **CI / Composite Action** (`.github/actions/check-release/action.yml`):
>   - Migrate full release-conflict logic into the action (no inputs required), using GitHub context.
>   - Add steps to compute merge base, detect release commits, resolve PR number for `pull_request` and `merge_group`, determine target ref, scan for released packages vs. files changed ahead of the PR, and fail on conflicts.
>   - Post/manage PR comments for conflicts; re-checkout after `action-is-release`; switch to `actions/checkout@v5` and fetch full history.
>   - Fix branch ref when computing `BEFORE` (`refs/remotes/origin/$PULL_REQUEST_BRANCH`).
> - **Workflow** (`.github/workflows/main.yml`):
>   - Simplify `check-release` job to call the composite action directly for non-push events, removing duplicated inline logic (merge-base, is-release, PR number, inputs).
>   - Keep `is-release` (push) and other jobs unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4feecb7eb78206f89e5b1186cc00f917f344228. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/check-release/action.yml
+++ b/.github/actions/check-release/action.yml
@@ -25,6 +25,13 @@ runs:
         commit-message: ${{ github.event.pull_request.title }}
         before: ${{ steps.merge-base.outputs.MERGE_BASE }}
 
+    # `action-is-release` checks out the repository with `fetch-depth: 2`, so we
+    # need to check it out again with full history.
+    - name: Checkout repository
+      uses: actions/checkout@v5
+      with:
+        fetch-depth: 0
+
     - name: Get pull request number
       id: pr-number
       if: steps.is-release.outputs.IS_RELEASE == 'true'
@@ -72,8 +79,6 @@ runs:
         TARGET: ${{ steps.get-target.outputs.TARGET }}
       run: |
         set -euo pipefail
-
-        git fetch --unshallow origin main
 
         mapfile -t PACKAGES < <(find packages -maxdepth 2 -name "package.json" -not -path "*/node_modules/*")
         RELEASED_PACKAGES=()


.github/actions/check-release/action.yml
```diff
@@ -31,14 +64,16 @@ runs:
...
       env:
         GH_TOKEN: ${{ github.token }}
-        PULL_REQUEST: ${{ inputs.pull-request }}
+        PULL_REQUEST: ${{ steps.pr-number.outputs.PR_NUMBER }}
         TARGET: ${{ steps.get-target.outputs.TARGET }}
       run: |
         set -euo pipefail
-        git fetch origin main
+
+        git fetch --unshallow origin main
```
Mrtenz: Not sure why, but if we don't unshallow here, the `git merge-base` below will fail.

.github/workflows/main.yml
```diff
@@ -72,52 +72,10 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
```
Gudahtt, line 75: Nit: Curious that we were using v5 here, but now we're using v4 in the action.

Not related to your refactor, but I'm guessing this was accidental

Mrtenz: Yeah, this was accidental. We bumped to `v5` around the time I was working on this action, and I forgot to update my PR. Fixed it here: e2ce58e9dae8ff433d667efeb6899c5ab8dca565

.github/actions/check-release/action.yml
```diff
@@ -1,21 +1,54 @@
...
+          echo "PR_NUMBER=$PULL_REQUEST_NUMBER" >> "$GITHUB_OUTPUT"
+        elif [ "$EVENT_NAME" = "merge_group" ]; then
+          PR_NUMBER_REGEX='/pr-([0-9]+)-'
+          if [[ "$MERGE_GROUP_HEAD_REF" =~ $PR_NUMBER_REGEX ]]; then
+            echo "PR_NUMBER=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+          else
+            echo "::error::Could not extract PR number from merge group head ref: $MERGE_GROUP_HEAD_REF."
+            exit 1
+          fi
+        fi
```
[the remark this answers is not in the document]
Mrtenz: We only run this for pull request and merge queue events.

.github/actions/check-release/action.yml
Gudahtt: Maybe it's because the pull request branch wasn't fetched? It's not totally clear but it seems from reading the docs that `--unshallow` (when used in a non-shallow repo) ends up fetching all remote branches.

You might get away with removing it and speeding this up a bit by fetching the pull request branch later, after you get the branch name. Or move the "# Fetch the pull request branch" step above `git fetch`, and fetch both at once

Mrtenz: I've tried several variations of `git fetch "$PULL_REQUEST_BRANCH"`, but couldn't get it to work without `--unshallow` here. For example, fetching both `main` and `$PULL_REQUEST_BRANCH` [here](https://github.com/ts-bridge/core-merge-queue-test/actions/runs/19432299204/job/55594057970?pr=9).

.github/workflows/main.yml
```diff
@@ -72,52 +72,10 @@ jobs:
...
-            PR_NUMBER_REGEX='/pr-([0-9]+)-'
-            if [[ "$MERGE_GROUP_HEAD_REF" =~ $PR_NUMBER_REGEX ]]; then
-              echo "PR_NUMBER=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
-            else
-              echo "::error::Could not extract PR number from merge group head ref: $MERGE_GROUP_HEAD_REF."
-              exit 1
-            fi
-          fi
-
+        uses: actions/checkout@v4
```
Gudahtt: Nit: Also using v4 here

Mrtenz: 62079ac592d98bc283ff89bf0843e8da50a67c8f

.github/actions/check-release/action.yml
Gudahtt: Interesting. From local testing, it seems like it will work if you fetch the pull request branch with the syntax `git fetch ${branch}:${branch}`, then drop the `refs/remotes/origin/` prefix in the `merge-base` command (i.e. just `git merge-base "refs/remotes/origin/main" "$PULL_REQUEST_BRANCH"`

I don't completely follow _why_ though.

Mrtenz: Looked into it more, and it looks like `actions/checkout` will fetch all branches anyway? The additional `git fetch --unshallow` doesn't seem to refetch all branches.

Gudahtt: If `actions/checkout` fetches all branches, why would `git merge-base` fail 🤔

Mrtenz: I'm confused about it as well 😅 It's fetching everything with `fetch-depth: 0`, but still seems to be shallow somehow?

Gudahtt: Ah, maybe it's because we're checking out again in `action-is-release` with a depth of 2: https://github.com/MetaMask/action-is-release/blob/d55be15c02da4b30aa080aee4dd96fde1dbdbd30/action.yml#L27

Mrtenz: Oh, yeah, that would explain why it suddenly behaves differently after this refactor. Not sure how to go about this, do we want to update `action-is-release` to clone to a subfolder?

Gudahtt: Hmm, interesting idea. We can more simply workaround it for now by doing another checkout after calling `action-is-release`, but I like the idea of preserving the existing checkout (if there is one) by using a subfolder.

Mrtenz: This seems to work: a0c342a1db6331b561b4e247c01b5200c31ba913

We can remove this later if we change `action-is-release` to checkout in a different folder.

PR #7185, opened 2025-11-18
chore(bridge-controller): faster utils

Nodonisko:
## Explanation

While working on https://github.com/MetaMask/metamask-mobile/pull/22744 I noticed that if do this small change of pre-converting values I can save up to 25% of time when looping though big array of tokens where this functions were used multiple times per one loop.

## References

* Related to https://github.com/MetaMask/metamask-mobile/pull/22744 

## Checklist

---

> [!NOTE]
> Precomputes chain ID string constants and uses them in `isSolanaChainId`, `isBitcoinChainId`, and `isTronChainId` within `utils/bridge.ts`.
> 
> - **utils/bridge.ts**:
>   - Add constants: `SOLANA_MAINNET_CHAIN_ID_STRING`, `SOLANA_CHAIN_ID_STRING`, `BITCOIN_MAINNET_CHAIN_ID_STRING`, `BITCOIN_CHAIN_ID_STRING`, `TRON_MAINNET_CHAIN_ID_STRING`, `TRON_CHAIN_ID_STRING`.
>   - Update `isSolanaChainId`, `isBitcoinChainId`, `isTronChainId` to compare against precomputed strings instead of inline `toString()` calls.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 68f77e1f1848bae3e01d2fc3179fcaa782789a66. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-controller/src/utils/bridge.ts
+++ b/packages/bridge-controller/src/utils/bridge.ts
@@ -168,6 +168,8 @@ export const isNativeAddress = (address?: string | null) =>
       (reference) => address.includes(reference) || reference.endsWith(address),
     ));
 
+const SOLANA_MAINNET_CHAIN_ID_STRING = SolScope.Mainnet.toString();
+const SOLANA_CHAIN_ID_STRING = ChainId.SOLANA.toString();
 /**
  * Checks whether the chainId matches Solana in CaipChainId or number format
  *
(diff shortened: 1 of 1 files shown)


packages/bridge-controller/src/utils/bridge.ts
```diff
@@ -178,25 +180,31 @@ export const isSolanaChainId = (
   chainId: Hex | number | CaipChainId | string,
 ) => {
   if (isCaipChainId(chainId)) {
-    return chainId === SolScope.Mainnet.toString();
+    return chainId === SOLANA_MAINNET_CHAIN_ID_STRING;
```
mcmire, line 183: Why do we have to convert these chain IDs to strings at all? Shouldn't they already be strings? (If the reason is because we're getting a lint error due to enums, we should be able to bypass that.)

Nodonisko: Yes, these Scopes are strings actually but ChainId enum lower are numbers. I can probably use typecast for these if it's preferred.

PR #7186, opened 2025-11-18
Keep network statuses up to date as requests are made

mcmire:
## Explanation

NetworkController has the ability to analyze an RPC endpoint and capture its availability status (that is, whether the controller is able to make successful requests to the endpoint). However, this step occurs automatically only once, when the RPC endpoint's network is switched to, and so any changes in status while the network is being used will not be reflected in state. This problem can be mitigated by periodically calling `lookupNetwork` manually, but this is awkward, and usage of this method should be kept in check so as not to create too many requests.

Ideally, the controller should keep track of network status as requests are made. This commit implements this change by hooking into network client events added in a previous commit.

Note that this PR does not remove `lookupNetwork` or touch the existing behavior for this method. So with these changes there are now two strategies at play for updating the network status. This should be okay for the time being, although we should look to refactor this in the future.

## References

## Checklist

---

> [!NOTE]
> Automatically updates network status metadata in response to chain-level RPC events; introduces Degraded status, refactors metadata update API, and adds comprehensive provider tests.
> 
> - **NetworkController**:
>   - Auto-subscribes to `NetworkController:rpcEndpointChainUnavailable`, `...ChainDegraded`, and `...ChainAvailable` to update `networksMetadata[networkClientId].status` in real time.
>   - Refactors `#updateMetadataForNetwork` to accept a `metadata` object and updates all call sites (`#lookupGivenNetwork`, `#lookupSelectedNetwork`).
> - **Constants**:
>   - Add `NetworkStatus.Degraded` and clarify enum semantics.
> - **Tests**:
>   - Add `tests/NetworkController.provider.test.ts` covering status transitions (available ↔ degraded ↔ unavailable), retries, failover, and recovery.
>   - Move/export `withController` helper to `tests/helpers.ts` and remove inline duplicate.
> - **Changelog**:
>   - Document automatic status updates via chain-level RPC events.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e623211520efd0b84e86125db995a0b3bd3736dd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/network-controller/tests/NetworkController.provider.test.ts
+++ b/packages/network-controller/tests/NetworkController.provider.test.ts
@@ -357,7 +357,7 @@ describe('NetworkController provider tests', () => {
     );
   });
 
-  it('does not transition the status of a network client from "degraded" the first time a failover is activated but it does not return a 2xx response', async () => {
+  it('does not transition the status of a network client from "degraded" the first time a failover is activated if it returns a non-2xx response', async () => {
     const primaryEndpointUrl = 'https://first.endpoint';
     const secondaryEndpointUrl = 'https://second.endpoint';
     const networkClientId = 'AAAA-AAAA-AAAA-AAAA';
@@ -441,7 +441,7 @@ describe('NetworkController provider tests', () => {
     );
   });
 
-  it('does not transition the status of a network client from "degraded" the first time a failover is activated but requests are slow to complete', async () => {
+  it('does not transition the status of a network client from "degraded" the first time a failover is activated if requests are slow to complete', async () => {
     const primaryEndpointUrl = 'https://first.endpoint';
     const secondaryEndpointUrl = 'https://second.endpoint';
     const networkClientId = 'AAAA-AAAA-AAAA-AAAA';


packages/network-controller/src/NetworkController.ts
```diff
@@ -1965,12 +1992,15 @@ export class NetworkController extends BaseController<
...
-      const meta = state.networksMetadata[networkClientId];
-      meta.status = networkStatus;
-      if (isEIP1559Compatible === undefined) {
-        delete meta.EIPS[1559];
-      } else {
-        meta.EIPS[1559] = isEIP1559Compatible;
+      const newMetadata = state.networksMetadata[networkClientId];
+      newMetadata.status = metadata.networkStatus;
+
+      if ('isEIP1559Compatible' in metadata) {
```
Gudahtt, line 1998: Is it intended that we no longer reset `isEIP1559Compatible` if it's omitted from the metadata object? That is a functional change from before

Gudahtt: Oh I see, it is intended, I get it now. The old invocations all include this property, so this doesn't functionally change those invocations.

packages/network-controller/tests/NetworkController.provider.test.ts
```diff
@@ -0,0 +1,894 @@
...
+              path: ['networksMetadata', 'AAAA-AAAA-AAAA-AAAA', 'status'],
+              value: 'available',
+            },
+          ],
+        );
+      },
+    );
+  });
+
+  it('does not transition the status of a network client from "degraded" the first time a failover is activated but it does not return a 2xx response', async () => {
```
Gudahtt: Nit: I found this difficult to understand because of the awkward "but" and the double-negative. Here's an attempt at simplifying:

```suggestion
  it('does not transition the status of a network client from "degraded" the first time a failover is activated if it returns a non-2xx response', async () => {
```

packages/network-controller/tests/NetworkController.provider.test.ts
```diff
@@ -0,0 +1,894 @@
...
+        await expect(provider.request(request)).rejects.toThrow(expectedError);
+
+        expect(controller.state.networksMetadata[networkClientId].status).toBe(
+          NetworkStatus.Degraded,
+        );
+      },
+    );
+  });
+
+  it('does not transition the status of a network client from "degraded" the first time a failover is activated but requests are slow to complete', async () => {
```
Gudahtt: Nit: Another attempted simplification:

```suggestion
  it('does not transition the status of a network client from "degraded" the first time a failover is activated if requests are slow to complete', async () => {
```

packages/network-controller/src/NetworkController.ts
cryptodev-2s: I believe this functional change to make `isEIP1559Compatible` optional instead of required. Previously, it was always processed (even when `undefined`, which would delete `EIPS[1559]`). Now, if omitted, `EIPS[1559]` is left unchanged.

So yes this was intentional to enable partial updates. The new event handlers (lines 1446, 1454, 1462) only update `networkStatus` and should not reset EIP-1559 compatibility. Existing call sites (`#lookupGivenNetwork`, `#lookupSelectedNetwork`) still pass `isEIP1559Compatible`, preserving their behavior.

Event handlers can update network status without affecting EIP-1559 compatibility, while `lookupNetwork` methods continue to update both fields as before.

packages/network-controller/tests/NetworkController.provider.test.ts
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7186/commits/e623211520efd0b84e86125db995a0b3bd3736dd

packages/network-controller/tests/NetworkController.provider.test.ts
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7186/commits/e623211520efd0b84e86125db995a0b3bd3736dd

Gudahtt:
>Note that this PR does not remove lookupNetwork or touch the existing behavior for this method. So with these changes there are now two strategies at play for updating the network status. This should be okay for the time being, although we should look to refactor this in the future.

I don't think leaving `lookupNetwork` here helps us. It just makes the network status harder to understand. But still, dealing with it in a later PR sounds good to me. It doesn't seem like it would make the status less accurate than it is today at least.

Plus I'd hesitate to remove it completely without finding a more elegant way of dealing with the EIP1559 compatibility check.

Edit: Maybe in a future PR, we could consider updating `lookupNetwork` to just update the EIP1559 compatibility, and rely on the network client to indirectly _also_ update the network status if appropriate.

Gudahtt (approved):
LGTM!

PR #7192, opened 2025-11-19
feat: Add StorageService for offloading large controller data

andrepimenta:
# Add StorageService for Large Controller Data

**Current state**: MetaMask Mobile Engine state is 10.79 MB, with 92% (9.94 MB) concentrated in just 2 controllers:
- **SnapController**: 5.95 MB of snap source code (55% of total state)
- **TokenListController**: 3.99 MB of token metadata cache (37% of state)

This data is **rarely accessed** after initial load but stays in Redux state, causing:
- Slow app startup (parsing 10.79 MB on every launch)
- High memory usage (all data loaded even if not needed)
- Slow persist operations (up to 6.26 MB written per controller change)

**Why change**: Controllers need a way to store large, infrequently-accessed data outside of Redux state while maintaining platform portability and testability.

**New package**: `@metamask/storage-service`

A platform-agnostic service that allows controllers to offload large data from state to persistent storage via messenger actions.

**How it works**:
1. Controllers call `StorageService:setItem` via messenger to store large data
2. StorageService saves to platform-specific storage (FilesystemStorage on mobile, IndexedDB on extension)
3. StorageService publishes events (`StorageService:itemSet:{namespace}`) so other controllers can react
4. Controllers call `StorageService:getItem` to load data lazily (only when needed)

**Storage adapter pattern**:
```typescript
// Service accepts platform-specific adapter (like ErrorReportingService)
const service = new StorageService({
  messenger,
  storage: filesystemStorageAdapter, // Mobile provides this
});
```

**Example controller usage**:
```typescript
// Store data (out of state)
await this.messenger.call(
  'StorageService:setItem',
  'MyController',
  'dataKey',
  largeData,
);

// Load on demand (lazy loading)
const data = await this.messenger.call(
  'StorageService:getItem',
  'MyController',
  'dataKey',
);

// Subscribe to changes (optional)
this.messenger.subscribe(
  'StorageService:itemSet:MyController',
  (key, value) => {
    // React to storage changes
  },
);
```

### Why this architecture?

**Platform-agnostic**: Service defines `StorageAdapter` interface; clients provide implementation (mobile: FilesystemStorage, extension: IndexedDB, tests: in-memory)

**Messenger-integrated**: Controllers access storage via messenger actions, no direct dependencies

**Event-driven**: Controllers can subscribe to storage changes without coupling

**Testable**: InMemoryAdapter provides zero-config testing (no mocking needed)

**Proven pattern**: Follows ErrorReportingService design (accepts platform-specific function)

### Expected impact?

**With both controllers optimized**:
- State: 10.79 MB → 0.85 MB (**92% reduction**)
- App startup: 92% faster state parsing
- Memory: 9.94 MB freed
- Disk I/O: Up to 9.94 MB less per persist

**This PR adds infrastructure** - Controllers can now use StorageService. Separate PRs will integrate with SnapController and TokenListController.

## References

- **ADR**: [0017-storage-service-large-data.md](https://github.com/MetaMask/decisions/blob/core/0017-storage-service-large-data/decisions/core/0017-storage-service-large-data.md)
- **Mobile PR**: https://github.com/MetaMask/metamask-mobile/pull/22943

## Checklist

  - 100% test coverage (44 tests)
  - Tests for all storage operations, namespace isolation, events, error handling
  - Real-world usage test (simulates 6 MB snap source code)
  - Complete README with examples
  - JSDoc for all public APIs
  - Architecture documentation in ADR
  - CHANGELOG.md created for initial release
  - No breaking changes (new package)
  - N/A - No breaking changes
  - Consumer PRs will be created after this is merged and released

---

> [!NOTE]
> Introduces `@metamask/storage-service` with messenger-based APIs and an in-memory adapter to offload large controller data, plus repo wiring and ownership updates.
> 
> - **New Package**: `packages/storage-service`
>   - `StorageService`: Messenger-exposed methods `setItem`, `getItem`, `removeItem`, `getAllKeys`, `clear`; publishes `StorageService:itemSet:{namespace}` events.
>   - `InMemoryStorageAdapter`: Default non-persistent adapter implementing `StorageAdapter` with key prefix `storageService:`.
>   - Types/exports: `StorageAdapter`, `StorageGetResult`, messenger types, action types file, constants `SERVICE_NAME`, `STORAGE_KEY_PREFIX`.
> - **Tests & Docs**:
>   - Comprehensive unit tests for service and adapter (namespace isolation, events, errors); Jest config with 100% coverage thresholds.
>   - `README.md` with usage/examples; `CHANGELOG.md` initialized; dual-license files.
> - **Repo Wiring**:
>   - Add package to `tsconfig.build.json`, `yarn.lock`, exports/index, and TypeDoc config.
>   - Update `.github/CODEOWNERS` and `teams.json` to include `storage-service` ownership.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6bf384a8f6ffda07be30b30a1e19b0e8ae54e805. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/storage-service/src/StorageService-method-action-types.ts
+++ b/packages/storage-service/src/StorageService-method-action-types.ts
@@ -38,7 +38,7 @@ export type StorageServiceSetItemAction = {
  *
  * @param namespace - Controller namespace (e.g., 'SnapController').
  * @param key - Storage key (e.g., 'npm:@metamask/example-snap:sourceCode').
- * @returns Parsed JSON data or null if not found.
+ * @returns StorageGetResult: { result } if found, {} if not found, { error } on failure.
  */
 export type StorageServiceGetItemAction = {
   type: `StorageService:getItem`;


packages/storage-service/CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- Initial release of `@metamask/storage-service`
```
Gudahtt: Feel free to include whatever details you want here, but usually we only list "Initial release of ___" for the first release, since nothing has really _changed_ but that.

packages/storage-service/LICENSE
Gudahtt, line 1: Sorry this shouldn't be the default license, I keep meaning to change that. We use a dual license of MIT+Apache2 by default, as requested by legal. See the analytics controller for a recent example of how this is setup: https://github.com/MetaMask/core/tree/main/packages/analytics-controller

MIT is just used for the package before we adopted this policy

packages/storage-service/README.md
```diff
@@ -0,0 +1,377 @@
+# `@metamask/storage-service`
+
+A platform-agnostic service for storing large, infrequently accessed controller data outside of memory.
+
+## Problem
+
+Controllers store large, infrequently-accessed data in Redux state, causing:
```
Gudahtt: Some of this content seems like... marketing copy? Or like an attempt to explain the merits of this change to the business, e.g. like in the ADR. Both here and the "Real world impact" section later.

The README ideally would explain purpose and usage, but we don't need to make a business case here, it just gets in the way. All of this data will be obsolete very soon anyway.

packages/storage-service/README.md
```diff
@@ -0,0 +1,377 @@
...
+const service = new StorageService({
+  messenger: myMessenger,
+  storage: new InMemoryStorageAdapter(),
+});
+
+await service.setItem('MyController', 'myKey', { data: 'value' });
+const data = await service.getItem('MyController', 'myKey');
+```
+
+## API
```
Gudahtt: Nit: Given that we use inline documentation for public methods, we don't typically document an API in the README.

We used to regularly generate dedicated API docs from inline docs. This is broken right now but we plan on fixing it soon. Documenting it here seems like more trouble than it would be worth - we'll have to keep it up-to-date with each change, and contributors are not accustomed to looking for it here since we don't include it elsewhere

packages/storage-service/README.md
```diff
@@ -0,0 +1,377 @@
...
+export type StorageAdapter = {
+  getItem(namespace: string, key: string): Promise<unknown>;
+  setItem(namespace: string, key: string, value: unknown): Promise<void>;
+  removeItem(namespace: string, key: string): Promise<void>;
+  getAllKeys(namespace: string): Promise<string[]>;
+  clear(namespace: string): Promise<void>;
+};
+```
+
+## When to Use
```
Gudahtt: Nit: Personally I'd put this higher up, this seems important

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * @param namespace - Controller namespace (e.g., 'SnapController').
+   * @param key - Storage key (e.g., 'npm:@metamask/example-snap:sourceCode').
+   * @returns Parsed data or null if not found.
+   * @template T - The type of the value being retrieved.
+   */
+  async getItem<T>(namespace: string, key: string): Promise<T | null> {
+    // Adapter handles deserialization and unwrapping
+    const result = await this.#storage.getItem(namespace, key);
+    return result as T | null;
```
Gudahtt: The types don't seem correct here. How do we know this will be `T | null`?

packages/storage-service/src/types.ts
```diff
@@ -0,0 +1,220 @@
...
+ * Storage key prefix for all keys managed by StorageService.
+ * Keys are formatted as: {STORAGE_KEY_PREFIX}{namespace}:{key}
+ * Example: 'storageService:SnapController:snap-id:sourceCode'
+ */
+export const STORAGE_KEY_PREFIX = 'storageService:';
+
+/**
+ * Action for storing data in the storage service.
+ */
+export type StorageServiceSetItemAction = {
```
Gudahtt: Nit: We have a `generate-method-action-types` script that can generate these automatically from the service, and keep them updated automatically as the code changes.

See here for an example of how to use this: https://github.com/MetaMask/core/blob/3db312fb349bd1d01caad15bc49ba9694f494c4e/packages/profile-metrics-controller/src/ProfileMetricsService.ts#L139

It relies on an exposed constant, `MESSENGER_EXPOSED_METHODS`, which can also be used to bulk-register the actions

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,231 @@
...
+   */
+  async setItem<T>(namespace: string, key: string, value: T): Promise<void> {
+    // Adapter handles serialization and wrapping with metadata
+    await this.#storage.setItem(namespace, key, value as never);
+
+    // Publish event so other controllers can react to changes
+    // Event type: StorageService:itemSet:namespace
+    // Payload: [value, key]
+    this.#messenger.publish(
+      `${SERVICE_NAME}:itemSet:${namespace}` as `${typeof SERVICE_NAME}:itemSet:${string}`,
```
Gudahtt: The cast can be avoided here using `as const`, which preserves type safety:
```suggestion
      `${SERVICE_NAME}:itemSet:${namespace}` as const,
```

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,231 @@
...
+   */
+  async removeItem(namespace: string, key: string): Promise<void> {
+    // Adapter builds full storage key (e.g., mobile: 'storageService:namespace:key')
+    await this.#storage.removeItem(namespace, key);
+
+    // Publish event so other controllers can react to removal
+    // Event type: StorageService:itemRemoved:namespace
+    // Payload: [key]
+    this.#messenger.publish(
+      `${SERVICE_NAME}:itemRemoved:${namespace}` as `${typeof SERVICE_NAME}:itemRemoved:${string}`,
```
Gudahtt: Likewise here, no cast is needed
```suggestion
      `${SERVICE_NAME}:itemRemoved:${namespace}` as const,
```

packages/storage-service/src/StorageService.ts
Gudahtt: This problem seems widespread. The storage service interactions all have incorrect types

Gudahtt: Hmm. Since we have no schema or validation for the data, perhaps it's easiest to adopt the types used by the storage adapter, and always treat the values as `unknown`.

Gudahtt: Alternatively, we could pass in the expected state shape (across _all_ namespaces), and have proper type checking, like we do for controllers. But the types would get a bit complex if we went this route.

Also if we went that route, we don't really have a way to _validate_ the data, so the _returns_ would have to be unknown (otherwise the types would be lying to the client, in the event the data was corrupted or migrated from an earlier version or something). Or something limited to the validation we _can_ do, like "not null".

packages/storage-service/src/types.ts
```diff
@@ -0,0 +1,220 @@
...
+ * messenger.subscribe('StorageService:itemSet:TokenListController', (value, key) => {
+ *   // value = the data that was set
+ *   // key = 'cache:0x1', 'cache:0x38', etc.
+ *   if (key.startsWith('cache:')) {
+ *     const chainId = key.replace('cache:', '');
+ *     // React to cache change for specific chain
+ *   }
+ * });
+ */
+export type StorageServiceItemSetEvent = {
```
Gudahtt, line 152: Interesting, did you have any use-cases in mind for these two events?

No objections to including them of course, they're simple enough. I just wasn't sure if you had a purpose in mind for them or not.

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Clear all data for a namespace.
+   * Delegates to storage adapter which handles clearing.
+   *
+   * @param namespace - Controller namespace (e.g., 'SnapController').
+   */
+  async clear(namespace: string): Promise<void> {
+    await this.#storage.clear(namespace);
+  }
```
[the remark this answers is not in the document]
Gudahtt: Good point

packages/storage-service/src/types.ts
andrepimenta: Hey @Gudahtt, I was looking at the TokenListController to understand how we could potentially move `tokensChainsCache` out of the controller state to disk via StorageService. And I noticed there were other controllers subscribing to changes in state of the TokenListController, specifically to do something with the `tokensChainsCache` data when it changed. 

So if `tokensChainsCache` moved to StorageService and potentially not live in the Controller state anymore, we would need a way to let those other controllers know when it changed. In other words, to have a way for them to subscribe to the itemSet event so they know when new data is available.

Of course, controllers could also just emit their own custom events via the messenger (i.e. `tokensChainsCacheChanged` event), but I think it's one less thing they would need to implement. It would be nice to support this functionality out of the box for the StorageService.

One thing I am thinking about is removing the itemRemoved event. A controller can subscribe to itemSet to know when new data is available. But I believe it doesn't typically need to know when a controller removes its own data. In this way, we also don't need to take care of the "clear" event which simplifies things as well.

packages/storage-service/src/StorageService.ts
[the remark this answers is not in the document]
andrepimenta: Thinking about removing the itemRemoved event, check thread here for rationale:
https://github.com/MetaMask/core/pull/7192#discussion_r2566653551

packages/storage-service/src/types.ts
Gudahtt: Right... good point, I guess we do need to know when the data changes.

I was thinking that we don't really need to know when the data is _persisted_, but it's not the persistence that we care about here, it's the change.

Gudahtt: We may need to know when state is removed as well then. We may want to delete a token from the ignore list that is no longer present in metadata for example (hypothetically - not sure that we actually want to do that).

andrepimenta: For the record, we have decided to remove `itemRemoved` events as they are not as useful. If data changes in any way and gets rewritten to disc, that is still going to be an `itemSet` event, so that covers all cases.

packages/storage-service/LICENSE
andrepimenta: Thanks @Gudahtt, this should be done now if you'd like to confirm.

packages/storage-service/src/StorageService.ts
andrepimenta: Hey @Gudahtt, thanks! This should now be fixed if you'd like to confirm, before resolving the conversation.

packages/storage-service/package.json
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  "homepage": "https://github.com/MetaMask/core/tree/main/packages/storage-service#readme",
+  "bugs": {
+    "url": "https://github.com/MetaMask/core/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MetaMask/core.git"
+  },
+  "license": "MIT",
```
Gudahtt: This field needs an update to reference both licenses as well:

```suggestion
  "license": "(MIT OR Apache-2.0)",
```

I noticed this is wrong in the example I cited too 😅 I'll fix this field for the analytics controller in a separate PR

packages/storage-service/src/StorageService.ts
Gudahtt: Looks good!

packages/storage-service/package.json
andrepimenta: Hahah good catch!

packages/storage-service/src/types.ts
andrepimenta: @Gudahtt, this should be fixed, thank you. Please confirm if it's ok now before we resolve the conversation

packages/storage-service/CHANGELOG.md
andrepimenta: Thanks @Gudahtt, this is now fixed!

packages/storage-service/README.md
andrepimenta: Thanks @Gudahtt, I've improved the readme!

packages/storage-service/README.md
andrepimenta: Thanks @Gudahtt, I've improved the readme!

packages/storage-service/README.md
andrepimenta: Thanks @Gudahtt, I've improved the readme! Let me know what you think so we can resolve the conversation

packages/storage-service/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@metamask/storage-service",
+  "version": "1.0.0",
```
Gudahtt: Our release automation assumes that unpublished packages start at version 0, so we'll need to leave this at zero until the first publish.

It'll get bump as part of the publish, we can choose to make it 1.0.0 at that point

```suggestion
  "version": "0.0.0",
```

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,217 @@
...
+   */
+  async setItem(namespace: string, key: string, value: unknown): Promise<void> {
+    // Adapter handles serialization and wrapping with metadata
+    await this.#storage.setItem(namespace, key, value);
+
+    // Publish event so other controllers can react to changes
+    // Event type: StorageService:itemSet:namespace
+    // Payload: [key, value]
+    this.#messenger.publish(
+      `${SERVICE_NAME}:itemSet:${namespace}` as const,
```
Gudahtt, line 172: Nit: It looks like we are missing tests for this event. Maybe we can add this to the `setItem` block of tests?

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,217 @@
...
+   * Returns `unknown` since there's no schema validation.
+   * Callers should validate or cast the result to the expected type.
+   *
+   * @param namespace - Controller namespace (e.g., 'SnapController').
+   * @param key - Storage key (e.g., 'npm:@metamask/example-snap:sourceCode').
+   * @returns Parsed data or null if not found. Type is `unknown` - caller must validate.
+   */
+  async getItem(namespace: string, key: string): Promise<unknown> {
+    // Adapter handles deserialization and unwrapping
+    return await this.#storage.getItem(namespace, key);
```
Gudahtt, line 187: If we are to use the `itemSet` event to notify subscribers of removals, we'll need to publish here as well

packages/storage-service/README.md
Gudahtt: Looks great!

packages/storage-service/src/StorageService.ts
andrepimenta: Not sure I follow this one, can you explain in more detail?

Let's say the controller does:
```
const largeData = [1, 2 , 3 ,... , N]

await this.messenger.call(
  'StorageService:setItem',
  'MyController',
  'data-key',
  largeData,
);
```

Then it can do:
```
const largeData = [1, 2 , 3 ,... , N - 1]

await this.messenger.call(
  'StorageService:setItem',
  'MyController',
  'data-key',
  largeData,
);
```

So `largeData` is a list with one less value (or one value removed) in it. But the `setItem` doesn't care about what changes happened to the data itself, it just cares about receiving some data to persist. Meaning, the `setItem` event just shows that there was data with a certain key, persisted to disk.

The `getItem` reads the data from disk with a certain key, normally this data was already persisted to disk via `setItem`, so there was already an event for it. So there is no event that may be useful that comes from `getItem`? Maybe I'm missing something?

Gudahtt: .... Oh my, I posted this comment on the completely wrong method. I meant this for `removeItem`, sorry about that

Gudahtt: Essentially what i meant was that if you do something like this:

```
await this.messenger.call(
  'StorageService:removeItem',
  'MyController',
  'data-key',
);
```

Then we should emit `null` (or undefined?) for the event `StorageService:MyController:itemSet:data-key` so that subscribers have a way to discover it was deleted

packages/storage-service/src/types.ts
```diff
@@ -0,0 +1,167 @@
...
+export type StorageAdapter = {
+  /**
+   * Retrieve an item from storage.
+   * Adapter is responsible for building the full storage key.
+   *
+   * @param namespace - The controller namespace (e.g., 'SnapController').
+   * @param key - The data key (e.g., 'snap-id:sourceCode').
+   * @returns The value as a string, or null if not found.
+   */
+  getItem(namespace: string, key: string): Promise<unknown>;
```
Gudahtt: We may want to consider using `Json` for the value type everywhere instead of `unknown`, because it's not easy for us to store non-JSON data with extension and mobile storage APIs. We'd have to serialize it somehow, and if we just do a naive `JSON.stringify`, then it may throw an error or discard non-JSON data.

While we could implement serialization logic to handle this in all cases, I don't think we have any need for that. Simpler to always use JSON. At least until we find a reason to expand the supported data types.

packages/storage-service/src/StorageService.ts
andrepimenta: In my opinion, this event is not as useful. `removeItem` removes the entire data, with a certain key, from disk, so it's not really an update, as an update to the data is done with an `itemSet`.

Let's say a list in disk is `[1, 2, 3, 4]`. And now the list got updated and it's empty `[]`.

The controller would still use `itemSet` action with data as `[]`.

The only reason I would see a controller doing `removeItem` is to completely delete data when it is not used anymore (so, no need to keep it in disk). And if this happens, other controllers were already refactored not to use it. Makes sense? `itemRemoved` is like instantiating a controller with a different state/stateMetadata, I don't think it's a "subscribeable" event if it makes sense?

Gudahtt: Ah OK, sorry I had misunderstood this detail in our earlier discussion about the `itemRemoved` event. I had thought you meant `itemSet` to replace it, but no, rather it's that there is no need for this event at all.

packages/storage-service/src/types.ts
andrepimenta: I thought about that. But I think if it's a json, in the future we might loose on optimizations if we change the persistence system to use a proper DB that handles more types than a string or a json.

What I am doing in mobile is basically this:

```
setItem(data){
  
  const dataToStore = JSON.stringify({
    data,
    ...otherMetadata
  })
  
  Filesystem.setItem(dataToStore)
}
```

The only thing this wouldn't handle is functions and things like that? Do you think there is a chance that devs will try to save that type of data?

packages/storage-service/src/StorageService.ts
andrepimenta: Yes exactly. What I meant is if there is one less element in a certain piece of data, it is still updated/stored to disk via itemSet which triggers the event. Makes sense?

Gudahtt: It still seems like an event for removal would be useful, but that's fine, we can add it later easily if we find any need for it. Proceeding without it sounds good to me!

~I have one more minor suggestion to make related to this that I will make in a separate comment~

Nevermind, I was thinking we should automatically remove subscribers as well, but that does not apply here (the event type is per namespace not key).

packages/storage-service/src/types.ts
andrepimenta: You are right, using json for now is much easier. Once/if we support more types later, we can expand this then.

packages/storage-service/src/StorageService.ts
andrepimenta: True, just added some tests to it, let me know if it's ok and we can resolve the conversation

packages/storage-service/src/types.ts
Gudahtt: Yes, we've seen accidental and intentional use of non-JSON data many times. Mostly with `undefined`, but we also had a crash just last week from trying to serialize a click event (by accident). I can easily see functions, DOM references, etc. end up embedded in the data accidentally, and intentional use of `undefined` values as well.

We started using types to more strictly require only JSON data because it met all use-cases we had, and it avoided a lot of different types of bugs. Most concerning was that different platforms handled non-JSON data in different ways (e.g. Chrome normalizes `undefined` values to `null`, but Firefox excludes them, which led to bugs that only appeared on one browser or the other and thus evaded testing).

Good point about this preventing certain optimizations though 🤔

andrepimenta: Sorry I have removed the comment I made before, after thinking about it a bit more, it's better to json for now. If we ever do optimizations later, we can expand the types we support then.

andrepimenta: Ok @Gudahtt I believe this is now complete as well

andrepimenta:
@metamaskbot publish-preview

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Retrieve an item from in-memory storage.
+   * Deserializes JSON data from storage.
+   *
+   * @param namespace - The controller namespace.
+   * @param key - The data key.
+   * @returns The parsed JSON data, or null if not found.
+   */
+  async getItem(namespace: string, key: string): Promise<Json | null> {
```
Gudahtt: Nit: Redundant type union, `null` is already part of `Json`

```suggestion
  async getItem(namespace: string, key: string): Promise<Json> {
```

packages/storage-service/src/StorageService.ts
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  /**
+   * Retrieve JSON data from storage.
+   *
+   * @param namespace - Controller namespace (e.g., 'SnapController').
+   * @param key - Storage key (e.g., 'npm:@metamask/example-snap:sourceCode').
+   * @returns Parsed JSON data or null if not found.
+   */
+  async getItem(namespace: string, key: string): Promise<Json | null> {
```
Gudahtt: Nit: Redundant type union, `Json` includes `null`

```suggestion
  async getItem(namespace: string, key: string): Promise<Json> {
```

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    const fullKey = `${STORAGE_KEY_PREFIX}${namespace}:${key}`;
+    const serialized = this.#storage.get(fullKey);
+
+    if (!serialized) {
+      return null;
```
Gudahtt: Hmm. This could be ambiguous for the caller, there would be no way to differentiate data stored as `null` versus non-existent data.

Maybe we could return the data _wrapped_ in an object, e.g. `{ data: ___ }`, so that the `null` return can more clearly mean that it's not present?

Or we could use `undefined` as the empty return value, but that seems less than idea given that we'll be calling this across process boundaries on extension, and that would end up coerced to `null`.

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    try {
+      return JSON.parse(serialized) as Json;
```
Gudahtt: Unnecessary type cast:

```suggestion
      return JSON.parse(serialized);
```

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    } catch (error) {
+      // istanbul ignore next - defensive error handling for corrupted data
+      console.error(`Failed to parse stored data for ${fullKey}:`, error);
```
Gudahtt, line 64: This seems a bit dangerous. The caller would have no way to differentiate empty data from data that we failed to retrieve. The caller might then proceed when it's unsafe to do so, or overwrite the data, or something like that.

Given that we don't know precisely why the data is being stored/retrieved, or what the consequences of this failure might be, it would be safer to _not_ catch the error and let the caller deal with it. We can highlight that it _can_ throw with a TSDoc `@throws` directive in the doc comment for `getItem` (on both the service and the storage adapter) so that it's more obvious to the caller that they should expect failure some of the time.

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    const serialized = this.#storage.get(fullKey);
+
+    if (!serialized) {
+      return null;
+    }
+
+    try {
+      return JSON.parse(serialized) as Json;
+    } catch (error) {
+      // istanbul ignore next - defensive error handling for corrupted data
```
Gudahtt: Nit: This is a legitimate test case, we should be able to test this (even if we do remove the test block). Storage corruption can happen.

packages/storage-service/src/InMemoryStorageAdapter.ts
Gudahtt: Or I guess we could go Result type with this, e.g. `{ result }` or `{ error }` or `{}` (empty), to capture the error case as well. If you wanted this to "not crash by default", but in a way that still allows callers to handle the error.

packages/storage-service/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
```
Gudahtt: The changelog check is failing because it's expecting the PR number to be referenced in the entry:

```suggestion
- Initial release of `@metamask/storage-service` ([#7192](https://github.com/MetaMask/core/pull/7192))
```

Gudahtt: We could omit this and use the `no-changelog` label instead, since _anything_ before initial release can be assumed as part of this entry 🤷 

Whatever you prefer is fine; either way would get CI passing

packages/storage-service/src/InMemoryStorageAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (serialized === undefined) {
+      return {};
+    }
+
+    try {
+      const result = JSON.parse(serialized);
+      return { result };
+    } catch (error) {
+      console.error(`Failed to parse stored data for ${fullKey}:`, error);
+      return { error: error as Error };
```
Gudahtt, line 65: Normally I don't like to cast if we can avoid it, and we _can_ avoid it here by validating the error type.

But... it _is_ pretty easy to tell in this case that the only possible error here is `SyntaxError`. So.... this seems pretty unobjectionable.

I might propose an improvement here later, but I can do it as a separate PR

Gudahtt (approved):
Looks great!

Gudahtt (dismissed):

PR #7195, opened 2025-11-19
chore: Add Expiry Extraction Utilities to gator-permissions-controller

hanzel98:
## Explanation

This PR export the delegation framework version and skips the rules from the sanitization of the permissions response. The rules include the expiry.

### Solution
This PR introduces:

1. **`Skip rules from sanitization`**
   - The rules is part of the permission stored in the profile sync it includes the expiration date

3. **`DELEGATION_FRAMEWORK_VERSION` constant**
   - Moved to new `constants.ts` file to prevent circular dependencies
   - Ensures version consistency across implementations

## References

- Related to MetaMask/metamask-extension#37874 (where this functionality will be consumed)

## Checklist

- [ ] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes (not applicable - no breaking changes)

---

> [!NOTE]
> Keeps `rules` in sanitized permission responses, introduces strict TimestampEnforcer expiry parsing/validation, and exports `DELEGATION_FRAMEWORK_VERSION` from a new `constants.ts`.
> 
> - **Decode/validation (BREAKING)**
>   - Add `extractExpiryFromCaveatTerms` and enforce strict `TimestampEnforcer` terms: exact length, `timestampAfterThreshold` must be 0, `timestampBeforeThreshold` > 0, and safe-integer checks; update error messages accordingly.
> - **Controller & Types**
>   - Sanitization now only removes `dependencyInfo` and `signer`; preserves `permissionResponse.rules`.
>   - `PermissionResponseSanitized` updated to retain `rules`.
> - **Exports & Constants**
>   - Move `DELEGATION_FRAMEWORK_VERSION` to new `constants.ts` and export via package index; update imports.
> - **Tests**
>   - Update tests to verify preserved `rules` and new expiry validation scenarios.
> - **Changelog**
>   - Document breaking validation changes and new exports.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9313be30d1ff92792d611611cea4b6f9df8d688d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -223,6 +259,56 @@ export const getPermissionDataAndExpiry = ({
...
+      (caveat) => getChecksumAddress(caveat.enforcer) === timestampEnforcer,
+    );
+
+    if (!timestampCaveat) {
+      return null;
+    }
+
+    return extractExpiryFromCaveatTerms(timestampCaveat.terms);
+  } catch {
+    return null;
```
mj-kiwi: Is there a reason we're silencing the error here? It might be better to log the error before return null;

mj-kiwi (approved):
LGTM

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
hanzel98: Added the log

MoMannn (changes requested):
I think that instead of a util function the expiry should be decoded and saved under Rules in the permission. So that anyone fetching the permissions would already get decoded expiry and does not need to decode every time.

packages/gator-permissions-controller/src/decodePermission/extractExpiry.test.ts
MoMannn, line 1: For consistency I would move this tests to decodePermission.test.ts since the method is in that coresponsing file.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -77,6 +77,37 @@ export const identifyPermissionByEnforcers = ({
...
+ * Based on the TimestampEnforcer contract encoding:
+ * - Terms are 32 bytes total (64 hex characters without '0x')
+ * - First 16 bytes (32 hex chars): timestampAfterThreshold (uint128) - must be 0
+ * - Last 16 bytes (32 hex chars): timestampBeforeThreshold (uint128) - this is the expiry
+ *
+ * @param terms - The hex-encoded terms from a TimestampEnforcer caveat
+ * @returns The expiry timestamp in seconds, or null if no valid expiry exists
+ * @throws If the terms are not exactly 32 bytes or if the timestampAfterThreshold is non-zero
+ */
+export const extractExpiryFromCaveatTerms = (terms: Hex): number | null => {
```
MoMannn: This probably does not need to be exported now? Or if we still want it to be usable as a package then it would also need to be exported in index but I don't think we need that now.

packages/gator-permissions-controller/src/decodePermission/extractExpiry.test.ts
hanzel98: sure

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
hanzel98: yes

packages/gator-permissions-controller/src/GatorPermissionsController.test.ts
mj-kiwi: The test expects `rules` to be undefined, but the PR's purpose is to include rules in sanitized responses. Should we update this test to meet the purpose of this PR?

hanzel98: sure

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -451,7 +446,8 @@ export default class GatorPermissionsController extends BaseController<
   }
 
   /**
-   * Sanitizes a stored gator permission by removing the fields that are not expose to MetaMask client.
+   * Sanitizes a stored gator permission for client exposure.
```
jeffsmale90, line 479: Not this change, but "Sanitize" seems like a strange choice of word here... is it really just creating a view model?

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -77,6 +77,37 @@ export const identifyPermissionByEnforcers = ({
...
+
+  const [after, before] = splitHex(terms, [16, 16]);
+
+  if (hexToNumber(after) !== 0) {
+    throw new Error('Invalid expiry: timestampAfterThreshold must be 0');
+  }
+
+  const expiry = hexToNumber(before);
+
+  return expiry === 0 ? null : expiry;
```
jeffsmale90: if the expiry is 0 should we consider the request invalid?

I'm not sure if it's possible to specify expiry of zero within the picker, but the permission decoding logic tends to be very strict, because part of it's responsibility is to ensure that the request is a valid request from the gator permissions system.

packages/gator-permissions-controller/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `PermissionResponseSanitized` now includes `rules` property for stronger typing support ([#7195](https://github.com/MetaMask/core/pull/7195))
+- Export `DELEGATION_FRAMEWORK_VERSION` constant for version consistency across implementations ([#7195](https://github.com/MetaMask/core/pull/7195))
```
jeffsmale90, line 21: Is it worth including "More meaningful validation errors returned when expiry information is invalid" or something similar under `### Changed`?

jeffsmale90 (approved):
Looks good!

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
hanzel98: reverting on 0 now

packages/gator-permissions-controller/CHANGELOG.md
hanzel98: I added it

packages/gator-permissions-controller/CHANGELOG.md
```diff
@@ -7,10 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING:** Permission decoding now rejects `TimestampEnforcer` caveats with zero `timestampBeforeThreshold` values
+- Validation errors for `TimestampEnforcer` include: invalid terms length, non-zero `timestampAfterThreshold`, and zero `timestampBeforeThreshold`
```
jeffsmale90: Can you add links to this PR for these changes?

hanzel98: sure

packages/gator-permissions-controller/CHANGELOG.md
jeffsmale90: I note that this was erroneously added to the previous 0.7.0 release notes - will be fixed in 0.8.0

PR #7198, opened 2025-11-20
add tron filtering options to token-selectors

Prithpal-Sooriya:
## Explanation

Currently we have fragmented logic in mobile and extension selectors due to TRON staked assets (energy and bandwidth).
This PR unifies this logic via an optional selector option to filter out TRON staked assets.

Default behaviour is to filter TRON staked assets.

## Checklist

---

> [!NOTE]
> Adds an optional (default on) filter to exclude Tron staking resources in `selectAssetsBySelectedAccountGroup`, with tests and changelog update.
> 
> - **Selectors**:
>   - Add Tron resource constants and filtering (`TRON_RESOURCE`, `TRON_RESOURCE_SYMBOLS_SET`).
>   - Extend `selectAssetsBySelectedAccountGroup` to accept options `{ filterTronStakedTokens }` (defaults to `true`) and filter Tron staking tokens across `TrxScope` networks.
>   - Enable memoization for selector args/results via `weakMapMemoize`.
> - **Tests**:
>   - Add Tron fixtures in `src/selectors/__fixtures__/arrange-tron-state.ts`.
>   - Add unit tests validating filtering on/off in `token-selectors.test.ts`.
> - **Changelog**:
>   - Document new optional Tron filtering for `selectAssetsBySelectedAccountGroup`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e6a0939baad479a1368aba7e32636e1375d1915d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/selectors/token-selectors.ts
+++ b/packages/assets-controllers/src/selectors/token-selectors.ts
@@ -463,6 +463,10 @@ export type SelectAccountGroupAssetOpts = {
   filterTronStakedTokens: boolean;
 };
 
+const defaultSelectAccountGroupAssetOpts: SelectAccountGroupAssetOpts = {
+  filterTronStakedTokens: true,
+};
+
 const filterTronStakedTokens = (assetsByAccountGroup: AccountGroupAssets) => {
   const newAssetsByAccountGroup = { ...assetsByAccountGroup };
 
@@ -495,7 +499,7 @@ export const selectAssetsBySelectedAccountGroup = createAssetListSelector(
     (state) => state.accountTree,
     (
       _state,
-      opts: SelectAccountGroupAssetOpts = { filterTronStakedTokens: true },
+      opts: SelectAccountGroupAssetOpts = defaultSelectAccountGroupAssetOpts,
     ) => opts,
   ],
   (groupAssets, accountTree, opts) => {
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added optional filtering of Tron staking tokens (energy, bandwidth) in asset selectors `selectAssetsBySelectedAccountGroup` (defaults to `true`) ([#7198](https://github.com/MetaMask/core/pull/7198))
+
 ## [89.0.1]
 
 ### Fixed


packages/assets-controllers/src/selectors/token-selectors.test.ts
```diff
@@ -918,5 +921,108 @@ describe('token-selectors', () => {
 
       expect(result).toStrictEqual(expectedMockResult);
     });
+
+    const arrangeTronState = () => {
```
Prithpal-Sooriya, line 925: I really hate this test suite...

Like sure this uses realistic state data, but MAN its ridiculously difficult to add new tests since you need to understand the WHOLE STATE.

It would be much much nicer if we can:
1. Create JSON objects for each state we want to test. We have a fixtures folder that we can shove this in.
2. Break up the selector parts into modular atomic functions that we can mock and test in isolation.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -438,14 +459,66 @@ const selectAllAssets = createAssetListSelector(
...
+
 export const selectAssetsBySelectedAccountGroup = createAssetListSelector(
-  [selectAllAssets, (state) => state.accountTree],
-  (groupAssets, accountTree) => {
+  [
+    selectAllAssets,
+    (state) => state.accountTree,
+    (
+      _state,
+      opts: SelectAccountGroupAssetOpts = defaultSelectAccountGroupAssetOpts,
```
bergarces, line 502: Ok for now, but I'm not a big fan of this as it can force recalculations based on filters. Also, if we add more chains that have different needs we are going to end up with very messy logic here.

IMO, the problem with Tron should have been solved by expanding the asset type to contain underlying metada (the tron resources), so all these Tron resources would have been added inside the main Tron asset, which will always be present. By adding those "resources" to the list of assets, even though we don't want to display them most of the time, we are causing situations like this one.

Prithpal-Sooriya: Yep I 100% agree with this.

R.E. selector recalculations - yep noticed this too. Converted to a weakmap memo on arguments to reduce this. Likewise I decoupled the filtering logic, but I can imagine this selector getting real messy down the line.

TRON tokens are a bit of a mess, I agree it would be good if either:
1. There was a separation between normal tokens and staking tokens
2. Tron team does not expose staking tokens immediately (most flows should not show these tokens), only on some rare screens should they be visible (which can be a separate call not going through our controllers)

PR #7244, opened 2025-11-26
feat(gator-permissions-controller): add submitDirectRevocation for already-disabled delegations

hanzel98:
## Explanation

This PR adds support for revoking already-disabled delegations that don't require an on-chain transaction. Previously, the revocation flow required creating a transaction and waiting for it to confirm, but for delegations that are already disabled on-chain, we can submit the revocation directly to the snap without waiting for a transaction.

### What this PR does:

1. **Adds submitDirectRevocation method**: This new method combines adding a pending revocation to state (to disable the UI button) and immediately submitting the revocation to the snap. This simplifies the caller code in metamask-extension since they no longer need to call two separate methods.

2. **Adds isPendingRevocation helper method**: A utility method to check if a permission context is currently in the pending revocations list. This can be used by the UI to determine button states.

3. **Improves error handling in submitRevocation**: Refactored to use a finally block to ensure pending revocations are always cleared when the revocation succeeds, even if the refresh operation fails. This prevents stuck Revocation Pending states in the UI.

### Key implementation details:

- submitDirectRevocation uses a placeholder txId format (no-tx-${permissionContext}) that won't conflict with real transaction IDs
- The pending revocation is removed by permissionContext (not txId), so the placeholder doesn't affect functionality
- If submitRevocation fails, the pending revocation stays in state so the UI can show an error/retry state
- The finally block ensures cleanup happens even if the refresh fails after successful revocation
- All new functionality is fully tested with comprehensive test coverage

### Usage:

Before (required two calls):
await gatorPermissionsController.addDirectPendingRevocation({ permissionContext });
await gatorPermissionsController.submitRevocation({ permissionContext });

After (single call):
await gatorPermissionsController.submitDirectRevocation({ permissionContext });

### Testing:

- All existing tests pass (45 tests)
- Added 4 new tests for submitDirectRevocation
- Added 3 new tests for isPendingRevocation
- Added 1 test for finally block behavior in submitRevocation
- All tests passing ✅

## Related

This change is needed to support revoking already-disabled delegations in the MetaMask extension UI.

---

> [!NOTE]
> Adds `submitDirectRevocation` and `isPendingRevocation`, and improves revocation cleanup and error handling.
> 
> - **Controller (`src/GatorPermissionsController.ts`)**:
>   - **New APIs**: `submitDirectRevocation(params)` for already-disabled delegations; `isPendingRevocation(permissionContext)` helper.
>   - **Actions/Handlers**: Register `GatorPermissionsController:submitDirectRevocation` and `:isPendingRevocation`.
>   - **Revocation flow**: `submitRevocation` now refreshes permissions after submit, clears pending in a `finally` block, and wraps refresh failures with `GatorPermissionsFetchError` (specific message).
>   - **Pending state**: Removal by `permissionContext` is now case-insensitive.
> - **Exports (`src/index.ts`)**: Export new action types for direct revocation and pending check.
> - **Tests**: Add tests for `submitDirectRevocation`, `isPendingRevocation`, and `submitRevocation` cleanup-on-refresh-failure.
> - **Changelog**: Note addition of `submitDirectRevocation` in Unreleased.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ea1f04efc7c6b4ed9aea4da173758e81abd9acd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/gator-permissions-controller/src/GatorPermissionsController.ts
+++ b/packages/gator-permissions-controller/src/GatorPermissionsController.ts
@@ -372,7 +372,8 @@ export default class GatorPermissionsController extends BaseController<
     this.update((state) => {
       state.pendingRevocations = state.pendingRevocations.filter(
         (pendingRevocations) =>
-          pendingRevocations.permissionContext !== permissionContext,
+          pendingRevocations.permissionContext.toLowerCase() !==
+          permissionContext.toLowerCase(),
       );
     });
   }


jeffsmale90 (approved):
looks good to me!

PR #7266, opened 2025-11-28
fix(multichain-account-service)!: wait for Snap platform to be ready

ccharly:
## Explanation

Snap requests are now delayed after the onboarding, to avoid starting requests "too soon" on the clients, we now have a logic to wait for the Snap platform to be ready.

This makes sure that any calls that uses retries or timeouts will start their calls AFTER the Snap platform is ready to process anything.

We stumbled upon this bug during the concurrent requests/Snap state inconsistency investigation. Initial account providers requests were usually timing out because they were hitting their limit because they were starting a bit "late" since the Snap platform was not ready yet. (This was also still causing concurrent requests, despite the `maxConcurrency: 1`, since Snaps were still processing the timed out request while receiving new ones).

To future-proof this and fully await for the Snap platform to be ready, we now send 1 simple Snap request and wait for it to be executed (being successful or not). Then, we start processing all Snap requests with our usual logic (timeouts, semaphore, retries, etc...)

## References

- https://github.com/MetaMask/metamask-mobile/issues/23807

## Checklist

---

> [!NOTE]
> Ensures Snap-based flows wait for platform readiness and refactors providers to use a guarded Snap access pattern.
> 
> - **Add** `SnapPlatformWatcher` to track Snap platform readiness and gate operations; expose `MultichainAccountService:ensureCanUseSnapPlatform` action/method and register its handler
> - **Refactor (BREAKING)**: make `SnapAccountProvider.client` private and introduce `SnapAccountProvider.withSnap` to access `{ client, keyring }`; update `Sol/Btc/TrxAccountProvider` to use `withSnap`, move creation helpers private, and use restricted keyring for `createAccount`/`removeAccount`
> - **Service updates**: instantiate watcher in `MultichainAccountService` and route readiness checks; wire new action; minor resync adjustments to use guarded keyring ops
> - **Tests**: add comprehensive unit tests for watcher and provider readiness gating; adapt existing provider tests to new helpers/mocks
> - **Misc**: update `CHANGELOG.md`; add `lodash` dependency
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3c898b2cc58e5ea54bff0ca1b82af8003b6de79d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/multichain-account-service/src/snaps/SnapPlatformWatcher.ts
+++ b/packages/multichain-account-service/src/snaps/SnapPlatformWatcher.ts
@@ -37,22 +37,25 @@ export class SnapPlatformWatcher {
   #watch(): void {
     const logReadyOnce = once(() => log('Snap platform is ready!'));
 
-    const state = this.#messenger.call('SnapController:getState');
-
     // If already ready, resolve immediately.
-    if (state.isReady) {
+    const initialState = this.#messenger.call('SnapController:getState');
+    if (initialState.isReady) {
       this.#isReady = true;
       this.#isReadyOnce.resolve();
     }
 
     // We still subscribe to state changes to keep track of the platform's readiness.
-    this.#messenger.subscribe('SnapController:stateChange', ({ isReady }) => {
-      this.#isReady = isReady;
+    this.#messenger.subscribe(
+      'SnapController:stateChange',
+      (isReady: boolean) => {
+        this.#isReady = isReady;
 
-      if (isReady) {
-        logReadyOnce();
-        this.#isReadyOnce.resolve();
-      }
-    });
+        if (isReady) {
+          logReadyOnce();
+          this.#isReadyOnce.resolve();
+        }
+      },
+      (state) => state.isReady,
+    );
   }
 }


packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -42,6 +48,9 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
 
   readonly #trace: TraceCallback;
 
+  protected static ensureSnapPlatformIsReadyPromise: Promise<void> | null =
```
ccharly: Mainly using a `protected static` here so we can write tests and update this without exposing it completely publicly (consumers could still hack this easily though..)

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -222,14 +286,32 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
...
+  }): Promise<Bip44Account<KeyringAccount>[]> {
+    await this.ensureSnapPlatformIsReady();
+
+    return await this.runDiscoverAccounts(options);
+  }
+
   abstract isAccountCompatible(account: Bip44Account<InternalAccount>): boolean;
 
-  abstract createAccounts(options: {
+  abstract runCreateAccounts(options: {
```
ccharly: Now providers have to implements those `run*` methods, so we can automatically call `this.ensureSnapPlatformIsReady` for any of those actions.

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: Question: Should we have a 1 promise per Snap? This would allow to wait for the Snap to be ready rather than the Snap platform itself?

hmalik88: Wouldn't the snap be ready if the platform is?

ccharly: I guess so yes, so I think we can keep this as-is!

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: Actually, I reverted this change in favor of a new pattern (`withSnap`)

packages/multichain-account-service/src/snaps/SnapPlatformWatcher.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    // We still subscribe to state changes to keep track of the platform's readiness.
+    this.#messenger.subscribe('SnapController:stateChange', ({ isReady }) => {
+      this.#isReady = isReady;
+
+      if (isReady) {
+        logReadyOnce();
+        this.#isReadyOnce.resolve();
+      }
+    });
```
ccharly: TODO: Actually, I missed something here, we want to have a "selector" to avoid triggering this handler for other unrelated state changes!

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
ccharly: Resolving since we now use a completely different logic

packages/multichain-account-service/src/snaps/SnapPlatformWatcher.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // We still subscribe to state changes to keep track of the platform's readiness.
+    this.#messenger.subscribe(
+      'SnapController:stateChange',
+      (isReady: boolean) => {
+        this.#isReady = isReady;
+
+        if (isReady) {
+          logReadyOnce();
+          this.#isReadyOnce.resolve();
```
ccharly, line 55: For reviewers: One could argue that this could "block" the entire service until the Snap platform becomes ready. However, we already have some mechanism that prevent calling any Snap requests if Basic Functionality is OFF, so this should, never really happen.

I just wonder if we could expose a new `:isReady` action on the service itself to allow any caller to check for this before interacting with the service 🤔

But this can be done later.

hmalik88 (approved):

PR #7277, opened 2025-12-01
feat: Implement version-gated feature flag

### LINKED ISSUE #7285: Add version gated feature flag option into `RemoteFeatureFlagController`
We need version-gated feature flag filtering in order to handle LD flags appropriately. 

The idea is to allow feature flags to define a minimum supported app version before they are activated. For example, a flag could include a fromVersion field to indicate when it becomes effective:

```
{
  "featureA": {
    "fromVersion": "13.10.0",
    "flagX": "X"
  }
}
```

As we need this feature in the both clients and we are using different LD projects for each, we want to add this feature into `RemoteFeatureFlagController` so both clients can benefit from it.

OGPoyraz:
## Explanation

This PR implements version-gated feature flags that allow feature flags to specify minimum app version requirements and support multiple version-specific values. This enables safer progressive feature rollouts and ensures features only activate on compatible app versions.

Feature flags can now include multiple version entries, each with its own value and version requirement:

**Important note**: All versions we pass into `versions` must be valid 3 part semver string. (E.g `13.5.2`)

```
// New supported format - Launch Darkly flag
{
    "versions": {
      "13.0.0": { "ui": "legacy" },
      "13.1.0": { "ui": "improved" },
      "13.2.0": { "ui": "modern", "animations": true }
    }
}

// Version selection logic:
// Client v13.0.5 → Gets "legacy" UI (only 13.0.0 qualifies)
// Client v13.1.8 → Gets "improved" UI (13.0.0 and 13.1.0 qualify, highest selected) 
// Client v13.2.1 → Gets "modern" UI (all qualify, 13.2.0 is highest)
// Client v12.9.0 → Feature excluded (no versions qualify - flag will not be added to state)
```

## References

Fixes: https://github.com/MetaMask/core/issues/7285

## Checklist

---

> [!NOTE]
> Introduce multi-version feature flag processing with SemVer-based selection and make `clientVersion` a required constructor param.
> 
> - **Feature flags (core behavior)**:
>   - Add support for multi-version flags: `{ versions: { "13.1.0": ..., "13.2.0": ... } }`.
>   - Select highest qualifying version via SemVer comparison; exclude flag if none qualify.
>   - Integrate version-selected values with threshold-based A/B grouping.
> - **Controller (`src/remote-feature-flag-controller.ts`)**:
>   - Require and validate `clientVersion` (3-part SemVer) in constructor. [BREAKING]
>   - Process flags via `#processVersionBasedFlag` and merge with existing scoped selection logic.
> - **Utilities (`src/utils/version.ts`)**:
>   - New helpers: `isVersionFeatureFlag()`, `getVersionData()`, `isVersionAtLeast()` and `MULTI_VERSION_FLAG_KEYS`.
> - **Types (`remote-feature-flag-controller-types.ts`)**:
>   - Add `MultiVersionFeatureFlagValue` and use `SemVerVersion` for versioned entries.
> - **Tests**:
>   - Add comprehensive tests for versioned flag detection/selection and controller integration.
> - **Docs**:
>   - Update `CHANGELOG.md` with new feature and breaking change details.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9ab90dfb126b631fffb029bb8e819a0d8791baed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +231,19 @@ export class RemoteFeatureFlagController extends BaseController<
...
+          this.#appVersion,
+        );
+
+        if (!selectedVersion) {
+          continue;
+        }
+
+        processedValue = selectedVersion.value;
+      }
+
```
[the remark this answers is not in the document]
OGPoyraz: This could be handled on different PR as the aim is just implementing version gated filtering in this PR.

packages/remote-feature-flag-controller/src/utils/version-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @param entry - The object to check
+ * @returns true if the entry is a valid VersionEntry, false otherwise
+ */
+function isValidVersionEntry(entry: unknown): entry is VersionEntry {
+  return (
+    typeof entry === 'object' &&
+    entry !== null &&
+    !Array.isArray(entry) &&
+    'fromVersion' in entry &&
```
matthewwalsh0: Should we put all these reserved keys in constants for visibility?

packages/remote-feature-flag-controller/src/utils/version-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  return version.split('.').map(Number);
+}
+
+/**
+ * Checks if an object is a valid VersionEntry.
+ *
+ * @param entry - The object to check
+ * @returns true if the entry is a valid VersionEntry, false otherwise
+ */
+function isValidVersionEntry(entry: unknown): entry is VersionEntry {
```
matthewwalsh0: Minor, `isVersionEntry`?

packages/remote-feature-flag-controller/src/utils/version.ts
matthewwalsh0, line 1: Minor, `/version.ts`?

packages/remote-feature-flag-controller/src/utils/version-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Selects the appropriate version value from a multi-version feature flag based on current app version.
+ * Returns the value from the highest version that the app version meets or exceeds.
+ *
+ * @param multiVersionFlag - The multi-version feature flag
+ * @param currentAppVersion - The current application version
+ * @returns The selected version entry, or null if no version requirements are met
+ */
+export function selectVersionFromMultiVersionFlag(
```
matthewwalsh0: Minor, `getVersionData` to avoid confusing with selector?

packages/remote-feature-flag-controller/src/utils/version-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  const eligibleVersions = multiVersionFlag.versions.filter((versionEntry) =>
+    isVersionAtLeast(currentAppVersion, versionEntry.fromVersion),
+  );
+
+  if (eligibleVersions.length === 0) {
+    return null;
+  }
+
+  // Compare versions - we want the highest version that still qualifies hence the first sort is descending
+  return eligibleVersions.sort((a, b) => {
```
matthewwalsh0: Minor, would it be conceptually simpler to sort the entries first, then return the first match?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -129,13 +136,15 @@ export class RemoteFeatureFlagController extends BaseController<
...
     getMetaMetricsId,
+    appVersion,
   }: {
     messenger: RemoteFeatureFlagControllerMessenger;
     state?: Partial<RemoteFeatureFlagControllerState>;
     clientConfigApiService: AbstractClientConfigApiService;
     getMetaMetricsId: () => string;
     fetchInterval?: number;
     disabled?: boolean;
+    appVersion: string;
```
matthewwalsh0: Minor, `clientVersion` to align with terminology?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +231,19 @@ export class RemoteFeatureFlagController extends BaseController<
...
+        const selectedVersion = selectVersionFromMultiVersionFlag(
+          remoteFeatureFlagValue,
+          this.#appVersion,
+        );
+
+        if (!selectedVersion) {
+          continue;
+        }
+
+        processedValue = selectedVersion.value;
```
matthewwalsh0: If we only need the `value`, should we update `selectVersionFromMultiVersionFlag` / `getVersionData` to just return the data?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
```diff
@@ -24,6 +24,17 @@ export enum EnvironmentType {
   Exp = 'exp',
 }
 
+/** Type representing a single version entry in a version-gated feature flag */
+export type VersionEntry = {
+  fromVersion: string;
+  value: Json;
```
matthewwalsh0: Minor, `data` rather than `value` given we know it's an object?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +231,19 @@ export class RemoteFeatureFlagController extends BaseController<
     ] of Object.entries(remoteFeatureFlags)) {
       let processedValue = remoteFeatureFlagValue;
 
+      if (isMultiVersionFeatureFlagValue(remoteFeatureFlagValue)) {
```
matthewwalsh0: Minor, is it worth putting this logic in a local private method to break up this large block?

packages/remote-feature-flag-controller/src/utils/version-utils.test.ts
```diff
@@ -0,0 +1,198 @@
+import {
+  isVersionAtLeast,
+  isMultiVersionFeatureFlagValue,
+  selectVersionFromMultiVersionFlag,
+} from './version-utils';
+
+describe('isVersionAtLeast', () => {
```
matthewwalsh0: Minor, should we be exporting this just to test it? Are we just considering it high risk?

packages/remote-feature-flag-controller/src/utils/version-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  VersionEntry,
+} from '../remote-feature-flag-controller-types';
+
+/**
+ * Checks if a feature flag value is a multi-version gated flag (contains versions array).
+ *
+ * @param value - The feature flag value to check
+ * @returns true if the value is a multi-version feature flag, false otherwise
+ */
+export function isMultiVersionFeatureFlagValue(
```
matthewwalsh0: Minor, `isVersionFeatureFlag`?

matthewwalsh0 (changes requested):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
```diff
@@ -24,6 +24,17 @@ export enum EnvironmentType {
...
+};
+
+/** Type representing a feature flag with multiple version entries */
+export type MultiVersionFeatureFlagValue = {
+  versions: VersionEntry[];
```
matthewwalsh0: Apologies, I suggested the array, but should we instead use an object keyed by version to reduce the syntax and prevent duplicate entries?

We're already sorting the entries so order shouldn't matter.

The only downside is future-proofing if we want additional config per version in future, but we could always support multiple version schemas if necessary.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +231,19 @@ export class RemoteFeatureFlagController extends BaseController<
...
+          this.#clientVersion,
+        );
+
+        if (!selectedValue) {
+          continue;
+        }
+
+        processedValue = selectedValue;
+      }
+
```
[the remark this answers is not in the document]
OGPoyraz: Replied above

packages/remote-feature-flag-controller/src/utils/version-utils.ts
OGPoyraz: Done - only moved `versions` as we don't have `fromVersion` or `data` now

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
OGPoyraz: `data` or `value` is not existing with new interface now - so it's fixed

packages/remote-feature-flag-controller/src/utils/version-utils.test.ts
OGPoyraz: Yes it's just testing purpose - I want to put exclusive tests as there are many cases

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
OGPoyraz: This totally make sense - it will prevent putting duplicated LD flags so kinda validates upfront. I converted PR to this approach now

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +243,14 @@ export class RemoteFeatureFlagController extends BaseController<
...
       let processedValue = remoteFeatureFlagValue;
 
+      const versionBasedValue = this.#processVersionBasedFlag(
+        remoteFeatureFlagValue,
+      );
+      if (versionBasedValue === null) {
+        continue;
+      }
+      processedValue = versionBasedValue;
+
```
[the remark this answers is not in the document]
OGPoyraz: This can be handled separately not in this PR.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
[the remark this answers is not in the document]
matthewwalsh0: Could these errors be fixed by just replacing any references to `remoteFeatureFlagValue` with `processedValue`?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -221,6 +243,14 @@ export class RemoteFeatureFlagController extends BaseController<
     ] of Object.entries(remoteFeatureFlags)) {
       let processedValue = remoteFeatureFlagValue;
 
+      const versionBasedValue = this.#processVersionBasedFlag(
+        remoteFeatureFlagValue,
+      );
+      if (versionBasedValue === null) {
+        continue;
+      }
+      processedValue = versionBasedValue;
```
matthewwalsh0: Minor, could we assign to this directly on line 246? Should be identical since the version based value returns the original value if not a version flag.

matthewwalsh0 (dismissed):

packages/remote-feature-flag-controller/src/utils/version.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * Compares two semantic version strings.
+ *
+ * @param currentVersion - The current version (e.g., "13.9.5")
+ * @param requiredVersion - The required minimum version (e.g., "13.10.0")
+ * @returns true if currentVersion >= requiredVersion, false otherwise
+ */
+export function isVersionAtLeast(
```
FrederikBolding: Can we use `semver` for this instead?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -151,6 +158,7 @@ export class RemoteFeatureFlagController extends BaseController<
     this.#disabled = disabled;
     this.#clientConfigApiService = clientConfigApiService;
     this.#getMetaMetricsId = getMetaMetricsId;
+    this.#clientVersion = clientVersion;
```
FrederikBolding: To whoever implements this in the clients, remember to remove the `-flask.0` parts of the version as it causes problems when comparing semantic versions

packages/remote-feature-flag-controller/src/utils/version.ts
FrederikBolding: We have some utils already in `metamask/utils`: https://github.com/MetaMask/utils/blob/main/src/versions.ts

packages/remote-feature-flag-controller/src/utils/version.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  });
+
+  it('returns false when current version is less than required version', () => {
+    expect(isVersionAtLeast('13.9.0', '13.10.0')).toBe(false);
+    expect(isVersionAtLeast('13.5.0', '13.5.2')).toBe(false);
+    expect(isVersionAtLeast('12.0.0', '13.0.0')).toBe(false);
+  });
+
+  it('returns true when versions are equal', () => {
+    expect(isVersionAtLeast('13.10.0', '13.10.0')).toBe(true);
```
dan437: nit: since both params represent a version, it would be little more readable to do this:
```isVersionAtLeast({ currentVersion: '13.10.0', requiredVersion: '13.10.0' })```

**Comment by dan437** [comment]
Thanks for adding this, it will be very valuable for us and other teams. One question: does this support gradual rollout (% of users who see a feature): https://github.com/MetaMask/contributor-docs/blob/main/docs/remote-feature-flags.md#choosing-the-right-feature-flag-type

File: packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts [hunk 19]
```diff
@@ -221,6 +249,14 @@ export class RemoteFeatureFlagController extends BaseController<
...
       let processedValue = remoteFeatureFlagValue;
 
+      const versionBasedValue = this.#processVersionBasedFlag(
+        remoteFeatureFlagValue,
+      );
+      if (versionBasedValue === null) {
+        continue;
+      }
+      processedValue = versionBasedValue;
+
```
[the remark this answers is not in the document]
OGPoyraz: Fixed

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -219,7 +247,12 @@ export class RemoteFeatureFlagController extends BaseController<
       remoteFeatureFlagName,
       remoteFeatureFlagValue,
     ] of Object.entries(remoteFeatureFlags)) {
-      let processedValue = remoteFeatureFlagValue;
+      let processedValue = this.#processVersionBasedFlag(
+        remoteFeatureFlagValue,
+      );
+      if (processedValue === null) {
+        continue;
+      }
```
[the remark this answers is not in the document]
OGPoyraz: Fixed

packages/remote-feature-flag-controller/src/utils/version.ts
OGPoyraz: Thanks for raising @FrederikBolding - majority of the version checks removed and re-used from utils now 
https://github.com/MetaMask/core/pull/7277/commits/ebd77d625931234836aa6379f7816db524c5fcbc

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
OGPoyraz: Also added a tiny validation in constructor for this to not throw type errors later
https://github.com/MetaMask/core/pull/7277/commits/ebd77d625931234836aa6379f7816db524c5fcbc

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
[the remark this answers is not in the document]
OGPoyraz: I think you're right @matthewwalsh0 
Now we combine both features, also added a test for this

https://github.com/MetaMask/core/pull/7277/commits/d4933d18ac1cdf642ad96e65a2418706c0bc780e

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
OGPoyraz: Addressed here https://github.com/MetaMask/core/pull/7277/commits/997582cac371022d05fc65384cd4f3bd12ac5f27

OGPoyraz:
Good call @dan437 - I recently added that support too, related commit is https://github.com/MetaMask/core/commit/d4933d18ac1cdf642ad96e65a2418706c0bc780e

packages/remote-feature-flag-controller/src/utils/version.test.ts
OGPoyraz: I removed all the tests for this function, it's not exported and much more tiny now. Hence I am skipping this nit

matthewwalsh0 (dismissed):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -74,6 +75,7 @@ function createController(
       options.clientConfigApiService ?? buildClientConfigApiService(),
     disabled: options.disabled,
     getMetaMetricsId: options.getMetaMetricsId ?? (() => MOCK_METRICS_ID),
+    clientVersion: options.clientVersion ?? '13.10.0',
```
DDDDDanica: Could you make `13.10.0` a const as MOCK_BASE_VERSION so it can be accessed easily in other tests

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
DDDDDanica: could we export a util function from controller and access from client side easily?

packages/remote-feature-flag-controller/src/utils/version.ts
```diff
@@ -0,0 +1,93 @@
...
+    value === null ||
+    Array.isArray(value) ||
+    !(MULTI_VERSION_FLAG_KEYS.VERSIONS in value)
+  ) {
+    return false;
+  }
+
+  const versions = (value as Record<string, unknown>)[
+    MULTI_VERSION_FLAG_KEYS.VERSIONS
+  ];
```
Gudahtt: This assertion doesn't appear to be necessary
```suggestion
  const versions = value[MULTI_VERSION_FLAG_KEYS.VERSIONS];
```

packages/remote-feature-flag-controller/src/utils/version.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param multiVersionFlag - The multi-version feature flag
+ * @param currentAppVersion - The current application version
+ * @returns The selected version value, or null if no version requirements are met
+ */
+export function getVersionData(
+  multiVersionFlag: MultiVersionFeatureFlagValue,
+  currentAppVersion: string,
+): Json | null {
+  const sortedVersions = Object.entries(multiVersionFlag.versions).sort(
```
Gudahtt: Nit: I saw the notice below in `isVersionAtLeast` about how the params were already validated as SemVer strings, and I tried to update it to carry the type through from when it was validated so that TypeScript could check this for us (so that we don't break it in the future), and the main challenge was this `Object.entries` call. Even if we change the index type to `SemVerVersion`, it gets widened to `string` here.

It ended up being pretty easy to write a utility function to call `Object.entries` while preserving the type though:

```
/**
 * A utility function that calls `Object.entries` while preserving the index type.
 *
 * @param input - The object to get the entries of.
 * @returns - The object entries.
 */
function getObjectEntries<Input extends Record<string, unknown>>(
  input: Input,
): [keyof Input, Input[keyof Input]][] {
  // Use cast to preserve index type. Object.entries always widens to string.
  return Object.entries(input) as [keyof Input, Input[keyof Input]][];
}
```

it still relies upon a type assertion, but at least it can be easily validated as correct just by understanding that one line. This seems easier to maintain than the assertion we use below, where the basis for the assertion is that the caller of the caller (in another file) had already validated the types.

Perhaps we could use this utility function, and update `MultiVersionFeatureFlagValue` to use `SemVerVersion` as the index type? Then we could use the type throughout this controller and let TypeScript do the work for us of ensuring it was validated.

Gudahtt: Also I ended up having to update the type of the `#clientVersion` property to `SemVerVersion` when I attempted this, but that was fine, that was already validated as well.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Gudahtt: The `isValidSemVerVersion` function is already easily accessible from clients, it's from `@metamask/utils`

Gudahtt (commented):
Looks great! Just a few suggestions about the types

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
FrederikBolding: To be clear. the Flask versions are still valid semantic versions, they are just considered pre-release so comparison is different

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
OGPoyraz: Adressed in https://github.com/MetaMask/core/pull/7277/commits/7b9b9c76b293ea7613d4a14a2173d1490605cf35

Gudahtt (dismissed):
LGTM!

packages/remote-feature-flag-controller/src/utils/version.ts
OGPoyraz: Thank you for the suggestion @Gudahtt !

I adjusted the type of versions to be `Record<SemVerVersion, Json>;`, fixed here https://github.com/MetaMask/core/pull/7277/commits/7b9b9c76b293ea7613d4a14a2173d1490605cf35

packages/remote-feature-flag-controller/src/utils/version.ts
OGPoyraz: Fixed in https://github.com/MetaMask/core/commit/7b9b9c76b293ea7613d4a14a2173d1490605cf35

Gudahtt (approved):

PR #7280, opened 2025-12-01
chore: Use stricter TypeScript configuration for linting

Gudahtt:
## Explanation

The `tsconfig.json` file has been updated to use stricter compiler options, matching what is used in other TSConfig files already. This was a blocker for using the rule
`@typescript-eslint/no-unnecessary-boolean-literal-compare` with newer versions of `typescript-eslint` (it requires `strictNullChecks` to work correctly).

To avoid duplicating this configuration, it was moved to a new file (`tsconfig.base.json`), and extended elsewhere. I've added a comment block to each root-level `tsconfig` file explaining its purpose as well, since it was getting difficult to keep track of them all.

## References

N/A

## Checklist

---

> [!NOTE]
> Introduce shared `tsconfig.base.json` and update repo configs and Jest mapping to extend/use it, consolidating strict TS settings.
> 
> - **TypeScript configuration**
>   - Add shared `tsconfig.base.json` with strict compiler options.
>   - Update root `tsconfig.json` to extend base and serve lint/editor use; remove duplicated options.
>   - Update `tsconfig.packages.json` to extend base; keep `paths` for `@metamask/*`.
>   - Update `tsconfig.packages.build.json` to extend packages config; no functional changes beyond inheritance.
>   - Add `tsconfig.build.json` to reference all package build configs (documentation clarified).
>   - Update `tsconfig.scripts.json` to extend base and drop redundant options.
> - **Tooling**
>   - Adjust Jest `moduleNameMapper` comment to reference `tsconfig.base.json`.
>   - Update `.gitattributes` to mark `tsconfig.**.json` as `jsonc` for linguist.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 22e5a48c54219138e13d7e8b7364fc269c22009b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,10 @@
 
 yarn.lock linguist-generated=false
 
+# `tsconfig.json` is already recognized as JSONC. This ensures our other `tsconfig` files are as
+# well. They all use this naming convention.
+tsconfig.**.json linguist-language=jsonc
+
 # yarn v3
 # See: https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
 /.yarn/releases/** binary
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,0 +1,14 @@
+{
+  /**
+   * This configuration is extended by all other TypeScript configurations.
+   */
+  "compilerOptions": {
+    "composite": true,
+    "esModuleInterop": true,
+    "lib": ["ES2020", "DOM"],
+    "module": "Node16",
+    "moduleResolution": "Node16",
+    "strict": true,
+    "target": "ES2020"
+  }
+}
--- a/tsconfig.scripts.json
+++ b/tsconfig.scripts.json
@@ -1,17 +1,22 @@
 {
+  /**
+   * This configuration is intended for the `scripts/` directory, both for linting and for editor
+   * TypeScript-related features.
+   *
+   * It's currently not actually used for that purpose, but it will be in a future PR.
+   *
+   * This is also extended by the `tsconfig.json` file in `scripts/create-package/`, which _is_
+   * actively used to support TypeScript-related editor features in that directory.
+   */
+  "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "baseUrl": "./",
-    "esModuleInterop": true,
     "exactOptionalPropertyTypes": true,
     "forceConsistentCasingInFileNames": true,
     "lib": ["ES2020"],
-    "module": "Node16",
-    "moduleResolution": "Node16",
     "noEmit": true,
     "noErrorTruncation": true,
-    "noUncheckedIndexedAccess": true,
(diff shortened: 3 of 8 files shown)


eslint-suppressions.json
```diff
@@ -100,6 +100,11 @@
       "count": 13
     }
   },
+  "packages/chain-agnostic-permission/src/scope/constants.ts": {
```
Gudahtt: I don't entirely know why, but this error appeared after making the `tsconfig.json` file more strict. This rule must depend upon one of the compiler options.

tsconfig.base.json
```diff
@@ -0,0 +1,25 @@
+{
+  /**
```
Gudahtt, line 2: TypeScript tolerates comments in `tsconfig` files, and even includes them in the `tsconfig` generated by `tsc --init`. We already had a comment in `tsconfig.packages.json` as well. Apparently they use something like "JSONC" (i.e. JSON with comments), though they have their own parser so it may not be identical.

It seemed OK to add more.

.gitattributes
```diff
@@ -2,6 +2,10 @@
 
 yarn.lock linguist-generated=false
 
+# `tsconfig.json` is already recognized as JSONC. This ensures our other `tsconfig` files are as
```
Gudahtt, line 5: This doesn't seem to be taking effect yet, but it should after this is merged.

This should ensure the comments in TSConfig files aren't highlighted in red on GitHub.

tsconfig.scripts.json
```diff
@@ -1,17 +1,22 @@
...
+  /**
+   * This configuration is intended for the `scripts/` directory, both for linting and for editor
+   * TypeScript-related features.
+   *
+   * It's currently not actually used for that purpose, but it will be in a future PR.
+   *
+   * This is also extended by the `tsconfig.json` file in `scripts/create-package/`, which _is_
+   * actively used to support TypeScript-related editor features in that directory.
+   */
+  "extends": "./tsconfig.base.json",
```
Gudahtt, line 11: This does result in two minor additions:

* `composite`
* `paths`

These seem fine to include here, and they don't seem to have any noticeable impact

Gudahtt: Nevermind on `paths`, I reverted that part because it resolved to a relative directory from a _package_, so it was the wrong directory when referenced in a root tsconfig file.

The only addition here now is `composite: true`

eslint-suppressions.json
Gudahtt: Nevermind, this was due to the `paths` property being set at a root-level tsconfig file and mapping `@metamask` to `../`, so it was mapping to another local repo I had checked out.

cryptodev-2s (approved):
LGTM!

PR #7284, opened 2025-12-02
feat(seedless-onboarding): add dataType support for secret data items

huggingbot:
## Explanation

The metadata service now supports a `dataType` column for categorizing secret data (PrimarySrp, ImportedSrp, ImportedPrivateKey). This enables clients to distinguish between different types of backed-up secrets

Changes include:
- Add `dataType` parameter to insert operations
- Add `updateSecretDataItem` and `batchUpdateSecretDataItems` for updating existing items
- Update `fetchAllSecretData` to return storage metadata (`itemId`, `dataType`) alongside secret data
- Refactor `SecretMetadata` to separate local metadata from storage-level metadata

## References

https://consensyssoftware.atlassian.net/browse/SL-350

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces breaking API changes and new migration logic that updates remote secret metadata and changes secret ordering/validation, so regressions could affect backup/restore correctness for existing users.
> 
> **Overview**
> Secret backups now carry a storage-level `dataType` (`PrimarySrp`/`ImportedSrp`/`ImportedPrivateKey`) via `@metamask/toprf-secure-backup@^1.0.0`, and `addNewSecretData` is **breaking**: it accepts `dataType: EncAccountDataType` (rejecting `PrimarySrp`) and persists that classification when adding items.
> 
> `fetchAllSecretData` now consumes the SDK’s richer `FetchedSecretDataItem` shape, returns `SecretMetadata` enriched with `itemId`/`dataType`/`createdAt`/`storageVersion`, and changes ordering/validation to use `PrimarySrp`-first sorting with TIMEUUID timestamp extraction (fixing previous lexicographic TIMEUUID sorting).
> 
> Adds migration plumbing: controller state gains persisted `migrationVersion` plus `runMigrations`/`setMigrationVersion`, with a `#migrateDataTypes` routine that updates legacy items’ `dataType` (single vs batch updates) and marks migration completion; `SecretMetadata` is refactored accordingly (removing old batch/parse/sort helpers and `version` in favor of `storageVersion`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71413498ff1dc034b2a2a6eb3bb6c00a912a8a63. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/seedless-onboarding-controller/CHANGELOG.md
+++ b/packages/seedless-onboarding-controller/CHANGELOG.md
@@ -7,13 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
-
-- Bump `@metamask/keyring-controller` from `^25.1.1` to `^25.2.0` ([#8363](https://github.com/MetaMask/core/pull/8363))
-- Bump `@metamask/messenger` from `^1.0.0` to `^1.1.1` ([#8364](https://github.com/MetaMask/core/pull/8364), [#8373](https://github.com/MetaMask/core/pull/8373))
-
-## [9.1.0]
-
 ### Added
 
 - Add `runMigrations` method to run pending data migrations for legacy secrets ([#7284](https://github.com/MetaMask/core/pull/7284))
@@ -25,6 +18,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `SecretMetadata.compareByTimestamp` static method for comparing metadata by timestamp ([#7284](https://github.com/MetaMask/core/pull/7284))
 - Add `SecretMetadata.matchesType` static method for checking if metadata matches a given type ([#7284](https://github.com/MetaMask/core/pull/7284))
 - Re-export `EncAccountDataType` from `@metamask/toprf-secure-backup` ([#7284](https://github.com/MetaMask/core/pull/7284))
+
+### Changed
+
+- Bump `@metamask/keyring-controller` from `^25.1.1` to `^25.2.0` ([#8363](https://github.com/MetaMask/core/pull/8363))
+- Bump `@metamask/messenger` from `^1.0.0` to `^1.1.1` ([#8364](https://github.com/MetaMask/core/pull/8364), [#8373](https://github.com/MetaMask/core/pull/8373))
+- Bump `@metamask/toprf-secure-backup` from `^0.11.0` to `^1.0.0` ([#7284](https://github.com/MetaMask/core/pull/7284))
+- **BREAKING:** Change `addNewSecretData` method signature to require `dataType: EncAccountDataType` instead of `type: SecretType` ([#7284](https://github.com/MetaMask/core/pull/7284))
+  - `SecretType` is now derived internally from `EncAccountDataType`
+  - Encrypted payload still includes `type` for backward compatibility with older clients
+- **BREAKING:** Remove `parseSecretsFromMetadataStore`, `fromBatch`, and `sort` methods from `SecretMetadata` ([#7284](https://github.com/MetaMask/core/pull/7284))
+  - Use `SecretMetadata.compare` or `SecretMetadata.compareByTimestamp` for sorting
+  - Use `SecretMetadata.matchesType` for filtering
+- **BREAKING:** Change `SecretMetadata.fromRawMetadata` signature to require `storageMetadata` parameter ([#7284](https://github.com/MetaMask/core/pull/7284))
+- **BREAKING:** Remove `version` getter from `SecretMetadata`; use `storageVersion` instead ([#7284](https://github.com/MetaMask/core/pull/7284))
+
+### Fixed
+
+- Fix TIMEUUID sorting by extracting actual timestamps instead of using lexicographic comparison ([#7284](https://github.com/MetaMask/core/pull/7284))
+
+## [9.1.0]
+
+### Added
+
 - Expose all public `SeedlessOnboardingController` methods through its messenger ([#8219](https://github.com/MetaMask/core/pull/8219))
   - The following actions are now available:
     - `SeedlessOnboardingController:fetchMetadataAccessCreds`
@@ -81,21 +97,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- **BREAKING:** Change `addNewSecretData` method signature to require `dataType: EncAccountDataType` instead of `type: SecretType` ([#7284](https://github.com/MetaMask/core/pull/7284))
-  - `SecretType` is now derived internally from `EncAccountDataType`
-  - Encrypted payload still includes `type` for backward compatibility with older clients
-- **BREAKING:** Remove `parseSecretsFromMetadataStore`, `fromBatch`, and `sort` methods from `SecretMetadata` ([#7284](https://github.com/MetaMask/core/pull/7284))
-  - Use `SecretMetadata.compare` or `SecretMetadata.compareByTimestamp` for sorting
-  - Use `SecretMetadata.matchesType` for filtering
-- **BREAKING:** Change `SecretMetadata.fromRawMetadata` signature to require `storageMetadata` parameter ([#7284](https://github.com/MetaMask/core/pull/7284))
-- **BREAKING:** Remove `version` getter from `SecretMetadata`; use `storageVersion` instead ([#7284](https://github.com/MetaMask/core/pull/7284))
 - `refreshAuthTokens` now coalesces concurrent calls — if a refresh is already in-flight, subsequent callers share the same promise rather than issuing duplicate HTTP requests with the same token ([#7989](https://github.com/MetaMask/core/pull/7989))
 - `refreshAuthTokens` now uses the live `state.refreshToken` value when calling `authenticate` after a successful HTTP refresh, preventing a stale-capture bug that could overwrite a concurrently-renewed token with an older one ([#7989](https://github.com/MetaMask/core/pull/7989))
 - `renewRefreshToken` now queues the old token for revocation only after `#createNewVaultWithAuthData` succeeds, ensuring tokens are not queued if vault creation fails ([#7989](https://github.com/MetaMask/core/pull/7989))
 
 ### Fixed
 
-- Fix TIMEUUID sorting by extracting actual timestamps instead of using lexicographic comparison ([#7284](https://github.com/MetaMask/core/pull/7284))
 - Fixed `refreshAuthTokens` throwing the generic `FailedToRefreshJWTTokens` error for all HTTP failures — a 401 response (permanently revoked token) now throws `InvalidRefreshToken` instead, giving callers a distinct signal to distinguish permanent from transient failures ([#7989](https://github.com/MetaMask/core/pull/7989))
 
 ## [8.0.0]


packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -14,6 +18,29 @@ import type {
...
+
+  /**
+   * The server-assigned item ID for this row.
+   */
+  itemId?: string;
+
+  /**
+   * The client-assigned data type classification.
+   */
+  dataType?: EncAccountDataType;
```
lwin-kyaw: Why do we need this extra type `SecretMetadata` already has it has `type`

packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -14,6 +18,29 @@ import type {
...
+> = {
+  /**
+   * The parsed secret metadata (serialized data).
+   */
+  secret: SecretMetadata<DataType>;
+
+  /**
+   * The server-assigned item ID for this row.
+   */
+  itemId?: string;
```
lwin-kyaw: Should add to the `SecretMetadata` instead of creating another wrapping types

packages/seedless-onboarding-controller/src/types.ts
```diff
@@ -14,6 +18,29 @@ import type {
   SecretType,
   Web3AuthNetwork,
 } from './constants';
+import type { SecretMetadata } from './SecretMetadata';
+
+/**
+ * A secret data item with storage-level metadata from the metadata store.
+ */
+export type SecretDataItemWithMetadata<
```
lwin-kyaw: Please add the required metadata to the `SecretMetadata` class instead of creating another type defs.

The purpose the metadata class is to store and do things with extra non-secret data.

packages/seedless-onboarding-controller/src/index.ts
```diff
@@ -23,3 +24,5 @@ export {
 } from './constants';
 export { SecretMetadata } from './SecretMetadata';
 export { RecoveryError } from './errors';
+
+export { EncAccountDataType } from '@metamask/toprf-secure-backup';
```
lwin-kyaw, line 58: Not sure why we export this, but this should not be determined from the users (ext or mobile).

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -438,12 +441,14 @@ export class SeedlessOnboardingController<
...
    * @param seedPhrase - The initial seed phrase (Mnemonic) created together with the wallet.
    * @param keyringId - The keyring id of the backup seed phrase
+   * @param dataType - Optional data type for categorizing the secret data. Defaults to PrimarySrp.
    * @returns A promise that resolves to the encrypted seed phrase and the encryption key.
    */
   async createToprfKeyAndBackupSeedPhrase(
     password: string,
     seedPhrase: Uint8Array,
     keyringId: string,
+    dataType: EncAccountDataType = EncAccountDataType.PrimarySrp,
```
lwin-kyaw: `createToprfKeyAndBackupSeedPhrase` only used for new users scenario, this param should not be exposed at all and type is always `primary`.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -495,13 +501,15 @@ export class SeedlessOnboardingController<
...
    * @param options.keyringId - The keyring id of the backup keyring (SRP).
+   * @param options.dataType - Optional data type for categorizing the secret data.
    * @returns A promise that resolves to the success of the operation.
    */
   async addNewSecretData(
     data: Uint8Array,
     type: SecretType,
     options?: {
       keyringId?: string;
+      dataType?: EncAccountDataType;
```
lwin-kyaw: we have `type` field in 2nd param, we don't need new field here.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -532,14 +540,113 @@ export class SeedlessOnboardingController<
...
+   *
+   * This method updates metadata fields for an existing secret data item
+   * without modifying the encrypted data itself.
+   *
+   * @param params - The parameters for updating the secret data item.
+   * @param params.itemId - The item ID of the secret data to update.
+   * @param params.dataType - The data type to set for the item.
+   * @returns A promise that resolves when the update is complete.
+   */
+  async updateSecretDataItem(params: {
```
lwin-kyaw: Can we specify why and when we need to call this? since in the clients, `update*` is not something we usually do with SRPs

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -532,14 +540,113 @@ export class SeedlessOnboardingController<
...
+   * Batch update fields for multiple existing secret data items by their itemIds.
+   *
+   * This method updates metadata fields for multiple existing secret data items
+   * without modifying the encrypted data itself.
+   *
+   * @param params - The parameters for batch updating secret data items.
+   * @param params.updates - Array of objects containing itemId and fields to update.
+   * @returns A promise that resolves when all updates are complete.
+   */
+  async batchUpdateSecretDataItems(params: {
```
lwin-kyaw: Same here.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1180,16 +1287,27 @@ export class SeedlessOnboardingController<
...
+
+      // Sort by timestamp (oldest first)
+      results.sort((a, b) =>
+        SecretMetadata.compareByTimestamp(a.secret, b.secret, 'asc'),
+      );
+
       // validate the primary secret data is a mnemonic (SRP)
-      const primarySecret = secrets[0];
-      if (primarySecret.type !== SecretType.Mnemonic) {
+      if (!SecretMetadata.matchesType(results[0].secret, SecretType.Mnemonic)) {
```
lwin-kyaw: Since we have `Primary` type, why don't we use here?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1268,6 +1387,7 @@ export class SeedlessOnboardingController<
     authKeyPair: KeyPair;
     options?: {
       keyringId?: string;
+      dataType?: EncAccountDataType;
```
lwin-kyaw: we already have `type` in the params object. Please use it.

packages/seedless-onboarding-controller/src/types.ts
huggingbot: `type` is in encrypted data, which the server can't access to
`data_type` is a server-side column
the server needs to know the type of key for the deletion api: we're preventing clients from deleting the default srp

packages/seedless-onboarding-controller/src/types.ts
huggingbot: `itemId` is not part of the secret data, hence it shouldn't belong inside `SecretMetadata`. Only `data`, `timestamp`, `type` and `version` are part of `SecretMetadata`

packages/seedless-onboarding-controller/src/types.ts
huggingbot: `itemId` and `dataType` are not part of the secret data. They belong outside of it, hence the wrapper type `SecretDataItemWithMetadata` is introduced.

> The purpose the metadata class is to store and do things with extra non-secret data.

by the metadata class, did u mean the `SecretMetadata` class?

packages/seedless-onboarding-controller/src/index.ts
huggingbot: it's exported so clients can use it in `createToprfKeyAndBackupSeedPhrase`, `addNewSecretData`, `updateSecretDataItem` and `batchUpdateSecretDataItems`. The clients do determine the type of secret, e.g. in `addNewSecretData`, there's a `type` param of type `SecretType` where the client can decide if it's a `mnemonic` or `privateKey`

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: `SecretType` and `EncAccountDataType` are slightly different:
- `SecretType` only has 2 types: SRP or private key; `EncAccountDataType` has 3: Primary SRP, imported SRP, imported private key
- Even if we were to extend `SecretType` to include a primary SRP, using string values should be avoided to protect user privacy when data is transmitted to the server

we can attempt to migrate `SecretType` from string to number, making it essentially the same as `EncAccountDataType`. Though they are essentially different things, and in the future we might want to remove `type` from the encrypted data since we will use `dataType` as the source of truth. What do u think?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: it's used to migrate existing users to update the `data_type` column in the metadata service. Currently the server doesn't know the type of secret (it's encrypted). It needs to know for the deletion API to work: to prevent clients from deleting the default SRP

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: related question and answer here: https://github.com/MetaMask/core/pull/7284#discussion_r2580885153

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/85275f56407e4d0342f74565b587c83c379da193)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: added more notes [here](https://github.com/MetaMask/core/pull/7284/commits/fc73107ccdf8d53ebdd56b5ba0bee5c87592136b)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: added more notes [here](https://github.com/MetaMask/core/pull/7284/commits/fc73107ccdf8d53ebdd56b5ba0bee5c87592136b)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/3afa87907945bad3407d16c50e75cc646365fe40)
now we're validating both the `.type` in the encrypted data and also `dataType`

packages/seedless-onboarding-controller/src/types.ts
huggingbot: moved `itemId` and `dataType` [here](https://github.com/MetaMask/core/pull/7284/commits/11c09f898346bb4107164f7d46b967dbc1b8fb13)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1180,16 +1287,48 @@ export class SeedlessOnboardingController<
...
+      const results: SecretMetadata[] = secretDataItems.map((item) =>
+        SecretMetadata.fromRawMetadata(item.data, {
+          itemId: item.itemId,
+          dataType: item.dataType,
+          createdAt: item.createdAt,
+        }),
+      );
+
+      // Sort: PrimarySrp first, then by createdAt/timestamp (oldest first)
+      results.sort((a, b) => {
```
huggingbot: sorting is required here even though /get in metadata already does sorting because we need this for the migration since `dataType` and `createdAt` could be `null` thus we'll need a fallback to use the client-side timestamp

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -53,50 +63,35 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
-  constructor(data: DataType, options?: Partial<SecretMetadataOptions>) {
+  constructor(
+    data: DataType,
+    options?: Partial<SecretMetadataOptions>,
+    storageMetadata?: StorageMetadata,
+  ) {
     this.#data = data;
     this.#timestamp = options?.timestamp ?? Date.now();
     this.#type = options?.type ?? SecretType.Mnemonic;
     this.#version = options?.version ?? SecretMetadataVersion.V1;
```
lwin-kyaw: We should maintain the metadata type version with this.

The existing (old) items will have version number `SecretMetadataVersion.V1` after decryption. We can do the increment on version number (V2) for new structure.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -122,42 +117,16 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
    * @param rawMetadata - The raw metadata.
+   * @param storageMetadata - Storage-level metadata from the metadata store.
    * @returns The parsed secret metadata.
    */
   static fromRawMetadata<DataType extends SecretDataType>(
     rawMetadata: Uint8Array,
+    storageMetadata?: StorageMetadata,
   ): SecretMetadata<DataType> {
     const serializedMetadata = bytesToString(rawMetadata);
     const parsedMetadata = JSON.parse(serializedMetadata);
```
lwin-kyaw, line 143: Please update the `SecretMetadata.assertIsValidSecretMetadataJson` with the new structure.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -175,31 +144,47 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
-  ): SecretMetadata<DataType>[] {
-    return data.sort((a, b) => {
-      if (order === 'asc') {
-        return a.timestamp - b.timestamp;
-      }
-      return b.timestamp - a.timestamp;
-    });
+  ): number {
+    return order === 'asc'
+      ? a.timestamp - b.timestamp
```
lwin-kyaw, line 177: shouldn't this be compared using the server timestamp (`createdAt`?)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: If we intend to do the migration at the `unlock` and `rehydration`, during this `addNewSecretData` method call, we can assume that the user data is already migrated. Hence, we don't need both `SecretType` and `EncAccountDataType` to be present.

We can use `EncAccountDataType` type for 2nd param `type`.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Imo, I don't think we should expose this method. This is purely internal controller logic. Rather than adding migration code to the client sides (ext/mobile), we should handle this internally inside the controller package (with proper versioning).

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -547,12 +553,113 @@ export class SeedlessOnboardingController<
...
+   *
+   * This is the batch version of updateSecretDataItem, used for migrating
+   * multiple legacy secrets in a single operation. Useful when a user with
+   * multiple SRPs/private keys upgrades from an older SDK version.
+   *
+   * @param params - The parameters for batch updating secret data items.
+   * @param params.updates - Array of objects containing itemId and dataType to assign.
+   * @returns A promise that resolves when all updates are complete.
+   */
+  async batchUpdateSecretDataItems(params: {
```
lwin-kyaw: Please move the migration logics to internal scope.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1308,48 @@ export class SeedlessOnboardingController<
...
+      const results: SecretMetadata[] = secretDataItems.map((item) =>
+        SecretMetadata.fromRawMetadata(item.data, {
+          itemId: item.itemId,
+          dataType: item.dataType,
+          createdAt: item.createdAt,
+        }),
+      );
+
+      // Sort: PrimarySrp first, then by createdAt/timestamp (oldest first)
+      results.sort((a, b) => {
```
lwin-kyaw: Why don't we move the whole logic into SecretMetadata class?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1308,48 @@ export class SeedlessOnboardingController<
...
+        // Use server-side createdAt if available (TIMEUUID requires timestamp extraction)
+        if (a.createdAt && b.createdAt) {
+          return compareTimeuuid(a.createdAt, b.createdAt);
+        }
+        // Fall back to client-side timestamp
+        return SecretMetadata.compareByTimestamp(a, b, 'asc');
+      });
+
+      // Validate the first item is the primary SRP
+      const firstItem = results[0];
```
lwin-kyaw, line 1633: I think we can refactor this logics into SecretMetadata class.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1308,48 @@ export class SeedlessOnboardingController<
...
+      const isDataTypePrimary =
+        firstItem.dataType === undefined || // Legacy data (before dataType was introduced)
+        firstItem.dataType === EncAccountDataType.PrimarySrp;
+      const isMnemonic = SecretMetadata.matchesType(
```
lwin-kyaw, line 1638: After the migration, this become invalid right? We won't have the `type` anymore and we will be relying on the EncAccountDataType.PrimarySrp.

So, this will incorrectly result in `false` and the below error will be thrown.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1463,56 @@ export class SeedlessOnboardingController<
...
+
+      // Sort: PrimarySrp first, then by createdAt/timestamp (oldest first)
+      results.sort((a, b) => {
+        // PrimarySrp always comes first
+        if (a.dataType === EncAccountDataType.PrimarySrp) {
+          return -1;
+        }
+        if (b.dataType === EncAccountDataType.PrimarySrp) {
+          return 1;
+        }
```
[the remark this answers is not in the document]
huggingbot: there can only be a single primary srp

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -42,6 +43,22 @@ type SecretMetadataJson<DataType extends SecretDataType> = Omit<
...
+  itemId?: string;
+  dataType?: EncAccountDataType;
+  createdAt?: string;
+  /**
+   * The storage-level version from the SDK ('v1' or 'v2').
+   * This is different from the encrypted data version (SecretMetadataVersion).
+   * - 'v1': Legacy items created before dataType was introduced
+   * - 'v2': Items with dataType set (either new or migrated)
+   */
+  storageVersion?: 'v1' | 'v2';
```
lwin-kyaw: nit: would be nice if we can import this from the sdk.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -53,50 +70,38 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
   readonly #version: SecretMetadataVersion;
 
+  // Storage-level metadata (not encrypted)
+  readonly #itemId?: string;
+
+  readonly #dataType?: EncAccountDataType;
+
+  readonly #createdAt?: string;
+
+  readonly #storageVersion?: 'v1' | 'v2';
```
lwin-kyaw: Should we avoid hardcoding version number?

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -122,42 +127,16 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
   /**
    * Parse and create the SecretMetadata instance from the raw metadata bytes.
    *
    * @param rawMetadata - The raw metadata.
+   * @param storageMetadata - Storage-level metadata from the metadata store.
    * @returns The parsed secret metadata.
    */
   static fromRawMetadata<DataType extends SecretDataType>(
     rawMetadata: Uint8Array,
+    storageMetadata?: StorageMetadata,
```
lwin-kyaw: Looks like this can be a required param.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
lwin-kyaw: May I know are there any major difference between `#storageVersion` and `#version`?
If not we can remove either one of them.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/7634383b6fc3520916573264ddd153ed02d9bc40)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/7634383b6fc3520916573264ddd153ed02d9bc40)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -53,50 +70,38 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
    * @param options - The options for the secret metadata.
    * @param options.timestamp - The timestamp when the secret was created.
    * @param options.type - The type of the secret.
+   * @param options.version - The version of the secret metadata.
+   * @param storageMetadata - Storage-level metadata from the metadata store.
    */
-  constructor(data: DataType, options?: Partial<SecretMetadataOptions>) {
+  constructor(
+    data: DataType,
+    options?: Partial<SecretMetadataOptions>,
```
lwin-kyaw: Are we keeping this options for the backward compatibility?

Just wanna check if we can remove this and replace with `storageMetadata`? Something like this

```ts
// V1
const v1StorageMetadata = {
  ...,
  storageVersion: 'v1',
  dataType: convertToV2DataType(type_v1),
}
```

I don't think we don't need to maintain different structs for different versions as many of them share the same traits.
`SecretMetadata` class does not need to strictly follow the actual storage structure and it's flexible at this level.

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -175,31 +154,47 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
     }
 
-    return new SecretMetadata<DataType>(data, {
-      timestamp: parsedMetadata.timestamp,
-      type,
-      version,
-    });
+    return new SecretMetadata<DataType>(
+      data,
+      {
```
lwin-kyaw: We can convert this object to `StorageMetadata`.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -510,13 +522,15 @@ export class SeedlessOnboardingController<
...
    * @param options.keyringId - The keyring id of the backup keyring (SRP).
+   * @param options.dataType - Optional data type for categorizing the secret data.
    * @returns A promise that resolves to the success of the operation.
    */
   async addNewSecretData(
     data: Uint8Array,
     type: SecretType,
     options?: {
       keyringId?: string;
+      dataType?: EncAccountDataType;
```
lwin-kyaw: As we have this `EncAccountDataType`, I think we can remove the 2nd param, `type`, right?
Users do not need to provide two types. wdyt?

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/12d0ed0ff7c45d5a47a40dfc0aa9f097c5df1104)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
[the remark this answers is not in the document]
huggingbot: there can only be a single primary srp

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1463,56 @@ export class SeedlessOnboardingController<
...
+        }
+        // Both null: fall back to client-side timestamp
+        return SecretMetadata.compareByTimestamp(a, b, 'asc');
+      });
+
+      // Validate the first item is the primary SRP
+      const firstItem = results[0];
+      const isDataTypePrimary =
+        firstItem.dataType === undefined || // Legacy data (before dataType was introduced)
+        firstItem.dataType === EncAccountDataType.PrimarySrp;
```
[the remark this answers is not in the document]
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/e820ef57d31b9df25c056ba275e681a1c515c10f)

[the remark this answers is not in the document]
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/e820ef57d31b9df25c056ba275e681a1c515c10f)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/changes/22b695e7426b57e262db68aa64b9675e3195ac70)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/c0dcad7912a238d9de2a7fe9e66bc502164cb0f9)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/c0dcad7912a238d9de2a7fe9e66bc502164cb0f9)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -547,12 +559,260 @@ export class SeedlessOnboardingController<
...
+      this.#assertIsUnlocked();
+
+      if (
+        this.state.migrationVersion <
+        SeedlessOnboardingMigrationVersion.DataType
+      ) {
+        await this.#migrateDataTypes();
+      }
+    });
+  }
```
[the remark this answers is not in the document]
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/6e5a3cebab4ed6d17ec4c90ea563ff25f1a35850)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -51,52 +78,33 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
+    this.#dataType = options?.dataType;
+    this.#createdAt = options?.createdAt;
+    this.#storageVersion = options?.storageVersion;
+
+    // Derive type from dataType (new clients), or use provided type (V1 compat)
+    if (options?.dataType === undefined) {
+      this.#type = options?.type ?? SecretType.Mnemonic;
+    } else {
+      this.#type = getSecretTypeFromDataType(options.dataType);
+    }
```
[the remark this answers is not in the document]
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/ed0177958f8bdf0f5e823e3542e6e0ec1402c6b9)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -547,12 +558,266 @@ export class SeedlessOnboardingController<
...
+   * their secrets were stored without dataType classification. This method allows
+   * clients to retroactively assign the correct dataType (e.g., PrimarySrp,
+   * ImportedSrp) when the wallet is unlocked.
+   *
+   * @param params - The parameters for updating the secret data item.
+   * @param params.itemId - The server-assigned item ID from fetchAllSecretData.
+   * @param params.dataType - The data type classification to assign.
+   * @returns A promise that resolves when the update is complete.
+   */
+  async updateSecretDataItem(params: {
```
lwin-kyaw: We have `runMigrations` method, do we still need this `updateSecretDataItem`?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Should we move this logic to `SecretMetadata` class? wdyt?

packages/seedless-onboarding-controller/src/constants.ts
```diff
@@ -20,8 +20,8 @@ export enum SecretType {
   PrivateKey = 'privateKey',
 }
 
-export enum SecretMetadataVersion {
-  V1 = 'v1',
+export enum SeedlessOnboardingMigrationVersion {
+  DataType = 1,
```
lwin-kyaw: I think we should rename this, to avoid confusion.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1201,16 +1466,57 @@ export class SeedlessOnboardingController<
...
+
+      // Sort: PrimarySrp first, then by createdAt/timestamp (oldest first)
+      results.sort((a, b) => {
+        // PrimarySrp always comes first
+        if (a.dataType === EncAccountDataType.PrimarySrp) {
+          return -1;
+        }
+        if (b.dataType === EncAccountDataType.PrimarySrp) {
+          return 1;
+        }
```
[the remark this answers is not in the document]
huggingbot: there's only 1 primary srp

packages/seedless-onboarding-controller/src/constants.ts
huggingbot: what do u suggest the name to be?

huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/changes/2f02601aa6623a1fab2384bb556b13f71195dd2c)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/changes/10c0c09a5f6d6656ed0e5a1b0f37668d9d8498dd)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
```diff
@@ -178,28 +156,81 @@ export class SecretMetadata<DataType extends SecretDataType = Uint8Array>
...
-      return b.timestamp - a.timestamp;
-    });
+  ): number {
+    // PrimarySrp always comes first (regardless of order direction)
+    if (a.dataType === EncAccountDataType.PrimarySrp) {
+      return -1;
+    }
+    if (b.dataType === EncAccountDataType.PrimarySrp) {
+      return 1;
+    }
```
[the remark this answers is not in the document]
huggingbot: there can only be one primary srp.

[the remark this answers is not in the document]
huggingbot: there can only be one primary srp.

packages/seedless-onboarding-controller/src/types.ts
huggingbot: have moved this wrapper into `SecretMetadata`

packages/seedless-onboarding-controller/src/index.ts
huggingbot: it's now not export as it's used in the controller only

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: > We can use EncAccountDataType type for 2nd param type.

fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/changes/40193282fa44ebf59067cdd66ae483d94a9e3411)

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: removed `type` from encrypted data entirely

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: no new items were added into `parsedMetadata` so no update should be needed

packages/seedless-onboarding-controller/src/SecretMetadata.ts
huggingbot: we need a fallback during migration where `createdAt` is empty

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: we are keeping it as it is

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: we'd still need `type` for backward compatibility

lwin-kyaw (dismissed):

himanshuchawla009 (approved):

packages/seedless-onboarding-controller/src/SecretMetadata.ts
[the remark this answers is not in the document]
huggingbot: there is only 1 primary srp

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -547,12 +557,160 @@ export class SeedlessOnboardingController<
...
+        if (!secret.itemId || secret.itemId === 'PW_BACKUP') {
+          continue;
+        }
+
+        // Skip items that are already migrated (v2 with dataType set)
+        // Check both storageVersion and dataType since this migration is specific to dataType
+        const isAlreadyMigrated =
+          secret.storageVersion === 'v2' &&
+          secret.dataType !== undefined &&
+          secret.dataType !== null;
```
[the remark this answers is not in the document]
huggingbot: this is fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -547,12 +557,163 @@ export class SeedlessOnboardingController<
...
+        ) {
+          this.#setMigrationVersion(SeedlessOnboardingMigrationVersion.V1);
+          return;
+        }
+        throw error;
+      }
+
+      let hasPrimarySrp = secretDatas.some(
+        (secret) => secret.dataType === EncAccountDataType.PrimarySrp,
+      );
```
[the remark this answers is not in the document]
huggingbot: this is fixed

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -504,17 +514,17 @@ export class SeedlessOnboardingController<
...
-   * @param type - The type of the secret data.
+   * @param dataType - The data type classification for the secret data.
    * @param options - Optional options object, which includes optional data to be added to the metadata store.
    * @param options.keyringId - The keyring id of the backup keyring (SRP).
    * @returns A promise that resolves to the success of the operation.
    */
   async addNewSecretData(
     data: Uint8Array,
-    type: SecretType,
+    dataType: EncAccountDataType,
```
lwin-kyaw, line 683: nit: would be better not to accept the `EncAccountDataType.PrimarySrp` here to avoid any issues in metadata side.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -474,7 +484,7 @@ export class SeedlessOnboardingController<
         // encrypt and store the secret data
         await this.#encryptAndStoreSecretData({
           data: seedPhrase,
-          type: SecretType.Mnemonic,
+          dataType: EncAccountDataType.PrimarySrp,
```
lwin-kyaw, line 640: I think we should do the `this.setMigrationVersion` after the successful backup creation.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/commits/614be60d032656111a905879063c68300dff1cbf)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
huggingbot: fixed [here](https://github.com/MetaMask/core/pull/7284/changes/c1e534ce6445fa24e7a53b11675fa58271f2aaad)

lwin-kyaw (approved):
LGTM!!! 🚀

himanshuchawla009 (dismissed):

packages/seedless-onboarding-controller/src/utils.ts
```diff
@@ -146,3 +148,75 @@ export function compareAndGetLatestToken(
...
+ * @returns The 60-bit timestamp as a bigint.
+ */
+export function getTimestampFromTimeuuid(uuid: string): bigint {
+  const parts = uuid.split('-');
+  const timeLow = parts[0]; // 32 bits (least significant)
+  const timeMid = parts[1]; // 16 bits
+  const timeHi = parts[2].slice(1); // 12 bits (remove version nibble '1')
+  // Reconstruct timestamp: timeHi | timeMid | timeLow
+  return BigInt(`0x${timeHi}${timeMid}${timeLow}`);
+}
```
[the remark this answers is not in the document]
huggingbot: `createdAt` will always be in a uuid format

PR #7295, opened 2025-12-02
Release/705.0.0

micaelae:
## Explanation

Releases major versions of the bridge-controller and bridge-status-controller

## Checklist

---

> [!NOTE]
> Release major versions of bridge controllers and update dependent packages and changelogs.
> 
> - **Packages**:
>   - **`@metamask/bridge-controller@64.0.0`**:
>     - Added: export `fetchTokens` and `type SwapsToken`; handle USDT insufficient non‑zero allowance via `resetApproval` (incl. fee calc).
>     - Changed: BREAKING remove `SWAPS_TESTNET_CHAIN_ID` (use `CHAIN_IDS.LOCALHOST`); bump `network-controller` and `transaction-controller`; bump `assets-controllers`.
>     - Removed: BREAKING public `getBridgeERC20Allowance` action.
>     - Fixed: BREAKING add `usd_amount_source` to QuotesRequested metrics; add MON/SEI integer chain IDs.
>   - **`@metamask/bridge-status-controller@64.0.0`**:
>     - Changed: BREAKING bump `@metamask/bridge-controller` to `^64.0.0`; type‑safety improvements; submit `resetApproval` before approval; bump `network-controller` and `transaction-controller`.
>   - **`@metamask/transaction-pay-controller`**:
>     - Changed: bump `@metamask/bridge-controller` to `^64.0.0` and `@metamask/bridge-status-controller` to `^64.0.0`; bump `assets-controllers`; update changelog.
> - **Repo**:
>   - Bump monorepo version to `705.0.0`; update lockfile.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1e2b27291ee1374b49ce793fa6d24c02f4668c7c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


SteP-n-s (approved):

packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [64.0.0]
+
 ### Changed
 
+- **BREAKING:** Bump `@metamask/bridge-controller` from `^63.2.0` to `^64.0.0` ([#7295](https://github.com/MetaMask/core/pull/7295))
```
Gudahtt, line 14: Is this still a breaking change? If so, why?

Note that it's no longer a peer dependency, so these dependency updates aren't breaking changes every time anymore.

micaelae: @Gudahtt bridge-controller has 3 breaking changes:
- The getBridgeERC20Allowance handler has been made private. Prior to this version bump, bridge-status-controller was calling that handler. However this new version of status-controller removes the call
- The new required `usd_amount_source` context parameter in the updateBridgeQuotes handler will cause a type error in the clients, but doesn't break bridge-status-controller
- Removing previously exported `SWAPS_TESTNET_CHAIN_ID` constants

Only the first breaking change directly affects bridge-status-controller. Should it still be marked as a BREAKING change if the "break" is addressed in this new bridge-status-controller version?

matthewwalsh0 (approved):

packages/bridge-status-controller/CHANGELOG.md
Gudahtt: Not a huge deal, but for future reference, each "breaking" entry should explain _why_/_how_ it's breaking.

e.g. if I were to update the `@metamask/bridge-status-controller` (on it's own, without updating anything else including `@metamask/bridge-controller`), what would break? How should the user of this package address the breakage? That's what we're trying to convey here.

It sounds like maybe the messenger type for `bridge-status-controller` changed in a breaking way in this case? So we could explain that there was a change there, and that to address it you can remove the action that is no longer being used. Something like that. If the package would break without also updating `bridge-controller` (e.g. even without producing a type error), that's _definitely_ worth calling out_. 

And this isn't really the correct entry to list it under either - this is due to #7228, it's not due to the dependency change. Ideally this would have been added in that PR.

PR #7415, opened 2025-12-09
feat: Adding requestId field to dapp confirmations

jpuri:
## Explanation

The PR adds id from the JSON RPC request to the transaction meta object. Thus change is done at 2 places:
1. Transaction controller
2. EIP-5792 middleware

## References

Related to [#67890](https://github.com/MetaMask/MetaMask-planning/issues/6345)

## Checklist

- [X] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs)
- [X] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> Propagates the JSON-RPC request id into transaction metadata for single and batched flows, wiring through EIP-5792 middleware and TransactionController with updated types, tests, and changelogs.
> 
> - **Transaction Controller**:
>   - Add `requestId` to `TransactionMeta`, `TransactionBatchMeta`, `AddTransactionOptions`, and `TransactionBatchRequest`.
>   - Plumb `requestId` through `addTransaction`, `addTransactionBatch`, and batch utilities (`utils/batch.ts`).
>   - Update tests to reflect new metadata field.
> - **EIP-5792 Middleware**:
>   - Extract `req.id` and pass as `requestId` to `addTransaction`/`addTransactionBatch` in `processSendCalls`.
>   - Adjust tests to assert `requestId` propagation.
> - **Changelogs**:
>   - Document addition of `requestId` across both packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e854e886ab0853c934bdea3cefb3dc523debbb5a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/eip-5792-middleware/src/hooks/processSendCalls.ts
```diff
@@ -101,6 +101,8 @@ export async function processSendCalls(
   const securityAlertId = uuid();
   const validateSecurity = validateSecurityHook.bind(null, securityAlertId);
 
+  const requestId = String(req.id);
```
OGPoyraz: What if we use `JsonRpcId` type from `@metamask/utils` in the `TransactionController` types? As v2 JSONRpcEngine referring the same types https://github.com/MetaMask/core/blob/main/packages/json-rpc-engine/src/v2/JsonRpcEngineV2.ts#L4

Then we wouldn't really need to cast here and just pass as is?

packages/eip-5792-middleware/src/hooks/processSendCalls.ts
OGPoyraz, line 1: Do we want to note additions in changelog on this package and `transaction-controller`?

packages/eip-5792-middleware/src/hooks/processSendCalls.ts
jpuri: String is I think more useful even later when we use it in the UI

packages/eip-5792-middleware/src/hooks/processSendCalls.ts
jpuri: Good point, let me do that.

packages/eip-5792-middleware/src/hooks/processSendCalls.ts
OGPoyraz: Sorry not sure I follow, what makes string to be useful and how's that related with UI? 

It's a minor thing for sure, but am not completely convinced if casting here necessary. So I will defer this to @dan437

dan437: I would prefer to use the `JsonRpcId` type and not cast it into String for consistency reasons.
```
import {
  type JsonRpcId,
} from '@metamask/utils';
```

jpuri: We are using JSOP RPC request ID here but I do not want to tie completely to it, we may extend this field with some custom value and we are going to use it a lot in UI. `string` is more generic type, that is also reason it was casted for actionId.

ffmcgee725 (approved):

PR #7417, opened 2025-12-09
chore(lint): Fix suppressed ESLint errors in `sample-controllers` package

### LINKED ISSUE #7389: Fix suppressed lint errors in `sample-controllers`
Fix suppressed lint errors in this package.

Mrtenz:
## Explanation

This fixes all suppressed ESLint errors in the `sample-controllers` package.

## References

Closes #7389.

## Checklist

---

> [!NOTE]
> Adds explicit return types and error capture in sample controllers/services, updates tests accordingly, and removes related ESLint suppressions.
> 
> - **Controllers**:
>   - `SampleGasPricesController`
>     - Add explicit return types to `updateGasPrices` and `#onSelectedNetworkClientIdChange`.
>     - On network client change, call `updateGasPrices(...).catch(...)` and forward errors to `messenger.captureException`.
>   - `SamplePetnamesController`
>     - Add explicit `void` return type to `assignPetname`.
> - **Service**:
>   - `SampleGasPricesService`
>     - Add explicit `IDisposable` return types for `onRetry`, `onBreak`, `onDegraded`.
> - **Tests**:
>   - `sample-gas-prices-controller.test.ts`: mock `captureException`; add test asserting it’s called on fetch failure.
>   - `sample-gas-prices-service.test.ts`: make `onRetry` callbacks non-async and handle `clock.nextAsync()` via `.catch(console.error)`.
> - **Lint**:
>   - Remove ESLint suppressions for `sample-controllers` files in `eslint-suppressions.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 82ce74518de24420cf053f86d7ad0d09c2e0e356. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sample-controllers/src/sample-gas-prices-controller.test.ts
+++ b/packages/sample-controllers/src/sample-gas-prices-controller.test.ts
@@ -202,7 +202,7 @@ describe('SampleGasPricesController', () => {
       });
     });
 
-    it('calls `ErrorReportingService:captureException` if fetching gas prices fails', async () => {
+    it('calls `messenger.captureException` if fetching gas prices fails', async () => {
       await withController(async ({ rootMessenger }) => {
         const chainId = '0x42';
 
@@ -214,15 +214,10 @@ describe('SampleGasPricesController', () => {
           throw new Error(`Unrecognized chain ID '${givenChainId}'`);
         });
 
-        const captureException = jest.fn();
+        const captureException = jest.spyOn(rootMessenger, 'captureException');
         rootMessenger.registerActionHandler(
           'SampleGasPricesService:fetchGasPrices',
           fetchGasPrices,
-        );
-
-        rootMessenger.registerActionHandler(
-          'ErrorReportingService:captureException',
-          captureException,
         );
 
         rootMessenger.registerActionHandler(
@@ -438,7 +433,10 @@ type WithControllerOptions = {
  * @returns The root messenger.
  */
 function getRootMessenger(): RootMessenger {
-  return new Messenger({ namespace: MOCK_ANY_NAMESPACE });
+  return new Messenger({
+    namespace: MOCK_ANY_NAMESPACE,
+    captureException: jest.fn(),
+  });
 }
 
 /**
@@ -457,7 +455,6 @@ function getMessenger(
   });
   rootMessenger.delegate({
     actions: [
-      'ErrorReportingService:captureException',
       'NetworkController:getNetworkClientById',
       'SampleGasPricesService:fetchGasPrices',
     ],
--- a/packages/sample-controllers/src/sample-gas-prices-controller.ts
+++ b/packages/sample-controllers/src/sample-gas-prices-controller.ts
@@ -4,7 +4,6 @@ import type {
   StateMetadata,
 } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
-import { ErrorReportingServiceCaptureExceptionAction } from '@metamask/error-reporting-service';
 import type { Messenger } from '@metamask/messenger';
 import type {
   NetworkClientId,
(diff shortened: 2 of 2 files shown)


packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -109,6 +110,7 @@ export type SampleGasPricesControllerActions =
  * Actions from other messengers that {@link SampleGasPricesMessenger} calls.
  */
 type AllowedActions =
+  | ErrorReportingServiceCaptureExceptionAction
```
Gudahtt: We'd have to document this as a breaking change. I know these are meant to be examples, but they're published and actually used in the clients as part of an example UI feature as well.

Though.... there is an alternative that is both preferable and non-breaking, and that is to use `messenger.captureException` (see https://github.com/MetaMask/core/blob/206599b079eb0a20b5162e4cf11a065384245973/packages/messenger/src/Messenger.ts#L263)
This was added essentially to replace the error reporting service, and it does the same thing

packages/sample-controllers/src/sample-gas-prices-controller.ts
```diff
@@ -293,9 +295,9 @@ export class SampleGasPricesController extends BaseController<
    *
    * @param selectedNetworkClientId - The globally selected network client ID.
    */
-  async #onSelectedNetworkClientIdChange(
```
Gudahtt, line 296: Not sure how I feel about this "no misused promises" rule. There wasn't really any misused promise here, it was intentional for this to be async.

I guess it's theoretically possible that a function that is expected to return `void` would break something if it returned a value, like a Promise. I see the value of aligning on the type. Hmm.

Maybe we can consider updating the messenger to allow `Promise<void>` for subscribers as well. That wouldn't solve this for 3rd party libraries though.

Gudahtt: Tempted to suggest that we disable the `checksVoidReturn` option: https://typescript-eslint.io/rules/no-misused-promises/

Mrtenz: My main concern is unhandled promise rejections if the caller doesn't handle them, which won't be the case if the caller expects a void return. Not sure about disabling `checksVoidReturn`.

packages/sample-controllers/src/sample-gas-prices-controller.ts
Mrtenz: I wasn't aware it was intended to replace the error reporting service. Wondering if we should deprecate that then, to make it clear?

Mrtenz: Updated in this commit: 82ce74518de24420cf053f86d7ad0d09c2e0e356.

packages/sample-controllers/src/sample-gas-prices-controller.ts
Gudahtt: Yeah, fair. Especially problematic for mobile, where it would crash the app.

packages/sample-controllers/src/sample-gas-prices-controller.ts
Gudahtt: Definitely overdue to deprecate that package

Mrtenz: Will take care of that tomorrow!

packages/sample-controllers/src/sample-gas-prices-controller.test.ts
```diff
@@ -201,6 +201,52 @@ describe('SampleGasPricesController', () => {
...
+
+        const fetchGasPrices = jest.fn(async (givenChainId) => {
+          if (givenChainId === chainId) {
+            throw new Error('Failed to fetch gas prices');
+          }
+
+          throw new Error(`Unrecognized chain ID '${givenChainId}'`);
+        });
+
+        const captureException = jest.spyOn(rootMessenger, 'captureException');
```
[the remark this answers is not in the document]
Mrtenz: This works as expected.

PR #7420, opened 2025-12-09
Fix getNetworkClientById to be less strict about network client ID

mcmire:
## Explanation

Currently, it is expected that IDs for Infura network clients match a known property in the `InfuraNetworkType` enum and that IDs for custom network clients follow a UUID format.

The way that these expectations are enforced is inconsistent: while an Infura network client ID must match an entry in the aforementioned enum — an error will be produced upon initialization otherwise — there is no such restriction for a custom network client ID.

But `getNetworkClientById` is more strict, and assumes that if given a property of `InfuraNetworkType`, it _must_ correspond to an Infura network client, failing if it cannot find one.

This causes issues with the `transaction-controller` package, because it uses a flattened version of the network client registry to know which networks should be used for tracking transactions. The flattened network client registry holds a list of network client IDs without knowing which are Infura and which are custom. When `transaction-controller` attempts to look them up again, `getNetworkClientById` throws an error. Since this operation occurs when the client starts, this ends up bricking the wallet.

Ideally, we could fix this by enforcing that custom network client IDs do _not_ match `InfuraNetworkType`. However, it is too late to make this change, and anyway, in the future, we want to consolidate Infura and custom network clients, and at that point, the network client ID format won't matter.

For these reasons, this commit changes `getNetworkClientById` to treat an Infura network client ID as a custom network client ID if no Infura network client is found.

## References

This PR unblocks releasing this change in `controller-utils` so that they are no longer breaking changes: https://github.com/MetaMask/core/pull/7067. There is some discussion about this in Slack.

## Checklist

---

> [!NOTE]
> Updates `getNetworkClientById` to fall back to custom clients when an Infura-typed ID isn’t found and standardizes the error to `No network client was found with the ID "<id>"`; adjusts tests and changelog accordingly.
> 
> - **Network Controller**:
>   - Update `getNetworkClientById` in `packages/network-controller/src/NetworkController.ts` to:
>     - Return Infura client if present; otherwise check and return a custom client with the same ID.
>     - Throw a unified error: `No network client was found with the ID "<id>"`.
> - **Tests**:
>   - Adjust tests in `packages/network-controller/tests/NetworkController.test.ts` to reflect fallback behavior and new error message.
>   - Update `packages/assets-controllers/src/AssetsContractControllerWithNetworkClientId.test.ts` to expect the unified error message across ERC-20/721/1155 paths.
> - **Changelog**:
>   - Add Fixed entry in `packages/network-controller/CHANGELOG.md` documenting the `getNetworkClientById` fallback to custom client.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1da819ab5a842e066c460ef99c6f76115c44564. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/AssetsContractControllerWithNetworkClientId.test.ts
+++ b/packages/assets-controllers/src/AssetsContractControllerWithNetworkClientId.test.ts
@@ -31,7 +31,7 @@ describe('AssetsContractController with NetworkClientId', () => {
           'invalidNetworkClientId',
         ),
     ).rejects.toThrow(
-      `No custom network client was found with the ID "invalidNetworkClientId".`,
+      `No network client was found with the ID "invalidNetworkClientId".`,
     );
     messenger.clearEventSubscriptions('NetworkController:networkDidChange');
   });
@@ -46,7 +46,7 @@ describe('AssetsContractController with NetworkClientId', () => {
           'invalidNetworkClientId',
         ),
     ).rejects.toThrow(
-      `No custom network client was found with the ID "invalidNetworkClientId".`,
+      `No network client was found with the ID "invalidNetworkClientId".`,
     );
     messenger.clearEventSubscriptions('NetworkController:networkDidChange');
   });
@@ -154,7 +154,7 @@ describe('AssetsContractController with NetworkClientId', () => {
           undefined,
           'invalidNetworkClientId',
         ),
-    ).rejects.toThrow('No custom network client was found');
+    ).rejects.toThrow('No network client was found');
     messenger.clearEventSubscriptions('NetworkController:networkDidChange');
   });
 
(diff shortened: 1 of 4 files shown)


Gudahtt:
>When `transaction-controller` attempts to look them up again, `getNetworkClientById` throws an error. Since this operation occurs when the client starts, this ends up bricking the wallet.

Sorry I'm not following this scenario.

I understand that `getNetworkClientById` throws an error if given an ID that appears to be an Infura network type but no Infura network client exists. This was how it was supposed to work, the Infura network clients should always exist.

How does that relate to the Transaction Controller? Why does it matter whether the internal registry of clients used there is type-aware or not, why would it need to now the type?

Gudahtt:
Why isn't `monad-testnet` an Infura network client though? I don't understand why we'd want to add it as an Infura network type, but not as a network client. That very distinction is exactly what `InfuraNetworkType` is for.

It still seems to me that adding `monad-testnet` to `InfuraNetworkType` (or _not_ adding it as an Infura network client) is the bug here, not the behavior of `getNetworkClientById`.

PR #7469, opened 2025-12-12
fix: improve add tokens through polling

salimtb:
## Explanation

- Fix `TokenBalancesController` to evaluate `allowExternalServices` dynamically instead of only at constructor time.
  - Previously, the `#balanceFetchers` array was built once in the constructor, so changes to `allowExternalServices()` after initialization were not reflected
  - Now `allowExternalServices()` is stored as a function and evaluated dynamically in the fetcher's `supports` method
- Fix `TokenDetectionController` methods `addDetectedTokensViaPolling` and `addDetectedTokensViaWs` to refresh token metadata cache before use 
  - Previously, these methods used a potentially stale/empty `#tokensChainsCache` from construction time
  - Now they fetch the latest `tokensChainsCache` from `TokenListController:getState` before looking up token metadata
- Add `isOnboarded` constructor option to `TokenBalancesController` and `AccountTrackerController`
  - When `isOnboarded()` returns `false`, balance updates are skipped to prevent unnecessary API calls during onboarding
  - `TokenBalancesController.isActive` now also checks `isOnboarded()` in addition to `isUnlocked`
  - `AccountTrackerController.#refreshAccounts` now checks `isOnboarded()` before fetching balances

Extension UI:  https://github.com/MetaMask/metamask-extension/pull/38841
Mobile UI: https://github.com/MetaMask/metamask-mobile/pull/24005

## Checklist

---

> [!NOTE]
> Introduce onboarding-aware balance fetching, dynamic allowExternalServices checks, and fresh token metadata usage, with refactors and tests across balances and detection controllers.
> 
> - **TokenBalancesController**:
>   - Add `isOnboarded` option; `isActive` now requires unlocked + onboarded.
>   - Always include Accounts API fetcher; `supports()` now evaluates `allowExternalServices()` dynamically.
>   - Preserve original polling intents; improve interval grouping and error handling.
> - **AccountTrackerController**:
>   - Add `isOnboarded` option; skip `refresh`/`syncBalanceWithAddresses` when not onboarded.
>   - Refresh balances for both `from` and `to` addresses on tx events.
>   - Handle `unprocessedChainIds` fallback; tighten typings, privatize `syncAccounts`, safer event handling.
> - **TokenDetectionController**:
>   - Refresh `tokensChainsCache` at call-time for `addDetectedTokensViaPolling`/`addDetectedTokensViaWs`.
>   - Keep RPC detection path; skip when external services disabled/unavailable.
> - **Tests/Changelog**:
>   - Extensive unit tests for onboarding gating, dynamic external-services, fallbacks, and cache refresh.
>   - Update `CHANGELOG.md` with added options and fixes; add mocks tests for Accounts API balances.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb4a07d593a77975e745011ec831f95df18a46ff. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/CHANGELOG.md
+++ b/packages/assets-controllers/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `TokenDetectionController` methods `addDetectedTokensViaPolling` and `addDetectedTokensViaWs` to refresh token metadata cache before use ([#7469](https://github.com/MetaMask/core/pull/7469))
   - Previously, these methods used a potentially stale/empty `#tokensChainsCache` from construction time
   - Now they fetch the latest `tokensChainsCache` from `TokenListController:getState` before looking up token metadata
+- Fix `AccountTrackerController.syncBalanceWithAddresses` to also check `isOnboarded()` before fetching balances ([#7469](https://github.com/MetaMask/core/pull/7469))
+  - Previously, only `#refreshAccounts` checked `isOnboarded()`, but `syncBalanceWithAddresses` would still make RPC calls during onboarding
+  - Now `syncBalanceWithAddresses` returns an empty object when `isOnboarded()` returns `false`
 
 ## [94.0.0]
 
--- a/packages/assets-controllers/src/AccountTrackerController.test.ts
+++ b/packages/assets-controllers/src/AccountTrackerController.test.ts
@@ -1634,6 +1634,71 @@ describe('AccountTrackerController', () => {
         },
       );
     });
+
+    it('should skip balance fetching when isOnboarded returns false', async () => {
+      await withController(
+        {
+          options: { isOnboarded: () => false },
+          isMultiAccountBalancesEnabled: true,
+          selectedAccount: ACCOUNT_1,
+          listAccounts: [],
+        },
+        async ({ controller }) => {
+          // Reset query mock to track calls
+          mockedQuery.mockClear();
+
+          const result = await controller.syncBalanceWithAddresses([
+            ADDRESS_1,
+            ADDRESS_2,
+          ]);
+
+          // Should return empty object without making any RPC calls
+          expect(result).toStrictEqual({});
+
+          // Verify no RPC calls were made (query should not have been called for getBalance)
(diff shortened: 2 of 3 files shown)


packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -333,14 +335,14 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
...
+    this.#allowExternalServices = allowExternalServices;
     this.#defaultInterval = interval;
     this.#websocketActivePollingInterval = websocketActivePollingInterval;
     this.#chainPollingConfig = { ...chainPollingIntervals };
 
+    // Always include AccountsApiFetcher - it dynamically checks allowExternalServices() in supports()
     this.#balanceFetchers = [
-      ...(accountsApiChainIds().length > 0 && allowExternalServices()
-        ? [this.#createAccountsApiFetcher()]
-        : []),
```
salimtb, line 343: we should not do this at the constructor level , this is init only when the worker restart , the new approach will have it changing even when the worker doesn't restart

packages/assets-controllers/src/TokenBalancesController.ts
```diff
@@ -537,7 +539,9 @@ export class TokenBalancesController extends StaticIntervalPollingController<{
     );
 
     return {
+      // Dynamically check allowExternalServices() at call time, not just at construction time
       supports: (chainId: ChainIdHex): boolean =>
+        this.#allowExternalServices() &&
```
salimtb, line 550: we consider chains not supported for api fetch if basic func is off

packages/assets-controllers/src/TokenDetectionController.ts
```diff
@@ -817,6 +817,13 @@ export class TokenDetectionController extends StaticIntervalPollingController<To
       return;
     }
 
+    // Refresh the token cache to ensure we have the latest token metadata
+    // This fixes a bug where the cache from construction time could be stale/empty
+    const { tokensChainsCache } = this.messenger.call(
+      'TokenListController:getState',
+    );
```
salimtb, line 824: same here , we should fetch this value from the state to ensure we get last updated data

PR #7503, opened 2025-12-15
feat: Attach metadata when submitting a revocation to the permission provider snap

V00D00-child:
## Explanation

This PR extends the `GatorPermissionsController` to pass the hash of the transaction that revoked a permission (if available).

## References

Requires(gator snap): [feat: Store metadata when revoking a permission](https://github.com/MetaMask/snap-7715-permissions/pull/228)
Required by(MM client): [chore: Bump @metamask/gator-permissions-controller to 0.9.0](https://github.com/MetaMask/metamask-extension/pull/39076)

## Checklist

---

> [!NOTE]
> Introduces tx-aware revocations and related type/state updates.
> 
> - **BREAKING:** `submitRevocation` now sends `txHash` via `RevocationParams`; stored permissions may include `revocationMetadata` (exported types updated) in `types.ts` and `index.ts`
> - GatorPermissionsController: listens for `TransactionStatus`, skips submit on non-confirmed/failed, passes confirmed `hash` to `permissionsProvider_submitRevocation`, always refreshes permissions after terminal states
> - Tests expanded to cover confirmed/failed paths and metadata propagation; minor typing/utility refinements in `decodePermission/utils` (explicit return types, generics)
> - Changelog updated; bumps `@metamask/transaction-controller`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7183271566e03ccd07a2de8f545f91e801001203. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/gator-permissions-controller/CHANGELOG.md
+++ b/packages/gator-permissions-controller/CHANGELOG.md
@@ -9,8 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **BREAKING:** Extends the `GatorPermissionsController` to enable attaching metadata(transaction hash) when submitting a permission revocation. ([#7503](https://github.com/MetaMask/core/pull/7503))
-  - `submitRevocation` revocationParams object now requires a `revocationMetadata` property.
+- **BREAKING:** Now passes the hash of the transaction that revoked the permission, when invoking the Gator Permissions Snap to mark the stored permission as revoked. ([#7503](https://github.com/MetaMask/core/pull/7503))
 
 ### Changed
 


packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -948,9 +951,34 @@ export default class GatorPermissionsController extends BaseController<
...
+          txHash: transactionMeta.hash,
         });
 
-        this.submitRevocation({ permissionContext })
+        // Attach metadata by parsing the confirmed transactionMeta
+        let metadata: RevocationMetadata | undefined;
+        const { hash, blockTimestamp } = transactionMeta;
+        if (hash === undefined || blockTimestamp === undefined) {
+          controllerLog(
+            'Failed to attach transaction has after revocation transaction confirmed',
```
jeffsmale90:
```suggestion
            'Failed to attach transaction hash after revocation transaction confirmed',
```

packages/gator-permissions-controller/src/types.ts
```diff
@@ -237,6 +237,14 @@ export type DelegationDetails = Pick<
   'caveats' | 'delegator' | 'delegate' | 'authority'
 >;
 
+/**
+ * Represents the metadata for confirmed transaction revocation.
+ */
+export type RevocationMetadata = {
+  txHash: Hex;
+  blockTimestamp: string;
```
jeffsmale90: is blockTimestamp a decimal string? 

If that's the case, any reason not to represent it as a number? As is, it's a bit ambiguous.

packages/gator-permissions-controller/src/types.ts
```diff
@@ -245,6 +253,11 @@ export type RevocationParams = {
    * The permission context as a hex string that identifies the permission to revoke.
    */
   permissionContext: Hex;
+
+  /**
+   * The metadata associated with the permission revocation transaction.
+   */
+  metadata?: RevocationMetadata;
```
jeffsmale90: Shouldn't this be required? Is there a scenario where we want to persist the revocation status, but don't have revocation metadata?

As per other comments, I think we should name this explicitly `revocationMetadata` to make it clear to what it is pertaining.

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -948,9 +951,34 @@ export default class GatorPermissionsController extends BaseController<
...
+        } else {
+          metadata = {
+            txHash: hash as Hex,
+            blockTimestamp,
+          };
+        }
+
+        const revocationParams = metadata
+          ? { permissionContext, metadata }
+          : { permissionContext };
```
jeffsmale90: As per previous comment - should we accept revocations where we don't have the metadata? In what scenario can this happen?

jeffsmale90 (commented):
Should we also add `revocationMetadata` to `StoredGatorPermission` (and as per my comments on the related PR, perhaps remove the `isRevoked` flag?

This would probably require some changes to how we load the permission from storage also, but IMO worthwhile,

packages/gator-permissions-controller/src/types.ts
V00D00-child: There is an edge case in extension where we call submitRevocation with only permissionContext. In this case, we would not pass any revocation metadata, so we have to make that value optional. 

Here is a link to the code for more context: https://github.com/MetaMask/metamask-extension/blob/main/ui/hooks/gator-permissions/useRevokeGatorPermissions.ts#L176

Edge case: Before submitting a revocation transaction, we check whether the delegation is already disabled on-chain. If it is disabled, we don't submit a transaction; instead, we pass the permissionContext to the Snap revocation method.

packages/gator-permissions-controller/src/GatorPermissionsController.ts
V00D00-child: In this case, where we are in the context of a handler that runs when a transaction is confirmed, we should accept revocations even when we don't have the metadata. Not sure why the `hash` on `TransactionMeta` type coming from `transaction-controller` is marked as optional, but I would expect it to be present when the transaction is confirmed.

packages/gator-permissions-controller/src/types.ts
V00D00-child: I dropped the block timestamp from the metadata. Seems unnecessary to pass this data to our Snap to store in profile sync when we can always fetch using the transaction hash.

jeffsmale90: It _might_ be interesting to include the timestamp of when it was marked as revoked - especially if there's no hash 🤷

packages/gator-permissions-controller/src/types.ts
jeffsmale90: That makes sense - I still think it makes more sense to set the `revocationMetadata`, but just make the `txHash` property optional. It reduces the data redundancy, making it more difficult to get inconsistent data, or to mis-interpret the data.

packages/gator-permissions-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Extends the `GatorPermissionsController` to enable attaching metadata(transaction hash, block timestamp) when submitting a permission revocation. ([#7503](https://github.com/MetaMask/core/pull/7503))
```
mj-kiwi: Please make sure to update the CHANGELOG accordingly, as the implementation does not include a block timestamp.

V00D00-child: Nice call out; I updated the CHANGELOG to reflect the current implementation

packages/gator-permissions-controller/src/types.ts
V00D00-child: Updated the `RevocationParams` to make `revocationMetadata` required and made `txHash` optional on `RevocationMetadata`.

V00D00-child: Making revocationMetadata required on `RevocationParams` will introduce a breaking change to the `submitRevocation()` interface. I will open a daft PR on the MM client to update calls to `submitRevocation()` to match the new interface.

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -948,9 +951,28 @@ export default class GatorPermissionsController extends BaseController<
         controllerLog('Transaction confirmed, submitting revocation', {
```
jeffsmale90, line 945: not strictly related to this PR - but do we need to check `transactionMeta.status` here? 

When a transaction is _confirmed_, do we explicitly guard against failed transactions?

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -948,9 +951,28 @@ export default class GatorPermissionsController extends BaseController<
...
+              txId,
+              permissionContext,
+              error: new Error(
+                'Confirmed transaction is missing transaction hash',
+              ),
+            },
+          );
+        } else {
+          revocationMetadata.txHash = hash as Hex;
+        }
```
jeffsmale90: nit: you could simplify this - moving the assignment out of the else block, and creating the revocationMetadata in a single statement:


```suggestion
        const { hash } = transactionMeta;
        
        const revocationMetadata: RevocationMetadata = {
          txHash: hash as Hex | undefined,
        };
        
        if (hash === undefined) {
          controllerLog(
            'Failed to attach transaction hash after revocation transaction confirmed',
            {
              txId,
              permissionContext,
              error: new Error(
                'Confirmed transaction is missing transaction hash',
              ),
            },
          );
        }
```

jeffsmale90 (approved):
This included a couple of linting changes, so I suggest you review by changeset.

PR #7511, opened 2025-12-16
feat: Make A/B tests independent

MarioAslau:
## Explanation

Right now, only the ID is used as a seed to determine what bucket the user falls into in an A/B test.

Due to this, if there’s more than 1 A/B test configured, they will never be independent of each other

**Solution:**
Optimizes threshold-based feature flag processing to ensure independent user assignment across different flags. Define the distribution using profile/metametrics ID + A/B test ID as a seed instead of just profile/metametrics ID

**Key changes:**
- Use sha256 from @metamask/utils@11.9.0 to generate unique threshold per flag (native crypto.subtle.digest - 10-100x faster)
- Persist threshold values in controller state to avoid recalculating on app restart
- Skip cryptographic operations for non-threshold arrays
- Batch cache updates and cleanup into single state change
- Automatically remove stale cache entries when flags are deleted
- Remove @noble/hashes dependency (now available in @metamask/utils)

## Checklist

---

> [!NOTE]
> Implements per-flag deterministic A/B assignment with cached thresholds in the Remote Feature Flag Controller and upgrades @metamask/utils to ^11.9.0 repo-wide.
> 
> - **Remote Feature Flag Controller**:
>   - Implement per-flag threshold selection using `sha256(metaMetricsId + flagName)` via `@metamask/utils@^11.9.0`.
>   - Add `thresholdCache` to state; batch cache updates and clean up stale entries; skip processing for non-threshold arrays.
>   - Update tests and expectations; remove reliance on external hashing utility.
> - **Dependencies**:
>   - Upgrade `@metamask/utils` to `^11.9.0` across packages (yarn.lock updated).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 297369a476c8a5608c694273498aa1b9f0f176c9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/remote-feature-flag-controller/CHANGELOG.md
+++ b/packages/remote-feature-flag-controller/CHANGELOG.md
@@ -9,15 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511)) for native `crypto.subtle.digest` optimization ([#7511](https://github.com/MetaMask/core/pull/7511))
-- Remove `@noble/hashes` dependency since hashing utilities are now available in upgraded `@metamask/utils` ([#7511](https://github.com/MetaMask/core/pull/7511))
-- Changes to exported types ([#7511](https://github.com/MetaMask/core/pull/7511)):
-  - Add optional field `thresholdCache` to `RemoteFeatureFlagControllerState`
-- Optimize threshold-based feature flag processing to ensure independent user assignment across different flags ([#7511](https://github.com/MetaMask/core/pull/7511)):
+- **BREAKING:** Improve threshold-based feature flag processing to ensure independent user assignment across different flags ([#7511](https://github.com/MetaMask/core/pull/7511)):
   - Persist threshold values in controller state to avoid recalculating on app restart
   - Skip cryptographic operations for non-threshold arrays
   - Batch cache updates and cleanup into single state change
   - Automatically remove stale cache entries when flags are deleted
+- Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511)) for native `crypto.subtle.digest` optimization ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Remove `@noble/hashes` dependency since hashing utilities are now available in upgraded `@metamask/utils` ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Changes to exported types ([#7511](https://github.com/MetaMask/core/pull/7511)):
+  - Add optional field `thresholdCache` to `RemoteFeatureFlagControllerState`
 - Bump `@metamask/controller-utils` from `^11.16.0` to `^11.17.0` ([#7534](https://github.com/MetaMask/core/pull/7534))
 
 ## [3.1.0]


packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -24,6 +26,19 @@ import { isVersionFeatureFlag, getVersionData } from './utils/version';
 export const controllerName = 'RemoteFeatureFlagController';
 export const DEFAULT_CACHE_DURATION = 24 * 60 * 60 * 1000; // 1 day
 
+/**
```
asalsys: add this function in the user-segmentation-utils with the generateDeterministicRandomNumber

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -632,10 +634,11 @@ describe('RemoteFeatureFlagController', () => {
...
         controller.state.remoteFeatureFlags;
       // Should select 13.1.0 version and then apply A/B testing to that array
-      // With MOCK_METRICS_ID threshold, should select groupC (threshold 1.0)
+      // With MOCK_METRICS_ID + 'multiVersionABFlag' hashed:
+      // Threshold = 0.094878, which falls in groupA range (t <= 0.3)
       expect(multiVersionABFlag).toStrictEqual({
-        name: 'groupC',
-        value: { feature: 'C', enabled: true },
+        name: 'groupA',
+        value: { feature: 'A', enabled: true },
```
asalsys, line 788: add a test for createDeterministicSeed is deterministic
and a test for test processRemoteFeatureFlags adding you to different groups when you have different flags and same group when you have the same flag

asalsys (changes requested):

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
```diff
@@ -19,6 +21,29 @@ const MIN_UUID_V4_BIGINT = uuidStringToBigInt(MIN_UUID_V4);
...
+export function createDeterministicSeed(
+  metaMetricsId: string,
+  featureFlagName: string,
+): string {
+  if (!metaMetricsId) {
+    throw new Error('MetaMetrics ID cannot be empty');
+  }
+
+  const seed = metaMetricsId + featureFlagName;
+  const hashBuffer = sha256(seed);
```
FrederikBolding: If async is OK here we should consider using `crypto.subtle.digest`

FrederikBolding: Just upstreamed [this](https://github.com/MetaMask/utils/blob/main/src/hashing.ts#L11) which may be useful here

MarioAslau: @FrederikBolding can you expand on that ?

I've used sha256 from noble/hashes to keep consistency with what I've noticed in other packages.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,7 +291,12 @@ export class RemoteFeatureFlagController extends BaseController<
         continue;
       }
 
-      if (Array.isArray(processedValue) && thresholdValue) {
+      if (Array.isArray(processedValue)) {
+        const deterministicSeed = createDeterministicSeed(
+          metaMetricsId + remoteFeatureFlagName,
```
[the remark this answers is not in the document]
MarioAslau: outdated

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
MarioAslau: It doesn't seem to be published/exported yet in the current version of metamask/utils

asalsys (approved):

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
FrederikBolding: It was published earlier today!

FrederikBolding: SHA-256 using native code will perform better across mobile and extension! We automatically fall back to noble if unavailable

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,7 +291,13 @@ export class RemoteFeatureFlagController extends BaseController<
         continue;
       }
 
-      if (Array.isArray(processedValue) && thresholdValue) {
+      if (Array.isArray(processedValue)) {
+        const deterministicSeed = createDeterministicSeed(
```
Gudahtt: This is a slow cryptographic operation and it runs once per flag, but we don't need to run this for all flags. It's unnecessary unless the flag has a threshold.

Perhaps we can calculate this later, after we've determined that there is a threshold on this specific flag? That way it can be skipped for non-threshold flags, which is the majority of them.

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
```diff
@@ -19,6 +21,33 @@ const MIN_UUID_V4_BIGINT = uuidStringToBigInt(MIN_UUID_V4);
...
 const UUID_V4_VALUE_RANGE_BIGINT = MAX_UUID_V4_BIGINT - MIN_UUID_V4_BIGINT;
 
+/**
+ * Creates a deterministic hex ID by hashing the input seed.
+ * This ensures consistent group assignment for A/B testing across different feature flags.
+ * Normalizes inputs to lowercase to ensure case-insensitive consistency.
+ *
+ * @param metaMetricsId - The user's MetaMetrics ID (must be non-empty)
+ * @param featureFlagName - The feature flag name to create unique seed per flag
+ * @returns A hex string with '0x' prefix suitable for generateDeterministicRandomNumber
```
Gudahtt: The `0x`-prefixed input for `generateDeterministicRandomNumber` is meant to support legacy metrics IDs. We had hoped to remove that functionality when dropping support for those legacy IDs. Migrating to a format we planned to deprecate is not ideal.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Gudahtt: It's also not ideal to run this every time flags get updated 🤔.

asalsys (changes requested):
Add a change description of the change to the remote-feature-flag-controller changelog

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
asalsys: @Gudahtt To the first point that is made: Array.isArray(processedValue) is filtering out the threshold flags because it is understood that the Array is an array of options that is processed to a single option after the threshold value is calculated and used to select the correct option.

packages/remote-feature-flag-controller/package.json
```diff
@@ -51,7 +51,8 @@
     "@metamask/base-controller": "^9.0.0",
     "@metamask/controller-utils": "^11.16.0",
     "@metamask/messenger": "^0.3.0",
-    "@metamask/utils": "^11.8.1",
+    "@metamask/utils": "^11.9.0",
+    "@noble/hashes": "^1.8.0",
```
Gudahtt: Is this still being used?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -150,6 +150,8 @@ export class RemoteFeatureFlagController extends BaseController<
 
   readonly #clientVersion: SemVerVersion;
 
+  readonly #thresholdCache: Map<string, number> = new Map();
```
Gudahtt: Nit: We could use a more specific type here

```suggestion
  readonly #thresholdCache: Map<`${string}:${string}`, number> = new Map();
```

Gudahtt: This will cache the values for a session, but not across sessions. We'll be regenerating these values every app start.

Maybe we can put them in controller state and persist them instead?

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
```diff
@@ -19,6 +20,43 @@ const MIN_UUID_V4_BIGINT = uuidStringToBigInt(MIN_UUID_V4);
...
+  // Normalize to lowercase to ensure case-insensitive consistency
+  const normalizedId = metaMetricsId.toLowerCase();
+  const normalizedFlagName = featureFlagName.toLowerCase();
+  const seed = normalizedId + normalizedFlagName;
+
+  // Hash the combined seed
+  const encoder = new TextEncoder();
+  const hashBuffer = await sha256(encoder.encode(seed));
+
+  // Convert hash bytes directly to 0-1 range without intermediate hex format
```
Gudahtt: I get what you were trying to say 😅 The intermediary format+conversion we used before is gone. But this is a little confusing to say because we are still converting the hash to a hex number to get a number between 0-1.

```suggestion
  // Convert hash bytes directly to 0-1 range
```

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
```diff
@@ -19,6 +20,43 @@ const MIN_UUID_V4_BIGINT = uuidStringToBigInt(MIN_UUID_V4);
...
+  featureFlagName: string,
+): Promise<number> {
+  if (!metaMetricsId) {
+    throw new Error('MetaMetrics ID cannot be empty');
+  }
+
+  // Normalize to lowercase to ensure case-insensitive consistency
+  const normalizedId = metaMetricsId.toLowerCase();
+  const normalizedFlagName = featureFlagName.toLowerCase();
+  const seed = normalizedId + normalizedFlagName;
```
Gudahtt: Nit: Is this step necessary? Do we expect to get the same ID in alternating mixed/lower case? That would be surprising.

```suggestion
  const normalizedFlagName = featureFlagName.toLowerCase();
  const seed = metaMetricsId + normalizedFlagName;
```

yarn.lock
```diff
@@ -5180,6 +5181,25 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@metamask/utils@npm:^11.9.0":
```
Gudahtt: Looks like you need to run `yarn dedupe`

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Gudahtt: >it is understood that the Array is an array of options that is processed to a single option after the threshold value is calculated and used to select the correct option.

Ah I see. I don't think that is a valid assumption to make here though.

Flags can have array values even if they have no thresholds set.

packages/remote-feature-flag-controller/package.json
MarioAslau: removed!

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +292,37 @@ export class RemoteFeatureFlagController extends BaseController<
...
+      if (Array.isArray(processedValue)) {
+        // Validate array has valid threshold items before doing expensive crypto operation
+        const hasValidThresholds = processedValue.some(
+          isFeatureFlagWithScopeValue,
+        );
+
+        if (!hasValidThresholds) {
+          // Skip this flag - no valid threshold configuration
+          continue;
+        }
```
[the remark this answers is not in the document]
MarioAslau: This concern has been addressed

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -273,12 +281,45 @@ export class RemoteFeatureFlagController extends BaseController<
...
+   * @param currentFlagNames - Array of flag names currently in the remote response
+   */
+  #cleanupStaleThresholdCache(currentFlagNames: string[]): void {
+    const metaMetricsId = this.#getMetaMetricsId();
+    const staleCacheKeys: string[] = [];
+
+    // Find cache keys that don't match any current flag
+    for (const cacheKey of Object.keys(this.state.thresholdCache)) {
+      // Cache key format: "metaMetricsId:flagName"
+      const [cachedMetaMetricsId, cachedFlagName] = cacheKey.split(':');
```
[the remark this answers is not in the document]
Gudahtt: True. could do something like this instead:

```suggestion
      const [cachedMetaMetricsId, ...cachedFlagNameParts] = cacheKey.split(':');
      const cachedFlagName = cachedFlagNameParts.join(':');
```

packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts
```diff
@@ -19,6 +20,40 @@ const MIN_UUID_V4_BIGINT = uuidStringToBigInt(MIN_UUID_V4);
...
+export async function calculateThresholdForFlag(
+  metaMetricsId: string,
+  featureFlagName: string,
+): Promise<number> {
+  if (!metaMetricsId) {
+    throw new Error('MetaMetrics ID cannot be empty');
+  }
+
+  const normalizedFlagName = featureFlagName.toLowerCase();
+  const seed = metaMetricsId + normalizedFlagName;
```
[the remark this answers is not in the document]
Gudahtt: May as well ignore or error upon encountering an empty flag name I guess 🤔 I doubt this would be allowed by launch darkly though

packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -7,10 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add persistent threshold caching to avoid recalculating SHA-256 hashes on every app restart
```
Gudahtt: These entries are missing the PR number+link (e.g. `([#7511](https://github.com/MetaMask/core/pull/7511))`)

packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -7,10 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add `calculateThresholdForFlag` utility function that directly converts hash to 0-1 threshold value
+- Add automatic cleanup mechanism to remove stale threshold cache entries for deleted flags
+
 ### Changed
 
+- Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` for native `crypto.subtle.digest` optimization
```
Gudahtt: We'll need to duplicate this in the changelog of each package where it was updated (i.e. most of them)

packages/remote-feature-flag-controller/CHANGELOG.md
Gudahtt: Most of these entries don't seem relevant. e.g. this one talks about a problem that never existed in the published package ("recalculcating SHA-256 hashes on every app restart"), and the next talks about an internal function, and the next again talks about a feature to remove caches that never existed before this release.

The changes here should be for an audience of _users of this package_, and should be relative to the last published version.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -58,6 +59,12 @@ const remoteFeatureFlagControllerMetadata = {
     includeInDebugSnapshot: true,
     usedInUi: false,
   },
+  thresholdCache: {
+    includeInStateLogs: false,
```
Gudahtt, line 63: Nit: Usually we default this to true unless there's a specific reason to omit it.

Not a big deal, as I don't see this as being _likely_ useful, but it won't hurt to collect either.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +332,43 @@ export class RemoteFeatureFlagController extends BaseController<
...
+        );
+
+        if (!hasValidThresholds) {
+          // Not a threshold array - preserve as-is
+          processedRemoteFeatureFlags[remoteFeatureFlagName] = processedValue;
+          continue;
+        }
+
+        // Check cache first, calculate only if needed
+        const cacheKey: `${string}:${string}` = `${metaMetricsId}:${remoteFeatureFlagName}`;
```
Gudahtt: Nit: This should also work:
```suggestion
        const cacheKey = `${metaMetricsId}:${remoteFeatureFlagName}` as const;
```

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +332,43 @@ export class RemoteFeatureFlagController extends BaseController<
...
+        let thresholdValue = this.state.thresholdCache[cacheKey];
+
+        if (thresholdValue === undefined) {
+          // eslint-disable-next-line require-atomic-updates
```
Gudahtt: Nit: We should explain why this is safe (it does seem safe in this case)

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +332,43 @@ export class RemoteFeatureFlagController extends BaseController<
...
+          thresholdValue = await calculateThresholdForFlag(
+            metaMetricsId,
+            remoteFeatureFlagName,
+          );
+
+          // Update state to persist the threshold
+          this.update((state) => {
```
Gudahtt: Nit: This will create one state change event per update, which could mean dozens of changes in a short time frame. It'll then change _again_ when we clear out stale flags.

To reduce the frequency of these state change events, we should be able to consolidate this into a single update (with all new thresholds added, and new ones removed).

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +332,43 @@ export class RemoteFeatureFlagController extends BaseController<
...
+          continue;
+        }
+
+        // Check cache first, calculate only if needed
+        const cacheKey: `${string}:${string}` = `${metaMetricsId}:${remoteFeatureFlagName}`;
+        let thresholdValue = this.state.thresholdCache[cacheKey];
+
+        if (thresholdValue === undefined) {
+          // eslint-disable-next-line require-atomic-updates
+          thresholdValue = await calculateThresholdForFlag(
```
Gudahtt, line 352: Nit: We should be able to calculate these thresholds in parallel, rather than one-at-a-time. Given that there is this async step, it's possible this could significantly speed up the operation

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
MarioAslau: I've updated the check to reflect this! This problem is now solved

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
MarioAslau: we are now batching udpdates

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
MarioAslau: the comment is no longer needed after refactor

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -291,14 +326,40 @@ export class RemoteFeatureFlagController extends BaseController<
...
+        );
+
+        if (!hasValidThresholds) {
+          // Not a threshold array - preserve as-is
+          processedFlags[remoteFeatureFlagName] = processedValue;
+          continue;
+        }
+
+        // Check cache first, calculate only if needed
+        const cacheKey = `${metaMetricsId}:${remoteFeatureFlagName}` as const;
```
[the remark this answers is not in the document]
MarioAslau: outdated

packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511)) for native `crypto.subtle.digest` optimization ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Remove `@noble/hashes` dependency since hashing utilities are now available in upgraded `@metamask/utils` ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Changes to exported types ([#7511](https://github.com/MetaMask/core/pull/7511)):
+  - Add optional field `thresholdCache` to `RemoteFeatureFlagControllerState`
+- Optimize threshold-based feature flag processing to ensure independent user assignment across different flags ([#7511](https://github.com/MetaMask/core/pull/7511)):
+  - Persist threshold values in controller state to avoid recalculating on app restart
+  - Skip cryptographic operations for non-threshold arrays
+  - Batch cache updates and cleanup into single state change
+  - Automatically remove stale cache entries when flags are deleted
```
Gudahtt: I've added the `BREAKING` prefix and moved this one to the top.

Also a minor nit, I replaced "Optimize" with "Improve" because to me optimize means "improve performance", and this will not improve performance (though the negative impact has been very effectively mitigated). This is more of a functional improvement

```suggestion
- **BREAKING:** Improve threshold-based feature flag processing to ensure independent user assignment across different flags ([#7511](https://github.com/MetaMask/core/pull/7511)):
  - Persist threshold values in controller state to avoid recalculating on app restart
  - Skip cryptographic operations for non-threshold arrays
  - Batch cache updates and cleanup into single state change
  - Automatically remove stale cache entries when flags are deleted
- Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511)) for native `crypto.subtle.digest` optimization ([#7511](https://github.com/MetaMask/core/pull/7511))
- Remove `@noble/hashes` dependency since hashing utilities are now available in upgraded `@metamask/utils` ([#7511](https://github.com/MetaMask/core/pull/7511))
- Changes to exported types ([#7511](https://github.com/MetaMask/core/pull/7511)):
  - Add optional field `thresholdCache` to `RemoteFeatureFlagControllerState`
```

Gudahtt (approved):
LGTM!

PR #7514, opened 2025-12-16
ci: Skip merge queue if pull request is up-to-date

Mrtenz:
## Explanation

This adds a new action which determines whether a pull request in the merge queue is up-to-date, meaning:

- The pull request is based on the latest commit on `main`.
- The pull request is the first in the merge queue.

In this case, all status checks have already passed on the branch, and running in the merge queue would be redundant, meaning we can skip the merge queue checks.

To accomplish this, I've added an output to the action (`up-to-date`) which we can use in all steps in the main workflow to determine if they should run. If `up-to-date` is `true`, all jobs except `all-jobs-pass` will be skipped, and the pull request will be merged much quicker.

## References

WPC-184

## Checklist

---

> [!NOTE]
> Adds a composite action to detect up-to-date PRs in the merge queue and gates workflow jobs to skip queue checks when eligible.
> 
> - **CI**:
>   - **New composite action** `/.github/actions/check-merge-queue`:
>     - Extracts PR number from `head-ref`, queries GraphQL for `headRefName` and `mergeQueueEntry.position`, and outputs `up-to-date` when branch is based on latest `base-ref` and queue position is `1`.
>   - **Workflow updates** `/.github/workflows/main.yml`:
>     - Adds `check-skip-merge-queue` job (merge_group only) exposing `skip-merge-queue`.
>     - Gates `check-workflows` with conditional `needs`/`if` to skip when `skip-merge-queue` is `true`.
>     - Updates `all-jobs-pass` to depend on `check-skip-merge-queue` and succeed if either all jobs passed or skipping is allowed via `PASSED` env.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ceccfa8f59158f2701f3443f3f70f8f55e0fe766. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,15 +13,16 @@ concurrency:
 jobs:
   check-skip-merge-queue:
     name: Check if pull request can skip merge queue
-    if: github.event_name == 'merge_group'
     runs-on: ubuntu-latest
     outputs:
       skip-merge-queue: ${{ steps.check-skip-merge-queue.outputs.up-to-date }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
+        if: github.event_name == 'merge_group'
       - name: Check pull request merge queue status
         id: check-skip-merge-queue
+        if: github.event_name == 'merge_group'
         uses: ./.github/actions/check-merge-queue
         with:
           head-ref: ${{ github.event.merge_group.head_ref }}
@@ -32,7 +33,7 @@ jobs:
     name: Check workflows
     needs:
       - check-skip-merge-queue
-    if: always() && github.event_name != 'merge_group' || needs.check-skip-merge-queue.outputs.skip-merge-queue != 'true'
+    if: github.event_name != 'merge_group' || needs.check-skip-merge-queue.outputs.skip-merge-queue != 'true'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5


.github/workflows/main.yml
```diff
@@ -11,8 +11,32 @@ concurrency:
...
+        id: check-skip-merge-queue
+        if: github.event_name == 'merge_group'
+        uses: ./.github/actions/check-merge-queue
+        with:
+          head-ref: ${{ github.event.merge_group.head_ref }}
+          base-ref: main
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set skip-merge-queue output to false
+        if: github.event_name != 'merge_group'
+        run: echo "skip-merge-queue=false" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
Mrtenz: Looks like this can just be removed.

.github/workflows/main.yml
```diff
@@ -11,8 +11,29 @@ concurrency:
...
 jobs:
+  check-skip-merge-queue:
+    name: Check if pull request can skip merge queue
+    runs-on: ubuntu-latest
+    outputs:
+      skip-merge-queue: ${{ steps.check-skip-merge-queue.outputs.up-to-date }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        if: github.event_name == 'merge_group'
```
Gudahtt, line 22: Nit: can we move this condition to the job level so that we don't need to repeat it twice?

Mrtenz: Unfortunately not, that would skip `all-jobs-pass` when not running in a merge queue.

Gudahtt: Ah I see, yeah I missed that this was listed under `needs` for other steps.

`all-jobs-pass` doesn't need this to run because it has an `always()` condition, but the other steps do.

It's unfortunate to have to spin up an empty VM each time and do nothing with it. Ah well, it works at least.

Mrtenz: At least we're no longer spinning up a bunch of other VMs unnecessarily 😅

.github/workflows/main.yml
```diff
@@ -11,8 +11,29 @@ concurrency:
...
+        with:
+          head-ref: ${{ github.event.merge_group.head_ref }}
+          base-ref: main
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
   check-workflows:
     name: Check workflows
+    needs:
+      - check-skip-merge-queue
+    if: github.event_name != 'merge_group' || needs.check-skip-merge-queue.outputs.skip-merge-queue != 'true'
```
Gudahtt: Nit: I believe this would let us skip `check-skip-merge-queue` completely on non-merge-queue triggers:


```suggestion
    if: ${{ always() && github.event_name != 'merge_group' || needs.check-skip-merge-queue.outputs.skip-merge-queue != 'true' }}
```

We'd also have to remove `check-skip-merge-queue` from other jobs that list it under `needs` (except `all-jobs-pass` which also uses `always()`), but the fact that they depend on `check-workflows` should suffice for ensuring they don't run when we want to skip them.

.github/workflows/main.yml
Gudahtt: True! We may be able to do better though, suggestion about it here: https://github.com/MetaMask/core/pull/7514/files#r2624541043

.github/workflows/main.yml
Mrtenz: Good point, pushed here: 3b2eb1daa66823d1ec2d2e9c39e02bd6c25e6f97.

Wondering if `always()` is necessary? 🤔

Gudahtt: Oh sorry, to clarify, the `always()` suggestion was meant to let us move the `if` condition on `check-skip-merge-queue` to the job-level (which would eliminate the "spin up VM then shut it down again" step).

It's not needed in https://github.com/MetaMask/core/commit/3b2eb1daa66823d1ec2d2e9c39e02bd6c25e6f97 but it would be if we made that additional change to `check-skip-merge-queue`

.github/workflows/main.yml
```diff
@@ -11,8 +11,29 @@ concurrency:
...
+        uses: actions/checkout@v6
+        if: github.event_name == 'merge_group'
+      - name: Check pull request merge queue status
+        id: check-skip-merge-queue
+        if: github.event_name == 'merge_group'
+        uses: ./.github/actions/check-merge-queue
+        with:
+          head-ref: ${{ github.event.merge_group.head_ref }}
+          base-ref: main
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
Gudahtt, line 30: Nit: Given that this is a secret, we can pass it as a secret instead:

```suggestion
        secrets:
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

We don't want GitHub to treat this as non-sensitive, and log it in the workflow logs or something.

This would require changes to the check-merge-queue action as well, to change it from an input to a secret

cryptodev-2s: Good catch, we may want to update other actions to use this pattern too

.github/workflows/main.yml
Mrtenz: Ah, I see how that works now. Updated here: 285ce9bf9890119c162b9de0e4ff087907c40d94

.github/workflows/main.yml
Mrtenz: This is one of the confusing differences between actions and workflows, but actions don't accept secrets, and using inputs is actually the recommended way to provide secrets to actions. GitHub will still treat them as sensitive values.

For reference, from [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idsecrets):

> When a job is used to call a reusable workflow, you can use `secrets` to provide a map of secrets that are passed to the called workflow.

Notice how there's no `jobs.<job_id>.steps[*].secrets`, and `jobs.<job_id>.uses` is only valid for workflows.

.github/workflows/main.yml
Mrtenz: Hmm, [looks like GitHub skips everything](https://github.com/MetaMask/core/actions/runs/20282342613/job/58247371269?pr=7514) 🤔

Mrtenz: I don't think we should add `always()` to all jobs. I will revert this change unless you have a better idea?

.github/actions/check-merge-queue/action.yml
```diff
@@ -0,0 +1,91 @@
...
+          });
+
+          if (!result.repository.pullRequest) {
+            return core.setFailed(`Pull request #${number} not found in repository "${context.repo.owner}/${context.repo.repo}".`);
+          }
+
+          const position = result.repository.pullRequest.mergeQueueEntry?.position;
+          if (!position) {
+            return core.setFailed(`Pull request #${number} is not in the merge queue.`);
+          }
```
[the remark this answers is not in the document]
Mrtenz: The API is not 0-indexed.

.github/actions/check-merge-queue/action.yml
```diff
@@ -0,0 +1,91 @@
...
+    - name: Check if pull request is up-to-date with base branch
+      id: up-to-date
+      shell: bash
+      env:
+        BASE_REF: ${{ inputs.base-ref }}
+        PR_BRANCH: ${{ steps.pr-details.outputs.pr-branch }}
+        MERGE_QUEUE_POSITION: ${{ steps.pr-details.outputs.merge-queue-position }}
+      run: |
+        set -euo pipefail
+        if git merge-base --is-ancestor "origin/${BASE_REF}" "origin/${PR_BRANCH}" && [ "${MERGE_QUEUE_POSITION}" -eq 1 ]; then
```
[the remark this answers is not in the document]
Mrtenz: > The actions/checkout with fetch-depth: 0 only fetches the current ref's history, not all remote branches.

This is not true. `fetch-depth: 0` fetches all branches.

.github/workflows/main.yml
Gudahtt: Yeah, `always` on each step would be bad.

I was experimenting with this on a separate branch, and made the unfortunate discovery that `success()` (the default) does skip if _any_ previous job was skipped, not just the immediate previous job. The suggestion was based on the assumption that it just looked at immediate `needs` jobs. Oh well.

I am tempted to still looking for a solution here that doesn't require the empty job step, but it'll probably require reusable workflow shenanigans (potentially not worth effort/complexity even if I get something working), so it's definitely not worth waiting for that. Your solution here is good.

PR #7525, opened 2025-12-17
ci: Add action to validate changelog diffs after merging

Mrtenz:
## Explanation

This fixes an issue in changelogs where changelog entries may be incorrectly moved after merging into the main branch. Given the following (partial) changelog for example:

```md
## [Unreleased]

### Changed

- Changelog entry A ([#1234](...))
- Changelog entry B ([#5678](...))
- Changelog entry C ([#1337](...))
```

After a release, this changelog may look like this:

```diff
  ## [Unreleased]

+ ## [1.0.0]

  ### Changed

  - Changelog entry A ([#1234](...))
  - Changelog entry B ([#5678](...))
  - Changelog entry C ([#1337](...))
```

But if another PR based on the original changelog also adds an entry to the unreleased section like this:

```diff
  ## [Unreleased]

  ### Changed

  - Changelog entry A ([#1234](...))
  - Changelog entry B ([#5678](...))
  - Changelog entry C ([#1337](...))
+ - Changelog entry D ([#0000](...))
```

The final changelog will look like this, since Git can merge the entry based on the other entries around it:

```md
## [Unreleased]

## [1.0.0]

### Changed

- Changelog entry A ([#1234](...))
- Changelog entry B ([#5678](...))
- Changelog entry C ([#1337](...))
- Changelog entry D ([#0000](...))
```

Here it appears that the release included "entry D", but in reality that was meant to be under the unreleased section.

To address this issue, I've added a simple workflow that checks for changelog diffs. It uses a minimal algorithm to parse the markdown files, find entries that are included in unreleased originally, and checks if they still are after merging. If not, the workflow will fail and show an error such as this:

> Checking changelog file: packages/some-package/CHANGELOG.md
> The following lines added in the PR are missing from the "Unreleased" section after merge:
>
>    - Changelog entry D ([#0000](...))
>
> Please update your pull request and ensure that new changelog entries remain in the "Unreleased" section.

Long term we should solve this problem by using changesets instead, in which case the unreleased changes will never be merged or conflict with each other, but until we figure out how to implement those in our release workflows, we can use this workflow.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk CI-only change that adds a new workflow gate; main failure mode is false positives/negatives that could block merges.
> 
> **Overview**
> Adds a new composite GitHub Action (`check-merge-queue-changelogs`) and Node script to detect when changelog lines introduced in a PR get moved out of the `## [Unreleased]` section after merge (a common merge-queue conflict artifact).
> 
> Updates `lint-build-test.yml` to run a new `validate-changelog-diffs` job on `pull_request` and `merge_group` events, failing the workflow when modified `CHANGELOG.md` files lose newly-added Unreleased entries in the merged result.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ef5a18960def0bf5e52b1301a195becf53a3c863. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
+++ b/.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
@@ -75,11 +75,7 @@ function getUnreleasedSection(content) {
  */
 function getAddedLines(oldLines, newLines) {
   return Array.from(newLines).filter(
-    (line) =>
-      line.length > 0 &&
-      !oldLines.has(line) &&
-      !line.startsWith('#') &&
-      !line.startsWith('['),
+    (line) => line.length > 0 && !oldLines.has(line) && !line.startsWith('#'),
   );
 }
 


.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
Mrtenz, line 1: This is a plain JS file so we can just call `node check-changelog-diff.cjs` without installing dependencies etc.

.github/actions/check-merge-queue-changelogs/action.yml
```diff
@@ -0,0 +1,100 @@
...
+        GH_TOKEN: ${{ inputs.github-token }}
+      run: |
+        BRANCH=$(gh api "/repos/${REPOSITORY}/pulls/${PR_NUMBER}" --jq=.head.ref)
+        echo "pr-branch=$BRANCH" >> "$GITHUB_OUTPUT"
+
+    - name: Check changelog changes
+      id: changelog-check
+      shell: bash
+      env:
+        HEAD_REF: ${{ github.event.pull_request.head.ref || github.event.merge_group.head_ref }}
```
Gudahtt: This seems to be unused

```suggestion
```

.github/actions/check-merge-queue-changelogs/action.yml
```diff
@@ -0,0 +1,100 @@
...
+        BASE_REF: ${{ github.event.pull_request.base.ref || github.event.merge_group.base_ref }}
+        PR_BRANCH: ${{ steps.pr-branch.outputs.pr-branch }}
+        REPOSITORY: ${{ github.repository }}
```
Gudahtt: This seems to be unused

```suggestion
```

.github/actions/check-merge-queue-changelogs/action.yml
```diff
@@ -0,0 +1,100 @@
...
+
+        # Strip invalid prefix from `BASE_REF`
+        # It comes prefixed with `refs/heads/`, but the branch is not checked out in this context
+        # We need to express it as a remote branch
+        PREFIXED_REF_REGEX='refs/heads/(.+)'
+        if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+          BASE_REF="${BASH_REMATCH[1]}"
+        fi
+
+        TARGET_REF=$(git merge-base "origin/$BASE_REF" "origin/$PR_BRANCH")
```
Gudahtt, line 71: I'm not following how this ref will help us find changelog updates. It looks like we're always getting the same diff we see in the PR here, so it would miss changes due to updates on the base branch.

e.g. if I'm fixing a bug in the network-controller, then a release happens and moves that under a release header, we won't see that here. The `TARGET_REF` will be the commit on `main` _before_ the release, and the `PR_BRANCH` won't have the release on it either (the branch wasn't updated since the release; if it was, the pre-existing changelog linter would have caught this!), so we're comparing two commits here that are both too old to catch the problem.

Gudahtt: It seems like what we want here is to compare the changelog in the PR with what it would be _after_ merge.

The "before" case here would be `PR_BRANCH`, and the "after" would be the commit that triggered the workflow (which is post-merge for both `merge_group` and `pull_request`)

.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
```diff
@@ -0,0 +1,105 @@
...
+  console.error(
+    'Usage: tsx check-changelog-diff.mts <base-file> <pr-file> <merged-file>',
+  );
+
+  // eslint-disable-next-line n/no-process-exit
+  process.exit(1);
+}
+
+/* eslint-disable n/no-sync */
+// The type of these is inferred as `Buffer` when using "utf-8" directly instead
```
Gudahtt, line 16: Is this comment from an earlier draft when this was written in TypeScript?

Mrtenz: It's not, I'm using my IDE to check the types (even though the file is JS):
image

.github/actions/check-merge-queue-changelogs/action.yml
Mrtenz: I believe this is actually correct. The script does a 3-way diff, comparing the merge base file (`git show "$TARGET_REF":"$FILE"`), pull request file (`git show origin/"$PR_BRANCH":"$FILE"`), and merged file (`$FILE`, on-disk from the earlier `actions/checkout`).

The script uses the merge base file and pull request file to get the lines added to the changelog, and uses the contents of the merged file to check if those lines are still in the "Unreleased" section.

Gudahtt: Ah right, I was missing that the "after" commit was currently checked out

.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
Gudahtt: Hmm. What type is being inferred as `Buffer` in this case? I'm not understanding what the comment means.

Mrtenz: `fs.readFileSync(process.argv[2], 'utf-8')` is `Buffer`. `fs.readFileSync(process.argv[2], { encoding: 'utf-8' })` is `string`.

.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
```diff
@@ -0,0 +1,105 @@
...
+ * @param {Set<string>} newLines - The PR changelog content.
+ * @returns {string[]} The added lines as an array of strings.
+ */
+function getAddedLines(oldLines, newLines) {
+  return Array.from(newLines).filter(
+    (line) =>
+      line.length > 0 &&
+      !oldLines.has(line) &&
+      !line.startsWith('#') &&
+      !line.startsWith('['),
```
Gudahtt: How would we end up with a line starting with `[` 🤔

Mrtenz: I was thinking of the version headings for some reason, but they start with `#` 😅. Removed in ef5a18960def0bf5e52b1301a195becf53a3c863.

Mrtenz: Actually, the references on the bottom start with `[`. This may not be unnecessary.

Mrtenz: Well, this function should only process the unreleased section of the changelog. So it should be ok to remove this check.

.github/actions/check-merge-queue-changelogs/check-changelog-diff.cjs
Gudahtt: This comment was initially confusing because the phrasing made it sound like `Buffer` was the intended type.

Also I checked the inferred type with `utf-8` as a positional parameter, and it's still `string` when I test it locally :thinking:

Screenshot 2026-03-26 at 14 02 38

Mrtenz: Strange. 🤔 Could be a bug in my IDE. 
image

PR #7539, opened 2025-12-19
feat(ramps): adds controller methods for region eligibility

georgeweiler:
## Explanation

**Problem:**
The RampsController needs to determine whether a user's region is eligible for ramps (on/off-ramp services). This requires fetching eligibility information from the regions API and making it available to consumers.

**Solution:**
This PR adds region eligibility support to the RampsController:

1. **New `Eligibility` type** - Defines eligibility information returned from `/regions/countries/{isoCode}` endpoint with `aggregator`, `deposit`, and `global` flags indicating what services are available for a region.

2. **New `getEligibility` service method** - Calls `/regions/countries/{isoCode}` to fetch eligibility for a specific region (e.g., `fr`, `us-ny`). The ISO code is normalized to lowercase.

3. **Automatic eligibility fetching** - `updateGeolocation()` now automatically fetches and stores eligibility in controller state after successfully getting geolocation. This ensures eligibility is always up-to-date with the user's current location.

4. **New `getCountries` controller method** - Fetches the list of supported countries from `/regions/countries` and filters them using OnRampSDK aggregator logic:
   - For `buy`: filters by `country.supported`
   - For `sell`: filters by `country.supported` OR if any state has `supported !== false`

5. **Eligibility in controller state** - Eligibility is stored as a persisted property in `RampsControllerState`, making it accessible to consumers via `controller.state.eligibility`.

**Notable implementation details:**
- Eligibility is automatically fetched when geolocation is updated, keeping them in sync
- The eligibility endpoint supports both country codes (`fr`) and state codes (`us-ny`)
- `getCountries` uses OnRampSDK's aggregator filtering logic for consistency with existing SDK behavior
- Both geolocation and eligibility are persisted in controller state
## References

- TRAM-2923

## Checklist

---

> [!NOTE]
> Introduces region eligibility support and country listing with caching, and auto-syncs eligibility with geolocation.
> 
> - Adds `eligibility` to `RampsController` state (persisted/visible) and `updateEligibility(isoCode)` with normalized caching; prevents stale overwrites and clears on failures
> - `updateGeolocation()` now automatically fetches eligibility and retains request caching/TTL semantics
> - Adds `getCountries(action)` to controller with cached responses
> - Expands `RampsService`: new `getCountries` (regions cache URLs, support filtering), `getEligibility(isoCode)`, shared request helper adding `sdk/controller/context` params; introduces `RampsApiService`, `RAMPS_SDK_VERSION`, and `context` option
> - Updates messenger action types/permissions and public exports; enables `resolveJsonModule` to read `package.json` version
> - Comprehensive tests added/updated and changelog entry for eligibility state
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18dd92bbcb113204e9ac2fce56cca53f46f3adf7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ramps-controller/src/RampsService.ts
+++ b/packages/ramps-controller/src/RampsService.ts
@@ -22,7 +22,7 @@ export type CountryPhone = {
  */
 export type State = {
   /**
-   * State identifier. Can be in path format (e.g., "/regions/us-ut") or ISO code format (e.g., "ut").
+   * State identifier. Can be in path format (e.g., "/regions/us-ut") or ISO code format (e.g., "us-ut").
    */
   id?: string;
   /**
@@ -464,10 +464,6 @@ export class RampsService {
     }
 
     return countries.filter((country) => {
-      if (action === 'buy') {
-        return country.supported;
-      }
-
       if (country.states && country.states.length > 0) {
         const hasSupportedState = country.states.some(
           (state) => state.supported !== false,
--- a/packages/ramps-controller/src/RampsService.test.ts
+++ b/packages/ramps-controller/src/RampsService.test.ts
@@ -448,6 +448,51 @@ describe('RampsService', () => {
       expect(countriesResponse[0]?.states?.[0]?.supported).toBe(true);
     });
 
+    it('includes country with unsupported country but supported state for buy action', async () => {
+      const mockCountriesWithUnsupportedCountry = [
+        {
+          isoCode: 'US',
+          id: '/regions/us',
+          flag: '🇺🇸',
+          name: 'United States',
+          phone: { prefix: '+1', placeholder: '', template: '' },
+          currency: 'USD',
+          supported: false,
+          states: [
+            {
+              id: '/regions/us-tx',
+              stateId: 'TX',
+              name: 'Texas',
+              supported: true,
+            },
+            {
+              id: '/regions/us-ny',
+              stateId: 'NY',
+              name: 'New York',
+              supported: false,
+            },
+          ],
+        },
+      ];
+      nock('https://on-ramp-cache.uat-api.cx.metamask.io')
+        .get('/regions/countries')
+        .query({
+          action: 'buy',
+          sdk: '2.1.6',
+          controller: '2.0.0',
+          context: 'mobile-ios',
+        })
+        .reply(200, mockCountriesWithUnsupportedCountry);
+      const { service } = getService();
+
+      const countriesResponse = await service.getCountries('buy');
+
+      expect(countriesResponse).toHaveLength(1);
+      expect(countriesResponse[0]?.isoCode).toBe('US');
+      expect(countriesResponse[0]?.supported).toBe(false);
+      expect(countriesResponse[0]?.states?.[0]?.supported).toBe(true);
+    });
+
     it('throws if the countries API returns an error', async () => {
       nock('https://on-ramp-cache.uat-api.cx.metamask.io')
         .get('/regions/countries')


georgeweiler:
cursor review

packages/ramps-controller/src/RampsService.ts
```diff
@@ -6,6 +6,109 @@ import { createServicePolicy, HttpError } from '@metamask/controller-utils';
...
+  template: string;
+};
+
+/**
+ * Represents a state/province within a country.
+ */
+export type State = {
+  /**
+   * State identifier. Can be in path format (e.g., "/regions/us-ut") or ISO code format (e.g., "ut").
+   */
```
wachunei, line 26: is the format for the example meant to be`us-ut`?

wachunei: aah, this is for a state only.

georgeweiler: I could use us-ca or us-ny if you'd like? Utah is a bit random.

packages/ramps-controller/src/RampsService.ts
```diff
@@ -241,33 +363,124 @@ export class RampsService {
...
-      const url = new URL('geolocation', this.#baseUrl);
-      const localResponse = await this.#fetch(url);
-      if (!localResponse.ok) {
+  #addCommonParams(url: URL, action?: 'buy' | 'sell'): void {
+    if (action) {
+      url.searchParams.set('action', action);
+    }
+    url.searchParams.set('sdk', RAMPS_SDK_VERSION);
+    url.searchParams.set('controller', packageJson.version);
+    url.searchParams.set('context', 'mobile-ios');
```
pkowalski: Should we hardcode mobile-ios since we have some anrdoid specific logic like for enabling/disabling gpay

packages/ramps-controller/src/RampsService.ts
wachunei: it does not matter about the content but the format on the comment, a state id is the full id. for example (hypothetically speaking) if "ca" exists in US and in CL then the states must be `us-ca` and `cl-ca` since `ca` is not an identifier of its own

packages/ramps-controller/src/RampsService.ts
wachunei: good catch, we should not, context is a construction param coming from the client (mobile-ios, mobile-android, web, extension, anything in the future)

packages/ramps-controller/src/RampsController.ts
```diff
@@ -391,6 +410,80 @@ export class RampsController extends BaseController<
...
+    this.update((state) => {
+      if (state.geolocation === null) {
+        state.eligibility = eligibility;
+      } else {
+        const currentGeolocation = state.geolocation.toLowerCase().trim();
+        if (currentGeolocation === normalizedIsoCode) {
+          state.eligibility = eligibility;
+        }
+      }
+    });
```
wachunei, line 461: the result of every branch here is the same, please refactor
```ts
this.update((state) => {
  if (
    state.geolocation === null || 
    state.geolocation.toLowerCase().trim() === normalizedIsoCode
  ) {
    state.eligibility = eligibility;
  }
});
```

packages/ramps-controller/src/RampsService.ts
georgeweiler: I pushed a commit to make the context configurable

packages/ramps-controller/src/RampsService.ts
```diff
@@ -6,6 +6,109 @@ import { createServicePolicy, HttpError } from '@metamask/controller-utils';
...
+  placeholder: string;
+  template: string;
+};
+
+/**
+ * Represents a state/province within a country.
+ */
+export type State = {
+  /**
+   * State identifier. Can be in path format (e.g., "/regions/us-ut") or ISO code format (e.g., "ut").
```
wachunei:
```suggestion
   * State identifier. Can be in path format (e.g., "/regions/us-ut") or ISO code format (e.g., "us-ut").
```

packages/ramps-controller/src/RampsService.ts
```diff
@@ -241,33 +373,124 @@ export class RampsService {
...
+      }
+
+      if (country.states && country.states.length > 0) {
+        const hasSupportedState = country.states.some(
+          (state) => state.supported !== false,
+        );
+        return country.supported || hasSupportedState;
+      }
+
+      return country.supported;
```
wachunei: Regarding early return for action buy and not check for the states in line 467-469: I would expect this early return only if no states are there.

why are we early returning the value for buy?

packages/ramps-controller/src/RampsController.ts
wachunei, line 1: One feature we have discussed and I don't see is the per-method TTL in the cache, we have discussed eligibility or countries data can last for 15mins and quotes 15 seconds. In these methods I see the cache key being created but not the TTL specification per-method. 

This is also missing in `updateRequestState`, I don't see timestamp comparison with TTL amount, only sorting by timestamp and removing according to the size

wachunei: TTL has a 15 min value default. It will be needed only for methods that required a different value.

`updateRequestState` expiration cleanup will be added in a new PR

PR #7582, opened 2026-01-07
fix(STX-340): use medium fee for gas fee in submission

rarquevaux:
## Problem

Users could submit bridge/swap transactions that fail with "insufficient funds" despite the UI showing sufficient balance. This occurred because:

- **Validation** (in `bridge-controller`) uses `medium` gas price estimates
- **Submission** (in `bridge-status-controller`) was using `high` gas price estimates

This mismatch allowed transactions to pass validation but fail immediately at submission time. In the STX case it times-out in Sentinel. 

## Example (uses a non STX transaction for simplicity but the same happens with STX)

A swap of ~0.002854 ETH → USDC on Base:

{
  "quote": {
    "srcTokenAmount": "2829027500000000",
    "gasLimit": 726516,
    "effectiveGas": 431877
  }
}Submission failed with:

{
  "error": {
    "code": -32003,
    "message": "insufficient funds for gas * price + value: have 2860507066881008 want 2864770358251460"
  }
}The user had **0.00286 ETH** but the transaction required **0.00286 ETH** (4.26 gwei more than available) due to the higher `high` gas price at submission.

## Fix

Changed `bridge-status-controller` to use `medium` gas estimates on submission (matching validation):

An alternate fix would be to use `high` on validation but it would prevent more swaps to go through. 
NB: this happens in the non-gasless cases so irrespective of the fix we will see less and less occurrences of this issue as we release gasless swaps and auto-upgrade on all networks and clients.

## Checklist

---

> [!NOTE]
> Aligns gas estimation used at submission with validation to prevent mismatches (e.g., insufficient funds).
> 
> - In `gas.ts`, select `maxFeePerGas`/`maxPriorityFeePerGas` from `txGasFeeEstimates[BRIDGE_PREFERRED_GAS_ESTIMATE]` instead of hardcoded `high`
> - Tests updated in `gas.test.ts` to validate preferred estimate usage, adjust fixtures to `Medium`, and improve missing-estimate handling
> - Changelog updated under Unreleased to note the fix
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fcfac09fe7868cf276ad9c519e87d1510c7fe562. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-status-controller/src/utils/gas.ts
+++ b/packages/bridge-status-controller/src/utils/gas.ts
@@ -1,3 +1,4 @@
+import { BRIDGE_PREFERRED_GAS_ESTIMATE } from '@metamask/bridge-controller';
 import type { TokenAmountValues, TxData } from '@metamask/bridge-controller';
 import { toHex } from '@metamask/controller-utils';
 import type {
@@ -21,7 +22,8 @@ const getTransaction1559GasFeeEstimates = (
   txGasFeeEstimates: FeeMarketGasFeeEstimates,
   estimatedBaseFee: string,
 ) => {
-  const { maxFeePerGas, maxPriorityFeePerGas } = txGasFeeEstimates?.high ?? {};
+  const { maxFeePerGas, maxPriorityFeePerGas } =
+    txGasFeeEstimates?.[BRIDGE_PREFERRED_GAS_ESTIMATE] ?? {};
 
   const baseAndPriorityFeePerGas = maxPriorityFeePerGas
     ? new BigNumber(estimatedBaseFee, 10)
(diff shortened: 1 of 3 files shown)


packages/bridge-status-controller/src/utils/gas.ts
```diff
@@ -21,7 +21,8 @@ const getTransaction1559GasFeeEstimates = (
   txGasFeeEstimates: FeeMarketGasFeeEstimates,
   estimatedBaseFee: string,
 ) => {
-  const { maxFeePerGas, maxPriorityFeePerGas } = txGasFeeEstimates?.high ?? {};
```
micaelae, line 24: Can you replace this with the `BRIDGE_PREFERRED_GAS_ESTIMATE` constant exported by the bridge-controller? This way they will always be in sync

rarquevaux: done!

micaelae (approved):

PR #7587, opened 2026-01-08
chore: create new assets controller package

salimtb:
## Explanation

This PR introduces a new `@metamask/assets-controller` package as a placeholder for future consolidation of asset tracking functionality.

**Current state:**
Asset tracking functionality (account balances, token balances, asset detection) is currently spread across multiple controllers in the `@metamask/assets-controllers` package.

**Solution:**
This PR creates an empty `@metamask/assets-controller` package as a foundation for future work to consolidate asset tracking into a single, unified controller. The package currently exports nothing and serves as a placeholder with TODOs indicating the intended scope:
- Account balance tracking
- Token balance tracking
- Asset detection

No functional changes are included in this PR.

## Checklist

---

> [!NOTE]
> Introduces a scaffolded `@metamask/assets-controller` package (empty controller, exports, tests, build/test/docs configs, changelog, license) as a foundation for future asset tracking consolidation.
> 
> - Adds `packages/assets-controller` with `AssetsController` stub, unit tests, TypeScript/Jest/Typedoc configs, and package metadata
> - Updates monorepo wiring: README package list and dependency graph, `tsconfig.json`/`tsconfig.build.json` references, `yarn.lock`, `.github/CODEOWNERS`, and `teams.json`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 65d631f3c2ae04c87e8b3ba5d515e391d2f2e801. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.test.ts
+++ b/packages/assets-controller/src/AssetsController.test.ts
@@ -29,28 +29,6 @@ describe('AssetsController', () => {
     });
   });
 
-  describe('AssetsController:getState', () => {
-    it('returns the current state', async () => {
-      await withController(({ rootMessenger }) => {
-        const state = rootMessenger.call('AssetsController:getState');
-        expect(state).toMatchInlineSnapshot(`Object {}`);
-      });
-    });
-  });
-
-  describe('AssetsController:stateChange', () => {
-    it('allows subscribing to state changes', async () => {
-      await withController(({ rootMessenger }) => {
-        const listener = jest.fn();
-        rootMessenger.subscribe('AssetsController:stateChange', listener);
-
-        // Since state is empty and there's no way to change it yet,
-        // we just verify the subscription works without errors
-        expect(listener).not.toHaveBeenCalled();
-      });
-    });
-  });
-
   describe('metadata', () => {
     it('includes expected state in debug snapshots', async () => {
       await withController(({ controller }) => {


packages/assets-controller/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "@types/jest": "^27.4.1",
+    "deepmerge": "^4.2.2",
+    "jest": "^27.5.1",
+    "ts-jest": "^27.1.4",
+    "typedoc": "^0.24.8",
+    "typedoc-plugin-missing-exports": "^2.0.0",
+    "typescript": "~5.3.3"
+  },
+  "peerDependencies": {
+    "@metamask/messenger": "^0.3.0"
```
cryptodev-2s: Nit: this should be a dependency

tsconfig.build.json
```diff
@@ -1,11 +1,4 @@
 {
-  /**
```
cryptodev-2s, line 2: any reasons why these where removed ?

tsconfig.json
```diff
@@ -1,12 +1,6 @@
 {
-  /**
-   * This configuration is used by the `lint` script in `package.json`, and by editors such as
```
cryptodev-2s, line 3: Nit: any reasons why these comments where removed ?

packages/assets-controller/src/AssetsController.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  });
+
+  describe('actions', () => {
+    it('should respond to AssetsController:getState action', () => {
+      const rootMessenger = getRootMessenger();
+      const messenger = getMessenger(rootMessenger);
+
+      // eslint-disable-next-line no-new
+      new AssetsController({ messenger });
```
cryptodev-2s: What do you think about starting this scaffolded with withController pattern ?

packages/assets-controller/src/AssetsController.ts
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @returns The default state.
+ */
+export const getDefaultAssetsControllerState =
+  (): AssetsControllerState => ({});
+
+/**
+ * The metadata for the state of the {@link AssetsController}.
+ */
+const assetsControllerMetadata: StateMetadata<AssetsControllerState> = {};
```
mcmire: Can we use `satisfies` instead of a type assertion here? See `SamplePetnamesController` in the `sample-controllers` package: https://github.com/MetaMask/core/blob/5db273fee57f84c05b6bbf65564a2c6058fc7671/packages/sample-controllers/src/sample-petnames-controller.ts#L42-L49

``` suggestion
const assetsControllerMetadata = {} satisfies StateMetadata<AssetsControllerState>;

File: packages/assets-controller/src/index.ts [hunk 6]
```diff
@@ -0,0 +1,13 @@
...
+  AssetsControllerState,
+  AssetsControllerGetStateAction,
+  AssetsControllerActions,
+  AssetsControllerStateChangeEvent,
+  AssetsControllerEvents,
+  AssetsControllerMessenger,
+} from './AssetsController';
+export {
+  AssetsController,
+  controllerName,
```
mcmire: Generally we do not export names of controllers, `controllerName` is an internal variable only. If there are multiple controllers in a package then there would be multiple `controllerName` exports. Would you mind removing this?

``` suggestion
```

packages/assets-controller/src/AssetsController.ts
mcmire, line 1: Two questions:

1. What do you think about copying and pasting one of the controllers in the `sample-controllers` package? This way we can keep the general look and feel for a controller consistent across all packages.

2. I see that we are creating a new package called `assets-controller`, when we already have a package called `assets-controllers` (with an "s"). I'm a bit worried this will be a constant source of confusion, and so perhaps we shouldn't make a new package yet? Or — are there any near-term plans to split apart `assets-controllers`?

salimtb: @mcmire yes, there are near-term plans to split apart `assets-controllers`. The goal is to deprecate the existing `assets-controllers` (plural) package and break it into several focused, single-responsibility packages:

- `@metamask/assets-controller` - consolidated asset tracking (balances, detection)
- Other controllers from the package will be split into their own dedicated packages as well

The naming is intentional:
- `assets-controllers` = legacy "kitchen sink" package (to be deprecated)
- `assets-controller` = new single unified controller

This follows the pattern we've used elsewhere in the monorepo where packages contain a single controller (e.g., `@metamask/network-controller`, `@metamask/keyring-controller`).

Once the migration is complete and `assets-controllers` is deprecated and deleted, the confusion will be resolved. In the meantime, we can add a note to the `assets-controllers` README indicating it's being deprecated and pointing to the new packages.

mcmire: @salimtb Okay, that sounds like a great plan. Thanks for clarifying!

packages/assets-controller/src/AssetsController.test.ts
```diff
@@ -0,0 +1,180 @@
...
+        const listener = jest.fn();
+        rootMessenger.subscribe('AssetsController:stateChange', listener);
+
+        // Since state is empty and there's no way to change it yet,
+        // we just verify the subscription works without errors
+        expect(listener).not.toHaveBeenCalled();
+      });
+    });
+  });
+
```
mcmire: Nit: Typically we don't test events and actions that are supplied via BaseController. They are already tested there. So you can remove these tests if you like.

``` suggestion
```

PR #7588, opened 2026-01-08
Update preview build instructions

mcmire:
## Explanation

Preview builds are represented as prereleases, which means that care must be taken to accurately simulate what the dependency tree of a project will look like when a new production version is released. Unfortunately, the existing instructions for using preview builds do not achieve this. This problem has always existed but was exacerbated by the recent shift from peer dependencies to dependencies.

This commit updates the instructions to fix these issues. There are now separate instructions for testing breaking changes vs. non-breaking changes. An addendum for working with patches, which was missing before, is also now included. Finally, instructions for using local builds has been split off from the preview builds instructions for ease of readability.

**View rendered versions:**

- [`local-builds.md`](https://github.com/MetaMask/core/blob/update-preview-builds-instructions/docs/processes/local-builds.md)
- [`preview-builds.md`](https://github.com/MetaMask/core/blob/update-preview-builds-instructions/docs/processes/local-builds.md)

## Checklist

---

> [!NOTE]
> Consolidates and clarifies testing workflows, separating preview builds from local builds and refining guidance for breaking vs. non-breaking changes.
> 
> - Adds `docs/processes/preview-builds.md` with detailed steps (including patched deps) and independent contributor flow
> - Adds `docs/processes/local-builds.md` with Yarn `file:` workflow
> - Simplifies `docs/processes/testing-changes-in-other-projects.md` to point to the two new guides
> - Updates `docs/processes/breaking-changes.md` to link to `preview-builds.md`
> - Updates `scripts/generate-preview-build-message.ts` to point to the new `preview-builds.md` URL
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa249f42547c5dde926442940f968b6daa73400b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/processes/preview-builds.md
+++ b/docs/processes/preview-builds.md
@@ -37,28 +37,51 @@ Note that the steps above will only work if you are a member of the MetaMask eng
 
 ## Using preview builds
 
-To simulate production as best as possible, there two different paths to take depending on whether you anticipate making breaking or non-breaking changes to the package you want to test.
+To simulate production as best as possible, there are two different paths to take depending on whether you anticipate making breaking or non-breaking changes to the package you want to test.
 
 ### Testing non-breaking changes to a package
 
-If you're in a MetaMask client repo (e.g. `metamask-extension` or `metamask-mobile`), run:
+If you're in a MetaMask client repo (e.g. `metamask-extension` or `metamask-mobile`):
 
-```bash
-yarn use-preview-build <name-of-package> <preview-version> --type non-breaking
-```
+1. Open `package.json` and locate the `previewBuilds` section. If it doesn't exist, then add it (it should be an object).
+2. Within `previewBuilds`, add an entry that looks like this:
 
-For example:
+   ```
+   "<name-of-package>": {
+     "type": "non-breaking",
+     "previewVersion": "<preview-version>"
+   }
+   ```
 
-```bash
-yarn use-preview-build @metamask/permission-controller 13.0.2-preview-940c934 --type non-breaking
-```
+   For example:
+
+   ```
+   "@metamask/permission-controller": {
+     "type": "non-breaking",
+     "previewVersion": "13.0.2-preview-940c934"
+   }
+   ```
+
+3. Run `yarn install`. You're now using the preview build.
 
 <details>
 <summary><strong>Manual steps</strong></summary>
 
-If you have made non-breaking changes to the package you want to test, and therefore plan on bumping the _minor_ or _patch_ part of that package's version, follow these steps:
-
 1. In the project, open `package.json` and locate the entry in `dependencies` for the package. Take note of the major part of the version.
+
+   - Note that the dependency may be patched. Patched dependencies look like this:
+
+     ```json
+     "@metamask/<PACKAGE_NAME>": "patch:@metamask/<PACKAGE_NAME>@npm:<ESCAPED_VERSION_RANGE>#~/.yarn/patches/<PATCH_NAME>.patch"
+     ```
+
+     where:
+
+     - `PACKAGE_NAME` is the name of the package.
+     - `ESCAPED_VERSION_RANGE` is the version range, but where characters like `^` are escaped (for instance, `^` will appear as `%3A`).
+     - `PATCH_NAME` is the filename for a patch. It is usually dash-separated, starts with the package name, and ends with the commit ID from which the patch was created.
+
+     In this case the major version you are targeting will come from `ESCAPED_VERSION_RANGE`.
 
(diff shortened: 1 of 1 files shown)


docs/processes/preview-builds.md
```diff
@@ -0,0 +1,204 @@
...
+
+To simulate production as best as possible, there are different instructions for using preview builds depending on whether the changes to a package you are testing are breaking or non-breaking.
+
+### Testing non-breaking changes to a package
+
+If you have made non-breaking changes to the package you want to test, and therefore plan on bumping the _minor_ or _patch_ part of that package's version, follow these steps:
+
+1. In the project, open `package.json` and locate the entry in `dependencies` for the package. Take note of the major part of the version.
+
+2. Run `yarn why @metamask/<PACKAGE_NAME>`, replacing `PACKAGE_NAME` as appropriate. You will see output like this:
```
mcmire, line 86: This is a new process that I'm introducing in this PR. The previous process recommended that the engineer look at the existing dependency entry and target whatever version range was used there. Now I'm recommended that all version ranges in the dependency tree that are compatible with the one used in the dependency tree are targeted. As I mention at the top of this section, this is to mimic production as close as possible.

So on one hand, it's more accurate. On the other hand, it's more steps that before. Do we have any thoughts on this? It would nice if we could automate this somehow, but I'm not sure how to do that yet.

docs/processes/local-builds.md
```diff
@@ -0,0 +1,41 @@
+# Using local builds
+
+If you are unable to use [preview builds](./preview-builds.md) for testing, you can always build a package locally and link it to a project.
+
+1. First, build the monorepo by running `yarn build`.
+
+2. In the project, open `package.json` and locate the entry in `dependencies` for the package you want to test.
+
+3. Replace the version range in the right-hand side of the entry to point to the local build:
```
Mrtenz, line 9: I think `yarn link -A` does this as well, did you try that?

docs/processes/preview-builds.md
```diff
@@ -0,0 +1,236 @@
...
+   yarn build
+   yarn publish-previews
+   ```
+
+   You will see a list of the newly published packages along with their versions.
+   </details>
+
+## Using preview builds
+
+To simulate production as best as possible, there two different paths to take depending on whether you anticipate making breaking or non-breaking changes to the package you want to test.
```
Mrtenz: Out of curiosity, what's the difference between testing non-breaking and breaking changes?

mcmire: When you're testing breaking changes to a package, you're going to be bumping the version by a major, so when you release the new version you're testing, that version will not get deduplicated with versions of that package that already exist in the dependency tree. When you're testing non-breaking changes, you're going to be bumping by a minor or patch, so the new version _will_ get deduplicated.

When you are using preview builds, you need to mimic this same behavior, especially for breaking changes. The previous instructions were not doing this effectively, so transitive dependencies would get updated to use preview builds, too, and we might not see the full effects of those breaking changes. For instance:

- We are making breaking changes to one of the public types in `@metamask/network-controller` and plan on releasing those changes
- A client is already using `^27.0.0`
- We add `"@metamask/network-controller@^27.0.0": "npm:@metamask-previews/network-controller@27.0.0-preview-abcd123"` to `resolutions`
- But `@metamask/transaction-controller` is also using `^27.0.0`, so now it gets updated to use the preview build too
- The breaking changes that we made "magically" work in `@metamask/transaction-controller`, and we get no type errors, so we don't know that when we release `network-controller` we will also need to bump and release `transaction-controller` too.

docs/processes/local-builds.md
mcmire: `yarn link` (or, using the `portal:` protocol) doesn't always work in my experience. Sometimes Yarn will claim that transitive dependencies are incompatible between the package being tested and the project. I've found that `file:` always works, even if it's less convenient.

For instance if you run `yarn link -A ../create-release-branch` in `core` you get:

```
➤ YN0071: │ Cannot link @metamask/create-release-branch into @metamask/core-monorepo@workspace:. dependency @metamask/utils@npm:9.3.0 conflicts with parent dependency @metamask/utils@npm:11.9.0
➤ YN0071: │ Cannot link @metamask/create-release-branch into @metamask/core-monorepo@workspace:. dependency execa@npm:8.0.1 conflicts with parent dependency execa@npm:5.1.1
```

docs/processes/preview-builds.md
Mrtenz: I see, thanks for the clarification!

PR #7603, opened 2026-01-12
feat: align with 7715 specification

MoMannn:

The `@metamask/eth-json-rpc-middleware` package currently supports two EIP-7715 methods:
- `wallet_requestExecutionPermissions` — for requesting new execution permissions
- `wallet_revokeExecutionPermission` — for revoking existing execution permissions

However, the EIP-7715 specification also defines two additional "discoverability" methods that allow dApps to query the wallet about its execution permission capabilities. Without these methods, dApps have no way to:
1. Discover which permission types the wallet supports and on which chains
2. Retrieve the list of currently granted (non-revoked) permissions for the user

This PR adds support for the two missing EIP-7715 discoverability methods:

- **`wallet_getSupportedExecutionPermissions`** — Returns an object keyed by permission type (e.g., `native-token-allowance`, `erc20-token-allowance`) with their supported chain IDs and rule types. This allows dApps to understand what the wallet can handle before making permission requests.

- **`wallet_getGrantedExecutionPermissions`** — Returns an array of all currently active (non-revoked) execution permissions. Each permission includes chain ID, addresses, permission details, context, dependencies, and delegation manager information.

The implementation follows the established patterns in this package:
- Each method has a dedicated handler file with a factory function (`createWallet*Handler`)
- Hook-based architecture allows consumers to provide their own processing logic
- Superstruct schemas are defined for result types, enabling runtime validation
- Types and structs are exported from the package index for consumer use

#### EIP-7715 spec alignment for `wallet_requestExecutionPermissions`

The request/response schema has been updated to align with the latest EIP-7715 specification:

- **Simplified address fields**: Replaced the nested `address` + `signer: { type, data: { address } }` structure with flat `from` (optional) and `to` (required) fields
- **Removed `isAdjustmentAllowed` from rules**: This property is only relevant at the permission level, not individual rules
- **Extended result type**: Added `dependencies` (array of `{ factory, factoryData }`) and `delegationManager` to the response
- **New export**: Added `PermissionDependency` type export for consumers
- **Consistent address validation**: Updated `GrantedExecutionPermissionStruct` to use `HexChecksumAddressStruct` for address fields

The handlers don't perform parameter validation since these methods don't accept parameters (they're pure query methods). This is intentional and differs from `wallet_requestExecutionPermissions` and `wallet_revokeExecutionPermission` which validate their params using Superstruct schemas.

## References

- Needed for: https://github.com/MetaMask/metamask-extension/pull/39176
- Related to: https://github.com/MetaMask/snap-7715-permissions/pull/249

## Checklist

---

> [!NOTE]
> Implements EIP-7715 discoverability methods and integrates them into the wallet middleware.
> 
> - New handlers: `createWalletGetSupportedExecutionPermissionsHandler` and `createWalletGetGrantedExecutionPermissionsHandler` using hook-based processing and `NoParamsStruct` validation
> - Exports new types and structs from `index.ts` (`SupportedExecutionPermissionConfig*`, `GrantedExecutionPermission*`, results)
> - Adds `EmptyArrayStruct`/`NoParamsStruct` in `utils/structs` for "no params" validation
> - Wires methods into `createWalletMiddleware` via optional `processGet*` hooks
> - Unit tests for both methods: hook invocation, result passthrough, missing-hook error, and params validation
> - Changelog updated to note new RPC support
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e94f86b2eb84817944e0ae1f0140e36b6f893c1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/eth-json-rpc-middleware/src/methods/wallet-request-execution-permissions.test.ts
+++ b/packages/eth-json-rpc-middleware/src/methods/wallet-request-execution-permissions.test.ts
@@ -119,6 +119,18 @@ describe('wallet_requestExecutionPermissions', () => {
     );
   });
 
+  it('supports null rules', async () => {
+    params[0].rules = null as never;
+
+    await callMethod();
+
+    expect(processRequestExecutionPermissionsMock).toHaveBeenCalledWith(
+      params,
+      request,
+      context,
+    );
+  });
+
   it('supports optional from', async () => {
     params[0].from = undefined;
 
--- a/packages/eth-json-rpc-middleware/src/methods/wallet-request-execution-permissions.ts
+++ b/packages/eth-json-rpc-middleware/src/methods/wallet-request-execution-permissions.ts
@@ -4,10 +4,12 @@ import type { Infer } from '@metamask/superstruct';
 import {
   array,
   boolean,
+  literal,
   object,
   optional,
   record,
   string,
+  union,
   unknown,
 } from '@metamask/superstruct';
 import { HexChecksumAddressStruct, StrictHexStruct } from '@metamask/utils';
@@ -32,7 +34,7 @@ const PermissionRequestStruct = object({
   from: optional(HexChecksumAddressStruct),
   to: HexChecksumAddressStruct,
   permission: PermissionStruct,
-  rules: optional(array(RuleStruct)),
+  rules: optional(union([array(RuleStruct), literal(null)])),
 });
 
 export const RequestExecutionPermissionsStruct = array(PermissionRequestStruct);


packages/eth-json-rpc-middleware/src/methods/wallet-get-granted-execution-permissions.ts
```diff
@@ -0,0 +1,93 @@
...
+  processGetGrantedExecutionPermissions?: ProcessGetGrantedExecutionPermissionsHook;
+}): JsonRpcMiddleware<JsonRpcRequest, Json, WalletMiddlewareContext> {
+  return async ({ request, context }) => {
+    if (!processGetGrantedExecutionPermissions) {
+      throw rpcErrors.methodNotSupported(
+        'wallet_getGrantedExecutionPermissions - no middleware configured',
+      );
+    }
+
+    return await processGetGrantedExecutionPermissions(request, context);
```
mj-kiwi, line 105: The two new methods (`wallet_getSupportedExecutionPermissions` and `wallet_getGrantedExecutionPermissions`) currently don’t perform parameter validation.

Although these methods do not expect any parameters, it’s still recommended to call `validateParams` in the handler to explicitly assert that `params` is empty or `undefined`. This helps protect against unexpected or potentially malicious input.

For reference, `wallet_revokeExecutionPermission` follows this pattern by validating parameters in the handler:

```ts
validateParams(params, RevokeExecutionPermissionRequestParamsStruct);
```

You could apply the same approach in the two new handlers:

```ts
const { params } = request;
validateParams(params, GetSupportedExecutionPermissionsParamsStruct);
```

with an empty params struct defined as:

```ts
export const GetSupportedExecutionPermissionsParamsStruct = object({});
```

This keeps the behavior consistent with existing handlers and ensures invalid requests are caught early in the middleware layer, with clearer and more predictable error handling.

mj-kiwi (dismissed):
Let's go

packages/eth-json-rpc-middleware/src/methods/wallet-get-granted-execution-permissions.ts
```diff
@@ -0,0 +1,107 @@
...
+  return async ({ request, context }) => {
+    if (!processGetGrantedExecutionPermissions) {
+      throw rpcErrors.methodNotSupported(
+        'wallet_getGrantedExecutionPermissions - no middleware configured',
+      );
+    }
+
+    const { params } = request;
+
+    validateParams(params, GetGrantedExecutionPermissionsParamsStruct);
```
jiexi, line 103: we can probably just forgo validation in this case?

packages/eth-json-rpc-middleware/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for `wallet_getSupportedExecutionPermissions` and `wallet_getGrantedExecutionPermissions` RPC methods ([#7603](https://github.com/MetaMask/core/pull/7603))
```
jiexi, line 12: PermissionDependency is also exported now

packages/eth-json-rpc-middleware/src/methods/wallet-get-granted-execution-permissions.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        factoryData: '0xabcdef' as Hex,
+      },
+    ],
+    delegationManager: '0xAb8483F64d9C6d1EcF9b849Ae677dD3315835cb2' as Hex,
+  },
+];
+
+const REQUEST_MOCK = {
+  params: [],
+} as unknown as JsonRpcRequest;
```
jiexi, line 36: nit. if we add, `id`, `jsonrpc`, and `method`, we can drop the typecast here and the several below. Seems like the trouble might be worth it

PR #7609, opened 2026-01-13
feat: add CAIP-19 asset identifiers for each chain into the state

salimtb:
### Current State
The `NetworkEnablementController` tracks which networks are enabled/disabled but lacks information about the native asset identifier for each chain. This information is needed to properly identify native assets across different blockchain networks using a standardized format.

Extension UI PR: https://github.com/MetaMask/metamask-extension/pull/39234

### Solution
This PR adds a new state property `nativeAssetIdentifiers` to the `NetworkEnablementController` that maps CAIP-2 chain IDs to CAIP-19-like native asset identifiers. 

For example:
- `eip155:1` → `eip155:1/slip44:60` (Ethereum uses ETH, SLIP-44 coin type 60)
- `eip155:56` → `eip155:56/slip44:714` (BSC uses BNB, SLIP-44 coin type 714)
- `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp` → `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501`

### Implementation Details
1. **New `Slip44Service`**: A service that wraps the `@metamask/slip44` package to look up SLIP-44 coin types by symbol (e.g., `ETH` → `60`, `BTC` → `0`).

2. **State Updates**: The `nativeAssetIdentifiers` state is automatically updated when:
   - A network is added (`NetworkController:networkAdded`)
   - A network is removed (`NetworkController:networkRemoved`)
   - A network's native currency symbol is changed (`NetworkController:stateChange`)

3. **Default State**: Popular networks are pre-populated with their correct native asset identifiers based on their native currency symbols.

4. **Graceful Handling**: If a symbol has no SLIP-44 mapping, the entry is removed from the state rather than storing an invalid identifier.

### Dependency Addition
Added `@metamask/slip44` package to dynamically look up SLIP-44 coin types instead of hardcoding them, ensuring accuracy and maintainability.

## References

- [SLIP-44 Standard](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)
- [CAIP-2 Chain ID Specification](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md)
- [CAIP-19 Asset Identifier Specification](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-19.md)
- [@metamask/slip44 npm package](https://www.npmjs.com/package/@metamask/slip44)

## Checklist

---

> [!NOTE]
> Introduces native asset tracking and supporting utilities.
> 
> - Adds `nativeAssetIdentifiers` mapping (CAIP-2 -> `.../slip44:<coinType>`) to `NetworkEnablementController` state and metadata
> - New `Slip44Service` (uses `@metamask/slip44` and chainid.network) to resolve coin types by symbol/chainId; exports helpers
> - Makes `init()` async and populates EVM identifiers; new `initNativeAssetIdentifiers(networks)` API for bulk initialization
> - Updates network lifecycle: on add (async) computes and stores identifier; on remove deletes identifier
> - Extends public exports (types, service) and adds `@metamask/slip44` dependency
> - Updates/extends tests for controller, selectors, utils, and new service
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8cdcfc266bab0da0043592ea003a3243b56de68. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


salimtb:
@metamaskbot publish-preview

packages/network-enablement-controller/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `nativeAssetIdentifiers` state property that maps CAIP-2 chain IDs to CAIP-19-like native asset identifiers (e.g., `eip155:1/slip44:60`) ([#7609](https://github.com/MetaMask/core/pull/7609))
+- Add `Slip44Service` to look up SLIP-44 coin types by native currency symbol ([#7609](https://github.com/MetaMask/core/pull/7609))
+- Add `@metamask/slip44` dependency for SLIP-44 coin type lookups ([#7609](https://github.com/MetaMask/core/pull/7609))
+- Subscribe to `NetworkController:stateChange` to update `nativeAssetIdentifiers` when a network's native currency changes ([#7609](https://github.com/MetaMask/core/pull/7609))
```
sahar-fehri, line 15: Hey @salimtb nice work, is this more related to unified controllers work or to the network picker discussion we had yesterday

salimtb: @sahar-fehri it's related to the unified controllers , and mainly the RPC data source , the mapping for native tokens will use slip-44 instead of the 0 address now

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -134,6 +181,9 @@ const getDefaultNetworkEnablementControllerState =
         [TrxScope.Shasta]: false,
       },
     },
+    // nativeAssetIdentifiers is initialized as empty and should be populated
+    // by the client using initNativeAssetIdentifiers() during controller init
+    nativeAssetIdentifiers: {},
```
Prithpal-Sooriya, line 186: hmm, actually not a bad place to add the native asset identifiers (close to the actual chains a user might have).

I was wondering if this could be stateless -- just a simple cache inside middleware/a module instead of inside a controller persisted state.

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -184,13 +240,52 @@ export class NetworkEnablementController extends BaseController<
...
 
     messenger.subscribe('NetworkController:networkRemoved', ({ chainId }) => {
       this.#removeNetworkEntry(chainId);
     });
+
+    messenger.subscribe(
+      'NetworkController:stateChange',
+      (_newState, patches) => {
+        this.#onNetworkControllerStateChange(patches);
+      },
```
Prithpal-Sooriya: Can you double check, I'm pretty sure core does not want us to subscribe or execute against all controller state changes.

Can we instead use the selector approach for state changes instead of patches?
```
messenger.subscribe(
  '...',
  (pickedStuff, prevPickedStuff) => {...doSomething},
  (state: State) => pickOutStateYouAreTestingAgainst
)
```

packages/network-enablement-controller/src/NetworkEnablementController.ts
salimtb: I’m trying to avoid processing this on every app launch. The goal would be to compute it once, then only update it when a network is added or removed or edited, rather than relying on a purely stateless in-memory cache and compute this every time i open the app.
also this information will be used out of the middleware , some other teams need it

packages/network-enablement-controller/src/NetworkEnablementController.ts
salimtb: hmmm maybe i need to export a new event here , because the goal is to listen into ticker change , let me try to see if we have it, i don't want to make change into NetworkController then release it now , let me check if then already export what i need

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -100,6 +124,29 @@ export type NetworkEnablementControllerMessenger = Messenger<
...
+ *
+ * @param caipChainId - The CAIP-2 chain ID (e.g., 'eip155:1', 'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp')
+ * @param slip44CoinType - The SLIP-44 coin type number
+ * @returns The native asset identifier string (e.g., 'eip155:1/slip44:60')
+ */
+function buildNativeAssetIdentifier(
+  caipChainId: CaipChainId,
+  slip44CoinType: number,
+): NativeAssetIdentifier {
+  return `${caipChainId}/slip44:${slip44CoinType}`;
```
Prithpal-Sooriya, line 138: Oh interesting! So we can actually build the slip44 identifier? I was under the presumption that this is something we need to fetch and index?

salimtb: yes ,the index is what we have on slip44CoinType , but we are building the assetsIdentifier here , example:
`eip155:1/slip44:60` `<caipChainId>/slip44:<index slip-44 fetched from the npm package>`
so yes we fetch the index and [this](https://www.npmjs.com/package/@metamask/slip44) package is used

packages/network-enablement-controller/src/NetworkEnablementController.ts
salimtb: we don't have such event on the NetworkController , i should create it , merge it and release a new version of the networkController , i'll do this separately and not include it on this PR

Prithpal-Sooriya (dismissed):

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -184,13 +240,52 @@ export class NetworkEnablementController extends BaseController<
...
+   * Handles NetworkController state changes to detect symbol updates.
+   *
+   * @param patches - The patches describing what changed
+   */
+  #onNetworkControllerStateChange(
+    patches: { op: string; path: (string | number)[]; value?: unknown }[],
+  ): void {
+    // Look for patches that replace a network configuration
+    // Path format: ['networkConfigurationsByChainId', chainId]
+    for (const patch of patches) {
```
FrederikBolding: Generally we should refrain from inspecting patches like this, as it is error prone.

packages/network-enablement-controller/src/services/Slip44Service.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @example
+   * ```typescript
+   * const ethCoinType = Slip44Service.getSlip44BySymbol('ETH');
+   * // Returns 60
+   *
+   * const btcCoinType = Slip44Service.getSlip44BySymbol('BTC');
+   * // Returns 0
+   * ```
+   */
+  static getSlip44BySymbol(symbol: string): number | undefined {
```
FrederikBolding, line 173: If useful we could consider exposing more files in `metamask/slip44`, such as one keyed by symbol. Unsure if they are guaranteed to be unique though

packages/network-enablement-controller/src/services/Slip44Service.ts
```diff
@@ -0,0 +1,220 @@
...
+    return undefined;
+  }
+
+  /**
+   * Gets the SLIP-44 entry for a given coin type index.
+   *
+   * @param index - The SLIP-44 coin type index (e.g., 60 for ETH, 0 for BTC)
+   * @returns The SLIP-44 entry with metadata, or undefined if not found
+   */
+  static getSlip44Entry(index: number): Slip44Entry | undefined {
```
bergarces, line 206: Is there a use case for this method at the client? I cannot see it being currently used and the parameter name (`index`) is a bit confusing.

If there are no plans to use it I'd use remove it until it's needed, but otherwise we can rename `index` to `chainId` like in `getSlip44ByChainId`.

salimtb: i don't need this method anymore , the one we need is the `getSlip44ByChainId`, i'll remove it in a separated PR now

packages/network-enablement-controller/src/services/Slip44Service.ts
```diff
@@ -0,0 +1,220 @@
...
+    // Check chainId cache first
+    const cached = this.#chainIdCache?.get(chainId);
+    if (cached !== undefined) {
+      return cached;
+    }
+
+    // Fall back to symbol lookup if provided
+    if (symbol) {
+      return this.getSlip44BySymbol(symbol);
+    }
```
bergarces, line 150: There are quite a few entries with repeated symbols in the imported json. This does not look very safe.

LBTC
LKY
LTO
LUM
MONK
MTR
NXT
ONE
PAC
RYO
SAFE
SAMA
SBC
SCR
SUM
TAO
UC
VAL
VARA
XCH
XEC
XRD

salimtb: i'm actually using this only for non evm ( BTC, SOL, Tron ) , they are unique , for evm the provider https://chainid.network/chains.json is used

salimtb: so the case of duplicated symbols is handled

salimtb: maybe i should remove the fallback here , is this your point ?

PR #7623, opened 2026-01-14
feat: add  new `ConnectivityController`

cryptodev-2s:
## Explanation

This PR introduces a new `@metamask/connectivity-controller` package that provides a centralized way to track and manage device internet connectivity status across MetaMask clients (extension and mobile).

- Have a single source of truth for connectivity state
- Share connectivity status across controllers and UI components
- Test connectivity-related functionality in isolation
- Maintain consistent behavior across platforms

The `ConnectivityController` is a platform-agnostic controller that:
- Stores the current connectivity status (`online` or `offline`) in its state
- Uses a service injection pattern where platform-specific `ConnectivityService` implementations are injected at construction time
- Subscribes to connectivity changes from the service and automatically updates its state
- Emits `stateChange` events when connectivity status changes, allowing other controllers and UI components to react accordingly

The controller is designed to work with different service implementations:
- **Mobile**: `NetInfoConnectivityService` using `@react-native-community/netinfo`
- **Extension (same context)**: `BrowserConnectivityService` using browser APIs

- The `connectivityStatus` field in state is prefixed with "connectivity" to avoid conflicts when state is flattened in Redux
- The state is configured with `persist: false` since connectivity status is ephemeral and should not be restored from storage

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WPC-210

## Checklist

---

> [!NOTE]
> Introduce a new, platform-agnostic controller to track device connectivity with adapter-driven status updates.
> 
> - New package `@metamask/connectivity-controller` exposing `ConnectivityController`, `CONNECTIVITY_STATUSES`, and `ConnectivityAdapter` interface
> - Controller initializes from adapter `getStatus()` and updates state on `onConnectivityChange`, with non-persistent `connectivityStatus` metadata
> - Comprehensive tests for initialization and status change handling; 100% coverage config
> - Added package docs (README, CHANGELOG, LICENSE) and build/test configs (tsconfig, typedoc, jest)
> - Monorepo integration: README package list/graph, `CODEOWNERS`, `teams.json`, tsconfig references, and `yarn.lock` entry
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16c51b55a5cc3e4435f70715cbecedfe7e04fdde. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/connectivity-controller/src/ConnectivityController.test.ts
+++ b/packages/connectivity-controller/src/ConnectivityController.test.ts
@@ -7,45 +7,9 @@ import type {
 } from '@metamask/messenger';
 
 import type { ConnectivityControllerMessenger } from './ConnectivityController';
-import {
-  ConnectivityController,
-  controllerName,
-} from './ConnectivityController';
+import { ConnectivityController } from './ConnectivityController';
 import { CONNECTIVITY_STATUSES } from './types';
 import type { ConnectivityAdapter, ConnectivityStatus } from './types';
-
-/**
- * A test implementation of ConnectivityAdapter.
- * Allows manual control of connectivity status via setStatus.
- */
-class TestConnectivityAdapter implements ConnectivityAdapter {
-  #status: ConnectivityStatus;
-
-  #onConnectivityChangeCallbacks: ((status: ConnectivityStatus) => void)[] = [];
-
-  constructor(
-    initialStatus: ConnectivityStatus = CONNECTIVITY_STATUSES.Online,
-  ) {
-    this.#status = initialStatus;
-  }
-
-  getStatus(): ConnectivityStatus {
-    return this.#status;
-  }
-
-  onConnectivityChange(callback: (status: ConnectivityStatus) => void): void {
-    this.#onConnectivityChangeCallbacks.push(callback);
-  }
-
-  setStatus(status: ConnectivityStatus): void {
-    this.#status = status;
-    this.#onConnectivityChangeCallbacks.forEach((callback) => callback(status));
-  }
-
-  destroy(): void {
-    this.#onConnectivityChangeCallbacks = [];
-  }
-}
 
 describe('ConnectivityController', () => {
   describe('constructor', () => {
@@ -80,38 +44,6 @@ describe('ConnectivityController', () => {
           expect(controller.state.connectivityStatus).toBe(
             CONNECTIVITY_STATUSES.Offline,
           );
-        },
-      );
-    });
-
-    it('subscribes to service connectivity changes', async () => {
-      const mockAdapter: ConnectivityAdapter = {
-        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
-        onConnectivityChange: jest.fn(),
-        destroy: jest.fn(),
-      };
-
-      await withController(
-        { options: { connectivityAdapter: mockAdapter } },
-        () => {
-          expect(mockAdapter.onConnectivityChange).toHaveBeenCalledWith(
-            expect.any(Function),
-          );
-        },
-      );
-    });
-
-    it('has correct name property', async () => {
-      const mockAdapter: ConnectivityAdapter = {
-        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
-        onConnectivityChange: jest.fn(),
-        destroy: jest.fn(),
-      };
-
-      await withController(
-        { options: { connectivityAdapter: mockAdapter } },
-        ({ controller }) => {
-          expect(controller.name).toBe(controllerName);
         },
       );
     });
@@ -179,32 +111,28 @@ describe('ConnectivityController', () => {
     });
   });
 
-  describe('service callbacks', () => {
+  describe('when connectivity changes via the adapter', () => {
     it('updates state when service reports offline', async () => {
-      const mockOnConnectivityChange = jest.fn();
+      let onConnectivityChangeCallback: (
+        connectivityStatus: ConnectivityStatus,
+      ) => void;
       const mockAdapter: ConnectivityAdapter = {
         getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
-        onConnectivityChange: mockOnConnectivityChange,
+        onConnectivityChange(
+          callback: (connectivityStatus: ConnectivityStatus) => void,
+        ) {
+          onConnectivityChangeCallback = callback;
+        },
         destroy: jest.fn(),
       };
-
       await withController(
         { options: { connectivityAdapter: mockAdapter } },
         ({ controller }) => {
           expect(controller.state.connectivityStatus).toBe(
             CONNECTIVITY_STATUSES.Online,
           );
-
-          // Get the callback that was passed to onConnectivityChange
-          const capturedCallback = mockOnConnectivityChange.mock
-            .calls[0]?.[0] as
-            | ((status: ConnectivityStatus) => void)
-            | undefined;
-          expect(capturedCallback).toBeDefined();
-
           // Simulate service reporting offline
-          capturedCallback?.(CONNECTIVITY_STATUSES.Offline);
-
+          onConnectivityChangeCallback(CONNECTIVITY_STATUSES.Offline);
           expect(controller.state.connectivityStatus).toBe(
             CONNECTIVITY_STATUSES.Offline,
           );
@@ -213,119 +141,28 @@ describe('ConnectivityController', () => {
     });
 
     it('updates state when service reports online', async () => {
-      const mockOnConnectivityChange = jest.fn();
+      let onConnectivityChangeCallback: (
+        connectivityStatus: ConnectivityStatus,
+      ) => void;
       const mockAdapter: ConnectivityAdapter = {
         getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Offline),
-        onConnectivityChange: mockOnConnectivityChange,
+        onConnectivityChange(
+          callback: (connectivityStatus: ConnectivityStatus) => void,
+        ) {
+          onConnectivityChangeCallback = callback;
+        },
         destroy: jest.fn(),
       };
-
       await withController(
         { options: { connectivityAdapter: mockAdapter } },
         ({ controller }) => {
           expect(controller.state.connectivityStatus).toBe(
             CONNECTIVITY_STATUSES.Offline,
           );
-
-          // Get the callback that was passed to onConnectivityChange
-          const capturedCallback = mockOnConnectivityChange.mock
-            .calls[0]?.[0] as
-            | ((status: ConnectivityStatus) => void)
-            | undefined;
-          expect(capturedCallback).toBeDefined();
-
           // Simulate service reporting online
-          capturedCallback?.(CONNECTIVITY_STATUSES.Online);
-
+          onConnectivityChangeCallback(CONNECTIVITY_STATUSES.Online);
           expect(controller.state.connectivityStatus).toBe(
             CONNECTIVITY_STATUSES.Online,
-          );
-        },
-      );
-    });
-
-    it('emits stateChange event when status changes', async () => {
-      const mockOnConnectivityChange = jest.fn();
-      const mockAdapter: ConnectivityAdapter = {
-        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
-        onConnectivityChange: mockOnConnectivityChange,
-        destroy: jest.fn(),
-      };
-
-      await withController(
-        { options: { connectivityAdapter: mockAdapter } },
-        ({ controllerMessenger }) => {
-          const eventHandler = jest.fn();
-          controllerMessenger.subscribe(
-            `${controllerName}:stateChange`,
-            eventHandler,
-          );
-
-          // Get the callback that was passed to onConnectivityChange
-          const capturedCallback = mockOnConnectivityChange.mock
-            .calls[0]?.[0] as
-            | ((status: ConnectivityStatus) => void)
-            | undefined;
-          expect(capturedCallback).toBeDefined();
-
-          capturedCallback?.(CONNECTIVITY_STATUSES.Offline);
-
-          expect(eventHandler).toHaveBeenCalledWith(
-            { connectivityStatus: CONNECTIVITY_STATUSES.Offline },
-            expect.any(Array),
-          );
-        },
-      );
-    });
-
-    it('does not emit event when status does not change', async () => {
-      const mockOnConnectivityChange = jest.fn();
-      const mockAdapter: ConnectivityAdapter = {
-        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
-        onConnectivityChange: mockOnConnectivityChange,
-        destroy: jest.fn(),
-      };
-
-      await withController(
-        { options: { connectivityAdapter: mockAdapter } },
-        ({ controllerMessenger }) => {
-          const eventHandler = jest.fn();
-          controllerMessenger.subscribe(
-            `${controllerName}:stateChange`,
-            eventHandler,
-          );
-
-          // Get the callback that was passed to onConnectivityChange
-          const capturedCallback = mockOnConnectivityChange.mock
-            .calls[0]?.[0] as
-            | ((status: ConnectivityStatus) => void)
-            | undefined;
-          expect(capturedCallback).toBeDefined();
-
-          // Report online when already online
-          capturedCallback?.(CONNECTIVITY_STATUSES.Online);
-
-          expect(eventHandler).not.toHaveBeenCalled();
-        },
-      );
-    });
-  });
-
-  describe('with TestConnectivityAdapter', () => {
-    it('updates state when adapter.setStatus is called', async () => {
-      const adapter = new TestConnectivityAdapter();
-
-      await withController(
-        { options: { connectivityAdapter: adapter } },
-        ({ controller }) => {
-          expect(controller.state.connectivityStatus).toBe(
-            CONNECTIVITY_STATUSES.Online,
-          );
-
-          adapter.setStatus(CONNECTIVITY_STATUSES.Offline);
-
-          expect(controller.state.connectivityStatus).toBe(
-            CONNECTIVITY_STATUSES.Offline,
           );
         },
       );
@@ -405,7 +242,11 @@ async function withController<ReturnValue>(
     args.length === 2 ? args : [{}, args[0]];
   const rootMessenger = getRootMessenger();
   const controllerMessenger = getMessenger(rootMessenger);
-  const defaultAdapter = new TestConnectivityAdapter();
+  const defaultAdapter: ConnectivityAdapter = {
+    getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
+    onConnectivityChange: jest.fn(),
+    destroy: jest.fn(),
+  };
   const controller = new ConnectivityController({
     messenger: controllerMessenger,
     connectivityAdapter: defaultAdapter,


packages/connectivity-controller/src/ConnectivityController.ts
```diff
@@ -0,0 +1,185 @@
...
+ */
+export const controllerName = 'ConnectivityController';
+
+/**
+ * State for the {@link ConnectivityController}.
+ */
+export type ConnectivityControllerState = {
+  /**
+   * The current device connectivity status.
+   * Named with 'connectivity' prefix to avoid conflicts when state is flattened in Redux.
```
mcmire, line 25: Good idea.

packages/connectivity-controller/src/ConnectivityController.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Connectivity service for platform-specific detection.
+   *
+   * The controller subscribes to the service's `onConnectivityChange`
+   * callback to receive connectivity updates.
+   *
+   * Platform implementations:
+   * - Mobile: Use `NetInfoConnectivityService` with `@react-native-community/netinfo`
+   * - Extension (same context): Use `BrowserConnectivityService`
+   * - Extension (cross-context): Use `PassiveConnectivityService` and call
+   * `setStatus()` from the UI context
```
mcmire: Do we need to include this here? It seems that we would need to update this documentation if the implementation in the client changes.

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.ts
```diff
@@ -0,0 +1,185 @@
...
+ * and updates its state accordingly. All connectivity updates flow through
+ * the service, ensuring a single source of truth.
+ *
+ * **Platform implementations:**
+ *
+ * - **Mobile:** Inject `NetInfoConnectivityService` using `@react-native-community/netinfo`
+ * - **Extension:** Inject `PassiveConnectivityService` in the background.
+ * Status is updated via the `setDeviceConnectivityStatus` API, which is called from:
+ * - MV3: Offscreen document (where browser events work reliably)
+ * - MV2: Background page (where browser events work directly)
```
mcmire: Similar to above — do we need to include this in the documentation?

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.ts
```diff
@@ -0,0 +1,185 @@
...
+        connectivityStatus: initialStatus,
+      },
+    });
+
+    connectivityService.onConnectivityChange((status) => {
+      this.update((draftState) => {
+        draftState.connectivityStatus = status;
+      });
+    });
+  }
```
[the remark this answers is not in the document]
mcmire: That's true. We haven't been consistent about implementing `destroy`. But maybe we should do this?

If the ConnectivityService used the messenger then we could call `unsubscribe`. Otherwise, we would need ConnectivityService to implement EventEmitter so that it could call `removeEventListener` in its `destroy`.

Maybe ConnectivityService _should_ use the messenger? Curious to know your thoughts. (But also see my comment below.)

packages/connectivity-controller/src/types.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Connectivity status constants.
+ * Used to represent whether the device has internet connectivity.
+ */
+export const ConnectivityStatus = {
```
mcmire: What do you think about calling this `CONNECTIVITY_STATUSES`? We don't have formal guidelines on these "quasi-enums", but I recently found [this guide](https://mkosir.github.io/typescript-style-guide/#enums--const-assertion) and have found it to be useful.

packages/connectivity-controller/src/ConnectivityController.ts
```diff
@@ -0,0 +1,185 @@
...
+   */
+  connectivityStatus: ConnectivityStatus;
+};
+
+/**
+ * The metadata for each property in {@link ConnectivityControllerState}.
+ */
+const connectivityControllerMetadata = {
+  connectivityStatus: {
+    persist: false,
```
mcmire, line 35: Hmm, so I thought we _were_ going to persist this state.

If don't plan on persisting _any_ state, then do we need a controller? I'm wondering if it makes more sense to use a service instead. If we have some place where we need to respond to state changes, we can mimic that by having the service store the state internally and publishing an event when the state changes.

Here in `core`, we could define a messenger type and interface for the ConnectivityService, but leave the implementation to the clients. Other controllers like NetworkController can allowlist the action type for the new service, as they would for a controller.

I guess the argument against creating a service is, if we are mimicking a controller by storing state and publishing an event when the state changes, why not use a controller? My answer is that this particular scenario isn't typical. We have a case where the implementation for ConnectivityService doesn't live in `core`, it's client-specific. This controller, then, is just a wrapper around the service, so it seems like an extra layer to me.

What do you think though?

packages/connectivity-controller/src/types.ts
cryptodev-2s: I totally agree, I find it also very strange to have type and const having the same name

packages/connectivity-controller/src/ConnectivityController.ts
cryptodev-2s: Actually I initially started this as a service. Then switched back to a controller for 2 main reasons.
1. Easily access the status from the UI (as currently accessed from redux state to hide the banner).
2. Make it by default available in debug snapshot which could add highly appreciated output from sentry issues.

packages/connectivity-controller/src/types.ts
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/7623/changes/9d171f5a60034c219129f29fcbf0b5f928b81efa

packages/connectivity-controller/src/ConnectivityController.ts
cryptodev-2s: Make sense removed here https://github.com/MetaMask/core/pull/7623/changes/eda16e41722eec8742eb145f6139160134dd0b96

packages/connectivity-controller/src/ConnectivityController.ts
cryptodev-2s: Make sense removed here https://github.com/MetaMask/core/pull/7623/changes/eda16e41722eec8742eb145f6139160134dd0b96

packages/connectivity-controller/src/ConnectivityController.ts
mcmire: Oh, right, I forgot about the UI. Yeah, that makes a lot of sense for it to be a controller then.

packages/connectivity-controller/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Service interface for platform-specific connectivity detection.
+ *
+ * Each platform (extension, mobile) implements this interface using
+ * platform-specific APIs:
+ * - Extension: `navigator.onLine` and `online`/`offline` events
+ * - Mobile: `@react-native-community/netinfo`
+ *
+ * The service is injected into the ConnectivityController, which
+ * subscribes to connectivity changes and updates its state accordingly.
```
mcmire: Perhaps we don't need this here? Maybe we can go to the clients if we need to know how they implement this service.

``` suggestion
```

packages/connectivity-controller/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export type ConnectivityService = {
```
mcmire: Since this doesn't follow the service pattern, I am hesitant to call it one.

Maybe "adapter" is a better name? We follow a similar pattern in AnalyticsController: https://github.com/MetaMask/core/blob/a7fad314c101ce05b83eb5b57b9c80cc5ee4112c/packages/analytics-controller/src/AnalyticsController.ts#L214

``` suggestion
export type ConnectivityAdapter = {

File: packages/connectivity-controller/src/ConnectivityController.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (cryptodev-2s) [reply packages/connectivity-controller/src/ConnectivityController.ts:167]: @mcmire I am not sure what I need to do ?

File: packages/connectivity-controller/src/ConnectivityController.ts [hunk 9]
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Constructs a new {@link ConnectivityController}.
+   *
+   * @param args - The arguments to this controller.
+   * @param args.messenger - The messenger suited for this controller.
+   * @param args.connectivityService - The connectivity service to use.
+   */
+  constructor({
+    messenger,
+    connectivityService,
```
mcmire: Since this doesn't follow the service pattern, maybe `connectivityAdapter` is a better name? This is what `AnalyticsController` uses: https://github.com/MetaMask/core/blob/a7fad314c101ce05b83eb5b57b9c80cc5ee4112c/packages/analytics-controller/src/AnalyticsController.ts#L214

packages/connectivity-controller/src/ConnectivityController.ts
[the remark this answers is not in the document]
mcmire: The comment from Cursor noted that some controllers use a `destroy` method to clean up event subscriptions, but we don't do that here, because the connectivity service doesn't let us do this. I was debating whether this was necessary. If it is, then perhaps it makes sense for the connectivity service to implement the messenger pattern so this is easier to do.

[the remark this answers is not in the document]
mcmire: However, I think we can ignore this Cursor comment and revisit this later. So there is nothing to do.

packages/connectivity-controller/src/ConnectivityController.ts
cryptodev-2s: Renamed here https://github.com/MetaMask/core/pull/7623/changes/85ecacbae6d8790d7be4727fee5923f74b5e4d99

packages/connectivity-controller/src/types.ts
cryptodev-2s: Renamed here https://github.com/MetaMask/core/pull/7623/changes/85ecacbae6d8790d7be4727fee5923f74b5e4d99

packages/connectivity-controller/src/types.ts
cryptodev-2s: removed here https://github.com/MetaMask/core/pull/7623/changes/85ecacbae6d8790d7be4727fee5923f74b5e4d99

cryptodev-2s:
@metamaskbot publish-preview

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+        ).toMatchInlineSnapshot(`
+          Object {
+            "connectivityStatus": "online",
+          }
+        `);
+      });
+    });
+  });
+
+  describe('service callbacks', () => {
```
mcmire: Since there is only one callback we are testing, should we be more direct here?

``` suggestion
  describe('when connectivity changes via the adapter', () => {

File: packages/connectivity-controller/src/ConnectivityController.test.ts [hunk 11]
```diff
@@ -0,0 +1,415 @@
...
+          );
+
+          adapter.setStatus(CONNECTIVITY_STATUSES.Offline);
+
+          expect(controller.state.connectivityStatus).toBe(
+            CONNECTIVITY_STATUSES.Offline,
+          );
+        },
+      );
+    });
```
mcmire: Do we need this test, or TestConnectivityAdapter for that matter? There's no requirement that `setStatus` in the adapter triggers the `onConnectivityStatus` callback. I wonder if just testing `onConnectivityStatus` is good enough.

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+
+          // Simulate service reporting offline
+          capturedCallback?.(CONNECTIVITY_STATUSES.Offline);
+
+          expect(controller.state.connectivityStatus).toBe(
+            CONNECTIVITY_STATUSES.Offline,
+          );
+        },
+      );
+    });
```
mcmire, line 141: Since we create the mock adapter, do we need to use the Jest API to go back through and find a mock function? I think we could simplify this test:

``` suggestion
    it('updates state when service reports offline', async () => {
      let onConnectivityChangeCallback: (
        connectivityStatus: ConnectivityStatus,
      ) => void;
      const mockAdapter: ConnectivityAdapter = {
        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Online),
        onConnectivityChange(
          callback: (connectivityStatus: ConnectivityStatus) => void,
        ) {
          onConnectivityChangeCallback = callback;
        },
        destroy: jest.fn(),
      };

      await withController(
        { options: { connectivityAdapter: mockAdapter } },
        ({ controller }) => {
          expect(controller.state.connectivityStatus).toBe(
            CONNECTIVITY_STATUSES.Online,
          );

          // Simulate service reporting offline
          onConnectivityChangeCallback(CONNECTIVITY_STATUSES.Offline);

          expect(controller.state.connectivityStatus).toBe(
            CONNECTIVITY_STATUSES.Offline,
          );
        },
      );
    });
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+
+          // Simulate service reporting online
+          capturedCallback?.(CONNECTIVITY_STATUSES.Online);
+
+          expect(controller.state.connectivityStatus).toBe(
+            CONNECTIVITY_STATUSES.Online,
+          );
+        },
+      );
+    });
```
mcmire, line 169: Similar as above:

``` suggestion
    it('updates state when service reports online', async () => {
      let onConnectivityChangeCallback: (
        connectivityStatus: ConnectivityStatus,
      ) => void;
      const mockAdapter: ConnectivityAdapter = {
        getStatus: jest.fn().mockReturnValue(CONNECTIVITY_STATUSES.Offline),
        onConnectivityChange(
          callback: (connectivityStatus: ConnectivityStatus) => void,
        ) {
          onConnectivityChangeCallback = callback;
        },
        destroy: jest.fn(),
      };

      await withController(
        { options: { connectivityAdapter: mockAdapter } },
        ({ controller }) => {
          expect(controller.state.connectivityStatus).toBe(
            CONNECTIVITY_STATUSES.Offline,
          );

          // Simulate service reporting online
          onConnectivityChangeCallback(CONNECTIVITY_STATUSES.Online);

          expect(controller.state.connectivityStatus).toBe(
            CONNECTIVITY_STATUSES.Online,
          );
        },
      );
    });
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+
+          capturedCallback?.(CONNECTIVITY_STATUSES.Offline);
+
+          expect(eventHandler).toHaveBeenCalledWith(
+            { connectivityStatus: CONNECTIVITY_STATUSES.Offline },
+            expect.any(Array),
+          );
+        },
+      );
+    });
```
mcmire: Do we need this test? We don't typically test that the `:stateChange` event is published when the controller state changes — usually it is enough just to test that the state changes, which you've already done.

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+            | undefined;
+          expect(capturedCallback).toBeDefined();
+
+          // Report online when already online
+          capturedCallback?.(CONNECTIVITY_STATUSES.Online);
+
+          expect(eventHandler).not.toHaveBeenCalled();
+        },
+      );
+    });
```
mcmire: Similarly, do we need this test?

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+async function withController<ReturnValue>(
+  ...args:
+    | [WithControllerCallback<ReturnValue>]
+    | [WithControllerOptions, WithControllerCallback<ReturnValue>]
+): Promise<ReturnValue> {
+  const [{ options = {} }, testFunction] =
+    args.length === 2 ? args : [{}, args[0]];
+  const rootMessenger = getRootMessenger();
+  const controllerMessenger = getMessenger(rootMessenger);
+  const defaultAdapter = new TestConnectivityAdapter();
```
mcmire: Maybe we don't need to make a separate class, but can make an object on the fly as you've done in some of the tests above?

``` suggestion
  const defaultAdapter = {
    getStatus: jest.fn(),
    onConnectivityChange: jest.fn(),
    destroy: jest.fn(),
  };
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+      await withController(
+        { options: { connectivityAdapter: mockAdapter } },
+        () => {
+          expect(mockAdapter.onConnectivityChange).toHaveBeenCalledWith(
+            expect.any(Function),
+          );
+        },
+      );
+    });
+
```
mcmire: Do we need this test? We already test subscribing to connectivity change below, so we are testing this implicitly.

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.test.ts
```diff
@@ -0,0 +1,415 @@
...
+        destroy: jest.fn(),
+      };
+
+      await withController(
+        { options: { connectivityAdapter: mockAdapter } },
+        ({ controller }) => {
+          expect(controller.name).toBe(controllerName);
+        },
+      );
+    });
```
mcmire: Do we need this test? We already test that `name` gets set in the tests for BaseController.

``` suggestion
```

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Updated here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Updated here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/7623/changes/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

packages/connectivity-controller/src/ConnectivityController.test.ts
cryptodev-2s: Updated here https://github.com/MetaMask/core/commit/16c51b55a5cc3e4435f70715cbecedfe7e04fdde

PR #7624, opened 2026-01-14
feat!: add time-based delay for first metrics sending

mikesposito:
## Explanation

`ProfileMetricsController` is being overhauled to rely on a timestamp-based delay for the first sending of metrics, rather than waiting for a second App Unlock event. This change ensures that metrics are sent even if the user only unlocks the app once during a session.

The delay is skipped immediately if:
- A transaction is submitted
- The user is already opted-in when `KeyringController:unlock` is fired (this covers both users who have opted-in in a previous session, and users who just onboarded and opted-in during onboarding).
- The consumer client calls the `.skipInitialDelay()` method or the `ProfileMetricsController:skipInitialDelay` action via messenger.

## Checklist

---

> [!NOTE]
> Introduces a timestamp-based initial delay for first profile metrics send, with configurable duration and multiple skip paths.
> 
> - Add `initialDelayEndTimestamp` state and default 10-minute delay; constructor accepts `initialDelayDuration`
> - New `skipInitialDelay()` method and `ProfileMetricsController:skipInitialDelay` action; auto-skip on `KeyringController:unlock` when already opted-in and on `TransactionController:transactionSubmitted`
> - Polling now starts on unlock; `_executePoll` exits until delay completes; state metadata updated to persist/expose new field
> - BREAKING: messenger must allow `TransactionController:transactionSubmitted`; add `@metamask/transaction-controller` dependency and TS references; export new method-action types; tests and README dependency graph updated
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0237820b5034bb8ad6b0adb98dae5b0d6d83d40f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/profile-metrics-controller/package.json
+++ b/packages/profile-metrics-controller/package.json
@@ -55,7 +55,7 @@
     "@metamask/messenger": "^0.3.0",
     "@metamask/polling-controller": "^16.0.2",
     "@metamask/profile-sync-controller": "^27.0.0",
-    "@metamask/transaction-controller": "^62.9.1",
+    "@metamask/transaction-controller": "^62.9.2",
     "@metamask/utils": "^11.9.0",
     "async-mutex": "^0.5.0"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -4506,7 +4506,7 @@ __metadata:
     "@metamask/messenger": "npm:^0.3.0"
     "@metamask/polling-controller": "npm:^16.0.2"
     "@metamask/profile-sync-controller": "npm:^27.0.0"
-    "@metamask/transaction-controller": "npm:^62.9.1"
+    "@metamask/transaction-controller": "npm:^62.9.2"
     "@metamask/utils": "npm:^11.9.0"
     "@ts-bridge/cli": "npm:^0.6.4"
     "@types/jest": "npm:^27.4.1"


packages/profile-metrics-controller/src/ProfileMetricsController.ts
```diff
@@ -27,6 +29,11 @@ import type { AccountWithScopes } from './ProfileMetricsService';
  */
 export const controllerName = 'ProfileMetricsController';
 
+/**
+ * The default delay duration before data is sent for the first time, in milliseconds.
+ */
+export const DEFAULT_INITIAL_DELAY_DURATION = 10 * 60 * 1000; // 10 minutes
```
Mrtenz: We have a nice method in `@metamask/utils` for this:

```suggestion
export const DEFAULT_INITIAL_DELAY_DURATION = inMilliseconds(10, Duration.Minute);
```

packages/profile-metrics-controller/src/ProfileMetricsController.ts
```diff
@@ -188,22 +212,30 @@ export class ProfileMetricsController extends StaticIntervalPollingController()<
...
 
     this.messenger.subscribe('KeyringController:unlock', () => {
-      this.#queueFirstSyncIfNeeded().catch(console.error);
       if (this.#assertUserOptedIn()) {
-        this.startPolling(null);
+        // If the user has already opted in at the start of the session,
+        // it must have opted in during onboarding, or during a previous session.
+        this.skipInitialDelay();
       }
+      this.#queueFirstSyncIfNeeded().catch(console.error);
```
Mrtenz: Should this use `messenger.captureException`?

packages/profile-metrics-controller/src/ProfileMetricsController.ts
mikesposito: That's cool! changed in 750feaa770ce

packages/profile-metrics-controller/src/ProfileMetricsController.ts
mikesposito: Definitely, changed to `this.messenger.captureException ?? console.error` in a70fe32f24f6

Mrtenz: `messenger.captureException` should always be defined in practice, so I think just using `messenger?.captureException` is ok, but this is fine too.

mikesposito: It seems to be defined as optional: https://github.com/MetaMask/core/blob/0237820b5034bb8ad6b0adb98dae5b0d6d83d40f/packages/messenger/src/Messenger.ts#L263-L263
Without using `console.error` as fallback, I still get `no-floating-promises` errors

PR #7626, opened 2026-01-14
fix: clean up slip 44 service logic

salimtb:
## Explanation

The `Slip44Service` was introduced in #7609 to look up SLIP-44 coin types for building native asset identifiers (CAIP-19). However, the initial API had some complexity that wasn't necessary:

1. The `getSlip44ByChainId` method had an optional `symbol` parameter for fallback, but this fallback logic was confusing since EVM networks should use chain ID lookup and non-EVM networks should use symbol lookup - mixing them in one method added unnecessary complexity.

2. The method could return `undefined` for unknown EVM chains, but since EVM networks are all Ethereum-compatible, defaulting to 60 (Ethereum's SLIP-44 coin type) is a reasonable and safe default.

3. The `getSlip44Entry` method and `Slip44Entry` type export weren't being used anywhere.

This PR cleans up the service by:
- Renaming `getSlip44ByChainId` → `getEvmSlip44` to make it explicit that it's for EVM networks only
- Removing the optional `symbol` fallback parameter (use `getSlip44BySymbol` directly for non-EVM networks)
- Always returning `60` (Ethereum) as the default for unknown EVM chain IDs instead of `undefined`
- Removing the unused `getSlip44Entry` method and `Slip44Entry` type export
- Updating the `NetworkEnablementController` to no longer pass the native currency symbol to the lookup function

## References

- Follow-up cleanup to #7609

## Checklist

---

> [!NOTE]
> Streamlines SLIP-44 resolution and aligns controller usage.
> 
> - Rename `getSlip44ByChainId` → `getEvmSlip44` (EVM-only) and always default to coin type `60` when unmapped
> - Remove symbol fallback from EVM lookup; use `getSlip44BySymbol` only for non-EVM networks
> - Delete unused `getSlip44Entry` and its type/export; update `services/index.ts` and package `index.ts` exports
> - Update `NetworkEnablementController` to call `Slip44Service.getEvmSlip44`, stop passing native currency to add-network handler, and parse references correctly
> - Adjust tests to mock/use `getEvmSlip44`, update expectations (e.g., `eip155:2` → `slip44:60`), and remove unused entry tests
> - Add changelog entry under "Fixed" for Slip44Service cleanup
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47cfb4e3cd811abb115b1890257b393f0652dbfc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/network-enablement-controller/src/NetworkEnablementController.ts
+++ b/packages/network-enablement-controller/src/NetworkEnablementController.ts
@@ -460,7 +460,6 @@ export class NetworkEnablementController extends BaseController<
       const numericChainId = parseInt(chainId, 16);
 
       // EVM networks: use getEvmSlip44 (chainid.network data)
-      // Defaults to 60 (Ethereum) if no specific mapping is found
       const slip44CoinType = await Slip44Service.getEvmSlip44(numericChainId);
 
       evmNativeAssetUpdates.push({
@@ -547,7 +546,6 @@ export class NetworkEnablementController extends BaseController<
 
       if (namespace === 'eip155') {
         // EVM networks: use getEvmSlip44 (chainid.network data)
-        // Defaults to 60 (Ethereum) if no specific mapping is found
         const numericChainId = parseInt(reference, 10);
         slip44CoinType = await Slip44Service.getEvmSlip44(numericChainId);
       } else {
@@ -713,7 +711,6 @@ export class NetworkEnablementController extends BaseController<
     const numericChainId = parseInt(reference, 10);
 
     // EVM networks: use getEvmSlip44 (chainid.network data)
-    // Defaults to 60 (Ethereum) if no specific mapping is found
     const slip44CoinType = await Slip44Service.getEvmSlip44(numericChainId);
 
     this.update((state) => {


packages/network-enablement-controller/src/services/Slip44Service.ts
```diff
@@ -144,20 +147,18 @@ export class Slip44Service {
...
     }
 
-    // Fall back to symbol lookup if provided
-    if (symbol) {
-      return this.getSlip44BySymbol(symbol);
-    }
-
-    return undefined;
+    // Default to 60 (Ethereum) for EVM networks without specific mapping
+    return 60;
```
bergarces, line 151: 💀 

Is undefined not better here? If not, at least the behaviour is very clear here.

bergarces: Forget what I said, I see now that it defaulted to 60 anyway in the consumer.

packages/network-enablement-controller/src/NetworkEnablementController.ts
```diff
@@ -462,13 +459,9 @@ export class NetworkEnablementController extends BaseController<
...
 
-      // EVM networks: use getSlip44ByChainId (chainid.network data)
-      // Default to 60 (Ethereum) if no specific mapping is found
-      const slip44CoinType =
-        (await Slip44Service.getSlip44ByChainId(
-          numericChainId,
-          config.nativeCurrency,
-        )) ?? 60;
+      // EVM networks: use getEvmSlip44 (chainid.network data)
+      // Defaults to 60 (Ethereum) if no specific mapping is found
```
bergarces:
```suggestion
```
Can remove this comment as it's now dealt with at service level.

packages/network-enablement-controller/src/services/Slip44Service.ts
salimtb: I don’t think `undefined` is better here, because it pushes the ambiguity to runtime.

Defaulting to `slip44:60` makes the behavior explicit and deterministic for EVM chains, which matches how wallets already derive and sign accounts. An `undefined` value would still require downstream logic to infer or guess the coin type, and that inference would almost certainly end up being 60 anyway.

bergarces (approved):

Issue #7618, opened 2026-01-14
feat: create analytics privacy controller


NicolasMassart:
## Description

Create a new `AnalyticsPrivacyController` in the `@metamask/core` repository to manage analytics privacy and data deletion functionality. This controller will handle GDPR/CCPA compliance features including data deletion requests, tracking deletion status, and managing data recording flags.

### State Management

The controller should extend `BaseController` and manage the following state:

```typescript
interface AnalyticsPrivacyControllerState {
  dataRecorded: boolean;
  deleteRegulationId: string | undefined;
  deleteRegulationDate: string | undefined; // DD/MM/YYYY format
}
```

### Actions

The controller should expose the following actions:

1. **`createDataDeletionTask()`**
   - Creates a new delete regulation for the user via Segment API
   - Returns: `Promise<IDeleteRegulationResponse>`
   - Purpose: GDPR/CCPA compliance - initiates data deletion request

2. **`checkDataDeleteStatus()`**
   - Checks the latest delete regulation status
   - Returns: `Promise<IDeleteRegulationStatus>`
   - Purpose: Get current status of data deletion request

3. **`getDeleteRegulationCreationDate()`**
   - Gets the latest delete regulation request date
   - Returns: `string | undefined` (format: DD/MM/YYYY)
   - Purpose: Retrieve when deletion request was created

4. **`getDeleteRegulationId()`**
   - Gets the latest delete regulation request ID
   - Returns: `string | undefined`
   - Purpose: Retrieve Segment regulation ID for status checks

5. **`isDataRecorded()`**
   - Indicates if events have been recorded since the last deletion request
   - Returns: `boolean`
   - Purpose: Track if analytics data was collected after deletion request

6. **`updateDataRecordingFlag(saveDataRecording?: boolean)`**
   - Updates the data recording flag after tracking events
   - Returns: `void`
   - Purpose: Maintain data recording state for deletion workflows

### Events

The controller should emit the following events:

```typescript
interface AnalyticsPrivacyControllerEvents {
  /**
   * Emitted when controller state changes
   */
  stateChange: (state: AnalyticsPrivacyControllerState) => void;
  
  /**
   * Emitted when data deletion task is created
   */
  dataDeletionTaskCreated: (response: IDeleteRegulationResponse) => void;
  
  /**
   * Emitted when data recording flag is updated
   */
  dataRecordingFlagUpdated: (isRecorded: boolean) => void;
}
```

### Controller Options

The controller should accept the following options during initialization:

[...]

PR #7643, opened 2026-01-15
Feat: new analytics data regulation controller

NicolasMassart:
## Explanation

This PR introduces a new `@metamask/analytics-data-regulation-controller` package that provides GDPR/CCPA data deletion functionality for analytics data. The package allows to extract the logic from the mobile app (and will be compatible with extension too)

**Current state:** MetaMask mobile app currently has a dedicated mechanism to handle user data deletion requests for analytics data in compliance with GDPR and CCPA regulations.

**Solution:** This package introduces:
- **AnalyticsDataRegulationController**: A controller that manages the lifecycle of data deletion requests, tracks whether new data has been recorded since the last deletion request, and stores deletion regulation metadata (ID and timestamp)
- **AnalyticsDataRegulationService**: A service that communicates with Segment's Regulations API via a proxy endpoint to create deletion tasks and check their status
- **State management**: Tracks `hasCollectedDataSinceDeletionRequest` flag, `deleteRegulationId`, and `deleteRegulationTimestamp` to support compliance workflows
- **Selectors**: Provides reusable selectors for accessing controller state

**Implementation details:**
- The controller takes `analyticsId` as a constructor parameter (provided by the consumer)
- It delegates to `AnalyticsDataRegulationService` via messenger actions to make HTTP requests to Segment's Regulations API
- The service uses `createServicePolicy` from `@metamask/controller-utils` for retry logic and error handling
- State is persisted and can be used to determine if new analytics events have been recorded since the last deletion request
- The package includes comprehensive test coverage (100% branch, function, line, and statement coverage)

## References

see also https://github.com/MetaMask/metamask-mobile/issues/22016

Fixes https://github.com/MetaMask/core/issues/7618

## Checklist

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 52f4a2cad1c312fcc8cbd987cb55d801bab1bf75. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-data-regulation-controller/src/types.ts
+++ b/packages/analytics-data-regulation-controller/src/types.ts
@@ -50,4 +50,3 @@ export type DeleteRegulationStatus = {
   hasCollectedDataSinceDeletionRequest: boolean;
   dataDeletionRequestStatus: DataDeleteStatus;
 };
-


packages/analytics-privacy-controller/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "test:clean": "NODE_OPTIONS=--experimental-vm-modules jest --clearCache",
+    "test:verbose": "NODE_OPTIONS=--experimental-vm-modules jest --verbose",
+    "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
+  },
+  "dependencies": {
+    "@metamask/analytics-controller": "^1.0.0",
+    "@metamask/base-controller": "^9.0.0",
+    "@metamask/controller-utils": "^7.0.0",
+    "@metamask/messenger": "^0.3.0",
+    "@metamask/utils": "^11.9.0"
```
NicolasMassart: I have to update these as they do not match what other package have in their package.json

NicolasMassart:
@metamaskbot publish-preview

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationService.ts
```diff
@@ -0,0 +1,394 @@
...
+   * Constructs a new AnalyticsPrivacyService object.
+   *
+   * @param options - The constructor options.
+   */
+  constructor(options: AnalyticsPrivacyServiceOptions) {
+    this.name = serviceName;
+    this.#messenger = options.messenger;
+    this.#fetch = options.fetch;
+    this.#segmentSourceId = options.segmentSourceId;
+    this.#segmentRegulationsEndpoint = options.segmentRegulationsEndpoint;
```
mcmire, line 232: My thought around data services is that they should represent an API, and the implication is that the exact API they represent should be obvious. However, it seems that this class accepts any URL. That means the knowledge about which external system this is talking to is somewhere else.

Is it possible to include the URL in this same file, perhaps as a constant? If it's based on environment, perhaps the class can take an environment and use that to build the URL?

packages/analytics-data-regulation-controller/src/logger.ts
mcmire, line 1: Nit: Since there is no class here, is there a reason why this file needs to be `PascalCase`? Thoughts on calling this file `logger.ts`?

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+// === STATE ===
+
+/**
+ * Describes the shape of the state object for {@link AnalyticsPrivacyController}.
+ */
+export type AnalyticsPrivacyControllerState = {
+  /**
+   * Indicates if data has been recorded since the last deletion request.
+   */
+  dataRecorded: boolean;
```
mcmire: In `checkDataDeleteStatus`, this is renamed to `hasCollectedDataSinceDeletionRequest`. I think this name does a good job of describing the purpose. Thoughts on renaming the state property?

``` suggestion
  hasCollectedDataSinceDeletionRequest: boolean;

File: packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts [hunk 4]
```diff
@@ -0,0 +1,375 @@
...
+   */
+  async createDataDeletionTask(): Promise<IDeleteRegulationResponse> {
+    try {
+      const analyticsControllerState = this.messenger.call(
+        'AnalyticsController:getState',
+      );
+      const { analyticsId } = analyticsControllerState;
+
+      if (!analyticsId || analyticsId.trim() === '') {
+        const error = new Error('Analytics ID not found');
```
mcmire: Nit: Do we want a more descriptive message which instructs the engineer on how to fix this? Perhaps something like "You need to set up AnalyticsController with an analytics ID. You can do this by..." (etc.)

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+      }
+
+      const response = await this.messenger.call(
+        'AnalyticsPrivacyService:createDataDeletionTask',
+        analyticsId,
+      );
+
+      if (
+        response.status === DataDeleteResponseStatus.Success &&
+        response.regulateId &&
```
mcmire: We seem to be validating the response data. Is there a reason why we are doing that here? Ideally, the service should return data that is usable, and if not, should throw an error.

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+      }
+
+      return response;
+    } catch (error) {
+      log('Analytics Deletion Task Error', error);
+      const errorMessage =
+        error instanceof Error
+          ? error.message
+          : 'Analytics Deletion Task Error';
+      return {
```
mcmire: Is there a reason why we are returning a result object rather than throwing an error? This isn't a pattern that we've been following. Not to say it's necessarily bad, but at the moment an error would accomplish the same thing. Do we anticipate adding more properties to this object in the future?

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+      hasCollectedDataSinceDeletionRequest: false,
+    };
+
+    if (!this.state.deleteRegulationId) {
+      return status;
+    }
+
+    try {
+      const dataDeletionTaskStatus = await this.messenger.call(
+        'AnalyticsPrivacyService:checkDataDeleteStatus',
```
mcmire: This seems to be a thin wrapper around a similarly named method on the service. Is there a reason why the service can't do what this method is already doing?

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+
+    return status;
+  }
+
+  /**
+   * Get the latest delete regulation request timestamp.
+   *
+   * @returns The timestamp (in milliseconds since epoch), or undefined
+   */
+  getDeleteRegulationCreationTimestamp(): number | undefined {
```
mcmire: I see you've already got some selectors for these state properties, so do we need these controller methods?

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts
```diff
@@ -0,0 +1,375 @@
...
+   * @param saveDataRecording - Whether to save the data recording flag (default: true)
+   */
+  updateDataRecordingFlag(saveDataRecording: boolean = true): void {
+    if (saveDataRecording && !this.state.dataRecorded) {
+      this.update((state) => {
+        state.dataRecorded = true;
+      });
+
+      this.messenger.publish(
+        `${controllerName}:dataRecordingFlagUpdated`,
```
mcmire, line 329: Is there a reason why we need to publish a specific event for this state property? Usually, to respond to state property changes on the consuming side, it is enough to listen to `:stateChange` and pass a selector function. See: https://github.com/MetaMask/core/blob/main/docs/code-guidelines/controller-guidelines.md#use-selectors-to-reduce-the-scope-of-state-change-listeners

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts
```diff
@@ -0,0 +1,375 @@
...
+ */
+export type DataDeletionTaskCreatedEvent = {
+  type: `${typeof controllerName}:dataDeletionTaskCreated`;
+  payload: [IDeleteRegulationResponse];
+};
+
+/**
+ * Event emitted when the data recording flag is updated.
+ */
+export type DataRecordingFlagUpdatedEvent = {
```
mcmire, line 123: All actions and events should be prefixed with the name of the controller or service that they belong to. See: https://github.com/MetaMask/core/blob/main/docs/code-guidelines/controller-guidelines.md#use-standard-naming-scheme-for-event-types

``` suggestion
export type AnalyticsControllerDataDeletionTaskCreatedEvent = {
  type: `${typeof controllerName}:dataDeletionTaskCreated`;
  payload: [IDeleteRegulationResponse];
};

/**
 * Event emitted when the data recording flag is updated.
 */
export type AnalyticsControllerDataRecordingFlagUpdatedEvent = {
```

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
```diff
@@ -0,0 +1,375 @@
...
+  }
+
+  /**
+   * Update the data recording flag if needed.
+   * This method should be called after tracking events to ensure
+   * the data recording flag is properly updated for data deletion workflows.
+   *
+   * @param saveDataRecording - Whether to save the data recording flag (default: true)
+   */
+  updateDataRecordingFlag(saveDataRecording: boolean = true): void {
```
mcmire: Hmm. The logic here implies that if you call `updateDataRecordingFlag(false)` and `this.state.dataRecorded` is false, then nothing will happen. Why would this be useful?

packages/analytics-privacy-controller/src/selectors.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Selects the delete regulation ID from the controller state.
+ *
+ * @param state - The controller state
+ * @returns The regulation ID, or undefined if not set
+ */
+const selectDeleteRegulationId = (
+  state: AnalyticsPrivacyControllerState,
+): string | undefined => state.deleteRegulationId ?? undefined;
```
mcmire: It seems that you are converting `null` to `undefined` here. Is there a reason why this is? Why not use `undefined` to represent no value to begin with instead of `null`? (Same for `selectDeleteRegulationTimestamp`).

packages/analytics-privacy-controller/src/selectors.ts
```diff
@@ -0,0 +1,40 @@
+import type { AnalyticsPrivacyControllerState } from './AnalyticsPrivacyController';
+
+/**
+ * Selects the data recorded flag from the controller state.
+ *
+ * @param state - The controller state
+ * @returns Whether data has been recorded since the last deletion request
+ */
+const selectDataRecorded = (state: AnalyticsPrivacyControllerState): boolean =>
```
mcmire: Going along with my comment about the `dataRecorded` state property, what are your thoughts on renaming this to `selectHasCollectedDataSinceDeletionRequest`?

packages/analytics-privacy-controller/src/types.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Status values for data deletion requests from Segment API.
+ * Enum values match Segment API response values exactly.
+ */
+export enum DataDeleteStatus {
```
mcmire: What are your thoughts on using a type object rather than an enum? We are considering banning enums in `core` (and other places) because they have a lot of issues, one of which is that two enums with the same name and same content aren't assignable to each other (nominal typing), which causes unexpected type errors. You can read more here: https://github.com/MetaMask/eslint-config/issues/417

For instance, you could replace this enum with:

``` typescript
export enum DATA_DELETE_STATUSES = {
  Failed: 'FAILED',
  Finished: 'FINISHED',
  Initialized: 'INITIALIZED',
  Invalid: 'INVALID',
  NotSupported: 'NOT_SUPPORTED',
  PartialSuccess: 'PARTIAL_SUCCESS',
  Running: 'RUNNING',
  Unknown: 'UNKNOWN',
} as const;

export type DataDeleteStatus =
  (typeof DataDeleteStatus)[keyof typeof DataDeleteStatus];
```

packages/analytics-data-regulation-controller/src/types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Timestamp for deletion regulation creation (milliseconds since epoch).
+ */
+export type DataDeleteTimestamp = number | undefined;
+
+/**
+ * Regulation ID from Segment API.
+ */
+export type DataDeleteRegulationId = string | undefined;
```
mcmire: Can an ID be `undefined`? That seems interesting 🤔

packages/analytics-privacy-controller/src/AnalyticsPrivacyService.ts
```diff
@@ -0,0 +1,394 @@
...
+        new Error('Malformed response from Segment API'),
+      );
+      return {
+        status: DataDeleteResponseStatus.Failure,
+        error: 'Analytics Deletion Task Error',
+      };
+    } catch (error) {
+      log('Analytics Deletion Task Error', error);
+      return {
+        status: DataDeleteResponseStatus.Failure,
```
mcmire: Same question I posed in the controller — is there a reason why throwing an error would not achieve the same thing? Or in this case, since the service is supposed to be lower-level, is there a reason why it's useful to catch errors at all? Why  not let them bubble up?

packages/analytics-privacy-controller/src/constants.ts
mcmire, line 1: Nit: These constants seem to only be used in the service. Would it be easier to maintain if they lived there?

packages/analytics-privacy-controller/src/AnalyticsPrivacyService.ts
```diff
@@ -0,0 +1,394 @@
...
+
+      const jsonResponse =
+        (await response.json()) as GetRegulationStatusResponse;
+
+      const rawStatus = jsonResponse?.data?.data?.regulation?.overallStatus;
+      const dataDeleteStatus = Object.values(DataDeleteStatus).includes(
+        rawStatus as DataDeleteStatus,
+      )
+        ? (rawStatus as DataDeleteStatus)
+        : DataDeleteStatus.Unknown;
```
mcmire: Nit: If you had an `isDataDeleteStatus` function then you could simplify this and wouldn't need to use typecasting:

``` typescript
function isDataDeleteStatus(status: unknown): status is DataDeleteStatus {
  const dataDeleteStatuses: string[] = Object.values(DataDeleteStatus);
  return dataDeleteStatuses.includes(status);
}
```

``` suggestion
      const dataDeleteStatus = isDataDeleteStatus(rawStatus)
        ? rawStatus
        : DataDeleteStatus.Unknown;
```

packages/analytics-privacy-controller/src/AnalyticsPrivacyController.ts
mcmire: A broader question: You say in the JSDoc block "This method should be called after tracking events to ensure the data recording flag is properly updated for data deletion workflows." What do you mean by this, can you provide an example? The reason I'm asking is that it's a bit strange for this method to exist. It feels like this sort of thing should be automatic somehow.

NicolasMassart: This will be used by analytics controller when track event is called, but we had to create this in the controller first. It will not be something devs have to deal with eventually. Just a separation of concerns

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.test.ts
```diff
@@ -0,0 +1,832 @@
...
+            ),
+          ),
+      );
+
+      rootMessenger.delegate({
+        messenger: analyticsDataRegulationControllerMessenger,
+        actions: ['AnalyticsDataRegulationService:createDataDeletionTask'],
+      });
+
+      const controller = new AnalyticsDataRegulationController({
```
mcmire, line 385: Nit: Can we use `setupController` in this test to remove some of the boilerplate?  There are some other tests in this file that are also creating the controller and messengers manually?

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.test.ts
```diff
@@ -0,0 +1,832 @@
...
+        }),
+      );
+
+      rootMessenger.delegate({
+        messenger: analyticsDataRegulationControllerMessenger,
+        actions: ['AnalyticsDataRegulationService:checkDataDeleteStatus'],
+      });
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const _controller = new AnalyticsDataRegulationController({
```
mcmire, line 651: Same here. It seems if we used `setupController` we could only destructure the messenger and avoid from needing an `eslint-disable` comment.

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.test.ts
```diff
@@ -0,0 +1,832 @@
...
+        eventListener,
+      );
+
+      rootMessenger.call(
+        'AnalyticsDataRegulationController:updateDataRecordingFlag',
+      );
+
+      expect(eventListener).not.toHaveBeenCalled();
+    });
+  });
```
mcmire, line 831: Do we need this `describe`? Typically we don't need to test that the `stateChange` event is triggered for controllers, as this behavior is already tested in the BaseController tests.

``` suggestion
```

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationService.test.ts
```diff
@@ -0,0 +1,539 @@
...
+
+describe('AnalyticsDataRegulationService', () => {
+  let clock: SinonFakeTimers;
+  const segmentSourceId = 'test-source-id';
+  const segmentRegulationsEndpoint = 'https://proxy.example.com/v1beta';
+
+  beforeEach(() => {
+    clock = useFakeTimers();
+    cleanAll();
+    disableNetConnect();
```
mcmire, line 23: Do you need to do this? It should already be happening for all tests in `tests/setupAfterEnv/nock.ts`.

``` suggestion
```

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationService.test.ts
```diff
@@ -0,0 +1,539 @@
...
+  });
+
+  afterEach(() => {
+    clock.restore();
+    cleanAll();
+    enableNetConnect();
```
mcmire, line 29: Do you need to do this? It should already be happening for all tests in `tests/setupAfterEnv/nock.ts`.

``` suggestion
```

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationService.test.ts
```diff
@@ -0,0 +1,539 @@
...
+          data: {
+            data: {
+              regulateId,
+            },
+          },
+        });
+
+      const { rootMessenger } = getService();
+
+      const response = await rootMessenger.call(
```
mcmire, line 49: Nit: Would it make more sense to use the messenger here rather than the root messenger, since the messenger represents the service?

``` suggestion
      const { messenger } = getService();

      const response = await messenger.call(
```

(Same goes for the rest of these tests.)

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationService.test.ts
```diff
@@ -0,0 +1,539 @@
...
+          'test-analytics-id',
+        ),
+      ).rejects.toThrow('Creating data deletion task failed');
+
+      expect(onRetryListener).toHaveBeenCalled();
+    });
+  });
+
+  describe('onBreak', () => {
+    it('calls break listener when circuit breaker opens after multiple failures', async () => {
```
mcmire, line 378: Nit: The "it" here is the `onBreak` listener, so maybe this would be more accurate?

``` suggestion
    it('is called when circuit breaker opens after multiple failures', async () => {
```

(Similar for the other `onBreak` and `onDegraded` tests)

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts
```diff
@@ -0,0 +1,334 @@
...
+    regulateId: string;
+  }> {
+    if (!this.#analyticsId || this.#analyticsId.trim() === '') {
+      const error = new Error(
+        'Analytics ID not found. You need to provide a valid analytics ID when initializing the AnalyticsDataRegulationController.',
+      );
+      log('Analytics Deletion Task Error', error);
+      throw error;
+    }
+
```
mcmire, line 255: Is this runtime check necessary? There are other instances within `core` where we perform runtime checks, but I think a lot of those instances are left over from when the code used to be written in JavaScript. I believe we have been largely relying on TypeScript to verify that required inputs have been provided. Of course if you think it's necessary, then that's fine, but I wanted to double-check first.

``` suggestion
```

packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts
```diff
@@ -0,0 +1,334 @@
...
+   * Check the latest delete regulation status.
+   *
+   * @returns Promise containing the timestamp, delete status and collected data flag
+   */
+  async checkDataDeleteStatus(): Promise<DeleteRegulationStatus> {
+    // Capture all state values before async call to ensure consistency
+    // in case createDataDeletionTask() completes concurrently
+    const { deleteRegulationId } = this.state;
+    const { deleteRegulationTimestamp } = this.state;
+    const { hasCollectedDataSinceDeletionRequest } = this.state;
```
mcmire, line 288: Can these be grouped together?

``` suggestion
    const {
      deleteRegulationId,
      deleteRegulationTimestamp,
      hasCollectedDataSinceDeletionRequest
    } = this.state;

File: packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts (hunk 9 above)
Comment (mcmire) [reply packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts:329]: @NicolasMassart Just checking on this again, to see if this is necessary?

File: packages/analytics-data-regulation-controller/src/AnalyticsDataRegulationController.ts [hunk 27]
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    // Service validates and throws on all errors, so if we reach here, the response
+    // is guaranteed to be a success response with dataDeleteStatus present
+    const dataDeletionTaskStatus = await this.messenger.call(
+      'AnalyticsDataRegulationService:checkDataDeleteStatus',
+      deleteRegulationId,
+    );
+
+    status.dataDeletionRequestStatus = dataDeletionTaskStatus.dataDeleteStatus;
```
mcmire, line 307: Nit: I think this comment might be misplaced? Maybe it was meant to go after the `call`.

``` suggestion
    const dataDeletionTaskStatus = await this.messenger.call(
      'AnalyticsDataRegulationService:checkDataDeleteStatus',
      deleteRegulationId,
    );

    // Service validates and throws on all errors, so if we reach here, the response
    // is guaranteed to be a success response with dataDeleteStatus present
    status.dataDeletionRequestStatus = dataDeletionTaskStatus.dataDeleteStatus;

mcmire (approved):
Had some more comments, but they are fairly minor at this point, implementation looks pretty good to me. I'll go ahead and approve.

PR #7659, opened 2026-01-19
fix: set isExternalSign to true for delegated transactions

khanti42:
## Explanation

EIP-7702 delegated transactions were incorrectly using the user's EOA nonce, causing nonce conflicts and preventing transactions from appearing in the activity tab. In EIP-7702, the delegation address manages its own nonce sequence, not the user's EOA.

This PR sets `isExternalSign` for EIP-7702 txs. 

## Checklist

---

> [!NOTE]
> Addresses incorrect nonce handling for EIP-7702 delegated/batch transactions.
> 
> - Set `isExternalSign = true` when `nestedTransactions` are present to bypass EOA nonce assignment for EIP-7702 batches
> - Record `delegationAddress` for metrics without using it to infer `isExternalSign`
> - Update `CHANGELOG.md` under Fixed to reflect the behavior change
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 99b756f7d1246a6d701733b43feba6ba514f2b1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/transaction-controller/src/utils/nonce.ts
```diff
@@ -31,6 +32,13 @@ export async function getNextNonce(
     return [undefined, undefined];
   }
 
+  // Skip nonce assignment for delegated transactions (EIP-7702)
+  // The delegation address manages its own nonce sequence externally
```
matthewwalsh0: Even if the account is upgraded, it can still sign and submit transactions normally meaning the nonces are valid.

This just records the current delegated address for any transaction, for metrics.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1363,6 +1365,7 @@ export class TransactionController extends BaseController<
           deviceConfirmedOn,
           disableGasBuffer,
           id: random(),
+          isExternalSign: isEIP7702Batch,
```
matthewwalsh0, line 1368: Sorry, this is the same issue as before.

EIP-7702 batch transactions can still be signed locally by the EOA itself and more typically are.

The only exceptions to that are when the gas station EIP-7702 flow is used as it's signed by our relay.

If we're missing a scenario when a transaction is signed externally, then we need to update `isExternalSign` externally (as we do for gas station EIP-7702 currently) or trigger it more specifically internally, assuming we have that context internally without unnecessary coupling.

PR #7670, opened 2026-01-20
fix(keyring-controller): Only drop target keyring on `removeAccount`

matthiasgeihs:
## Explanation

Before, `removeAccount` would clear all empty keyrings from the keyring controller, leading to undesirable side effects. (E.g., when a keyring has been initialized as empty, it would be dropped here)

Changing `removeAccount` to only drop the target keyring if its last account has been removed. (And no longer touch the other keyrings.)

---
Also fixes an issue where Snap Keyring accounts would not be removed in order (because their `removeAccount` function is async, but this was previously not awaited).

## Checklist

---

> [!NOTE]
> Refines account removal behavior and hardens keyring lookups.
> 
> - Update `removeAccount` to await `keyring.removeAccount` and remove only the targeted keyring when it becomes empty; delete `#removeEmptyKeyrings`
> - Add `#findKeyringIndexForAccount` with case-insensitive address matching; prevents unintended primary keyring deletion with non-normalized addresses
> - Simplify "No keyring found" error handling; update `getKeyringForAccount` to use the new lookup helper
> - Expand tests: non-normalized address protection, preserve unrelated empty keyrings, and ensure async `removeAccount` completes before keyring removal
> - Changelog updated under Changed/Fixed
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 074a6c110224302fa26da68ba8d385d6ea1696c1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/keyring-controller/CHANGELOG.md
+++ b/packages/keyring-controller/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Prevent the accidental removal of unrelated empty keyrings when deleting an account ([#7670](https://github.com/MetaMask/core/pull/7670))
   - Empty keyrings were removed during account deletion, regardless of which account was being targeted.
 
+### Fixed
+
+- Fixed a bug where `removeAccount` would not prevent deletion of the primary keyring because of missing address normalization ([#7670](https://github.com/MetaMask/core/pull/7670))
+
 ## [25.0.0]
 
 ### Added


packages/keyring-controller/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Upgrade `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Do not remove other empty keyrings when removing an account ([#7670](https://github.com/MetaMask/core/pull/7670))
```
mikesposito: nit:
```suggestion
- Prevent the accidental removal of unrelated empty keyrings when deleting an account ([#7670](https://github.com/MetaMask/core/pull/7670))
  - Empty keyrings were removed during account deletion, regardless of which account was being targeted.

File: packages/keyring-controller/CHANGELOG.md [hunk 2]
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Await the `removeAccount` method of the keyring ([#7670](https://github.com/MetaMask/core/pull/7670))
```
mikesposito: nit: this seems to be an internal detail, transparent to consumers:

```suggestion
```

packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -1645,6 +1645,88 @@ describe('KeyringController', () => {
...
+      it('should not remove other empty keyrings when removing an account', async () => {
+        await withController(async ({ controller }) => {
+          // Import an account, creating a Simple keyring with 1 account
+          const importedAccount = await controller.importAccountWithStrategy(
+            AccountImportStrategy.privateKey,
+            [privateKey],
+          );
+
+          // Add an empty Simple keyring (no accounts)
+          await controller.addNewKeyring(KeyringTypes.simple);
```
mikesposito, line 1669: Passing a custom vault to `withController` would avoid the need of using `importAccountWithStrategy` and `addNewKeyring`, making the test scenario clearer as we can directly see what's in the vault when the test is run. Though this works as well, so you can consider this a NIT suggestion

mikesposito (approved):
Some small suggestions, but overall looks good.

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1260,15 +1260,18 @@ export class KeyringController<
         );
       }
 
-      // The `removeAccount` method of snaps keyring is async. We have to update
-      // the interface of the other keyrings to be async as well.
-      // FIXME: We do cast to `Hex` to makes the type checker happy here, and
-      // because `Keyring<State>.removeAccount` requires address to be `Hex`. Those
-      // type would need to be updated for a full non-EVM support.
-      keyring.removeAccount(address as Hex);
+      // FIXME #1: We do cast to `Hex` to make the type checker happy here, and
```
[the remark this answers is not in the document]
Gudahtt: Seems like a legitimate pre-existing bug that could break our attempt to detect whether an account is part of the primary keyring or not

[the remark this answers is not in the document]
matthiasgeihs: That is actually a critical bug.

It was already there previously, but since the keyringIndex was only used to detect the primary keyring, it wasn't that critical as it would have been here. Still worth fixing in any case.

PR #7688, opened 2026-01-21
feat: Add INK support

bizk:
## Explanation

We want to enable INK into price API

## Checklist

---

> [!NOTE]
> **Low Risk**
> Small, additive change limited to chain-ID mapping tables and a changelog update; main risk is incorrect chain/contract identifier causing failed price/balance lookups on Ink.
> 
> **Overview**
> Adds Ink Mainnet (chain `0xdef1`) support to `@metamask/assets-controllers` so `CodefiTokenPricesServiceV2` can fetch native ETH spot prices for that network.
> 
> This is done by registering Ink’s CAIP-19 native asset identifier in `token-prices-service/codefi-v2.ts` and adding the Multicall3 contract address for `0xdef1` in `multicall.ts`, with a changelog entry documenting the new support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ed3e58dc22e4b9a438489862fe5be103f33c0b7b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/controller-utils/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add INK (0xed88) mapping to eip155:57073/erc20:0x0000000000000000000000000000000000000000 ([#7688](https://github.com/MetaMask/core/pull/7688))
+
```
gabrieledm: Is this correct ?
From my understanding the `CHANGELOG` should be updated only if there there are some changes in the related package.

packages/assets-controllers/src/multicall.test.ts
```diff
@@ -276,6 +276,30 @@ describe('multicall', () => {
         'Contract call failed',
       );
     });
+
```
gabrieledm: Is this test needed ?
From my understanding if the business logic is not updatd, then is not needed to add tests for new supported network.

bizk: I don't think it is needed but, is it bad to add a new test? what do you suggest?

packages/controller-utils/CHANGELOG.md
bizk: no it isn't my bad

packages/assets-controllers/src/multicall.test.ts
gabrieledm: Maybe the best thing to do is to follow the Tech Guide and it says that if there isn't any change in the business logic, but only in configuration, the addition of new tests it's not needed

bizk: understood will remove

gabrieledm (dismissed):
LGTM

packages/assets-controllers/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add INK (0xdef1) mapping to eip155:57073/slip44:60 ([#7688](https://github.com/MetaMask/core/pull/7688))
```
mikesposito: True that we added an object property, but what we are doing from the perspective of a package consumer is _changing_ the `SPOT_PRICES_SUPPORT_INFO` exported variable. Should this line go under the CHANGED category?

mikesposito: hmm actually, the constant I was referring to doesn't seem to be exported by the package, so in theory this change is transparent to consumers

mcmire: True. What effect does this have on the service class? Were there errors in attempting to use this chain before? Does it change the return type of something? etc.

bizk: I followed the steps to add a new supported chain as we have been doing so with the rest of the other chains. 

Could you be more specific in your concerns please?

mcmire: The purpose of a changelog — at least how we are using them at MetaMask — is not to communicate changes to the implementation of a package, but to document changes to the public API. It's the public API that you interact with when you use a package.

That's why Michele is making a suggestion for how this change is represented in the changelog, and why I am asking what effect this has. What does this do?

To me "Add INK (0xdef1) mapping to eip155:57073/slip44:60" is an implementation detail. It seems that what we are really doing is making it so that you can fetch prices for ETH on Ink Mainnet (chain `0xdef1`). And we're also updating `CodefiTokenPricesServiceV2`. So this feels like a better description:

``` suggestion
- `CodefiTokenPricesServiceV2` now supports fetching prices of ETH on Ink Mainnet (chain `0xdef1`) ([#7688](https://github.com/MetaMask/core/pull/7688))

bizk: Oh... I see.. I makes total sense. Thank you.

salimtb (approved):

PR #7716, opened 2026-01-24
feat: use dynamic supported networks from Price API

salimtb:
### Current State

The `CodefiTokenPricesServiceV2` in `assets-controllers` uses a hardcoded list (`SPOT_PRICES_SUPPORT_INFO`) to determine which blockchain networks are supported by the Price API's `/v3/spot-prices` endpoint. This approach has several drawbacks:

1. **Maintenance overhead**: Every time a new network is added to the Price API, a code change is required to update the hardcoded list
2. **Deployment delays**: New network support requires a release cycle before users can benefit from it
3. **Potential staleness**: The hardcoded list can become out of sync with the actual API capabilities

Additionally, the native asset CAIP-19 identifiers (e.g., `eip155:1/slip44:60` for ETH) were also hardcoded, duplicating data that already exists in `NetworkEnablementController.state.nativeAssetIdentifiers`.

### Solution

This PR introduces dynamic fetching of supported networks from the Price API's `/v2/supportedNetworks` endpoint:

1. **New API integration**: Added `fetchSupportedNetworks()` function that fetches from `/v2/supportedNetworks` and caches the response
2. **Graceful fallback**: If the API call fails, the existing hardcoded list is used as a fallback
3. **Background refresh**: The supported networks list is refreshed in the background during each `fetchTokenPrices()` call
4. **Native asset integration**: Added `setNativeAssetIdentifiers()` method to receive native asset mappings from `NetworkEnablementController`, eliminating duplication
5. **TokenRatesController integration**: The controller now automatically retrieves and passes `nativeAssetIdentifiers` from `NetworkEnablementController` to the token prices service

### Changes Summary

- **`codefi-v2.ts`**: Added dynamic network fetching, caching, and native asset identifier support
- **`abstract-token-prices-service.ts`**: Added `NativeAssetIdentifiersMap` type and optional `setNativeAssetIdentifiers()` method to interface
- **`TokenRatesController.ts`**: Integrated with `NetworkEnablementController` to pass native asset identifiers
- **Dependencies**: Added `@metamask/network-enablement-controller` as a dependency

## References

- Related to Price API v2 supportedNetworks endpoint: `https://price.api.cx.metamask.io/v2/supportedNetworks`
- Uses `NetworkEnablementController.state.nativeAssetIdentifiers` for CAIP-19 native token mappings

## Checklist

---

> [!NOTE]
> Introduces dynamic network support and native asset identifier integration for token pricing.
> 
> - Add `fetchSupportedNetworks()`, `getSupportedNetworks()`, and `resetSupportedNetworksCache()` in token-prices-service (fallback to hardcoded list; background refresh during `fetchTokenPrices`)
> - Update `CodefiTokenPricesServiceV2` to:
>   - accept `setNativeAssetIdentifiers()` and use CAIP-19 IDs for native tokens
>   - validate chains against the dynamic networks list
>   - expose `updateSupportedNetworks()`
> - `TokenRatesController` now calls `NetworkEnablementController:getState` at init to pass `nativeAssetIdentifiers` to the token prices service
> - Export `NativeAssetIdentifiersMap`; add `@metamask/network-enablement-controller` dependency and TS references; expand tests for new behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 863782eebd42fc037c786b4826e6d103f5a8aed9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/TokenRatesController.test.ts
+++ b/packages/assets-controllers/src/TokenRatesController.test.ts
@@ -112,9 +112,8 @@ describe('TokenRatesController', () => {
     it('should not fail if tokenPricesService does not have setNativeAssetIdentifiers', async () => {
       const tokenPricesService = buildMockTokenPricesService();
       // Explicitly remove setNativeAssetIdentifiers to simulate an old service
-      delete (
-        tokenPricesService as Partial<AbstractTokenPricesService>
-      ).setNativeAssetIdentifiers;
+      delete (tokenPricesService as Partial<AbstractTokenPricesService>)
+        .setNativeAssetIdentifiers;
 
       await withController(
         {


packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -542,26 +684,53 @@ export class CodefiTokenPricesServiceV2
     assets: EvmAssetAddressWithChain<SupportedChainId>[];
     currency: SupportedCurrency;
   }): Promise<EvmAssetWithMarketData<SupportedChainId, SupportedCurrency>[]> {
+    // Refresh supported networks in background (non-blocking)
+    // This ensures the list stays fresh during normal polling
+    // Note: fetchSupportedNetworks handles errors internally and always resolves
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    fetchSupportedNetworks();
```
juanmigdr: I dont understand why we are calling the fetchSupportedNetworks function here. That function is not storing anything in state so why would we need to call it here if we are not making use of the response at all

salimtb: `fetchSupportedNetworks()` updates a module-level cache (`lastFetchedSupportedNetworks`). The background call keeps this cache fresh so that subsequent calls to `getSupportedChainIdsV3AsHex()` return up-to-date data.
This is the same pattern used for `fetchSupportedCurrencies()` in `fetchExchangeRates()`.

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
```diff
@@ -343,20 +345,132 @@ export const SUPPORTED_CHAIN_IDS = Object.keys(
...
+  return numberToHex(parseInt(match[1], 10));
+}
+
+/**
+ * Fetches the list of supported networks from the API.
+ * Falls back to the hardcoded list if the fetch fails.
+ *
+ * @returns The supported networks response.
+ */
+export async function fetchSupportedNetworks(): Promise<SupportedNetworksResponse> {
```
Kriys94, line 390: You could find the call in core-backend

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
juanmigdr: Oh missed that, thanks! Why did you change it to be blocking instead of non-blocking now?

packages/assets-controllers/src/token-prices-service/codefi-v2.ts
Prithpal-Sooriya, line 1: Should we finally rename codefi-v2 😄  

We can do this later.

PR #7725, opened 2026-01-26
refactor: refactor intent order management

oscarwroche:
## Explanation

This PR moves intent-related functions to their own folder in `bridge-status-controller` . This is to prepare moving more intent-specific logic from the `metamask-mobile` repo to the `core` repo. 

## Checklist

---

> [!NOTE]
> No code changes in this PR; the diff is empty.
> 
> - No files modified, added, or removed
> - No behavioral, API, or build impacts
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0ad33a138a3ae331ed69e2a0c80a7b7affde266. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/bridge-status-controller/CHANGELOG.md
+++ b/packages/bridge-status-controller/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [65.0.0]
+
+### Changed
+
+- Bump `@metamask/bridge-controller` from `^64.8.2` to `^65.0.0` ([#7731](https://github.com/MetaMask/core/pull/7731))
+
 ## [64.4.5]
 
 ### Changed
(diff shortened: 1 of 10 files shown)


packages/bridge-status-controller/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump `@metamask/bridge-controller` from `^64.8.1` to `^64.8.2` ([#7722](https://github.com/MetaMask/core/pull/7722))
+- Move intent handling into a dedicated intent module with an internal status manager ([#7725](https://github.com/MetaMask/core/pull/7725))
```
mcmire: Is this visible to consuming code, i.e., does this change the public API of this package? If not, then you don't need to add a changelog entry (and can add the `no-changelog` label).

``` suggestion
```

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -178,6 +176,39 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
         config?.customBridgeApiBaseUrl ?? BRIDGE_PROD_API_BASE_URL,
     };
     this.#trace = traceFn ?? (((_request, fn) => fn?.()) as TraceCallback);
+    this.#intentStatusManager = new IntentStatusManager({
```
micaelae: Nice work organizing the intent code - seems much easier to find and test

I'm concerned that there's significant duplication between BridgeStatusController and IntentStatusManager, specifically in the `fetchBridgeTxStatus` method. Both implement nearly identical logic for history updates, "final" status handling, event tracking. The real differences are just API endpoints and response formats.

Risks of current approach:
- Future changes need to be made in two places (easy to miss one or have them diverge)
- Bug fixes might only get applied to one path
- If we need to change how final status is handled or add new event tracking, we'd need to remember to update both implementations
- The duplication makes it harder to see what's actually different vs what's the same
- I see similar issues in `submitIntent`

An example of a bug-fix that has not been applied to intents due to the current divergence: `actionId` history lookups that enable event publishing if the approval fails before the swap trade is attempted. Imo we need to work towards making these code paths share common logic where possible instead

Suggestions:
- extract shared utilities with updateHistoryFromStatus() and handleFinalStatus() functions
- keep only the API-specific code (fetching, validation, mapping, submitIntent fetch call) in separate modules

PR #7729, opened 2026-01-27
fix: correctly validate erc20-token-revocation terms when decoding permission

jeffsmale90:
## Explanation

Decoding a permission is performed in two steps: first we identify the permission type, based on the required and optional caveat enforcers. Second we decode the terms of the enforcers specific for the permission type into the permission data. The decoded permission data is then shown to the user to confirm.

Because the `erc20-token-revocation` permission doesn't have permission data (aside from the expiry) to show the user, the encoded terms are considered _inherent_ to the permission. It's important that we validate these terms to ensure that the permission being signed is valid.

Previous to this PR we validated the caveat _types_ included in the permission, but not the terms of the caveats.

Note: I considered breaking out the data decoding for each permission type into it's own file, but decided to keep it inline in order to reduce the amount of change in this PR.

Fixes: #7738

## Checklist

---

> [!NOTE]
> Strengthens decoding for `erc20-token-revocation` by validating inherent caveat terms.
> 
> - Enforces `AllowedCalldataEnforcer` includes `approve` selector and zero-amount constraints; `ValueLteEnforcer` terms must be zero (`0x00…00`) in `decodePermission.ts`
> - Adds tests for valid/invalid term combinations and updates fixtures/imports in `decodePermission.test.ts`
> - Updates `CHANGELOG.md` under Unreleased → Fixed
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 859df6cc6cb3a1abb3d302782096f1c8aa8e42be. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -267,6 +269,55 @@ export const getPermissionDataAndExpiry = ({
...
+
+      // 0 amount starting at index 24
+      const ERC20_APPROVE_ZERO_AMOUNT_TERMS =
+      '0x00000000000000000000000000000000000000000000000000000000000000240000000000000000000000000000000000000000000000000000000000000000' as const;
+
+      const allowedCalldataCaveats = checksumCaveats.filter(
+        (caveat) => caveat.enforcer === allowedCalldataEnforcer,
+      );
+
+      if (allowedCalldataCaveats.length !== 2) {
```
MoMannn: this is probably not necessary, the check that identifies this is `erc20-token-revocation` already validates that there need to be 2 `allowedCalldataCaveats`.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -267,6 +269,55 @@ export const getPermissionDataAndExpiry = ({
       break;
     }
     case 'erc20-token-revocation': {
+      // 0 value for ValueLteEnforcer
+      const ZERO_32_BYTES =
+      '0x0000000000000000000000000000000000000000000000000000000000000000' as const;
```
MoMannn: This should probably be moved to some constants in the future so it can be reused.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -267,6 +269,55 @@ export const getPermissionDataAndExpiry = ({
       break;
     }
     case 'erc20-token-revocation': {
+      // 0 value for ValueLteEnforcer
+      const ZERO_32_BYTES =
+        '0x0000000000000000000000000000000000000000000000000000000000000000' as const;
```
MoMannn, line 274: This should probably be moved to some constants in the future so they are reusable.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
```diff
@@ -267,6 +269,55 @@ export const getPermissionDataAndExpiry = ({
...
+      }
+
+      const valueLteTerms = getTermsByEnforcer({
+        caveats: checksumCaveats,
+        enforcer: valueLteEnforcer,
+      });
+
+      if (valueLteTerms !== ZERO_32_BYTES) {
+        throw new Error('Invalid ValueLteEnforcer terms: maxValue must be 0');
+      }
```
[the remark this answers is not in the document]
jeffsmale90: This is fine - there will never be a capital prefix, but there might be capital hexadecimal values.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
jeffsmale90: Yeah, good call! I've removed that check.

packages/gator-permissions-controller/src/decodePermission/decodePermission.ts
jeffsmale90: Agreed - I considered moving this entire block into a separate file but decided to defer that to minimise change in this PR. We should move those consts at the same time.

PR #7746, opened 2026-01-27
feat: adds ai digest controller

Bigshmow:
## Explanation

Adds AiDigestController to the @metamask/ai-controllers package for fetching and caching AI-generated asset digests.

What it does:
- AiDigestService - Fetches digest data from GET /digests/assets/:assetId/latest
- AiDigestController - Manages state, caching (10 min TTL, 50 entry max), and FIFO eviction
- Follows existing controller patterns (BaseController, messenger actions, etc.)

Design decisions:
- DigestData type mirrors the API response structure from va-mmcx-digest-api - we'll iterate inline as the API evolves
- baseUrl is required (no default) - consuming apps must provide their API endpoint
- Eviction handles stale entries, error/loading states, and invalid entries from corrupted persistence

This is a V1 implementation - the API response shape may evolve as the backend is finalized.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds new persisted controller state and a network-backed fetch flow; correctness of TTL/FIFO eviction and API error handling affects UI behavior and storage growth, but does not touch auth or funds.
> 
> **Overview**
> Adds a new `AiDigestController` that fetches asset digests via an injected `DigestService`, persists them in controller state, and exposes messenger actions for `fetchDigest`, `clearDigest`, and `clearAllDigests`.
> 
> Implements caching with a 10-minute TTL (`CACHE_DURATION_MS`) and a 50-entry max (`MAX_CACHE_ENTRIES`) with stale + FIFO eviction, and introduces `AiDigestService` to call `GET /digests/assets/:assetId/latest` with explicit error handling.
> 
> Updates `@metamask/ai-controllers` exports, tests (replaces the placeholder `greeter` test), adds `base-controller`/`messenger` deps and TS project references, and notes the addition in the changelog.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6a24fd38dde333314a9d93528bdc5809b0bacf38. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/ai-controllers/src/AiDigestController.ts
```diff
@@ -0,0 +1,225 @@
...
+      `${controllerName}:clearDigest`,
+      this.clearDigest.bind(this),
+    );
+    this.messenger.registerActionHandler(
+      `${controllerName}:clearAllDigests`,
+      this.clearAllDigests.bind(this),
+    );
+  }
+
+  async fetchDigest(coingeckoSlug: string): Promise<DigestEntry> {
```
xavier-brochard: I'm tempted to rename this `assetId` instead of `coingeckoSlug`.

We would loose expressiveness, but at least it wouldn't require a rename once we stop using CoinGecko as data source in the backend. Ideally we use internal data sources, like Token API DB, and Price API DB.

packages/ai-controllers/src/AiDigestController.ts
```diff
@@ -0,0 +1,225 @@
...
+      return entry;
+    } catch (error) {
+      const entry: DigestEntry = {
+        asset: coingeckoSlug,
+        status: DIGEST_STATUS.ERROR,
+        error: error instanceof Error ? error.message : 'Unknown error',
+      };
+
+      this.update((state) => {
+        state.digests[coingeckoSlug] = entry;
```
xavier-brochard: Oh this is smart! 🦉 Let's us troubleshoot by checking the MM state

packages/ai-controllers/src/AiDigestController.ts
```diff
@@ -0,0 +1,225 @@
...
+    this.update((state) => {
+      delete state.digests[coingeckoSlug];
+    });
+  }
+
+  clearAllDigests(): void {
+    this.update((state) => {
+      state.digests = {};
+    });
+  }
```
xavier-brochard: Do you expect a use case for these? 🤔

packages/ai-controllers/src/AiDigestController.ts
```diff
@@ -0,0 +1,225 @@
...
+    const entries = Object.entries(state.digests);
+    const keysToDelete: string[] = [];
+
+    // Collect fresh entries (with fetchedAt) and mark stale/error/loading entries for deletion
+    const freshEntries: [string, DigestEntry & { fetchedAt: number }][] = [];
+    for (const [key, entry] of entries) {
+      // Evict error and loading entries to prevent unbounded accumulation
+      if (
+        entry.status === DIGEST_STATUS.ERROR ||
+        entry.status === DIGEST_STATUS.LOADING
```
xavier-brochard: If we are to evict them at next eviction cycle (= during next update), should we even bother storing `ERROR` and `LOADING` entries in the state?

packages/ai-controllers/src/AiDigestController.test.ts
```diff
@@ -0,0 +1,268 @@
...
+  it('returns default state', () => {
+    expect(getDefaultAiDigestControllerState()).toStrictEqual({ digests: {} });
+  });
+
+  it('fetches and caches a digest', async () => {
+    const mockService = { fetchDigest: jest.fn().mockResolvedValue(mockData) };
+    const controller = new AiDigestController({
+      messenger: createMessenger(),
+      digestService: mockService,
+    });
```
xavier-brochard, line 39: Nitpick, but we could save a fair amount of repeated code by factorizing the setup:
```
let mockService: DigestService
let controller: AiDigestController

beforeEach() {
  mockService = { fetchDigest: jest.fn() };
  controller = new AiDigestController({
      messenger: createMessenger(),
      digestService: mockService,
    });
}


it('does stuff', () => {
   jest.spyOn(mockService, 'fetchDigest').mockResolvedValue(...); // mock however we need for this test case
  ...
}) 
```

packages/ai-controllers/src/AiDigestController.ts
Bigshmow: agreed, I'll update

packages/ai-controllers/src/AiDigestController.ts
Bigshmow: These are just common patterns really.  Some example usage: `clearAllDigests` for logout/account-switch cleanup, `clearDigest` for a forced refetch CTA.  Not exposed in UI _yet_ so if we'd rather not add them I understand.

packages/ai-controllers/src/AiDigestController.ts
Bigshmow: That's valid.  `LOADING` was stored so UI could show per-asset loading state, but the client can handle that locally.  `ERROR` was stored for debugging, but errors are more useful in the API/Grafana than in client.  I'll simplify such that the controller will only store successful digests, making it a pure cache.  `LOADING`/`ERROR` can be handled by the caller.

packages/ai-controllers/src/AiDigestController.test.ts
Bigshmow: Good point, happy to refactor to shared beforeEach.  Tbh these were new tests I wrote alongside claude to get us to that 100% mark.

packages/ai-controllers/src/AiDigestController.ts
xavier-brochard: Ahh I see. Hummm, I would think it's safer to remove them as long as we don't need them (don't allow external code to tamper with the internal controller state), and we can easily add them back later on if needed. WYT?

PR #7759, opened 2026-01-28
feat: add network fallback mechanism to IncomingTransactionHelper

jpuri:
## Explanation

In incoming transaction helper if any chain goes down on web sockets, start polling for it.

## References

Fixes [#12345](https://github.com/MetaMask/MetaMask-planning/issues/6834)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes when/what chains are polled for incoming transactions and adds new event-driven state around network availability, which could impact transaction discovery if chain ID conversion or status handling is off.
> 
> **Overview**
> When WebSocket-based incoming transaction retrieval is enabled, `IncomingTransactionHelper` now listens for `AccountActivityService:statusChanged` and **starts/stops fallback polling** based on which supported networks are reported *down*, polling only the affected chains.
> 
> This threads an optional `chainIds` filter through `RemoteTransactionSourceRequest` and updates `AccountsApiRemoteTransactionSource` to query only those chains, plus adds tests and updates `TransactionController`’s allowed events/types to include the new status event (and updates the changelog).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 72bf51c88c7b793cbc146b9a1e689f024aec9e8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
```diff
@@ -155,7 +183,7 @@ export class IncomingTransactionHelper {
     }
 
     this.#onInterval().catch((error) => {
-      log('Initial polling failed', error);
+      log('Polling failed', error);
```
matthewwalsh0: This was intentional to distinguish that the first attempt failed, rather than a future interval based one.

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
```diff
@@ -383,4 +419,70 @@ export class IncomingTransactionHelper {
...
+    if (!this.#useWebsockets) {
+      return;
+    }
+
+    let hasChanges = false;
+
+    for (const caip2ChainId of chainIds) {
+      const hexChainId = this.#caip2ToHex(caip2ChainId);
+
+      if (!hexChainId || !SUPPORTED_CHAIN_IDS.includes(hexChainId)) {
```
matthewwalsh0, line 451: Are we adding the dynamic supported chains via `/supportedNetworks` in a separate ticket?

matthewwalsh0: Can we add a `log` here for helpful debug, maybe separately if chain ID not recognised or if not supported?

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
```diff
@@ -383,4 +419,70 @@ export class IncomingTransactionHelper {
...
+        const index = this.#chainsToPoll.indexOf(hexChainId);
+        if (index !== -1) {
+          this.#chainsToPoll.splice(index, 1);
+          hasChanges = true;
+          log('Supported network came up, removed from polling list', {
+            chainId: hexChainId,
+          });
+        }
+      } else if (!this.#chainsToPoll.includes(hexChainId)) {
+        // status === 'down'
```
matthewwalsh0: Should we be explicit in case they ever add more statuses?

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
```diff
@@ -383,4 +419,70 @@ export class IncomingTransactionHelper {
...
+        this.#chainsToPoll.push(hexChainId);
+        hasChanges = true;
+        log('Supported network went down, added to polling list', {
+          chainId: hexChainId,
+        });
+      }
+    }
+
+    if (!hasChanges) {
+      return;
```
matthewwalsh0, line 484: Another `log` for good debug?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
jpuri: PR is updated

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
jpuri: PR is updated

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
jpuri: For dynamic chain support yes a separate task will be better as it will require changes in AccountsApiRemoteTransactionSource

I added log to the PR

packages/transaction-controller/src/helpers/IncomingTransactionHelper.ts
jpuri: PR is updated

matthewwalsh0 (dismissed):

PR #7764, opened 2026-01-28
fix(AssetController): refactor Snap data source

Kriys94:
## Explanation

This PR simplifies and cleans up `SnapDataSource` by:

1. **Dynamic snap discovery via PermissionController**: Instead of hardcoded snap registry entries, `SnapDataSource` now dynamically discovers installed snaps that have the `endowment:keyring` permission and extracts their supported chain IDs from permission caveats. This makes the data source more flexible and extensible for future snaps.

2. **Simplified fetch routing**: Changed from complex chain-based routing to account-based routing using `account.metadata.snap.id`. This aligns with how `MultichainBalancesController` handles snap accounts and is more direct since each account already knows which snap it belongs to.

3. **Added `KeyringClient`**: Replaced manual `SnapController:handleRequest` calls with `KeyringClient` from `@metamask/keyring-snap-client`, which provides cleaner typed methods (`listAccountAssets`, `getAccountBalances`). This follows the same pattern used in `MultichainBalancesController`.

4. **Removed unused/redundant code**:
   - Removed `DEFAULT_SNAP_POLL_INTERVAL` constant (never used)
   - Removed `#groupChainsBySnapId` method (no longer needed with account-based routing)
   - Removed `#fetchFromSnapById` method (inlined into `fetch`)
   - Removed `#accountSupportsChain` method (unused after simplification)
   - Removed redundant chain filtering in `fetch` (already done by middleware/subscribe callers)
   - Removed unused `InternalAccount` import

5. **Used proper types**: Replaced inline type `Record<string, { amount: string; unit: string }>` with `GetAccountBalancesResponse` and `Balance` from `@metamask/keyring-api`.

**Dependencies added**:
- `@metamask/keyring-api`: For `Balance` and `CaipAssetType` types
- `@metamask/keyring-snap-client`: For `KeyringClient` to make typed snap calls

## References

- Related to the `SnapDataSource` dynamic snap discovery refactoring
- Pattern aligned with `MultichainBalancesController` and `MultichainAssetsController` implementations

## Checklist

---

> [!NOTE]
> **High Risk**
> High risk because it changes SnapDataSource discovery/routing behavior and removes the `snapProvider` dependency and many exported constants/types, which may break consumers and alter which chains/accounts get queried for balances.
> 
> **Overview**
> **Refactors `SnapDataSource` to use dynamic snap discovery** instead of a hardcoded Solana/Bitcoin/Tron registry: it now queries `SnapController:getRunnableSnaps` and `PermissionController:getPermissions` to build a `chainToSnap` map from the `endowment:assets` `chainIds` caveat, and keeps `activeChains` empty until discovery succeeds (re-running on `PermissionController:stateChange`).
> 
> **Simplifies balance fetching to be account-driven** by using `account.metadata.snap.id` and calling snap keyring methods via `KeyringClient` (`@metamask/keyring-snap-client`) over `SnapController:handleRequest`, plus updates event filtering and cleanup/unsubscribe logic accordingly.
> 
> Updates initialization to delegate the new Snap/Permission controller actions/events, removes `snapProvider` from `initDataSources` options/tests, trims Snap exports from `index.ts`, and adds new dependencies/tsconfig references for permission/keyring snap support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0376a352eb68904a86a5e340dd9bbe4b7ab1a494. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/eth-json-rpc-middleware/CHANGELOG.md
+++ b/packages/eth-json-rpc-middleware/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add prototype pollution validation for `signTypedData` methods (V1, V3, V4) to block dangerous properties (`__proto__`, `constructor`, `prototype`, etc.) in message data. ([#7732](https://github.com/MetaMask/core/pull/7732))
+
 ### Changed
 
 - Bump `@metamask/eth-block-tracker` from `^15.0.0` to `^15.0.1` ([#7642](https://github.com/MetaMask/core/pull/7642))
--- a/packages/eth-json-rpc-middleware/src/utils/validation.ts
+++ b/packages/eth-json-rpc-middleware/src/utils/validation.ts
@@ -3,6 +3,7 @@ import type { Struct, StructError } from '@metamask/superstruct';
 import { validate } from '@metamask/superstruct';
 import type { Hex } from '@metamask/utils';
 
+import { parseTypedMessage } from './normalize';
 import type { WalletMiddlewareContext } from '../wallet';
 
 /**
@@ -95,4 +96,94 @@ function formatValidationError(error: StructError, message: string): string {
         `${failure.path.join(' > ')}${failure.path.length ? ' - ' : ''}${failure.message}`,
     )
     .join('\n')}`;
+}
+
+export const DANGEROUS_PROTOTYPE_PROPERTIES = [
+  '__proto__',
+  'constructor',
+  'prototype',
+  '__defineGetter__',
+  '__defineSetter__',
+  '__lookupGetter__',
+  '__lookupSetter__',
+] as const;
+
+/**
+ * Checks if a property name is dangerous for prototype pollution.
+ *
+ * @param key - The property name to check
+ * @returns True if the property name is dangerous
+ */
+function isDangerousProperty(key: string): boolean {
+  return (DANGEROUS_PROTOTYPE_PROPERTIES as readonly string[]).includes(key);
+}
+
+/**
+ * Recursively checks an object for dangerous prototype pollution properties.
+ *
+ * @param obj - The object to check
+ * @throws rpcErrors.invalidInput() if a dangerous property is found
+ */
+function checkObjectForPrototypePollution(obj: unknown): void {
+  if (obj === null || obj === undefined) {
+    return;
+  }
+
+  if (Array.isArray(obj)) {
+    for (const item of obj) {
+      checkObjectForPrototypePollution(item);
+    }
+    return;
+  }
+
+  if (typeof obj === 'object') {
+    for (const key of Object.getOwnPropertyNames(
+      obj as Record<string, unknown>,
+    )) {
+      if (isDangerousProperty(key)) {
+        throw rpcErrors.invalidInput();
+      }
+      checkObjectForPrototypePollution((obj as Record<string, unknown>)[key]);
+    }
+  }
+}
+
+/**
+ * Validates V1 typed data (array format) for prototype pollution attacks.
+ * V1 format: [{ type: 'string', name: 'fieldName', value: 'data' }, ...]
+ *
+ * @param data - The V1 typed data array to validate
+ * @throws rpcErrors.invalidInput() if prototype pollution is detected
+ */
+export function validateTypedDataV1ForPrototypePollution(
+  data: Record<string, unknown>[],
+): void {
+  if (!data || !Array.isArray(data)) {
+    return;
+  }
+
+  for (const item of data) {
+    if (item && typeof item === 'object') {
+      // Only check the 'value' field (the message data) for dangerous properties
+      if (item.value !== null && typeof item.value === 'object') {
+        checkObjectForPrototypePollution(item.value);
+      }
+    }
+  }
+}
+
+/**
+ * Validates V3/V4 typed data (EIP-712 format) for prototype pollution attacks.
+ * Only checks the message field for dangerous properties.
+ *
+ * @param data - The stringified typed data to validate
+ * @throws rpcErrors.invalidInput() if prototype pollution is detected
+ */
+export function validateTypedDataForPrototypePollution(data: string): void {
+  const { message } = parseTypedMessage(data);
+
+  // Check message recursively for dangerous properties
+  if (message !== undefined) {
+    checkObjectForPrototypePollution(message);
+  }
 }
--- a/packages/eth-json-rpc-middleware/src/wallet.test.ts
+++ b/packages/eth-json-rpc-middleware/src/wallet.test.ts
@@ -8,6 +8,7 @@ import type {
   TypedMessageV1Params,
 } from '.';
 import { createWalletMiddleware } from '.';
+import { DANGEROUS_PROTOTYPE_PROPERTIES } from './utils/validation';
 import { createHandleParams, createRequest } from '../test/util/helpers';
 
 const testAddresses = [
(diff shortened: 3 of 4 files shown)


packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -467,447 +369,186 @@ export class SnapDataSource extends AbstractDataSource<
...
     try {
-      const snaps = await this.#snapProvider.request<
-        Record<string, { version: string }>
-      >({
-        method: 'wallet_getSnaps',
-        params: {},
-      });
-
-      return snaps;
+      const allSnaps = this.#messenger.call('SnapController:getAll');
```
FrederikBolding: `SnapController:getRunnableSnaps` does the filtering for you

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -467,447 +369,186 @@ export class SnapDataSource extends AbstractDataSource<
...
+        // Must have endowment:keyring permission
+        if (!permissions[KEYRING_PERMISSION]) {
+          continue;
+        }
 
-    for (const scope of scopes) {
-      const [scopeNamespace, scopeReference] = (scope as string).split(':');
+        // Find chainIds caveat in any permission
+        let chainIds: ChainId[] | null = null;
+        for (const permissionConstraint of Object.values(permissions)) {
```
FrederikBolding: If we only care about keyrings here we don't need to iterate over all permissions. We can just inspect `permissions[KEYRING_PERMISSION]`

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -467,447 +369,186 @@ export class SnapDataSource extends AbstractDataSource<
...
+      if (!snapId) {
+        continue;
+      }
 
-      // Skip accounts that don't support any of the requested chains
-      if (accountSupportedChains.length === 0) {
+      // Skip accounts whose snap doesn't support any of the requested chains
+      const snapSupportsRequestedChains = request.chainIds.some(
+        (chainId) => this.state.chainToSnap[chainId] === snapId,
+      );
```
FrederikBolding, line 534: I don't fully understand how this is supposed to work. What if the `chainToSnap` mapping isn't already populated? Are the inputs BIP-44 account groups or individual accounts?

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done. Thanks

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done. THanks

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: `chainToSnap` is populated at the initialization of MetaMask, much before that we are fetching assets. So `What if the chainToSnap mapping isn't already populated?` is unlikely to say impossible.

`Are the inputs BIP-44 account groups or individual accounts?`
request.accounts is the selected account group

Prithpal-Sooriya (dismissed):
Approving on first pass. This data source is completely new and not in production (we have time to refine).

Lets have the snaps team to review before merging. Maybe a short loom/recording to walkthrough changes?

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -413,8 +316,8 @@ export class SnapDataSource extends AbstractDataSource<
         response.assetsBalance[accountId] = {};
```
FrederikBolding: If statement above here seems pointless?

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -413,8 +316,8 @@ export class SnapDataSource extends AbstractDataSource<
...
```
FrederikBolding: We can delay setting allocating until we know we need it, then we also don't need to have an extra branch later to clean it up.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -349,28 +253,29 @@ export class SnapDataSource extends AbstractDataSource<
...
+  /** Bound handler for snap keyring balance updates, stored for cleanup */
+  readonly #handleSnapBalancesUpdatedBound: (
+    payload: AccountBalancesUpdatedEventPayload,
+  ) => void;
 
   constructor(options: SnapDataSourceOptions) {
-    const configuredNetworks =
-      options.configuredNetworks ?? ALL_DEFAULT_NETWORKS;
-
     super(SNAP_DATA_SOURCE_NAME, {
```
FrederikBolding, line 270: Not necessarily Snaps-related, but I think you should loop in @mcmire to look at the DataSource architecture. We should ensure that this pattern is aligned with the approach we generally want for data services.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -467,447 +380,183 @@ export class SnapDataSource extends AbstractDataSource<
...
+    try {
+      const runnableSnaps = this.#getRunnableSnaps();
+      const chainToSnap: Record<ChainId, string> = {};
+      const supportedChains: ChainId[] = [];
 
-    // Extract namespace and reference from chainId
-    const [chainNamespace, chainReference] = chainId.split(':');
+      for (const snap of runnableSnaps) {
+        const permissions = this.#getSnapPermissions(snap.id);
+        if (!permissions) {
```
FrederikBolding: Nit, but we can combine this branch and the one below by using optional chaining

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -349,28 +253,29 @@ export class SnapDataSource extends AbstractDataSource<
...
 
     this.#registerActionHandlers();
     this.#subscribeToSnapKeyringEvents();
 
-    // Check availability for all snaps on initialization
-    this.#checkAllSnapsAvailability().catch(() => {
-      // Silently ignore availability check failures on init
-    });
+    // Discover keyring-capable snaps and populate activeChains dynamically
+    this.#discoverKeyringSnaps();
```
FrederikBolding, line 288: Since this is wrapped in try-catch internally we should consider the fallback behavior. In that case `chainToSnap` would not be populated correctly.

Additionally, potentially minor, but the PermissionController state can change during runtime, so more keyring Snaps could be added at runtime. We don't need to support this right away, but I wanna make sure that it at least is known.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -467,447 +380,183 @@ export class SnapDataSource extends AbstractDataSource<
...
       }
 
       const accountId = account.id;
       try {
-        // Step 1: Get the list of assets for this account
-        log(`${snapType} snap calling keyring_listAccountAssets`, {
-          snapId: config.snapId,
-          accountId,
-        });
+        const client = this.#getKeyringClient(snapId);
```
FrederikBolding, line 541: Potentially we should consider re-using the clients per Snap, so a call with many accounts don't need to instantiate a keyring client per account?

packages/assets-controller/src/data-sources/SnapDataSource.ts
mcmire: @FrederikBolding: Thanks for the mention, @Kriys94 messaged me privately about the DataSource code in AssetsController code. I'm looking through it now to understand how it all fits together. @Kriys94: I'll let you know when I have something to share.

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done. Thanks

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: I have a diagram about what we are trying to build https://consensyssoftware.atlassian.net/wiki/spaces/PRDC/pages/400284352527/Architecture+proposal+34+AssetsController+the+cross-network-asset-type+controller#High-Level-Architecture

Here we call "data source" a piece of logic that extends the AssetsController and implements a specific interface to work with external services (Backend  REST APIs, Snap, RPC). The code is related to the business logic, here assets balance, price and metadata.

For @mcmire and this ticket https://consensyssoftware.atlassian.net/browse/WPC-26, I think it refers more to a new service/client I just implemented in core-backend https://github.com/MetaMask/core/tree/main/packages/core-backend#http-api . The code in core-backend is agnostic form business logic.

Kriys94: In any case, feel free to comment on the overall code :) It's not released yet so making a braking change is easy :)

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done. Thanks

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done. I added the subscription to state changes in the PermissionController

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done 👍

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Done

packages/assets-controller/package.json
```diff
@@ -54,12 +54,17 @@
...
     "@metamask/keyring-controller": "^25.1.0",
     "@metamask/keyring-internal-api": "^9.0.0",
+    "@metamask/keyring-snap-client": "^8.0.0",
     "@metamask/messenger": "^0.3.0",
     "@metamask/network-controller": "^29.0.0",
     "@metamask/network-enablement-controller": "^4.1.0",
+    "@metamask/permission-controller": "^12.2.0",
     "@metamask/polling-controller": "^16.0.2",
+    "@metamask/snaps-controllers": "^17.2.0",
+    "@metamask/snaps-utils": "^11.7.0",
```
FrederikBolding, line 67: We just released a new major version of snaps-utils, please use that!

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -1081,16 +747,74 @@ export class SnapDataSource extends AbstractDataSource<
...
+   * Caches clients per snap ID to avoid re-instantiation across multiple calls.
+   *
+   * @param snapId - ID of the Snap to get the client for.
+   * @returns A `KeyringClient` for the Snap.
+   */
+  #getKeyringClient(snapId: string): KeyringClient {
+    const cachedClient = this.#keyringClientCache.get(snapId);
+    if (cachedClient) {
+      return cachedClient;
+    }
```
FrederikBolding, line 765: Nice!

PR #7777, opened 2026-01-29
feat(AssetController): add hiddenAssets state

Kriys94:
## Explanation

This PR adds support for hidden assets in the `AssetsController`, similar to how `allIgnoredAssets` works in `MultichainAssetsController`.

**Current State:**
Users currently have no way to hide assets they don't want to see or track in the new AssetsController architecture. The `MultichainAssetsController` already has this capability via `allIgnoredAssets`, but the new unified `AssetsController` was missing this feature.

**Solution:**
Added a `hiddenAssets` field to the `AssetsController` state that stores hidden assets per account (CAIP-19 asset IDs). This allows users to hide assets they don't want to track and/or see.

**Changes:**
1. **State**: Added `hiddenAssets: { [accountId: string]: string[] }` to `AssetsControllerState`
2. **Methods**: Added `hideAsset()`, `unhideAsset()`, `getHiddenAssets()`, and `isAssetHidden()` methods
3. **Actions**: Added corresponding action types (`AssetsControllerHideAssetAction`, `AssetsControllerUnhideAssetAction`, `AssetsControllerGetHiddenAssetsAction`) and registered handlers
4. **Custom Assets Integration**: When a custom asset is added via `addCustomAsset()`, it's automatically unhidden if it was previously hidden (matching the behavior of `MultichainAssetsController.addAssets()`)
5. **RPC Tracking**: Updated `RpcDataSource` and `BalanceFetcher` to filter out hidden assets from balance polling, so hidden assets won't be tracked or updated

**Type Updates:**
- Added `hiddenAssets` to `AssetsControllerState`
- Added `hiddenAssets` to `AssetsControllerStateInternal`
- Added `hiddenAssets` to `AssetsBalanceState` in evm-rpc-services types

## References

- Related to the Hide Token feature for the new unified AssetsController architecture
- Pattern follows `allIgnoredAssets` in `MultichainAssetsController`

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new persisted `assetPreferences` state slice and changes `getAssets` to filter hidden assets, which can affect UI output and downstream consumers relying on prior state shape/types.
> 
> **Overview**
> Adds a new persisted `assetPreferences` state (with `AssetPreferences`) to store per-asset UI flags like `hidden`, and introduces `hideAsset`/`unhideAsset` actions to manage this globally.
> 
> Updates `getAssets` to exclude hidden assets from returned results (while leaving balance/price tracking intact), and ensures `addCustomAsset` automatically unhides an asset if it was previously hidden.
> 
> Tightens `AssetsControllerState` typing from generic `Json`/`string[]` to semantic types (`AssetMetadata`, `AssetBalance`, `AssetPrice`, `Caip19AssetId[]`), and adjusts `RpcDataSource`/tests to match the narrowed `AssetsController:getState` expectations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a661f7e5438b0b54d7b092b8fa5d86fecd621f8c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.test.ts
+++ b/packages/assets-controller/src/AssetsController.test.ts
@@ -201,7 +201,7 @@ describe('AssetsController', () => {
         assetsBalance: {},
         assetsPrice: {},
         customAssets: {},
-        hiddenAssets: {},
+        assetPreferences: {},
       });
     });
   });
@@ -214,7 +214,7 @@ describe('AssetsController', () => {
           assetsBalance: {},
           assetsPrice: {},
           customAssets: {},
-          hiddenAssets: {},
+          assetPreferences: {},
         });
       });
     });
@@ -293,6 +293,7 @@ describe('AssetsController', () => {
 
       // Controller should still have default state (from super() call)
       expect(controller.state).toStrictEqual({
+        assetPreferences: {},
         assetsMetadata: {},
         assetsBalance: {},
         assetsPrice: {},
@@ -313,6 +314,7 @@ describe('AssetsController', () => {
       await withController(({ controller, messenger }) => {
         // Controller should have default state
         expect(controller.state).toStrictEqual({
+          assetPreferences: {},
           assetsMetadata: {},
           assetsBalance: {},
           assetsPrice: {},
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -19,7 +19,6 @@ import type {
   NetworkEnablementControllerEvents,
   NetworkEnablementControllerState,
 } from '@metamask/network-enablement-controller';
-import type { Json } from '@metamask/utils';
 import { parseCaipAssetType } from '@metamask/utils';
 import { Mutex } from 'async-mutex';
 import BigNumberJS from 'bignumber.js';
@@ -39,6 +38,7 @@ import { projectLogger, createModuleLogger } from './logger';
 import type { DetectionMiddlewareGetAssetsMiddlewareAction } from './middlewares/DetectionMiddleware';
 import type {
   AccountId,
+  AssetPreferences,
   ChainId,
   Caip19AssetId,
   AssetMetadata,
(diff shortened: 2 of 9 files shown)


packages/assets-controller/src/data-sources/evm-rpc-services/services/BalanceFetcher.ts
```diff
@@ -123,13 +123,21 @@ export class BalanceFetcher extends StaticIntervalPollingControllerOnly<BalanceP
...
     const chainIdDecimal = parseInt(chainId, 16);
     const caipChainPrefix = `eip155:${chainIdDecimal}/`;
 
     const tokenMap = new Map<string, TokenFetchInfo>();
 
     for (const assetId of Object.keys(accountBalances)) {
+      // Skip hidden assets
+      if (hiddenAssets.has(assetId)) {
+        continue;
+      }
```
Prithpal-Sooriya: Hmm does this mean we _don't_ fetch and update balances for hidden assets?
Might be problematic for the swaps flow (you could be trying to swap a hidden asset, where the balance is outdated?)

High level, my initial thoughts are that "Hidden Assets" are a UI lever/control, we should still be performing data fetching if there are places that do show this asset.

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -179,6 +179,7 @@ type AssetsControllerGetStateAction = {
     allIgnoredTokens: Record<string, Record<string, string[]>>;
     assetsMetadata: Record<Caip19AssetId, AssetMetadata>;
     assetsBalance: Record<string, Record<string, { amount: string }>>;
+    hiddenAssets: Record<string, string[]>;
```
Prithpal-Sooriya: nit (not a blocker) - I wonder if we add the JSDOC like we did for `packages/assets-controller/src/data-sources/evm-rpc-services/types/state.ts`, it helped readability on what this `Record<string, string[]>` is.

Prithpal-Sooriya: (optional, needs discussion), the Accounts team export a type alias, that could assist with readability
```ts
export type AccountId = string;
```

...Hesitant since it doesn't actually provide any additional type safety... maybe JSDOC is better?

Prithpal-Sooriya: Actually I actually much more prefer the approach you used below in packages/assets-controller/src/AssetsController.ts
```ts
type Foo = { [accountId: string]: string[] };
```

I think this is easier to read/understand then `Record<string, string[]>`

packages/assets-controller/src/AssetsController.ts
```diff
@@ -931,6 +987,85 @@ export class AssetsController extends BaseController<
...
+   * @param accountId - The account ID to check.
+   * @param assetId - The CAIP-19 asset ID to check.
+   * @returns True if the asset is hidden, false otherwise.
+   */
+  isAssetHidden(accountId: AccountId, assetId: Caip19AssetId): boolean {
+    const normalizedAssetId = normalizeAssetId(assetId);
+    return (
+      this.state.hiddenAssets[accountId]?.includes(normalizedAssetId) ?? false
+    );
+  }
```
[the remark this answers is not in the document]
Prithpal-Sooriya: not a huge deal, we can add some tests to cover this.

packages/assets-controller/src/AssetsController.test.ts
Prithpal-Sooriya, line 1: Can we add some tests on the new exposed methods from this controller?

packages/assets-controller/src/data-sources/RpcDataSource.test.ts
```diff
@@ -187,10 +187,6 @@ async function withController<ReturnValue>(
 
   // Mock AssetsController:getState
   messenger.registerActionHandler('AssetsController:getState', () => ({
-    allTokens: {},
-    allDetectedTokens: {},
-    allIgnoredTokens: {},
-    assetsMetadata: {},
```
Kriys94, line 193: This is just cleaning because we only need balance and metadata

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -170,14 +170,10 @@ type TokenListControllerGetStateAction = {
...
 
-// AssetsController:getState action (for user tokens state and assets balance)
+// AssetsController:getState action (for assets balance and metadata)
 type AssetsControllerGetStateAction = {
   type: 'AssetsController:getState';
   handler: () => {
-    allTokens: Record<string, Record<string, { address: string }[]>>;
-    allDetectedTokens: Record<string, Record<string, { address: string }[]>>;
-    allIgnoredTokens: Record<string, Record<string, string[]>>;
-    assetsMetadata: Record<Caip19AssetId, AssetMetadata>;
```
Kriys94, line 180: This is just cleaning because we only need balance and metadata

packages/assets-controller/src/AssetsController.ts
```diff
@@ -90,19 +92,22 @@ export type AssetsControllerState = {
   assetsPrice: { [assetId: string]: Json };
   /** Custom assets added by users per account (CAIP-19 asset IDs) */
   customAssets: { [accountId: string]: string[] };
+  /** UI preferences per asset (e.g. hidden) */
+  assetPreferences: { [assetId: string]: Json };
```
Prithpal-Sooriya: nit/Q is this type meant to be wide (untyped Json type?). Could it be narrowed to `AssetPreferences`?


```suggestion
  assetPreferences: { [assetId: string]: AssetPreferences };
```

Prithpal-Sooriya (approved):

packages/assets-controller/src/AssetsController.ts
Kriys94: Much better. Thanks

PR #7783, opened 2026-01-30
feat: Withdrawal to any token (Predict)

dan437:
## Explanation
Adds support for "post-quote" withdrawal flows, enabling users to withdraw funds to any token/chain (e.g., Predict withdrawals).

Changes:

- Add isPostQuote flag to transaction data for withdrawal-first flows
- Add setIsPostQuote controller action to enable post-quote mode
- Support EXACT_INPUT trade type in Relay strategy for withdrawals (user specifies source amount, receives quote for destination)
- Build post-quote requests in buildQuoteRequests when isPostQuote is true
- Handle same-token-same-chain withdrawals without requiring a bridge
- Add source amount tracking for post-quote flows
- Put the deposit into a batch with the original transaction

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches quote construction and Relay batch submission logic (including transaction typing and gas handling), which can impact correctness of on-chain execution for withdrawals. Changes are feature-gated by the new `isPostQuote` flag but broaden behavior in core transaction-pay flows.
> 
> **Overview**
> Adds *post-quote* (withdrawal) support across Transaction Pay by introducing `isPostQuote` into `TransactionData`/`QuoteRequest` and persisting it onto `TransactionMeta.metamaskPay` (also extending `MetamaskPayMetadata` in `transaction-controller`).
> 
> Replaces `setIsMaxAmount` with a callback-based `setTransactionConfig`, and ensures toggling `isPostQuote` re-triggers source-amount and quote refresh.
> 
> Updates Relay quoting/submission to handle post-quote semantics: keep `targetAmountMinimum=0` requests when `isPostQuote`, use `EXACT_INPUT`, skip transaction preprocessing, avoid Hyperliquid-deposit normalization in post-quote mode, and submit as a batch that prepends the original transaction ahead of relay steps with corrected type/gasLimit handling.
> 
> Adds post-quote-specific source amount derivation (including new `source*` fields and same-token-same-chain skip logic via `isSameToken`), includes original transaction gas as `fees.transactionGas` in totals, and fixes a nested-transfer detection edge case (`findIndex` returning `-1`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 685847e231fc88e642a1dc7e6df355023eb14d14. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -59,8 +59,13 @@ export async function getRelayQuotes(
 
   try {
     const normalizedRequests = requests
-      // Ignore gas fee token requests
-      .filter((singleRequest) => singleRequest.targetAmountMinimum !== '0')
+      // Ignore gas fee token requests (which have both target=0 and source=0)
+      // but keep post-quote requests (which have target=0 but source>0)
+      .filter(
```
dan437, line 64: The new logic keeps requests where either target > 0 (normal deposit flows) or source > 0 (withdrawal/post-quote flows). Only requests with both at zero (gas fee token placeholders) get filtered out.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -32,6 +32,112 @@ const FALLBACK_HASH = '0x0' as Hex;
...
+};
+
+/**
+ * Extract and validate relay params from a quote.
+ *
+ * @param quote - Relay quote.
+ * @param messenger - Controller messenger.
+ * @returns Submit context with normalized params and metadata.
+ */
+export function getSubmitContext(
```
dan437: I've abstracted shared code between `submitPostQuoteTransactions` and `submitTransactions` functions.

packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `isPostQuote` field to `TransactionData` type for withdrawal flows ([#7773](https://github.com/MetaMask/core/pull/7773))
+- Add `setIsPostQuote` action type ([#7773](https://github.com/MetaMask/core/pull/7773))
```
matthewwalsh0: We can nest changes here, so could we have a high level summary like `Support quote execution after transaction in Relay strategy` with details in nested lines?

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -80,6 +80,12 @@ export class TransactionPayController extends BaseController<
     });
   }
 
+  setIsPostQuote(transactionId: string, isPostQuote: boolean): void {
```
matthewwalsh0: Rather than adding more individual setters and actions, could we define a new `TransactionConfig` type and a single `setTransactionConfig` that uses a callback to allow custom changes to the config without defining it each time?

packages/transaction-pay-controller/src/utils/token.ts
```diff
@@ -141,14 +144,14 @@ export function getTokenInfo(
...
 
-  if (!token && !isNative) {
-    return undefined;
-  }
-
   if (token && !isNative) {
     return { decimals: Number(token.decimals), symbol: token.symbol };
   }
 
+  if (!token && !isNative) {
```
matthewwalsh0, line 165: Can we remove the changes to this file as they seem identical?

packages/transaction-pay-controller/src/utils/source-amounts.ts
```diff
@@ -58,6 +67,25 @@ export function updateSourceAmounts(
...
+ */
+function calculatePostQuoteSourceAmounts(
+  tokens: TransactionPayRequiredToken[],
+): TransactionPaySourceAmount[] {
+  return tokens
+    .filter((token) => !token.skipIfBalance)
+    .map((token) => ({
+      sourceAmountHuman: token.amountHuman,
+      sourceAmountRaw: token.amountRaw,
+      targetTokenAddress: token.address,
```
matthewwalsh0: This should be the payment token?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +287,83 @@ function buildQuoteRequests({
...
+  from: Hex;
+  isMaxAmount: boolean;
+  destinationToken: TransactionPaymentToken;
+  sourceAmounts: TransactionPaySourceAmount[] | undefined;
+  tokens: TransactionPayRequiredToken[];
+  transactionId: string;
+}): QuoteRequest[] {
+  // For withdrawals, the required token (e.g., Polygon USDC.e) becomes the SOURCE
+  // and the destinationToken (paymentToken) becomes the TARGET/destination
+  const sourceToken = tokens.find((token) => !token.skipIfBalance);
```
matthewwalsh0: Since we've already done this to generate the source amounts, should we instead just look for a source amount with `targetTokenAddress` matching the payment token / destinationToken?

packages/transaction-pay-controller/src/utils/source-amounts.ts
```diff
@@ -58,6 +67,25 @@ export function updateSourceAmounts(
...
+ *
+ * @param tokens - Required tokens from the transaction.
+ * @returns Array of source amounts.
+ */
+function calculatePostQuoteSourceAmounts(
+  tokens: TransactionPayRequiredToken[],
+): TransactionPaySourceAmount[] {
+  return tokens
+    .filter((token) => !token.skipIfBalance)
+    .map((token) => ({
```
matthewwalsh0, line 109: Can we do the same is same token and zero skips here, as we do below?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +287,83 @@ function buildQuoteRequests({
...
+      sourceToken.address.toLowerCase(),
+  );
+
+  // Check if source and target are the same token on the same chain
+  const isSameToken =
+    sourceToken.address.toLowerCase() ===
+      destinationToken.address.toLowerCase() &&
+    sourceToken.chainId === destinationToken.chainId;
+
+  if (isSameToken) {
```
matthewwalsh0: As mentioned in other comment, this can go in `source-amounts.ts`.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +287,83 @@ function buildQuoteRequests({
...
+    // For same-token-same-chain, no quote is needed - the withdrawal goes directly to user
+    // Return empty to indicate no bridging required
+    log('Same token same chain - no bridge needed', { transactionId });
+    return [];
+  }
+
+  const request: QuoteRequest = {
+    from,
+    isMaxAmount,
```
matthewwalsh0, line 336: If we want max support, we can also do that in `source-amounts.ts` and set it to the required token balance.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +287,83 @@ function buildQuoteRequests({
...
+    sourceBalanceRaw: sourceToken.balanceRaw,
+    sourceTokenAmount: sourceAmount?.sourceAmountRaw ?? sourceToken.amountRaw,
```
matthewwalsh0: Should always use the `sourceAmount` here and return early above if there isn't one.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +287,83 @@ function buildQuoteRequests({
...
+    // For post-quote withdrawals, use EXACT_INPUT - user specifies how much to withdraw,
+    // and we show them how much they'll receive after fees
+    targetAmountMinimum: '0',
+    targetChainId: destinationToken.chainId,
+    targetTokenAddress: destinationToken.address,
+  };
+
+  log('Post-quote request built', { transactionId, request });
+
+  return [request];
```
matthewwalsh0, line 353: Maybe worth a comment to clarify we don't have multiple token support yet as that would require multiple quotes for each required token.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -59,8 +59,13 @@ export async function getRelayQuotes(
...
+        (singleRequest) =>
```
matthewwalsh0, line 65: For simplicity, could we remove this logic and provide `isPostQuote` in the pay strategy requests for clarity?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -111,15 +116,20 @@ async function getSingleQuote(
   );
 
   try {
+    // For post-quote (withdrawal) flows, use EXACT_INPUT - user specifies how much
```
matthewwalsh0: I'm concerned we keep mentioning withdrawal everywhere. It's definitely a good example of a post-quote flow, but conceptually we could be moving funds after a transaction for any reason, to a different recipient for example.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -111,15 +116,20 @@ async function getSingleQuote(
...
+    // to withdraw, and we show them how much they'll receive after fees.
+    // For regular flows with a target amount, use EXPECTED_OUTPUT.
+    const useExactInput = isMaxAmount === true || targetAmountMinimum === '0';
```
matthewwalsh0: As mentioned above, could just check `isPostQuote` directly.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -140,7 +150,7 @@ async function getSingleQuote(
 
     log('Fetched relay quote', quote);
 
-    return normalizeQuote(quote, request, fullRequest);
+    return await normalizeQuote(quote, request, fullRequest);
```
matthewwalsh0, line 158: Is this functionally any different if we're returning from an async function?

packages/transaction-pay-controller/src/types.ts
```diff
@@ -142,7 +147,20 @@ export type TransactionData = {
   /** Whether the user has selected the maximum amount. */
   isMaxAmount?: boolean;
 
-  /** Source token selected for the transaction. */
+  /**
+   * Whether this is a post-quote transaction (e.g., withdrawal flow).
+   * When true, the paymentToken represents the destination token,
+   * and the quote source is derived from the transaction's output token.
+   * Used for Predict/Perps withdrawals where funds flow:
```
matthewwalsh0: Ideally, we avoid coupling to specific usages / transaction types and keep the descriptions abstract?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
matthewwalsh0: Why are these two functions exported if they're only used internally?

We shouldn't export just for unit tests, but test them via their callers.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -73,6 +179,8 @@ async function executeSingleQuote(
 ): Promise<{ transactionHash?: Hex }> {
   log('Executing single quote', quote);
 
+  const isPostQuote = transaction.metamaskPay?.isPostQuote;
```
matthewwalsh0: Better to determine this via the strategy request directly, rather than the transaction where the properties are just used for transaction details.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -84,7 +192,14 @@ async function executeSingleQuote(
...
   );
 
-  await submitTransactions(quote, transaction.id, messenger);
+  // For post-quote (withdrawal) flows, create an atomic batch with:
+  // 1. Original transaction (e.g., Safe withdrawal)
+  // 2. Relay deposit transaction
+  if (isPostQuote) {
+    await submitPostQuoteTransactions(quote, transaction, messenger);
+  } else {
+    await submitTransactions(quote, transaction.id, messenger);
```
matthewwalsh0: Rather than branching here, would it not be much simpler to keep the original flow and code, but just add a tiny branch in `submitTransactions` to include the original transaction data as the first entry in `normalizedParams` if `isPostQuote` is `true`?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -104,6 +219,159 @@ async function executeSingleQuote(
...
+      maxPriorityFeePerGas?: Hex;
+      to: Hex;
+      value?: Hex;
+    };
+    type?: TransactionType;
+  }[] = [];
+
+  // Add original transaction (e.g., Safe execTransaction call)
+  // Note: We use txParams (the outer call) not nestedTransactions (internal calls)
+  // because nestedTransactions are executed BY the Safe, not by the user's EOA
```
matthewwalsh0: We only use `nestedTransactions` for batch transactions to user EOAs, meaning it should be empty anyway?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0, line 158: Doesn't have to be this PR, but since we're adding a transaction to the Relay deposit transaction, we'll need to update the source gas fees in the normalized quote to include the gas fee of the original transaction.

Technically, we may also need to move the fee to the target property, and set source to zero.

Depends if we consider the fees relative to the transaction or the quote 😅

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -111,15 +116,20 @@ async function getSingleQuote(
...
 
   try {
+    // For post-quote (withdrawal) flows, use EXACT_INPUT - user specifies how much
+    // to withdraw, and we show them how much they'll receive after fees.
+    // For regular flows with a target amount, use EXPECTED_OUTPUT.
+    const useExactInput = isMaxAmount === true || targetAmountMinimum === '0';
+
     const body: RelayQuoteRequest = {
-      amount: isMaxAmount ? sourceTokenAmount : targetAmountMinimum,
+      amount: useExactInput ? sourceTokenAmount : targetAmountMinimum,
```
matthewwalsh0, line 126: I think we also need to update `processTransactions` to skip changing the quote at all, since we never want to also execute the same transaction data on the target chain of the quote.

I assume it is skipping now since it's just a token transfer for the Predict withdraw.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
dan437: The await is needed because this return is inside a try/catch. Without it, rejections from normalizeQuote would bypass the catch block since the promise would be returned unwrapped. With await, errors are properly caught and logged before re-throwing.

dan437: Thanks for pointing this out. Will create a ticket for it.

packages/transaction-pay-controller/src/types.ts
```diff
@@ -85,10 +83,26 @@ export type TransactionPayControllerUpdatePaymentTokenAction = {
...
+   * and the quote source is derived from the transaction's output token.
+   */
+  isPostQuote?: boolean;
+};
+
+/** Callback to update transaction config. */
+export type TransactionConfigCallback = (config: TransactionConfig) => void;
+
+/** Action to update transaction configuration using a callback. */
+export type TransactionPayControllerSetTransactionConfigAction = {
```
matthewwalsh0: Minor, could we keep all the actions defined together?

packages/transaction-controller/src/types.ts
matthewwalsh0, line 1: Did we accidentally remove the changelog updates for both controllers?

packages/transaction-pay-controller/src/utils/source-amounts.ts
```diff
@@ -95,14 +153,9 @@ function calculateSourceAmount(
...
   const strategy = getStrategyType(transactionId, messenger);
-
-  const isSameTokenSelected =
-    token.address.toLowerCase() === paymentToken.address.toLowerCase() &&
-    token.chainId === paymentToken.chainId;
-
   const isAlwaysRequired = isQuoteAlwaysRequired(token, strategy);
 
-  if (isSameTokenSelected && !isAlwaysRequired) {
+  if (isSameToken(token, paymentToken) && !isAlwaysRequired) {
```
matthewwalsh0, line 158: 👍

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -263,6 +286,71 @@ function buildQuoteRequests({
...
+  // Find the source amount where targetTokenAddress matches the destination token
+  const sourceAmount = sourceAmounts?.find(
+    (amount) =>
+      amount.targetTokenAddress.toLowerCase() ===
+      destinationToken.address.toLowerCase(),
+  );
+
+  // Same-token-same-chain cases are already filtered in source-amounts.ts
+  if (
+    !sourceAmount?.sourceTokenAddress ||
```
matthewwalsh0: Minor, these are all required fields, so should we just check if `sourceAmount` is defined?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -111,19 +116,28 @@ async function getSingleQuote(
   );
 
   try {
+    // For post-quote or max amount flows, use EXACT_INPUT - user specifies how much to send,
+    // and we show them how much they'll receive after fees.
+    // For regular flows with a target amount, use EXPECTED_OUTPUT.
+    const useExactInput = isMaxAmount === true || request.isPostQuote === true;
```
matthewwalsh0: Minor, since these are both booleans, can we remove the `=== true`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -111,19 +116,28 @@ async function getSingleQuote(
...
       slippageTolerance,
-      tradeType: isMaxAmount ? 'EXACT_INPUT' : 'EXPECTED_OUTPUT',
+      tradeType: useExactInput ? 'EXACT_INPUT' : 'EXPECTED_OUTPUT',
       user: from,
     };
 
-    await processTransactions(transaction, request, body, messenger);
+    // Skip transaction processing for post-quote flows - the original transaction
+    // will be included in the batch separately, not as part of the quote
+    if (!request.isPostQuote) {
```
matthewwalsh0, line 139: Minor, since this is a local function, we could also early return inside it as we have a few branches like that already.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -304,6 +438,20 @@ async function submitTransactions(
 
   end();
 
+  // For post-quote flows, mark original transaction as handled by nested batch
+  if (isPostQuote) {
```
matthewwalsh0: Is this not already the default path for all transactions in the Relay strategy?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
matthewwalsh0: We've still got a considerable refactor in this file, my point was also to greatly simplify the diff.

Starting from the original code, could we change nothing except line 193 in `submitTransactions` where we would just add our additional params / transaction from the target transaction, then everything else is unchanged?

It would automatically go into the batch flow since we'd always have multiple params?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/types.ts
dan437: I've removed them temporarily, because I kept getting conflicts there. It seems the PR is getting ready to be merged, so I will add them back.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
dan437: An early return here would require duplicating or extracting the fetch + normalize logic that follows, since both paths need it. Keeping the guard keeps the diff smaller.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -32,6 +32,39 @@ const FALLBACK_HASH = '0x0' as Hex;
...
+/**
+ * Determine the transaction type for a given index in the batch.
+ *
+ * @param isPostQuote - Whether this is a post-quote flow.
+ * @param index - Index of the transaction in the batch.
+ * @param originalType - Type of the original transaction (used for post-quote index 0).
+ * @param relayParamCount - Number of relay-only params (excludes prepended original tx).
+ * @returns The transaction type.
+ */
+function getTransactionType(
```
matthewwalsh0, line 44: Minor, exported functions at the top, internal below?

dan437: Will include it in the next PR!

packages/transaction-pay-controller/src/types.ts
```diff
@@ -311,6 +350,14 @@ export type TransactionPayFees = {
...
   /** Network fee for transactions on the target network. */
   targetNetwork: FiatValue;
+
+  /**
+   * Gas cost of the original user transaction on its native network.
+   * Always calculated from the transaction's own gas params via
+   * `calculateTransactionGasCost`, independent of any bridge quote fees.
+   * Only populated on totals, not on individual quotes.
+   */
+  transactionGas?: FiatValue;
```
matthewwalsh0, line 360: Afraid I totally missed this, was this added between reviews?

We shouldn't need to expose the transaction specific gas externally, as it becomes part of the Relay quote deposit, so it should be encapsulated by the `sourceNetwork` fee.

We can remove in the upcoming gas fee PR.

PR #7800, opened 2026-02-02
fix: persist new `accessToken` in SeedlessOnboarding Vault after tokens refresh

lwin-kyaw:
## Explanation

This PR address 
- fix new `accessToken` isn't being persisted to vault after the tokens refresh (`refreshAuthToken()`).
- add a new public method getAccessToken() with automatic token refresh capability to prevent clients from using expired access tokens when retrieving them from state.
#### Issue:
Previously, clients could directly access state.accessToken from the controller state, which could potentially return an expired token. This could lead to failed API calls and poor user experience when the token had expired but wasn't refreshed.

#### Solution:
- Added a new public method `getAccessToken()` that automatically refreshes expired tokens: Uses the #executeWithTokenRefresh wrapper to check token expiration and refresh if needed before returning
- Persisted new `accessToken` into Encrypted vault after token refresh.

#### Persist `newAccessToken` in the encrypted vault

The persist flow can be different based on the wallet (vault) status when `refreshAuthTokens()` is called.

1. When wallet is unlocked (vault `encryptionKey` is available in the state).
When wallet is unlocked, vault encryption key is temporarily cached in the controller state and the values inside the encrypted vault are also flooded into the controller state, for performance and to avoid the password dependency
We can use these cache values and create a new encrypted vault (with updated accessToken).

Screenshot 2026-02-03 at 1 47 36 AM

2. When wallet is locked (Password must be provided)

While the wallet is locked, `refreshAuthTokens()` can be called under one condition, i.e, check for password sync (`getAuthPubKey` RPC call) when user tries to unlock the wallet. 
In this case, we can't immediately update the encrypted vault as the wallet is locked. So, we will temporarily store the new `accessToken` in the controller state, then persist the stored value later in the vault unlock (**only** when user provides correct password) and update the vault.

Screenshot 2026-02-03 at 12 22 38 PM

Based on the password sync status, the persist flow can be a little bit different.

2.1. Password is in sync

When user's current device password is in sync, we will update the vault when user unlocks via `submitPassword`. 
Vault is decrypted using the submitted password, we will compare the access token value from decrypted vault (`accessToken1`) and recently set new access token (`accessToken2`). If the accessToken is different and detected as new, we will update the encrypted vault value in the state (using password). 
Please note that this requires an additional encryption step to update the latestToken in encrypted vault.

Screenshot 2026-02-03 at 12 29 21 PM

2.2. Password is not in sync (`OutdatedPassword`).

When user's current device password is out of sync and user submits the latest `globalPassword`, we do the `pwEncKey` recovery first so that we can unlock the current vault. Similar to the flow above (2.1), we decrypt the current vault and compare the two tokens (`accessToken1` and `accessToken2`). If the accessToken is different and detected as new, we will update the controller state and encrypted vault.
Unlike the flow (2.1.), this doesn't require additional encryption, as we can include the latest accessToken when we sync the latest global enc keys into the controller.

Screenshot 2026-02-03 at 3 53 31 PM

## References

Fixes https://github.com/MetaMask/core/issues/7805, https://github.com/MetaMask/metamask-extension/issues/39566

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches vault encryption/update paths and token refresh behavior, so bugs could lead to incorrect vault contents or token state during unlock/refresh flows, though changes are localized and well-covered by tests.
> 
> **Overview**
> Fixes token persistence by updating the seedless onboarding vault when `refreshAuthTokens()` obtains a new `accessToken`, using `encryptWithKey` when the wallet is already unlocked and reconciling state vs vault tokens during `submitPassword`/`submitGlobalPassword` via a new `compareAndGetLatestToken` helper.
> 
> Adds a new public `getAccessToken()` action/messenger handler that returns the current access token and triggers the existing refresh flow when tokens are expired, and introduces `assertIsValidPassword` plus test refactors/new cases (including a shared `createMockJWTToken`) to cover these behaviors. **Breaking:** the injected vault `encryptor` must now implement `encryptWithKey`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 09a0ddf6c2aa0c6c377c47d7dce34cf0be40e1b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -745,7 +866,50 @@ export class SeedlessOnboardingController<
...
+          accessTokenAfterUnlock,
+        );
+      }
+
+      // update the state and vault with the latest access token if it's different from the current access token in the state.
+      if (latestAccessToken && latestAccessToken !== accessTokenAfterUnlock) {
+        // update the access token in the state with the latest access token if it's different from the decrypted access token after unlocking
+        this.update((state) => {
+          state.accessToken = latestAccessToken;
+        });
```
matthiasgeihs: Instead of doing all of this, would it be possible to simply overwrite the access token?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -856,25 +1020,51 @@ export class SeedlessOnboardingController<
...
+        const latestAccessToken = compareAndGetLatestToken(
+          accessTokenBeforeUnlock,
+          accessTokenFromDecryptedVault,
+        );
+        // update the access token in the state with the latest access token if it's different from the decrypted access token after unlocking
+        // later when we call `syncLatestGlobalPassword`, the encrypted vault will be updated with the latest access token.
+        this.update((state) => {
+          state.accessToken = latestAccessToken;
+        });
+      }
```
matthiasgeihs: Similar comment as above: instead of doing this, could you simply have one access token that is overwritten when you refresh it before unlock?

If this is not possible, it would be nice if you could move the checking code into a dedicated function to reduce code duplication.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -856,25 +1020,51 @@ export class SeedlessOnboardingController<
...
+
+      this.#resetPasswordOutdatedCache();
```
matthiasgeihs: What's that doing?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1970,6 +2163,23 @@ export class SeedlessOnboardingController<
...
   }
 
+  /**
+   * Get the access token from the state.
+   *
+   * If the tokens are expired, the method will refresh them and return the new access token.
+   *
+   * @returns The access token.
+   */
+  async getAccessToken(): Promise<string | undefined> {
```
matthiasgeihs, line 2246: When is this public function used by the MM client?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: You mean overwrite with the new accessToken from state, without validating it?
Yes, we can though. It just a sanity check.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: For the Auth-service `/marketing_opt_in` API calls (when we create/import wallet or in the settings). We provide the `accessToken` as an Authorization header.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: This must been accidentally included when I tried to combine `submitGlobalPassword` and `syncLatestPassword`. (But I didn't, for the atomic reasons.)

I will remove it back.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
matthiasgeihs: OK, up to you. But in any case would be good to have this logic in a dedicated function to improve readability of the code and reduce duplication.

Also, I was a bit confused that `#updateVaultAfterAuthTokenRefresh` doesn't actually update the vault when the controller is locked. You could rename this to `#updateStateAfterAuthTokenRefresh`.

lwin-kyaw: > But in any case would be good to have this logic in a dedicated function to improve readability of the code and reduce duplication.

Yes, I will definitely refactor it.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Removed it here, https://github.com/MetaMask/core/pull/7800/commits/77d255e6a336320d7050cfda453315e63c574938

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Thanks for the suggestion. I've refactored it here, https://github.com/MetaMask/core/pull/7800/commits/77d255e6a336320d7050cfda453315e63c574938

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Refactored here, https://github.com/MetaMask/core/pull/7800/commits/77d255e6a336320d7050cfda453315e63c574938

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -64,6 +69,117 @@ import {
...
+    typeof controllerName,
+    SeedlessOnboardingControllerState
+  >;
+export type SeedlessOnboardingControllerEvents =
+  SeedlessOnboardingControllerStateChangeEvent;
+
+type AllowedEvents = never;
+
+// Messenger
+export type SeedlessOnboardingControllerMessenger = Messenger<
```
matthiasgeihs, line 111: Why did you move the types `SeedlessOnboardingControllerMessenger` and `SeedlessOnboardingControllerOptions`?

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2094,6 +2293,84 @@ export class SeedlessOnboardingController<
     }
   }
 
+  async #updateVaultAfterAuthTokenRefresh(accessToken: string): Promise<void> {
```
matthiasgeihs: You should merge this function with `#updateVault`.
Having the updateVault logic duplicated across two functions is a recipe for disaster. Imagine updating the one without updating the other.

lwin-kyaw: Hi matthias, thanks for the review.

I have updated the `#updateVault` in this commit, https://github.com/MetaMask/core/pull/7800/commits/39708a097ddf16105ed7f04985674954bcd2a6c5

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Hi, I move these types into controller file for the new messenger action, [SeedlessOnboardingControllerGetAccessTokenAction](https://github.com/MetaMask/core/blob/39708a097ddf16105ed7f04985674954bcd2a6c5/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts#L86).

If I added the messenger action in the `/types.ts`, it would create a circular import as I need to reference the controller in the `/types.ts`.
```ts
export type SeedlessOnboardingControllerGetAccessTokenAction = {
  type: `${typeof controllerName}:getAccessToken`;
  handler: SeedlessOnboardingController< // <=== Coz of this reference, I can't declare this action is `/types.ts`.
    encryptionUtils.EncryptionKey,
    encryptionUtils.KeyDerivationOptions
  >['getAccessToken'];
};
```

lwin-kyaw:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1603,40 +1798,97 @@ export class SeedlessOnboardingController<
...
-      this.#cachedDecryptedVaultData = vaultData;
+      const { vaultEncryptionKey, vaultEncryptionSalt, vault } = this.state;
 
-      const serializedVaultData = serializeVaultData(vaultData);
+      const updatedState: Partial<SeedlessOnboardingControllerState> = {
+        vault,
+        vaultEncryptionKey,
+        vaultEncryptionSalt,
+        encryptedSeedlessEncryptionKey:
+          this.state.encryptedSeedlessEncryptionKey,
```
tuna1207, line 1815: is there a reason we are not getting this `encryptedSeedlessEncryptionKey` from the state destructuring on line 1808
https://github.com/MetaMask/core/pull/7800/changes#diff-7ec8aa1b0ce65286f76f0cc6467231ec9f79c7cefd409bb13bb0d20bd258b52aR1808

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1603,40 +1798,97 @@ export class SeedlessOnboardingController<
...
+        updatedState.vaultEncryptionKey = exportedKeyString;
+        updatedState.vaultEncryptionSalt = JSON.parse(updatedEncVault).salt;
+
+        // encrypt the seedless encryption key with the password encryption key from TOPRF network
+        updatedState.encryptedSeedlessEncryptionKey =
+          this.#encryptSeedlessEncryptionKey(exportedKeyString, pwEncKey);
+      } else if (vaultEncryptionKey && vaultEncryptionSalt) {
+        const encryptionKey =
+          await this.#vaultEncryptor.importKey(vaultEncryptionKey);
+        const updatedEncVault = await this.#vaultEncryptor.encryptWithKey(
```
tuna1207, line 1843: since this is unserialized vault encrypt result and would need to be serialized later unlike `updatedEncVault` on line 1827
https://github.com/MetaMask/core/pull/7800/changes#diff-7ec8aa1b0ce65286f76f0cc6467231ec9f79c7cefd409bb13bb0d20bd258b52aR1827
i think we should name this `unserializedUpdatedEncVault` to be more clear

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: ah, no reason. I just wanna see the whole method without scrolling. hehe 😅

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: This is the encrypted result resulted from encrypting the Serialized Vault Data.

matthiasgeihs (approved):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1603,40 +1798,97 @@ export class SeedlessOnboardingController<
...
-      // Note that vault encryption using the password is a very costly operation as it involves deriving the encryption key
-      // from the password using an intentionally slow key derivation function.
-      // We should make sure that we only call it very intentionally.
-      const { vault, exportedKeyString } =
-        await this.#vaultEncryptor.encryptWithDetail(
-          password,
+      // if the password is provided (not undefined), encrypt the vault with the password
+      // We gonna prioritize the password encryption here, in case of the operation is `Change Password`.
+      // We don't wanna re-use the old encryption key from the state.
+      if (password !== undefined) {
```
chaitanyapotti, line 1821: best practice is to use typeof password !== "undefined" when checking for existence of a variable especially globals.
This is fine for this use case.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2094,6 +2362,29 @@ export class SeedlessOnboardingController<
...
+   */
+  #checkTokensExpired(): boolean {
+    // proactively check for expired tokens and refresh them if needed
+    const isNodeAuthTokenExpired = this.checkNodeAuthTokenExpired();
+    const isMetadataAccessTokenExpired = this.checkMetadataAccessTokenExpired();
+    // access token is only accessible when the vault is unlocked
+    // so skip the check if the vault is locked
+    let isAccessTokenExpired = false;
+    if (this.#isUnlocked) {
+      isAccessTokenExpired = this.checkAccessTokenExpired();
```
chaitanyapotti, line 2378: Ideally, we must check expiry with server as part of oauth2 spec 
since, client time can be incorrect or session might have been invalidated by server.
this particular implementation is non-standard and can be revisited at a later date

PR #7804, opened 2026-02-02
feat: add rwa data to tokens in tokens controller

oscarwroche:
## Explanation

  - Propagate cached `rwaData` onto tokens when metadata is updated.
  - Thread optional `rwaData` through `addToken` so suggested assets can carry RWA data.
  - Set token `rwaData` from the supplied argument instead of token metadata (which doesn't contain `rwaData`)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches token state write paths (`addToken`/`watchAsset`) and cache-to-state propagation, which can affect persisted token metadata and UI display if `rwaData` is set/cleared incorrectly, but scope is limited and covered by new tests.
> 
> **Overview**
> Adds optional `rwaData` plumbing to `TokensController` token additions so `watchAsset`/`addToken` can persist RWA metadata, including allowing `rwaData: undefined` to *clear* previously-stored RWA data on re-add.
> 
> Also updates the TokenListController cache sync logic to key lookups by lowercased token address and to propagate cached `rwaData` onto existing tokens, with tests covering overwrite/clear behavior and cache-driven updates; changelog is updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cea397b40f58275454358274c105cfb341828024. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/eip-5792-middleware/CHANGELOG.md
+++ b/packages/eip-5792-middleware/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/transaction-controller` from `^62.7.0` to `^62.14.0` ([#7596](https://github.com/MetaMask/core/pull/7596), [#7602](https://github.com/MetaMask/core/pull/7602), [#7604](https://github.com/MetaMask/core/pull/7604), [#7642](https://github.com/MetaMask/core/pull/7642), [#7737](https://github.com/MetaMask/core/pull/7737), [#7760](https://github.com/MetaMask/core/pull/7760), [#7775](https://github.com/MetaMask/core/pull/7775), [#7802](https://github.com/MetaMask/core/pull/7802), [#7832](https://github.com/MetaMask/core/pull/7832))
 - Bump `@metamask/utils` from `^11.8.1` to `^11.9.0` ([#7511](https://github.com/MetaMask/core/pull/7511))
+- Make `wallet_sendCalls` more reliable ([#7816](https://github.com/MetaMask/core/pull/7816))
 
 ## [2.1.0]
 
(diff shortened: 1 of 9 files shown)


packages/assets-controllers/src/TokensController.ts
```diff
@@ -288,6 +289,9 @@ export class TokensController extends BaseController<
                 if (cachedToken && cachedToken.name && !token.name) {
                   token.name = cachedToken.name; // Update the token name
                 }
+                if (cachedToken?.rwaData) {
+                  token.rwaData = cachedToken.rwaData; // Update the token RWA data
+                }
```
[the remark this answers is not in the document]
oscarwroche: This is on purpose

packages/assets-controllers/jest.config.js
```diff
@@ -32,4 +32,7 @@ module.exports = merge(baseConfig, {
 
   // We rely on `window` to make requests
   testEnvironment: '<rootDir>/jest.environment.js',
+
+  // Watchman isn't available in some environments (e.g. sandboxed CI/containers)
```
mcmire: Nit: Should this be in its own PR? It doesn't seem to be related to the changes here and might get buried. It would also be good to understand what you were doing when you encountered the problem that this fixes.

oscarwroche: It's just codex that added this, good catch, will get rid of it :)

packages/assets-controllers/src/TokensController.ts
Akaryatrh: For some reason, `address` from `token` is not lowercased (while those from `chainData` are), hence `cachedToken` returns `undefined`.
```suggestion
                const cachedToken = chainData[token.address.toLowerCase()];
```

This is maybe specific to RWA tokens. In the meantime we can apply this change but i guess there's an API where address is not forced to lowercase format.

Prithpal-Sooriya (approved):

PR #7905, opened 2026-02-11
chore: migrate Jest from v29 to v30

cryptodev-2s:
## Explanation

- Update jest to ^30.4.2, @types/jest to ^30.0.0, jest-silent-reporter to ^0.6.0, ts-jest to ^29.4.11 across all packages
- Update jest-environment-jsdom to ^30.4.1
- Remove prettier-2 alias and prettierPath config (Jest 30 supports Prettier 3 natively)
- Fix toThrow() assertions for Jest 30's stricter error comparison in eth-json-rpc-provider, eth-json-rpc-middleware, and permission-controller

### Why `PollingBlockTracker.checkForLatestBlock` now dedupes

`checkForLatestBlock` should return the same ongoing promise when called concurrently, but the old implementation did not. It was a plain `async` method, so each call started its own `eth_blockNumber` request and returned a different promise. The unit test claimed "should return the same promise" but used `toStrictEqual`, which compares promises structurally. Two promises have no enumerable own properties, so `toStrictEqual` treats any two as equal on both Jest 29 and Jest 30. The assertion always passed, so the missing dedupe went unnoticed.

The fix caches the promise and switches the assertion to `toBe` (reference identity), the only matcher that actually verifies two calls return the same promise. Net effect for consumers: concurrent `checkForLatestBlock` calls now share one request instead of each issuing their own. Recorded as a fix in the changelog.

<details>
<summary>Runnable repro — save as <code>repro.mjs</code> in the repo root, then <code>node repro.mjs</code></summary>

```js
#!/usr/bin/env node
/**
 * Repro: `PollingBlockTracker.checkForLatestBlock` was supposed to dedupe
 * concurrent calls (return the SAME in-flight promise), but the old
 * implementation did not -- it fired a separate request per call. The unit
 * test claimed "should return the same promise" yet passed anyway, because it
 * used `toStrictEqual`, which compares promises structurally and cannot tell
 * two distinct promises apart. This is true on BOTH Jest 29 and Jest 30, so the
 * bug was invisible regardless of Jest version. Only `toBe` (reference
 * identity) catches it, which is why the fix pairs the dedupe with `toBe`.
 *
 * Run from the repo root: node <path>/block-tracker-dedup-repro.mjs
 */
import { createRequire } from 'node:module';

// Resolve the repo's own `expect` (Jest 30) so we can show its real behavior.
const require = createRequire(import.meta.url);
let expect;
try {
  ({ expect } = require('expect'));
} catch {
  expect = undefined; // script still works without it
}

// A fake RPC that returns an incrementing block and counts how many times it is
// actually hit. A duplicate request shows up as an extra count.
let requestCount = 0;
async function fetchLatestBlock() {
  requestCount += 1;
  await new Promise((resolve) => setTimeout(resolve, 5));
  return `0x${requestCount}`;
}

// OLD: plain async, no dedupe. Every call starts its own request.
class OldTracker {
  async checkForLatestBlock() {
    return fetchLatestBlock();
  }
}

// NEW: cache the in-flight promise so concurrent callers share one request.
class NewTracker {
  #pending;
  checkForLatestBlock() {
    if (!this.#pending) {
      this.#pending = fetchLatestBlock().finally(() => {
        this.#pending = undefined;
      });
    }
    return this.#pending;
  }
}

function matcher(name, a, b) {
  if (!expect) return '(expect unavailable)';
  try {
    if (name === 'toBe') expect(a).toBe(b);
    else expect(a).toStrictEqual(b);
    return 'PASS';
  } catch {
    return 'FAIL';
  }
}

async function run(label, tracker) {
  requestCount = 0;
  const p1 = tracker.checkForLatestBlock();
  const p2 = tracker.checkForLatestBlock(); // concurrent, before p1 settles
  await Promise.all([p1, p2]);

  console.log(`\n== ${label} ==`);
  console.log(`  RPC requests made          : ${requestCount}  ${requestCount === 1 ? '(deduped)' : '(DUPLICATE requests -> the bug)'}`);
  console.log(`  Object.is(p1, p2)          : ${Object.is(p1, p2)}`);
  console.log(`  expect().toStrictEqual()   : ${matcher('toStrictEqual', p1, p2)}   <- what the test used`);
  console.log(`  expect().toBe()            : ${matcher('toBe', p1, p2)}   <- identity check`);
  return { requestCount, sameRef: Object.is(p1, p2), toStrict: matcher('toStrictEqual', p1, p2) };
}

console.log('Two distinct promises have no enumerable own properties:');
console.log('  Object.keys(Promise.resolve(1)) =', JSON.stringify(Object.keys(Promise.resolve(1))));
console.log('so deep/structural equality cannot tell two promises apart.');

const oldResult = await run('OLD checkForLatestBlock (no dedupe)', new OldTracker());
const newResult = await run('NEW checkForLatestBlock (dedupe)', new NewTracker());

console.log('\n----------------------------------------------------------------');
console.log('The test asserts the two promises are "the same":');
console.log('  expect(p1).toStrictEqual(p2)   <- the assertion the test actually used');
console.log('');
console.log(`  OLD code: makes ${oldResult.requestCount} requests, returns 2 DIFFERENT promises,`);
console.log(`            yet toStrictEqual = ${oldResult.toStrict}. The bug is invisible.`);
console.log(`  NEW code: makes ${newResult.requestCount} request, returns the SAME promise.`);
console.log('');
console.log('Key point: toStrictEqual is STRUCTURAL. Two promises have no enumerable own');
console.log('properties, so it treats any two as equal -- on Jest 29 AND Jest 30. The Jest');
console.log('version does not matter here. Only expect().toBe() (reference identity) can tell');
console.log('them apart, which is why the fix pairs the dedupe with a toBe assertion.');

// Machine-checkable: the old bug is real (dupe requests, distinct promises) yet
// toStrictEqual still passes it.
const bugHiddenByToStrictEqual =
  oldResult.requestCount > 1 && !oldResult.sameRef && oldResult.toStrict === 'PASS';
console.log(`\nBug present AND hidden by toStrictEqual: ${bugHiddenByToStrictEqual}`);
process.exit(bugHiddenByToStrictEqual ? 1 : 0);
```

Output:

```
Two distinct promises have no enumerable own properties:
  Object.keys(Promise.resolve(1)) = []
so deep/structural equality cannot tell two promises apart.

== OLD checkForLatestBlock (no dedupe) ==
  RPC requests made          : 2  (DUPLICATE requests -> the bug)
  Object.is(p1, p2)          : false
  expect().toStrictEqual()   : PASS   <- what the test used
  expect().toBe()            : FAIL   <- identity check

== NEW checkForLatestBlock (dedupe) ==
  RPC requests made          : 1  (deduped)
  Object.is(p1, p2)          : true
  expect().toStrictEqual()   : PASS   <- what the test used
  expect().toBe()            : PASS   <- identity check

----------------------------------------------------------------
The test asserts the two promises are "the same":
  expect(p1).toStrictEqual(p2)   <- the assertion the test actually used

  OLD code: makes 2 requests, returns 2 DIFFERENT promises,
            yet toStrictEqual = PASS. The bug is invisible.
  NEW code: makes 1 request, returns the SAME promise.

Key point: toStrictEqual is STRUCTURAL. Two promises have no enumerable own
properties, so it treats any two as equal -- on Jest 29 AND Jest 30. The Jest
version does not matter here. Only expect().toBe() (reference identity) can tell
them apart, which is why the fix pairs the dedupe with a toBe assertion.

Bug present AND hidden by toStrictEqual: true
```

</details>

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Wide test-infra upgrade plus small production changes (block tracker dedupe, backup/sync promise returns) that affect RPC churn and sync call semantics; coverage threshold tweaks could mask regressions if misunderstood as product changes.
> 
> **Overview**
> Upgrades the monorepo to **Jest 30** (`jest` ^30.4.2, `@types/jest` ^30, `jest-environment-jsdom` ^30.4.1, `ts-jest` ^29.4.11, `jest-silent-reporter` ^0.6.0) across root and packages, drops the **prettier-2** alias and **`prettierPath`** (Jest 30 uses Prettier 3), and adds a Lavamoat allowlist entry for Jest’s **`unrs-resolver`**.
> 
> Test and tooling fallout: **`toThrow`** expectations now use message strings where Jest 30 no longer matches thrown error objects deeply; promise-deduping tests switch from **`toStrictEqual`** to **`toBe`** so same-reference promises are actually asserted. Many **`coverageThreshold`** values are retuned to match post-upgrade coverage.
> 
> **Runtime fixes** bundled with the migration: **`PollingBlockTracker.checkForLatestBlock`** caches an in-flight promise so concurrent callers share one `eth_blockNumber` request (changelog in `eth-block-tracker`). **`BackupAndSyncService`** `performFullSync` / `performFullSyncAtLeastOnce` return explicit **`Promise.resolve`** when disabled instead of `async`/`undefined`. **`recordCallsToSetTimeout`** no longer calls **`jest.runOnlyPendingTimers`** around timeout helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d264438178924c42e67a6418ee3327f155052ea5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/account-tree-controller/jest.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 100,
+      branches: 92,
```
cryptodev-2s: Coverage drop explanation here https://github.com/jestjs/jest/issues/15760#issuecomment-3645209409

packages/account-tree-controller/src/backup-and-sync/service/index.ts
```diff
@@ -215,9 +215,9 @@ export class BackupAndSyncService {
    *
    * @returns A promise that resolves when the sync is complete.
    */
-  async performFullSync(): Promise<void> {
+  performFullSync(): Promise<void> {
```
cryptodev-2s, line 218: Deliberately not `async`. They return the cached ongoing promise by reference (`#ongoingFullSyncPromise` / `#firstOngoingFullSyncPromise`), which is the dedupe. An `async` method wraps its return in a fresh promise every call, so the identity (and the dedupe) is lost. The test asserts this with `toBe`. The old `toStrictEqual` compared structurally and can pass for two different promises, so it never actually verified the dedupe. That is also why the disabled branch returns `Promise.resolve(undefined)` instead of a bare `undefined`.

packages/eth-block-tracker/tests/recordCallsToSetTimeout.ts
```diff
@@ -52,10 +52,6 @@ class SetTimeoutRecorder {
    * called.
    */
   async next(): Promise<void> {
-    // Resolve pending Promises first. Pending Promises always get resolved before a `setTimeout`
-    // callback in practice, so this better reflects a real scenario.
-    jest.runOnlyPendingTimers();
-
     await new Promise<void>((resolve) => {
```
cryptodev-2s, line 55: Removed the `jest.runOnlyPendingTimers()` calls here (and in `nextMatchingDuration`). This suite uses its own `setTimeout` mock and never calls `jest.useFakeTimers()`, so these only ever hit jest fake clock that is not installed. Jest 29 ignored that silently, jest 30 throws. They were no-ops, so removing them is behavior neutral.

packages/eth-json-rpc-provider/src/internal-provider.test.ts
```diff
@@ -202,11 +202,7 @@ describe.each([
       };
 
       await expect(async () => provider.request(request)).rejects.toThrow(
-        providerErrors.custom({
-          code: 1001,
-          message: 'Test error',
-          data: { cause: 'Test cause' },
-        }),
+        'Test error',
```
cryptodev-2s, line 205: Changed these `toThrow(...)` args from error objects to the message string. Jest 30 makes `toThrow(errorObject)` also compare the error `cause`, not just the message. The thrown error is wrapped by the JSON-RPC engine (gains `code: -32603` and a nested cause), so it no longer deep matches a freshly built `providerErrors.custom(...)` / `rpcErrors.internal(...)`. Matching on the message is stable and is what the test intends.

packages/eth-json-rpc-middleware/src/fetch.test.ts
```diff
@@ -120,14 +119,7 @@ describe('createFetchMiddleware', () => {
...
         ),
-      ).rejects.toThrow(
-        rpcErrors.internal({
-          data: {
-            code: -1000,
-            message: 'oops',
-          },
-        }),
-      );
+      ).rejects.toThrow('Internal JSON-RPC error.');
```
cryptodev-2s, line 122: Same jest 30 change as in `internal-provider.test.ts`: `toThrow(errorObject)` now compares the error `cause`, not just the message. These errors get wrapped by the middleware, so they no longer deep match a freshly built `rpcErrors.internal({...})`. Matching on the message string is stable and is what the test intends, so the `rpcErrors` import is dropped too.

cryptodev-2s:
@SocketSecurity ignore npm/data-urls@5.0.0
@SocketSecurity ignore npm/rrweb-cssom@0.8.0
@SocketSecurity ignore npm/@jest/snapshot-utils@30.4.1
@SocketSecurity ignore npm/@jest/diff-sequences@30.4.0
@SocketSecurity ignore npm/glob@10.5.0
@SocketSecurity ignore npm/jsdom@26.1.0
@SocketSecurity ignore npm/ts-jest@29.4.11

All transitive/dev deps from the Jest 30 + jsdom upgrade. The "obfuscated code" flags (data-urls, rrweb-cssom) are minified published bundles, the "AI signal" flags (jsdom, glob, ts-jest, @jest/snapshot-utils) are heuristics on widely used packages, and @jest/diff-sequences "publisher changed" is the expected Jest 30 republish. No runtime exposure in published packages.

packages/account-tree-controller/jest.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 100,
+      branches: 92.3,
```
Mrtenz, line 20: Why did this go down so much? 🤔 Can we get it back to 100%?

Mrtenz: I see now this happens in many packages. Have you looked into the cause?

packages/assets-controller/jest.config.js
```diff
@@ -18,10 +18,10 @@ module.exports = merge(baseConfig, {
   // TODO: Increase thresholds as more tests are added
```
Mrtenz: I guess we can remove this comment?

packages/eth-block-tracker/src/PollingBlockTracker.ts
```diff
@@ -316,9 +318,15 @@ export class PollingBlockTracker<
    * @deprecated Use {@link getLatestBlock} instead.
    * @returns A promise that resolves to the latest block number.
    */
-  async checkForLatestBlock(): Promise<string> {
-    await this.#updateLatestBlock();
-    return await this.getLatestBlock();
+  checkForLatestBlock(): Promise<string> {
```
Mrtenz, line 321: Why is this needed? 🤔

Mrtenz (commented):
Thought this would be more work. Looks good, just have some minor questions.

packages/account-tree-controller/jest.config.js
cryptodev-2s: I left a comment here but it doesn't show up anymore https://github.com/MetaMask/core/pull/7905#discussion_r2804522937

packages/eth-block-tracker/src/PollingBlockTracker.ts
cryptodev-2s: I left an explanation in the description but I believe that was not smart

Mrtenz: Ah, sorry, I missed that.

Mrtenz (approved):

packages/assets-controller/jest.config.js
cryptodev-2s: Yes done here https://github.com/MetaMask/core/pull/7905/changes/87c60f8412e79c49b321e7679d7668f8960804bf

PR #8032, opened 2026-02-24
fix: fix WebSocket balance conversion and add unit tests

salimtb:
## Explanation

`BackendWebsocketDataSource` was pushing WebSocket balance notifications into `assetsBalance` as **raw smallest-units** (e.g. `"10000000000000000000"` for 10 ETH, `"1000000"` for 1 USDC). The RPC and Accounts API data sources already store **human-readable** amounts (e.g. `"10"`, `"1"`). That mismatch meant that after a swap or other WebSocket-driven update, the UI could show raw numbers for some assets and correct amounts for others, and any logic that assumes a single format could break.

Balance updates in `BackendWebsocketDataSource` are now converted from raw to human-readable before being written to `assetsBalance`. We parse the raw value (hex or decimal string), divide by `10^decimals` using the asset’s `decimals` (defaulting to 18 when missing), and store the resulting string—matching the approach used in `RpcDataSource`. `assetsBalance` is now consistently human-readable across fetch and WebSocket updates.

The conversion lives in `#processBalanceUpdates` and reuses the same BigNumber-based divide-by-10^decimals pattern used elsewhere in the pipeline so that all data sources write the same shape into state.

**If your primary goal was to update one package but you found you had to update another one along the way, why did you do so?**

N/A — only `@metamask/assets-controller` was changed.

**If you had to upgrade a dependency, why did you do so?**

N/A — no dependency upgrades.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches balance normalization logic used to update user-visible asset balances; mistakes could display incorrect amounts, but the change is small and covered by new unit tests.
> 
> **Overview**
> WebSocket balance notifications processed by `BackendWebsocketDataSource` are now converted from raw smallest-units (hex or decimal) into human-readable strings by dividing by `10^decimals` before writing into `assetsBalance`.
> 
> Adds guardrails and coverage: updates lacking `asset.decimals` are skipped, and unit tests were updated/added to assert correct conversion for native and ERC-20 assets (including hex inputs) plus the skip behavior. Changelog updated to note the fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac51deb6b1d145e685fe59df483be3a0488e8cc9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
+++ b/packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
@@ -13,6 +13,7 @@ import {
   KnownCaipNamespace,
   toCaipChainId,
 } from '@metamask/utils';
+import BigNumberJS from 'bignumber.js';
 
 import { AbstractDataSource } from './AbstractDataSource';
 import type {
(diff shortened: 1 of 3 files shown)


packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
```diff
@@ -607,13 +608,19 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
...
 
-      // Parse balance amount (already in hex format like "0xc350")
-      const balanceAmount = postBalance.amount.startsWith('0x')
+      // Parse raw balance (hex like "0x26f0e5" or decimal string)
+      const rawBalanceStr = postBalance.amount.startsWith('0x')
         ? BigInt(postBalance.amount).toString()
         : postBalance.amount;
 
+      // Convert to human-readable using asset decimals (match RpcDataSource / pipeline format)
+      const decimals = asset.decimals ?? 18;
```
bergarces: Can decimals be undefined? I don't think we should be defaulting this value.

salimtb: in theory we should always have decimals , I hesitated to do this but I think I'll remove the fallback now

bergarces (approved):

PR #8171, opened 2026-03-10
feat: add address poisoning detection to PhishingController

AugmentedMode:
## Explanation

This PR adds address poisoning detection to `@metamask/phishing-controller` by:

- Introducing a `findSimilarAddresses` utility that compares a candidate address against known recipients using prefix/suffix matching heuristics
- Exposing a new `PhishingController:checkAddressPoisoning` messenger action that clients can call to check if a recipient address looks like a poisoning attempt
- Hydrating and maintaining a set of known recipient addresses from:
  - Confirmed transactions via `TransactionController` state
  - Address book entries via `AddressBookController` state
- Subscribing to state changes on both controllers to keep the known recipients set up to date
- Returning match metadata (prefix/suffix match lengths, poisoning score, diff indices) so consumers can make informed decisions about how to warn users

## References

- Related to address poisoning attack vector ([learn more](https://support.metamask.io/managing-my-tokens/moving-your-tokens/address-poisoning-scams/))

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds new recipient-hydration logic driven by `TransactionController`/`AddressBookController` state changes plus a new messenger-exposed method, so bugs could impact performance or client warning behavior though no funds-moving logic is changed.
> 
> **Overview**
> Adds address-poisoning detection to `@metamask/phishing-controller` by introducing `findSimilarAddresses` and a new messenger action/method `PhishingController:checkAddressPoisoning` that returns similarity metadata (prefix/suffix match lengths, score, diff indices).
> 
> `PhishingController` now depends on `@metamask/address-book-controller`, hydrates a set of *known recipient* addresses from confirmed transactions and address-book entries, and keeps it updated by subscribing to both controllers’ `stateChange` events (with patch-aware transaction updates and rebuild fallbacks).
> 
> Includes new unit tests for the address similarity utility and extensive controller tests covering hydration, incremental updates, and failure-path logging; also tightens API error-type checks in a few existing scan methods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 306647c96c808668b85f1445466451787da0afeb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/phishing-controller/src/PhishingController.ts
```diff
@@ -721,13 +785,133 @@ export class PhishingController extends BaseController<
...
+    state: TransactionControllerState,
+  ): void {
+    this.#transactionRecipients.clear();
+    for (const address of this.#getConfirmedTransactionRecipients(
+      state.transactions,
+    )) {
+      this.#transactionRecipients.add(address);
+    }
+    this.#rebuildKnownRecipients();
+  }
```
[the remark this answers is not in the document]
AugmentedMode: FP

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -670,6 +730,10 @@ export class PhishingController extends BaseController<
     }
   }
 
+  #onAddressBookControllerStateChange(state: AddressBookControllerState): void {
+    this.#setKnownRecipientsFromAddressBookState(state);
+  }
```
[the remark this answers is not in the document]
AugmentedMode: FP

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -694,13 +748,133 @@ export class PhishingController extends BaseController<
...
+
+    const transactionRecipient = this.#normalizeAddress(
+      transaction.txParams.to,
+    );
+    const transferRecipient = this.#normalizeAddress(
+      (
+        transaction.transferInformation as
+          | { recipientAddress?: string }
+          | undefined
+      )?.recipientAddress,
```
imblue-dabadee: `TransactionMeta.transferInformation` doesn't have `recipientAddress` as a value so this would always evaluate to `undefined`?

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -694,13 +748,133 @@ export class PhishingController extends BaseController<
...
+    this.#hydrateKnownRecipientsFromTransactionState();
+    this.#hydrateKnownRecipientsFromAddressBookState();
+  }
+
+  #hydrateKnownRecipientsFromTransactionState(): void {
+    try {
+      const state = this.messenger.call('TransactionController:getState');
+      this.#setKnownRecipientsFromTransactionState(state);
+    } catch {
+      // Some environments may not provide transaction state hydration.
```
imblue-dabadee: What environments? And seems valuable to console.error this so that callers know that `checkAddressPoisoning` wouldn't really work if there are environments.

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -694,13 +748,133 @@ export class PhishingController extends BaseController<
...
+    }
+  }
+
+  #hydrateKnownRecipientsFromAddressBookState(): void {
+    try {
+      const state = this.messenger.call('AddressBookController:getState');
+      this.#setKnownRecipientsFromAddressBookState(state);
+    } catch {
+      // Some environments may not provide address book state hydration.
```
imblue-dabadee: What environments? And seems valuable to console.error this so that callers know that `checkAddressPoisoning` wouldn't really work if there are environments.

packages/phishing-controller/src/PhishingController.test.ts
```diff
@@ -91,8 +93,14 @@ function setupMessenger(): {
   });
 
   rootMessenger.delegate({
-    actions: [],
-    events: ['TransactionController:stateChange'],
+    actions: [
```
imblue-dabadee, line 131: Setting up some no-op mocks for these seem fair so that tests aren't operating with an incomplete setup.

packages/phishing-controller/src/address-poisoning.ts
```diff
@@ -0,0 +1,104 @@
+import type { SimilarAddressMatch, SimilarityOptions } from './types';
+
+const DEFAULT_PREFIX_LEN = 4;
+const DEFAULT_SUFFIX_LEN = 4;
+
+function isHexAddress(address: string): address is `0x${string}` {
```
imblue-dabadee: Can use `isValidHexAddress` in `controller-utils` instead.

packages/phishing-controller/src/address-poisoning.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        knownAddress: '0x1234ccccdddd5678',
+        prefixMatchLength: 4,
+        suffixMatchLength: 4,
+        poisoningScore: 8,
+        diffIndices: [6, 7, 8, 9, 10, 11, 12, 13],
+      },
+    ]);
+  });
+
+  it('ignores non-hex and differently-sized addresses', () => {
```
imblue-dabadee: Test case says "differently-sized addresses" but doesn't test this (if you are trying to, splitting the test up makes sense to keep them focused).

packages/phishing-controller/src/address-poisoning.test.ts
```diff
@@ -0,0 +1,93 @@
+import { findSimilarAddresses } from './address-poisoning';
+
+describe('findSimilarAddresses', () => {
```
imblue-dabadee, line 7: Testing empty `knownAddresses` (a wallet could have no known addresses in it) seems fair

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -615,10 +654,21 @@ export class PhishingController extends BaseController<
...
   #onTransactionControllerStateChange(
-    _state: { transactions: TransactionMeta[] },
+    _state: TransactionControllerState,
     patches: Patch[],
-  ) {
+  ): void {
     try {
+      if (patches.some((patch) => patch.path[0] === 'transactions')) {
+        try {
+          this.#setKnownRecipientsFromTransactionState(_state);
```
imblue-dabadee: Is this rebuilding the known recipients by iterating through every confirmed transaction every time there's a patch? If yes, surely there's a more reasonable alternative -- I can see this being expensive for power users.

packages/phishing-controller/src/PhishingController.ts
AugmentedMode: Good catch, you’re right. transferInformation only has token transfer metadata, so this would always be undefined. I removed that read and now use the actual transaction recipient fields instead.

packages/phishing-controller/src/PhishingController.ts
AugmentedMode: Agreed, the comment was too vague. This was defensive against consumers/tests that delegate the action but don’t register a TransactionController:getState handler. I added explicit console.error logging so it’s visible when existing confirmed transactions can’t be hydrated for address poisoning checks.

packages/phishing-controller/src/PhishingController.test.ts
AugmentedMode: Agreed. I updated the shared test messenger setup to register no-op/default TransactionController:getState and AddressBookController:getState handler

packages/phishing-controller/src/address-poisoning.ts
AugmentedMode: Updated!

packages/phishing-controller/src/address-poisoning.test.ts
AugmentedMode: Fixed. I split this into focused tests for non-hex candidates, non-hex known addresses, and differently-sized known addresses

packages/phishing-controller/src/address-poisoning.test.ts
AugmentedMode: Added a test for empty knownAddresses

packages/phishing-controller/src/PhishingController.ts
AugmentedMode: Good callout. I changed this to update transaction recipients incrementally from the relevant transaction patch instead of rebuilding from every confirmed transaction on each update. It also tracks recipients per transaction id with counts, so removals and duplicate recipients are handled correctly.

imblue-dabadee (dismissed):

mindofmar (approved):

PR #8338, opened 2026-03-30
feat(messenger): add `delegateAll` with exhaustive type checking

cryptodev-2s:
## Explanation

`Messenger.delegate()` accepts a partial list of actions/events without compile-time enforcement that all required items are present. This means missing a delegation (e.g., a newly added `AllowedAction`) silently compiles and only fails at runtime.

`delegateAll()` wraps `delegate()` with a `RequireExhaustive` branded intersection type that produces a clear TypeScript error showing exactly which actions or events are missing from the list. Own-namespace actions of the delegatee are automatically excluded from the check.

Example error when an action is missing:
```
Property '__missing' is missing in type '[...]'
  but required in type '{ __missing: "KeyringController:signTypedMessage" }'.
```

### `@metamask/messenger`

- **Added**: `delegateAll()` method — strict alternative to `delegate()` that requires all external actions and events to be listed
- **Added**: `MessengerNamespace<M>` utility type — extracts the namespace string literal from a Messenger type
- **Added**: `RequireExhaustive<Required, Provided>` utility type — validates tuple exhaustiveness at compile time

## Checklist

---

> [!NOTE]
> **Low Risk**
> API addition only (`delegateAll` delegates to existing `delegate`); main risk is consumers adopting stricter typing, not runtime behavior changes.
> 
> **Overview**
> Adds **`Messenger.delegateAll`**, a typed wrapper around **`delegate`** that requires every **external** action and event on the delegatee messenger to be listed (own-namespace items are excluded). Missing entries fail at compile time via a **`RequireExhaustive`** branded type that surfaces **`__MISSING_DELEGATIONS__`**, and the source messenger must also type-check as able to provide those items.
> 
> Also exports **`MessengerNamespace`** for extracting a messenger’s namespace literal from its type.
> 
> **Testing/tooling:** Jest unit tests for runtime behavior, new **TSTyche** (`*.tst.ts`) compile-time tests, **`test`** split into **`test:unit`** + **`test:types`**, coverage/build/lint configs updated to exclude type-test files, and **yarn.config** allows the messenger package’s custom test script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22ed38bbbe3670bf3a6a923640e1d1432453c88a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/messenger/src/Messenger.test.ts
```diff
@@ -1931,6 +1931,133 @@ describe('Messenger', () => {
...
+      sourceMessenger.delegateAll({
+        messenger: childMessenger,
+        actions: ['Source:getValue'],
+        events: [],
+      });
+
+      expect(childMessenger.call('Source:getValue')).toBe(42);
+    });
+
+    it('produces a type error when an action is missing', () => {
```
mcmire: I have found that Jest doesn't always check type errors, so I am not sure this will work as intended.

What are your thoughts about adding `tstyche` to this package and writing compile-time tests? We could include this test and also the one before it (since it also seems like that is meant to test the type and not the behavior).

packages/messenger/src/Messenger.ts
```diff
@@ -1129,6 +1169,65 @@ export class Messenger<
...
+   * @template DelegatedActions - A const tuple of delegated action type
+   *   strings.
+   * @template DelegatedEvents - A const tuple of delegated event type strings.
+   */
+  delegateAll<
+    Delegatee extends Messenger<string, ActionConstraint, EventConstraint>,
+    const DelegatedActions extends readonly (MessengerActions<Delegatee> &
+      Action)['type'][],
+    const DelegatedEvents extends readonly (MessengerEvents<Delegatee> &
+      Event)['type'][],
```
mcmire: Do we need `const ... readonly` here? If we take these away we can remove the type assertions below.

``` suggestion
    DelegatedActions extends (MessengerActions<Delegatee> & Action)['type'][],
    DelegatedEvents extends (MessengerEvents<Delegatee> & Event)['type'][],
```

packages/messenger/src/Messenger.ts
```diff
@@ -1129,6 +1169,65 @@ export class Messenger<
...
+        (MessengerEvents<Delegatee> & Event)['type']
+      >,
+      DelegatedEvents
+    >;
+  }): void {
+    this.delegate({
+      actions: actions as (MessengerActions<Delegatee> & Action)['type'][],
+      events: events as (MessengerEvents<Delegatee> & Event)['type'][],
+      messenger,
+    });
```
mcmire:
```suggestion
    this.delegate({ actions, events, messenger });
```

packages/messenger/src/Messenger.ts
mcmire: At the very least, we should extract the `type` first, then perform an intersection of them. We found in the past that performing an intersection first led to a "union type too complex to represent" error: https://github.com/MetaMask/core/pull/8748

``` suggestion
    const DelegatedActions extends
      readonly (MessengerActions<Delegatee>['type'] & Action['type'])[],
    const DelegatedEvents extends readonly (MessengerEvents<Delegatee>['type'] &
      Event['type'])[],
```

Or if we go with the above suggestion:

```suggestion
    DelegatedActions extends (MessengerActions<Delegatee>['type'] &
      Action['type'])[],
    DelegatedEvents extends (MessengerEvents<Delegatee>['type'] &
      Event['type'])[],
```

packages/messenger/src/Messenger.ts
```diff
@@ -1129,6 +1169,65 @@ export class Messenger<
...
+  >({
+    actions,
+    events,
+    messenger,
+  }: {
+    messenger: Delegatee;
+    actions: RequireExhaustive<
+      NotNamespacedBy<
+        MessengerNamespace<Delegatee>,
+        (MessengerActions<Delegatee> & Action)['type']
```
mcmire: Similar optimization as above:

```suggestion
        MessengerActions<Delegatee>['type'] & Action['type']

File: packages/messenger/src/Messenger.ts [hunk 5]
```diff
@@ -1129,6 +1169,65 @@ export class Messenger<
...
+      >,
+      DelegatedActions
+    >;
+    events: RequireExhaustive<
+      NotNamespacedBy<
+        MessengerNamespace<Delegatee>,
+        (MessengerEvents<Delegatee> & Event)['type']
```
mcmire: Similar optimization as above:

```suggestion
        MessengerEvents<Delegatee>['type'] & Event['type']
```

packages/messenger/src/Messenger.ts
```diff
@@ -94,6 +94,46 @@ export type MessengerEvents<
...
+ * type T2 = RequireExhaustive<'A' | 'B' | 'C', readonly ['A', 'B']>;
+ * // => readonly ['A', 'B'] & { missingDelegations: 'C' }
+ * ```
+ */
+type RequireExhaustive<
+  Required extends string,
+  Provided extends readonly string[],
+> = [Exclude<Required, Provided[number]>] extends [never]
+  ? Provided
+  : Provided & { missingDelegations: Exclude<Required, Provided[number]> };
```
mcmire: Nit: Do we want to name this property something non-standard on purpose so engineers don't mistake it for a property they ought to set?

``` suggestion
  // eslint-disable-next-line @typescript-eslint/naming-convention
  : Provided & { __MISSING_DELEGATIONS__: Exclude<Required, Provided[number]> };
```

mcmire (commented):
Overall strategy seems sound, but I had some suggestions.

packages/messenger/src/Messenger.ts
cryptodev-2s: Done in 08fc1b35cb. Dropped `const`/`readonly` and switched to `MessengerActions<Delegatee>['type'] & Action['type']` (same for events), matching `delegate`/`revoke`. Casts are gone as a result.

packages/messenger/src/Messenger.ts
cryptodev-2s: Done. Now `this.delegate({ actions, events, messenger })`.

packages/messenger/src/Messenger.ts
cryptodev-2s: Done. Updated to `MessengerActions<Delegatee>['type'] & Action['type']`.

packages/messenger/src/Messenger.ts
cryptodev-2s: Done. Updated to `MessengerEvents<Delegatee>['type'] & Event['type']`.

packages/messenger/src/Messenger.ts
cryptodev-2s: Good idea. Renamed to `__MISSING_DELEGATIONS__` so it does not look like a real option.

packages/messenger/src/Messenger.test.ts
cryptodev-2s: Good call. Added `tstyche` in e2fca54f6c and moved the exhaustiveness cases (missing action + own namespace exclusion) into `Messenger.delegateAll.tst.ts`. `yarn test` now runs unit tests and type tests.

packages/messenger/tsconfig.build.json
```diff
@@ -6,5 +6,6 @@
     "rootDir": "./src"
   },
   "references": [],
-  "include": ["../../types", "./src"]
+  "include": ["../../types", "./src"],
+  "exclude": ["**/*.tst.ts"]
```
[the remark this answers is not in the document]
cryptodev-2s: Good catch. `exclude` replaces the inherited list, so I restored `**/*.test.ts` alongside `**/*.tst.ts`.

packages/messenger/src/Messenger.ts
```diff
@@ -1129,6 +1170,60 @@ export class Messenger<
...
+      >,
+      DelegatedActions
+    >;
+    events: RequireExhaustive<
+      NotNamespacedBy<
+        MessengerNamespace<Delegatee>,
+        MessengerEvents<Delegatee>['type'] & Event['type']
+      >,
+      DelegatedEvents
+    >;
```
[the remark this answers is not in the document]
cryptodev-2s: Good catch. The required set no longer intersects with the source Action/Event types, so unsupported dependencies still show up as missing. Added tstyche coverage for that case and for missing events.

packages/messenger/src/Messenger.ts
```diff
@@ -1129,6 +1170,65 @@ export class Messenger<
...
+   *   action type defined on the delegatee that is **not** under its own
+   *   namespace.
+   * @param args.events - The event types to delegate. Must include every event
+   *   type defined on the delegatee that is **not** under its own namespace.
+   * @param args.messenger - The messenger to delegate to.
+   * @template Delegatee - The messenger the actions/events are delegated to.
+   * @template DelegatedActions - An array of delegated action type strings.
+   * @template DelegatedEvents - An array of delegated event type strings.
+   */
+  delegateAll<
```
mcmire, line 1202: I see that you moved all of the tests to `Messenger.tst.ts`. Maybe we should still have one test which verifies that the behavior of `delegateAll` is as we expect (i.e., it delegates the requested actions and events to the given messenger)?

mcmire (commented):
One more thing.

packages/messenger/src/Messenger.ts
cryptodev-2s: We still have those in Messenger.test.ts under describe('delegateAll') https://github.com/MetaMask/core/pull/8338/changes#diff-7c70b8b38fb189e7496767fa9487bf0bbafe014e2fa5f9efb3224fda03fdea27R1934: one for actions + events, and one for actions with an empty events list. Only the compile-time cases moved to Messenger.tst.ts

mcmire: Oops, you're right. Thanks.

mcmire (dismissed):
LGTM!

cryptodev-2s:
@SocketSecurity ignore npm/tstyche@5.0.2
Used as a dev dependency only

mcmire (approved):
LGTM.

PR #8347, opened 2026-03-31
feat: Implement `FiatStrategy` submit flow with order polling and relay execution

OGPoyraz:
## Explanation

Implements `FiatStrategy` submit flow with order polling and relay execution.

- Implement `submitFiatQuotes` to poll the on-ramp order via `RampsController:getOrder` until completion, then re-quote and submit the relay leg with the settled crypto amount
- Add order validation that verifies the completed order's asset and chain match the expected fiat asset before proceeding with relay
- Update relay quote request filtering to support exact-input max amount requests used by the fiat submit re-quote path
- Add `orderCode` field to `TransactionFiatPayment` and `RampsControllerGetOrderAction` to `AllowedActions` to enable order polling from the fiat strategy
- Fail-closed on terminal order statuses (failed/cancelled), polling timeout with last-status reporting, asset mismatches, and invalid crypto amounts

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new fiat submission path that polls on-ramp orders and then re-quotes/submits Relay based on the settled amount; mistakes here can cause failed submissions or unexpected amounts. Also changes Relay quote filtering logic, which may affect which quote requests are processed.
> 
> **Overview**
> Implements the `FiatStrategy` submit flow: `submitFiatQuotes` now polls `RampsController:getOrder` until completion (with timeout, terminal failure handling, and transient error tolerance), validates the completed order’s asset/chain, converts the settled crypto amount to raw units, re-quotes Relay with an exact-input max-amount request, checks re-quote slippage (5% cap), and submits via `submitRelayQuotes`.
> 
> Updates `TransactionPayController` so changes to `fiatPayment` trigger quote refreshes, and changing the fiat payment method attempts to sync `RampsController:setSelectedToken` (best-effort, errors swallowed). Extends types to include `TransactionFiatPayment.orderId` and adds `RampsController:getOrder`/`RampsController:setSelectedToken` to `AllowedActions`, plus broad new test coverage for the fiat submit flow and token-selection behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45c2c795d92f32380e2c432ce32393f8632cf94d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
@@ -26,6 +26,7 @@ const log = createModuleLogger(projectLogger, 'fiat-submit');
 
 const ORDER_POLL_INTERVAL_MS = 1000;
 const ORDER_POLL_TIMEOUT_MS = 10 * 60 * 1000;
+const MAX_SLIPPAGE_PERCENT = 5;
 
 const TERMINAL_FAILURE_STATUSES: RampsOrderStatus[] = [
   RampsOrderStatus.Cancelled,
@@ -56,36 +57,35 @@ export async function submitFiatQuotes(
   }
 
   const state = messenger.call('TransactionPayController:getState');
-  const orderCode =
-    state.transactionData[transactionId]?.fiatPayment?.orderCode;
+  const orderId = state.transactionData[transactionId]?.fiatPayment?.orderId;
 
-  if (!orderCode) {
-    throw new Error('Missing order code for fiat submission');
+  if (!orderId) {
+    throw new Error('Missing order ID for fiat submission');
   }
 
-  const parsedOrderCode = parseOrderCode(orderCode);
+  const parsedOrder = parseOrderId(orderId);
 
-  if (!parsedOrderCode) {
-    throw new Error(`Invalid order code format: ${orderCode}`);
+  if (!parsedOrder) {
+    throw new Error(`Invalid order ID format: ${orderId}`);
   }
 
   log('Starting fiat order polling', {
-    orderCode,
-    providerCode: parsedOrderCode.providerCode,
+    orderId,
+    providerCode: parsedOrder.providerCode,
     transactionId,
   });
 
   const order = await waitForOrderCompletion({
     messenger,
-    orderCode: parsedOrderCode.orderCode,
-    providerCode: parsedOrderCode.providerCode,
+    orderCode: parsedOrder.orderCode,
+    providerCode: parsedOrder.providerCode,
     transactionId,
     walletAddress,
   });
 
   log('Fiat order completed', {
     cryptoAmount: order.cryptoAmount,
-    orderCode,
+    orderId,
     transactionId,
   });
 
@@ -93,15 +93,15 @@ export async function submitFiatQuotes(
 }
 
 /**
- * Parses a normalized order code string into its provider and order components.
+ * Parses a normalized order ID string into its provider and order components.
  *
- * @param orderCode - Order code in `/providers/{providerCode}/orders/{orderCode}` format.
+ * @param orderId - Order ID in `/providers/{providerCode}/orders/{orderCode}` format.
  * @returns The parsed provider and order codes, or `null` if the format is invalid.
  */
-function parseOrderCode(
-  orderCode: string,
+function parseOrderId(
+  orderId: string,
 ): { orderCode: string; providerCode: string } | null {
-  const parts = orderCode.split('/').filter(Boolean);
+  const parts = orderId.split('/').filter(Boolean);
 
   if (parts.length < 4 || parts[0] !== 'providers' || parts[2] !== 'orders') {
     return null;
@@ -178,6 +178,51 @@ function validateOrderAsset({
     throw new Error(
       `Fiat order chain mismatch for transaction ${transactionId}: ` +
         `expected ${expectedChainId}, got ${orderChainId}`,
+    );
+  }
+}
+
+/**
+ * Validates that the re-quoted relay target output hasn't drifted beyond the
+ * acceptable slippage threshold compared to the original quote shown to the user.
+ *
+ * @param options - The validation options.
+ * @param options.originalTargetRaw - Raw target amount from the original relay quote.
+ * @param options.reQuotedTargetRaw - Raw target amount from the re-quoted relay.
+ * @param options.transactionId - Transaction ID for error reporting.
+ */
+function validateRelaySlippage({
+  originalTargetRaw,
+  reQuotedTargetRaw,
+  transactionId,
+}: {
+  originalTargetRaw: string;
+  reQuotedTargetRaw: string;
+  transactionId: string;
+}): void {
+  const original = new BigNumber(originalTargetRaw);
+  const reQuoted = new BigNumber(reQuotedTargetRaw);
+
+  if (!original.gt(0) || !reQuoted.gt(0)) {
+    return;
+  }
+
+  const slippagePercent = original
+    .minus(reQuoted)
+    .dividedBy(original)
+    .multipliedBy(100);
+
+  log('Relay slippage check', {
+    originalTargetRaw,
+    reQuotedTargetRaw,
+    slippagePercent: slippagePercent.toFixed(2),
+    transactionId,
+  });
+
+  if (slippagePercent.gt(MAX_SLIPPAGE_PERCENT)) {
+    throw new Error(
+      `Relay re-quote slippage too high for transaction ${transactionId}: ` +
+        `${slippagePercent.toFixed(2)}% exceeds ${MAX_SLIPPAGE_PERCENT}% max`,
     );
   }
 }
@@ -312,6 +357,13 @@ async function submitRelayAfterFiatCompletion({
   if (!relayQuotes.length) {
     throw new Error('No relay quotes returned for completed fiat order');
   }
+
+  const originalRelayQuote = quotes[0].original.relayQuote;
+  validateRelaySlippage({
+    originalTargetRaw: originalRelayQuote.details.currencyOut.amount,
+    reQuotedTargetRaw: relayQuotes[0].original.details.currencyOut.amount,
+    transactionId,
+  });
 
   log('Received relay quotes for completed fiat order', {
     relayQuoteCount: relayQuotes.length,
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
@@ -234,8 +234,9 @@ async function waitForOrderCompletion({
     }
 
     if (Date.now() - startTime >= ORDER_POLL_TIMEOUT_MS) {
-      const statusDetail = lastStatus ? ` (last status: ${lastStatus})` : '';
-      throw new Error(`Fiat order polling timed out${statusDetail}`);
+      throw new Error(
+        `Fiat order polling timed out (last status: ${lastStatus})`,
+      );
     }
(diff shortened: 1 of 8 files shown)


packages/transaction-pay-controller/src/types.ts
```diff
@@ -211,6 +215,9 @@ export type TransactionFiatPayment = {
   /** Entered fiat amount for the selected payment method. */
   amountFiat?: string;
 
+  /** Order identifier - `orderCode` specifically used as RampsService:getOrder parameter in normalized format (/providers/{provider}/orders/{id}). */
+  orderCode?: string;
```
OGPoyraz: Originally we talked Ramps team to add metadata into fiat orders and also filtering mechanism.
Due to limitation of timeframe this may not be option to deliver besides headless ramp, hence we are adding `orderCode` (meaning order id) into `TransactionFiatPayment`

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -78,13 +78,15 @@ export async function getRelayQuotes(
...
-      )
+      .filter((singleRequest) => {
+        const hasTargetMinimum = singleRequest.targetAmountMinimum !== '0';
+        const isPostQuote = Boolean(singleRequest.isPostQuote);
+        const isExactInputRequest =
+          Boolean(singleRequest.isMaxAmount) &&
+          new BigNumber(singleRequest.sourceTokenAmount).gt(0);
+
+        return hasTargetMinimum || isPostQuote || isExactInputRequest;
+      })
```
OGPoyraz, line 89: The fiat re-quote request could get dropped by this filter if `targetAmountMinimum` happened to be '0' - it's not a post-quote, so the second condition wouldn't save it. The `isExactInputRequest` check distinguishes real exact-input requests (positive source amount + max amount flag) from empty gas fee token requests (zero source amount), ensuring the fiat re-quote passes through.

vinistevam (dismissed):

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,335 @@
...
+  QuoteRequest,
+  TransactionPayControllerMessenger,
+} from '../../types';
+import { getRelayQuotes } from '../relay/relay-quotes';
+import { submitRelayQuotes } from '../relay/relay-submit';
+import type { RelayQuote } from '../relay/types';
+
+const log = createModuleLogger(projectLogger, 'fiat-submit');
+
+const ORDER_POLL_INTERVAL_MS = 1000;
```
matthewwalsh0, line 26: Minor, I added feature flags for this on Relay, could extend for fiat in future.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,335 @@
...
-  return { transactionHash: undefined };
+  const { messenger, transaction } = request;
+  const transactionId = transaction.id;
+  const walletAddress = transaction.txParams.from as Hex | undefined;
+
+  if (!walletAddress) {
+    throw new Error('Missing wallet address for fiat submission');
+  }
+
+  const state = messenger.call('TransactionPayController:getState');
```
matthewwalsh0, line 58: Minor, should this be in the request to make it more "pure"?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,335 @@
...
+  providerCode: string;
+  transactionId: string;
+  walletAddress: string;
+}): Promise<RampsOrder> {
+  const startTime = Date.now();
+  let lastStatus: string | undefined;
+
+  while (true) {
+    const order = await messenger.call(
+      'RampsController:getOrder',
```
matthewwalsh0: To clarify, does this trigger a network request in the `RampsController`?

Would it be more efficient if they owned the polling and we listened to a messenger event like `RampsController:quoteStatusChange` or `RampsController:stateChange` if it's updated in state?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,335 @@
...
+  expectedAsset: TransactionPayFiatAsset;
+  orderCrypto: RampsOrderCryptoCurrency | undefined;
+  transactionId: string;
+}): void {
+  const orderAssetId = orderCrypto?.assetId?.toLowerCase();
+  const expectedAssetId = expectedAsset.caipAssetId.toLowerCase();
+  const expectedChainId = expectedAssetId.split('/')[0];
+  const orderChainId = orderCrypto?.chainId?.toLowerCase();
+
+  if (orderAssetId && orderAssetId !== expectedAssetId) {
```
matthewwalsh0, line 169: Oh we already have an order ID? Is this generated by the provider or Ramps?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,335 @@
...
+  };
+
+  log('Re-quoting relay from completed fiat order', {
+    completedOrderAmount: order.cryptoAmount,
+    relayRequest,
+    sourceAmountRaw,
+    transactionId,
+  });
+
+  const relayQuotes = await getRelayQuotes({
```
matthewwalsh0, line 358: Do we need to validate the final quote to check the fees are within our original slippage compared to what we showed the user?

Or is that a separate PR?

packages/transaction-pay-controller/src/types.ts
matthewwalsh0: Minor, `orderId` or `rampsId` to be more generic?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Good call, I'll extend the feature flags for fiat polling in a follow-up, not only this but also tokens per type.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Fair point - I can work on this on separate PR as this will touch beyond fiat scope.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Generated by the provider (e.g Transak). The mobile app receives it after checkout completion and writes it to state via `updateFiatPayment`. The normalized format `/providers/{provider}/orders/{id}` is what Ramps uses internally.

So I expect them to execute a callback after successfully checkout passed while starting headless ramps (meaning we have `orderId`) then regularly fetch in the TPC.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Yes it triggers a network request via `RampsController:getOrder`. Event-based via `stateChange` would be cleaner - agreed it's the better long-term approach. For now polling keeps fiat self-contained without coupling to RampsController's internal state updates. 

In fact this we should be able to just subscribe -order status change-, I will bring this topic once we proceed with other blockers.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Good catch - added `validateRelaySlippage` that compares the re-quoted relay's target output against the original. Fails if slippage exceeds 5%.

packages/transaction-pay-controller/src/types.ts
OGPoyraz: Done - renamed to `orderId`

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,397 @@
...
+  log('Relay slippage check', {
+    originalTargetRaw,
+    reQuotedTargetRaw,
+    slippagePercent: slippagePercent.toFixed(2),
+    transactionId,
+  });
+
+  if (slippagePercent.gt(MAX_SLIPPAGE_PERCENT)) {
+    throw new Error(
+      `Relay re-quote slippage too high for transaction ${transactionId}: ` +
```
matthewwalsh0: Minor, `transactionId` might be excessive since the error is stored on the transaction.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,397 @@
...
+    transaction,
+  });
+
+  if (!relayQuotes.length) {
+    throw new Error('No relay quotes returned for completed fiat order');
+  }
+
+  const originalRelayQuote = quotes[0].original.relayQuote;
+  validateRelaySlippage({
+    originalTargetRaw: originalRelayQuote.details.currencyOut.amount,
```
matthewwalsh0, line 371: How stable is the input amount returned from the fiat order?

If it under-performs and returns less, then we're also adding to the slippage since the output amount from Relay will be smaller.

Should we compare the output amount minus the input amount for a fairer comparison of provider cost slippage?

Not crucial now, can implement if needed in future PR.

matthewwalsh0 (dismissed):

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -1,14 +1,397 @@
...
+import type { RelayQuote } from '../relay/types';
+import type { TransactionPayFiatAsset } from './constants';
 import type { FiatQuote } from './types';
+import { deriveFiatAssetForFiatPayment } from './utils';
+
+const log = createModuleLogger(projectLogger, 'fiat-submit');
+
+const ORDER_POLL_INTERVAL_MS = 1000;
+const ORDER_POLL_TIMEOUT_MS = 10 * 60 * 1000;
+const MAX_SLIPPAGE_PERCENT = 5;
```
matthewwalsh0, line 28: Minor, valuable to be a feature flag in future?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Removed

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: I've been told it's not guaranteed that this will be the crypto amount user bought on-chain and we shouldn't rely on this to proceed the intent.

Instead we will read transaction receipt on-chain and use the transferred amount and it will be on separate PR. I will address slippage concern there.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: I will address this here: https://github.com/MetaMask/core/pull/8631

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

PR #8356, opened 2026-04-01
fix assetId in token selector

bergarces:
## Explanation

`assetId` is being populated with the evm address for evm assets, which is both confusing and makes it cumbersome to extract the real `assetId` in the clients.

## Checklist


bergarces:
@metamaskbot publish-previews

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -64,14 +73,15 @@ type MultichainAccountType = Exclude<
 export type Asset = (
   | {
       accountType: EvmAccountType;
-      assetId: Hex; // This is also the address for EVM tokens
+      assetId: CaipAssetType;
```
bergarces, line 77: This is a breaking change result from the bad decision in BIP44 days of setting `assetId` as the address for evm assets, which is very confusing and means we have no access to the assetId at all.

This is the main change of this PR.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -64,14 +73,15 @@ type MultichainAccountType = Exclude<
...
       address: Hex;
       chainId: Hex;
     }
   | {
       accountType: MultichainAccountType;
-      assetId: `${string}:${string}/${string}:${string}`;
-      chainId: `${string}:${string}`;
+      assetId: CaipAssetType;
+      address: CaipAssetType;
+      chainId: CaipChainId;
```
bergarces, line 85: No breaking changes here, just using the correct type.

Added `address` here too, because otherwise the type is annoying to handle (it won't let you access `asset.address` as it doesn't always exist. The PRs implementing this breaking change show how this can simplify the code.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -222,7 +237,15 @@ const selectAllEvmAccountNativeBalances = createAssetListSelector(
...
-          assetId: nativeToken.address,
+          assetId:
+            SPOT_PRICES_SUPPORT_INFO[
+              chainId as keyof typeof SPOT_PRICES_SUPPORT_INFO
+            ] ?? toCaipAssetType(
+              KnownCaipNamespace.Eip155,
+              hexToNumber(chainId).toString(),
+              'erc20',
+              ZERO_ADDRESS,
+            ),
```
bergarces, line 249: We need a default here, and `erc20:0x0000...0000` is a better default than `slip44:60`, as it usually works even when there is a slip44 entry and it doesn't assume it's ETH. It does assume that the native token is the zero address, but there is only a handful of chains for which that is not the case, and we hopefully have the captured in the settings.

bergarces:
@metamaskbot publish-previews

PR #8384, opened 2026-04-06
Ensure refs in tsconfig files are synced with internal deps

### LINKED ISSUE #966: Add linting for TypeScript configuration files
It is easy for the TypeScript files to be misconfigured due to requirements that aren't obvious.

At the project level, the following should hold true:

* In `tsconfig.json`, `references` should cover all public workspace packages, and all paths should point to valid files and end with `/tsconfig.json` [note: this is not the case now but we can add it for consistency with the next rule]
* In `tsconfig.build.json`, `references` should match `references` in `tsconfig.json` except that all paths should end with `/tsconfig.build.json`.

For each public workspace package, the following should hold true:

* `tsconfig.json` should be present and:
  * `extends` should be "../../tsconfig.packages.json"
  * `compilerOptions.baseUrl` should be "./"
  * all paths in `references` (if present) should point to valid files and end with `/tsconfig.json` [note: this is not the case now but we can add it for consistency with the next rule]
  * `include` should include "../../types" and "./src"
* `tsconfig.build.json` should be present and:
  * `extends` should be "../../tsconfig.packages.build.json"
  * `compilerOptions.baseUrl` should be "./"
  * `compilerOptions.dist` should be "./dist"
  * `compilerOptions.rootDir` should be "./src"
  * all paths in `references` (if present) should match those in `tsconfig.json` except they should end with `/tsconfig.build.json` and should be valid files
  * `include` should include "../../types" and "./src"

mcmire:
### Problem

We use project references to tell TypeScript and other tools about the structure of the monorepo. Both the root- and package-level tsconfig files have a `references` field which is used to link packages together. All packages that we want to build and publish need to be present in `references` within the root `tsconfig.json` and `tsconfig.build.json` files, and the `references` field in both root-level and package-level tsconfig files need to list all internal dependencies.

However, it's a pain to keep references up to date, especially as as packages are added or dependencies are updated. There are cases in the past, in fact, where packages were not published or type errors occurred because the references were not correctly kept in sync.

### Solution

We don't need to do this manual work. We can infer root references by scanning workspaces, and we can infer references for an individual package by scanning that package's `dependencies` for other workspace packages.

This commit adds a script that can be run on both the root workspace and child workspaces to check and/or regenerate the `references` field for each tsconfig file so that it is kept in sync with dependencies throughout the monorepo. There are also new package scripts for use:

- Root level
  - **`lint:tsconfigs`:** Validates the root `tsconfig.json` and `tsconfig.build.json`, erroring if any are out of sync.
  - **`lint:tsconfigs:all`:** Validates all `tsconfig.json` and `tsconfig.build.json` across the monorepo, erroring if any are out of sync.
  - **`lint:tsconfigs:fix`:** Regenerates the root `tsconfig.json` and `tsconfig.build.json`.
  - **`lint:tsconfigs:fix:all`:**  Regenerates `tsconfig.json` and `tsconfig.build.json` files across the monorepo.
- Package level
  - **`lint:tsconfigs`:** Validates `tsconfig.json` and `tsconfig.build.json` in the package, erroring if any are out of sync.
  - **`lint:tsconfigs:fix`:**  Regenerates `tsconfig.json` and `tsconfig.build.json` in the package.

As the names indicate, tsconfig file validation is also a part of the lint pipeline, so CI will now fail if any files are out of date. To make sure this doesn't happen in the future, this commit also corrects tsconfig files across the board to add missing references or remove extra references.

### Why hand-roll a solution?

There are some existing solutions for syncing references within the TypeScript community, some of which are documented in [this issue][1]. However, I reviewed them and rejected them for various reasons:

- **[`typescript-monorepo-toolkit`][2]:** Doesn't update references for individual packages based on `dependencies`
- **[`update-ts-references`][3]:** Works, but re-sorts all references, and drops the leading `./` from references in root tsconfig files; also, generated files fail Prettier validation
- **[`@monorepo-utils/workspaces-to-typescript-project-references`][4]:** Works, but re-sorts all references, and drops the leading `./` from references in root tsconfig files; also generated files fail Prettier validation

My thought is that by building our own script, we get to control the exact changes that are made to tsconfig files, and we get to run all files through Oxfmt so that engineers do not have to reformat them manually.

[1]: https://github.com/microsoft/TypeScript/issues/25376
[2]: https://github.com/Bnaya/typescript-monorepo-toolkit
[3]: https://github.com/eBayClassifiedsGroup/update-ts-references
[4]: https://github.com/azu/monorepo-utils/tree/master/packages/@monorepo-utils/workspaces-to-typescript-project-references

## References

Closes #966.

Also closes https://consensyssoftware.atlassian.net/browse/WPC-649.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Wide monorepo tsconfig and declared-dependency changes can affect composite builds; a few packages gain runtime deps to match messenger/types usage.
> 
> **Overview**
> Adds **`lint-tsconfigs`** (`scripts/lint-tsconfigs/lint-tsconfigs.mts`) to validate or regenerate `references` in root and package `tsconfig.json` / `tsconfig.build.json` from workspace layout and each package’s internal `dependencies`. Root **`lint`** and CI now run **`lint:tsconfigs:all`**; every package gets **`lint:tsconfigs`** / **`lint:tsconfigs:fix`** scripts.
> 
> This PR **realigns references repo-wide** (adds missing project refs, drops extras, normalizes path formatting) and fixes a few **dependency/tsconfig mismatches** the linter surfaced: e.g. **`@metamask/preferences-controller`** on `eip-5792-middleware`, **`@metamask/controller-utils`** on `keyring-controller`, and README dependency edges. ESLint now covers **`scripts/**/*.{ts,mts}`** and **`.mts`** for TypeScript rules.
> 
> Small follow-on code tweaks: stricter typing in **`permission-middleware`**, removal of a **`PerpsController`** `@ts-expect-error` after reference fixes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2886d3bd776bcb98e5ab33fba62cfd355cf0f387. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/accounts-controller/tsconfig.build.json
```diff
@@ -10,9 +10,18 @@
...
+      "path": "../keyring-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../network-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 23: This dependency is present in `devDependencies` but was missing as a reference.

packages/accounts-controller/tsconfig.json
```diff
@@ -10,8 +10,15 @@
...
-    { "path": "../network-controller" },
-    { "path": "../messenger" }
+    {
+      "path": "../network-controller"
+    },
+    {
+      "path": "../messenger"
+    },
+    {
+      "path": "../controller-utils"
```
mcmire, line 20: This dependency is present in `devDependencies` but was missing as a reference.

packages/assets-controller/tsconfig.json
```diff
@@ -4,16 +4,51 @@
...
+      "path": "../network-enablement-controller"
+    },
+    {
+      "path": "../phishing-controller"
+    },
+    {
+      "path": "../preferences-controller"
+    },
+    {
+      "path": "../controller-utils"
```
mcmire: All of these dependencies are present in `dependencies` but were missing as references.

packages/assets-controllers/tsconfig.build.json
```diff
@@ -6,22 +6,60 @@
...
+      "path": "../storage-service/tsconfig.build.json"
+    },
+    {
+      "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../phishing-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../multichain-account-service/tsconfig.build.json"
```
mcmire, line 58: These dependencies are present in `dependencies` but were missing as references.

packages/assets-controllers/tsconfig.json
```diff
@@ -5,22 +5,60 @@
...
+      "path": "../permission-controller"
+    },
+    {
+      "path": "../storage-service"
+    },
+    {
+      "path": "../transaction-controller"
+    },
+    {
+      "path": "../multichain-account-service"
```
mcmire, line 57: These dependencies are present in `dependencies` but were missing as references.

packages/base-controller/tsconfig.build.json
```diff
@@ -7,10 +7,10 @@
   },
   "references": [
     {
-      "path": "../controller-utils/tsconfig.build.json"
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 10: `base-controller` no longer depends on `controller-utils`. It now lists `json-rpc-engine` as a dev dependency.

packages/base-controller/tsconfig.json
```diff
@@ -4,9 +4,6 @@
     "baseUrl": "./"
   },
   "references": [
-    {
-      "path": "../controller-utils"
```
mcmire, line 8: `tsconfig.build.json` and `tsconfig.json` for `base-controller` were out of sync.

packages/bridge-controller/tsconfig.build.json
```diff
@@ -6,17 +6,48 @@
...
+      "path": "../multichain-network-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../remote-feature-flag-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../assets-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../eth-json-rpc-provider/tsconfig.build.json"
```
mcmire, line 43: This dependency is present in `devDependencies` but was missing as a reference.

packages/bridge-controller/tsconfig.build.json
```diff
@@ -6,17 +6,48 @@
...
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 46: These dependencies are present in `dependencies` but were missing as references.

packages/bridge-controller/tsconfig.json
```diff
@@ -5,18 +5,48 @@
...
+      "path": "../profile-sync-controller"
+    },
+    {
+      "path": "../remote-feature-flag-controller"
+    },
+    {
+      "path": "../assets-controller"
+    },
+    {
+      "path": "../eth-json-rpc-provider"
```
mcmire, line 45: `tsconfig.json` and `tsconfig.build.json` were out of sync for `bridge-controller`.

packages/bridge-status-controller/tsconfig.build.json
```diff
@@ -6,15 +6,39 @@
...
+      "path": "../gas-fee-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../polling-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../keyring-controller/tsconfig.build.json"
```
mcmire, line 37: These are present in `dependencies`.

packages/bridge-status-controller/tsconfig.json
```diff
@@ -5,16 +5,39 @@
...
+      "path": "../transaction-controller"
+    },
+    {
+      "path": "../profile-sync-controller"
+    },
+    {
+      "path": "../gas-fee-controller"
+    },
+    {
+      "path": "../keyring-controller"
```
mcmire, line 39: This file was out of sync with `tsconfig.build.json`.

packages/chain-agnostic-permission/tsconfig.build.json
```diff
@@ -7,9 +7,12 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../controller-utils/tsconfig.build.json" },
-    { "path": "../network-controller/tsconfig.build.json" },
```
mcmire, line 11: No longer a dependency.

packages/claims-controller/tsconfig.build.json
```diff
@@ -17,6 +17,9 @@
     },
     {
       "path": "../keyring-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 22: New dependency.

packages/claims-controller/tsconfig.json
```diff
@@ -5,16 +5,19 @@
...
     {
-      "path": "../profile-sync-controller/tsconfig.json"
+      "path": "../profile-sync-controller"
     },
     {
-      "path": "../keyring-controller/tsconfig.json"
+      "path": "../keyring-controller"
+    },
+    {
+      "path": "../controller-utils"
```
mcmire, line 20: For `tsconfig.json` files, the final `/tsconfig.json` is dropped from references.

Also, `controller-utils` is a new dependency.

packages/client-controller/tsconfig.json
```diff
@@ -1,5 +1,15 @@
 {
   "extends": "../../tsconfig.packages.json",
-  "compilerOptions": { "baseUrl": "." },
-  "include": ["src"]
+  "compilerOptions": {
+    "baseUrl": "."
+  },
+  "include": ["src"],
+  "references": [
```
mcmire, line 7: This did not include any references at all 😬

packages/composable-controller/tsconfig.build.json
```diff
@@ -11,6 +11,9 @@
     },
     {
       "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../json-rpc-engine/tsconfig.build.json"
```
mcmire, line 16: New dev dependency.

packages/controller-utils/tsconfig.build.json
```diff
@@ -5,5 +5,6 @@
     "outDir": "./dist",
     "rootDir": "./src"
   },
-  "include": ["../../types", "./src"]
+  "include": ["../../types", "./src"],
+  "references": []
```
mcmire: This property isn't required (`controller-utils` has no internal dependencies) but is merely added for completeness.

packages/earn-controller/tsconfig.build.json
```diff
@@ -20,6 +20,9 @@
     },
     {
       "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 25: New dependency.

packages/eip-5792-middleware/tsconfig.build.json
```diff
@@ -6,12 +6,15 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../network-controller/tsconfig.build.json" },
```
mcmire, line 9: This dependency is no longer present.

packages/eip-5792-middleware/tsconfig.build.json
```diff
@@ -6,12 +6,15 @@
...
-    { "path": "../transaction-controller/tsconfig.build.json" },
-    { "path": "../messenger/tsconfig.build.json" },
-    { "path": "../accounts-controller/tsconfig.build.json" },
```
mcmire, line 12: `accounts-controller` is no longer present.

packages/eip-7702-internal-rpc-middleware/tsconfig.build.json
```diff
@@ -6,12 +6,9 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../network-controller/tsconfig.build.json" },
```
mcmire, line 9: Many dependencies are no longer present.

packages/json-rpc-engine/tsconfig.build.json
```diff
@@ -5,5 +5,6 @@
     "outDir": "./dist",
     "rootDir": "./src"
   },
-  "include": ["../../types", "./src"]
+  "include": ["../../types", "./src"],
+  "references": []
```
mcmire: `json-rpc-engine` has no internal dependencies, but this is merely added for completeness.

packages/keyring-controller/tsconfig.build.json
```diff
@@ -9,9 +9,6 @@
     {
       "path": "../base-controller/tsconfig.build.json"
     },
-    {
-      "path": "../message-manager/tsconfig.build.json"
```
mcmire, line 13: No longer a dependency.

packages/multichain-account-service/tsconfig.build.json
```diff
@@ -6,10 +6,21 @@
...
+      "path": "../accounts-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../keyring-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 25: New dev dependency.

packages/multichain-api-middleware/tsconfig.build.json
```diff
@@ -7,10 +7,24 @@
...
+      "path": "../json-rpc-engine/tsconfig.build.json"
+    },
+    {
+      "path": "../network-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../permission-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 26: New dependency.

packages/multichain-api-middleware/tsconfig.build.json
```diff
@@ -7,10 +7,24 @@
...
+    },
+    {
+      "path": "../multichain-transactions-controller/tsconfig.build.json"
```
mcmire, line 29: New dev dependency.

packages/multichain-network-controller/tsconfig.build.json
```diff
@@ -6,11 +6,24 @@
...
+      "path": "../controller-utils/tsconfig.build.json"
+    },
+    {
+      "path": "../network-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../keyring-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 25: New dependency.

packages/name-controller/tsconfig.build.json
```diff
@@ -11,6 +11,9 @@
     },
     {
       "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 16: New dependency.

packages/notification-services-controller/tsconfig.build.json
```diff
@@ -18,6 +18,9 @@
     },
     {
       "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 26: New dependency.

packages/ramps-controller/tsconfig.json
```diff
@@ -4,6 +4,16 @@
...
-  "references": [{ "path": "../base-controller" }, { "path": "../messenger" }],
+  "references": [
+    {
+      "path": "../base-controller"
+    },
+    {
+      "path": "../messenger"
+    },
+    {
+      "path": "../controller-utils"
```
mcmire, line 21: This file is out of sync with `tsconfig.build.json`.

packages/remote-feature-flag-controller/tsconfig.build.json
```diff
@@ -6,8 +6,15 @@
...
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../messenger/tsconfig.build.json" }
+    {
+      "path": "../base-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 16: New dependency.

packages/seedless-onboarding-controller/tsconfig.build.json
```diff
@@ -10,10 +10,10 @@
       "path": "../base-controller/tsconfig.build.json"
     },
     {
-      "path": "../message-manager/tsconfig.build.json"
```
mcmire, line 13: `messenger-manager` removed as a dependency, `messenger` added.

packages/shield-controller/tsconfig.build.json
```diff
@@ -6,10 +6,21 @@
...
+      "path": "../signature-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 22: New dependency.

packages/signature-controller/tsconfig.build.json
```diff
@@ -21,9 +21,6 @@
     {
       "path": "../gator-permissions-controller/tsconfig.build.json"
     },
-    {
-      "path": "../message-manager/tsconfig.build.json"
```
mcmire, line 25: `message-manager` removed as a dependency.

packages/subscription-controller/tsconfig.build.json
```diff
@@ -20,6 +20,9 @@
     },
     {
       "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire, line 25: New dependency.

packages/transaction-controller/tsconfig.build.json
```diff
@@ -6,15 +6,42 @@
...
+      "path": "../network-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../remote-feature-flag-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../connectivity-controller/tsconfig.build.json"
```
mcmire, line 37: New dev dependencies.

packages/transaction-pay-controller/tsconfig.build.json
```diff
@@ -38,6 +38,9 @@
     },
     {
       "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 40: New dependency.

packages/user-operation-controller/tsconfig.build.json
```diff
@@ -29,6 +29,12 @@
     },
     {
       "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../approval-controller/tsconfig.build.json"
```
mcmire, line 34: New dependency.

packages/user-operation-controller/tsconfig.build.json
```diff
@@ -29,6 +29,12 @@
...
+    },
+    {
+      "path": "../eth-block-tracker/tsconfig.build.json"
```
mcmire, line 37: New dev dependency.

tsconfig.json
```diff
@@ -226,6 +226,18 @@
     },
     {
       "path": "./packages/user-operation-controller"
+    },
+    {
+      "path": "./packages/eip-5792-middleware"
```
mcmire, line 282: Not sure how these were never added to the root tsconfig 🤔

eslint.config.mjs
```diff
@@ -105,8 +105,7 @@ const config = createConfig([
       '**/*.test.{js,ts}',
       '**/test/**/*.{js,ts}',
       '**/tests/**/*.{js,ts}',
-      'scripts/*.ts',
-      'scripts/create-package/**/*.ts',
+      'scripts/**/*.ts',
```
mcmire: We've added a new subdirectory to `scripts` so rather than adding an extra entry here we can consolidate them all into one glob.

packages/keyring-controller/tsconfig.build.json
```diff
@@ -10,10 +10,10 @@
       "path": "../base-controller/tsconfig.build.json"
     },
     {
-      "path": "../message-manager/tsconfig.build.json"
+      "path": "../messenger/tsconfig.build.json"
     },
     {
-      "path": "../messenger/tsconfig.build.json"
+      "path": "../controller-utils/tsconfig.build.json"
```
mcmire: Not entirely a new dependency, but it should have been added before (see above).

packages/keyring-controller/package.json
```diff
@@ -50,6 +52,7 @@
     "@ethereumjs/util": "^9.1.0",
     "@metamask/base-controller": "^9.0.1",
     "@metamask/browser-passworder": "^6.0.0",
+    "@metamask/controller-utils": "^11.20.0",
```
mcmire: Adding this dependency was necessary to avoid a type error. This should have been added — and the type error should have been present — all along, but for some reason only popped up now. It's possible that with the new changes on this branch, `keyring-controller` was getting built before `controller-utils`; if so, this corrects it.

packages/eth-json-rpc-middleware/package.json
```diff
@@ -68,7 +70,6 @@
   },
   "devDependencies": {
     "@metamask/auto-changelog": "^3.4.4",
-    "@metamask/network-controller": "^30.0.1",
```
mcmire: It turns out that there has been a circular dependency between `network-controller` and `eth-json-rpc-middleware` ever since the latter package was moved from the `eth-json-rpc-middleware` repo, and the problem didn't pop up until now. 

`network-controller` is used in a test to confirm that a type that `eth-json-rpc-middleware` copies from `network-controller` is compatible with the same type from `network-controller` itself. I wasn't sure that this was necessary because now that `eth-json-rpc-middleware` is in this monorepo, if there are any type incompatibilities between types, they should show up locally (and any breaking changes to types should be spotted during development or at least during review). So I opted to remove the test file and thus the dependency on `network-controller`.

packages/eth-json-rpc-middleware/tsconfig.json
```diff
@@ -13,9 +13,6 @@
     {
       "path": "../json-rpc-engine"
     },
-    {
-      "path": "../network-controller"
```
mcmire, line 17: No longer a dependency.

packages/eip-5792-middleware/package.json
```diff
@@ -48,6 +50,7 @@
   },
   "dependencies": {
     "@metamask/messenger": "^1.1.1",
+    "@metamask/preferences-controller": "^23.1.0",
```
mcmire, line 57: Added to avoid a new type error. Similar to https://github.com/MetaMask/core/pull/8384#discussion_r3040833619.

mcmire:
Curious if our team should be co-codeowners of `tsconfig.json` and `tsconfig.build.json` files 🤔 This would make this PR merge a lot faster.

packages/assets-controller/tsconfig.build.json
```diff
@@ -6,17 +6,54 @@
...
+      "path": "../permission-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../phishing-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../preferences-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../accounts-controller/tsconfig.build.json"
```
mcmire, line 43: All references starting with `accounts-controller` are present in `dependencies` but were missing.

packages/assets-controller/tsconfig.json
```diff
@@ -4,16 +4,54 @@
...
+      "path": "../network-enablement-controller"
+    },
+    {
+      "path": "../phishing-controller"
+    },
+    {
+      "path": "../preferences-controller"
+    },
+    {
+      "path": "../accounts-controller"
```
mcmire, line 38: All references starting with `accounts-controller` are present in `dependencies` but were missing.

packages/base-data-service/tsconfig.build.json
```diff
@@ -6,8 +6,9 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../messenger/tsconfig.build.json" },
-    { "path": "../controller-utils/tsconfig.build.json" }
+    {
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 10: `@metamask/base-data-service` no longer relies on `controller-utils`.

packages/bitcoin-regtest-up/tsconfig.json
```diff
@@ -3,6 +3,10 @@
   "compilerOptions": {
     "baseUrl": "./"
   },
-  "references": [{ "path": "../local-node-utils/tsconfig.json" }],
+  "references": [
+    {
+      "path": "../local-node-utils"
```
mcmire, line 8: In `tsconfig.json` files, references don't end in `/tsconfig.json`. (They _could_, but that would be a bigger change.)

packages/eip-7702-internal-rpc-middleware/tsconfig.json
```diff
@@ -4,6 +4,10 @@
     "baseUrl": "./",
     "rootDir": "../.."
   },
-  "references": [],
+  "references": [
```
mcmire, line 7: This file was out of date.

packages/gator-permissions-controller/tsconfig.build.json
```diff
@@ -6,9 +6,18 @@
...
+      "path": "../base-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../network-controller/tsconfig.build.json"
```
mcmire, line 19: New dependency.

packages/java-tron-up/tsconfig.json
```diff
@@ -3,6 +3,10 @@
   "compilerOptions": {
     "baseUrl": "./"
   },
-  "references": [{ "path": "../local-node-utils/tsconfig.json" }],
+  "references": [
+    {
+      "path": "../local-node-utils"
```
mcmire, line 8: Non-standard.

packages/money-account-upgrade-controller/tsconfig.build.json
```diff
@@ -6,14 +6,27 @@
...
   },
   "references": [
-    { "path": "../authenticated-user-storage/tsconfig.build.json" },
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../chomp-api-service/tsconfig.build.json" },
-    { "path": "../delegation-controller/tsconfig.build.json" },
-    { "path": "../keyring-controller/tsconfig.build.json" },
-    { "path": "../messenger/tsconfig.build.json" },
-    { "path": "../network-controller/tsconfig.build.json" },
-    { "path": "../transaction-controller/tsconfig.build.json" }
```
mcmire, line 16: `transaction-controller` is no longer a dependency.

packages/perps-controller/tsconfig.build.json
```diff
@@ -8,18 +8,39 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../account-tree-controller/tsconfig.build.json" },
-    { "path": "../authenticated-user-storage/tsconfig.build.json" },
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../bridge-controller/tsconfig.build.json" },
```
mcmire, line 14: No longer a dependency.

packages/phishing-controller/tsconfig.build.json
```diff
@@ -6,10 +6,21 @@
...
+      "path": "../controller-utils/tsconfig.build.json"
+    },
+    {
+      "path": "../transaction-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../address-book-controller/tsconfig.build.json"
```
mcmire, line 22: New dependency.

packages/polling-controller/tsconfig.build.json
```diff
@@ -6,10 +6,15 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../controller-utils/tsconfig.build.json" },
```
mcmire, line 10: No longer a dependency.

packages/preferences-controller/tsconfig.build.json
```diff
@@ -6,10 +6,12 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../controller-utils/tsconfig.build.json" },
```
mcmire, line 10: `controller-utils` and `keyring-controller` are no longer dependencies.

packages/profile-metrics-controller/tsconfig.build.json
```diff
@@ -6,14 +6,30 @@
...
-    { "path": "../../packages/accounts-controller/tsconfig.build.json" },
-    { "path": "../../packages/base-controller/tsconfig.build.json" },
-    { "path": "../../packages/controller-utils/tsconfig.build.json" },
-    { "path": "../../packages/keyring-controller/tsconfig.build.json" },
-    { "path": "../../packages/messenger/tsconfig.build.json" },
-    { "path": "../../packages/polling-controller/tsconfig.build.json" },
-    { "path": "../../packages/profile-sync-controller/tsconfig.build.json" },
-    { "path": "../../packages/transaction-controller/tsconfig.build.json" }
+    {
+      "path": "../accounts-controller/tsconfig.build.json"
```
mcmire, line 10: Correcting these paths to be standard (no need to say `../../packages`)

packages/react-data-query/tsconfig.build.json
```diff
@@ -5,6 +5,13 @@
     "outDir": "./dist",
     "rootDir": "./src"
   },
-  "references": [{ "path": "../base-data-service/tsconfig.build.json" }],
+  "references": [
+    {
+      "path": "../base-data-service/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
```
mcmire, line 13: New dependency.

packages/sample-controllers/tsconfig.build.json
```diff
@@ -6,10 +6,21 @@
...
+      "path": "../base-data-service/tsconfig.build.json"
+    },
+    {
+      "path": "../controller-utils/tsconfig.build.json"
+    },
+    {
+      "path": "../messenger/tsconfig.build.json"
+    },
+    {
+      "path": "../network-controller/tsconfig.build.json"
```
mcmire, line 22: New dependency.

packages/smart-transactions-controller/tsconfig.build.json
```diff
@@ -6,20 +6,36 @@
     "rootDir": "./src"
   },
   "references": [
-    { "path": "../accounts-controller/tsconfig.build.json" },
```
mcmire, line 9: `accounts-controller` and `approval-controller` no longer dependencies.

packages/smart-transactions-controller/tsconfig.build.json
```diff
@@ -6,20 +6,36 @@
...
-    { "path": "../approval-controller/tsconfig.build.json" },
-    { "path": "../base-controller/tsconfig.build.json" },
-    { "path": "../controller-utils/tsconfig.build.json" },
-    { "path": "../eth-block-tracker/tsconfig.build.json" },
-    { "path": "../eth-json-rpc-provider/tsconfig.build.json" },
-    { "path": "../gas-fee-controller/tsconfig.build.json" },
```
mcmire, line 15: `gas-fee-controller` no longer a dependency.

packages/wallet-cli/tsconfig.json
```diff
@@ -4,10 +4,18 @@
     "baseUrl": "./"
   },
   "references": [
-    { "path": "../base-controller/tsconfig.json" },
-    { "path": "../remote-feature-flag-controller/tsconfig.json" },
-    { "path": "../storage-service/tsconfig.json" },
-    { "path": "../wallet/tsconfig.json" }
+    {
+      "path": "../base-controller"
```
mcmire, line 8: Standardizing these paths.

packages/wallet/tsconfig.json
```diff
@@ -4,18 +4,42 @@
...
-    { "path": "../connectivity-controller/tsconfig.json" },
-    { "path": "../controller-utils/tsconfig.json" },
-    { "path": "../keyring-controller/tsconfig.json" },
-    { "path": "../messenger/tsconfig.json" },
-    { "path": "../network-controller/tsconfig.json" },
-    { "path": "../remote-feature-flag-controller/tsconfig.json" },
-    { "path": "../storage-service/tsconfig.json" },
-    { "path": "../transaction-controller/tsconfig.json" }
+    {
+      "path": "../accounts-controller"
```
mcmire, line 8: Standardizing these paths.

scripts/lint-tsconfigs/utils.mts
```diff
@@ -0,0 +1,598 @@
+import { readJsonFile, readFile } from '@metamask/utils/node';
+import * as commentJson from 'comment-json';
+import execa from 'execa';
+import fs from 'fs';
+import type { FormatConfig } from 'oxfmt';
```
mcmire, line 5: This (and the other files) are `.mts` files because `oxfmt` is ESM-only. Fortunately, we don't need to transform `oxfmt`, we can run these files as ESM files (even though `type` is not `module` in `package.json`).

packages/permission-controller/src/permission-middleware.ts
```diff
@@ -52,11 +48,10 @@ export function createPermissionMiddleware({
   RestrictedMethodParameters,
   Json
 > {
-  const permissionsMiddleware = async (
-    request: JsonRpcRequest<RestrictedMethodParameters>,
-    response: PendingJsonRpcResponse,
-    next: AsyncJsonRpcEngineNextCallback,
-  ): Promise<void> => {
+  const permissionsMiddleware: AsyncJsonrpcMiddleware<
```
mcmire, line 51: This is an interesting one.

After making these changes, I found that building the entire monorepo failed at `perps-controller` with this error:

```
packages/permission-controller/src/permission-middleware.ts:75:32 - error TS2345: Argument of type '(request: JsonRpcRequest<RestrictedMethodParameters>, response: PendingJsonRpcResponse, next: AsyncJsonRpcEngineNextCallback) => Promise<void>' is not assignable to parameter of type 'AsyncJsonrpcMiddleware<(Record<string, Json> & ExactOptionalGuard) | (Record<string, Json> & ExactOptionalGuard & Json[]) | (Json[] & ... 1 more ... & Record<...>) | (Json[] & ... 1 more ... & Json[]), string | ... 4 more ... | null>'.
  Types of parameters 'response' and 'response' are incompatible.
    Type 'PendingJsonRpcResponse<string | number | boolean | Json[] | { [prop: string]: Json; } | null>' is not assignable to type 'PendingJsonRpcResponse'.
      Type 'PendingJsonRpcResponse<string | number | boolean | Json[] | { [prop: string]: Json; } | null>' is not assignable to type 'Omit<{ id: string | number | null; jsonrpc: "2.0"; result: unknown; error?: { data?: (Json & ExactOptionalGuard) | undefined; stack?: string | undefined; code: number; message: string; } | undefined; }, "result">'.
        Types of property 'error' are incompatible.
          Type '{ data?: (import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined; stack?: string | undefined; code: number; message: string; } | undefined' is not assignable to type '{ data?: (import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined; stack?: string | undefined; code: number; message: string; } | undefined'. Two different types with this name exist, but they are unrelated.
            Type '{ data?: (import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined; stack?: string | undefined; code: number; message: string; }' is not assignable to type '{ data?: (import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined; stack?: string | undefined; code: number; message: string; }'. Two different types with this name exist, but they are unrelated.
              Types of property 'data' are incompatible.
                Type '(import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined' is not assignable to type '(import("/Users/elliot/code/metamask/core/node_modules/@metamask/utils/dist/json").Json & ExactOptionalGuard) | undefined'. Two different types with this name exist, but they are unrelated.
                  Type 'string & ExactOptionalGuard' is not assignable to type '(Json & ExactOptionalGuard) | undefined'.
                    Type 'string & ExactOptionalGuard' is not assignable to type 'string & ExactOptionalGuard'. Two different types with this name exist, but they are unrelated.
                      Type 'string & ExactOptionalGuard' is not assignable to type 'ExactOptionalGuard'.
                        Types of property '_exactOptionalGuard' are incompatible.
                          Type 'typeof exactOptionalSymbol | undefined' is not assignable to type 'typeof exactOptionalSymbol | undefined'. Two different types with this name exist, but they are unrelated.
                            Type 'typeof exactOptionalSymbol' is not assignable to type 'typeof exactOptionalSymbol'. Two different types with this name exist, but they are unrelated.

75   return createAsyncMiddleware(permissionsMiddleware);
```

One thing that's different about `perps-controller` than other packages is that it uses `Bundler` as `moduleResolution`. It does this because it relies on a dependency that is ESM-only. This seems to cause TypeScript to see and load ESM versions of some dependencies instead of CommonJS versions. In doing this, it appears that TypeScript is trying to resolve the ESM version of `@metamask/utils` with the CJS version of itself. This causes issues because `@metamask/utils` uses a "branded type" — a type that is intended to be unique — called `ExactOptionalGuard`. One "instance" of `ExactOptionalGuard` is not assignable to the other, so it breaks down.

I am not sure why this error did not occur before, but TypeScript may be building packages in a slightly different order now and so the problem manifested now.

We could fix this in a different PR but I figured since it was related to this work it was better to do it here.

.oxfmtrc.json
```diff
@@ -11,6 +11,9 @@
       ["internal", "parent", "sibling", "index", "unknown"]
     ]
   },
+  "sortPackageJson": {
```
mcmire, line 14: I will revert this change before merging, it just made sorting each `package.json` much easier.

mcmire:
> Okay, I will put this PR on hold until I update the codeowners.

@cryptodev-2s Update: I think once we're happy with this PR, we can just force-merge it instead of updating codeowners. We don't update `tsconfig.json` often enough to warrant a change in codeowners. Plus I don't want to add more maintenance burden to the codeowners file as it's quite high right now.

scripts/lint-tsconfigs/utils.mts
```diff
@@ -0,0 +1,598 @@
...
+  const updatedContent = commentJson.stringify(
+    commentJson.assign(tsconfig.content, { references: [...newReferences] }),
+    null,
+    2,
+  );
+  const { code: updatedFormattedContent, errors } = await oxfmtFormat(
+    tsconfig.fileName,
+    updatedContent,
+    oxfmtRc,
+  );
```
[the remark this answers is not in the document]
mcmire: I think this is okay. I tested the script and it seems to format the JSON files fine.

mcmire:
This is now ready for review again!

scripts/lint-tsconfigs/lint-package-tsconfigs.mts
```diff
@@ -0,0 +1,103 @@
...
+  const tsconfigs = await Promise.all([
+    readTsconfig(packageRoot, 'tsconfig.json'),
+    readTsconfig(packageRoot, 'tsconfig.build.json'),
+  ]);
+  const expectedPackageNames = getExpectedWorkspaceDependencies({
+    manifest,
+    workspaces,
+  });
+  const sortedExpectedWorkspaces = [...expectedPackageNames]
+    .sort((a, b) => a.localeCompare(b))
```
Mrtenz, line 45: Does our target allow `toSorted`?

```suggestion
    .toSorted((a, b) => a.localeCompare(b))
```

scripts/lint-tsconfigs/lint-tsconfigs.mts
```diff
@@ -0,0 +1,81 @@
...
+
+import { lintPackageTsconfigs } from './lint-package-tsconfigs.mjs';
+import { lintRootTsconfigs } from './lint-root-tsconfigs.mjs';
+import { readPackageManifest } from './utils.mjs';
+
+/**
+ * Packages whose tsconfig files are intentionally excluded from linting and
+ * fixing, e.g. because their references are managed by hand.
+ */
+const EXCLUDED_PACKAGE_NAMES = new Set(['@metamask/snap-account-service']);
```
Mrtenz: Why do we exclude this one?

scripts/lint-tsconfigs/lint-tsconfigs.mts
```diff
@@ -0,0 +1,81 @@
...
+
+// Run this script!
+main().catch((error: unknown) => {
+  console.error(error);
+  process.exitCode = 1;
+});
```
Mrtenz: Since this is ESM, we can just await it.

```suggestion
await main();
```

scripts/lint-tsconfigs/utils.mts
```diff
@@ -0,0 +1,598 @@
...
+  currentWorkspaceRoot: string;
+}): Promise<TsconfigLintReport> {
+  const actualPackageNames = await getReferencedPackageNamesFromTsconfigFile({
+    tsconfig,
+    workspaces,
+  });
+
+  const missingReferencePaths = [...expectedPackageNames]
+    .filter((name) => !actualPackageNames.has(name))
+    .sort()
```
Mrtenz, line 286:
```suggestion
    .toSorted()
```

scripts/lint-tsconfigs/utils.mts
```diff
@@ -0,0 +1,598 @@
...
+        workspace,
+        tsconfig,
+        repoRoot,
+        currentWorkspaceRoot,
+      });
+    });
+
+  const extraReferencePaths = [...actualPackageNames]
+    .filter((name) => !expectedPackageNames.has(name))
+    .sort()
```
Mrtenz, line 302:
```suggestion
    .toSorted()
```

scripts/lint-tsconfigs/utils.mts
```diff
@@ -0,0 +1,598 @@
...
+ * @returns The normalized path.
+ */
+function normalizeReferencePath(refPath: string): string {
+  return refPath.replace(/\/tsconfig\.json$/u, '');
+}
+
+/**
+ * Produces an updated `references` list within a tsconfig file that preserves
+ * the order of existing references, removes extras, and adds missing ones in
+ * the order they appear in `newReferences`.
```
Mrtenz, line 382: Do we need to preserve order of existing references? I think it would be cleaner if we just alphabetically sort everything.

.oxfmtrc.json
Mrtenz: We should probably just enable this (in a separate PR). We did it previously with Prettier, not sure why we didn't enable this for Oxfmt.

tsconfig.json
Mrtenz: IIRC `ts-bridge` resolves these either way if they are specified as references in other packages. Makes sense to add them here though.

Mrtenz (changes requested):

scripts/lint-tsconfigs/lint-tsconfigs.mts
mcmire: It appears that `@metamask/snap-account-service` and `@metamask/account-tree-controller` have a circular dependency on each other. The Accounts team fixed this problem by copying types from `account-tree-controller` and leaving `@metamask/account-tree-controller` out from `tsconfig.json` / `tsconfig.build.json`: https://github.com/MetaMask/core/blob/60bb56c7726937935c94744d9b478adb8d9cbfc3/packages/snap-account-service/tsconfig.json.

Curiously, `snap-account-service`'s `package.json` still includes `account-tree-controller` in the list of `dependencies`. I don't think this is necessary. I have reached out to the Accounts team to see if I can remove this. If we can do this then we don't this exception.

scripts/lint-tsconfigs/lint-package-tsconfigs.mts
mcmire: Good callout. These are scripts, so it really depends on the Node version. It looks like `toSorted` was added in ES2023, which is supported at least in the latest version of Node 20.x: https://node.green/#ES2023. So I think we should be good to use `toSorted` here.

scripts/lint-tsconfigs/utils.mts
mcmire: I thought about that, but we haven't been enforcing sorting, so every single package's tsconfig files would likely have to be updated. I figured it would be more difficult to understand the diffs in this PR if I did this. I can turn this on in this PR, but what do you think if I did it in another PR?

scripts/lint-tsconfigs/lint-package-tsconfigs.mts
mcmire: Actually... this would require changing our `lib` setting (we use `ES2020` right now). We could change it to `ES2023` across the board, but I tried this and even though our `target` is still `ES2020` it doesn't seem that `toSorted` gets compiled down to `sort`. Plus as Cursor points out, technically we still support Node 18 in `node.engines`. So... maybe this is not a good idea. I've reverted these changes in [16d5dc1](https://github.com/MetaMask/core/pull/8384/commits/16d5dc1019122e7cf4c6f34032054a8b99dccec2).

scripts/lint-tsconfigs/utils.mts
mcmire: Initially accepted, but reverted in [16d5dc1](https://github.com/MetaMask/core/pull/8384/commits/16d5dc1019122e7cf4c6f34032054a8b99dccec2).

scripts/lint-tsconfigs/utils.mts
mcmire: Initially accepted, but reverted in [16d5dc1](https://github.com/MetaMask/core/pull/8384/commits/16d5dc1019122e7cf4c6f34032054a8b99dccec2).

.oxfmtrc.json
mcmire: Extracted here: https://github.com/MetaMask/core/pull/9492

scripts/lint-tsconfigs/lint-tsconfigs.mts
mcmire: I've fixed `snap-account-service` and have dropped the exclusion of this package from this PR: [13fb9ef](https://github.com/MetaMask/core/pull/8384/commits/13fb9efd9f973a89cc2ccecc5f798caad02f240e)

scripts/lint-tsconfigs/utils.mts
Mrtenz: Doing it in a separate PR makes sense to me.

Mrtenz (dismissed):

Mrtenz (approved):

PR #8388, opened 2026-04-07
fix: Fix hardware wallet MMPay on EIP-7702 chains by gating 7702 paths on account keyring capability

OGPoyraz:
## Explanation

Fix hardware wallet mUSD conversion on EIP-7702 chains by gating 7702 paths on account keyring capability, and fix batch transaction signing to prevent remaining hardware wallet prompts after a rejection.

#### `@metamask/transaction-controller`

- **BREAKING:** Add `KeyringControllerGetStateAction` to `AllowedActions` — clients must add `KeyringController:getState` to the TransactionController messenger's allowed actions
- Add `doesAccountSupportEIP7702` in `eip7702.ts` — checks keyring type via `KeyringController:getState`, returns `true` only for HD Key Tree and Simple Key Pair keyrings, falls back to `true` when keyring is unresolved
- Gate `addTransactionBatch` 7702 path on `doesAccountSupportEIP7702` — hardware wallet accounts fall back to sequential/hook batch flow instead of attempting EIP-7702 batch signing
- Sign batch transactions sequentially — each transaction waits for the previous signing to complete before calling `addTransaction` for the next, preventing remaining Ledger prompts from appearing after a rejection
- Add `CollectPublishHook.waitForSignedCount(count)` to support sequential signing gating in the batch loop
- Abort remaining in-flight transaction signings on batch failure via `abortTransactionSigning`

#### `@metamask/transaction-pay-controller`

- **BREAKING:** Add `KeyringControllerGetStateAction` to `AllowedActions` — clients must add `KeyringController:getState` to the TransactionPayController messenger's allowed actions
- Add `accountSupports7702` utility in `utils/7702.ts` and `KEYRING_TYPES_SUPPORTING_7702` constant in `types.ts`
- Thread `accountSupports7702` flag through `PayStrategyGetQuotesRequest` and `PayStrategyExecuteRequest` types
- Gate relay quote `useExecute` flag on `accountSupports7702` — ensures the relay API receives non-7702 requests for hardware wallets and returns quotes with proper individual gas limits
- Pass `accountSupports7702` from `TransactionPayPublishHook` into strategy `execute()` calls
- Pass `accountSupports7702` from fiat-quotes into relay-quotes
- Re-estimate gas individually in `estimateQuoteGasLimitsBatch` when batch simulation returns a combined 7702 gas limit but account doesn't support 7702

## References

- Fixes CONF-1151

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches transaction batching, signing flow control, and gas/quote estimation, plus introduces new messenger permissions; regressions could impact batch submission behavior especially across different keyring types.
> 
> **Overview**
> Fixes EIP-7702 batching/quote behavior for hardware-wallet accounts by **gating all 7702 paths on keyring capability**. Both `TransactionController` and `TransactionPayController` now require `KeyringController:getState` in messenger `AllowedActions`, and new helpers (`doesAccountSupportEIP7702` / `accountSupports7702`) treat only `HD Key Tree` and `Simple Key Pair` accounts as 7702-capable.
> 
> `addTransactionBatch` and `estimateGasBatch` now skip the EIP-7702 path when the account can’t sign 7702 authorizations, falling back to per-transaction/sequential behavior. Batch signing via hook is also made **sequential** by waiting for each transaction to reach a signed/submitted state before proceeding, reducing “extra” hardware wallet prompts after a rejection.
> 
> MMPay threads an `accountSupports7702` flag through strategy request/execute types and uses it to disable Relay “execute” (7702) request shaping and to produce non-7702 gas limit outputs when needed; tests and changelogs are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fa203c20fb6901c8373dc67babf011638ae1bb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


OGPoyraz:
@metamaskbot publish-preview

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1827,10 +1828,14 @@ export class KeyringController<
...
-   * Constructor helper for registering this controller's messeger
-   * actions.
-   */
+  async accountSupports7702(account: string): Promise<boolean> {
+    const keyringType = await this.getAccountKeyringType(account);
+    return (
+      keyringType === (KeyringTypes.hd as string) ||
+      keyringType === (KeyringTypes.simple as string)
+    );
+  }
```
gantunesr: Is there any reason we need this method to be part of the `KeyringController`?

OGPoyraz: Good question @gantunesr 

~~I haven't push my local changes into [mobile PR](https://github.com/MetaMask/metamask-mobile/pull/28330/) yet, but I am aiming to use this functionality exactly here: https://github.com/MetaMask/metamask-mobile/blob/9cae43a59fc69642b6329413a8d6ca79547ff274/app/components/Views/confirmations/hooks/transactions/useTransactionConfirm.ts#L50~~

~~Instead of determining `isHardwareWallet` in mobile and branching MMPay functionality depending on `isHardwareWallet`, I will create a hook in client directly gets if account supports 7702 and MMPay will depend on this.~~

~~Hopefully in near future, once we are ready enabling 7702 for hardware wallets, we will basically add keyring types here and MMPay will adapt, this is the end goal.~~

Removing my comment above, I will leave it as is. This is basically just used in `TransactionPayController` instead of adding callback to this controller. They are producing the same results with `isHardwareAccount` anyway.

- Core: uses `KeyringController:accountSupports7702` messenger action
- Mobile UI/hooks layer: uses `isHardwareAccount()` which has the same effective result for all current keyring types 

Depending on more feedback I can remove this and add `isHardwareAccount` callback in `TPC`.

packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/ramps-controller` from `^13.0.0` to `^13.1.0` ([#8407](https://github.com/MetaMask/core/pull/8407))
 
+### Fixed
+
+- Fix mUSD conversion for hardware wallets on EIP-7702 chains by gating relay and Across 7702 paths on `KeyringController:accountSupports7702` ([#8388](https://github.com/MetaMask/core/pull/8388))
```
matthewwalsh0: Could this ultimately be a client validation? So we extend the original client block to check the specific keyring supports 7702 otherwise they get a standard alert and can't continue meaning they can't trigger any of this logic?

So before it was all hardware wallets blocked, now it would just be all unless `TransactionType.musdConversion` and specific keyrings?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/CHANGELOG.md
OGPoyraz: So HW wallets are already blocked from the standard MMPay + 7702 gasless flow at the client level. The gap this PR fixes is different.

**The scenario**

HW wallet user does a mUSD conversion (or other same-chain operation) on a 7702-capable chain. The client lets them through because it's a valid MMPay transaction - it's not a 7702 flow. But internally, Relay/Across strategies see a 7702-enabled chain and naturally try to use 7702 paths (batch gas estimation, authorization lists, `originGasOverhead`). That fails because the HW wallet can't sign the authorization.

**Why a client block doesn't fully solve this**

The fix isn't "block the transaction", it's "allow the transaction but execute it differently." The strategies need to:
- Skip `originGasOverhead` in relay quotes
- Submit individually via `addTransaction` instead of `addTransactionBatch` (I might try to use `addTransactionBatch` with different config)
- Re-estimate gas per-transaction instead of using combined batch estimates
- Clear `batchTransactions:[]` so `TransactionController` doesn't try to batch

A client-side alert can only say yes or no. But it can't change how the strategy executes internally.

**The TransactionConfig flag alternative** 

We could add a `use7702?: boolean` flag to `TransactionConfig` (alongside `isMaxAmount`, `isPostQuote`, etc.) and have the client set it to `false` for HW wallets. But the conditional branches in the strategies stay identical - we'd just replace `messenger.call('KeyringController:accountSupports7702', from)` with `config.use7702`. Same logic, different call site. (this might be an answer to @gantunesr 's question above)

Having core check the keyring directly means a client bug (missed alert, wrong feature flag, new transaction type added later) can't accidentally send a 7702 authorization to a Ledger. Given that mobile have divergent gating patterns (different alert hooks, different publish hook order, different feature flag names), this feels worth keeping.

That said, if we want to decouple core from `KeyringController` and make this a client responsibility, the `TransactionConfig` flag approach is clean and extensible. Happy to go that route if preferred, but the strategy branching logic would remain the same.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -490,6 +493,7 @@ export type AllowedActions =
   | AccountsControllerGetSelectedAccountAction
   | AccountsControllerGetStateAction
   | ApprovalControllerAddRequestAction
+  | KeyringControllerAccountSupports7702Action
```
matthewwalsh0: Why are we adding this to the `TransactionController` if not used?

packages/transaction-pay-controller/src/strategy/across/across-quotes.ts
```diff
@@ -413,7 +413,43 @@ async function calculateSourceNetworkCost(
...
+
+  const accountSupports7702 = await messenger.call(
+    'KeyringController:accountSupports7702',
+    from,
+  );
+
+  // If the chain returned a combined 7702 gas limit but the account cannot sign
+  // EIP-7702 authorizations (e.g. hardware wallet), re-estimate each transaction
+  // individually so the submit path receives per-transaction gas limits.
+  if (gasEstimates.is7702 && !accountSupports7702) {
```
matthewwalsh0: Given the purpose of `estimateQuoteGasLimits` is to encapsulate the gas complexities, could we move this check into there so it can't return `is7702` to begin with is the account isn't compatible?

packages/transaction-pay-controller/src/strategy/across/across-submit.ts
```diff
@@ -196,17 +203,19 @@ async function submitTransactions(
...
-        'TransactionController:addTransaction',
-        transactions[0].params,
-        {
-          networkClientId,
-          origin: ORIGIN_METAMASK,
-          requireApproval: false,
-          type: transactions[0].type,
-        },
-      );
+    if (transactions.length === 1 || !accountSupports7702) {
```
matthewwalsh0: `addTransactionBatch` isn't exclusively for 7702, so ideally should the `TransactionController` do the keyring check before falling back to the STX or sequential hook?

So it's abstracted by any callers, they can still just say do this batch the best way you can?

packages/transaction-pay-controller/src/strategy/across/across-submit.ts
```diff
@@ -196,17 +203,19 @@ async function submitTransactions(
...
+      for (const { params, type } of transactions) {
```
matthewwalsh0: To clarify, `addTransactionBatch` already has an automatic fallback to sequential if not disabled via `disableSequential`.

So ideally `is7702` is never set in this scenario as the central gas util (we also use for Across) handles this check?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -221,7 +221,13 @@ async function getSingleQuote(
     // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
     const useExactInput = isMaxAmount || request.isPostQuote;
 
+    const accountSupports7702 = await messenger.call(
```
matthewwalsh0: Also here, if we've already done the gas limit check to see if 7702 is supported, could we avoid the repeated check here?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -536,24 +543,35 @@ async function submitViaTransactionController(
...
-        gasFeeToken,
-        networkClientId,
-        origin: ORIGIN_METAMASK,
-        requireApproval: false,
-        type: getRelayDepositType(transaction.type),
-      },
-    );
+  const results: { result: Promise<string> }[] = [];
+
+  if (allParams.length === 1 || !accountSupports7702) {
```
matthewwalsh0: Same comment as in Across, this could be encapsulated inside `addTransactionBatch` to remove any new logic here?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/CHANGELOG.md
matthewwalsh0: Thanks Goktug, I was indeed thinking this was gating the full functionality, rather than specific 7702 features 👍

packages/keyring-controller/src/KeyringController.ts
matthewwalsh0: I think I agree, that maybe it is safer for this not to be a `KeyringController` decision.

We can retrieve the keyring via the existing state and `KeyringController:getState`, but then there could be other nuances so even if a keyring supports 7702, we don't want to support it in MetaMask Pay.

Also decouples this from other controllers while we're developing and testing?

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: The reason of this addition was that the action is used on the transaction controller publish hooks hence the addition. But now I will add this as an individual callback to `TransactionPayController` config, so this will be removed.

packages/keyring-controller/src/KeyringController.ts
OGPoyraz: My goal was to centralising the account 7702 support decision in `KeyringController`. As per other comments, I will remove this and add it as `TransactionPayController` hook.

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: Removed, hence resolving this

packages/transaction-pay-controller/src/strategy/across/across-quotes.ts
OGPoyraz: Done, `estimateQuoteGasLimits` now accepts an `accountSupports7702` and handles the re-estimation internally - it never returns `is7702: true` when the account doesn't support it.

packages/transaction-pay-controller/src/strategy/across/across-submit.ts
OGPoyraz: Done. Removed the manual individual submission loop. Now always uses `addTransactionBatch` which falls back to sequential when 7702 isn't available.

packages/transaction-pay-controller/src/strategy/across/across-submit.ts
OGPoyraz: Same here, using `addTransactionBatch` now

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Fixed, `accountSupports7702` is now resolved once and passed through the request object. The gas util handles it centrally, and relay-quotes just reads it from the request.

packages/transaction-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add `KeyringControllerGetStateAction` to `AllowedActions` to support publish hooks that need to check keyring type for EIP-7702 compatibility ([#8388](https://github.com/MetaMask/core/pull/8388))
```
matthewwalsh0, line 12: Is this breaking since it requires new client messenger permissions?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -591,13 +607,13 @@ async function submitViaTransactionController(
     await messenger.call('TransactionController:addTransactionBatch', {
       from,
       disable7702: !gasLimit7702,
-      disableHook: Boolean(gasLimit7702),
+      disableHook: !gasLimit7702,
```
matthewwalsh0, line 610: We can still use STX for batch regardless of 7702 support, so shouldn't need this change?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -542,6 +558,8 @@ async function submitViaTransactionController(
   const { metamask } = quote.original;
   const { gasLimits } = metamask;
 
+  const is7702 = metamask.is7702 && accountSupports7702;
```
matthewwalsh0, line 561: My meaning before, was do we not need to move this to the `TransactionController` inside `addTransactionBatch` as this limitation will apply to all transaction batches, not just MM Pay?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -532,7 +546,9 @@ async function submitViaTransactionController(
     quote.original.details.currencyOut.currency.chainId;
 
   const authorizationList: AuthorizationList | undefined =
-    isSameChain && quote.original.request.authorizationList?.length
+    accountSupports7702 &&
+    isSameChain &&
+    quote.original.request.authorizationList?.length
```
matthewwalsh0, line 551: We should only get an `authorizationList` in the request if we tried to included transactions / intent, so we shouldn't need this check here, but instead to check in `processTransactions` when building the quote and throw if the keyring doesn't support intents but transactions are needed?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -453,6 +454,7 @@ async function normalizeQuote(
     messenger,
     request,
     fullRequest.transaction,
+    fullRequest.accountSupports7702,
```
matthewwalsh0, line 457: Minor, could destructure these above?

packages/transaction-pay-controller/src/strategy/across/across-submit.ts
```diff
@@ -215,7 +224,7 @@ async function submitTransactions(
 
       await messenger.call('TransactionController:addTransactionBatch', {
         disable7702: !gasLimit7702,
-        disableHook: Boolean(gasLimit7702),
+        disableHook: !gasLimit7702,
```
matthewwalsh0, line 227: Same as mentioned in Relay.

packages/transaction-pay-controller/src/helpers/QuoteRefresher.ts
```diff
@@ -30,19 +31,24 @@ export class QuoteRefresher {
 
   readonly #updateTransactionData: UpdateTransactionDataCallback;
 
+  readonly #accountSupports7702: AccountSupports7702Callback;
```
matthewwalsh0, line 34: Rather than passing this around as a callback, our standard architecture in this controller is to wrap in a util that requires the `messenger` as an argument, so it it's still encapsulated, but easier to invoke as we have a `messenger` reference almost everywhere already.

packages/transaction-pay-controller/src/helpers/TransactionPayPublishHook.ts
```diff
@@ -81,8 +82,18 @@ export class TransactionPayPublishHook {
     );
 
     const strategy = getStrategyByName(quotes[0].strategy);
+    const from = transactionMeta.txParams.from as Hex;
+
+    const { keyrings } = this.#messenger.call('KeyringController:getState');
```
matthewwalsh0, line 87: No need to duplicate here if we move this to a util as mentioned in other comment.

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -260,6 +263,17 @@ export class TransactionPayController extends BaseController<
     }
   }
 
+  async #accountSupports7702(account: string): Promise<boolean> {
```
matthewwalsh0, line 266: As mentioned, this could be in a util such as `utils/7702.ts`?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -491,6 +494,7 @@ export type AllowedActions =
   | AccountsControllerGetSelectedAccountAction
   | AccountsControllerGetStateAction
   | ApprovalControllerAddRequestAction
+  | KeyringControllerGetStateAction
```
matthewwalsh0, line 497: This still shouldn't be needed here as we're only using it in the hook which uses the client init messenger, rather than the controller messenger.

But I also suggested moving keyring check into `addTransactionBatch` so would need it in that case.

packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/assets-controllers` from `^104.0.0` to `^104.1.0` ([#8509](https://github.com/MetaMask/core/pull/8509))
 
+### Fixed
+
+- Fix mUSD conversion for hardware wallets on EIP-7702 chains by gating relay and Across 7702 paths on the account keyring type via `KeyringController:getState` ([#8388](https://github.com/MetaMask/core/pull/8388))
```
matthewwalsh0, line 16: Also breaking due to messenger permission in client?

packages/transaction-pay-controller/src/utils/quote-gas.ts
```diff
@@ -53,8 +55,40 @@ export async function estimateQuoteGasLimits({
   const useBatch = transactions.length > 1;
 
   if (useBatch) {
+    const result = await estimateQuoteGasLimitsBatch(transactions, messenger);
+
+    // If the batch returned a combined 7702 gas limit but the account cannot
+    // sign EIP-7702 authorizations (e.g. hardware wallet), re-estimate each
+    // transaction individually so callers receive per-transaction gas limits.
+    if (result.is7702 && !accountSupports7702) {
```
matthewwalsh0, line 63: Wouldn't this go in the above `estimateQuoteGasLimitsBatch` function?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -123,7 +129,7 @@ export async function updateQuotes(
     log('Calculated totals', { transactionId, totals });
 
     syncTransaction({
-      batchTransactions,
+      batchTransactions: supports7702 ? batchTransactions : undefined,
```
matthewwalsh0, line 132: Not needed as batch transactions don't have to be 7702.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-submit.test.ts
```diff
@@ -968,7 +968,6 @@ describe('Relay Submit Utils', () => {
 
         expect(addTransactionBatchMock).toHaveBeenCalledWith(
           expect.objectContaining({
-            disable7702: true,
```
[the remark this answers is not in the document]
OGPoyraz: The primary test (line 460) already asserts `disable7702` and `overwriteUpgrade` with an exact match.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -538,23 +547,33 @@ async function addTransactionBatchWithHook(
...
+        signingComplete.catch(reject);
+      });
+
+      await Promise.race([
+        collectHook.waitForSignedCount(index),
+        signingFailure,
+      ]);
     }
 
     const { signedTransactions } = await collectHook.ready();
```
OGPoyraz, line 642: After each `processTransactionWithHook`, the loop waits for that transaction's signing to complete (via `collectHook.waitForSignedCount`) before calling `addTransaction` for the next one. If signing fails (e.g. user rejects on Ledger), `signingFailure` wins the race and the loop throws - so the next transaction never reaches the hardware wallet.

This is explicitly fixing this case: https://consensys.slack.com/archives/C09T0L3SHCY/p1776671782000809

packages/transaction-controller/src/hooks/CollectPublishHook.ts
```diff
@@ -46,6 +48,20 @@ export class CollectPublishHook {
...
+   */
+  waitForSignedCount(count: number): Promise<void> {
+    if (this.#results.length >= count) {
+      return Promise.resolve();
+    }
+
+    return new Promise((resolve) => {
+      this.#signedCountWaiters.push({ count, resolve });
+    });
+  }
```
OGPoyraz: `waitForSignedCount` enables sequential signing in `addTransactionBatchWithHook`. Each time `#hook` is called (transaction finishes signing and reaches the publish stage), pending waiters whose threshold is met get resolved, allowing the batch loop to proceed to the next transaction.

packages/transaction-pay-controller/src/utils/7702.ts
```diff
@@ -0,0 +1,28 @@
...
+): boolean {
+  const { keyrings } = messenger.call('KeyringController:getState');
+  const keyring = keyrings.find((k) =>
+    k.accounts.some((a) => a.toLowerCase() === account.toLowerCase()),
+  );
+
+  return keyring
+    ? (KEYRING_TYPES_SUPPORTING_7702 as string[]).includes(keyring.type)
+    : true;
+}
```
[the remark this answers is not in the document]
OGPoyraz: This is expected.

packages/transaction-controller/CHANGELOG.md
OGPoyraz: Marked as BC now

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Reverted this

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Now the gate is part of TC

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Reverted this

packages/transaction-pay-controller/src/helpers/QuoteRefresher.ts
OGPoyraz: Put it into a helper now

packages/transaction-controller/src/TransactionController.ts
OGPoyraz: Exactly, leaving it now

packages/transaction-pay-controller/src/utils/quote-gas.ts
OGPoyraz: Fixed

packages/transaction-controller/src/utils/eip7702.ts
```diff
@@ -40,6 +40,32 @@ const ERC7579_EXEC_TYPE_TRY = '01';
...
+ * @param messenger - Controller messenger.
+ * @param account - The account address to check.
+ * @returns Whether the account supports EIP-7702.
+ */
+export function doesAccountSupportEIP7702(
+  messenger: TransactionControllerMessenger,
+  account: string,
+): boolean {
+  const { keyrings } = messenger.call('KeyringController:getState');
+  const keyring = keyrings.find((k: { type: string; accounts: string[] }) =>
```
matthewwalsh0: Minor, could we `return` this with a `some` instead?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -706,7 +743,7 @@ async function processTransactionWithHook(
     });
   }
 
-  const { transactionMeta } = await addTransaction(
+  const { transactionMeta, result: signingComplete } = await addTransaction(
```
matthewwalsh0: Is the goal here just to stop further signing if one of them is rejected at the signing stage? Meaning we ensure the whole thing (including any parent transaction via pay controller) fails?

If so, could we massively simplify by just awaiting the signing here so we ensure each one doesn't start until the previous is signed? Then no `Promise.race` nor changes to `CollectPublishHook` nor any need for `abortTransactionSigning`?

The only downside is probably a few quick gas requests for limits and fees that are no longer running in parallel, but the typical approval and swap case should always have gas limits anyway, and these are typically background processes anyway (usually 7702) so little impact.

We can't just `await result` since it won't resolve until the whole batch does.

But we could just make a small local function like `waitForSigned(transactionId)` that listens to the `transactionStatusUpdated` event and resolves if status is `signed` or throws if `failed`?

packages/transaction-pay-controller/src/utils/7702.ts
```diff
@@ -0,0 +1,28 @@
...
+  messenger: TransactionPayControllerMessenger,
+  account: string,
+): boolean {
+  const { keyrings } = messenger.call('KeyringController:getState');
+  const keyring = keyrings.find((k) =>
+    k.accounts.some((a) => a.toLowerCase() === account.toLowerCase()),
+  );
+
+  return keyring
+    ? (KEYRING_TYPES_SUPPORTING_7702 as string[]).includes(keyring.type)
```
matthewwalsh0: Minor, this is obviously duplicated in transaction controller, but it lets pay have its own criteria also. Could expose via action in transaction controller but not urgent, so better a future PR when needed.

packages/transaction-pay-controller/src/utils/quote-gas.ts
```diff
@@ -127,6 +137,38 @@ async function estimateQuoteGasLimitsBatch(
     0,
   );
   const is7702 = bufferedGasLimits.length === 1;
+
+  // If the batch returned a combined 7702 gas limit but the account cannot
+  // sign EIP-7702 authorizations (e.g. hardware wallet), re-estimate each
+  // transaction individually so callers never see is7702=true.
+  const supports7702 = accountSupports7702(messenger, firstTransaction.from);
```
matthewwalsh0: Is this all redundant if we check this inside `estimateGasBatch` in the transaction controller?

Since this too will be a limitation everywhere as with batch add?

We already check `isAtomicBatchSupported` for example, so would just check this too.

matthewwalsh0 (changes requested):

packages/transaction-controller/src/utils/batch.ts
OGPoyraz: > Is the goal here just to stop further signing if one of them is rejected at the signing stage?

Yes, the goal is to ensure transactions are signed sequentially so that hardware wallet prompts appear one at a time, and if a signing is rejected, the remaining transactions in the batch don't proceed (and remaining Ledger prompts don't appear).

> Meaning we ensure the whole thing (including any parent transaction via pay controller) fails?

That's correct. If any signing fails, the error propagates out of the for loop into the catch block, which calls `abortTransactionSigning` for any remaining transactions and then calls `collectHook.error()` + `resultCallbacks?.error()`, failing the entire batch including the parent pay controller flow.

> If so, could we massively simplify by just awaiting the signing here so we ensure each one doesn't start until the previous is signed? Then no Promise.race nor changes to CollectPublishHook nor any need for abortTransactionSigning?

I explored this approach and ran into a fundamental timing issue.

When `addTransaction` is called with `requireApproval: false`, it returns immediately - the signing chain (`#processApproval` > `#approveTransaction` > `#signTransaction`) runs as a detached async operation via the result promise. A `waitForSigned` helper that subscribes to `transactionStatusUpdated` after `addTransaction` returns has a race condition: the signed event can fire before the subscription is set up.

In practice this breaks the hardware wallet flow. The "Confirm on Ledger" modal never appears because the loop moves on before the signing flow has properly started - the event listener misses the status transitions entirely.

We can't subscribe before `addTransaction` either, because the `transactionId` isn't known until it returns.

The current `CollectPublishHook.waitForSignedCount` approach avoids this because the hook is passed into `addTransaction` and called from within the signing flow - it can't miss the event. The `Promise.race` with `signingFailure` handles the rejection path.
 
I think the current approach is the simplest one that's actually correct given the async topology here.

packages/transaction-pay-controller/src/utils/7702.ts
OGPoyraz: I agree, this is a duplication for now since the pay controller uses its own messenger type. Happy to expose it via a `TransactionController` action in a follow-up PR.

packages/transaction-pay-controller/src/utils/quote-gas.ts
OGPoyraz: Good call - moved the account support check into `estimateGasBatch` in the transaction controller so it applies everywhere. When the account doesn't support 7702, `estimateGasBatch` now skips the 7702 path entirely and falls through to per-transaction estimation. Removed the redundant check and the individual re-estimation fallback from the pay controller.

packages/transaction-controller/src/utils/eip7702.ts
OGPoyraz: Switched to `some` here. Also changed the fallback to return `false` when the keyring isn't found, since we'd rather be restrictive by default and not allow 7702 for accounts we can't resolve. Applied the same change to the pay controller duplicate.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -452,6 +458,77 @@ async function addTransactionBatchWith7702(
...
+      return false;
+    };
+
+    try {
+      const initialTx = getTransaction(transactionId);
+
+      if (checkStatus(initialTx)) {
+        return;
+      }
+    } catch {
```
matthewwalsh0: Minor, it should always exist right as we're awaiting the initial `addTransaction` first?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -544,17 +620,26 @@ async function addTransactionBatchWithHook(
...
+          publishHook,
+          request,
+          txBatchMeta,
+          index,
+        );
 
       hookTransactions.push(hookTransaction);
       index += 1;
+
+      if (!alreadySigned) {
```
matthewwalsh0: Can we remove this and rely just on `waitForTransactionStatus` for maximum simplicity and smallest diff?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -629,9 +714,12 @@ async function processTransactionWithHook(
   request: AddTransactionBatchRequest,
   txBatchMeta: TransactionBatchMeta | undefined,
   index: number,
-): Promise<
-  Omit<PublishBatchHookTransaction, 'signedTx'> & { type?: TransactionType }
-> {
+): Promise<{
```
matthewwalsh0: My hope is we wouldn't have to change anything in this function, but just accept the `submitted` or `confirmed` state also, so we have minimum branches and edge cases?

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -544,17 +620,26 @@ async function addTransactionBatchWithHook(
...
+          index,
+        );
 
       hookTransactions.push(hookTransaction);
       index += 1;
+
+      if (!alreadySigned) {
+        await waitForTransactionStatus(
+          String(hookTransaction.id),
+          TransactionStatus.signed,
```
matthewwalsh0, line 634: Could this be an array including `submitted` and `confirmed` so it covers all the bases including the existing transaction path?

packages/transaction-pay-controller/src/utils/quote-gas.test.ts
```diff
@@ -42,6 +43,17 @@ describe('quote gas estimation', () => {
   beforeEach(() => {
     jest.resetAllMocks();
 
+    getKeyringControllerStateMock.mockReturnValue({
```
matthewwalsh0: Minor, I guess this could mock the transaction controller action only?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/utils/batch.ts
OGPoyraz: Removed as it's unnecessary

packages/transaction-controller/src/utils/batch.ts
OGPoyraz: Agreed, no more `alreadySigned` prop and check

packages/transaction-controller/src/utils/batch.ts
OGPoyraz: Done

packages/transaction-controller/src/utils/batch.ts
OGPoyraz: Added these statuses as well

packages/transaction-pay-controller/src/utils/quote-gas.test.ts
OGPoyraz: Removed the unnecessary mocks

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -452,6 +458,77 @@ async function addTransactionBatchWith7702(
...
+    };
+
+    messenger.subscribe(
+      'TransactionController:stateChange',
+      handler,
+      (state: TransactionControllerState) =>
+        state.transactions.find((tx) => tx.id === transactionId),
+    );
+  });
+}
```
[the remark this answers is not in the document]
OGPoyraz: This is expected, we don't want to apply any timeout as it may take time user to complete confirmation on hardware wallet

matthewwalsh0 (approved):

PR #8412, opened 2026-04-09
Fix: assets controller startup empty accounts

salimtb:
## Explanation

**Current state:** On returning users (those with persisted accounts), `AssetsController.#start()` was called at keyring unlock time before `AccountTreeController.init()` had completed building the account tree. At that point, `#selectedAccounts` returned an empty array, so no subscriptions were set up and no asset fetch was triggered. The controller would then sit idle because:

- `AccountTreeController:selectedAccountGroupChange` only fires when the selected group *changes* — on returning users the persisted group is unchanged, so this event is skipped.
- `AccountTreeController:accountTreeChange` is not published by `init()` — it only fires when accounts are added/removed *after* initialization.

This meant asset balances, metadata, and prices were never loaded until the user manually switched accounts or networks.

**Solution:** Two changes working together:

1. **Subscribe to `AccountTreeController:stateChange`** — the base-controller event that is guaranteed to fire when `init()` calls `this.update()` to build the tree. When this fires, `#updateActive()` is called, which calls `#start()` now that accounts are available.

2. **Make `#start()` idempotent** — it now returns early if accounts/chains are not yet available (so the initial unlock call is a safe no-op), and also returns early if subscriptions are already active (so multiple subsequent `stateChange` firings — e.g. from snap accounts being added — don't trigger redundant fetches).

This follows the same pattern used by `DeFiPositionsController`, which also gets accounts lazily from `AccountTreeController` rather than eagerly at unlock time.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches `AssetsController` startup/lifecycle logic and subscription gating, which could change when/ how often asset fetches run if the new idempotency checks miss an edge case.
> 
> **Overview**
> Fixes a startup race where `AssetsController` could remain idle for returning users by additionally reacting to `AccountTreeController:stateChange` and re-evaluating active tracking once the account tree is initialized.
> 
> Makes `#start()` idempotent by returning early when no accounts/chains are available and when subscriptions are already active, preventing duplicate subscriptions/fetches from repeated events; adds a regression test for the unlock-then-tree-init flow and documents the fix in the assets-controller changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5ad3c3928adf6e1dc4a1a0f607a108f94c0c753. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -875,7 +875,6 @@ export class AssetsController extends BaseController<
     // when init() calls this.update(). #start() is idempotent so
     // repeated fires are safe.
     this.messenger.subscribe('AccountTreeController:stateChange', () => {
-      console.log('AccountTreeController:stateChanged +++++++++++');
       this.#updateActive();
     });
 
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -1,6 +1,7 @@
 import type {
   AccountTreeControllerGetAccountsFromSelectedAccountGroupAction,
   AccountTreeControllerSelectedAccountGroupChangeEvent,
+  AccountTreeControllerStateChangeEvent,
 } from '@metamask/account-tree-controller';
 import { BaseController } from '@metamask/base-controller';
 import type {
@@ -282,6 +283,7 @@ type AllowedActions =
 type AllowedEvents =
   // AssetsController
   | AccountTreeControllerSelectedAccountGroupChangeEvent
+  | AccountTreeControllerStateChangeEvent
   | ClientControllerStateChangeEvent
   | KeyringControllerLockEvent
   | KeyringControllerUnlockEvent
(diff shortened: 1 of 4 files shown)


Prithpal-Sooriya (approved):

packages/assets-controller/src/AssetsController.ts
```diff
@@ -865,6 +867,17 @@ export class AssetsController extends BaseController<
...
 
+    // Catch the initial tree build. On returning users,
+    // `selectedAccountGroupChange` does NOT fire when the persisted group
+    // is unchanged, and `accountTreeChange` doesn't fire either (init()
+    // rebuilds from persisted accounts without publishing it).
+    // The base-controller `:stateChange` event is guaranteed to fire
+    // when init() calls this.update(). #start() is idempotent so
+    // repeated fires are safe.
+    this.messenger.subscribe('AccountTreeController:stateChange', () => {
+      this.#updateActive();
```
Prithpal-Sooriya, line 878: Can we use the state change selector (3rd param of the subscribe func)

Prithpal-Sooriya: Not a blocker, we'll receive a new event by accounts team which should remove the need for this state change event

salimtb: yes it's just a temp solution for now

PR #8413, opened 2026-04-09
feat: add chomp API service

Jwhiles:
- **feat: add chomp-api-service**
- **feat: add stubs for all API methods**

## Explanation
In this PR we create an API client to allow our client applications to use the chomp API. It fixes [this ticket](https://consensyssoftware.atlassian.net/jira/software/c/projects/MUSD/boards/4999?assignee=712020%3Adbb286ee-d4ae-45d1-b54f-01a69f0a80a5&quickFilter=6335&selectedIssue=MUSD-291)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces a new authenticated HTTP client/service that fetches JWT bearer tokens via messenger and performs write operations against an external API, so incorrect request/response handling could impact downstream flows but existing packages are largely untouched.
> 
> **Overview**
> Adds a new `@metamask/chomp-api-service` package implementing `ChompApiService` (a `BaseDataService`) to call the CHOMP API with JWT bearer auth sourced from `AuthenticationController:getBearerToken`.
> 
> The service exposes messenger-callable methods for address association, account upgrade create/get, delegation verification, intent create/list, withdrawal creation, and service-details lookup, with runtime response validation via `superstruct`, explicit status handling (e.g., `409`/`404`), and comprehensive `nock`-based tests.
> 
> Repo wiring is updated to include the new package in TypeScript project references, the monorepo README package list/graph, `CODEOWNERS`/`teams.json`, and `yarn.lock`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b15c9f6beaf6b870ed74acd784c79dca44a1a21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,433 @@
...
+
+      if (!response.ok && !acceptedStatuses.includes(response.status)) {
+        throw new HttpError(
+          response.status,
+          `POST ${path} failed with status '${response.status}'`,
+        );
+      }
+
+      return response;
+    });
```
mcmire: After I gave you the suggest to use the service policy manually like this, I came across another PR that someone else posted here: https://github.com/MetaMask/core/pull/8260/changes#diff-df87c072da30a25c3d5d0291f962f6b7c52da863a4d1e79a11752d89bc471c4cR204.

Basically it seems that we _can_ use `fetchQuery`, but pass `staleTime: 0` to effectively turn off caching. Can you try that and see how that works? So this would look something like:

``` suggestion
    return this.fetchQuery({
      queryKey: [
        // ... you might want to pass this in ...
      ],
      queryFn: async () => {
        const response = await fetch(new URL(path, this.#baseUrl), {
          method: 'POST',
          headers,
          body: JSON.stringify(body),
        });

        if (!response.ok && !acceptedStatuses.includes(response.status)) {
          throw new HttpError(
            response.status,
            `POST ${path} failed with status '${response.status}'`,
          );
        }

        return response;
      }),
      staleTime: 0,
    });
```

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,433 @@
...
+  }) {
+    super({
+      name: serviceName,
+      messenger,
+      queryClientConfig,
+      policyOptions,
+    });
+
+    this.#baseUrl = baseUrl;
+    this.#getAccessToken = getAccessToken;
```
mcmire: Where will this access token come from in the clients? Will this come from another controller or service? If so, then it would be better to use the messenger to retrieve this value instead of taking a callback.

packages/chomp-api-service/README.md
```diff
@@ -0,0 +1,15 @@
+# `@metamask/chom-api-service`
```
mcmire: Typo?

``` suggestion
# `@metamask/chomp-api-service`

File: packages/chomp-api-service/tsconfig.build.json
Comment (mcmire) [inline packages/chomp-api-service/tsconfig.build.json:1]: In addition to modifying these files you should also add `chomp-api-service` to the root `tsconfig.json` and `tsconfig.build.json` files to ensure that TypeScript knows about it (and its relationship to other packages) in development.

**Review by mcmire — COMMENTED** [review]
This looks pretty good so far! I just had a few comments.

File: packages/chomp-api-service/CHANGELOG.md [hunk 4]
```diff
@@ -0,0 +1,14 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- Add `ChompApiService` ([#8361](https://github.com/MetaMask/core/pull/8413))
```
MoMannn: Pr number here is different then the link

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,433 @@
...
+      type: enums(['cash-deposit', 'cash-withdrawal']),
+    }),
+    createdAt: string(),
+  }),
+);
+
+const IntentEntryArrayStruct = array(
+  type({
+    account: string(),
+    delegationHash: string(),
```
MoMannn: can we use hex validations here as they are defined in types?

packages/chomp-api-service/src/types.ts
```diff
@@ -0,0 +1,127 @@
...
+  profileId: string;
+  address: string;
+  status: string;
+};
+
+export type CreateUpgradeResponse = {
+  signerAddress: string;
+  status: string;
+  createdAt: string;
+};
```
MoMannn: Could this and `GetUpgradeResponse` be merged into a single type since they are the same?

packages/chomp-api-service/src/chomp-api-service.ts
Jwhiles: This is was included in [the spec](https://consensyssoftware.atlassian.net/wiki/spaces/mUSD/pages/401283973127/08.+Account+Upgrade+CHOMP) written by @MoMannn - but I'm open to using the messenger. Tadej, is there a reason that you called for this design in particular?

Jwhiles:
@metamaskbot publish-preview

Jwhiles:
@metamaskbot publish-preview

Jwhiles:
@metamaskbot publish-preview

Jwhiles:
@metamaskbot publish-preview

packages/chomp-api-service/src/chomp-api-service-method-action-types.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Fetches intents by address via GET /v1/intent/account/:address.
+ */
+export type ChompApiServiceGetIntentsByAddressAction = {
+  type: `ChompApiService:getIntentsByAddress`;
+  handler: ChompApiService['getIntentsByAddress'];
+};
+
+/**
+ * Creates a withdrawal for card spend flows.
```
MoMannn: Don't think this is specifically for card spend flow. Card spend should actually not be doing withdrawal at all since they can directly use vmUSD

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,571 @@
...
+  typeof serviceName,
+  ChompApiServiceActions | AllowedActions,
+  ChompApiServiceEvents | AllowedEvents
+>;
+
+// === RESPONSE VALIDATION ===
+
+const AssociateAddressResponseStruct = type({
+  profileId: string(),
+  address: string(),
```
MoMannn: should this also be StrictHexStruct? and other addresses

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,571 @@
...
+);
+
+const CreateWithdrawalResponseStruct = type({
+  success: literal(true),
+});
+
+const ServiceDetailsProtocolStruct = type({
+  supportedTokens: array(
+    type({
+      tokenAddress: string(),
```
MoMannn: same for addresses here: StrictHexStruct

packages/chomp-api-service/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  salt: Hex;
+  signature: Hex;
+};
+
+// === REQUEST TYPES ===
+
+export type AssociateAddressRequest = {
+  signature: string;
+  timestamp: string;
+  address: string;
```
MoMannn: Here we are also mixing hex and string. Some addresses are string some are hex

MoMannn (changes requested):

MoMannn (dismissed):

shane-t (dismissed):

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,564 @@
...
+   * Associates an address with a CHOMP profile.
+   *
+   * POST /v1/auth/address
+   *
+   * @param request - The association request containing signature, timestamp,
+   * and address.
+   * @returns The profile association result. Returns on both 201 and 409.
+   */
+  async associateAddress(
+    request: AssociateAddressRequest,
```
mcmire: Nit: Calling this a "request" sounds a bit misleading in the context of this class. Maybe `params` would be a better name?

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,564 @@
...
+        }
+
+        return response.json();
+      },
+    });
+
+    return create(
+      jsonResponse,
+      ServiceDetailsResponseStruct,
+    ) as ServiceDetailsResponse;
```
mcmire: Nit: Same as above — type assertion?

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,564 @@
...
+            response.status,
+            `Get intents request failed with status '${response.status}'`,
+          );
+        }
+
+        return response.json();
+      },
+    });
+
+    return create(jsonResponse, IntentEntryArrayStruct) as IntentEntry[];
```
mcmire: Nit: Same as above — is there a way to remove the type assertion?

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,564 @@
...
+            response.status,
+            `POST /v1/intent failed with status '${response.status}'`,
+          );
+        }
+
+        return response.json();
+      },
+    });
+
+    return create(
```
mcmire: Nit: Is there a way to avoid using a type assertion here? Maybe there is a way that `create` can return the right type already.

packages/chomp-api-service/src/chomp-api-service.ts
```diff
@@ -0,0 +1,564 @@
...
+    const jsonResponse = await this.fetchQuery({
+      queryKey: [`${this.name}:getUpgrade`, address],
+      queryFn: async () => {
+        const headers = await this.#authHeaders();
+        const response = await fetch(
+          new URL(`/v1/account-upgrade/${address}`, this.#baseUrl),
+          { headers },
+        );
+
+        if (response.status === 404) {
```
mcmire, line 360: Interesting. Why ignore 404 here?

mcmire (dismissed):
Left some comments below. All are minor (some are nitpicks). The way you've used BaseDataService seems good to me and everything seems in order here.

packages/chomp-api-service/src/chomp-api-service.ts
Jwhiles: The intention was to have the method return null if the user isn't upgrade - because this is an 'expected failure' rather than error.. However maybe this isn't the right place to do that - and we should just let the caller handle 404s?

mcmire: If this were an unexpected error we would definitely encourage you bubble it up to the caller. But if it's expected, it seems fine to handle as is.

mcmire (approved):
LGTM!

shane-t (approved):

PR #8415, opened 2026-04-09
feat(keyring-controller): persist vault when keyring state changes during unlock

danroc:
## Explanation

Currently, if a keyring runs a migration inside `deserialize()` during unlock, the updated in-memory state is never written back to the vault. On the next unlock the migration silently re-runs against the same stale data.

This PR fixes that by comparing each keyring's re-serialized state against the original data after `#restoreKeyring`. If any keyring's state differs — whether due to a migration or missing metadata — `hasChanged` is propagated up through `#restoreSerializedKeyrings` and `#unlockKeyrings`, and the existing vault upgrade path in `submitPassword`/`submitEncryptionKey` re-persists the vault.

As part of this change, a `hasStateChanged` helper function is introduced and reused in `#persistOrRollback`, removing the dependency on lodash `isEqual` for that comparison.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches the unlock/vault-update path to trigger re-encryption/persistence based on detected keyring state changes, which could affect login performance and vault write frequency if the change detection is wrong.
> 
> **Overview**
> During `submitPassword`/`submitEncryptionKey`, the controller now tracks whether any keyring’s serialized data changes during restore (including missing metadata) and, if so, re-runs the existing vault update flow to persist the upgraded state.
> 
> Change detection is implemented in `#restoreKeyring` by comparing pre- and post-deserialization serialized payloads and is propagated via `#restoreSerializedKeyrings`/`#unlockKeyrings` as `hasChanged`; `#persistOrRollback` also switches from lodash `isEqual` to a JSON-string comparison for session-state diffs. Adds a regression test for an in-place migrating keyring and updates the changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e7502642ba7c8f5a7cc2b674ff8bf88cd88869b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-controller/src/KeyringController.test.ts
+++ b/packages/keyring-controller/src/KeyringController.test.ts
@@ -2949,8 +2949,9 @@ describe('KeyringController', () => {
       const oldState = { version: 1, foo: 'bar' };
       const newState = { version: 2, foo: 'bar' };
 
-      // A keyring that migrates its own state during deserialization:
-      // after deserialize(oldState), serialize() returns newState.
+      // A keyring that migrates its own state in-place during deserialization: after
+      // deserialize(oldState), the original `data` object is mutated and serialize()
+      // returns the updated state.
       class MigratingKeyring {
         static type = 'Migrating Keyring';
 
@@ -2963,7 +2964,9 @@ describe('KeyringController', () => {
         }
 
         async deserialize(data: Record<string, unknown>): Promise<void> {
-          this.#state = data.version === 1 ? { ...data, version: 2 } : data;
+          // Mutate in-place to simulate a keyring that upgrades its own format.
+          data.version = 2;
+          this.#state = data;
         }
 
         async getAccounts(): Promise<string[]> {
@@ -3002,6 +3005,8 @@ describe('KeyringController', () => {
 
           await controller.submitPassword(password);
 
+          // Migration should have triggered a new vault update that we need to
+          // re-encrypt:
           expect(encryptWithKeySpy).toHaveBeenCalledWith(
             defaultCredentials,
             expect.arrayContaining([
--- a/packages/keyring-controller/src/KeyringController.test.ts
+++ b/packages/keyring-controller/src/KeyringController.test.ts
@@ -2945,6 +2945,76 @@ describe('KeyringController', () => {
       );
     });
 
+    it('should update the vault if the keyring state changes during deserialization', async () => {
+      const oldState = { version: 1, foo: 'bar' };
+      const newState = { version: 2, foo: 'bar' };
+
+      // A keyring that migrates its own state during deserialization:
+      // after deserialize(oldState), serialize() returns newState.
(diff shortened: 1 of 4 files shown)


packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -2945,6 +2945,76 @@ describe('KeyringController', () => {
...
+          keyringBuilders: [
+            keyringBuilderFactory(MigratingKeyring as unknown as KeyringClass),
+          ],
+        },
+        async ({ controller, encryptor }) => {
+          const encryptWithKeySpy = jest.spyOn(encryptor, 'encryptWithKey');
+
+          await controller.submitPassword(password);
+
+          expect(encryptWithKeySpy).toHaveBeenCalledWith(
```
ccharly, line 3010: WDYT about adding this small comment? This that's our only way to know a vault update got executed?

```suggestion
          // Migration should have triggered a new vault update that we need to re-encrypt:
          expect(encryptWithKeySpy).toHaveBeenCalledWith(
```

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2496,30 +2496,36 @@ export class KeyringController<
...
     this.#assertControllerMutexIsLocked();
 
     try {
       const { type, data, metadata: serializedMetadata } = serialized;
-      let newMetadata = false;
-      let metadata = serializedMetadata;
+
       const keyring = await this.#createKeyring(type, data);
+      const restoredData = await keyring.serialize();
+      let hasChanged = JSON.stringify(data) !== JSON.stringify(restoredData);
```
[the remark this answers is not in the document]
ccharly: Seems a bit unlikely to happen, but that's not entirely false!

Maybe we could use the same pattern than `oldState` and `newState`. Like:

```ts
      const originalData = JSON.stringify(data);
      const keyring = await this.#createKeyring(type, data);
      const restoredData = JSON.stringify(await keyring.serialize());
      let hasChanged = originalData !== restoredData;
```

tsconfig.packages.json
```diff
@@ -4,6 +4,7 @@
    */
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
+    "types": ["jest"],
```
ccharly: IDK if we should push this change? I know this address some IDE problems, but I wonder if there are other more "global" ways of addressing this?

danroc: Yeah, I wasn't planning to push it.

packages/keyring-controller/src/KeyringController.ts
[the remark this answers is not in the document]
danroc: It's a good catch indeed

ccharly (approved):
LGTM!

PR #8416, opened 2026-04-09
feat(keyring-controller): add `withController` for atomic operations over multiple keyrings

ccharly:
## Explanation

Today there's no way to make multiple operations in an "atomic" (read transactional) way.

A good example of this is if you want to use a keyring using `withKeyring` that's not existing yet (I'm omitting the `createIfMissing` variants, as we wanted to move away from this pattern).

To do this in a safe way, you usually have to use your own lock to make sure you can get-or-create the keyring and prevent concurrent keyring creations.

This new `withController` is based on the `withKeyring` but with an access to a "restricted" state and methods of the controller. This way, you can interact with multiple keyring at once while being guarded (to prevent race-conditions) by the controller's global lock.

The former problem can then be written that way now:

```ts
const account = await keyringController.withController(async (controller) => {
  // Here, `controller.keyrings` is a "view" on the existing keyrings (instances), only valid
  // for this block.
  let keyring: MyKeyring | undefined = controller.keyrings.find(isMyKeyring);
  if (!keyring) {
    const { keyring: myKeyring } = await controller.addNewKeyring({ type: 'My Keyring', data: { ... }});
    keyring = myKeyring;
  }
  
  const [account] = await keyring.createAccounts(...);
  return account;
});
```

This will also be used to write the migration from the existing `SnapKeyring` (1 for ALL Snaps) to multiple `SnapKeyring` (1 PER Snap) in a safe way like:

```ts
await keyringController.withController(async (controller) => {
  const accounts: Map<SnapId, KeyringAccount[]> = new Map();

  // Get existing Snap accounts from the single Snap keyring instance we have today.
  const keyring: SnapKeyring | undefined = controller.keyrings.find(isSnapKeyring);
  if (keyring) {
    for (const account of keyring.listAccounts()) {
      accounts[account.metadata.snap.id] ??= [];
      accounts[account.metadata.snap.id].push(account);
    }
  }
  
  // Re-create all new Snap keyrings, 1 per Snap.
  for (const [snapId, snapAccounts] of accounts.entries()) {
    await controller.addNewKeyring({ type: 'Snap keyring', data: snapAccounts });
  }
  
  // We can safely remove the existing Snap keyring now.
  await controller.removeKeyring(...);
});
```

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new transactional API that can create/remove keyrings and then persist/rollback state, which touches keyring lifecycle and vault persistence paths. Risk is moderate due to potential edge cases around staged mutations, keyring destruction, and primary keyring protection.
> 
> **Overview**
> Adds `KeyringController.withController` (and messenger action `KeyringController:withController`) to run **single-lock, atomic operations across multiple keyrings** via a `RestrictedController` that exposes a live read-only view plus staged `addNewKeyring`/`removeKeyring` mutations.
> 
> On success it commits staged keyring list changes, persists via existing persist/rollback flow, and destroys removed keyrings; on error it rolls back and destroys any newly created keyrings. The PR also blocks removal of the primary HD keyring (`CannotRemovePrimaryKeyring`), extends `KeyringEntry`/callback typings, updates mocks to include `destroy`, and adds comprehensive unit + messenger tests for the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ae24f643bfb15f9356ca49c8802e4557ce87478. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-controller/src/KeyringController-method-action-types.ts
+++ b/packages/keyring-controller/src/KeyringController-method-action-types.ts
@@ -272,24 +272,6 @@ export type KeyringControllerWithKeyringAction = {
 };
 
 /**
- * Execute an operation against all keyrings as a mutually exclusive atomic
- * operation. The operation receives a {@link RestrictedController} instance
- * that exposes a read-only live view of all keyrings as well as
- * `addNewKeyring` and `removeKeyring` methods to stage mutations.
- *
- * The method automatically persists changes at the end of the function
- * execution, or rolls back the changes if an error is thrown.
- *
- * @param operation - Function to execute with the restricted controller.
- * @returns Promise resolving to the result of the function execution.
- * @template CallbackResult - The type of the value resolved by the callback function.
- */
-export type KeyringControllerWithControllerAction = {
-  type: `KeyringController:withController`;
-  handler: KeyringController['withController'];
-};
-
-/**
  * Select a keyring and execute the given operation with the selected
  * keyring, **without** acquiring the controller's mutual exclusion lock.
  *
@@ -394,6 +376,24 @@ export type KeyringControllerWithKeyringV2UnsafeAction = {
 };
 
 /**
+ * Execute an operation against all keyrings as a mutually exclusive atomic
+ * operation. The operation receives a {@link RestrictedController} instance
+ * that exposes a read-only live view of all keyrings as well as
+ * `addNewKeyring` and `removeKeyring` methods to stage mutations.
+ *
+ * The method automatically persists changes at the end of the function
+ * execution, or rolls back the changes if an error is thrown.
+ *
+ * @param operation - Function to execute with the restricted controller.
+ * @returns Promise resolving to the result of the function execution.
+ * @template CallbackResult - The type of the value resolved by the callback function.
+ */
+export type KeyringControllerWithControllerAction = {
+  type: `KeyringController:withController`;
+  handler: KeyringController['withController'];
+};
+
+/**
  * Union of all KeyringController action types.
  */
 export type KeyringControllerMethodActions =
@@ -417,7 +417,7 @@ export type KeyringControllerMethodActions =
   | KeyringControllerPatchUserOperationAction
   | KeyringControllerSignUserOperationAction
   | KeyringControllerWithKeyringAction
-  | KeyringControllerWithControllerAction
   | KeyringControllerWithKeyringUnsafeAction
   | KeyringControllerWithKeyringV2Action
-  | KeyringControllerWithKeyringV2UnsafeAction;
+  | KeyringControllerWithKeyringV2UnsafeAction
+  | KeyringControllerWithControllerAction;
--- a/packages/keyring-controller/src/KeyringController-method-action-types.ts
+++ b/packages/keyring-controller/src/KeyringController-method-action-types.ts
@@ -272,6 +272,24 @@ export type KeyringControllerWithKeyringAction = {
 };
 
 /**
+ * Execute an operation against all keyrings as a mutually exclusive atomic
+ * operation. The operation receives a {@link RestrictedController} instance
+ * that exposes a read-only live view of all keyrings as well as
+ * `addNewKeyring` and `removeKeyring` methods to stage mutations.
+ *
+ * The method automatically persists changes at the end of the function
+ * execution, or rolls back the changes if an error is thrown.
+ *
+ * @param operation - Function to execute with the restricted controller.
+ * @returns Promise resolving to the result of the function execution.
+ * @template CallbackResult - The type of the value resolved by the callback function.
+ */
+export type KeyringControllerWithControllerAction = {
+  type: `KeyringController:withController`;
+  handler: KeyringController['withController'];
+};
+
+/**
  * Select a keyring and execute the given operation with the selected
  * keyring, **without** acquiring the controller's mutual exclusion lock.
  *
@@ -399,6 +417,7 @@ export type KeyringControllerMethodActions =
   | KeyringControllerPatchUserOperationAction
   | KeyringControllerSignUserOperationAction
   | KeyringControllerWithKeyringAction
+  | KeyringControllerWithControllerAction
   | KeyringControllerWithKeyringUnsafeAction
   | KeyringControllerWithKeyringV2Action
   | KeyringControllerWithKeyringV2UnsafeAction;
--- a/packages/keyring-controller/src/KeyringController.ts
+++ b/packages/keyring-controller/src/KeyringController.ts
@@ -250,6 +250,43 @@ type KeyringEntry = {
 };
 
 /**
+ * A restricted view of the {@link KeyringController} exposed to the callback
+ * passed to {@link KeyringController.withController}.
+ *
+ * It provides a read-only live view of all keyrings and the ability to stage
+ * keyring additions and removals atomically within a single transaction.
+ */
+export type RestrictedController = {
+  /**
+   * Read-only live view of all keyrings in the current transaction (original
+   * keyrings plus any added, minus any removed so far in this callback).
+   */
+  readonly keyrings: ReadonlyArray<{
+    keyring: EthKeyring;
+    metadata: KeyringMetadata;
+  }>;
+  /**
+   * Create a new keyring of the given type and stage it for commit. The new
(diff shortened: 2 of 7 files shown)


packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1852,6 +1866,107 @@ export class KeyringController<
...
+        },
+      };
+
+      const destroyKeyrings = async (
+        entries: Iterable<KeyringEntry>,
+      ): Promise<void> => {
+        await Promise.all(
+          [...entries].map(({ keyring }) => this.#destroyKeyring(keyring)),
+        );
+      };
```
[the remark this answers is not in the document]
ccharly: 5c291ea1428c2ead5440cbc1ae33c8a474d02ff7

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1852,6 +1866,112 @@ export class KeyringController<
...
+      this.#keyrings = restrictedEntries;
+
+      // As usual, we want to prevent returning direct references to keyring instances, so we check
+      // the result for any unsafe direct access before returning.
+      for (const { keyring, keyringV2 } of this.#keyrings) {
+        this.#assertNoUnsafeDirectKeyringAccess(result, keyring);
+        if (keyringV2) {
+          this.#assertNoUnsafeDirectKeyringAccess(result, keyringV2);
+        }
+      }
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/8416/commits/ba3dbf4440decfae52478f013a1f87f8c661cf4b

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1760,13 +1792,7 @@ export class KeyringController<
...
   >(
     selector: KeyringSelector<SelectedKeyring>,
-    operation: ({
-      keyring,
-      metadata,
-    }: {
-      keyring: SelectedKeyring;
-      metadata: KeyringMetadata;
-    }) => Promise<CallbackResult>,
+    operation: ({ keyring, metadata }: KeyringEntry) => Promise<CallbackResult>,
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/8416/commits/019cc603983166f2709c5a8341e4b74738628a03

packages/keyring-controller/src/KeyringController-method-action-types.ts
```diff
@@ -271,6 +271,24 @@ export type KeyringControllerWithKeyringAction = {
...
+ * execution, or rolls back the changes if an error is thrown.
+ *
+ * @param operation - Function to execute with the restricted controller.
+ * @returns Promise resolving to the result of the function execution.
+ * @template CallbackResult - The type of the value resolved by the callback function.
+ */
+export type KeyringControllerWithControllerAction = {
+  type: `KeyringController:withController`;
+  handler: KeyringController['withController'];
+};
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/8416/commits/491434bf9fbd9689e3406a80300e506926dcc5d0

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2069,6 +2089,117 @@ export class KeyringController<
...
+            this.#destroyKeyring(keyring, keyringV2),
+          ),
+        );
+      };
+
+      let result: CallbackResult;
+      try {
+        result = await operation(restrictedController);
+      } catch (error) {
+        await destroyKeyrings(createdEntries);
```
hmalik88, line 2174: Hmm I think we're overlooking the side-effects of if `addNewKeyring` is used in a failed operation. If data is passed into deserialize the snap would have created phantom accounts.

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2069,6 +2089,117 @@ export class KeyringController<
...
+
+      // As usual, we want to prevent returning direct references to keyring instances, so we check
+      // the result for any unsafe direct access before returning.
+      for (const { keyring, keyringV2 } of [
+        ...this.#keyrings,
+        // We also check for keyrings that got removed during the operation, since the result could
+        // still have references to them.
+        ...removedEntries,
+      ]) {
+        this.#assertNoUnsafeDirectKeyringAccess(result, keyring);
```
hmalik88, line 2193: This check can be bypassed if the consumer does something like this no? 🤔:

```ts
  let leaked: RestrictedController;

  await keyringController.withController(async (rc) => {
    leaked = rc;               // return value is void, so the direct-access check never fires
    return undefined;
  });
  // lock released here

  // Now, totally outside any lock:
  const entry = leaked.keyrings[0];          // getter still works
  const keyring = entry.keyring;              // raw live keyring instance
  await keyring.addAccounts(5);               // mutation without controller lock
  // Concurrent `withKeyring` or `signTransaction` on another path
  // could be reading this exact instance
```

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2069,6 +2089,117 @@ export class KeyringController<
...
+          const entry = await this.#createKeyring(type, opts);
+
+          restrictedEntries.push(entry);
+          createdEntries.add(entry);
+
+          return entry;
+        },
+
+        // Method to remove a keyring from the restricted entries.
+        removeKeyring: async (id: string) => {
```
hmalik88, line 2143: WDYT about adding a guard to make sure we're not removing the last HD keyring here similar to `removeAccount`? Otherwise a consumer can stage the removal, the callback resolves, `destroyKeyrings(removedEntries)` tears down the original HD keyring instance, then the vault commit (`#updateVault`) fails and `#withRollback` recovers state via `#restoreSerializedKeyrings`, which destroys everything in `this.#keyrings` and rebuilds new instances from the snapshot. You then have:

1. The error surfacing at commit time as `NoHdKeyring`, not at the `removeKeyring()` line that caused it which would be harder to debug.
2. Any external code holding a reference to the keyring instance is now pointing at a destroyed instance.

packages/keyring-controller/src/KeyringController.ts
ccharly: Yep that's correct. But that's now "new" 😅 I just re-used the pattern we have in place for `withKeyring*` (so you can also bypass those check with `withKeyring` exactly using your same pattern 😬)

packages/keyring-controller/src/KeyringController.ts
ccharly: Well 🤔 I think that should be ok given that we would call `destroy` on each keyrings in that case, so the keyring "could" detect that the account that he just created has to be removed?

We could improve `#destroyKeyring` and also call `removeAccount` on each accounts the keyring holds but.. That could be wasteful too! It's probably faster to call 1 good `destroy` rather than N `removeAccount` + 1 `destroy`.

So I think `destroy` should solve this issue no?

packages/keyring-controller/src/KeyringController.ts
ccharly: That sounds like a really good idea, I'll add this!

ccharly: https://github.com/MetaMask/core/pull/8416/commits/0ae24f643bfb15f9356ca49c8802e4557ce87478

packages/keyring-controller/src/KeyringController.ts
hmalik88: But the accounts would remain in the accounts controller still no?

ccharly: Nope. The whole call `operation(...)` + `destroyKeyrings` is executed within `persistOrRollback`.

And since the `AccountsController` is replicating the accounts after `KeyringController:stateChange` occurred, then the controller won't see those new keyrings at all (since they're going to be destroyed before we persist anything).

My comment about `destroy` and `removeAccount` was more about allowing the keyring to do a proper "teardown" in case of failure (and in case this teardown was having other side-effects)

hmalik88: Ok that addresses my concern, thanks!

hmalik88 (approved):

PR #8417, opened 2026-04-09
feat: Add wallet library

FrederikBolding:
## Explanation

Feature branch. WIP.

Know limitations:

- Does not run on Node 18. Currently skipped in CI. Would need polyfill to fix but we should probably deprecate Node 18 instead.

## Checklist


packages/wallet/src/persistence/index.ts
Gudahtt, line 1: Should this persistence API be in this package? I don't expect this would be used in either client

FrederikBolding: It shouldn't. We were planning to move that before attempting to merge this

PR #8420, opened 2026-04-09
fix(wallet): Fix builds and most lint issues

rekmarks:
Builds and tests pass. All lint issues are fixed except a handful that are deferred pending future changes.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to new controller dependencies and changes to initialization wiring/typing that could affect runtime messaging and controller lifecycle. Test behavior now depends on external `INFURA_PROJECT_KEY` and mocked timers, which may introduce CI/environment sensitivity.
> 
> **Overview**
> **Stabilizes wallet builds/tests by wiring in missing controllers and config.** The wallet package now depends on additional controllers (accounts/approval/connectivity/network/remote feature flags/transaction) and updates TS project references accordingly.
> 
> **Improves runtime/test ergonomics.** Jest loads a local `.env` (with `.env.example` added and `.env` gitignored), `Wallet` exposes stronger typed `messenger`/`state` and adds `destroy()` to clean up controller instances; tests are updated to require `INFURA_PROJECT_KEY`, use fake timers, and properly teardown the wallet.
> 
> **Tightens initialization typing and controller wiring.** Adds `initialization/defaults.ts` for inferred `DefaultInstances`/`DefaultActions`/`DefaultEvents`, introduces `bindMessengerAction` to preserve action typings, and updates controller initializers (notably `TransactionController` and `RemoteFeatureFlagController`) to pass required options and bind messenger actions safely.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6529337b2daf27d06b4a9e4645ab88939576880. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet/src/utilities.ts
```diff
@@ -59,12 +60,12 @@ export async function sendTransaction(
...
   options: AddTransactionOptions,
-) {
-  const { transactionMeta, result } = await wallet.messenger.call(
+): Promise<{ transactionMeta: TransactionMeta; result: Promise<string> }> {
+  const { transactionMeta, result } = (await wallet.messenger.call(
     'TransactionController:addTransaction',
     transaction,
     options,
-  );
+  )) as { transactionMeta: TransactionMeta; result: Promise<string> };
```
FrederikBolding: We can cast temporarily, but the real fix here is properly typing the `wallet.messenger` to include at least the actions and events that are supplied by the default configurations.

This will also give us intellisense 😄

packages/wallet/src/initialization/instances/network-controller.ts
```diff
@@ -24,36 +25,38 @@ export const networkController: InitializationConfiguration<
   name: 'NetworkController',
   init: ({ state, messenger, options }) => {
     // TODO: This was gutted to simplify implementation for now.
-    const getRpcServiceOptions = () => {
-      const maxRetries = DEFAULT_MAX_RETRIES;
+    const getRpcServiceOptions: NetworkControllerOptions['getRpcServiceOptions'] =
```
FrederikBolding, line 28: Is this a lint fix? 🤔

packages/wallet/src/initialization/initialization.ts
```diff
@@ -19,7 +19,7 @@ export function initialize({
   messenger,
   initializationConfigurations = [],
   options,
-}: InitializeArgs) {
+}: InitializeArgs): Record<string, Record<string, unknown>> {
```
FrederikBolding: I guess fine for now, but we should come up with a way to type this to at least include the instances from `defaultConfigurations`

packages/wallet/src/Wallet.test.ts
```diff
@@ -22,8 +43,22 @@ async function setupWallet() {
...
 describe('Wallet', () => {
+  let wallet: Wallet;
+
+  beforeEach(() => {
+    jest.useFakeTimers({ doNotFake: ['nextTick', 'queueMicrotask'] });
+  });
+
+  afterEach(async () => {
+    await wallet?.destroy();
+    cleanAll();
```
FrederikBolding: Clean all is already done by default in `core` IIRC

packages/wallet/src/Wallet.test.ts
```diff
@@ -22,8 +43,22 @@ async function setupWallet() {
...
 
 describe('Wallet', () => {
+  let wallet: Wallet;
+
+  beforeEach(() => {
+    jest.useFakeTimers({ doNotFake: ['nextTick', 'queueMicrotask'] });
+  });
+
+  afterEach(async () => {
+    await wallet?.destroy();
```
FrederikBolding, line 53: Nit: Any concern with just doing `wallet.destroy()` in each test case for now?

packages/wallet/src/Wallet.test.ts
```diff
@@ -7,12 +13,27 @@ const TEST_PHRASE =
...
+  }
+
   const wallet = new Wallet({
     options: {
-      infuraProjectId: 'infura-project-id',
+      infuraProjectId: process.env.INFURA_PROJECT_KEY,
       clientVersion: '1.0.0',
-      showApprovalRequest: () => undefined,
+      showApprovalRequest: (): undefined => undefined,
+      clientConfigApiService: new ClientConfigApiService({
```
FrederikBolding, line 28: I wonder if we need to pass this entire thing in? Maybe just add `client`, `distribution`, `environment` to `WalletOptions` and we can construct `ClientConfigApiService` based on that?

packages/wallet/src/initialization/types.ts
```diff
@@ -1,25 +1,57 @@
...
+): (
+  ...args: ExtractActionParameters<MessengerActions<Msgr>, ActionType>
+) => ExtractActionResponse<MessengerActions<Msgr>, ActionType> {
+  return messenger.call.bind(messenger, actionType) as (
+    ...args: ExtractActionParameters<MessengerActions<Msgr>, ActionType>
+  ) => ExtractActionResponse<MessengerActions<Msgr>, ActionType>;
+}
 
+// Method syntax provides bivariant parameter checking, which is needed to
+// collect heterogeneous InitializationConfiguration values in a single array.
```
FrederikBolding: I don't understand any of this

packages/wallet/src/types.ts
```diff
@@ -3,14 +3,19 @@ import {
   EventConstraint,
   Messenger,
 } from '@metamask/messenger';
+import type { ClientConfigApiService } from '@metamask/remote-feature-flag-controller';
 
 export type RootMessenger<
-  AllowedActions extends ActionConstraint = ActionConstraint,
-  AllowedEvents extends EventConstraint = EventConstraint,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  AllowedActions extends ActionConstraint = any,
```
FrederikBolding: The previous code is the RootMessenger from the clients, why do we need `any`? 🤔

packages/wallet/src/Wallet.test.ts
```diff
@@ -47,7 +82,7 @@ describe('Wallet', () => {
       { networkClientId: 'sepolia' },
     );
 
-    const hash = await result;
+    const hash = await jest.advanceTimersByTimeAsync(60_000).then(() => result);
```
FrederikBolding, line 85: I'm confused as to why this is needed. Why specifically 60 seconds?

packages/wallet/src/Wallet.test.ts
rekmarks: As opposed to in `afterEach`? Why?

packages/wallet/src/initialization/types.ts
rekmarks: It's basically saying that `init` is defined as a method as opposed to function property because the types will break otherwise. I edited the comment and moved it to `init`: e91bbe2a1

packages/wallet/src/Wallet.test.ts
rekmarks: There's something in the call chain that relies on timers and if we don't advance the timers the test hangs. The amount of time is arbitrary. Added a comment: e91bbe2a1

packages/wallet/src/initialization/instances/network-controller.ts
rekmarks: Yes, for explicit function return type.

packages/wallet/src/Wallet.test.ts
FrederikBolding: I generally find that more readable but it also provides flexibility in the way that we can pass different arguments to `setupWallet` in each test case.

packages/wallet/src/Wallet.test.ts
rekmarks: I think merging the constructor params of `ClientConfigApiService` into the `Wallet` constructor params would be more trouble than it's worth. If we ultimately decide that we don't care about all of the options of the former, we can always fold it in later.

packages/wallet/src/Wallet.test.ts
rekmarks: I prefer keeping cleanup in the hooks because customizing it between different test cases has a tendency to cause pernicious bugs in the test suite, IME.

PR #8428, opened 2026-04-10
feat: MUSD-248 create money-account-balance-service

Matt561:
## Explanation

Adds `MoneyAccountBalanceService` to `@metamask/money-account-controller` to support the MetaMask Money feature (mUSD stablecoin and musdSHFvd Veda vault shares). The service exposes five messenger actions: fetching mUSD and musdSHFvd ERC-20 balances via RPC, reading the Veda Accountant contract exchange rate, computing mUSD-equivalent vault share value, and fetching vault APY from Veda's Seven Seas REST API. All on-chain reads use `@ethersproject/contracts` with a `Web3Provider` resolved through `NetworkController` messenger actions.

## Key design decisions

- **TanStack Query caching**: RPC calls use a 30-second staleTime; the Veda APY endpoint uses 5 minutes. `getMusdEquivalentValue` delegates to the two underlying cached queries and performs BigInt arithmetic over cached results rather than issuing its own `fetchQuery`, avoiding cache coordination issues.
- **`VedaResponseValidationError` excluded from retry policy**: Malformed API responses are a deterministic failure — retrying them wastes quota. A custom error class signals the service policy to skip retry for validation failures.
- **Loose Superstruct validation**: The Veda APY schema uses `type()` rather than `object()` so unknown fields don't cause validation failures, guarding against future API additions without breaking the service.
- **`@metamask/utils` promoted to production dependency**: Previously a dev dependency, it is now required at runtime for hex/BigInt utilities used in balance and exchange rate computation.

## References

- [MUSD-248: Create MoneyAccountBalanceService to fetch users Veda Vault balances](https://consensyssoftware.atlassian.net/browse/MUSD-248)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new cached data service that performs on-chain RPC reads and external REST fetches; correctness and error-handling/caching behavior could impact displayed balances/APY, but changes are largely additive and isolated to a new package.
> 
> **Overview**
> Introduces a new package, `@metamask/money-account-balance-service`, implementing `MoneyAccountBalanceService` (a `BaseDataService`) that exposes messenger actions to fetch mUSD and vault-share ERC-20 balances, read the Veda Accountant exchange rate, compute mUSD-equivalent vault-share value, and fetch/normalize vault APY from Veda’s Seven Seas REST API (with Superstruct response validation and a non-retryable `VedaResponseValidationError`).
> 
> Wires the new package into repo metadata (adds to `README` package list/dependency graph, `CODEOWNERS`, `teams.json`, and `yarn.lock`) and includes full package scaffolding (build/test/typedoc configs, changelog, and extensive Jest coverage including caching and error-path behavior).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2deb59b447f2b70d1627882bbe28c2df11af3699. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Matt561:
@metamaskbot publish-previews

Matt561:
@metamaskbot publish-previews

Matt561:
@metamaskbot publish-previews

packages/money-account-controller/src/money-account-balance-service/money-account-balance-service.ts
```diff
@@ -0,0 +1,399 @@
...
+        const provider = this.#getProvider();
+        const contract = new Contract(
+          this.#config.accountantAddress,
+          ACCOUNTANT_ABI,
+          provider,
+        );
+        const rate = await contract.getRate();
+        return { rate: rate.toString() };
+      },
+      staleTime: inMilliseconds(30, Duration.Second),
```
Matt561: Note: We'll likely want to get the live exchange rate at runtime when depositing and withdrawing.

Matt561: Added optional `staleTime` override arg to all imperative fetching of live exchange rate during withdrawal flow.

Matt561:
@metamaskbot publish-previews

Matt561:
@metamaskbot publish-previews

shane-t (dismissed):

.github/CODEOWNERS
```diff
@@ -39,6 +39,7 @@
 
 ## Earn Team
 /packages/earn-controller                   @MetaMask/earn
+/packages/money-account-balance-service     @MetaMask/earn @MetaMask/core-platform
```
shane-t: no big deal either way, but do core-platform need to be co-owners of the whole package? in the upgrade controller I think we just made them owners of the package files/changelog

Matt561: Agreed, I'll update this real quick.

packages/money-account-balance-service/src/money-account-balance-service.ts
```diff
@@ -0,0 +1,403 @@
...
+        retryFilterPolicy: handleWhen(
+          (error) => !(error instanceof VedaResponseValidationError),
+        ),
+        ...policyOptions,
+      },
+    });
+
+    this.#config = {
+      ...config,
+      vedaApiBaseUrl: config.vedaApiBaseUrl ?? VEDA_PERFORMANCE_API_BASE_URL,
```
mcmire: It looks like this service takes a base URL as an option. Is there a particular reason for this? We don't recommend this because it means that consumers (i.e. some other team, even) may use this class in a way that you didn't intend.

Matt561:
@metamaskbot publish-previews

packages/money-account-balance-service/src/money-account-balance-service.ts
```diff
@@ -0,0 +1,403 @@
...
+   * @param args - The constructor arguments.
+   * @param args.messenger - The messenger suited for this service.
+   * @param args.policyOptions - Options to pass to `createServicePolicy`,
+   * which is used to wrap each request.
+   * @param args.config - The configuration for the service.
+   */
+  constructor({
+    messenger,
+    policyOptions = {},
+    config,
```
mcmire: Nit: Is there a reason why you are taking a separate set of options here? We recommend that all options for your data service be in the same "options bag". It's true that `messenger` and `policyOptions` are passed to BaseDataService, but consumers don't really know that (nor should they). All options should be treated on the same level.

If you want to have a single type for all options, you could create a `MoneyAccountBalanceServiceOptions` type and add `messenger` and `policyOptions` to it.

packages/money-account-balance-service/src/money-account-balance-service.ts
```diff
@@ -0,0 +1,403 @@
...
+        const provider = this.#getProvider();
+        const contract = new Contract(
+          this.#config.accountantAddress,
+          ACCOUNTANT_ABI,
+          provider,
+        );
+        const rate = await contract.getRate();
+        return { rate: rate.toString() };
+      },
+      staleTime: options?.staleTime ?? inMilliseconds(30, Duration.Second),
```
mcmire: Nit: Does this also work?

``` suggestion
  async getExchangeRate({
    staleTime = inMilliseconds(30, Duration.Second)
  }: {
    staleTime?: number;
  } = {}): Promise<ExchangeRateResponse> {
    return this.fetchQuery({
      queryKey: [`${this.name}:getExchangeRate`],
      queryFn: async () => {
        const provider = this.#getProvider();
        const contract = new Contract(
          this.#config.accountantAddress,
          ACCOUNTANT_ABI,
          provider,
        );
        const rate = await contract.getRate();
        return { rate: rate.toString() };
      },
      staleTime,

File: packages/money-account-balance-service/src/money-account-balance-service.ts [hunk 6]
```diff
@@ -0,0 +1,403 @@
...
+        if (!response.ok) {
+          throw new HttpError(
+            response.status,
+            `Veda performance API failed with status '${response.status}'`,
+          );
+        }
+
+        const rawResponse = await response.json();
+
+        // Validate raw response inside queryFn to avoid poisoned cache.
```
mcmire, line 392: Hmm, this is a point we didn't consider when creating the data service examples. I guess that does make sense.

packages/money-account-balance-service/src/money-account-balance-service.ts
```diff
@@ -0,0 +1,403 @@
...
+ *     accountantAddress: '0x...',
+ *     underlyingTokenAddress: '0x...',
+ *     underlyingTokenDecimals: 6,
+ *   },
+ * });
+ *
+ * const { balance } = await service.getMusdBalance('0xYourMoneyAccount...');
+ * ```
+ */
+export class MoneyAccountBalanceService extends BaseDataService<
```
mcmire, line 148: Since this talks to the Veda API would it make more sense to be literal and call this `VedaApiService`? Or do we expect to wrap this API in the future?

packages/money-account-balance-service/src/requestNormalization.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Normalizes the raw response from the Veda performance API into the expected
+ * format.
+ *
+ * @param rawResponse - The raw response from the Veda performance API.
+ * @returns The normalized response.
+ */
+export function normalizeVaultApyResponse(
+  rawResponse: Infer<typeof VaultApyResponseStruct>,
+): VaultApyResponse {
```
mcmire: Nit: It might be confusing that `VaultApyResponse` and `Infer<typeof VaultApyResponseStruct>` are not the same thing.

packages/money-account-balance-service/src/index.ts
```diff
@@ -0,0 +1,22 @@
+export {
+  MoneyAccountBalanceService,
+  serviceName as moneyAccountBalanceServiceName,
```
mcmire: The name of the service is not conventionally exported, can you remove this?

``` suggestion
```

mcmire (commented):
This looks pretty good. Many of the comments below are minor.

packages/money-account-balance-service/src/index.ts
Matt561: This is something I copied from the SocialService [here](https://github.com/MetaMask/core/blob/572fecfdbd7841461c340e981e273dde33507dd6/packages/social-controllers/src/index.ts#L68)

I will remove though if that's preferred.

packages/money-account-balance-service/src/requestNormalization.ts
Matt561: I've renamed `VaultApyResponseStruct` to `VaultApyRawResponseStruct` and `VaultApyResponse` to `NormalizedVaultApyResponse`. I'd like to differentiate between the raw response structure we get from Veda's API and our internal usage.

packages/money-account-balance-service/src/money-account-balance-service.ts
Matt561: Good question. I ultimately picked this name to better convey intent/context. If it's okay, I'd prefer to leave as is for now.

packages/money-account-balance-service/src/money-account-balance-service.ts
Matt561: I thought the separate options bag would be more intuitive but I understand your point. I'll refactor to place all options at the same level.

packages/money-account-balance-service/src/money-account-balance-service.ts
Matt561: I thought this would be convenient should Veda change their URL (e.g. new version) so we could update via remote flag. I'll remove.

Matt561:
@metamaskbot publish-previews

packages/money-account-balance-service/src/money-account-balance-service.test.ts
```diff
@@ -0,0 +1,725 @@
...
+describe('MoneyAccountBalanceService', () => {
+  beforeEach(() => {
+    MockWeb3Provider.mockImplementation(() => ({}) as unknown as Web3Provider);
+    nockCleanAll();
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
```
mcmire, line 272: Nit: This happens in our test setup already, so this should not be necessary:

``` suggestion
```

packages/money-account-balance-service/src/money-account-balance-service.test.ts
```diff
@@ -0,0 +1,725 @@
...
+}
+
+// ============================================================
+// Tests
+// ============================================================
+
+describe('MoneyAccountBalanceService', () => {
+  beforeEach(() => {
+    MockWeb3Provider.mockImplementation(() => ({}) as unknown as Web3Provider);
+    nockCleanAll();
```
mcmire, line 266: Nit: This happens in our test setup already, so this should not be necessary:

``` suggestion
```

mcmire (approved):
Two more comments, but they are non-blocking. LGTM!

Matt561:
Triple-checking that functionality works in Mobile before merging.

PR #8430, opened 2026-04-12
fix: improve assets controller snap accounts subscription and websocket

salimtb:
### Current state

`AssetsController` had three related issues that prevented balances from loading correctly in certain scenarios:

1. **Snap accounts not subscribed after startup**: On app launch, `#start()` was called before `AccountTreeController.init()` had finished building the account tree, so snap accounts weren't yet in the selected account group. Because `#start()` is idempotent (it returns early when subscriptions already exist), snap accounts were never picked up when they became available, leaving Solana, Tron, and Bitcoin balances missing.

2. **Custom assets filtered by spam heuristics**: `TokenDataSource` applied the `MIN_TOKEN_OCCURRENCES` threshold (EVM ERC-20) and Blockaid bulk scan (non-EVM fungible tokens) to all tokens uniformly. This meant tokens manually imported by the user and stored in `customAssets` state were incorrectly filtered out as potential spam.

3. **No polling fallback when WebSocket is disabled**: `BackendWebsocketDataSource` and `AccountsApiDataSource` both call the same `fetchV2SupportedNetworks` API to determine which chains they support. Because `BackendWebsocketDataSource` has higher priority in the chain-claiming loop, it would claim all supported chains at initialization — even when the WebSocket never connected. This left `AccountsApiDataSource` with no chains to poll, so users with the WebSocket disabled or with a broken connection received no balance updates.

### Solution

1. **Re-subscribe on AccountTree state change**: Added `#handleAccountTreeStateChange()` which subscribes to `AccountTreeController:stateChange`. When the account tree updates (e.g. after snap accounts are hydrated), the method forces a full re-subscription and re-fetch with the complete account list. Added `AccountsController:getSelectedAccount` as a fallback in `#getSelectedAccounts()` for cases where the account tree is not yet initialized.

2. **Custom assets bypass spam filters**: `TokenDataSource` now reads `customAssets` from state and builds a `customAssetIds` set. Tokens in that set skip the `occurrences >= 3` EVM filter and are excluded from the non-EVM Blockaid scan input, ensuring user-imported assets always appear.

3. **Connection-aware chain claiming in BackendWebsocketDataSource**: Added an `#isConnected` flag (initially `false`). `#initializeActiveChains` and `#refreshActiveChains` now store supported chains in `#supportedChains` but only call `updateActiveChains` when connected. On disconnect, `activeChains` is cleared so the chain-claiming loop re-assigns those chains to `AccountsApiDataSource` for polling. On reconnect, chains are restored from `#supportedChains` before pending WebSocket subscriptions are reprocessed.

## References

* Related to MetaMask/metamask-extension (assets multichain rollout)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes asset subscription/chain-claiming behavior and adds onboarding-based RPC suppression, which can affect when and how balances update across multiple chains and account types.
> 
> **Overview**
> Fixes multi-chain balance tracking edge cases by making `AssetsController` re-subscribe/re-fetch when the account tree later gains accounts (e.g. snap accounts), using `AccountsController:getSelectedAccount` as a startup fallback and tracking `#lastKnownAccountIds` to avoid redundant work.
> 
> Improves data-source coordination: `BackendWebsocketDataSource` now only claims chains while connected, releases them on disconnect so `AccountsApiDataSource` can poll as fallback, and reclaims on reconnect (dropping stale pending subscriptions). `TokenDataSource` now exempts user-imported `customAssets` from occurrence/Blockaid spam filtering.
> 
> Adds an optional `isOnboarded` hook plumbed into `RpcDataSource` so `fetch`/`subscribe` become no-ops before onboarding completes, preventing premature on-chain RPC calls; tests, changelog, and dependency graph are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5ac722270f6b8bf9bd6464fd1fd4c10c6795a0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ linkStyle default opacity:0.5
   approval_controller --> base_controller;
   approval_controller --> messenger;
   assets_controller --> account_tree_controller;
+  assets_controller --> accounts_controller;
   assets_controller --> assets_controllers;
   assets_controller --> base_controller;
   assets_controller --> client_controller;
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `isOnboarded` option to `AssetsControllerOptions` and `RpcDataSourceConfig` ([#8430](https://github.com/MetaMask/core/pull/8430))
+  - When `isOnboarded` returns `false`, `RpcDataSource` skips `fetch` and `subscribe` calls, preventing on-chain RPC calls before onboarding is complete.
+  - Defaults to `() => true` so existing consumers are unaffected.
+
 ### Changed
 
 - Bump `@metamask/account-tree-controller` from `^7.0.0` to `^7.1.0` ([#8472](https://github.com/MetaMask/core/pull/8472))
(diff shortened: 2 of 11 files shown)


salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

Prithpal-Sooriya (dismissed):

Prithpal-Sooriya (approved):

PR #8431, opened 2026-04-13
feat: skip simulation when enforced simulations container is active

matthewwalsh0:
## Explanation

When a transaction has the `EnforcedSimulations` container type, the simulation is redundant since the enforced simulation flow handles balance change protection separately. This PR skips the standard simulation in that case to avoid unnecessary network calls and potential conflicts.

## Changes

- Skip balance change simulation when `containerTypes` includes `EnforcedSimulations`
- Re-check `containerTypes` in the update callback to handle the race condition where simulation starts before the container type is set
- Use fresh state from `skipSimulationTransactionIds` in the callback instead of a stale closure, fixing an issue where toggling enforced simulations off would not restore balance changes

## References

- Extension PR: https://github.com/MetaMask/metamask-extension/pull/41126

---

> [!NOTE]
> **Medium Risk**
> Changes when/if `simulationData` is produced and persisted during transaction updates, which could affect balance-change protection behavior and downstream consumers that expect simulation results. Scope is contained and covered by a new unit test, but it touches core transaction lifecycle logic.
> 
> **Overview**
> Transactions marked with `containerTypes: [EnforcedSimulations]` now **skip balance-change simulation** (and avoid `getBalanceChanges` calls) when editable params change and would otherwise trigger re-simulation.
> 
> The simulation update path now re-checks the latest `containerTypes`/skip state inside the state update callback via a new `#isBalanceChangesSkipped` helper to avoid race conditions and prevent persisting `simulationData` when simulation is meant to be bypassed. Changelog updated and a regression test added to assert simulation is not re-run in this mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1b0199344332668aa99d7b118e5202fb45d7acf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Skip simulation when transaction `containerTypes` includes `EnforcedSimulations` ([#8431](https://github.com/MetaMask/core/pull/8431))
 - Bump `@metamask/accounts-controller` from `^37.1.1` to `^37.2.0` ([#8363](https://github.com/MetaMask/core/pull/8363))
 - Bump `@metamask/messenger` from `^1.0.0` to `^1.1.1` ([#8364](https://github.com/MetaMask/core/pull/8364), [#8373](https://github.com/MetaMask/core/pull/8373))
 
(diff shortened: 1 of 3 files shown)


packages/transaction-controller/CHANGELOG.md
```diff
@@ -23,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Skip simulation when transaction `containerTypes` includes `EnforcedSimulations` ([#8431](https://github.com/MetaMask/core/pull/8431))
```
vinistevam: Merge conflict?
```suggestion
```

vinistevam (dismissed):

vinistevam (approved):

PR #8437, opened 2026-04-13
feat(account-tree-controller): persist `accountTree`

ccharly:
## Explanation

We used to not persist the account-tree before and we were re-constructing it "fresh" upon `init`. However, some other controllers are dependent on the tree and have to wait for it to be built before consuming it.

Since the tree should never really change between lock/unlock, that's ok to have a on-disk copy of it to speedup consumers.

We will still try to re-construct it during `init` though (in case rules have changed or that new accounts have appeared that went unnoticed).

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Persists and rehydrates core controller state, which can affect startup/lock-unlock behavior and introduce stale/mismatched tree data if persisted state diverges from current accounts.
> 
> **Overview**
> `AccountTreeController` now **persists `accountTree`** (state metadata `persist: true`) instead of always recomputing it on `init`.
> 
> The controller constructor now pre-initializes internal reverse lookup maps from the persisted tree via `#initTreeContext`, allowing methods like `getAccountContext`, `getAccountGroupObject`, and `getAccountsFromSelectedAccountGroup` to work *before* `init()` runs.
> 
> Tests and snapshots are updated to cover pre-populated persisted state behavior and the new persisted-state surface, and the changelog documents the new persistence behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65dedea5f0e4c83537780870f0be5d5b84e9a662. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/account-tree-controller/src/AccountTreeController.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.ts
@@ -65,7 +65,7 @@ const accountTreeControllerMetadata: StateMetadata<AccountTreeControllerState> =
   {
     accountTree: {
       includeInStateLogs: true,
-      persist: false, // We do re-recompute this state everytime.
+      persist: true,
       includeInDebugSnapshot: false,
       usedInUi: true,
     },
@@ -235,6 +235,12 @@ export class AccountTreeController extends BaseController<
       this.#createBackupAndSyncContext(),
     );
 
+    // We build the initial tree (from the state), but this might get updated
+    // after `init` got called. Having it available now, allow our consumers to
+    // re-use the state that was already available before we restart/lock the
+    // wallet.
+    this.#initTreeContext(this.state.accountTree);
+
     this.messenger.subscribe('AccountsController:accountsAdded', (accounts) => {
       this.#handleAccountsAdded(accounts);
     });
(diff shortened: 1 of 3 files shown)


salimtb:
@metamaskbot publish-preview

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -275,6 +281,49 @@ export class AccountTreeController extends BaseController<
...
+        AccountGroupId,
+        AccountGroupObject,
+      ][]) {
+        this.#groupIdToWalletId.set(groupId, walletId);
+
+        // We still need to go through accounts for the sort order.
+        for (const accountId of group.accounts) {
+          // The `AccountsController` also persists its accounts, so we
+          // can fetch them too!
+          const account = this.messenger.call(
```
montelaidev: We can reduce this to one call by getting all the accounts

ccharly: I initially wanted to use `:getAccounts`, but didn't want to introduce a breaking change for this

However, I see we still have the old `listMultichainAccounts`, so we can probably use that + make a `Map` from it for fast indexing!

Good idea, let me change this!

hmalik88: @montelaidev beat me to it. I would prefer the one `getAccounts` call as well. I don't think it makes sense to rebuild essentially what is the same map to avoid a breaking change

hmalik88: something like this: 

```ts
        const accounts = this.messenger.call(
          'AccountsController:getAccounts',
          group.accounts,
        );

        group.accounts.forEach((accountId, idx) => {
          const account = accounts[idx];
          let sortOrder = MAX_SORT_ORDER;
          if (account) {
            sortOrder = ACCOUNT_TYPE_TO_SORT_ORDER[account.type];
          }

          this.#accountIdToContext.set(accountId, {
            walletId,
            groupId,
            sortOrder,
          });
        });
```

montelaidev (approved):

packages/account-tree-controller/src/AccountTreeController.ts
ccharly: I went for `:listMultichainAccounts` since adding `:getAccounts` would involve a breaking change for this controller and I'd like to keep it non-breaking for now to avoid the bubbling effect with this change 😅

Though, with `:listMultichainAccounts` + a `Map` we can get something pretty similar (I went for that).

ccharly: We can always change that later, but my goal for now is to unblock the assets team, we can introduce `:getAccounts` later for sure 👍

PR #8454, opened 2026-04-14
feat: add accountOverride to override default recipient or delegator for MM Pay transactions

jpuri:
## Explanation

Add accountOverride in transaction config to override default recipient or delegator for MM Pay transactions

## References

Fixes: https://consensyssoftware.atlassian.net/browse/CONF-1188
Fixes: https://consensyssoftware.atlassian.net/browse/CONF-1204

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes which account is used for balance checks and transaction/delegation signing, so incorrect wiring could cause quotes to be computed for the wrong account or submissions to be signed with an unexpected `from`.
> 
> **Overview**
> Adds a new per-transaction `accountOverride` config/state field that overrides the default `from` address used across MM Pay flows.
> 
> When `accountOverride` changes, the controller now re-runs source amount/quote updates and clears any selected `paymentToken`; quote building, payment-token balance lookups, Relay execute delegation (`txParams.from`), and HyperLiquid withdraw signing now all use the overridden address. Tests and changelog are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23839548549ac907788df591e5a5caa35843a72f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -127,6 +128,7 @@ export class TransactionPayController extends BaseController<
       transactionData.isPostQuote = config.isPostQuote;
       transactionData.isHyperliquidSource = config.isHyperliquidSource;
       transactionData.refundTo = config.refundTo;
+      transactionData.accountOverride = config.accountOverride;
```
matthewwalsh0: Minor, alphabetical.

packages/transaction-pay-controller/src/strategy/across/across-actions.ts
```diff
@@ -85,13 +85,13 @@ export function getAcrossDestination(
...
 
-  const { from } = request;
+  const { accountOverride, from } = request;
   const destinationCalls = getDestinationCalls(transaction);
   const swapRecipientTransferCallIndex = destinationCalls.findIndex((call) =>
     isExtractableOutputTokenTransferCall(call, request),
   );
   const callsForActions = [...destinationCalls];
-  let recipient = from;
+  let recipient = accountOverride ?? from;
```
matthewwalsh0: For a standard deposit flow, should `accountOverride` just be the funding account and gas payer, but the recipient of the funds should remain as the original `from`?

So in the money account example, `from` is money account so it gets the funds, but the source transactions are submitted by `accountOverride`.

packages/transaction-pay-controller/src/types.ts
```diff
@@ -100,6 +100,11 @@ export type TransactionConfig = {
    * go back to that account rather than the EOA.
    */
   refundTo?: Hex;
+
+  /** Optional address to override the default account used by the transaction.
```
matthewwalsh0: Can we elaborate here and confirm that the `accountOverride` provides the funds and pays for gas, and if `isPostQuote` is also the recipient.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -234,7 +234,10 @@ async function getSingleQuote(
...
         ? { originGasOverhead: getRelayOriginGasOverhead(messenger) }
         : {}),
-      recipient: from,
+      recipient:
+        request.isPostQuote && request.accountOverride
+          ? request.accountOverride
+          : from,
       slippageTolerance,
       tradeType: useExactInput ? 'EXACT_INPUT' : 'EXPECTED_OUTPUT',
       user: from,
```
matthewwalsh0, line 240: Do we also need to change this `user` since the `accountOverride` will be the one providing the funds and signing the source transactions?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0, line 1: For audit sake, just confirming that `processTransactions` shouldn't need changing as it already uses the original `transactionMeta` to build the delegations where the `from` would be the money account?

packages/transaction-pay-controller/src/strategy/across/across-actions.ts
matthewwalsh0, line 1: Do we need to update `across-quotes` to set the `depositor` to the `accountOverride`?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -93,6 +95,7 @@ export async function updateQuotes(
     });
 
     const requests = buildQuoteRequests({
+      accountOverride,
```
matthewwalsh0: Could we abstract this from the strategies by setting `from` here to `accountOverride` instead?

The delegations on the target side should still work as that uses the original full `transactionMeta`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0, line 1: For the fully gasless flow, do we need to update `submitViaRelayExecute` in `relay-submit` to use the `accountOverride` also since it would be providing the funds?

matthewwalsh0, line 1: Can do in a separate PR if easier, but for the withdraw flows, if we have a `accountOverride`, then we want that to sign the withdraw transaction rather than the original `from`.

So do we need to convert the withdraw transaction data to delegations on line 353 of `relay-submit.ts`?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/across/across-actions.ts
jpuri: That is good catch, let me fix this.

But I am bit confused here why we are using `from` account and not `to` account as destination ?

packages/transaction-pay-controller/src/TransactionPayController.ts
jpuri: PR is updated to address this.

packages/transaction-pay-controller/src/strategy/across/across-actions.ts
jpuri: PR is updated to address this.

packages/transaction-pay-controller/src/types.ts
jpuri: PR is updated to address this.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: PR is updated to address this.

packages/transaction-pay-controller/src/utils/quotes.ts
jpuri: This is nice simplification, PR is updated.

packages/transaction-pay-controller/src/strategy/across/across-actions.ts
jpuri: Since `depositor` is set to `from` which we updated in quotes.ts to use accountOverride. This change is not needed separately.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: This should also already work as `from` is now updated in quote.

jpuri: This should also already work since from is `quote` is updated.

packages/transaction-pay-controller/src/types.ts
```diff
@@ -100,6 +100,13 @@ export type TransactionConfig = {
    * go back to that account rather than the EOA.
    */
   refundTo?: Hex;
+
+  /**
+   * Optional address to override the default account used by the transaction.
+   * When `isPostQuote` is true, used as the recipient of the MM Pay transfer.
+   * When `isPostQuote` is false, used as the delegator for the transaction, it provides the funds and pays for gas.
```
matthewwalsh0: Slightly ambigious as the `delegation` is still done on behalf of the original `from` and not `accountOverride`. Plus this also still applies when `isPostQuote` is `true`.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0: This flow just uses the `transaction` and not `from` so I think we need to fix this manually.

matthewwalsh0: This is also a special case as while the data is currently using the original `from` via `transaction` which is what we want, we now need to convert it to delegations with `getDelegationTransaction` so our `accountOverride` can submit it.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/types.ts
jpuri: Valid point, I updated the PR

jpuri:
> This is also a special case as while the data is currently using the original from via transaction which is what we want, we now need to convert it to delegations with getDelegationTransaction so our accountOverride can submit it.

@matthewwalsh0 : PR is updated to address this.

matthewwalsh0:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: PR is updated to address this

jpuri: PR is updated to address this

matthewwalsh0 (approved):

PR #8455, opened 2026-04-14
feat: add `duration` and `traceId` to degraded events

cryptodev-2s:
## Explanation

RPC endpoint degraded events (`NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointChainDegraded`) now include two new optional properties in their payloads:

- **`duration`** (`number | undefined`): The policy execution time in milliseconds when the request succeeded but was slow (i.e., exceeded the `degradedThreshold`). This is `undefined` when the degraded event was caused by retries being exhausted.

- **`traceId`** (`string | undefined`): The value of the `x-trace-id` response header from the last request attempt. This enables correlating degraded events with backend traces for debugging RPC health issues. It is `undefined` when no response was received or the header was not present.

### How it works

The `duration` value originates from Cockatiel's `retryPolicy.onSuccess` callback in `createServicePolicy`. Previously, it was only used for the threshold comparison — now it is also emitted as part of the degraded event data.

The `traceId` is captured in `RpcService` from `response.headers.get('x-trace-id')` in the same `finally` block that tracks `rpcMethodName`, ensuring it reflects the last completed request attempt (handling concurrent request race conditions correctly).

Both values are threaded through the event chain: `createServicePolicy` → `RpcService` → `RpcServiceChain` → `createNetworkClient` → messenger events.

### Breaking change

The `RpcServiceRequestable` type's `onDegraded` listener signature now includes `duration?: number` and `traceId?: string` in its data parameter. Implementors of this interface will need to accept the new fields in their `onDegraded` callback signature.

## References

- https://consensyssoftware.atlassian.net/browse/WPC-963

## Checklist

---

> [!NOTE]
> **Medium Risk**
> This is a breaking, cross-package event/type change that affects all `onDegraded`/messenger subscribers and could cause downstream TypeScript or runtime assumptions to fail if payload handling isn’t updated. Logic changes are small but touch retry/degraded signaling and request metadata capture.
> 
> **Overview**
> **Degraded-event payloads now include more diagnostics.** `createServicePolicy`’s `onDegraded` event no longer emits `void` for slow successes; it emits `{ duration: number }`, and tests were updated accordingly.
> 
> **Network degraded messenger events are extended.** `NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointChainDegraded` payloads add optional `duration` (slow success) and `traceId` (last response `X-Trace-Id` header), with plumbing added from `RpcService` through `RpcServiceChain`/`createNetworkClient` and expanded test coverage (including avoiding stale `traceId` when a retry throws before a response).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9867d45505185d37ca55c8eb276ce1a4972b1b82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -655,6 +672,8 @@ export class RpcService implements AbstractRpcService {
             // before `onDegraded` gets called, no matter the outcome of the
             // request.
             this.#currentRpcMethodName = rpcMethodName;
+            this.#currentTraceId =
+              response?.headers.get('x-trace-id') ?? undefined;
```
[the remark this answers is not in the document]
cryptodev-2s: Good catch! Fixed in e61cfa5cb — `response` is now reset to `undefined` at the start of each retry attempt, before the `try` block. This ensures that if `this.#fetch` throws before assigning a new response (e.g., a connection error), the `finally` block sees `undefined` rather than a stale response from a previous attempt, so `#currentTraceId` correctly becomes `undefined`.

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/controller-utils/src/create-service-policy.test.ts
```diff
@@ -1500,6 +1500,27 @@ describe('createServicePolicy', () => {
               expect(onDegradedListener).toHaveBeenCalledTimes(1);
             });
 
+            it('calls onDegraded listeners with { duration } when the request succeeds but is slow', async () => {
```
mcmire: Nit:

``` suggestion
            it('calls onDegraded listeners with duration when the request succeeds but is slow', async () => {

File: packages/controller-utils/CHANGELOG.md [hunk 3]
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING:** The `ServicePolicy` type's `onDegraded` event now emits `{ duration: number }` instead of `void` when the service succeeds but takes longer than the `degradedThreshold` ([#8455](https://github.com/MetaMask/core/pull/8455))
```
mcmire, line 12: Another big release 🙈🙈🙈

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Should we update this comment to reflect that we are setting two properties now?

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -619,6 +636,10 @@ export class RpcService implements AbstractRpcService {
       );
       const jsonDecodedResponse = await this.#policy.execute(
         async (context) => {
+          // Reset response so that if this attempt throws before
```
mcmire, line 656: I'm not sure I understand how it could be possible to create a stale response. If the attempt throws before assigning a new response, there will be no response. And if the attempt succeeds, there _will_ be a response.

Can you write a test for this?

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -443,16 +452,24 @@ export class RpcService implements AbstractRpcService {
...
+        endpointUrl: this.endpointUrl.toString(),
+        rpcMethodName: this.#currentRpcMethodName,
+        traceId: this.#currentTraceId,
+      };
+      if (hasProperty(data, 'duration')) {
         listener({
-          endpointUrl: this.endpointUrl.toString(),
-          rpcMethodName: this.#currentRpcMethodName,
+          ...commonFields,
+          duration: data.duration as number,
```
mcmire: Why are we typecasting here?

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -443,16 +452,24 @@ export class RpcService implements AbstractRpcService {
     listener: Parameters<AbstractRpcService['onDegraded']>[0],
   ): IDisposable {
     return this.#policy.onDegraded((data) => {
-      if (data === undefined) {
+      // Two event shapes reach here:
+      // - Slow-success: data is { duration: number }
+      // - Retries-exhausted: data is FailureReason ({ error } or { value })
```
mcmire: This is true, but is this the right place for this knowledge? We only test for `duration` here anyway, we don't care whether there is `error` or `value` (we care about that later).

Also `slow_success` and `retries_exhausted` have to do with the metrics. We don't use that terminology in this file.

Maybe we don't need these comments? 

``` suggestion
```

packages/controller-utils/src/create-service-policy.test.ts
cryptodev-2s: Fixed in 63c1956f2.

packages/network-controller/src/rpc-service/rpc-service.ts
cryptodev-2s: Updated in 63c1956f2.

packages/network-controller/src/rpc-service/rpc-service.ts
cryptodev-2s: The `response` variable is declared outside the `policy.execute` callback, so it persists across retry attempts. If attempt 1 gets a 503 (setting `response`), and attempt 2's `fetch` throws a network error before assigning, `response` still holds the stale value from attempt 1. Added a test in 63c1956f2.

packages/network-controller/src/rpc-service/rpc-service.ts
cryptodev-2s: Because `hasProperty` narrows `data.duration` to `unknown` (not `number`), since `duration` isn't in `keyof` of the union. Replaced with a `typeof` check to narrow naturally without a cast in 63c1956f2.

packages/network-controller/src/rpc-service/rpc-service.ts
cryptodev-2s: Removed the comments in 63c1956f2.

cryptodev-2s:
@metamaskbot publish-preview

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Ah, okay. Thanks for the clarification.

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Ah yes. I don't like this about `hasProperty`. Makes sense to add an extra check. Although there might still be some type errors here.

packages/network-controller/src/rpc-service/rpc-service.ts
```diff
@@ -443,16 +452,21 @@ export class RpcService implements AbstractRpcService {
...
         });
       } else {
         listener({
           ...data,
-          endpointUrl: this.endpointUrl.toString(),
-          rpcMethodName: this.#currentRpcMethodName,
+          ...commonFields,
+          duration: undefined,
         });
       }
```
mcmire: There still seem to be some type errors here. I think this can just be:

``` suggestion
      listener({
        ...data,
        endpointUrl: this.endpointUrl.toString(),
        rpcMethodName: this.#currentRpcMethodName,
        traceId: this.#currentTraceId,
      });
```

packages/network-controller/CHANGELOG.md
```diff
@@ -24,9 +24,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `NetworkController:loadBackup`
   - Corresponding action types are available as well.
 - Add `getEthQuery` method to `NetworkController` ([#8350](https://github.com/MetaMask/core/pull/8350))
+- Add `duration` and `traceId` to `NetworkController:rpcEndpointDegraded` and `NetworkController:rpcEndpointChainDegraded` event payloads ([#8455](https://github.com/MetaMask/core/pull/8455))
+  - `duration` contains the policy execution time in milliseconds when the request succeeded but was slow. It is `undefined` when retries were exhausted.
+  - `traceId` contains the value of the `x-trace-id` response header from the last request attempt, or `undefined` if the header was not present.
 
 ### Changed
 
+- **BREAKING:** The `RpcServiceRequestable` type's `onDegraded` listener now receives `duration?: number` and `traceId?: string` in its data parameter ([#8455](https://github.com/MetaMask/core/pull/8455))
```
mcmire: I've been meaning to get rid of this type as we don't really need it anymore. If we change `RpcService` not to implement this interface anymore then we don't need to change this type at all, and we can avoid the breaking change here. May be worth doing that before merging this.

mcmire: I've made the change I mentioned here: https://github.com/MetaMask/core/pull/8475

packages/network-controller/src/rpc-service/rpc-service.ts
mcmire: Fixed here: [0e695d0](https://github.com/MetaMask/core/pull/8455/commits/0e695d0a3f82e1e10142704fd9c5465b634e590b)

packages/network-controller/CHANGELOG.md
mcmire: I've removed the changes from `RpcServiceRequestable` as well as the changelog here: [0e695d0](https://github.com/MetaMask/core/pull/8455/commits/0e695d0a3f82e1e10142704fd9c5465b634e590b)

packages/network-controller/src/NetworkController.ts
```diff
@@ -562,18 +572,22 @@ export type NetworkControllerRpcEndpointChainDegradedEvent = {
...
  * when the endpoint became degraded.
+ * @param payload.traceId - The value of the `x-trace-id` response header from
+ * the last request attempt, or `undefined` if the header was not present.
  */
 export type NetworkControllerRpcEndpointDegradedEvent = {
   type: 'NetworkController:rpcEndpointDegraded';
   payload: [
     {
       chainId: Hex;
+      duration?: number;
```
mcmire, line 583: I have to check to make sure that this also isn't a breaking change. I thought at first it wasn't, but then given that event payloads show up as arguments to callbacks, I am having second doubts.

mcmire: Traditionally we have not treated additional payload properties as breaking, but I think we should, so that's what I've done here.

mcmire (dismissed):
LGTM.

mcmire (dismissed):

mcmire (dismissed):

mcmire (approved):
LGTM!

PR #8462, opened 2026-04-15
feat: implement quote status update logic

GeorgeGkas:
### Context

The Bridge API requires clients to report the lifecycle of a swap quote — when a transaction is submitted and when it finalizes (success or failure). Previously, `BridgeStatusController` had no mechanism to send these reports, so the Bridge API had no visibility into quote outcomes.

Additionally, batch EVM transactions (Smart Transactions / 7702-delegated flows) do not have a `txHash` available at `submitTx` time, meaning status reporting for those transactions was impossible without subscribing to a later event.

### Solution

This PR introduces `QuoteStatusUpdateManager` inside `bridge-status-controller` — a resilient, persisted retry queue that reports quote lifecycle events (`SUBMITTED`, `FINALIZED_SUCCESS`, `FINALIZED_FAILURE`) to the Bridge API.

Key behaviours:
- **Immediate retries on transient errors** — up to 6 attempts with a short delay before giving up or deferring.
- **Deferred retry queue for network failures** — entries are retried every 30 minutes for up to 12 hours, and the queue is persisted to `deferredStatusUpdates` controller state so it survives service-worker restarts.
- **Non-retryable errors are evicted** — errors like `SVM_TRADE_DESERIALIZE_FAILED` are dropped immediately rather than retried endlessly.
- **Feature-flagged** — a new `isQuoteStatusUpdateEnabled` constructor option (backed by the `bridgeQuoteStatusUpdateEnabled` remote feature flag in the extension) gates all reporting.
- **Error forwarding** — a new `onQuoteStatusUpdateError` constructor option lets callers (e.g. the extension) forward eviction errors to Sentry.

To support STX / 7702-delegated batch transactions, `BridgeStatusController` now subscribes to `TransactionController:transactionSubmitted` and triggers `SUBMITTED` status reporting when a hash becomes available after the fact.

`bridge-controller`'s `QuoteResponseSchema` is updated to treat `quoteId` as a required field, since it is the stable identifier used as the key for the retry queue.

### Non-obvious changes

- `BridgeStatusControllerMessenger.AllowedEvents` now requires `TransactionControllerTransactionSubmittedEvent` — consumers must add this event to their messenger `AllowedEvents` list.
- A bug in `submitIntent` was fixed along the way: `originalTransactionId` is now passed at the top level to `#addTxToHistory`, so the history item is correctly linked to the synthetic `TransactionController` entry rather than the `orderUid`.

## References

- Jira: https://consensyssoftware.atlassian.net/browse/SWAPS-4399
- Related: `isQuoteStatusUpdateEnabled` is gated by `remoteFeatureFlags.bridgeQuoteStatusUpdateEnabled`

## Checklist

---

> [!NOTE]
> **High Risk**
> Touches core bridge submission and status flows with breaking state/types, new persisted queue, and many lifecycle hooks where missed or duplicate API reports could affect Bridge backend visibility.
> 
> **Overview**
> Adds **resilient quote-status reporting** to the Bridge API (`SUBMITTED`, `FINALIZED_SUCCESS`/`FINALIZED_FAILURE`) via a new **`QuoteStatusManager`** module and wires it through **`BridgeStatusController`**.
> 
> **Breaking API surface:** `BridgeHistoryItem` now expects **`quoteId`** and **`reportedSubmittedTxHash`**; controller state gains persisted **`quoteUpdateStatusStore`**; constructor requires **`clientProduct`** (optional **`clientVersion`**, **`onQuoteStatusManagerError`**, **`isQuoteStatusManagerEnabled`**).
> 
> Reporting is triggered from multiple paths so hashes and outcomes are not missed: **early at `submitTx`** (including non-EVM), on **`TransactionStatus.submitted`** (EVM hash replacements), during **status polling** when a src/settlement hash appears, and on **confirm/fail/terminal poll** paths via **`reportFinalised`**. **`#reportSubmittedOnce`** dedupes per hash and skips items without `quoteId`.
> 
> The manager layer includes an **API service** (`quote/updateStatus`, `getQuoteStatus`) with **immediate retries** on transient backend errors and a **persisted deferred queue** (30‑minute retries, 12‑hour TTL), plus an **entry store** and **FSM** for allowed transitions. Public exports add quote-status **error types** and constants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5ccfda16c32d23ee5871d1706f0dcde86cdba24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -338,6 +411,11 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       txHash: txMeta.hash,
     });
 
+    // Report finalized success for swap/bridge transactions
+    if (txMeta.type && isCrossChainTx(txMeta.type)) {
+      this.#quoteStatusUpdateManager.reportFinalised(txMeta.id, true);
+    }
```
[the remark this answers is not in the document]
GeorgeGkas: This is expected, we care only about reporting the success/failure of the the source tx.

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add required `quoteId: string` to `QuoteResponseSchema`; quote responses without a `quoteId` will now fail validation ([#8462](https://github.com/MetaMask/core/pull/8462))
```
infiniteflower: This is a breaking change as it requires clients to change code if they upgrade to this version.

GeorgeGkas: How do we expect clients to update code? At least on extension there is no interface impacted by the inclusion of quoteId. Does it exist on mobile or is there another reason for marking it as breaking?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -269,6 +315,28 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+    this.messenger.subscribe(
+      'TransactionController:transactionSubmitted',
+      ({ transactionMeta }) => {
+        const { type, id: txMetaId, hash, actionId } = transactionMeta;
+        if (hash && type && isCrossChainTx(type)) {
+          const historyItem =
+            this.state.txHistory[txMetaId] ??
+            (actionId ? this.state.txHistory[actionId] : undefined);
+          if (historyItem?.quoteId) {
+            this.#quoteStatusUpdateManager.reportSubmitted(
```
infiniteflower: Just wondering why this is called twice, once here at `transactionSubmitted` and once at `transactionStatusUpdated` (`submitted` status) above? Aren't those 2 the same and would there be duplicated reports?

packages/bridge-status-controller/src/types.ts
```diff
@@ -264,8 +270,18 @@ export type StartPollingForBridgeTxStatusArgsSerialized = Omit<
 
 export type SourceChainTxMetaId = string;
 
+export type DeferredStatusUpdateEntry = {
+  quoteId: string;
+  srcTxHash: string;
+  pendingStatuses: string[];
```
infiniteflower: Do we want to make this an explicit enum so we understand the full spectrum of possible statuses?

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: @infiniteflower they are indeed similar. There was a change recently introduced on the transaction controller where they added a universal event `TransactionController:transactionStatusUpdated`. I originally made the implementation using the separate submitted event listener and later they introduced the updated event. I kept both for completeness, note there are not issues with duplicate calls, I already tested that.

GeorgeGkas: @infiniteflower I gave it another look and we indeed require both. TransactionController:transactionStatusSubmitted is called immediately once we broadcast the tx while the update event is triggered on later time. If we omit the submitted event listener, there is an edge case where the quote status manager won't be able to create a record if user closes the app immediately after submitting a swap (or loose internet connection). We'll never get back the tx status in that case.

packages/bridge-status-controller/src/types.ts
GeorgeGkas: good call, I planned to do it today but then I worked on test failures and forgot. We should use `QuoteStatusUpdateStatus` on pendingStatuses rather than generic string.

packages/bridge-controller/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
```
micaelae: Please mark this as a breaking change. Client e2e tests will need to be updated

packages/bridge-status-controller/src/bridge-status-controller.ts
micaelae: the `#onTransactionStatusChange` handler in TransactionController (line 3337) publishes `transactionStatusUpdated` right after submitted, so I think this does duplicate the event handling

micaelae: We use the generic transaction statusUpdated event instead of specific ones like `transactionSubmitted` because the initial hash generated during submission can be replaced before it's confirmed on chain

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -642,6 +724,10 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       this.stopPollingByPollingToken(pollingToken);
       delete this.#pollingTokensByTxMetaId[bridgeTxMetaId];
 
+      // Finalize the quote status update as a failure since we can no longer
+      // determine the outcome via polling.
+      this.#quoteStatusUpdateManager.reportFinalised(bridgeTxMetaId, false);
```
micaelae: The controller restarts polling for these transactions next time the wallet restarts so it's too soon to report finalization here. This will also result in duplicate finalization status updates

micaelae: The #handleOldHistoryItem handler deletes the history item to permanently stop polling so the report should happen there

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -232,6 +260,24 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+          case TransactionStatus.submitted:
+            // For batch EVM (STX/gasIncluded7702), the hash is not available at submitTx time.
+            // Call reportSubmitted when the transaction reaches 'submitted' status (hash is now available).
+            if (
+              txMeta.hash &&
+              txMeta.type &&
+              isCrossChainTx(txMeta.type) &&
+              !isApprovalTxMeta
+            ) {
+              if (historyItem?.quoteId && txMeta.hash) {
```
micaelae: There are multiple places in the code where we assume `quoteId` is defined, but it won't be for older txHistory items. Can you make it optional in the BridgeHistoryItem type to ensure this undefined check is always done?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -338,6 +411,11 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       txHash: txMeta.hash,
     });
 
+    // Report finalized success for swap/bridge transactions
+    if (txMeta.type && isCrossChainTx(txMeta.type)) {
+      this.#quoteStatusUpdateManager.reportFinalised(txMeta.id, true);
```
micaelae: for bridge transactions, do we only care about the src transaction? or should finalized only be reported when both src and dest are confirmed?

packages/bridge-status-controller/src/quote-status-update-manager.ts
```diff
@@ -0,0 +1,601 @@
...
+        }),
+      },
+    );
+
+    if (res.ok) {
+      return undefined;
+    }
+
+    return (await res.json()) as QuoteStatusUpdateResponse;
+  };
```
micaelae: Please add a validator and infer the type for QuoteStatusUpdateResponse rather than casting the response. Validating ensures that the data matches the shape we expect. In the worst case the app or background worker may crash when we attempt to use unvalidated objects

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -163,6 +174,8 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
     addTransactionBatchFn,
     config,
     traceFn,
+    onQuoteStatusUpdateError,
```
micaelae: Is this logic going to be different across clients?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -163,6 +174,8 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+    isQuoteStatusUpdateEnabled,
```
micaelae: If this is only based on a client feature flag, I'd read the flag directly in the controller/update manager

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -295,6 +363,11 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
       txHash: isApprovalTxMeta ? undefined : txMeta.hash,
     });
 
+    // Report finalized failure for swap/bridge transactions
+    if (txMeta.type && isCrossChainTx(txMeta.type)) {
+      this.#quoteStatusUpdateManager.reportFinalised(txMeta.id, false);
```
micaelae: Have you tested this with a HW wallet rejection? If you submit the trade, but reject the tx on a HW wallet, does the `submitted` event fire? If it doesn't then this should happen after the `rejected` check below

packages/bridge-status-controller/src/quote-status-update-manager.ts
```diff
@@ -0,0 +1,601 @@
...
+    // This method uses `globalThis.fetch` and reads the raw
+    // `Response` (including JSON on non-2xx). Wrappers like `handleFetch` that
+    // throw on non-2xx would prevent typed error handling in callers.
+    const res = await globalThis.fetch(
+      `${this.#apiBaseUrl}/quote/updateStatus`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          ...getClientHeaders({
```
micaelae: Pass the clientVersion too

packages/bridge-status-controller/src/quote-status-update-manager.ts
```diff
@@ -0,0 +1,601 @@
...
+      type === QuoteStatusUpdateErrorType.InvalidStatusTransaction ||
+      type === QuoteStatusUpdateErrorType.QuoteStatusOnChainMismatch
+    ) {
+      const finalizationStatus = response.currentStatus;
+      entry.pendingStatuses = [finalizationStatus];
+      this.#persistToState();
+      this.#inFlight.delete(key);
+      await this.#attemptFinalizationWithRetries(key, entry);
+      return;
+    }
```
micaelae: Why post the status update again? If this happens, doesn't it mean the backend already has the fimnalizationStatus. So all we need to do is remove this from the deferred statuses

packages/bridge-status-controller/src/quote-status-update-manager.ts
```diff
@@ -0,0 +1,601 @@
...
+
+    // If there are items to be processed, start the poller and
+    // immediately attempt to process all entries (don't wait for
+    // the first interval tick).
+    if (this.#deferredRetryQueue.size > 0) {
+      this.#ensureRetryTimerRunning();
+      let delay = 0;
+      for (const key of this.#deferredRetryQueue.keys()) {
+        setTimeout(() => this.#processSingleEntry(key), delay);
+        delay += 125;
```
micaelae: ensureRetryTimerRunning also causes these entries to be processed. Why do both need to be started around the same time?

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: That was my initial thought as well, but I wanted to keep the controller as stateless as possible, which is why I prefer to pass callbacks and let clients handle initialization. While directly reading the flag might seem appropriate for now, it introduces tight coupling between our service and an external dependency (the LD flags), which complicates testing.  

My main concern, however, is that this won’t be the final requirement for the update manager. Since Ethan is planning to add more logic after v1, and we don’t know the exact changes yet, I wanted to keep the interface generic. That way, when we eventually update the implementation logic, the public API can remain stable, making future documentation simpler and more consistent.

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: Good catch, will have to do it.

GeorgeGkas:
From @micaelae : 

> I think this means the controller will need to publish the “submitted” event for intents after the submitOrder call as well

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: 👀

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: 👀

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: we only care about the src tx

packages/bridge-status-controller/src/quote-status-update-manager.ts
GeorgeGkas: 👀

packages/bridge-status-controller/src/bridge-status-controller.ts
GeorgeGkas: I do not have clear context yet, I think not, but even if they are the same, I want to keep the API lean so that we keep coupling of modules low, it's easier to test and more maintainable if we want to change the way we handle error reporting in the future.

packages/bridge-status-controller/src/quote-status-update-manager.ts
GeorgeGkas: this is a requirement, if we sent a `submit` status while it was already finalized on backend, then we need to follow up with the correct finalization status. This is the reason why we construct the pendingStatuses again and send it to queue.

packages/bridge-status-controller/src/quote-status-update-manager.ts
GeorgeGkas: both are needed, `ensureRetryTimerRunning` starts a timer, but does not process the entries immediately. This is done by the for loop.

packages/bridge-status-controller/src/quote-status-manager/quotes-status-manager.ts
```diff
@@ -0,0 +1,651 @@
...
+      // batch/7702 swaps whose type may still read as `batch` rather than `swap`.
+      const isCrossChainTrade =
+        (txMeta.type !== undefined && isCrossChainTx(txMeta.type)) ||
+        hasNestedSwapTransactions(txMeta);
+      if (!isCrossChainTrade) {
+        continue;
+      }
+
+      if (txMeta.status === TransactionStatus.confirmed) {
+        this.reportFinalised(entry.txMetaId, true);
```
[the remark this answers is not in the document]
GeorgeGkas: this is the expected behavior, we report finalized status for source txs on bridges.

packages/bridge-status-controller/src/quote-status-manager/quotes-status-manager.ts
```diff
@@ -0,0 +1,710 @@
...
+              sentStatus === QuoteStatusState.FinalizedFailed
+            ) {
+              // A finalized status was accepted; advance to the terminal
+              // `Completed` state and keep the entry so any later duplicate
+              // `reportSubmitted` for this quote is rejected instead of looping.
+              this.#markCompleted(current);
+              // Call getStatus to complete the flow on the backend.
+              // We discard any value for now, this is used purely
+              // to notify backend that the flow is complete.
+              this.getStatus(entry.quoteId, { maxRetries: 1 });
```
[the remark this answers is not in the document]
GeorgeGkas: N/A

SteP-n-s (dismissed):

SteP-n-s (approved):

PR #8464, opened 2026-04-15
chore: bump `accounts` deps

ccharly:
## Explanation

Bumping all `accounts` deps to the latest. Mostly including the new `/v2` exports for the on-going work with the keyring API v2.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches account/keyring-adjacent packages and upgrades multiple keyring dependencies, including switching some types/imports to `@metamask/keyring-api/v2`, which could cause subtle type/runtime integration issues if consumers or downstream deps assume old exports.
> 
> **Overview**
> Updates multiple controllers/services to newer keyring/account package versions (notably `@metamask/keyring-api@^23.0.1`, `@metamask/keyring-internal-api@^10.1.1`, `@metamask/keyring-snap-client@^9.0.1`, `@metamask/eth-snap-keyring@^21.0.1`, and `eth-*` keyrings), with corresponding changelog and lockfile updates.
> 
> Adopts new `/v2` exports in key places: `keyring-controller` now imports V2 keyring classes/types from `@metamask/eth-hd-keyring/v2`, `@metamask/eth-simple-keyring/v2`, and `@metamask/keyring-api/v2`, and `multichain-account-service` updates `KeyringCapabilities` typing to come from `@metamask/keyring-api/v2`.
> 
> Adds Stellar account-type ordering in `account-tree-controller` by including `XlmAccountType.Account` in `ACCOUNT_TYPE_TO_SORT_ORDER`, and tightens `chain-agnostic-permission` typing by explicitly enumerating supported non-wallet CAIP namespaces rather than using `Exclude`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc56540fa11144059de1d3f4bb2bf92bd6a87725. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
Added `no-changelog` since I made an internal change in the `chain-agnostic-permission` package (impacted by the `metamask/utils` dedupe, one of the `enum` got changed, and that can be seen as breaking change if you use this enum in your types directly...)

Not sure if there's a better way to handle this for changelogs?

jest.config.packages.js
```diff
@@ -83,6 +83,16 @@ module.exports = {
...
     ],
+    // Required for the keyring API v2 transition until `/v2` is merged with the top-level: {
+    '^@metamask/eth-hd-keyring/v2$':
+      '<rootDir>/../../node_modules/@metamask/eth-hd-keyring/dist/v2/index.cjs',
+    '^@metamask/eth-simple-keyring/v2$':
+      '<rootDir>/../../node_modules/@metamask/eth-simple-keyring/dist/v2/index.cjs',
+    '^@metamask/keyring-api/v2$':
+      '<rootDir>/../../node_modules/@metamask/keyring-api/dist/v2/index.cjs',
+    '^@metamask/keyring-sdk/v2$':
+      '<rootDir>/../../node_modules/@metamask/keyring-sdk/dist/v2/index.cjs',
```
ccharly: Somehow, jest is not able to resolve that without doing it explicitly 😅

packages/chain-agnostic-permission/src/scope/types.ts
```diff
@@ -116,10 +116,14 @@ export const parseScopeString = (
 /**
  * CAIP namespaces excluding "wallet" currently supported by/known to the wallet.
  */
-export type NonWalletKnownCaipNamespace = Exclude<
-  KnownCaipNamespace,
-  KnownCaipNamespace.Wallet
->;
+export type NonWalletKnownCaipNamespace =
```
montelaidev, line 119: Do we need one for stellar?

ccharly: Well, I wasn't sure if there was some more logic needed in the package (not my domain here 😅)

So I preferred to just use an explicit enumeration instead (to avoid having other breaking change in the future). But I'll let the wallet API team to guide me on this one.

But indeed, given the name, we could re-use the `Exclude` and declare the missing field for the type that relies on it (but that could be seen as a breaking change though..)

jiexi (dismissed):
`packages/chain-agnostic-permission/src/scope/types.ts` LGTM

jest.config.packages.js
ccharly: Not required anymore!

jiexi (approved):

gantunesr (approved):

bfullam (approved):

Jwhiles (approved):

geositta (approved):

mcmire:
@ccharly Given that we are switching to a `/v2` import for `keyring-api` and advancing a bunch of packages by a couple of major versions I just want to verify that this is correct and that these bumps won't cause any type incompatibilities? There don't seem to be any breaking changes noted here so I wanted to be extra sure.

ccharly:
Yep, that's correct @mcmire. No breaking changes with this bump. We're keeping the retro-compatibility with the previous version for now.

The idea is to migrate v1 to v2 while keeping both alive for a short-period of time.

Once everything gets migrated, we might remove the `/v2` part and merge it in the main "exports".

For now, we're just "enabling" the `/v2` everywhere so we can do this integration!

packages/chain-agnostic-permission/src/scope/types.ts
mcmire: @ccharly I know you posted a comment about this, but given that [Stellar is present in `KnownCaipNamespace`](https://github.com/MetaMask/utils/blob/9c4af7e84571bdb12f430a6998f829bdd508e449/src/caip-types.ts#L134) and we are removing it from this type, this does seem like a breaking change we should note in the changelog. What do you think?

sahar-fehri (approved):

packages/chain-agnostic-permission/src/scope/types.ts
mcmire: I guess you could make the argument that since `chain-agnostic-permission` still depends on `@metamask/utils` `^11.9.0`, this is not a breaking change since there's no requirement that consumers _must_ use 11.11.0 (which includes the addition of Stellar to `KnownCaipNamespace`. So maybe this is fine.

mcmire (approved):
LGTM!

packages/chain-agnostic-permission/src/scope/types.ts
ccharly: Yep I know that's a bit tricky 😅 but this change should prevent us from treating new update of this enum as a breaking change in the future.

For now, Stellar is not supported yet on any clients. So we'll have to enable it once we're ready.

PR #8471, opened 2026-04-15
feat(wallet): Non-EVM support

FrederikBolding:
## Checklist


packages/wallet/src/initialization/instances/keyring-controller.ts
```diff
@@ -148,15 +169,40 @@ export const keyringController: InitializationConfiguration<
...
-  messenger: (parent) =>
-    new Messenger<'KeyringController', never, never, typeof parent>({
+  messenger: (parent) => {
+    const controllerMessenger: KeyringControllerMessenger = new Messenger({
       namespace: 'KeyringController',
       parent,
-    }),
+    });
+
+    // TODO: We only need to delegate here for the SnapKeyring, decide if we wanna do that
```
rekmarks, line 194: I don't understand this comment.

FrederikBolding: Updated the comment. We have both `messenger` and `initMessenger` on the clients to provide additional messenger actions besides what the controller directly needs. Right now I am just using the same `messenger` for everything. Undecided what the best path forward is.

FrederikBolding: I've continued the pattern from the feature branch where we use `InitActions` but only have one shared messenger for both the instance and any hooks it needs

PR #8479, opened 2026-04-16
feat: stop polling when bridge history item is stale or has an invalid src hash

micaelae:
### Problem
Bridge txHistory persists invalid or premature src tx hashes (specially STX), and pending items could be polled indefinitely, spamming providers and leaving bad state after restarts.

### Solution

- Subscribe to TransactionController:transactionStatusUpdated instead of transactionFailed / transactionConfirmed, and update history item matching the tx id, action id, hash, batch, or approval

- Skips persisting STX src hashes on submit. Set the hash when the tx is confirmed, or backfill from TransactionController as needed

- If a transaction is deemed "too old", fetch the status once, then check if the tx has been mined on chain. If The src hash is nonexistent, stop polling and delete the history item to prevent re-polling on the next restart

## References

Fixes https://consensyssoftware.atlassian.net/browse/SWAPS-4380

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes bridge status polling/state management and switches to `TransactionController:transactionStatusUpdated`, which can affect analytics, history reconciliation, and when polling starts/stops. Also deletes old pending history entries based on a remote-configured age threshold, so misconfiguration or matching bugs could drop legitimate items.
> 
> **Overview**
> Prevents indefinite bridge status polling by introducing an age threshold (`maxPendingHistoryItemAgeMs`, default 2 days) sourced from remote feature flags and, for stale pending items, fetching status once then checking `eth_getTransactionReceipt`; if the src hash appears invalid, polling is stopped, a new `invalid_transaction_hash` polling reason is tracked, and the history entry is removed to avoid re-polling on restart.
> 
> Refactors BridgeStatusController transaction subscriptions to use `TransactionController:transactionStatusUpdated` (replacing `transactionFailed`/`transactionConfirmed`), adds more robust history matching (by `id`, `actionId`, `batchId`, `txHash`, and `approvalTxId`), and avoids persisting smart-transaction src hashes until a finalized status is observed; metrics generation is updated accordingly (centralized polling-status properties and improved `error_message` formatting).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbfbe65e3132b983efacbc66d412830cd96a1310. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


micaelae:
@metamaskbot publish-preview

micaelae:
@metamaskbot publish-preview

packages/bridge-status-controller/src/utils/bridge-status.ts
```diff
@@ -112,6 +115,54 @@ export const shouldSkipFetchDueToFetchFailures = (
...
+ * @param messenger - The messenger to use to get the transaction meta by hash or id
+ * @param historyItem - The history item to check
+ *
+ * @returns true if the src tx hash is valid or we should still wait for it, false otherwise
+ */
+export const shouldWaitForFinalBridgeStatus = async (
+  messenger: BridgeStatusControllerMessenger,
+  historyItem: BridgeHistoryItem,
+): Promise<boolean> => {
+  if (isNonEvmChainId(historyItem.quote.srcChainId)) {
```
infiniteflower, line 147: Just curious why don't we wait for final bridge status for non evm? Is it just not possible with the way snaps work?

packages/bridge-status-controller/src/utils/history.ts
```diff
@@ -119,7 +174,11 @@ export const getInitialHistoryItem = (
       status: StatusTypes.PENDING,
       srcChain: {
         chainId: quoteResponse.quote.srcChainId,
-        txHash: bridgeTxMeta?.hash,
+        // STX transactions don't have a final tx hash until they are mined, so we don't set it initially
```
infiniteflower: I know it's not the same as being confirmed, but STX does return a tx hash immediately now though, should we mention that in the comments?

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -606,43 +646,72 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
+      // Skip if the history item is finalized or not old
+      !(
+        this.state.txHistory[bridgeTxMetaId] &&
+        isHistoryItemTooOld(
+          this.messenger,
+          this.state.txHistory[bridgeTxMetaId],
+        ) &&
+        [StatusTypes.PENDING, StatusTypes.UNKNOWN].includes(
+          this.state.txHistory[bridgeTxMetaId].status.status,
+        )
```
infiniteflower: Maybe this can be extracted to its own helper function or be broken down into variables?

packages/bridge-status-controller/src/utils/history.ts
micaelae: That's right - updated the comment to indicate why we don't save the initial hash

packages/bridge-status-controller/src/utils/bridge-status.ts
micaelae: We don't have a way to verify that non-EVM hashes exist on chain because snaps don't expose that functionality. So if a transaction is pending for too long we just remove it from txHistory to stop polling, then rely on the transaction status provided by the snap

packages/bridge-status-controller/src/bridge-status-controller.ts
micaelae: Good call! moved it to the `shouldWaitForFinalBridgeStatus` util

infiniteflower (approved):

PR #8480, opened 2026-04-16
feat(wallet): Add split persistence backed by better-sqlite3

rekmarks:
## Summary
- Add synchronous, per-property SQLite persistence to the wallet package using better-sqlite3
- Each controller state property with `persist: true` metadata gets its own row in a `kv` table (key format: `ControllerName.propertyName`)
- Writes happen synchronously within the same call stack as `controller.update()` via `stateChanged` event subscriptions, eliminating data loss windows
- Defaults to `:memory:` when no database path is provided

## Test plan
- [ ] `yarn workspace @metamask/wallet exec jest --no-coverage --watchman=false src/persistence/` — 22 unit tests covering KeyValueStore CRUD, loadState grouping, persist filtering, StateDeriver application, patch-based diffing, unsubscribe, and multi-controller scenarios
- [ ] `yarn workspace @metamask/wallet exec tsc --noEmit` — no new type errors
- [ ] Integration test with file-backed DB: create Wallet with a file path, perform operations, create second Wallet from same path, verify state restoration

---

> [!NOTE]
> **Medium Risk**
> Introduces a new persistence layer and changes wallet lifecycle/messenger semantics, plus adds a native dependency (`better-sqlite3`) that can affect build/install reliability across environments.
> 
> **Overview**
> Adds a new `persistence` subpath export that provides synchronous SQLite-backed state storage via `KeyValueStore`, plus `loadState` (reconstruct controller state from `Controller.prop` keys) and `subscribeToChanges` (persist only `persist`-flagged controller properties based on Immer patches and delete on removal).
> 
> Updates `Wallet` construction/typing to accept optional preloaded `state`, exposes `controllerMetadata` for initialized controllers, switches messenger namespace to `Wallet`, and makes `destroy()` idempotent while publishing a new `Wallet:destroyed` event after best-effort controller teardown.
> 
> Wires in the native `better-sqlite3` dependency (with LavaMoat allow-scripts), adds test/CI setup to install required binaries (`anvil` + `better-sqlite3` prebuild) and documents how to rebuild the native addon.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 772c1e63ec2f25201332e28f3b8772812e7dd481. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rekmarks:
@SocketSecurity ignore npm/better-sqlite3@11.10.0

Yes, it has native binaries.

rekmarks:
@SocketSecurity ignore-all

All alerts are due to `better-sqlite3` or its transitives. It is a reputable package that we use elsewhere.

packages/wallet/src/Wallet.ts
```diff
@@ -22,12 +22,30 @@ export class Wallet {
 
   readonly #instances: DefaultInstances;
 
-  constructor({ state = {}, options }: WalletConstructorArgs) {
+  readonly #store: KeyValueStore;
+
+  readonly #unsubscribePersistence: () => void;
+
+  constructor({ options, databasePath = ':memory:' }: WalletConstructorArgs) {
```
FrederikBolding: If `Wallet` is to be used in all clients it should be agnostic to persistence. I think these changes should target the CLI branch and be part of the CLI class.

packages/wallet/src/persistence/persistence.ts
```diff
@@ -0,0 +1,148 @@
...
+    const handler = (state: Record<string, Json>, patches: Patch[]): void => {
+      const changed = getChangedProperties(patches);
+
+      for (const prop of changed) {
+        if (!persistedProperties.has(prop)) {
+          continue;
+        }
+
+        const key = `${controllerName}.${prop}`;
+        const persistFlag = metadata[prop]?.persist;
```
FrederikBolding: We should use `deriveStateFromMetadata`

packages/wallet/src/Wallet.ts
FrederikBolding: Alternatively we can come up with an interface where each client can pass in their own persistence, but it may be simpler to not deal with it at all in `Wallet`.

packages/wallet/src/persistence/persistence.ts
rekmarks: They're similar but I don't think it's a good fit. `deriveStateFromMetadata` iterates over every top-level state property and errors on missing properties. The `subscribeToChanges` handlers iterate over changed properties and deletes missing properties from persistence, which shouldn't happen in practice but that is not for the persistence layer to know. Then there's also the complex error handling logic in `deriveStateFromMetadata` which doesn't buy us anything here.

packages/wallet/src/Wallet.ts
rekmarks: Made some benign modifications to the `Wallet` interface and made it agnostic of persistence. I kept the `persistence` files in the wallet package for now to avoid creating a complicated PR stack (or bloating the CLI PR). We can move the persistence stuff out of the wallet package once this is merged.

packages/wallet/src/persistence/persistence.ts
```diff
@@ -0,0 +1,192 @@
...
+      continue;
+    }
+
+    const eventType = `${controllerName}:stateChanged`;
+
+    const handler = (state: Record<string, Json>, patches: Patch[]): void => {
+      const changed = getChangedProperties(patches, persistedProperties);
+
+      for (const prop of changed) {
+        const key = storeKey(controllerName, prop);
```
FrederikBolding, line 100: The clients do persistence using KV where the key is the controller name. Do we think it is preferable to do at the property level? If so, why?

packages/wallet/src/persistence/persistence.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @param patches - Immer patches from a state update.
+ * @param persistedProperties - The set of persist-flagged property names.
+ * @returns A set of top-level property names that were modified.
+ */
+function getChangedProperties(
+  patches: Patch[],
+  persistedProperties: ReadonlySet<string>,
+): ReadonlySet<string> {
+  const changed = new Set<string>();
+  for (const patch of patches) {
```
FrederikBolding, line 181: Generally we try to not inspect the patches as it has proven error-prone, it may be fine in this case, but flagging regardless.

packages/wallet/src/Wallet.ts
```diff
@@ -40,17 +52,30 @@ export class Wallet {
...
-          return instance.destroy();
+          return (async (): Promise<void> => await instance.destroy())();
         }
         /* istanbul ignore next */
-        return undefined;
+        return Promise.resolve();
       }),
     );
+
+    this.messenger.publish('Root:walletDestroyed');
```
FrederikBolding: Maybe we should reconsider the root messenger namespace, perhaps it should be `Wallet:` instead?

packages/wallet/src/Wallet.test.ts
```diff
@@ -169,4 +167,90 @@ describe('Wallet', () => {
...
+
+      const names = Object.keys(wallet.controllerMetadata);
+      expect(names).toStrictEqual(Object.keys(wallet.state));
+      for (const name of names) {
+        expect(wallet.controllerMetadata[name]).toBeDefined();
+      }
+    });
+
+    it('omits instances without a metadata property from controllerMetadata', () => {
+      const fakeMetadata = { foo: { persist: true, anonymous: false } };
```
FrederikBolding:
```suggestion
      const fakeMetadata = { foo: { persist: true, includeInDebugSnapshot: false } };
```

packages/wallet/src/Wallet.ts
```diff
@@ -40,17 +52,30 @@ export class Wallet {
...
+    }
+    this.#destroyed = true;
+
+    await Promise.allSettled(
       Object.values(this.#instances).map((instance) => {
         // @ts-expect-error Accessing protected property.
         if (typeof instance.destroy === 'function') {
           // @ts-expect-error Accessing protected property.
-          return instance.destroy();
+          return (async (): Promise<void> => await instance.destroy())();
```
FrederikBolding: Maybe simpler to make the map function async?

packages/wallet/src/persistence/persistence.ts
rekmarks: It's faster. SQLite stores a text representation of the JSON values under the hood. Every time we modify a value stored at a given key in the state table, we overwrite the value of that key with `JSON.stringify(newValue)` (see `KeyValueStore.ts`). Consequently, by splitting controller state across multiple keys, there is less persistence overhead. There are ways to make this vastly more efficient, but this should be fine for know.

I don't know what the tradeoffs look like for this in the browser or React Native.

packages/wallet/src/persistence/persistence.ts
rekmarks: I hear that, although in this case we're not deeply inspecting the patches but relying on a couple of invariants:
- If the patch path length is 0, the entire value has been replaced
- The first element of the path array is the name of the first modified property

FrederikBolding (approved):

PR #8486, opened 2026-04-16
feat!: Replace ESLint with direct call to Prettier or Oxfmt in messenger CLI

Mrtenz:
## Explanation

Rather than calling ESLint, which was only really needed for formatting anyway, we can directly call Prettier for formatting the generated method action types files. This also means we no longer need to write a temporary file to disk for comparisons, the content is now formatted in-memory. In addition, I've added a `--formatter` option, which can be set to either "oxfmt" or "prettier" (default). While they should produce equal content, it avoids a hard dependency on Prettier.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes the generation/checking pipeline and output formatting behavior (and makes formatting async), which can cause diffs in generated files and new runtime dependency expectations on consumer projects.
> 
> **Overview**
> Switches `@metamask/messenger-cli` generated file formatting from an ESLint-based workflow to direct in-memory formatting via **Prettier (default)** or **Oxfmt**, selected through a new `--formatter` CLI option.
> 
> `generateActionTypesContent` is now async and always formats output; `--check` compares formatted content without temp files, and `--generate` no longer runs ESLint or returns a success boolean. Dependencies/docs/changelog are updated accordingly (drops ESLint as a peer dep; adds `prettier`/`oxfmt` peers), and tests/snapshots are updated to the new formatted output and API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd8ef95ca98518f31a120e2dd0b6ca1b91428edc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/messenger-cli/src/generate-content.ts
```diff
@@ -1,14 +1,109 @@
...
+ * @param contents - The source code to format.
+ * @returns The formatted source code.
+ */
+async function oxfmt(contents: string): Promise<string> {
+  try {
+    const { format } = await import('oxfmt');
+    const result = await format(
+      'file.ts',
+      contents,
+      DEFAULT_FORMATTING_OPTIONS,
```
Mrtenz: Oxfmt doesn't have a `resolveConfig` function, so I'm just assuming all consumers of the CLI using Oxfmt use our default options.

packages/messenger-cli/src/check.ts
```diff
@@ -3,103 +3,73 @@ import * as path from 'node:path';
...
   let hasErrors = false;
 
   const fileComparisonJobs: {
-    expectedTempFile: string;
+    expectedContent: string;
     actualFile: string;
     baseFileName: string;
   }[] = [];
 
-  try {
```
Mrtenz, line 27: Most of these changes are just indentation. I removed the outer `try / finally` block, since we no longer need to write a temporary file to disk.

packages/messenger-cli/src/generate-content.ts
```diff
@@ -1,14 +1,108 @@
...
+    return await format(contents, {
+      parser: 'typescript',
+      ...config,
+    });
+  } catch (error) {
+    const message = getErrorMessage(error);
+    throw new Error(
+      `Failed to format source code with Prettier. Is Prettier installed?\n\n${message}`,
+    );
+  }
```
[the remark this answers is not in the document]
Mrtenz: The original error is included, so I don't think this needs a separate try / catch.

packages/messenger-cli/jest.config.js
```diff
@@ -20,10 +20,10 @@ module.exports = merge(baseConfig, {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 100,
+      branches: 97.33,
```
mcmire, line 23: :( Is there a way to bring this back up to 100%?

Mrtenz: The only branches that are not covered are the error handling blocks in the `oxfmt` and `prettier` functions. I couldn't find a good way to test this, since Oxfmt and Prettier are dynamically imported, and Jest mocks doesn't seem to affect that.

mcmire: Oh right, I see, makes sense.

mcmire (approved):
LGTM!

PR #8491, opened 2026-04-16
refactor(multichain-account-service)!: replace `withKeyring` usage with `withKeyringV2`

hmalik88:
### Changes

**Base provider & types**

- Add `KeyringControllerWithKeyringV2Action` to the messenger's AllowedActions
- Split `BaseBip44AccountProvider.withKeyring` into two helpers:
- `withKeyring`; calls `KeyringController:withKeyring` (V1). Used by `SnapAccountProvider` until its V2 migration lands.
- `withKeyringV2<SelectedKeyring extends KeyringV2>`; calls `KeyringController:withKeyringV2`. Used by the EVM provider.

**EvmAccountProvider**

- All call sites moved to `withKeyringV2`
- Callbacks use V2 keyring methods: `getAccounts()` returns `KeyringAccount[]`, `createAccounts({ type: 'bip44:derive-index', entropySource, groupIndex })` replaces `addAccounts()`
- Range creation loops with `bip44:derive-index` since `HdKeyringV2` does not support `bip44:derive-index-range`
- `#getAddressFromGroupIndex` is now typed against `HdKeyring` from `@metamask/eth-hd-keyring/v2` and peeks via `keyring.root.deriveChild(groupIndex)` (the V2 wrapper forwards the legacy root getter)
- `discoverAccounts` keeps the V1-style peek-then-create pattern: peek the address → check on-chain activity outside the keyring mutex → create the account only if there is activity → look up the `InternalAccount` from `AccountsController` after the `withKeyringV2` callback completes
- `KeyringCapabilities` and `Keyring` types now come from `@metamask/keyring-api/v2`

## References

N/a

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the EVM account-creation and discovery path to use the V2 keyring interface and different keyring methods; regressions could affect account derivation/indexing and discovery behavior.
> 
> **Overview**
> **BREAKING:** Migrates the EVM account provider from `KeyringController:withKeyring` (V1) to `KeyringController:withKeyringV2` (V2) and updates types/messenger wiring to allow the new action.
> 
> `EvmAccountProvider` now uses V2 keyring APIs (`getAccounts` returning `KeyringAccount[]`, `createAccounts` replacing `addAccounts`, and deriving IDs from returned accounts) and changes range creation to loop per index (since `HdKeyringV2` doesn’t support range creation). Discovery now “peeks” the derived address via the V2 HD keyring `root` before creating the account, and adds explicit assertions/error handling when keyring creation returns no account.
> 
> Tests were updated accordingly, including new V2 keyring mocks and deterministic derivation fixtures, and the changelog documents the breaking migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdf3ac56cab9507c8f4dcf221fa517b3b9b9a973. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -385,39 +363,52 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
-        const account = this.messenger.call(
-          'AccountsController:getAccount',
-          accoundId,
+            const account = this.messenger.call(
+              'AccountsController:getAccount',
+              accountId,
+            );
+            assertInternalAccountExists(account);
+            assertIsBip44Account(account);
+            this.accounts.add(account.id);
```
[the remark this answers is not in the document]
hmalik88: [44bf6db](https://github.com/MetaMask/core/pull/8491/commits/44bf6dbb589db0bdfd8135f00fee952b99be937c)

packages/multichain-account-service/src/providers/BaseBip44AccountProvider.ts
```diff
@@ -166,7 +166,7 @@ export abstract class BaseBip44AccountProvider<
     }) => Promise<CallbackResult>,
   ): Promise<CallbackResult> {
     const result = await this.messenger.call(
-      'KeyringController:withKeyring',
+      'KeyringController:withKeyringV2',
```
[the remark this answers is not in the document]
hmalik88: [462990e](https://github.com/MetaMask/core/pull/8491/commits/462990e23bf0a52470d5b47eac89e193c6ef7407)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -385,34 +363,59 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+              if (created) {
+                await keyring.deleteAccount(created.id);
+              }
+              return null;
+            }
+
+            // Activity found — account stays. Return the address so we can
+            // look it up in the AccountsController after the callback persists.
+            return address;
+          },
```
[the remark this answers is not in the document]
hmalik88: The prior implementation peeked at addresses via `keyring.root` (V1 internal state) outside the lock. V2 keyrings don't expose HD internals, so we create the account inside the callback to discover the address, and delete it if inactive. Both must happen in the same callback so that `#persistOrRollback` sees zero net state change, as a result this holds the lock during the rpc call

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Replace `KeyringController:withKeyring` with `KeyringController:withKeyringV2` across all account providers ([#8491](https://github.com/MetaMask/core/pull/8491))
```
```suggestion
- **BREAKING:** Replace `KeyringController:withKeyring` with `KeyringController:withKeyringV2` for the EVM account provider ([#8491](https://github.com/MetaMask/core/pull/8491))
```

packages/multichain-account-service/src/providers/BaseBip44AccountProvider.ts
```diff
@@ -156,7 +156,7 @@ export abstract class BaseBip44AccountProvider<
   }
 
   protected async withKeyring<SelectedKeyring, CallbackResult = void>(
-    selector: KeyringSelector,
+    selector: KeyringSelectorV2,
```
ccharly: I wonder if we should have introduced a `withKeyringV2` here too? This way we can add a constraint with `SelectedKeyring extends KeyringV2`?

The old one would still remain and be used by the `SnapAccountProvider` for now until we have proper support for `SnapKeyringV2`.

WDYT?

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -34,90 +31,87 @@ import {
...
-
-function mockNextDiscoveryAddress(address: string): void {
-  jest.mocked(publicToAddress).mockReturnValue(createBytes(address as Hex));
+    id: account.id,
+    address: account.address,
+    type: account.type,
+    options: account.options,
+    methods: account.methods,
+    scopes: account.scopes ?? [],
+  } as unknown as KeyringAccount;
```
ccharly: An alternative without casting would be to use:
```ts
  const {
    metadata, // `InternalAccount` is just a `KeyringAccount` + `metadata.
    ...keyringAccount;
  };
  
  return keyringAccount;
```

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -34,90 +31,87 @@ import {
...
+  readonly capabilities = {
+    scopes: [EthScope.Eoa],
+    bip44: { deriveIndex: true },
   };
-}
 
-class MockEthKeyring implements EthKeyring {
-  readonly type = 'MockEthKeyring';
+  // Internal test-only state — not part of the Keyring interface.
+  readonly accounts: InternalAccount[];
```
ccharly: Then we could just use `KeyringAccount` right away no?

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -34,90 +31,87 @@ import {
...
+      throw new Error(`Account not found: ${accountId}`);
+    }
+    return toKeyringAccount(account);
+  });
 
-  addAccounts = jest.fn().mockImplementation((numberOfAccounts: number) => {
-    const newAccountsIndex = this.accounts.length;
+  createAccounts = jest
+    .fn()
+    .mockImplementation((options: Record<string, unknown>) => {
```
ccharly: We can properly type this with `CreateAccountOptions` (from the `keyring-api`)

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -572,6 +561,9 @@ describe('EvmAccountProvider', () => {
     ).toStrictEqual([]);
 
     expect(provider.getAccounts()).toStrictEqual([]);
+    // The account was created to peek at the address, then deleted
+    // because there was no on-chain activity.
+    expect(keyring.deleteAccount).toHaveBeenCalledTimes(1);
```
ccharly: Oh, I completely forgot about this, and actually we might have to tweak this a bit...

We don't want to create an account + remove it after, this cause some weird glitch where a group would be created out of this account and be removed 1s after...

So maybe we need to re-export the same getters than the `HdKeyring` v1 to be able to derive the address ourselves for now 🤔 (or add a `peekAddress` inside the v2 wrapper for now.

NOTE: We might need to re-export `root` and other getters, cause I think we use them on the clients for the Snap platform 😅

WDYT?

packages/multichain-account-service/CHANGELOG.md
hmalik88: [0578780](https://github.com/MetaMask/core/pull/8491/commits/0578780dc4a4435aa98eef2c1ac0df21db24be86)

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
hmalik88: [df99513](https://github.com/MetaMask/core/pull/8491/commits/df99513557a50bdffdee3e44e1ea3517471f706a)

packages/multichain-account-service/src/providers/BaseBip44AccountProvider.ts
hmalik88: [f53848d](https://github.com/MetaMask/core/pull/8491/commits/f53848d768d05b6bf10694a30f873463cb222f4c)

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
hmalik88: [60de56a](https://github.com/MetaMask/core/pull/8491/commits/60de56a9ba50623906db18219cda67870c202ee2)

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
hmalik88: [60de56a](https://github.com/MetaMask/core/pull/8491/commits/60de56a9ba50623906db18219cda67870c202ee2)

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
hmalik88: [60de56a](https://github.com/MetaMask/core/pull/8491/commits/60de56a9ba50623906db18219cda67870c202ee2)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -224,21 +225,23 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+            let groupIndex = Math.max(range.from, existing.length);
+            groupIndex <= range.to;
+            groupIndex++
+          ) {
+            const [created] = await keyring.createAccounts({
+              type: AccountCreationType.Bip44DeriveIndex,
+              entropySource,
+              groupIndex,
+            });
+            result.push(this.#getAccountId(created.address));
```
ccharly: 1. This is slightly unsafe, if for some reason, `createAccounts` returns `[]`, the `create.address` will throw! We should probably check for it to be "defined" before we use it. For now, we could log a warning + skip it if it's not defined?

2. Since `created` is an account, we can use `created.id` instead of `#getAccountId`) no?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -402,17 +418,17 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
         const [address] = await this.#createAccount({
```
ccharly: Maybe `#createAccount` should now return a proper `KeyringAccount` object? So we can use the `id` directly too?

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -1537,7 +1537,7 @@ describe('MultichainAccountService', () => {
         );
 
         rootMessenger.registerActionHandler(
-          'KeyringController:withKeyring',
+          'KeyringController:withKeyringV2',
           async (_, operation) => {
             const newKeyring = mocks.KeyringController.keyrings.find(
               (keyring) => keyring.type === 'HD Key Tree',
```
ccharly: Maybe it's time to also change the `'HD Key Tree'`, new keyring v2 are using `KeyringType` enum from the `keyring-api/v2`. Also, we should probably stop casting to `EthKeyring` as well (below)?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -262,17 +269,15 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
       groupIndex,
       throwOnGap: true,
     });
 
-    const accountId = this.#getAccountId(address);
-
     const account = this.messenger.call(
       'AccountsController:getAccount',
-      accountId,
+      created?.id,
```
ccharly: That's unsafe, and I'm even surprised the compiler is not failing on that 😅 `:getAccount` cannot be called with `undefined` here!

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -262,17 +269,15 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
     // Handle Bip44DeriveIndex (single account creation).
     const { groupIndex } = options;
 
-    const [address] = await this.#createAccount({
+    const [created] = await this.#createAccount({
```
ccharly, line 253: We should use the same logic than what we used for range. If somehow, we cannot create the account, we log something and skip it, WDYT?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -399,20 +417,19 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
         // We have some activity on this address, we try to create the account.
-        const [address] = await this.#createAccount({
+        const [created] = await this.#createAccount({
           entropySource,
           groupIndex,
+          throwOnGap: false,
         });
         assert(
-          addressFromGroupIndex === address,
+          addressFromGroupIndex === created.address,
```
ccharly, line 411: That's also unsafe, if somehow `created` is `undefined`, this will throw.

Not sure we should skip it here 🤔 maybe we can use `assert(created, "Account could not be created")` first, so we make sure `asserts created`?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -224,21 +223,29 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
           ) {
             if (groupIndex < existing.length) {
               // Account already exists.
-              result.push(this.#getAccountId(existing[groupIndex]));
+              result.push(this.#getAccountId(existing[groupIndex].address));
```
ccharly: We should also be using `existing[groupIndex].id` here, so we can get rid of `#getAccountId` entirely (like we can remove it)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: honestly, I'm surprised too haha I was half expecting it to complain but then the fact that it didn't made me think that the function would be able to handle undefined and didn't think twice.

packages/multichain-account-service/src/providers/EvmAccountProvider.test.ts
```diff
@@ -34,90 +38,115 @@ import {
...
+    .mockImplementation((options: CreateAccountOptions) => {
+      const newAccounts: KeyringAccount[] = [];
+
+      if (options.type === `${AccountCreationType.Bip44DeriveIndex}`) {
+        const groupIndex = this.accounts.length;
+        const { metadata, ...keyringAccount } =
+          makeDerivedHdAccount(groupIndex);
+        this.accounts.push(keyringAccount);
+        newAccounts.push(keyringAccount);
+      }
```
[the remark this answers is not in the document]
hmalik88: [620381d](https://github.com/MetaMask/core/pull/8491/commits/620381df727bff0768c245b1e1dc05a150208c78)

ccharly (approved):

PR #8504, opened 2026-04-17
feat(profile-sync-controller): add profile pairing & canonical profile id management

mathieuartu:
## References

- https://consensyssoftware.atlassian.net/browse/MUL-1722

### Client PRs

- [Extension](https://github.com/MetaMask/metamask-extension/pull/42120)
- [Mobile](https://github.com/MetaMask/metamask-mobile/pull/29357)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches authentication and user-storage request scoping, adding new network calls and state fields (`canonicalProfileId`) that can affect login/session validity and storage partitioning if incorrect or mismatched with backend behavior.
> 
> **Overview**
> Adds SRP *profile pairing* support by calling `POST /api/v2/profile/pair` after `performSignIn` when 2+ SRPs exist, propagating a new `canonicalProfileId` across cached SRP sessions and emitting a new `AuthenticationController:profileSignIn` event when aliases are returned or the canonical changes.
> 
> Extends the SRP auth flow to request pairing (`X-MetaMask-Profile-Pairing: enabled` on `/srp/login`), parse and surface `profile_aliases`, and resolve the original per-SRP `profileId` from aliases (via new `computeIdentifierId`) while storing `canonicalProfileId` separately; sessions missing `canonicalProfileId` now force re-login.
> 
> Updates `UserStorage` to optionally scope requests by sending `x-profile-id` when `profileId !== canonicalProfileId` (with an escape hatch `useCanonicalScope`), adds `refreshCanonicalProfileId` to force a fresh canonical fetch, and introduces validation/utilities (`validatePairResponse`) plus expanded fixtures/tests for the new pairing behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abc619c92c04169d5962fe7b6e7a69654ff4e142. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/flow-srp.ts
```diff
@@ -185,6 +195,11 @@ export class SRPJwtBearerAuth implements IBaseAuth {
       return null;
     }
 
+    // get canonical profile id from server if not present in the cached session
+    if (!auth.profile.canonicalProfileId) {
+      return null;
+    }
```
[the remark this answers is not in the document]
mathieuartu: SiWE isn't used, doesn't participate in pairing, and doesn't go through the SRP login endpoint.

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/src/sdk/user-storage.ts
```diff
@@ -343,6 +366,10 @@ export class UserStorage {
     const entropySourceId = options?.entropySourceId;
     try {
       const headers = await this.#getAuthorizationHeader(entropySourceId);
+      const profileScopeHeader = await this.#getProfileScopeHeader(
```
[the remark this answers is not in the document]
mathieuartu: This is fine, the ACL covers us for those cases. We can read / delete anything owned by the aliases as long as we have the `sha(storageKey + profileId)`. GET ALL and both PUT and PUT ALL were a different beast.

mathieuartu:
@metamaskbot publish-preview

mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/services.ts
```diff
@@ -230,6 +251,63 @@ export async function pairIdentifiers(
...
+    const pairResponse = await response.json();
+
+    return {
+      profile: {
+        identifierId: pairResponse.profile.identifier_id,
+        metaMetricsId: pairResponse.profile.metametrics_id ?? '',
+        profileId: pairResponse.profile.profile_id,
+        canonicalProfileId: pairResponse.profile.profile_id,
+      },
+      profileAliases: parseProfileAliases(pairResponse.profile_aliases ?? []),
```
gantunesr, line 306: Should there be a struct to validate this object shape?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -295,9 +310,72 @@ export class AuthenticationController extends BaseController<
...
+          this.messenger.publish('AuthenticationController:profileSignIn', {
+            profileId: newCanonical ?? '',
+            profileAliases,
+            profileIdChanged,
+          });
+        }
+      } catch {
+        // Pairing failure is non-fatal — retry on next performSignIn
+      }
+    }
```
gantunesr, line 316: Should this be its own pairing function?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -337,6 +424,50 @@ export class AuthenticationController extends BaseController<
...
+
+    this.#propagateCanonical(canonical);
+    return canonical;
+  }
+
+  #invalidateSrpSession(entropySourceId: string): void {
+    this.update((state) => {
+      const entry = state.srpSessionData?.[entropySourceId];
+      if (entry?.profile) {
+        entry.profile.canonicalProfileId = '';
```
gantunesr, line 478: So if the `canonicalProfileId` is an empty string, then the session is invalid. If the value is `undefined` or `null`, does it mean the session is also invalid?

mathieuartu: Yes because of a internal check in `performLogin` that verifies if `canonicalProfileId` exists. Added a comment in [fc4f039](https://github.com/MetaMask/core/pull/8504/commits/fc4f03978de2e3ffddfd58d712a9c9f6af1bbd3a)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Done in [fc4f039](https://github.com/MetaMask/core/pull/8504/commits/fc4f03978de2e3ffddfd58d712a9c9f6af1bbd3a)

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/services.ts
mathieuartu: Yes, good call! Did it the same form we did it in the past with `validate-login-response.ts` ([fc4f039](https://github.com/MetaMask/core/pull/8504/commits/fc4f03978de2e3ffddfd58d712a9c9f6af1bbd3a))

ccharly: Should `validatePairResponse` use `asserts input is SomeType` to have proper typing over it?

IDK if we did that for the others, but here we're still manipulating JSON objects and we could have typos and stuff 😅 I think `asserts ... is ...` would be a nice addition!

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/utils/identifier.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(result).toBe(expected);
+  });
+
+  it('produces SHA256(publicKey + salt) for prd environment', () => {
+    const result = computeIdentifierId(MOCK_PUBLIC_KEY, Env.PRD);
+    const expected = createSHA256Hash(
+      MOCK_PUBLIC_KEY + IDENTIFIER_SALT[Env.PRD],
+    );
+    expect(result).toBe(expected);
+  });
```
ccharly: Nit: but this can be done with `it.each` to avoid duplicating the test 3 times

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -337,6 +428,53 @@ export class AuthenticationController extends BaseController<
...
+   * from any single SRP login.
+   *
+   * @returns The refreshed canonical profile ID.
+   */
+  public async refreshCanonicalProfileId(): Promise<string> {
+    this.#assertIsUnlocked('refreshCanonicalProfileId');
+
+    const primaryId = await this.#getPrimaryEntropySourceId();
+    this.#invalidateSrpSession(primaryId);
+    await this.#auth.getAccessToken(primaryId);
```
gantunesr: is it ok if `getAccessToken` throws just after the session has been invalidated? I guess this is recoverable scenario

mathieuartu: The `!canonicalProfileId` guard in `#getAuthSession` ensures the next `getAccessToken` call triggers a fresh `#performLogin` regardless of what threw, so the worst outcome is that we get an error and the next request recovers.

packages/profile-sync-controller/src/sdk/user-storage.ts
```diff
@@ -575,4 +609,18 @@ export class UserStorage {
...
+  async #getProfileScopeHeader(
+    entropySourceId?: string,
+    useCanonicalScope?: boolean,
+  ): Promise<Record<string, string>> {
+    if (useCanonicalScope) {
+      return {};
+    }
+    const profile = await this.config.auth.getUserProfile(entropySourceId);
+    if (profile.profileId !== profile.canonicalProfileId) {
+      return { 'x-profile-id': profile.profileId };
```
ccharly, line 626: Maybe add a comment there to explain that we need to access user storage data scope by a `profileId` (scoped for a specific SRP) rather than the canonical one (which is not fully ready yet)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -295,9 +310,76 @@ export class AuthenticationController extends BaseController<
...
+      const profileAliases = await this.#pairSrpProfiles(accessTokens);
+
+      const newCanonical = this.#getCanonicalProfileId();
+      const profileIdChanged = previousCanonical !== newCanonical;
+      const shouldEmitProfileSignInEvent =
+        profileIdChanged || profileAliases.length > 0;
+
+      if (shouldEmitProfileSignInEvent) {
+        this.messenger.publish('AuthenticationController:profileSignIn', {
+          profileId: newCanonical ?? '',
```
ccharly: Given that `?? ''` is used to make the compiler happy, we should probably use this instead:

```suggestion
      if (shouldEmitProfileSignInEvent && newCanonical) {
        this.messenger.publish('AuthenticationController:profileSignIn', {
          profileId: newCanonical,
```

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.test.ts
```diff
@@ -193,6 +195,240 @@ describe('AuthenticationController', () => {
...
+      ]);
+
+      const controller = new AuthenticationController({
+        messenger,
+        metametrics,
+      });
+
+      await controller.performSignIn();
+
+      expect(controller.state.isSignedIn).toBe(true);
```
ccharly, line 255: Nit: But we're not really checking that we NOT calling the `mockPairProfiles` endpoint

Should we add this (and define a dummy `mockEndpoints.mockPairProfilesUrl`)?
```ts
expect(mockEndpoints.mockPairProfilesUrl.isDone()).toBe(false);
```

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -295,9 +310,76 @@ export class AuthenticationController extends BaseController<
...
+  }
+
+  async #pairSrpProfiles(accessTokens: string[]): Promise<ProfileAlias[]> {
+    if (accessTokens.length < 2) {
+      return [];
+    }
+    const {
+      profileAliases,
+      profile: { canonicalProfileId },
+    } = await this.#auth.pairSrpProfiles(accessTokens, accessTokens[0]);
```
ccharly: Nit: But just to emphasis on the "primary token" === "primary SRP", we could maybe do:

```suggestion
    const primaryAccessToken = accessToken[0]; // Associated with primary SRP.
    const {
      profileAliases,
      profile: { canonicalProfileId },
    } = await this.#auth.pairSrpProfiles(accessTokens, primaryAccessToken);
```

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -295,9 +310,76 @@ export class AuthenticationController extends BaseController<
...
+      return;
+    }
+
+    this.update((state) => {
+      for (const key of Object.keys(state.srpSessionData ?? {})) {
+        const entry = state.srpSessionData?.[key];
+        if (entry?.profile) {
+          entry.profile.canonicalProfileId = canonicalProfileId;
+        }
+      }
```
ccharly: Nit: We could use `Object.values` no?

```suggestion
      for (const entry of Object.values(state.srpSessionData ?? {})) {
        if (entry?.profile) {
          entry.profile.canonicalProfileId = canonicalProfileId;
        }
      }
```

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -295,9 +310,76 @@ export class AuthenticationController extends BaseController<
...
+
+  #getCanonicalProfileId(): string | null {
+    const { srpSessionData } = this.state;
+    if (!srpSessionData) {
+      return null;
+    }
+    const firstKey = Object.keys(srpSessionData)[0];
+    return firstKey
+      ? (srpSessionData[firstKey]?.profile?.canonicalProfileId ?? null)
+      : null;
```
ccharly: Nit: Same here?

```suggestion
    const primarySessionData = Object.values(srpSessionData)[0];
    return primarySessionData?.profile?.canonicalProfileId ?? null;
```

ccharly: WDYT about having an optional `entropySourceId` parameter so we can re-use this function can be used as a real getter for other SRPs too? (since the path to access the is quite complex, it's nice to read `this.#getCanonicalProfileId(entropySourceId)` instead).

Bonus point: if somehow we need to add more check when accessing it, all call points would benefit from that too

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -337,6 +428,53 @@ export class AuthenticationController extends BaseController<
...
+   */
+  public async refreshCanonicalProfileId(): Promise<string> {
+    this.#assertIsUnlocked('refreshCanonicalProfileId');
+
+    const primaryId = await this.#getPrimaryEntropySourceId();
+    this.#invalidateSrpSession(primaryId);
+    await this.#auth.getAccessToken(primaryId);
+
+    const canonical =
+      this.state.srpSessionData?.[primaryId]?.profile?.canonicalProfileId;
```
ccharly: Related to my other comment, we could re-use `#getCanonicalProfileId` here too

packages/profile-sync-controller/src/sdk/utils/validate-pair-response.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Validates that the raw API response from the profile pairing endpoint has
+ * the expected shape before we attempt to map it into a `PairProfilesResponse`.
+ *
+ * @param input - The raw JSON parsed from the pair endpoint response body.
+ * @returns true if the response has all required fields; false otherwise.
+ */
+export function validatePairResponse(input: unknown): boolean {
```
ccharly: Could probably use this pattern (if we have a type for it):

```suggestion
export function validatePairResponse(input: unknown): asserts input is PairReponse {
```

This would improve type-safety when using this function!

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/flow-srp.test.ts
```diff
@@ -209,4 +218,300 @@ describe('SRPJwtBearerAuth rate limit handling', () => {
...
+    });
+
+    return { auth, store };
+  };
+
+  beforeEach((): void => {
+    jest.clearAllMocks();
+    mockGetNonce.mockResolvedValue(MOCK_NONCE_RESPONSE);
+    mockAuthorizeOIDC.mockResolvedValue(MOCK_OIDC_RESPONSE);
+    mockComputeIdentifierId.mockReturnValue('computed-identifier-hash');
```
ccharly: Could you add a `MOCK_COMPUTED_IDENTIFIER_ID` by any chance? Since that's the value that's "driving" the logic in the test (I mean that's the value that's being tracked/filtered, so it's easier to identify the patterns)

packages/profile-sync-controller/src/sdk/authentication-jwt-bearer/services.test.ts
```diff
@@ -626,6 +718,135 @@ describe('services', () => {
...
+
+      expect(result.profile.profileId).toBe('canonical-1');
+      expect(result.profile.identifierId).toBe('id-canonical');
+      expect(result.profileAliases).toHaveLength(2);
+      expect(result.profileAliases[0]).toStrictEqual({
+        aliasProfileId: 'p1',
+        canonicalProfileId: 'canonical-1',
+        identifierIds: [{ id: 'h1', type: 'SRP' }],
+      });
+    });
```
ccharly, line 875: We could also check for `[1]` (with `'p2'`) no?

packages/profile-sync-controller/src/sdk/user-storage.test.ts
ccharly, line 1: Nit: Some of the tests setup could be more DRY!

ccharly (approved):
LGTM! 🚀 thanks for making those changes 💪

PR #8506, opened 2026-04-17
feat: Add `createMethodMiddleware` function

FrederikBolding:
## Explanation

This PR implements a `createMethodMiddleware` function which allows usage of **both** hooks and the messenger. Currently hooks are the only way JSON-RPC implementations can leverage controllers, services etc. They require more glue code to be written and are more difficult to type. This PR aims to provide a shared utility for constructing middlewares with multiple method implementations, while also allowing these implementations to use the messenger, cutting down the amount of glue code required in each client significantly.

## References

https://consensyssoftware.atlassian.net/browse/WPC-981

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces a new public middleware construction path and messenger delegation behavior, which could affect RPC handling/integration patterns for v2 consumers despite being additive.
> 
> **Overview**
> Adds a new `v2` utility, `createMethodMiddleware`, for wiring multiple JSON-RPC method handlers that can consume a *restricted subset* of provided hooks (`selectHooks`) and a per-method delegated `@metamask/messenger` instance.
> 
> Updates `@metamask/json-rpc-engine` to depend on `@metamask/messenger`, exports the new APIs from `v2/index`, adds coverage for hook/messenger delegation behavior, and updates build TS project references plus the changelog/README dependency graph accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a96f8823f26c19a4b19c722ca05ba3ac7feda998. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/json-rpc-engine/CHANGELOG.md
+++ b/packages/json-rpc-engine/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `createOriginMiddleware` utility ([#8522](https://github.com/MetaMask/core/pull/8522))
+- Add `createOriginMiddleware` utility to `v2` ([#8522](https://github.com/MetaMask/core/pull/8522))
+- Add `createMethodMiddleware` utility to `v2` ([#8506](https://github.com/MetaMask/core/pull/8506))
+  - This utility allows JSON-RPC method implementations to use both the hooks pattern and the messenger.
 
 ## [10.2.4]
 
--- a/packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
+++ b/packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
@@ -1,0 +1,48 @@
+import { Messenger, MOCK_ANY_NAMESPACE } from '@metamask/messenger';
+
+import { makeRequest } from '../../tests/utils';
+import {
+  createMethodMiddleware,
+  MethodHandler,
+} from './createMethodMiddleware';
+import { JsonRpcEngineV2 } from './JsonRpcEngineV2';
+
+const getValueA = {
+  hookNames: { testHook: true },
+  implementation: ({ hooks }) => hooks.testHook(),
+} satisfies MethodHandler;
+
+const getValueB = {
+  actionNames: ['Example:TestAction'],
+  implementation: ({ messenger }) => messenger.call('Example:TestAction'),
+} satisfies MethodHandler;
+
+const messenger = new Messenger({ namespace: MOCK_ANY_NAMESPACE });
+
+messenger.registerActionHandler('Example:TestAction', async () => 'B');
+
+const middleware = createMethodMiddleware({
+  handlers: { getValueA, getValueB },
+  hooks: { testHook: async () => 'A' },
+  messenger,
+});
+
+const engine = JsonRpcEngineV2.create({ middleware: [middleware] });
+
+describe('createMethodMiddleware', () => {
+  it('passes in the requested hooks', async () => {
+    const result = await engine.handle(makeRequest({ method: 'getValueA' }));
+    expect(result).toBe('A');
+  });
+
+  it('passes in a delegated messenger', async () => {
+    const result = await engine.handle(makeRequest({ method: 'getValueB' }));
+    expect(result).toBe('B');
+  });
+
+  it('skips unrecognized methods', async () => {
+    await expect(
+      engine.handle(makeRequest({ method: 'getValueC' })),
+    ).rejects.toThrow('Nothing ended request');
+  });
+});
--- a/packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
+++ b/packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
@@ -12,43 +12,53 @@ type TestAction = {
   handler: () => Promise<string>;
 };
 
-const getValueA = {
-  hookNames: { testHook: true },
-  implementation: ({ hooks }): Promise<string> => hooks.testHook(),
-} satisfies MethodHandler<{ testHook: () => Promise<string> }>;
+function setup(): { engine: JsonRpcEngineV2 } {
+  const getValueA = {
+    hookNames: { testHook: true },
+    implementation: ({ hooks }): Promise<string> => hooks.testHook(),
+  } satisfies MethodHandler<{ testHook: () => Promise<string> }>;
 
-const getValueB = {
-  actionNames: ['Example:TestAction'],
-  implementation: ({ messenger }): Promise<string> =>
-    messenger.call('Example:TestAction'),
-} satisfies MethodHandler<never, TestAction>;
+  const getValueB = {
+    actionNames: ['Example:TestAction'],
+    implementation: ({ messenger }): Promise<string> =>
+      messenger.call('Example:TestAction'),
+  } satisfies MethodHandler<never, TestAction>;
 
-const messenger = new Messenger<string, TestAction>({
-  namespace: MOCK_ANY_NAMESPACE,
-});
+  const messenger = new Messenger<string, TestAction>({
+    namespace: MOCK_ANY_NAMESPACE,
+  });
 
-messenger.registerActionHandler('Example:TestAction', async () => 'B');
+  messenger.registerActionHandler('Example:TestAction', async () => 'B');
 
-const middleware = createMethodMiddleware({
-  handlers: { getValueA, getValueB },
-  hooks: { testHook: async () => 'A' },
-  messenger,
-});
+  const middleware = createMethodMiddleware({
+    handlers: { getValueA, getValueB },
+    hooks: { testHook: async () => 'A' },
+    messenger,
+  });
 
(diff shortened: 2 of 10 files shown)


packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
```diff
@@ -0,0 +1,151 @@
...
+  Context extends ContextConstraint = ContextConstraint,
+> = {
+  implementation: (options: {
+    request: Readonly<JsonRpcRequest<Parameters>>;
+    context: Context;
+    next: Next<JsonRpcRequest>;
+    messenger: Messenger<string, MessengerActions>;
+    hooks: Hooks;
+  }) => Promise<Result> | Result;
+  hookNames?: { [Key in keyof Hooks]: true };
```
FrederikBolding: I'm wondering why this wasn't just a list of names in our current production implementations. Open to changing it here.

packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+messenger.registerActionHandler('Example:TestAction', async () => 'B');
+
+const middleware = createMethodMiddleware({
+  handlers: { getValueA, getValueB },
+  hooks: { testHook: async () => 'A' },
+  messenger,
+});
+
+const engine = JsonRpcEngineV2.create({ middleware: [middleware] });
```
mcmire: Should we be sharing values across tests like this? Just curious whether it would be less error-prone in the future to create them inside of each test, perhaps using a function to encapsulate the setup.

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
```diff
@@ -0,0 +1,165 @@
...
+ * @returns A JsonRpcEngineV2 middleware.
+ */
+export function createMethodMiddleware<
+  Handlers extends Record<string, AnyMethodHandler>,
+  Context extends ContextConstraint,
+>(
+  options: CreateMethodMiddlewareOptions<Handlers>,
+): JsonRpcMiddleware<JsonRpcRequest, Json, Context> {
+  const { messenger: rootMessenger } = options;
+  const allHooks = options.hooks as Record<string, unknown>;
```
mcmire, line 95: I was curious why we needed this type assertion. I asked Claude, and if I understand correctly, because we are using a conditional type in `HandlerHooks`, TypeScript doesn't have enough information to resolve `UnionToIntersection<HandlerHooks<Handlers[keyof Handlers]>>` to `Record<string, unknown>` at this point in time. I'm not sure what TypeScript would need or how else we could set up the types for `createMethodMiddleware`. But the suggestion I was given was that we could change `CreateMethodMiddlewareOptions` from this:

``` typescript
export type CreateMethodMiddlewareOptions<
  Handlers extends Record<string, AnyMethodHandler>,
> = {
  handlers: Handlers;
  messenger: Messenger<string, HandlerActions<Handlers[keyof Handlers]>>;
  hooks: UnionToIntersection<HandlerHooks<Handlers[keyof Handlers]>>;
};
```

to this:

```typescript
export type CreateMethodMiddlewareOptions<
  Handlers extends Record<string, AnyMethodHandler>,
> = {
  handlers: Handlers;
  messenger: Messenger<string, HandlerActions<Handlers[keyof Handlers]>>;
  hooks: UnionToIntersection<HandlerHooks<Handlers[keyof Handlers]>> &
    Record<string, unknown>;
};
```

and then here we can remove the type assertion:

```suggestion
  const allHooks = options.hooks;
```

My interpretation of this is we are telling TypeScript, "hey, just so you know, `hooks` is also a `Record<string, unknown>`, so treat it like that too".

What do you think?

packages/json-rpc-engine/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `createMethodMiddleware` ([#8506](https://github.com/MetaMask/core/pull/8506))
```
mcmire: Should we be explicit that this was added to the `v2` path? Something like:

```suggestion
- Add `createMethodMiddleware` to `@metamask/json-rpc-engine/v2` ([#8506](https://github.com/MetaMask/core/pull/8506))
```

or even just

```suggestion
- Add `createMethodMiddleware` to `v2` ([#8506](https://github.com/MetaMask/core/pull/8506))
```

mcmire (commented):
A few minor suggestions, but I tested this out and it seems to work pretty well.

packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
rekmarks: I don't think it's ever justified. Even if I know that the value is stateless, I still create a `makeX()` utility or do `let x; beforeEach(() => { x = ... });`.

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
```diff
@@ -0,0 +1,165 @@
...
+  Parameters extends JsonRpcParams = JsonRpcParams,
+  Result extends Json = Json,
+  Context extends ContextConstraint = ContextConstraint,
+> = {
+  implementation: (options: {
+    request: Readonly<JsonRpcRequest<Parameters>>;
+    context: Context;
+    next: Next<JsonRpcRequest>;
+    messenger: Messenger<string, MessengerActions>;
+    hooks: Hooks;
```
rekmarks, line 41: Should these be optional? Could they be optional?

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
rekmarks: This is because, at the time it was first implemented, you couldn't create an array type like `[Key in keyof Hooks]`. Idk if this has changed.

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
FrederikBolding: The tricky thing here is balancing the types used inside the function with how they are enforced when being passed in. If we let `hooks` be `UnionToIntersection<HandlerHooks<Handlers[keyof Handlers]>> &
    Record<string, unknown>` it will allow developers to specify unnecessary hooks and break the inference which enforces that each hook specified in the handlers is also specified in `hooks`.

The cast strictly exists because of the type of `selectHooks`.

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
FrederikBolding: Ah. That is still not possible AFAIK

packages/json-rpc-engine/src/v2/createMethodMiddleware.test.ts
FrederikBolding: https://github.com/MetaMask/core/pull/8506/commits/a8f3aa675913bc3e8e161b91cd854a59759dff44

packages/json-rpc-engine/CHANGELOG.md
FrederikBolding: https://github.com/MetaMask/core/pull/8506/commits/7f3e3bb6bbb298a8a7250b67b8ebd8a30db57e3f

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
FrederikBolding: Theoretically they could, but it makes the types way more complicated, we would need to have ternary types for each of them and only allow them to be undefined if no hooks or messenger actions are required for all handlers. Additionally we would need more typing that enforces that the entire hooks/messenger property is available for each of the implementation types.

Not sure it is worth the trouble at this point.

FrederikBolding: I needed to do this for `hookNames` and `actionNames`: https://github.com/MetaMask/core/pull/8506/commits/a96f8823f26c19a4b19c722ca05ba3ac7feda998

Theoretically we could continue doing this, but it gets hard to read quite fast 😓

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
mcmire: > If we let `hooks` be `UnionToIntersection<HandlerHooks<Handlers[keyof Handlers]>> & Record<string, unknown>` it will allow developers to specify unnecessary hooks and break the inference which enforces that each hook specified in the handlers is also specified in `hooks`.

Will this happen even when using `&`? But I understand. If this doesn't make sense then never mind :)

mcmire (approved):
LGTM!

packages/json-rpc-engine/src/v2/createMethodMiddleware.ts
FrederikBolding: Seems like it does, yeah 😄

rekmarks (approved):

PR #8513, opened 2026-04-17
feat(accounts-controller): add support for Snap keyring v2

ccharly:
## Explanation

Adding support for Snap keyring v2 for this controller. For now, those keyrings are not in use, but we will need this to be able to the accounts associated with those new keyrings.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches account synchronization logic and adds a new path for constructing internal accounts from Snap v2 keyrings; incorrect matching or metadata injection could cause missing/incorrect accounts during keyring state updates.
> 
> **Overview**
> Adds support for `SnapKeyring` v2 accounts so the `AccountsController` can materialize Snap v2 keyring addresses into internal accounts during `KeyringController:stateChange` and `updateAccounts`.
> 
> This introduces Snap v2 type detection (`KeyringType.Snap`), looks up accounts via `SnapKeyringV2.lookupByAddress` across per-snap keyring instances, and injects minimal `metadata` (including `metadata.snap.id`) when converting those results into internal accounts. Utilities and tests are updated to recognize/label the v2 keyring type and cover edge cases where the v2 keyring instance/account cannot be found, and the changelog documents the new support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c49b5a72f263b508495574a046fe3394f72542e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/accounts-controller/src/AccountsController.ts
```diff
@@ -887,32 +935,16 @@ export class AccountsController extends BaseController<
...
-      normal: generatePatch(),
-    };
-
-    // Gets the patch object based on the keyring type (since Snap accounts and other accounts
-    // are handled differently).
-    const patchOf = (type: string): StatePatch => {
-      if (isSnapKeyringType(type)) {
-        return patches.snap;
-      }
-      return patches.normal;
```
ccharly, line 909: Just removed all this since we don't need that separation at all!

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -847,6 +852,49 @@ export class AccountsController extends BaseController<
...
+              lastSelected: 0,
+              keyring: {
+                type: KeyringType.Snap,
+              },
+              snap: {
+                name: keyring.snapId,
+                enabled: true,
+                id: keyring.snapId,
+              },
+            },
```
ccharly, line 887: The new Snap keyring v2 only use `KeyringAccount`, thus we don't have any metadata for those!

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -847,6 +850,49 @@ export class AccountsController extends BaseController<
...
+              importTime: Date.now(),
+              lastSelected: 0,
+              keyring: {
+                type: KeyringType.Snap,
+              },
+              snap: {
+                name: keyring.snapId,
+                enabled: true,
+                id: keyring.snapId,
+              },
```
ccharly, line 886: We could query this from the `SnapController`, but that requires a new allowed action, thus a breaking change (but at least that would mimic what the current big `SnapKeyring` is doing)

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -847,6 +850,52 @@ export class AccountsController extends BaseController<
...
+      if (keyring instanceof SnapKeyringV2) {
+        // We use the synchronous method here since this method is used during `:stateChange` that are
+        // use synchronous handlers.
+        const account = keyring.lookupByAddress(address);
+        if (account) {
+          return {
+            ...account,
+            // We still have to use internal account for now, so we inject some metadata.
+            metadata: {
+              name: '',
```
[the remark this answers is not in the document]
ccharly: That's fine, we're in te process of deprecating account names on those objects anyway.

packages/accounts-controller/src/AccountsController.ts
ccharly: No longer needed, we removed `enabled` and `name` from those metadata

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -841,6 +844,66 @@ export class AccountsController extends BaseController<
...
+   */
+  #getAccountFromSnapKeyringV2(address: string): InternalAccount | undefined {
+    const keyrings = this.messenger.call(
+      'KeyringController:getKeyringsByType',
+      KeyringType.Snap,
+    );
+
+    // Snap keyring v2 are "per-Snaps", so we need to iterate over all of them to find the account.
+    for (const keyring of keyrings) {
+      if (keyring instanceof SnapKeyringV2) {
```
gantunesr, line 880: This seems inefficient, can we keep a map in memory for faster lookups?

gantunesr: maybe thats an overkill given that there won't be hundreds of snap keyrings

gantunesr (approved):
lgtm

packages/accounts-controller/src/AccountsController.ts
ccharly: No actually I added that a safe-guard (with a small runtime cost yep) but `:getKeyringsByType` should """guarantee""" us that we get `SnapKeyringV2` instances.

Also, I couldn't use `:withKeyring*` here because this code has to be synchronous 😬 Ultimately, account storage will get delegated to the `KeyringController` that will keep all `KeyringAccount`s in memory (for synchronous access).

We won't need that trick anymore once everything will be migrated to v2!

PR #8527, opened 2026-04-20
fix(transaction-pay-controller): ignore synthetic across gas legs

pedronfigueiredo:
## Problem observed in testing

The perps direct deposit flow could include a synthetic gas-funding leg, generated internally for native ETH on Arbitrum, with targetAmountMinimum: '0'. That leg was optional and not the real asset the user was trying to deposit, but the Across strategy support check still evaluated it as if it were a required quote leg. As a result, a valid perps deposit request could be marked unsupported before quote resolution, because the optional gas leg made the whole request set fail strategy validation.

The fix was to make Across treat those zero-minimum synthetic gas legs consistently everywhere. A shared helper was introduced to identify “real” Across quote requests, and both quote fetching and strategy support now filter out requests where targetAmountMinimum === '0' unless it is a max-amount flow. In practice, that means the real deposit leg can proceed while the optional gas top-up leg no longer disqualifies the route. The turn also added a regression test covering exactly that case: one real perps deposit leg plus one synthetic native-token gas leg, and the strategy now correctly returns supported.

## Summary
- ignore synthetic zero-target gas legs when evaluating Across support
- share the same request filter between Across support checks and quote building
- add coverage for perps direct deposits that include a synthetic gas leg

## Testing
- yarn jest --config packages/transaction-pay-controller/jest.config.js --watchman=false packages/transaction-pay-controller/src/strategy/across/AcrossStrategy.test.ts --coverage=false

---

> [!NOTE]
> **Medium Risk**
> Changes Across strategy request filtering, which affects whether routes are considered supported and which legs are quoted; mistakes could cause valid quotes to be skipped or unsupported routes to be attempted.
> 
> **Overview**
> Fixes Across strategy validation/quoting to **ignore synthetic “gas” legs** (requests with `targetAmountMinimum: '0'`) so they don’t cause perps direct deposits to be marked unsupported.
> 
> Introduces a shared `isAcrossQuoteRequest` helper and applies it consistently in both `AcrossStrategy.supports` and `getAcrossQuotes`, and adds regression tests covering mixed real+synthetic legs, all-synthetic legs, and `isMaxAmount` zero-minimum behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d773fd302489d4734eceac39bb012f771c9d816c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/src/strategy/across/requests.ts
+++ b/packages/transaction-pay-controller/src/strategy/across/requests.ts
@@ -1,0 +1,9 @@
+import type { QuoteRequest } from '../../types';
+
+export function isAcrossQuoteRequest(request: QuoteRequest): boolean {
+  return (
+    request.isMaxAmount === true ||
+    (request.targetAmountMinimum !== undefined &&
+      request.targetAmountMinimum !== '0')
+  );
+}
--- a/packages/transaction-pay-controller/CHANGELOG.md
+++ b/packages/transaction-pay-controller/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Ignore synthetic gas legs when determining Across support for perps direct deposits ([#8527](https://github.com/MetaMask/core/pull/8527))
 - Route Across status polling through the configured Across API base and support `depositTxnRef`/`fillTxnRef` for Across status responses ([#8512](https://github.com/MetaMask/core/pull/8512))
 
 ## [19.2.1]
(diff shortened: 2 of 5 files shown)


packages/transaction-pay-controller/src/strategy/across/requests.ts
```diff
@@ -0,0 +1,9 @@
+import type { QuoteRequest } from '../../types';
+
+export function isAcrossQuoteRequest(request: QuoteRequest): boolean {
+  return (
+    request.isMaxAmount === true ||
+    (request.targetAmountMinimum !== undefined &&
```
vinistevam, line 6: Nit: `(request.targetAmountMinimum ?? '0') !== '0'` 
Any real amount maps to true.

vinistevam (approved):

PR #8542, opened 2026-04-21
feat(analytics-controller): persist analyticsId in controller state

gauthierpetetin:
## Explanation

Set analyticsId metadata persist flag to true so persisted stores include the identifier alongside optedIn.
This is needed by Extension, where we want to move the ID from MetaMetricsController to AnalyticsController.

## References

None

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes persistence behavior for `analyticsId`, which affects how a user identifier is saved/restored across sessions and could impact client migrations or state storage expectations.
> 
> **Overview**
> **`analyticsId` is now persisted by `AnalyticsController`** by marking it `persist: true` in state metadata, so persisted controller compositions save/restore it alongside `optedIn`.
> 
> Docs and JSDoc are updated to reflect the new persistence contract (platform supplies the initial UUIDv4, then the persisted store retains it), and tests add explicit coverage for metadata-derived outputs (debug snapshots, state logs, persisted state, and UI-exposed state).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12136bf9d3bc2b7c5d301efc03abb9b90b873269. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


NicolasMassart (commented):
How does this work with the mobile app that is already passing a UUID at controller construction?
We will have 2 sources of UUID: storage and state. How do we make sure they are in sync now?

gauthierpetetin:
@NicolasMassart , in Mobile, isolated storage remains the source of truth for `analyticsId`.

At startup, `EngineService` calls `getAnalyticsId()`, which reads/generates the value from `ANALYTICS_ID`, then passes that value into `Engine.init`. `analytics-controller-init` then constructs `AnalyticsController` state with that `analyticsId` argument and only takes `optedIn` from `persistedState.AnalyticsController`.

So even if `persistedState.AnalyticsController.analyticsId` exists, Mobile does not use it to initialize the controller. The storage value wins.

Once `analyticsId` is marked `persist: true`, Mobile’s existing controller persistence flow will include the storage-sourced `analyticsId` in the persisted `AnalyticsController` state. During init, `setupEnginePersistence` compares the controller’s current persistent state against the initial persisted state; if `analyticsId` is missing or different, it queues a write. That means persisted controller state becomes a mirror of the storage value.

Also, `AnalyticsController` does not mutate `analyticsId` after construction, so there is no runtime path where the two values can independently drift. The only reconciliation point needed is startup, and Mobile already makes storage the source of truth there.

NicolasMassart (approved):
Looks good to me.

PR #8554, opened 2026-04-23
fix: quote request should not be sent of amount of token requested by the user is 0

jpuri:
## Explanation

Money account page updated transaction a few times, this change ensure that we get quote only of token amount requested is not 0.

## References

Fixes https://consensyssoftware.atlassian.net/browse/CONF-1241

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes required-token detection behavior used to drive quoting/eligibility, which could alter downstream flows that relied on the synthetic gas-fee token being present. Scope is small and covered by updated tests, but behavior change is potentially breaking for consumers.
> 
> **Overview**
> `parseRequiredTokens` no longer synthesizes a native gas-fee “required token” when `requiredAssets` are absent; it now returns **only** the ERC-20 transfer token (or `[]` if none is detected).
> 
> This removes the `getGasFeeToken` calculation (and related dependencies) and updates unit tests/changelog to reflect that invalid/unsupported transfer data yields an empty required-token list rather than a gas-fee fallback token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c84f3a878eaaaaf54a292df9d577258b784be8f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -78,6 +76,19 @@ export async function updateQuotes(
...
+      (token) => token.skipIfBalance === false && token.amountRaw !== '0',
+    );
+
+  if (!hasNonZeroTokenAmount) {
+    log('Skipping quotes, all token amounts are zero', { transactionId });
+    return false;
+  }
+
+  log('Updating quotes', { transactionId });
+
```
jpuri: On money account page token / accountOverride can be updated few times. We need to ensure that quotes is not refetched unless user enters a non-zero token amount. Thus this change is made.

Another approach could be to change methods `setTransactionConfig`, 'setPaymentToken' to not refresh quote is amount of token requested is zero.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -84,6 +82,18 @@ export async function updateQuotes(
     tokens,
   } = transactionData;
 
+  const hasNonZeroTokenAmount =
```
matthewwalsh0: We already have this logic in `source-amounts.ts`, so this was working previously, so is this a unique issue for account override somehow?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/utils/quotes.ts
jpuri: When we update transactionConfig - like update accountOverride, it triggers updating payToken and getting new quote. So this was happening frequently on money account pages.

Also, money account transaction I found was request 0 value for required token and a small value for another token with `token.skipIfBalance` true. I guess it was for gas fees.

matthewwalsh0: But we only generate quote requests for any source amounts we have, and the logic in `calculatePostQuoteSourceAmounts` should already skip tokens if its `skipIfBalance` or if the amount is zero?

So we need to debug why that existing logic isn't working.

jpuri: Yep, let me debug existing code.

jpuri: I found that we are updating source amount here: https://github.com/MetaMask/core/blob/3733d0c796bd8bb793a5e0ed93b342722fac0fbd/packages/transaction-pay-controller/src/TransactionPayController.ts#L258

But despite zero amount we send request to api.

I think we can simplify code added in this PR to just check if all sourceAmounts are '0' we do not send request.

matthewwalsh0 (approved):

PR #8557, opened 2026-04-23
fix(messenger): Throw different error for missing delegated actions

FrederikBolding:
## Explanation

The error thrown when using `messenger.call` on delegated messengers is misleading as more often than not it is a missing delegation causing the error being thrown. This PR modifies the error message if the action being called is not in the messengers own namespace.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes the error message thrown by `Messenger.call` for non-local (delegated) actions, which could break consumers/tests that assert on exact error strings. Logic change is small and covered by updated unit tests.
> 
> **Overview**
> Adjusts `Messenger.call` to throw a **more specific error** when an action is missing because it was never delegated (or was revoked), instead of always reporting it as “not registered”.
> 
> Updates delegation/revocation unit tests to assert the new message and records the fix in the `@metamask/messenger` changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0481f2592b3c6e6faff2ab6238baaa353bf9a99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/messenger/src/Messenger.ts
+++ b/packages/messenger/src/Messenger.ts
@@ -421,7 +421,11 @@ export class Messenger<
       ActionType
     >;
     if (!handler) {
-      throw new Error(`A handler for ${actionType} has not been registered`);
+      throw new Error(
+        this.#isInCurrentNamespace(actionType)
+          ? `A handler for ${actionType} has not been registered`
+          : `A handler for ${actionType} has not been delegated to ${this.#namespace}`,
+      );
     }
     return handler(...params);
   }
--- a/packages/messenger/CHANGELOG.md
+++ b/packages/messenger/CHANGELOG.md
@@ -7,14 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-
-- Throw different error for missing delegated actions ([#8557](https://github.com/MetaMask/core/pull/8557))
-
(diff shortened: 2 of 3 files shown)


packages/messenger/src/Messenger.ts
```diff
@@ -421,7 +421,11 @@ export class Messenger<
       ActionType
     >;
     if (!handler) {
-      throw new Error(`A handler for ${actionType} has not been registered`);
+      throw new Error(
+        this.#isInCurrentNamespace(actionType)
+          ? `A handler for ${actionType} has not been registered`
+          : `A handler for ${actionType} has not been delegated to ${this.#namespace}`,
```
GuillaumeRx, line 427: Do we really need to mention the namespace here ?

GuillaumeRx (dismissed):
Small nit

packages/messenger/src/Messenger.ts
FrederikBolding: I think it makes it easier to track down where it is being called. Now you'll know that if the handler isn't registered, you can look at the namespaced messenger for that handler, if it is a missing delegation you will get the name of the messenger missing the delegation.

mcmire (approved):
Makes sense!

Gudahtt (approved):
LGTM!

PR #8575, opened 2026-04-24
feat: Add `subscribeOnce` and `waitUntil` methods

FrederikBolding:
## Explanation

Add `subscribeOnce` and `waitUntil` utility methods to `Messenger`. These are useful for subscribing to and waiting for a certain event to fire without wanting to listen for more than one invocation.

Additionally the methods support a `condition` parameter which can be used as a predicate for whether the event should be consumed in the first place.

A similar approach is already in-use on mobile, this is a "native" replacement.

## References

https://consensyssoftware.atlassian.net/browse/WPC-985

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core event-subscription behavior by introducing auto-unsubscription and Promise-based waiting, which could subtly affect event delivery/ordering and handler lifecycle. Coverage is added, but the new logic interacts with the existing selector/cache and unsubscribe paths.
> 
> **Overview**
> Adds two new event utilities to `Messenger`: `subscribeOnce` to auto-unsubscribe after the first matching event (optionally gated by a `condition`, and supporting selector-based subscriptions), and `waitUntil` to return a Promise that resolves on the next matching event using the same selector/condition options.
> 
> Updates the test suite with coverage for one-shot subscriptions, selector behavior (including previous value), conditional consumption, and Promise resolution, and records the feature in the `@metamask/messenger` changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e57a2546b778129982e15ed3c585667e2fac0e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/messenger/src/Messenger.test.ts
+++ b/packages/messenger/src/Messenger.test.ts
@@ -1036,6 +1036,131 @@ describe('Messenger', () => {
     });
   });
 
+  describe('subscribeOnce', () => {
+    it('unsubscribes automatically after receiving the first event', () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [string];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const handler = jest.fn();
+      messenger.subscribeOnce('Fixture:message', handler);
+      messenger.publish('Fixture:message', 'foo');
+      messenger.publish('Fixture:message', 'bar');
+
+      expect(handler).toHaveBeenCalledWith('foo');
+      expect(handler).not.toHaveBeenCalledWith('bar');
+      expect(handler.mock.calls).toHaveLength(1);
+    });
+
+    it('supports selectors', () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [{ value: string }];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const handler = jest.fn();
+      messenger.subscribeOnce('Fixture:message', handler, ({ value }) => value);
+      messenger.publish('Fixture:message', { value: 'foo' });
+      messenger.publish('Fixture:message', { value: 'bar' });
+
+      expect(handler).toHaveBeenCalledWith('foo');
+      expect(handler).not.toHaveBeenCalledWith('bar');
+      expect(handler.mock.calls).toHaveLength(1);
+    });
+
+    it('supports conditions', () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [{ value: string }];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const handler = jest.fn();
+      messenger.subscribeOnce(
+        'Fixture:message',
+        handler,
+        ({ value }) => value,
+        (value) => value === 'bar',
+      );
+      messenger.publish('Fixture:message', { value: 'foo' });
+      messenger.publish('Fixture:message', { value: 'bar' });
+
+      expect(handler).not.toHaveBeenCalledWith('foo');
+      expect(handler).toHaveBeenCalledWith('bar');
+      expect(handler.mock.calls).toHaveLength(1);
+    });
+  });
+
+  describe('waitUntil', () => {
+    it('resolves the promise when the event fires', async () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [string];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const promise = messenger.waitUntil('Fixture:message');
+      messenger.publish('Fixture:message', 'foo');
+
+      expect(await promise).toBe('foo');
+    });
+
+    it('supports selectors', async () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [{ value: string }];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const promise = messenger.waitUntil(
+        'Fixture:message',
+        ({ value }) => value,
+      );
+      messenger.publish('Fixture:message', { value: 'foo' });
+
+      expect(await promise).toBe('foo');
+    });
+
+    it('supports conditions', async () => {
+      type MessageEvent = {
+        type: 'Fixture:message';
+        payload: [{ value: string }];
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const promise = messenger.waitUntil(
+        'Fixture:message',
+        ({ value }) => value,
+        (value) => value === 'bar',
+      );
+      messenger.publish('Fixture:message', { value: 'foo' });
+      messenger.publish('Fixture:message', { value: 'bar' });
+
+      expect(await promise).toBe('bar');
+    });
+  });
+
   describe('clearEventSubscriptions', () => {
     it('does not call subscriber after clearing event subscriptions', () => {
       type MessageEvent = { type: 'Fixture:message'; payload: [string] };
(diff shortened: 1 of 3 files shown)


packages/messenger/src/Messenger.test.ts
```diff
@@ -1010,6 +1010,131 @@ describe('Messenger', () => {
...
+      };
+
+      const messenger = new Messenger<'Fixture', never, MessageEvent>({
+        namespace: 'Fixture',
+      });
+
+      const promise = messenger.waitUntil('Fixture:message');
+      messenger.publish('Fixture:message', 'foo');
+
+      expect(await promise).toBe('foo');
```
FrederikBolding: When not using a selector this just returns the first parameter of the event. The other option is returning the entire array, but that feels a bit complicated 🤔

packages/messenger/src/Messenger.ts
```diff
@@ -656,6 +656,125 @@ export class Messenger<
...
+   * @param condition - An optional function which is executed with the selector's return
+   * value. The promise only resolves when this returns `true`.
+   * @template EventType - A type union of Event type strings.
+   * @template SelectorReturnValue - The selector return value.
+   * @returns A promise that resolves with the selector's return value.
+   */
+  waitUntil<EventType extends Event['type'], SelectorReturnValue>(
+    eventType: EventType,
+    selector: SelectorFunction<Event, EventType, SelectorReturnValue>,
+    condition?: (value: SelectorReturnValue) => boolean,
```
Mrtenz: What if you want to specify a condition, but not a selector? Like this for example:

```ts
// Subscribe to `KeyringController:unlock` and wait if condition is not met
await messenger.waitUntil(
  'KeyringController:unlock',
  () => !engineContext.KeyringController.isUnlocked()
);
```

Maybe it's easier to have an options object, so you could do something like this:

```ts
await messenger.waitUntil(
  'KeyringController:unlock',
  {
    condition: () => !engineContext.KeyringController.isUnlocked()
  }
);
```

FrederikBolding: I'm confused by this question, if `KeyringController.isUnlocked` is true you wouldn't want to wait for `unlock` in the first place. The `condition` is meant as a condition on the value itself and is only checked when an event fires, e.g.

```typescript
await messenger.waitUntil(
  'TransactionController:transactionSuccess',
  (tx) => tx.hash === 'foo'
);
```

FrederikBolding: That being said, it would be nice if you could pass a condition without a selector. Potentially worth doing an options bag regardless.

Mrtenz: Then I was misunderstanding what the condition is supposed to do 😅

Mrtenz: I was under the impression that these would be equivalent:

```ts
await messenger.waitUntil(
  'KeyringController:unlock',
  {
    condition: () => !engineContext.KeyringController.isUnlocked()
  }
);
```

and

```ts
if (!engineContext.KeyringController.isUnlocked()) {
  await messenger.waitUntil('KeyringController:unlock');
}
```

FrederikBolding: I don't see an obvious way to solve for both that and the `TransactionController` example from above 🤔

Mrtenz: Yeah, we'd need to introduce a third property. But in hindsight, I'm not sure if it's necessary as the if statement seems clean enough.

FrederikBolding: Certainly better than what we have today

FrederikBolding: https://github.com/MetaMask/core/pull/8575/commits/5907ef9b57940ce88870dd89e224cfe5ed5de484

packages/messenger/src/Messenger.test.ts
mcmire: Hmm, why does it feel complicated? If it's easy to support, I say we give consumers the whole payload.

packages/messenger/src/Messenger.ts
```diff
@@ -656,6 +656,161 @@ export class Messenger<
...
+  waitUntil<EventType extends Event['type'], SelectorReturnValue>(
+    eventType: EventType,
+    options?: {
+      selector?: SelectorFunction<Event, EventType, SelectorReturnValue>;
+      condition?:
+        | ((...payload: ExtractEventPayload<Event, EventType>) => boolean)
+        | ((value: SelectorReturnValue) => boolean);
+    },
+  ): Promise<SelectorReturnValue | ExtractEventPayload<Event, EventType>[0]> {
+    return new Promise((resolve) => {
```
mcmire, line 854: How we expect `waitUntil` to be used? In Slack you gave this example:

``` typescript
  getUnlockPromise: () => {
    if (engineContext.KeyringController.isUnlocked()) {
      return Promise.resolve();
    }
    return new Promise<void>((resolve) => {
      controllerMessenger.subscribeOnceIf(
        'KeyringController:unlock',
        resolve,
        () => true,
      );
    });
  },
```

Is the idea that with this method you could shorten this to the following?

``` typescript
  getUnlockPromise: () => {
    return await controllerMessenger.waitUntil(
      'KeyringController:unlock',
      {
        condition: () => {
          return engineContext.KeyringController.isUnlocked();
        }
      }
    );
  },
```

Or is it that we would expect people to write this instead?

``` typescript
  getUnlockPromise: () => {
    return await controllerMessenger.waitUntil(
      'KeyringController:stateChange',
      {
        selector: (state) => state.isUnlocked,
      }
    );
  },
```

mcmire: Either way would we want to check the condition beforehand to make sure is not true before subscribing? (I guess the same goes for `subscribeOnce` as well.)

FrederikBolding: The condition is only applied when a state update comes in, it is meant for a scenario like this:

```ts
await messenger.waitUntil(
  'TransactionController:transactionSuccess',
  { condition: (tx) => tx.hash === 'foo' }
);
```

FrederikBolding: For the unlock example it would likely have to be:

```ts
  getUnlockPromise: () => {
    if (engineContext.KeyringController.isUnlocked()) {
      return Promise.resolve();
    }
    return controllerMessenger.waitUntil('KeyringController:unlock');
  },
```

packages/messenger/src/Messenger.test.ts
FrederikBolding: Ah, I should have clarified. What I was worried about is the ergonomics of dealing with the entire payload array versus just the first parameter (since more often than not the first parameter is probably what you want). But it may be fine. 

See https://github.com/MetaMask/core/pull/8575/commits/0ab75e38e93c919315097f775de8239e9f333607

packages/messenger/src/Messenger.ts
mcmire: Gotcha. If we mainly would use `condition` for waiting for an object to reach a certain status, then do we need `condition` at all? I feel like having a specialized event solves the same problem but without needing to check anything else at runtime:

```typescript
await messenger.waitFor(
  `TransactionController:transactionSuccess:${tx.hash}`,
);
```

FrederikBolding: Sure, that's another way of doing it. But I was under the impression that we agreed on having both tools would be useful and that forcing more granular events is not necessarily the path we wanted.

Mrtenz: The value you want to check for may also not necessarily be serialisable to a string.

packages/messenger/src/Messenger.ts
```diff
@@ -656,6 +656,161 @@ export class Messenger<
...
+   * @param options - Options bag.
+   * @param options.selector - The selector function used to select relevant data
+   * from the event payload. The type of the parameters for this selector must
+   * match the type of the payload for this event type.
+   * @param options.condition - An optional predicate evaluated against the
+   * selector's return value. The handler is only invoked when this returns `true`.
+   * @template EventType - A type union of Event type strings.
+   * @template SelectorReturnValue - The selector return value.
+   */
+  subscribeOnce<EventType extends Event['type'], SelectorReturnValue>(
```
Mrtenz, line 707: Maybe it's useful to add some JSDoc examples to these.

FrederikBolding: WDYT of these? https://github.com/MetaMask/core/pull/8575/commits/98eb62ac06e12e14be0701ed0703252f900ae72b

Mrtenz: Looks good, but maybe add one more example to `waitUntil` to show it can be used without selector/just for waiting for unlock for example?

FrederikBolding: Done!

Mrtenz (approved):

PR #8580, opened 2026-04-24
chore: fail-open-multichain assets controller filter logic

sahar-fehri:
## Explanation

Restores the original fail-open behavior in `MultichainAssetsController`'s Blockaid spam token filter, ensuring legitimate tokens aren't blocked due to API failures or missing data.

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes the spam/malicious token gating behavior so tokens may be added when the Blockaid bulk-scan call fails, which can reduce protection during outages and affects security-sensitive filtering logic.
> 
> **Overview**
> Updates Blockaid-based filtering for newly detected `token:` CAIP assets to be **fail-open**: if a `bulkScanTokens` batch rejects, tokens in that batch are now allowed through instead of being dropped.
> 
> Adjusts unit tests to reflect the new behavior, including scenarios where `bulkScanTokens` throws or where only some scan batches fail, while still filtering tokens explicitly marked `Malicious` in successful scan responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3be6a8cacc02a4a4be611268307ef1fac12a9628. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
+++ b/packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
@@ -826,7 +826,7 @@ export class MultichainAssetsController extends StaticIntervalPollingController<
   }
 
   /**
-   * Fail-closed Blockaid filter for newly detected `token:` assets (native/other namespaces unchanged).
+   * Fail-open Blockaid filter for newly detected `token:` assets (native/other namespaces unchanged).
    *
    * @param assets - CAIP assets to filter.
    * @returns Filtered list, original order preserved.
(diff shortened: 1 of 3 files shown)


packages/assets-controllers/src/MultichainAssetsController/MultichainAssetsController.ts
```diff
@@ -850,6 +850,10 @@ export class MultichainAssetsController extends StaticIntervalPollingController<
 
       for (const outcome of batchOutcomes) {
         if (outcome.status === 'rejected') {
+          // Fail-open: if API fails, allow all tokens in this batch through
+          for (const entry of outcome.entries) {
+            keptTokenAssets.add(entry.asset);
+          }
```
bergarces: nit: This probably will work, but it still assumes that a fulfilled response will include the token.

We could change the list from "keptTokenAssets" to "maliciousTokenAssets" and filter them out of the original list only if they are present in this one.

That way we guarantee that only if the response has the token as malicious, it gets filtered out.

bergarces (approved):

PR #8595, opened 2026-04-27
feat: track token_security_type_destination in Unified SwapBridge metrics

infiniteflower:
## Explanation

Adds a new `token_security_type_destination` field to every Unified SwapBridge analytics event that already carries `token_address_destination`. This lets the product team measure how often users swap into flagged tokens and quantify the impact of the destination-token security badge UI.

**Current state.** Unified SwapBridge events report which token a user is swapping into (`token_address_destination`) but not whether the client surfaced any security warning for that token. Without this signal we can't correlate the security badge with user behavior (e.g. drop-off after a "Malicious" classification).

**Solution.** A single string-or-`null` value is threaded through the metric pipeline:

1. Clients pass `token_security_type_destination` once when calling `BridgeController.updateBridgeQuoteRequestParams` (via the `context` arg).
2. `BridgeController` stores it on state (`tokenSecurityTypeDestination: string | null`, default `null`) so all pre-submit events (`InputChanged`, `QuotesRequested`, `QuotesReceived`, `QuotesError`, `QuotesValidationFailed`, etc.) automatically include it via `getRequestParams`.
3. Clients pass it again as an optional arg to `BridgeStatusController.submitTx` / `submitIntent`. The value is persisted onto the `BridgeHistoryItem` so post-submit events fired hours later (`Submitted`, `Completed`, `Failed`, `PollingStatusUpdated`, `StatusValidationFailed`) can read it back via `getRequestParamFromHistory` without depending on `BridgeController` state, which may have moved on by then.

**Non-obvious choices.**

- **`string | null`, not `string | undefined`.** `BridgeControllerState` must satisfy `@metamask/base-controller`'s `StateConstraint`, whose values are `Json`-compatible — and `Json` does not allow `undefined`. `null` also matches the existing convention in `RequestParams` (e.g. `token_address_destination: CaipAssetType | null`). The controller defensively coerces `undefined → null` at the `updateBridgeQuoteRequestParams` boundary so JS callers passing `undefined` don't end up with `"token_security_type_destination": undefined` keys getting dropped by `JSON.stringify`.
- **Conditional spread on `BridgeHistoryItem`.** Persisting unconditionally would have broken every `MockTxHistory` snapshot. The history item uses `...(tokenSecurityTypeDestination !== undefined && { tokenSecurityTypeDestination })` (matches the existing `abTests` / `activeAbTests` pattern); metric output is normalized with `?? null` at emit time.
- **`StatusValidationFailed` needs an explicit add.** Most post-confirm events spread `requestParamProperties`, but `StatusValidationFailed` builds its payload with explicit `Pick`s. The new field had to be added in two places: `getRequestParamFromHistory` AND the explicit pick in `#trackUnifiedSwapBridgeEvent`.
- **Why pass it explicitly to `submitTx` instead of reading from `BridgeController` state at submit time?** Considered but rejected. Reading state would require widening `bridge-status-controller`'s messenger `AllowedActions` with `BridgeController:getState` (a heavy permission) plus client wiring updates. The explicit-arg approach matches the existing pattern for every other analytics context piece on `submitTx` (`abTests`, `activeAbTests`, `location`, `quotesReceivedContext`) and keeps the cross-controller surface narrow.

https://github.com/user-attachments/assets/1b51fb95-ff36-4b20-a61a-c0341b857143

## References

- https://github.com/Consensys/segment-schema/pull/544
- https://consensyssoftware.atlassian.net/browse/SWAPS-4420
- https://consensyssoftware.atlassian.net/browse/SWAPS-4422

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it introduces **breaking** type/signature changes across `bridge-controller` and threads a new analytics field through pre- and post-submit flows, which could break downstream consumers if not updated.
> 
> **Overview**
> **Adds `token_security_type_destination` to Unified SwapBridge analytics payloads** so events that include `token_address_destination` also report the client-provided destination-token security classification.
> 
> In `bridge-controller`, this is stored on controller state (`tokenSecurityTypeDestination`) via the `context` argument of `updateBridgeQuoteRequestParams`, reset on `resetState`, and passed into `getRequestParams` (now a **breaking** 2-arg API) to populate pre-submit event properties; `InputSourceDestinationSwitched` is also updated to always include the field.
> 
> In `bridge-status-controller`, `submitTx`/`submitIntent` accept an optional `tokenSecurityTypeDestination`, persist it on `BridgeHistoryItem`, and emit it on post-submit analytics via `getRequestParamFromHistory`; tests/snapshots are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d4b2690c3ff27b1128a5a02e08aac38884024f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bridge-controller/src/constants/bridge.ts
```diff
@@ -91,6 +91,7 @@ export const DEFAULT_BRIDGE_CONTROLLER_STATE: BridgeControllerState = {
   assetExchangeRates: {},
   minimumBalanceForRentExemptionInLamports: '0',
   tokenWarnings: [],
+  tokenSecurityTypeDestination: null,
```
micaelae, line 94: What are the possible values for this? Are they enums or human-readable strings?

packages/bridge-controller/src/bridge-controller.test.ts
```diff
@@ -97,6 +97,7 @@ const metricsContext = {
   stx_enabled: true,
   security_warnings: [],
   warnings: [],
+  token_security_type_destination: null,
```
micaelae, line 100: What's the difference between `security_warnings` and `token_security_type_destination`?

infiniteflower: `security_warnings` is from `token.securityData.metadata.features.map(feature => feature.description)`

While `token_security_type_destination` is the `token.securityData.resultType`

packages/bridge-controller/src/constants/bridge.ts
infiniteflower: Yea I was debating on whether or not to make this an enum, but I'm leaning to `string` because Assets Team's `TrendingAsset.securityData.resultType` from the `assets-controlelrs` is a `string`.

When we select a trending token we get a `TrendingToken`. We only get `BridgeToken` when we call the Bridge API's `/popular` or `/search` endpoint.

micaelae (dismissed):

packages/bridge-status-controller/src/utils/metrics.ts
GeorgeGkas, line 193: @infiniteflower we need to add the type here  as well  else Unified SwapBridge Submitted  `swapbridge-v1` action will not include the result security type prop

GeorgeGkas (changes requested):

GeorgeGkas (dismissed):

packages/bridge-controller/src/bridge-controller.ts
```diff
@@ -1037,7 +1079,10 @@ export class BridgeController extends StaticIntervalPollingController<BridgePoll
         // Populate the properties that the error occurred before the tx was submitted
         return {
           ...baseProperties,
-          ...getRequestParams(this.state.quoteRequest),
+          ...getRequestParams(
+            this.state.quoteRequest,
+            this.state.tokenSecurityTypeDestination,
+          ),
```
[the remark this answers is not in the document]
infiniteflower: [7df2f18](https://github.com/MetaMask/core/pull/8595/commits/7df2f1839c674054a77da7fd2870925cd9912dfc)

GeorgeGkas (approved):

PR #8612, opened 2026-04-28
fix(transaction-pay-controller): abort stale quote requests

matthewwalsh0:
## Explanation

When `updateQuotes` is called twice in quick succession for the same transaction (e.g. user typing an amount with debounce, toggling max, or switching payment token), two quote fetches race with no cancellation. A slower older response can land after a faster newer one and silently overwrite `data.quotes`, `data.totals`, `tx.metamaskPay`, and `tx.batchTransactions` — so displayed fees no longer match the displayed amount.

This PR adds per-transaction `AbortController` tracking to `updateQuotes`. A fresh call aborts any in-flight predecessor, and aborted calls bail before mutating state. The signal is also forwarded into the Relay and Across `successfulFetch` calls so cancelled requests drop at the network level.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches quote-refresh control flow and state mutation ordering; incorrect abort handling could cause missed updates or loading-state inconsistencies, though changes are localized and well-covered by tests.
> 
> **Overview**
> Prevents stale quote refreshes from racing by **tracking per-transaction in-flight `updateQuotes` calls** and aborting any previous request when a newer call for the same `transactionId` starts, returning `false` for aborted runs and skipping state writes (including `isLoading` teardown) when cancelled.
> 
> Threads an optional `AbortSignal` through pay-strategy request types and forwards it into Relay (`fetchRelayQuote`) and Across (`requestAcrossApproval`) `successfulFetch` calls so cancelled quote requests also stop at the network layer; adds targeted unit tests covering cancellation behavior and signal propagation, and documents the change in the changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 307a0b6e4ba70c4cdceb2a825f5a094f44090d83. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/CHANGELOG.md
+++ b/packages/transaction-pay-controller/CHANGELOG.md
@@ -30,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Prevent stale quote responses from overwriting newer ones when `updateQuotes` is called multiple times in quick succession for the same transaction
+  - A fresh call now aborts any in-flight call for the same `transactionId` via `AbortController`; aborted calls bail before writing `data.paymentToken`, `data.quotes`, `data.totals`, `tx.metamaskPay`, or `tx.batchTransactions`
+  - Eliminates a race where a slower earlier quote (e.g. an older amount) could overwrite a faster later quote, causing displayed fees to mismatch the displayed amount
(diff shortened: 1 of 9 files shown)


matthewwalsh0:
@metamaskbot publish-preview

OGPoyraz (approved):

PR #8628, opened 2026-04-29
feat: pass explicit params to `RampsController:getQuotes` and persist `rampsQuote` on `fiatPayment`

OGPoyraz:
## Explanation

The fiat quote flow in `transaction-pay-controller` was calling `RampsController:getQuotes` without specifying the asset, provider, or currency - relying on defaults or a hardcoded `pickBestFiatQuote` helper that filtered for a specific staging provider.

This PR:

1. **Passes explicit parameters** to `RampsController:getQuotes`: `assetId` (from `fiatAsset.caipAssetId`), `providers` (from `RampsControllerState.providers.selected.id`), and `fiat: 'USD'`.
2. **Picks the first quote** from `quotes.success[0]` instead of searching by hardcoded provider name via `pickBestFiatQuote`.
3. **Persists the ramps quote** on `TransactionFiatPayment.rampsQuote` via `TransactionPayController:updateFiatPayment` so downstream consumers can access it.
4. **Extracts `getRampsQuote`** helper to isolate ramps state reading and quote fetching for readability.
5. **Removes `pickBestFiatQuote`** and the unused `FiatQuotesResponse` type alias since they are no longer needed.
6. **Adds `RampsControllerGetStateAction`** to allowed actions so the messenger can read ramps controller state.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes the fiat quoting path by selecting providers via `RampsController:getState` and altering how a ramps quote is chosen/persisted, which could affect quote availability and downstream consumers of `fiatPayment` state.
> 
> **Overview**
> Improves the fiat quote flow to call `RampsController:getQuotes` with explicit `assetId`, `fiat` (defaulting to `USD`), and an optional `providers` filter derived from `RampsController:getState` (instead of relying on implicit defaults and a hardcoded provider filter).
> 
> Persists the selected ramps quote onto per-transaction state (`TransactionFiatPayment.rampsQuote`) via `TransactionPayController:updateFiatPayment`, removes the now-unused `pickBestFiatQuote` helper/types, and updates tests to cover provider-selection behavior and persistence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e4c29b5bb5e0cba5494727c027104662bb0799f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
@@ -14,6 +14,7 @@ import type {
 import { computeRawFromFiatAmount, getTokenFiatRate } from '../../utils/token';
 import { getRelayQuotes } from '../relay/relay-quotes';
 import type { RelayQuote } from '../relay/types';
+import { DEFAULT_FIAT_CURRENCY } from './constants';
 import type { TransactionPayFiatAsset } from './constants';
 import type { FiatQuote } from './types';
 import { deriveFiatAssetForFiatPayment } from './utils';
@@ -171,7 +172,7 @@ async function getRampsQuote({
   const quotes = await messenger.call('RampsController:getQuotes', {
     amount: adjustedAmount,
     assetId: fiatAsset.caipAssetId,
-    fiat: 'USD',
+    fiat: DEFAULT_FIAT_CURRENCY,
     paymentMethods: [fiatPaymentMethod],
     providers: selectedProviderId ? [selectedProviderId] : undefined,
     walletAddress,
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
@@ -14,8 +14,9 @@ import type {
 import { computeRawFromFiatAmount, getTokenFiatRate } from '../../utils/token';
 import { getRelayQuotes } from '../relay/relay-quotes';
 import type { RelayQuote } from '../relay/types';
+import type { TransactionPayFiatAsset } from './constants';
 import type { FiatQuote } from './types';
-import { deriveFiatAssetForFiatPayment, pickBestFiatQuote } from './utils';
+import { deriveFiatAssetForFiatPayment } from './utils';
 
 const log = createModuleLogger(projectLogger, 'fiat-strategy');
 
@@ -115,22 +116,20 @@ export async function getFiatQuotes(
       transactionId,
     });
 
-    const quotes = await messenger.call('RampsController:getQuotes', {
-      amount: adjustedAmount,
-      paymentMethods: [fiatPaymentMethod],
+    const fiatQuote = await getRampsQuote({
+      adjustedAmount,
+      fiatAsset,
+      fiatPaymentMethod,
+      messenger,
       walletAddress,
     });
 
-    log('Fetched ramps quotes', {
-      rampsQuotesCount: quotes.success?.length ?? 0,
+    messenger.call('TransactionPayController:updateFiatPayment', {
+      callback: (fiatPayment) => {
+        fiatPayment.rampsQuote = fiatQuote;
+      },
       transactionId,
     });
-
-    const fiatQuote = pickBestFiatQuote(quotes);
-
-    if (!fiatQuote) {
-      throw new Error('No matching ramps quote found for selected provider');
-    }
 
     return [
       combineQuotes({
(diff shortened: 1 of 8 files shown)


packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -153,6 +152,45 @@ function getRequiredTokens(
...
+  walletAddress,
+}: {
+  adjustedAmount: number;
+  fiatAsset: TransactionPayFiatAsset;
+  fiatPaymentMethod: string;
+  messenger: PayStrategyGetQuotesRequest['messenger'];
+  walletAddress: string;
+}): Promise<RampsQuote> {
+  const rampsState = messenger.call('RampsController:getState');
+  const selectedProviderId = rampsState.providers.selected?.id;
```
OGPoyraz, line 170: There will always be a selected provider as Ramps team mentioned. Even though this is `undefined` we will pick first quote.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -153,6 +152,45 @@ function getRequiredTokens(
...
+
+  const quotes = await messenger.call('RampsController:getQuotes', {
+    amount: adjustedAmount,
+    assetId: fiatAsset.caipAssetId,
+    fiat: 'USD',
```
OGPoyraz: 'USD' will be hard lock for MMPay confirmations.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -115,23 +116,21 @@ export async function getFiatQuotes(
...
+      messenger,
       walletAddress,
     });
 
-    log('Fetched ramps quotes', {
-      rampsQuotesCount: quotes.success?.length ?? 0,
+    messenger.call('TransactionPayController:updateFiatPayment', {
+      callback: (fiatPayment) => {
+        fiatPayment.rampsQuote = fiatQuote;
+      },
```
OGPoyraz, line 131: Exact `fiatPayment.rampsQuote` will be required to start headless ramps operation in client, hence we will hold this in the `fiatPayment` state.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
amitabh94: Not sure if fiat has any effect here as the currency is determined by the region. @saustrie-consensys @pkowalski can check and confirm

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
wachunei: Although there's going to be a selected provider already this function restricts the current abilities of `RampsController:getQuotes`, so in case there is a need to set a specific provider this function would not allow it. I'm ok approving as is if this can be refactored later.

wachunei (dismissed):

saustrie-consensys (dismissed):

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
matthewwalsh0: To clarify, this insists the provider deal in USD? Do all the providers respect that? Or are the values / quotes converted in the `RampsController`?

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -153,6 +152,45 @@ function getRequiredTokens(
...
+}): Promise<RampsQuote> {
+  const rampsState = messenger.call('RampsController:getState');
+  const selectedProviderId = rampsState.providers.selected?.id;
+
+  const quotes = await messenger.call('RampsController:getQuotes', {
+    amount: adjustedAmount,
+    assetId: fiatAsset.caipAssetId,
+    fiat: 'USD',
+    paymentMethods: [fiatPaymentMethod],
+    providers: selectedProviderId ? [selectedProviderId] : undefined,
```
matthewwalsh0, line 177: Why do we need to retrieve state from the `RampsController` only to give it back?

Can this not be the default `providers` value if not set?

Or we use an optional property such as `useSelectedProvider: true`?

Not a blocker, but would help encapsulation and decoupling.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -153,6 +152,45 @@ function getRequiredTokens(
...
+    walletAddress,
+  });
+
+  log('Fetched ramps quotes', {
+    quotesCount: quotes.success?.length ?? 0,
+    selectedProviderId,
+  });
+
+  const quote = quotes.success?.[0];
```
matthewwalsh0, line 186: Just clarifying the obvious for audit sake that `success` is ordered by preference so first entry is best?

matthewwalsh0 (dismissed):

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
matthewwalsh0: And worth a constant?

OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: I agree, there are multiple ways to adjust the request in better shape, we should even consider having an explicit API to return single quote for MMPay use-case as we care just the best quote. 

These are all mentioned and alignments may follow in future.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: Yes that's right, but I believe we most probably never going to get more than one quote as limited to one provider while calling `getQuotes`.

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

PR #8633, opened 2026-04-29
feat(transaction-controller): gate internal validations on explicit `isInternal` flag

matthewwalsh0:
## Explanation

The `TransactionController` currently distinguishes "internal" (MetaMask-initiated) requests from "external" (dapp-initiated) requests by checking whether the request `origin` matches `ORIGIN_METAMASK`. Several validations in `addTransaction` and `addTransactionBatch` are skipped for internal requests:

- Permitted-address check
- EIP-7702 authorization rejection
- Calls-with-data targeting an internal account
- Duplicate `batchId` check
- Batch size limit
- Dapp-suggested gas fee handling and `userFeeLevel` selection

Inferring trust from the origin string couples the controller's security posture to whatever logic the consuming client happens to use when populating that field. A bug, refactor, or oversight further up the stack that lets a non-internal request reach the controller with `origin: ORIGIN_METAMASK` would silently bypass these validations.

This PR replaces that implicit check with an explicit, optional `isInternal: boolean` flag on both `AddTransactionOptions` and `TransactionBatchRequest`. The flag defaults to `false`, so the controller treats all requests as external unless the caller opts in. The flag is persisted on `TransactionMeta` and `TransactionBatchMeta` so downstream logic (notably `userFeeLevel` selection in `gas-fees.ts`) reads it directly rather than re-deriving trust from the origin.

The origin field is unchanged in shape and semantics — it still flows through to the `ApprovalController` as the display origin and is still attached to `TransactionMeta` for analytics, history, and dapp-suggested-gas display. It is simply no longer used as a trust signal.

## **BREAKING**: validations are no longer gated on `origin`

Callers that previously relied on `origin: ORIGIN_METAMASK` (or an absent origin) to suppress these validations must now pass `isInternal: true` explicitly. Without that flag the controller will:

- Reject EIP-7702 transactions
- Enforce the `permittedAddresses` check on the `from` address
- Enforce the batch size limit
- Reject calls-with-data targeting an internal account
- Reject duplicate `batchId`s
- Treat the transaction as dapp-suggested for `userFeeLevel`

Internal MetaMask consumers (extension, mobile, and other core packages such as `transaction-pay-controller`, `bridge-status-controller`, `perps-controller`, and `eip-5792-middleware`) need to audit their call sites and pass `isInternal: true` for every request that genuinely originates inside MetaMask.

## Checklist

---

> [!NOTE]
> **High Risk**
> High risk because it changes trust/validation gating for transaction and batch submission; missing `isInternal: true` on internal call sites could cause new rejections or different gas-fee behavior, while incorrect use could bypass security checks.
> 
> **Overview**
> **Replaces implicit “internal” detection based on `origin === ORIGIN_METAMASK` with an explicit `isInternal` flag** across `TransactionController` transaction and batch flows.
> 
> `TransactionController.addTransaction` now defaults `isInternal` to `false`, passes it into origin/batch validation, uses it to control duplicate `batchId` handling, and persists it onto `TransactionMeta`; batch utilities and types similarly carry `isInternal` through `TransactionBatchRequest`/`TransactionBatchMeta`. Gas-fee logic and validation tests are updated to key off `isInternal` (and no longer treat missing/MetaMask origin as trusted), and internal callers (bridge/earn/perps/transaction-pay) are updated to pass `isInternal: true` with corresponding snapshot/test updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83887f1d1d3e39ef3246191445d17acf9ca40860. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


matthewwalsh0:
@metamaskbot publish-preview

OGPoyraz (dismissed):

GeorgeGkas (approved):

jpuri (approved):
Changes look good

OGPoyraz (approved):

Matt561 (approved):

PR #8701, opened 2026-05-05
feat(analytics-controller): add optional invocation options for track, identify, view

gauthierpetetin:
## Explanation

Add the possibility to forward `context`, `callback`, `messageId`, and `timestamp` to AnalyticsController for trackEvent, identify, and trackView functions, , as this is needed in Extension.

## References

NA

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Updates the public `AnalyticsController` and `AnalyticsPlatformAdapter` method signatures to accept new optional invocation metadata, which may require downstream adapter/consumer updates. Adds special handling for `messageId` when emitting split anonymous/sensitive events, which could affect event deduplication behavior.
> 
> **Overview**
> Adds `AnalyticsInvocationOptions` (e.g., `context`, `callback`, `messageId`, `timestamp`) and threads it through `AnalyticsController.trackEvent`, `identify`, and `trackView` into `AnalyticsPlatformAdapter.track`/`identify`/`view`.
> 
> When anonymous-event splitting is enabled and a `messageId` is provided, the controller now derives a distinct `messageId` for the anonymous/sensitive counterpart to avoid SDK deduping. Tests, public exports (`index.ts`), generated method action types, and the package changelog are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94fc58c786f57f86e83bf4d3b51ec8c330ad88eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -28,6 +28,40 @@ export type AnalyticsTrackingEvent = {
...
+
+/**
+ * Optional metadata forwarded with track, identify, and screen or page calls.
+ * Platforms map these to their analytics SDK options.
+ */
+export type AnalyticsInvocationOptions = {
+  /**
+   * Optional context object attached to the invocation.
+   */
+  context?: AnalyticsContext;
```
NicolasMassart, line 50: Q: there's already a context attached by Segment SDK, how does this interfere with it?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -28,6 +28,40 @@ export type AnalyticsTrackingEvent = {
...
+  /**
+   * Optional callback when the invocation completes or fails.
+   */
+  callback?: AnalyticsInvocationCallback;
```
NicolasMassart, line 54: Q: What are the use cases of this? The tracking should never block or interfere with UX, so it's best effort and errors are handled inside controller, right?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -28,6 +28,40 @@ export type AnalyticsTrackingEvent = {
...
+  /**
+   * Optional stable identifier for deduplication or tracing.
+   */
+  messageId?: string;
```
NicolasMassart, line 58: Q; we already have a message ID in the Segment payload. How is this different?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
```diff
@@ -28,6 +28,40 @@ export type AnalyticsTrackingEvent = {
...
+  /**
+   * Optional event time. ISO strings, Unix timestamps in milliseconds, or Date values are typical.
+   */
+  timestamp?: Date | number | string;
```
NicolasMassart, line 62: Q: we already have a timestamp in Segment payload. How is this different?

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
gauthierpetetin: Extension adds context to events, that Segment SDK doesn't provide (cf. [this code](https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/controllers/metametrics-controller.ts#L1288-L1311)).
Segment can still add its own SDK context. We are not replacing that.
We are allowing the adapter to pass additional context through to Segment. If there are key collisions, that is a platform-adapter concern, so Extension should avoid overriding SDK-owned fields unless intentional.
So far, the only SDK-owned fields are:
```
context.library.name = '@segment/analytics-node'
context.library.version = <package version>
```

gauthierpetetin:
@NicolasMassart the reason why we need to pass `callback`, `messageId`, and `timestamp` is because we need them to implement a persisted event queue.

On Mobile, that’s achieved thanks to Segment SDK’s `storePersistor` option, but unfortunately there’s no equivalent option in the Segment SDK used by Extension.

That's quite problematic by the way, because it prevents us from fully getting rid of the existing MetaMetricController: we still need it besides AnalyticsController in order to persist the event queue.

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
gauthierpetetin: See message [here](https://github.com/MetaMask/core/pull/8701#issuecomment-4429770744).

We need a messageId that is defined by MetaMetricsController **before** sending the event to Segment, in order to persist the event queue.

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
gauthierpetetin: We need a timestamp that is defined by MetaMetricsController **before** sending the event to Segment, in order to preserve the original event time when the event is replayed later, thanks to the persisted event queue.

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
gauthierpetetin: See message [here](https://github.com/MetaMask/core/pull/8701#issuecomment-4429770744).

Also `callback` is not for UX and should not be awaited by product flows. It is only the cleanup signal: once Segment finishes processing the call, we remove that `messageId` from the persisted queue.

packages/analytics-controller/src/AnalyticsPlatformAdapter.types.ts
NicolasMassart: Will all these values be in the same context object in the final event payload?

NicolasMassart:
Only comment is if we could explore the Segment SDK plugin system more to see if we can have these event data added in a plugin to prevent changes in the event on the app side.

gauthierpetetin:
After discussion with @NicolasMassart , this PR is now replaced by #8835 and #8797

PR #8711, opened 2026-05-06
feat: fetch batched swap quotes

micaelae:
## Explanation

This PR implements quote polling for batched swap requests

#### Changes
- `updateBridgeQuoteRequest` now takes 2 extra params: the quote request’s index and the total quoteRequestCount within the batch. no changes in how we are using this handler for regular swaps
- the new `selectBatchSellQuotes` selector returns the recommended quote for each quoteRequest, and aggregated amounts for display purposes
- `isValidBatchSellQuoteRequest`

#### Usage
- the clients will need to maintain a list of requests, each one identified by an index. the index identifies the quoteRequest and its related side effects (quotes, metrics, submission etc) once it’s passed to updateBridgeQuoteRequest
- to access quotes for a single quoteRequest, use the same index used during quoteRequest update

#### Minimal client examples
- extension: https://github.com/MetaMask/metamask-extension/pull/42434. Checkout this branch and resolve the package locally to test
- mobile: https://github.com/MetaMask/metamask-mobile/pull/29831

## References

Fixes https://consensyssoftware.atlassian.net/browse/SWAPS-4443

## Checklist

---

> [!NOTE]
> **Medium Risk**
> This is a breaking API/state-shape change (`quoteRequest` becomes an array) that touches quote polling/streaming, analytics properties, and exchange-rate lookup logic; regressions could affect quote fetching and refresh behavior across swaps/bridges.
> 
> **Overview**
> **Adds BatchSell (batched swap) quote support** by changing `quoteRequest` state and polling inputs from a single request to an array, allowing callers to update a specific request via new `updateBridgeQuoteRequestParams(…, quoteRequestIndex, quoteRequestCount)` parameters.
> 
> Quote fetching/streaming is updated to accept multiple requests: SSE uses a new `POST /getBatchQuoteStream` path when batching, tags incoming quotes with `quoteRequestIndex`, traces via new Sentry trace name `Batch Sell Quotes Fetched`, and adjusts polling stop/refresh logic to continue as long as *any* request is sufficiently funded.
> 
> Selectors and rate lookup are extended for batching: adds `selectBatchSellQuotes` (per-request recommended quotes plus aggregated received/fee totals), switches exchange-rate selection to `selectExchangeRateByAssetId`, and exports `isValidBatchSellQuoteRequest`; tests/snapshots are updated and a new SSE batch test is added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e2a7fac51ccd55190ab84d51465a81f477ee680. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


micaelae:
@metamaskbot publish-preview

packages/bridge-controller/src/utils/fetch.ts
```diff
@@ -288,12 +302,26 @@ export const fetchAssetPrices = async (
...
+  srcTokenAddress,
+  destTokenAddress,
+}: QuoteRequest): string => {
+  return `${formatAddressToAssetId(srcTokenAddress, srcChainId)}-${formatAddressToAssetId(destTokenAddress, destChainId)}`.toLowerCase();
+};
+const getQuoteResponseId = ({
+  srcAsset: { address: srcTokenAddress, chainId: srcChainId },
+  destAsset: { address: destTokenAddress, chainId: destChainId },
+}: QuoteResponse['quote']): string => {
+  return `${formatAddressToAssetId(srcTokenAddress, srcChainId)}-${formatAddressToAssetId(destTokenAddress, destChainId)}`.toLowerCase();
```
infiniteflower: Do we need to worry that non-EVM addresses can be case sensitive here?

packages/bridge-controller/src/utils/fetch.ts
```diff
@@ -323,25 +351,42 @@ export async function fetchBridgeQuoteStream(
   },
   clientVersion?: string,
 ): Promise<void> {
-  const queryParams = formatQueryParams(request);
+  const isBatchRequest = quoteRequests.length > 1;
```
infiniteflower: Do we have any issues with any overlap this terminology has with the existing `submitBatch` terminology?

packages/bridge-controller/src/types.ts
```diff
@@ -358,7 +363,7 @@ export enum BridgeBackgroundAction {
 }
 
 export type BridgeControllerState = {
-  quoteRequest: Partial<GenericQuoteRequest>;
+  quoteRequest: Partial<GenericQuoteRequest>[];
   quotes: (QuoteResponse & L1GasFees & NonEvmFees)[];
```
infiniteflower, line 367: Interesting, so `quotes` is still a single dimensional array? For batch quote requests does each request get single or multiple quotes?

micaelae: Yea the backend returns a stream of all of them, which we match up with the quoteRequest as they arrive. There can be multiple quotes for each request but the batch selector only returns the best one for each

packages/bridge-controller/src/utils/fetch.ts
micaelae: the backend uses "batch" for the new endpoint too so it seemed like a good idea to keep referring to this feature in the same way. I do see how it could be confusing though. `addTransactionBatch` submission in the status controller also doesn't always mean that the user is submitting multiple quotes

packages/bridge-controller/src/utils/fetch.ts
micaelae: Good callout! removed the lowercasing

micaelae:
@metamaskbot publish-preview

infiniteflower (approved):

PR #8732, opened 2026-05-07
feat(snap-account-service): add migration logic + keyring v2 support

ccharly:
## Explanation

Add migration logic to migrate from the legacy Snap keyring to the new architecture with 1 Snap keyring v2 per Snaps.

Also enhance `:ensureReady` so consumers can safely call `:withKeyringV2` to get the associated Snap keyring v2 instance before using it.

## References

N/A

## Checklist

---

> [!NOTE]
> **High Risk**
> One-time keyring migration and removal of the legacy Snap keyring on unlock can affect all Snap-managed accounts; breaking messenger/API changes (`getLegacySnapKeyring`, `withKeyring` → `withKeyringV2`) touch core wallet keyring flows.
> 
> **Overview**
> **Migrates Snap account keyring architecture** from one legacy global Snap keyring (v1) to **one `SnapKeyring` v2 per Snap**, bumping `@metamask/eth-snap-keyring` to `^22.3.0`.
> 
> **`SnapAccountService`** replaces **`getLegacySnapKeyring`** with **`ensureMigrated`**, which runs on **`KeyringController:unlock`** (idempotent, single shared promise): legacy accounts are grouped by Snap ID, new v2 keyrings are added, then the legacy keyring is removed. **`ensureReady`** now awaits migration, **creates a v2 keyring** for the Snap if missing, then waits only for the Snap platform (no more waiting for a global Snap keyring in **`SnapPlatformWatcher`**). **`handleKeyringSnapMessage`** and **selected-account forwarding** route through **`withKeyringV2` / `withKeyringV2Unsafe`** per Snap (including **`GetSelectedAccounts`** filtered by **`hasAccount`**).
> 
> **`multichain-account-service`** Snap providers (**BTC/SOL/TRX** and **`SnapAccountProvider`**) drop **`KeyringController:withKeyring`** in favor of **`withKeyringV2`** with a Snap-type + **`snapId`** filter; account lifecycle uses v2 APIs (**`deleteAccount(id)`**, **`createAccount` / `createAccounts`** without a leading Snap ID argument). **`BaseBip44AccountProvider`** removes the v1 **`withKeyring`** helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c30aceec7a2200454ec25a7c6253cf518d989243. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -296,11 +316,130 @@ export class SnapAccountService {
...
+   * await the same promise.
+   *
+   * @returns A promise that resolves when the migration is complete.
+   */
+  async migrate(): Promise<void> {
+    if (!this.#migratePromise) {
+      this.#migratePromise = this.#migrate();
+    }
+    return await this.#migratePromise;
+  }
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/8732/commits/2d2d37bc10ffee549cce3843d5d51b7f1f122ecf

packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -296,11 +320,139 @@ export class SnapAccountService {
...
+        for (const state of states.values()) {
+          log(`Migrating accounts for Snap "${state.snapId}"...`);
+          await controller.addNewKeyring(
+            // IMPORTANT: The Snap keyring (v2) can also be used as a v1
+            // keyring. So the builder associated with the v2 keyring type is
+            // able to build both v1 and v2 keyrings.
+            KeyringType.Snap,
+            state,
+          );
+        }
```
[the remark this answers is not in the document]
ccharly: That seems incorrect.

The migration runs atomically. If it fails AFTER adding some keyrings and BEFORE removing the legacy one, everything will get rolled back, so we are ok!

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/snap-account-service/src/index.ts
```diff
@@ -9,7 +9,6 @@ export type {
 export type {
   SnapAccountServiceEnsureReadyAction,
   SnapAccountServiceGetSnapsAction,
-  SnapAccountServiceGetLegacySnapKeyringAction,
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/8732/commits/61b84463a2e4cb62fc393c135d1d8e33e924dd0b

ccharly:
@metamaskbot publish-preview

packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -296,56 +318,210 @@ export class SnapAccountService {
...
+    snapId: SnapId,
+    operation: (keyring: SnapKeyring) => Promise<Result>,
+  ): Promise<Result> {
+    return this.#messenger.call(
+      action,
+      {
+        filter: (keyring): keyring is SnapKeyring =>
+          isSnapKeyring(keyring) && keyring.snapId === snapId,
       },
+      async ({ keyring }) => operation(keyring as SnapKeyring),
```
hmalik88, line 541: why do we need the cast if we're using the type predicate in the filter?

packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -296,56 +318,210 @@ export class SnapAccountService {
     if (!this.#tracker.canUse(snapId)) {
       throw new Error(`Unknown snap: "${snapId}"`);
     }
+
+    // Migrate from the global v1 Snap keyring to the per-Snap v2 keyring
+    // before doing anything else.
+    await this.migrate();
```
hmalik88: can we not do this in the constructor? doing it anytime a snap provider gets used doesn't feel right even though it's a no-op.

ccharly: Well.. I put it there to ensure the migration would be run no matter who requires a Snap keyring.

Maybe I could try to:
- Rename `migrate` to `ensureMigrated`
- Hook `KeyringController:unlock` instead
- Run the migration right after the unlock (which could be a noop)

This way, we could guarantee the migration has been ran correctly.

We could add a guard like `if (!this.migrated) { throw new Error('Not migrated yet'); }` in case a `:ensureReady` comes before (but I don't think that would happen with the `:unlock` hook honestly)

packages/snap-account-service/src/SnapAccountService.ts
ccharly: I think we lose the type generics when using actions (that use generics) through the messenger

packages/snap-account-service/src/SnapAccountService.ts
hmalik88: Ok nice, that sounds good

packages/snap-account-service/src/SnapAccountService.ts
hmalik88: Ah yeah I figured, all good! I wonder if the typing can be improved in the messenger so we don't lose it, I imagine there would be so many castings dropped across all teams if so haha 😆

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -336,109 +359,207 @@ export class SnapAccountService {
...
+            KeyringType.Snap,
+            state,
+          );
         }
 
-        // The legacy Snap keyring is not compatible with `EthKeyring`, so we need to cast here.
-        return { snapKeyring } as unknown as Result;
+        // Remove the legacy Snap keyring after migration.
+        log('Removing legacy Snap keyring...');
+        await controller.removeKeyring(legacySnapKeyringEntry.metadata.id);
```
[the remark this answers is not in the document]
ccharly: That should really not be possible. All accounts coming out of the legacy `SnapKeyring` always have a `metadata.snap` (even though, type-wise, it's not enforced, that's true).

I guess though, if the Snap account tracker does not track the same accounts than the legacy keyring, then yes, we could lose accounts in the process.

That should never happen though...

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

hmalik88 (dismissed):

hmalik88 (approved):

PR #8735, opened 2026-05-07
fix(transaction-pay-controller, transaction-controller): support Predict same-chain withdraw on swap-only Relay routes

dan437:
## Explanation

Since the Polymarket pUSD migration on April 28 2026, Predict same-chain withdraws on Polygon (e.g. pUSD → USDC) sometimes fail with `transaction gas limit too high (cap: 33554432, tx: 42121000)`.

Two interacting bugs:

1. **`transaction-pay-controller`** — `calculateSourceNetworkCost` always uses `fromOverride = Safe proxy` for Predict withdraws. That works for deposit-style Relay routes but breaks for the swap-aggregator routes that same-chain destinations now use post-migration: DEX aggregators reject contract callers (anti-MEV `msg.sender == tx.origin` checks etc.), so the swap step reverts in simulation.
2. **`transaction-controller`** — `estimateGasBatch`'s 7702 path always calls `estimateGas`, even when every batch entry already has a `gas` value. The non-7702 path has had an `allTransactionsHaveGas` short-circuit since [#7405](https://github.com/MetaMask/core/pull/7405); the 7702 path was missed. When TPC tries the EOA `from`, the swap sub-call reverts (EOA has no source token until the prepended `Safe.execTransaction` sub-call runs mid-batch), the whole 7702 batch falls back to ~block-gas-limit × buffer ≈ 42M, and Polygon's per-tx `gascap` rejects it.

### Fix

- **TPC** `relay-quotes.ts`: gate `fromOverride = Safe proxy` on the route having a `deposit` step. Same-chain swap routes use the relay params' EOA `from` instead. Gas-fee-token lookup keeps using the Safe proxy (gated on `isPredictWithdraw && refundTo`) so the gasless flow still works for users with 0 POL.
- **TPC** `constants.ts`: add `POLYGON_PUSD_ADDRESS` and include it in `STABLECOINS[Polygon]` for correct stablecoin display.
- **TC** `gas.ts`: mirror the existing non-7702 short-circuit in the 7702 branch — when every transaction has a provided `gas`, sum them and skip simulation.

Verified end-to-end on mobile (yalc-linked): Polygon pUSD → USDC withdraw produces a sane fee both with POL and gasless (paid in pUSD). BNB cross-chain unchanged.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes gas estimation behavior for EIP-7702 batches and alters Relay quote simulation `from` selection for Predict withdraws; mistakes could impact fee estimation and transaction success for these flows.
> 
> **Overview**
> Fixes Predict same-chain withdraws on swap-only Relay routes by **gating the Safe-proxy `from` override** to only routes that include a `deposit` step, while still using the Safe proxy for gas-fee-token eligibility checks.
> 
> Updates `estimateGasBatch`’s EIP-7702 path to **fallback to summed caller-provided per-tx `gas` limits when node simulation fails** (instead of block-gas-limit fallback), via a new `getProvidedBatchGasLimits` helper and added test coverage.
> 
> Adds `POLYGON_PUSD_ADDRESS` and includes it in Polygon `STABLECOINS` for stablecoin display/fiat-rate handling, with new constants tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c1c30aea2c53c0aec05ae516fc0d987f5f814f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-controller/src/utils/gas.ts
```diff
@@ -245,6 +277,16 @@ export async function estimateGasBatch({
   }
 
   if (chainResult) {
+    const providedBatchGasLimits = getProvidedBatchGasLimits(transactions);
```
matthewwalsh0, line 281: This was intentional as the 7702 is a different transaction that can be more gas efficient since it has no intrinsic gas cost, so we could save a meaningful amount on mainnet.

However we could use this sum of the original gas values as a fallback if a 7702 estimation fails?

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

PR #8772, opened 2026-05-12
ci: Run changelog validation in a single job

Mrtenz:
## Explanation

Instead of running 83 jobs for changelog validation, we can run it in a single job, which should not affect the total CI time much.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk CI-only change; main risk is reduced parallelism and potential failures if the loop/jq parsing behaves differently than the previous matrix expansion.
> 
> **Overview**
> Reworks the `lint-build-test.yml` changelog validation to run in one `validate-changelogs` job instead of spawning a matrix job per workspace.
> 
> The new step reads the `child-workspace-package-names` JSON output and loops through each package, running `yarn workspace <pkg> changelog:validate` while aggregating failures into a single exit code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54f884dc0a220af75ad64dfa6a80931b32295b5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -40,7 +40,6 @@ jobs:
           - lint:teams
           - messenger-action-types:check
           - readme-content:check
-          - changelog:validate
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v2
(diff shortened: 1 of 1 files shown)


FrederikBolding (approved):

.github/workflows/lint-build-test.yml
```diff
@@ -58,21 +58,31 @@ jobs:
...
+          CHILD_WORKSPACE_PACKAGE_NAMES: ${{ needs.prepare.outputs.child-workspace-package-names }}
+        run: |
+          EXIT_CODE=0
+          for package_name in $(echo "$CHILD_WORKSPACE_PACKAGE_NAMES" | jq --raw-output '.[]'); do
+            echo "Validating changelog for $package_name"
+            if ! yarn workspace "$package_name" run changelog:validate; then
+              EXIT_CODE=1
+            fi
+          done
+          exit "$EXIT_CODE"
```
mcmire, line 85: If I recall correctly, the reason why we split this script up was because if any one package had an changelog issue, it was really hard to find it based on the job output.

That said, I now see that if you pass `--verbose` to `yarn workspaces foreach`, it prints output for a workspace pessimistically. So it will only show errors and will hide everything else. In fact, this is already what `yarn changelog:validate` is doing.

Furthermore `yarn workspaces foreach` will exit with 1 if any of the subcommands fails. Which is what you seem to be doing here too 🤔 

So... maybe we can add `changelog:validate` to the `lint` command and also to the `lint` step above? And then we can remove this `validate-changelog` job altogether.

PR #8820, opened 2026-05-14
feat(compliance-controller): add configured Compliance API support

geositta:
## Explanation

This updates `@metamask/compliance-controller` so clients can pass an explicit Compliance API URL into `ComplianceService`, allowing Extension and Mobile to source the endpoint from build configuration instead of hardcoding environment switches in application code. If no `apiUrl` is provided, the service keeps `env` as a fallback and defaults to production.

This also moves the blocked-wallet lookup behavior into the shared controller package. Cached EVM wallet compliance statuses now exact-match first, then fall back to case-insensitive matching for valid `0x` EVM addresses only; non-EVM addresses remain exact-match only. The package now also exports `selectAreAnyWalletsBlocked` so clients can share the same selector behavior.
## References

Related to Extension and Mobile compliance controller integration [work](https://github.com/MetaMask/metamask-extension/pull/42520). Client adoption can follow after this package change is released and consumed.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes affect OFAC compliance API routing and how blocked-wallet cache hits are resolved (including EVM casing), which can change enforcement behavior if misconfigured.
> 
> **Overview**
> **`ComplianceService`** now accepts an optional **`apiUrl`** (validated, optional base path, no query/fragment) so clients can point at build-configured endpoints; **`env`** is optional and defaults to **production** when **`apiUrl`** is omitted. Request paths are built with `URL` relative segments so custom base paths work.
> 
> **`getWalletComplianceStatus`** centralizes cache lookups: exact match first, then case-insensitive match for valid **`0x`** EVM addresses only. **`selectIsWalletBlocked`**, API-failure fallbacks in **`ComplianceController`**, and the new **`selectAreAnyWalletsBlocked`** export all use this helper. Docs and changelog reflect **`apiUrl`** and the removed bulk blocked-wallets example.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9d9a5ca05af1a18f05826b3706fd1f55a022e78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


abretonc7s (changes requested):
Is there a matching mobile PR? It should be split the compliance controller only on this PR. And separately sync mobile

abretonc7s (approved):

Issue #8780, opened 2026-05-12
wallet-cli: Add `mm wallet unlock` command


sirtimid:
Surfaced as a follow-up from PR #8446.

### Problem

Once #8776 lands, a daemon started with \`--password\`/\`MM_WALLET_PASSWORD\` will auto-unlock on subsequent runs. But the daemon can also be started without those flags (the env-var path makes them optional in practice), in which case the persisted vault stays locked and there's no CLI affordance to unlock it.

There's also the general case of a daemon that was unlocked, then locked manually via \`mm daemon call KeyringController:setLocked\`, and now needs to be unlocked again.

### Proposed shape

A new command at \`packages/wallet-cli/src/commands/wallet/unlock.ts\`:

\`\`\`text
\$ mm wallet unlock [--password \<pw\>]
\`\`\`

- \`--password\` flag with \`env: 'MM_WALLET_PASSWORD'\` fallback.
- If neither is supplied, prompt interactively via \`@inquirer/password\` (and document the \`prompts.ts\` helper pattern from \`packages/wallet-cli/src/daemon/prompts.ts\`).
- Dispatches \`KeyringController:submitPassword\` over the existing daemon socket via \`sendCommand\`.
- Surfaces wrong-password / no-vault errors with friendly messages (mirror \`mm daemon call\`'s ENOENT / ECONNREFUSED hint pattern).

This is the first command outside the \`daemon:\` topic, so the \`mm wallet\` topic itself needs to be set up in the oclif config — small but non-zero scaffolding.

### Acceptance

- \`mm wallet unlock --password \<pw\>\` succeeds against a running, locked daemon and \`mm daemon call AccountsController:listAccounts\` works afterwards.
- \`mm wallet unlock\` (no flag, no env) prompts and uses the supplied password.
- Wrong-password is reported as a single-line error, not a stack trace.
- Daemon-not-running is reported with the same friendly hint \`mm daemon call\` already uses.
- Test coverage matching the bar set by the other commands.

### Context

- Companion to #8776 (auto-unlock on subsequent daemon starts). The two together close the "daemon restart leaves wallet locked" gap from PR #8446.

sirtimid:
Done on  https://github.com/MetaMask/core/pull/8821

PR #8821, opened 2026-05-14
feat(wallet-cli): Auto-unlock keyring on daemon restart + add `mm wallet unlock`

### LINKED ISSUE #8776: wallet-cli: Auto-unlock the keyring on subsequent daemon starts
Today the wallet-cli daemon imports the SRP on first run (which leaves the keyring unlocked) but on subsequent runs it just hydrates state from \`<dataDir>/wallet.db\` and constructs the \`Wallet\`. The persisted \`KeyringController.vault\` is reused, but \`KeyringController.isUnlocked\` is \`persist: false\` and defaults back to \`false\`.

Net effect: after a daemon restart the wallet starts **locked**, and any messenger action that touches keyring-bound state (\`AccountsController:listAccounts\`, signing, etc.) fails. The daemon already has the password from the \`--password\` flag / \`MM_WALLET_PASSWORD\` env var, so the user can't actually use a restarted daemon without manual intervention.

### Proposed fix

In \`packages/wallet-cli/src/daemon/wallet-factory.ts\`, after \`subscribeToChanges\` and the existing \`if (wasFirstRun)\` branch, call \`KeyringController:submitPassword\` on the not-first-run path:

\`\`\`ts
if (wasFirstRun) {
  await importSecretRecoveryPhrase(wallet, password, srp);
} else {
  await wallet.messenger.call('KeyringController:submitPassword', password);
}
\`\`\`

### Acceptance

- A daemon that was started once, killed, and started again responds successfully to \`mm daemon call AccountsController:listAccounts\` without re-importing the SRP.
- Wrong-password case is surfaced as a clean startup error (likely propagated from \`submitPassword\`) rather than a wedged daemon.
- New test in \`wallet-factory.test.ts\` covering both branches.
- Optional follow-up: add a \`mm daemon unlock\` command for daemons started without \`--password\`.

### Context

- Surfaced during PR #8446 review as the single biggest user-visible gap in the CLI today.

sirtimid:
## Summary

Resolves #8776 (auto-unlock the keyring on subsequent daemon starts) and #8780 (add `mm wallet unlock`) — the two halves of the "daemon restart leaves the wallet locked" gap.

## Changes

**#8776 — Auto-unlock on subsequent daemon starts**
- `wallet-factory` unlocks the reused vault via `KeyringController:submitPassword` when a password is supplied, so keyring-bound messenger actions (`AccountsController:listAccounts`, signing, etc.) work immediately after a restart instead of failing against a locked keyring.
- `--password` / `MM_WALLET_PASSWORD` is now **optional** on `mm daemon start`. First run still requires it (enforced *before* the `Wallet` is constructed). On subsequent runs, omitting it starts the daemon locked; unlock later with `mm wallet unlock`.
- An empty-string password is normalised to "no password" at the factory boundary, so `--password ''` isn't submitted to the controller as a wrong password.
- `daemon-spawn` strips an inherited `MM_WALLET_PASSWORD` from the child env when the caller omits it, so a stray value can't override an explicit omission.

**#8780 — `mm wallet unlock`**
- New command under a `wallet` oclif topic that dispatches `KeyringController:submitPassword` over the daemon socket. `--password` flag with `MM_WALLET_PASSWORD` env fallback; prompts interactively (masked) when neither is supplied (or `--password ''`).
- Friendly errors: ENOENT/ECONNREFUSED → "Daemon is not running"; EACCES → permission-denied hint with an `MM_DAEMON_DATA_DIR` pointer; JSON-RPC failure → `Failed to unlock: <msg> (code <n>) data=<json>`.
- Adds `@inquirer/password@^5.0.5` (aligned to the version already in the tree, so the lockfile change is a single dependency line).

## Test plan

- [x] Unit tests: **311 pass, 100% coverage** maintained across the package.
- [x] Real-`Wallet` integration tests (no `@metamask/wallet` mocks) cover the on-disk lifecycle: auto-unlock on restart, subsequent-run-locked → unlock via `submitPassword`, and a wrong password rejecting without corrupting the DB (retry with the correct password succeeds).
- [x] Subprocess `daemon` e2e updated: a restart with a password now resumes **and** auto-unlocks, so the resume signature is the byte-identical persisted vault (a re-import would re-encrypt with a fresh salt) rather than the wallet coming back locked. 3/3 pass.
- [x] `yarn lint` clean (ESLint, oxfmt, constraints, dedupe); `yarn changelog:validate` clean.

---

> [!NOTE]
> **High Risk**
> Touches keyring unlock, vault persistence, and password handling across daemon spawn, factory startup, and a new CLI path that submits secrets over the socket.
> 
> **Overview**
> Closes the gap where a daemon restart left the keyring locked unless the password was always embedded in `daemon start`.
> 
> **Daemon start and wallet factory:** `--password` / `MM_WALLET_PASSWORD` are optional on `mm daemon start` after the first run. `createWallet` auto-unlocks a persisted vault via `KeyringController:submitPassword` when a password is provided; without one, the daemon stays up with a locked keyring (first run still requires a password before the `Wallet` is built). `daemon-spawn` no longer forwards a parent `MM_WALLET_PASSWORD` into the child when the caller omits a password, avoiding accidental unlock.
> 
> **New `mm wallet unlock`:** Dispatches `KeyringController:submitPassword` over the daemon socket, with `--password` / `MM_WALLET_PASSWORD` or a masked interactive prompt (`@inquirer/password`). Connection and JSON-RPC failures get the same friendly handling as other daemon commands.
> 
> Tests expand unit coverage, real-`Wallet` on-disk integration scenarios, and e2e for password-less restart plus CLI unlock (including wrong-password behavior).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3235ad9185894364ee9019f82c1473175f9aa4f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet-cli/src/commands/daemon/start.ts
```diff
@@ -18,9 +19,10 @@ export default class DaemonStart extends Command {
     }),
     password: Flags.string({
       description:
-        'Wallet password (testing only — use MM_WALLET_PASSWORD env var in production)',
+        'Wallet password (testing only — use MM_WALLET_PASSWORD env var in production). ' +
+        'Required on first run; on subsequent runs, omit to start with a locked keyring and use `mm wallet unlock`.',
       env: 'MM_WALLET_PASSWORD',
-      required: true,
+      required: false,
```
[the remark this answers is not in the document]
sirtimid: the behavior is correct

packages/wallet-cli/README.md
```diff
@@ -12,12 +12,24 @@ or
 
 ## Usage
 
-The CLI drives a long-lived background **daemon** that holds an unlocked `@metamask/wallet` in memory and exposes its messenger over a per-user Unix socket. All commands live under the `mm daemon` topic; run `mm --help` (or `mm daemon <command> --help`) for the full reference.
+The CLI drives a long-lived background **daemon** that holds a `@metamask/wallet` in memory and exposes its messenger over a per-user Unix socket. All commands live under the `mm daemon` and `mm wallet` topics; run `mm --help` (or `mm <topic> <command> --help`) for the full reference.
```
rekmarks:
```suggestion
The CLI drives a long-lived background daemon that holds a `@metamask/wallet` in memory and exposes its messenger over a per-user Unix socket. All commands live under the `mm daemon` and `mm wallet` topics; run `mm --help` (or `mm <topic> <command> --help`) for the full reference.
```

packages/wallet-cli/src/daemon/utils.ts
```diff
@@ -59,6 +59,20 @@ export function formatJsonRpcError(error: JsonRpcError): string {
...
 
+/**
+ * Treat an empty string the same as `undefined` — "no value supplied". Used
+ * to normalise `--password ''` and `MM_WALLET_PASSWORD=''` so they trigger
+ * the interactive prompt rather than sending an empty string to the keyring.
+ *
+ * @param value - The value to normalise.
+ * @returns `undefined` if `value` is `undefined` or `''`; `value` otherwise.
+ */
+export function blankToUndefined(
```
rekmarks: The docstring refers to the empty rather than a blank string, and so should the function name:
```suggestion
export function emptyToUndefined(
```

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      infuraProjectId: INFURA_PROJECT_ID,
+      log: () => undefined,
+    });
+
+    try {
+      expect(wallet.state.KeyringController?.isUnlocked).toBe(true);
+
+      // `getState` resolves synchronously; awaiting a non-thenable trips
+      // `@typescript-eslint/await-thenable`.
+      const { keyrings } = wallet.messenger.call('KeyringController:getState');
```
rekmarks: There's no await?

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Build a unique on-disk SQLite path under the OS temp dir and remember it
+ * for `afterEach` cleanup. Includes a random suffix so concurrent test runs
+ * cannot collide.
+ *
+ * @param label - A short label that makes the resulting filename traceable.
+ * @returns An absolute file path inside `os.tmpdir()`.
+ */
+function tempDbPath(label: string): string {
```
rekmarks:
```suggestion
function createTempDbPath(label: string): string {
```

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      const { keyrings } = wallet.messenger.call('KeyringController:getState');
+      expect(keyrings[0]?.accounts[0]).toMatch(/^0x[0-9a-fA-F]{40}$/u);
+    } finally {
+      await dispose();
+    }
+  });
+});
+
+describe('createWallet (integration)', () => {
+  afterEach(() => {
```
rekmarks: Shouldn't this be `afterAll`? If there are concurrent test runs, this could prematurely delete a db.

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
sirtimid: No, because the `getState` call is synchronous

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
sirtimid: Yes https://github.com/MetaMask/core/pull/8821/commits/eb5018e74c97018595396d6569e490dd9ff87813

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
sirtimid: Done https://github.com/MetaMask/core/pull/8821/commits/eb5018e74c97018595396d6569e490dd9ff87813

packages/wallet-cli/src/daemon/utils.ts
sirtimid: Ok https://github.com/MetaMask/core/pull/8821/commits/eb5018e74c97018595396d6569e490dd9ff87813

packages/wallet-cli/README.md
sirtimid: Applied https://github.com/MetaMask/core/pull/8821/commits/eb5018e74c97018595396d6569e490dd9ff87813

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
rekmarks: Right but what does the comment mean?

rekmarks: It looks like a comment you'd after before an `eslint-disable` directive but there's no such directive, and AFAICT no reason to expect an await in the first place.

sirtimid: it means that if we await it it will be rejected since we have the `@typescript-eslint/await-thenable` ESLint rule, but I guess it's over-explaining and it's confusing since you mislead it for a `eslint-disable` directive, I removed it 3235ad918

rekmarks (approved):

PR #8838, opened 2026-05-18
feat: Implement wallet initialization library

FrederikBolding:
## Explanation

This PR implements a narrowly-scoped (as compared to the original feature branch) version of the wallet initialization library that only includes initializing the `KeyringController`. This can eventually be used to demonstrate the integration of the library into the clients and serves as the base for future work.

Overall it works in a similarly to the initialization pattern used in extension and mobile today, with some differences:
- The entities initialized by the pattern are referred to as "instances", not "messenger clients".
- It attempts to be less verbose as the initialization of an instance can be done in a single file exporting a single object, the `InitializationConfiguration`. This object contains both a function to setup the messenger and the instance.
- It has no concept of "initialization messengers", the messenger returned from `InitializationConfiguration.messenger` is expected to have access to actions/events necessary to initialize _and_ operate the instance.
- ~There is no way to access the instances directly.~

The `Wallet` instance provides access to the instances within using the `messenger` while also exposing a limited set of useful properties like `state` and `controllerMetadata`.

## References

https://consensyssoftware.atlassian.net/browse/WPC-999

## Checklist

---

> [!NOTE]
> **High Risk**
> Introduces vault encryption, keyring bootstrap, and SRP import paths—security-sensitive wallet/key material handling even though scope is limited to KeyringController.
> 
> **Overview**
> Replaces the `@metamask/wallet` placeholder with a **wallet initialization library** centered on a `Wallet` class that wires a root messenger, runs pluggable `InitializationConfiguration` entries, and exposes aggregated `state`, `controllerMetadata`, messenger access, and deprecated `getInstance`.
> 
> The default stack currently boots **`KeyringController` only**, via `initialize()` merging default configs with optional overrides by name, per-instance messengers, hydrated `state`, and `instanceOptions` (custom encryptor / keyring builders). Keyring setup adds a **PBKDF2 (600k iterations) `encryptorFactory`** around `browser-passworder` and exports `importSecretRecoveryPhrase` for SRP restore through messenger actions.
> 
> Package surface: real **runtime dependencies** (`base-controller`, `keyring-controller`, `messenger`, etc.), TS project references, changelog “initial release”, README dependency graph edges, and **CODEOWNERS** for `keyring-controller` initialization. Tests cover wallet lifecycle, immutability guards, vault unlock/lock, encryptor behavior, and config overrides.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19bfd385f6f3969221d431ba2744c5df5b0f9a47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


FrederikBolding:
@metamaskbot publish-previews

FrederikBolding:
@metamaskbot publish-previews

packages/wallet/src/initialization/index.ts
mcmire, line 1: We have been advising teams not to add wildcard ("barrel") files. My understanding has been that it can increase the potential workflow of tools like ESLint (because it increases the number of routes these tools can take to find a file). There is an article I found (amazingly, from Atlassian) that goes into more detail here: https://www.atlassian.com/blog/atlassian-engineering/faster-builds-when-removing-barrel-files

What are your thoughts on this, is it possible to import these files directly where they are needed?

packages/wallet/src/initialization/types.ts
```diff
@@ -0,0 +1,29 @@
...
+export type InitFunctionArguments<Instance, InstanceMessenger> = {
+  state: InstanceState<Instance>;
+  messenger: InstanceMessenger;
+  options: InstanceOptions<Instance>;
+};
+
+export type InitializationConfiguration<Instance, InstanceMessenger> = {
+  name: InstanceName<Instance>;
+  init(args: InitFunctionArguments<Instance, InstanceMessenger>): Instance;
+  messenger(parent: RootMessenger): InstanceMessenger;
```
mcmire: What do you think about calling this method `getMessenger` instead of just `messenger`?

``` suggestion
  getMessenger(parent: RootMessenger): InstanceMessenger;
```

packages/wallet/src/Wallet.ts
```diff
@@ -0,0 +1,97 @@
...
+  // TODO: Expand default types when passing additionalConfigurations.
+  readonly #messenger: RootMessenger<DefaultActions, DefaultEvents>;
+
+  readonly #instances: DefaultInstances;
+
+  readonly #controllerMetadata: Readonly<
+    Record<string, Readonly<StateMetadataConstraint>>
+  >;
+
+  #destroyed = false;
```
mcmire: Nit: `isDestroyed`?

``` suggestion
  #isDestroyed = false;

File: packages/wallet/src/Wallet.ts [hunk 3]
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    this.#controllerMetadata = Object.fromEntries(
+      Object.entries(this.#instances)
+        .filter(([_, instance]) => hasProperty(instance, 'metadata'))
+        .map(([name, instance]) => [name, instance.metadata]),
+    );
+  }
+
+  get messenger(): Readonly<RootMessenger<DefaultActions, DefaultEvents>> {
```
mcmire, line 60: I am uneasy about using getters because I feel like if something is a method then it should be obvious when calling it, rather than hiding it behind property syntax. Are we using a getter, however, for compatibility with controllers/services?

mcmire:
@FrederikBolding 

> The entities initialized by the pattern are referred to as "instances", not "messenger clients".

Hmm. I wonder if "instances" sounds a bit too vague. It's true they are instances of classes, but then so are lots of other things. 

For context, we discussed a unified term for controllers and services in this PR: https://github.com/MetaMask/decisions/pull/41#discussion_r1809429486. I had some ideas such as "messaging actor" (or just "actor"), but I think "messenger client" was the least worst option (and the one that Mark also agreed upon). "Messageable" was also a contender. Any of these options appeal to you?

packages/wallet/src/initialization/instances/keyring-controller.ts
```diff
@@ -0,0 +1,158 @@
...
+  name: 'KeyringController',
+  init: ({ state, messenger, options }) =>
+    new KeyringController({
+      state,
+      messenger,
+      keyringBuilders: options.keyringBuilders,
+      encryptor: options.encryptor ?? encryptorFactory(600_000),
+    }),
+  messenger: (parent) =>
+    new Messenger<'KeyringController', never, never, typeof parent>({
```
mcmire: It would be nice to not have to repeat `parent` three times, but alas...

packages/wallet/src/initialization/defaults.ts
```diff
@@ -0,0 +1,57 @@
...
+  payload: [];
+};
+
+export type DefaultEvents =
+  | MessengerEvents<AllMessengers>
+  | WalletDestroyedEvent;
+
+export type RootMessenger<
+  AllowedActions extends ActionConstraint = ActionConstraint,
+  AllowedEvents extends EventConstraint = EventConstraint,
```
mcmire: Does this need to take type parameters? It would be nice if `RootMessenger` could never be unconstrained. Would it make sense to default these type parameters to `DefaultActions` and `DefaultEvents`? Or maybe we don't need to have defaults at all?

``` suggestion
  AllowedActions extends ActionConstraint,
  AllowedEvents extends EventConstraint,

File: packages/wallet/src/initialization/types.ts [hunk 6]
```diff
@@ -0,0 +1,29 @@
...
+  ? Name
+  : string;
+
+type InstanceOptions<Instance> =
+  InstanceName<Instance> extends keyof InstanceSpecificOptions
+    ? NonNullable<InstanceSpecificOptions[InstanceName<Instance>]>
+    : unknown;
+
+export type InitFunctionArguments<Instance, InstanceMessenger> = {
+  state: InstanceState<Instance>;
```
mcmire: Hmm. Services don't have state. So does this make sense as an argument to all init functions?

(Edit: I guess init functions for services don't need to care about this, is that the thought?)

packages/wallet/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import type { InitializationConfiguration } from './initialization';
+
+export type WalletOptions = {
+  state?: Record<string, Record<string, Json>>;
+  initializationConfigurations?: InitializationConfiguration<
+    unknown,
+    unknown
+  >[];
+  instanceOptions?: InstanceSpecificOptions;
```
mcmire, line 19: I'm a bit concerned that engineers will get confused on the difference between `initializationConfigurations` and `instanceOptions`.

Is it true that:

1. `initializationConfigurations` configures an array of instances which are added on the default set of instances?
2. `instanceOptions` configures constructors for default instances?

If so should we rename these options to something like this?

``` suggestion
  additionalInitializationConfigurations?: InitializationConfiguration<
    unknown,
    unknown
  >[];
  optionsForDefaultInstances?: InstanceSpecificOptions;
```

mcmire (commented):
Going in a good direction. Here are some thoughts and comments.

packages/wallet/src/Wallet.ts
FrederikBolding: I was thinking we would do something similar to `BaseController` where we create setters that throw so that the messenger, state etc aren't assigned to.

packages/wallet/src/initialization/instances/keyring-controller.ts
FrederikBolding: The generic parameters aren't actually necessary, so we can remove one instance 😄

packages/wallet/src/initialization/types.ts
FrederikBolding: > I guess init functions for services don't need to care about this, is that the thought

Essentially yes. Though we may be able to tighten the typing a bit by making `InstanceState` be `null` or something for services.

packages/wallet/src/initialization/types.ts
FrederikBolding: Yeah, that could make sense!

packages/wallet/src/initialization/defaults.ts
FrederikBolding: Eventually it'll be useful to expand the types beyond the defaults, if possible. I removed the defaults for now, but they could probably be `DefaultActions` and `DefaultEvents`.

FrederikBolding:
@metamaskbot publish-previews

FrederikBolding:
> @FrederikBolding
> 
> > The entities initialized by the pattern are referred to as "instances", not "messenger clients".
> 
> Hmm. I wonder if "instances" sounds a bit too vague. It's true they are instances of classes, but then so are lots of other things.
> 
> For context, we discussed a unified term for controllers and services in this PR: [MetaMask/decisions#41 (comment)](https://github.com/MetaMask/decisions/pull/41#discussion_r1809429486). I had some ideas such as "messaging actor" (or just "actor"), but I think "messenger client" was the least worst option (and the one that Mark also agreed upon). "Messageable" was also a contender. Any of these options appeal to you?

@mcmire I haven't spent much time thinking about this, when the prototyping started I don't believe we had adopted "messenger client" on the _MetaMask clients_ yet. If that already is our decided preferred naming, I can make changes, but "messenger client" seems similarly vague and maybe even an overloaded term considering API clients, the MetaMask clients themselves (extension/mobile) etc.

packages/wallet/src/types.ts
FrederikBolding: The idea was that you would use `initializationConfigurations` for either _overriding_ any default configuration or adding a client-specific configuration. E.g. for controllers that only live in one of the clients.

The `instanceOptions` are for any options that cannot be standardized across clients, for example `keyringBuilders`, `encryptor`, `infuraProjectId` etc.

`optionsForDefaultInstances` isn't technically true since you could use it to pass options to a configuration specified in `initializationConfigurations`. Though I guess that is a bit pointless 😅 

`additionalInitializationConfigurations` seems like confusing naming given the override behaviour mentioned above.

packages/wallet/src/initialization/index.ts
FrederikBolding: For the default configurations, we somewhat need a wildcard import: https://github.com/MetaMask/core/blob/ea54cecca2f7d08bea63bb5fc42a3787391df4bd/packages/wallet/src/initialization/defaults.ts#L9

Unless we want to have an ever-growing list of imports here.

Though it isn't strictly required anywhere else, if that is the recommendation, I'm happy to follow.

packages/wallet/src/initialization/types.ts
FrederikBolding: https://github.com/MetaMask/core/pull/8838/commits/a04182d08ff69e22c6ae0bc7163e87a6341f58b3

packages/wallet/src/initialization/types.ts
FrederikBolding: https://github.com/MetaMask/core/pull/8838/commits/b6651eda23f5974dcdb7c09e3f390b32eb3262ba

FrederikBolding:
@metamaskbot publish-previews

packages/wallet/src/initialization/index.ts
mcmire: I think it's fine to make an exception for `packages/wallet/src/initialization/instances/index.ts`, I can see the benefit there.

mcmire:
> I haven't spent much time thinking about this, when the prototyping started I don't believe we had adopted "messenger client" on the _MetaMask clients_ yet. If that already is our decided preferred naming, I can make changes, but "messenger client" seems similarly vague and maybe even an overloaded term considering API clients, the MetaMask clients themselves (extension/mobile) etc.

Yeah, it's not not a perfect name for sure. I purposefully avoided using the term "messenger client" in my presentation for similar reasons. I don't think we have adopted the term widely, so we could come up with a different one and no one would bat an eye.

I guess we could go with "instance" for now and maybe something else will pop out later.

packages/wallet/src/Wallet.ts
mcmire: I see, alright.

packages/wallet/src/types.ts
mcmire: Okay, gotcha, thanks for the clarification, I understand this better now.

mcmire:
@FrederikBolding This looks pretty good to me so far. Let me know when this is out of draft.

FrederikBolding:
@metamaskbot publish-previews

FrederikBolding:
@metamaskbot publish-previews

packages/wallet/src/initialization/defaults.ts
```diff
@@ -0,0 +1,57 @@
...
+  MessengerActions,
+  MessengerEvents,
+} from '@metamask/messenger';
+
+import * as defaultConfigurations from './instances';
+import type { InitializationConfiguration, InstanceState } from './types';
+
+export { defaultConfigurations };
+
+type ExtractInstance<Config> =
```
Mrtenz, line 17: Would be helpful to have some docs for these types, since it may not be immediately obvious what some of these do.

packages/wallet/src/initialization/types.ts
```diff
@@ -0,0 +1,40 @@
+import type { InstanceSpecificOptions } from '../types';
+import type { DefaultActions, DefaultEvents, RootMessenger } from './defaults';
+
+export type InstanceState<Instance> = Instance extends { state: unknown }
```
Mrtenz, line 7: Same here re: documentation.

packages/wallet/src/utilities.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param password - The password to the MetaMask wallet (not the SRP).
+ * @param phrase - The SRP as a string.
+ */
+export async function importSecretRecoveryPhrase(
+  wallet: Wallet,
+  password: string,
+  phrase: string,
+): Promise<void> {
+  const indices = phrase.split(' ').map((word) => wordlist.indexOf(word));
+  const mnemonic = new Uint8Array(new Uint16Array(indices).buffer);
```
Mrtenz, line 18: IIRC `key-tree` has a util function for this.

packages/wallet/src/Wallet.test.ts
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  it('omits instances without a metadata property from controllerMetadata', async () => {
+    const fakeMetadata = {
+      foo: { persist: true, includeInDebugSnapshot: false },
+    };
+    jest.spyOn(initializationModule, 'initialize').mockReturnValueOnce({
+      WithMeta: { state: {}, metadata: fakeMetadata },
+      NoMeta: { state: {} },
+    } as never);
```
Mrtenz: `// @ts-expect-error`?

packages/wallet/src/Wallet.ts
```diff
@@ -0,0 +1,144 @@
...
+          // @ts-expect-error Accessing protected property.
+          // eslint-disable-next-line @typescript-eslint/await-thenable
+          return await instance.destroy();
+        }
+        /* istanbul ignore next */
+        return undefined;
+      }),
+    );
+
+    this.messenger.publish('Root:destroyed');
```
Mrtenz: What's the use case for this?

packages/wallet/src/Wallet.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * @returns The root messenger of the wallet.
+   */
+  get messenger(): Readonly<RootMessenger<DefaultActions, DefaultEvents>> {
+    return this.#messenger;
+  }
+
+  set messenger(_) {
+    throw new Error('The messenger cannot be directly mutated.');
```
Mrtenz, line 65: Do we need to throw for these? Typically in JS trying to assign to a non-writable property is just a no-op.

FrederikBolding: We don't strictly need to do it, but we are already doing this for `BaseController`. So I was assuming we wanted it as developer guidance.

No strong opinion on whether to keep it though, only added it very recently.

packages/wallet/src/Wallet.ts
FrederikBolding: It comes from the feature branch, I believe it may have been needed for the CLI but I can't find where it is used right now.

Can remove and we can re-add if necessary

packages/wallet/src/Wallet.ts
Mrtenz: Fine with me to keep it if we do it elsewhere.

packages/wallet/src/Wallet.ts
Mrtenz: Just seems strange, since the controllers that could listen to it have already been destroyed, and the caller of `Wallet.destroy` can just await that. 😅 So I think we could remove this unless there's a good use case for it.

FrederikBolding: Removed: https://github.com/MetaMask/core/pull/8838/commits/a2c4a8c3a3ceddfdc4403743f75a702e3e40fbef

packages/wallet/src/initialization/types.ts
FrederikBolding: I added documentation for some of them, any specific ones you think are still confusing? https://github.com/MetaMask/core/pull/8838/commits/a2c4a8c3a3ceddfdc4403743f75a702e3e40fbef

packages/wallet/src/initialization/defaults.ts
FrederikBolding: https://github.com/MetaMask/core/pull/8838#discussion_r3281303653

packages/wallet/src/Wallet.ts
```diff
@@ -0,0 +1,144 @@
...
+  DefaultActions,
+  DefaultEvents,
+  DefaultInstances,
+  DefaultState,
+  RootMessenger,
+} from './initialization/defaults';
+import { initialize } from './initialization/initialization';
+import { WalletOptions } from './types';
+
+export class Wallet {
```
mcmire, line 15: Nit: Would be good to add JSDoc that briefly describes the purpose of this class and provides an example. Same for the README. This can be done in a future PR, however.

packages/wallet/jest.config.js
```diff
@@ -18,9 +18,9 @@ module.exports = merge(baseConfig, {
   coverageThreshold: {
     global: {
       branches: 100,
-      functions: 100,
-      lines: 100,
-      statements: 100,
+      functions: 89.65,
```
mcmire: Is it possible to reach 100%?

packages/wallet/src/initialization/instances/keyring-controller.ts
```diff
@@ -0,0 +1,183 @@
...
+  KeyringController,
+  KeyringControllerMessenger
+> = {
+  name: 'KeyringController',
+  init: ({ state, messenger, options }) =>
+    new KeyringController({
+      state,
+      messenger,
+      keyringBuilders: options.keyringBuilders,
+      encryptor: (options.encryptor ?? encryptorFactory(600_000)) as Encryptor<
```
mcmire, line 174: Nit: Why the typecast here?

packages/wallet/src/initialization/types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+type CamelCaseInstanceName<Instance> = LowerCaseFirstLetter<
+  InstanceName<Instance>
+>;
+
+/**
+ * Utility type for narrowing the InstanceSpecificOptions to just the options required for the instance.
+ */
+type InstanceOptions<Instance> =
+  CamelCaseInstanceName<Instance> extends keyof InstanceSpecificOptions
```
mcmire, line 36: I see that `instanceOptions` now takes properties which are camel-cased versions of instance names. I thought it made more sense for it to use the names directly (i.e. allow them to be PascalCased?) because I feel like this is what people will want to do naturally. However, did you change this because you didn't want them to constantly run up against our ESLint rule for property names?

mcmire (commented):
Looks basically good, I just had one question about test coverage.

packages/wallet/src/initialization/types.ts
FrederikBolding: Personally I would have preferred them as PascalCased too, but it does cause lint issues. Happy to revert, but then we'll need to disable the ESLint rule for the areas that define `instanceOptions`.

FrederikBolding:
@metamaskbot publish-previews

packages/wallet/src/initialization/instances/keyring-controller.ts
FrederikBolding: We are unfortunately forced to cast to one of the `Encryptor` types as it isn't possible to instantiate a `KeyringController` that supports them both at the same time.

It seemed OK to me to hide that detail with a cast while still preserving the ability to pass in both working `Encryptor` implementations.

packages/wallet/jest.config.js
FrederikBolding: Yes, done.

The `encryptorFactory` was holding us back because it includes some dead code that is required by the `Encryptor` type but not used in `KeyringController`. I have added tests until we can remove it.

Mrtenz (approved):

packages/wallet/src/initialization/types.ts
mcmire: Hmm, yeah, okay. I feel like somewhere we were discussing removing that rule altogether as there does seem to be a lot of exceptions. That doesn't really block this however. Your change is good for now, we can always change our mind later if need be.

packages/wallet/src/initialization/instances/keyring-controller.ts
mcmire: Okay sounds good.

packages/wallet/package.json
```diff
@@ -52,6 +52,14 @@
     "test:verbose": "NODE_OPTIONS=--experimental-vm-modules jest --verbose",
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
+  "dependencies": {
+    "@metamask/base-controller": "^9.1.0",
+    "@metamask/browser-passworder": "^6.0.0",
```
mcmire, line 57: I guess we're using this package in a way that works across platforms? Curious if it might be worth it at some point to extract the more platform-agnostic functions to some other library. Otherwise it might be a bit confusing. We can think about that later perhaps.

mcmire (approved):
LGTM!

packages/wallet/package.json
FrederikBolding: `@metamask/browser-passworder` works on most platforms except mobile, so it seemed fine to include. It's a good default option IMO.

PR #8846, opened 2026-05-18
refactor(wallet-cli): enforce RPC param validation at dispatch time via superstruct

### LINKED ISSUE #8777: wallet-cli: Parameterise RpcHandler with struct-validated dispatch

### Problem

\`RpcHandler\` (\`packages/wallet-cli/src/daemon/types.ts\`) is currently typed as \`(params: Json) =\> Promise\<Json | void\>\`. Every handler has to re-validate \`params\` from scratch — see the \`call\` handler in \`daemon-entry.ts\` doing \`Array.isArray(params) && typeof params[0] === 'string'\` before casting to \`[string, ...Json[]]\`. The next handler that needs structured params will copy-paste the same boilerplate.

The server-side \`isJsonRpcRequest\` check in \`rpc-socket-server.ts\` already validates the wire frame, but per-method param validation is still ad-hoc and the typed messenger surface is bypassed with an \`as any\` cast in the \`call\` handler.

### Proposed direction

Parameterise the handler type over its params and result, and add a small \`defineHandler\` helper that owns the struct guard:

\`\`\`ts
export type RpcHandler\<TParams extends JsonRpcParams, TResult extends Json\> =
  (params: TParams) =\> Promise\<TResult\>;

export function defineHandler\<TParams, TResult\>(
  paramsStruct: Struct\<TParams\>,
  run: (params: TParams) =\> Promise\<TResult\>,
): RpcHandler\<TParams \& JsonRpcParams, TResult \& Json\> {
  return async (params) =\> {
    const validated = assertStruct(params, paramsStruct);
    return run(validated);
  };
}
\`\`\`

Then rewrite each handler to declare its params struct (e.g. \`getStatus\` takes \`undefined\`, \`call\` takes a tuple of \`[string, ...Json[]]\`). The runtime guard runs once per request inside the server, and the handler body no longer needs the boilerplate.

### Acceptance

- All handlers in \`daemon-entry.ts\` go through the new helper.
- The \`as any\` cast on \`wallet.messenger.call\` is replaced by a typed dispatcher whose params type comes from the struct.
- \`rpc-socket-server.ts\`'s \`handleRequest\` validates params via the registered struct before calling the handler.
- Tests updated; coverage stays at 100%.

### Context

- Out of scope for #8446 because the change rewires every handler + test. Tracked as a deferred refactor in the PR description.

sirtimid:
## Summary

Resolves #8777.

- Adds `RpcHandler<TParams, TResult>`, `RpcHandlerDefinition<TParams, TResult>` (`{ paramsStruct, run }`), and a `defineHandler(paramsStruct, run)` helper in `daemon/types.ts`. `RpcHandlerMap` now maps method names to definitions.
- `rpc-socket-server.ts` validates `params` against the handler's struct via `superstruct.validate` *before* dispatch and returns `-32602 invalidParams` on shape mismatch. Each handler body can now trust the shape of its argument.
- `daemon-entry.ts`:
  - `getStatus` uses `literal(null)`.
  - `call` uses a `define`d struct that enforces a non-empty `[string, ...Json[]]` tuple, dropping the ad-hoc `Array.isArray / typeof` checks from the handler body.
  - Replaces the `wallet.messenger.call as any` cast with a single typed `RpcDispatcher = messenger.call.bind(messenger)` escape hatch.
- Adds `@metamask/superstruct ^3.1.0` to `wallet-cli` (already used across the monorepo at this version).

## Test plan

- [x] `yarn workspace @metamask/wallet-cli run test` — 248 tests pass, 100% line/branch/function/statement coverage on the touched files.
- [x] `yarn lint:eslint packages/wallet-cli/src` — clean.
- [x] `yarn constraints` — clean.
- [x] `yarn workspace @metamask/wallet-cli run changelog:validate` — clean.
- New server test covers the `-32602 invalidParams` path and verifies `run` is not invoked when the struct rejects.
- New struct tests on the `call` handler cover both the rejection cases (null, empty, non-string head, non-array) and the accepting case.

---

> [!NOTE]
> **Medium Risk**
> Changes the wallet daemon JSON-RPC dispatch path and how malformed `params` are surfaced (standard invalidParams vs handler errors), on a local socket that still exposes the full messenger surface.
> 
> **Overview**
> Daemon RPC handlers are now **`{ paramsStruct, run }` definitions** (via `defineHandler`) instead of raw functions. **`rpc-socket-server`** validates `params` with superstruct **before** calling `run` and responds with **`-32602 invalidParams`** when the shape does not match, so handlers no longer need ad-hoc checks.
> 
> **`daemon-entry`** wires **`getStatus`** / **`listActions`** with `literal(null)`, **`call`** with a struct for a non-empty `[string, ...]` action tuple, and routes messenger dispatch through a single **`RpcDispatcher`** bind instead of inline casts. Shutdown now **logs and continues** if **`dispose()`** fails during teardown.
> 
> Adds **`@metamask/superstruct`** and expands tests for invalid-params paths and struct acceptance/rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c971a5f3c7dde52b1e463c45233301023ab2b80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet-cli/src/daemon/types.ts
```diff
@@ -7,16 +8,66 @@ import type { Json } from '@metamask/utils';
...
+ * @param paramsStruct - Struct that validates `params` for this method.
+ * @param run - Handler invoked with the validated params.
+ * @returns An {@link RpcHandlerDefinition} suitable for an {@link RpcHandlerMap}.
+ */
+export function defineHandler<TParams, TResult extends Json | void>(
+  paramsStruct: Struct<TParams>,
+  run: RpcHandler<TParams, TResult>,
+): RpcHandlerDefinition<TParams, TResult> {
+  return { paramsStruct, run };
+}
```
rekmarks, line 68: Gets rid of the `any`, which is strongly discouraged:
```suggestion
 * A map of RPC method names to their handler definitions.
 *
 * `TParams` is erased to `unknown` here so definitions with different narrow
 * params types (e.g. `null` vs. a tuple) can coexist in the same map. Consumers
 * therefore see each `run` as accepting `unknown` and must validate `params`
 * against the paired `paramsStruct` before invoking it — which is exactly what
 * the server (see `rpc-socket-server.ts`) does. The concrete `TParams` is
 * captured inside {@link defineHandler}, where the struct and handler are bound
 * together.
 */
export type RpcHandlerMap = Record<
  string,
  RpcHandlerDefinition<unknown, Json | void>
>;

/**
 * Bundle a params struct with the handler that runs once `params` is
 * validated. The server invokes `run` only after `paramsStruct` accepts the
 * value, so `run` can trust the type of its argument.
 *
 * The returned definition erases `TParams` to `unknown` so heterogeneous
 * handlers can share an {@link RpcHandlerMap}.
 *
 * @param paramsStruct - Struct that validates `params` for this method.
 * @param run - Handler invoked with the validated params.
 * @returns An {@link RpcHandlerDefinition} suitable for an {@link RpcHandlerMap}.
 */
export function defineHandler<TParams, TResult extends Json | void>(
  paramsStruct: Struct<TParams>,
  run: RpcHandler<TParams, TResult>,
): RpcHandlerDefinition<unknown, TResult> {
  return { paramsStruct, run } as unknown as RpcHandlerDefinition<
    unknown,
    TResult
  >;
}
```

rekmarks (changes requested):
Looks good! Just one thing.

rekmarks:
@sirtimid actually, since we would have returned different errors prior to this work, we should add a changelog entry for the package.

packages/wallet-cli/src/daemon/types.ts
sirtimid: Applied https://github.com/MetaMask/core/pull/8846/commits/0c971a5f3c7dde52b1e463c45233301023ab2b80

rekmarks (approved):

PR #8847, opened 2026-05-19
refactor(wallet-cli): Replace { wallet, store } with { wallet, dispose }

### LINKED ISSUE #8779: wallet-cli: Replace { wallet, store } with { wallet, dispose }

### Problem

\`createWallet\` in \`packages/wallet-cli/src/daemon/wallet-factory.ts\` returns \`{ wallet, store }\`. The close-after-destroy ordering invariant is encoded only in the function's JSDoc:

\> Close \`store\` only \*\*after\*\* \`await wallet.destroy()\` has resolved; closing earlier will cause persistence writes during teardown to fail.

Today the invariant is consumed in two distinct cleanup ladders in \`packages/wallet-cli/src/daemon/daemon-entry.ts\` (the startup-failure \`catch\` block and the SIGTERM/SIGINT \`shutdown\` closure). Both implement the order correctly today, but if a third site is added — or if either ladder grows a new step (e.g. flushing the SQLite WAL) — the JSDoc isn't enforcement, and the two paths can drift.

### Proposed direction

Have \`createWallet\` return \`{ wallet, dispose }\` instead of \`{ wallet, store }\`. \`dispose\` is the single owner of teardown:

\`\`\`ts
export type CreateWalletResult = {
  wallet: Wallet;
  dispose: () =\> Promise\<void\>;
};
\`\`\`

\`dispose\` encapsulates \`await wallet.destroy()\` then \`store.close()\` (with per-step error logging) so neither call site can get the order wrong.

### Acceptance

- \`wallet-factory.ts\` no longer exposes \`store\` to consumers.
- \`daemon-entry.ts\`'s two cleanup ladders both call \`dispose\` instead of inlining \`wallet.destroy()\` + \`store.close()\`.
- Tests in \`wallet-factory.test.ts\` cover \`dispose\` semantics (idempotent, handles destroy failure, closes the store even when destroy throws).
- Coverage stays at 100%.

### Context

- Out of scope for #8446. Tracked as a deferred refactor in the PR description.

sirtimid:
## Summary

- `createWallet` in `daemon/wallet-factory` now returns `{ wallet, dispose }` instead of `{ wallet, store }`. `dispose` is the single owner of teardown — it runs `await wallet.destroy()` then `store.close()`, with per-step failures forwarded to the supplied logger (or `console.error` if none).
- The two daemon cleanup ladders in `daemon-entry.ts` (startup-failure `catch` block + SIGTERM/SIGINT `shutdown` closure) now both call `dispose` instead of inlining the order, so a future third call site or extra teardown step can't drift.
- The disposer is idempotent (concurrent and sequential calls coalesce onto a single teardown promise), closes the store even when `wallet.destroy()` rejects, and survives a throwing user-supplied logger.
- The pre-resolution catch block inside `createWallet` was tightened to route through the same reporter, so a `store.close()` throw during first-run cleanup can no longer mask the original failure (e.g. bad SRP).

Closes #8779. Builds on #8446 (the deferred refactor was tracked there).

## Test plan

- [x] `yarn workspace @metamask/wallet-cli run test` — 250 passing, 100% statements/branches/funcs/lines on `wallet-factory.ts` and `daemon-entry.ts`.
- [x] `yarn workspace @metamask/wallet-cli run build` — clean.
- [x] `yarn eslint packages/wallet-cli/src` — clean.
- [x] `yarn workspace @metamask/wallet-cli run changelog:validate` — clean (no entry; the package is unreleased and this is internal).
- New tests cover: destroy-before-close ordering, concurrent + sequential coalescence, store-closes-even-when-destroy-rejects, log routing for each step, console.error fallback when no logger supplied, console.error fallback when the supplied logger throws, and store.close throwing during first-run cleanup not masking the original error.

### DIFF
--- a/packages/wallet-cli/src/daemon/daemon-entry.test.ts
+++ b/packages/wallet-cli/src/daemon/daemon-entry.test.ts
@@ -55,7 +55,7 @@ function enoent(): NodeJS.ErrnoException {
 }
 
 /**
- * Create a mock createWallet result with a mocked wallet and store.
+ * Create a mock createWallet result with a mocked wallet and dispose callback.
  *
  * @returns A mock createWallet result.
  */
@@ -66,9 +66,7 @@ function createMockWallet(): MockCreateWalletResult {
       state: {},
       destroy: jest.fn().mockResolvedValue(undefined),
     },
-    store: {
-      close: jest.fn(),
-    },
+    dispose: jest.fn().mockResolvedValue(undefined),
   } as unknown as MockCreateWalletResult;
 }
 
@@ -306,7 +304,7 @@ describe('daemon-entry', () => {
     );
   });
 
-  it('cleans up wallet, store, and PID file when server fails to start', async () => {
+  it('disposes the wallet and removes the PID file when server fails to start', async () => {
     const result = createMockWallet();
     mockCreateWallet.mockResolvedValue(result);
     mockStartRpcSocketServer.mockRejectedValue(new Error('server failed'));
@@ -321,8 +319,7 @@ describe('daemon-entry', () => {
 
     await importDaemonEntry();
 
-    expect(result.wallet.destroy).toHaveBeenCalled();
-    expect(result.store.close).toHaveBeenCalled();
+    expect(result.dispose).toHaveBeenCalledTimes(1);
     expect(mockRm).toHaveBeenCalledWith('/tmp/daemon.pid', { force: true });
     expect(process.exitCode).toBe(1);
   });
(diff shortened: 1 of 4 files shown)


sirtimid:
Superseded by #9226, which lands `createWallet` + `daemon-entry` directly on `main` and folds in this `{ wallet, dispose }` refactor (idempotent, resilient teardown). It additionally adds the `unsubscribe` step that this branch's older `rekm/wallet-cli` base didn't need (on `main`, post-#9067 `subscribeToChanges` returns an unsubscribe fn rather than relying on a `Wallet:destroyed` event). #8779 is closed by #9226.

PR #8857, opened 2026-05-20
fix(assets-controller): improve SnapDataSource chain discovery and balance delivery

salimtb:
Read chainIds from endowment:keyring when assets permission lacks them, retry discovery on balance pushes, fail-open until activeChains populate, and allow subscriptions before discovery so snap balance events reach consumers. Subscribe to PermissionController:stateChanged. Remove debug console logging.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adjusts snap chain discovery and subscription behavior for non-EVM balances, which can change when/which balance updates propagate and how chains are assigned between Snap and RPC sources.
> 
> **Overview**
> Improves `SnapDataSource` resiliency so non-EVM snap balances are delivered even when snap/permission discovery is delayed or incomplete. Discovery now reads `chainIds` from `endowment:keyring` when missing on `endowment:assets`, retries discovery on early `AccountsController:accountBalancesUpdated` events, and preserves last known-good discovery state on rediscovery failures.
> 
> Subscriptions now *fail-open* before discovery completes (registering requested chains so push events reach `onAssetsUpdate`), and the data source can learn non-EVM chains from balance events (while explicitly excluding `eip155:*`). `AssetsController` also includes `SnapDataSource` active chains when building chain→account subscriptions so snap-backed accounts on non-enabled chains still receive balance subscriptions. Tests and changelog were updated accordingly, including requiring delegation of `PermissionController:stateChange`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d219a7228ab35e3026f160991f0a5e64931e5953. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- SnapDataSource now subscribes to `PermissionController:stateChange` instead of deprecated `PermissionController:stateChange`. Hosts that restrict or delegate events into the `AssetsController` messenger must delegate `PermissionController:stateChange`; delegating only `stateChange` will prevent permission-driven snap rediscovery. ([#8857](https://github.com/MetaMask/core/pull/8857))
 - Bump `@metamask/transaction-controller` from `^65.3.0` to `^66.0.0` ([#8796](https://github.com/MetaMask/core/pull/8796), [#8848](https://github.com/MetaMask/core/pull/8848))
 - Bump `@metamask/core-backend` from `^6.2.2` to `^6.3.0` ([#8813](https://github.com/MetaMask/core/pull/8813))
 - Bump `@metamask/phishing-controller` from `^17.1.2` to `^17.2.0` ([#8819](https://github.com/MetaMask/core/pull/8819))
@@ -17,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **SnapDataSource:** Read supported CAIP-2 chain IDs from `endowment:keyring` when `endowment:assets` has no `chainIds` caveat (common for some wallet snaps); retry discovery when a balance push arrives before `activeChains` is populated; accept balance payloads while discovery is still empty so updates are not dropped; register subscriptions in that window so `AccountsController:accountBalancesUpdated` can reach `onAssetsUpdate`. ([#8857](https://github.com/MetaMask/core/pull/8857))
 - Non-EVM assets with a `slip44` asset namespace (e.g. Bitcoin, Solana native, TRON) are now correctly typed as `native` instead of `erc20` in `assetsInfo` ([#8811](https://github.com/MetaMask/core/pull/8811))
 - Solana SPL tokens (CAIP-19 `solana:.../token:<address>`) are now correctly typed as `spl` instead of `erc20` in `assetsInfo` ([#8811](https://github.com/MetaMask/core/pull/8811))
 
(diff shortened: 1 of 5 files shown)


salimtb:
@metamaskbot publish-preview

packages/assets-controller/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **SnapDataSource:** Read supported CAIP-2 chain IDs from `endowment:keyring` when `endowment:assets` has no `chainIds` caveat (common for some wallet snaps); retry discovery when a balance push arrives before `activeChains` is populated; accept balance payloads while discovery is still empty so updates are not dropped; register subscriptions in that window so `AccountsController:accountBalancesUpdated` can reach `onAssetsUpdate`; subscribe to `PermissionController:stateChanged` instead of deprecated `stateChange`. ([#8857](https://github.com/MetaMask/core/pull/8857))
```
bergarces: I think this is a breaking change, as it involves a messenger with different permissions.

salimtb: ooh true , let me update now

bergarces: Ignore me. I can see that the permissions inside that type are the same.

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/AssetsController.ts
```diff
@@ -2565,9 +2561,13 @@ export class AssetsController extends BaseController<
...
     chainIds: ChainId[],
   ): void {
+    // Snap data source handles non-EVM chains that are never in #enabledChains.
+    // Include its active chains so snap-backed accounts (bip122, solana, tron…)
+    // appear in chainToAccounts and receive a subscription.
+    const snapActiveChains = this.#snapDataSource.getActiveChainsSync();
     const chainToAccounts = this.#buildChainToAccountsMap(
       accounts,
-      new Set(chainIds),
+      new Set([...chainIds, ...snapActiveChains]),
```
Prithpal-Sooriya, line 2570: is this change needed?

salimtb: no i don't think so , i'm cleaning now

packages/assets-controller/src/data-sources/SnapDataSource.ts
Prithpal-Sooriya, line 1: change is a bit larger than I initially thought. I thought the hotfix would just be.

```ts

// Rediscover keyring snaps when any snap gets installed
this.messenger.subscribe('SnapController:snapInstalled', () => {
  this.#discoverKeyringSnaps();
})

```

salimtb: this will only fix the timing race condition , when i added logs i found more than just that , happy to keep it small but keep in mind that 

- tron had no chainIds caveat on either permission
- activeSubscriptions=0 even for chains that were correctly discovered

the fact that chainId is not correctly received can cause other issues

Prithpal-Sooriya: Confirmed, this is also a valid fix - unless we need these additional checks?

Prithpal-Sooriya: Hmm okay... just cautious on complexity increase. If this is valid, happy to keep... but wondering if there is smaller changes needed or required.

Just feels like death by 1000 cuts, we keep piling on more and more complexity for every edge case, and this feels impossible to reason the full system with.

PR #8858, opened 2026-05-20
feat: add transaction data property useMoneyAccount

jpuri:
## Explanation

Add transactionData property useMoneyAccount, it will enable to know if source / target of transaction is money account.

## References

Related to https://consensyssoftware.atlassian.net/browse/CONF-1409

## Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk additive change that threads a new optional config/state field through `setTransactionConfig` and public exports; main risk is downstream consumers needing to handle the new enum value when present.
> 
> **Overview**
> Adds a new optional `paymentOverride` field (typed as `PaymentOverride`) to `TransactionConfig` and per-transaction `TransactionData`, allowing callers to override the payment source via `TransactionPayController.setTransactionConfig`.
> 
> Introduces the `PaymentOverride` enum (`moneyAccount`, `perps`, `predict`), exports it from the package public API, and updates tests and changelog to cover the new state update behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 924e329e0cae0a4323c0d3225a4eadaf9fcf0754. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/types.ts
```diff
@@ -140,6 +140,9 @@ export type TransactionConfig = {
    * When `isPostQuote` is false, it provides the funds and pays for gas.
    */
   accountOverride?: Hex;
+
+  /** Whether to use the Money Account (Money Keyring) as the payment source. */
+  useMoneyAccount?: boolean;
```
matthewwalsh0: Can we match the other existing configs so `isMoneyAccountSource` as we'll also likely need `isMoneyAccountTarget` in future too?

And should we add the callback at the same time so `getMoneyAccountWithdrawData` to the constructor?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/types.ts
jpuri: May be I change is to `isUsingMoneyAccount` , We can find source / target depending on postQuote.

I plan to work on separate PR to implement rest of changes, with this PR my intent is to unblock UI work.

On client side we need to know `isUsingMoneyAccount` value to set pay token and changes on review page also.

jpuri: Another option could be we add a more generic field called `additionTransactionsType: 'MONEY_ACCOUNT'` this can be indicator to get additional transaction data.

jpuri: I created this other draft PR, but it needs more testing before it can be merged: https://github.com/MetaMask/core/pull/8870

matthewwalsh0: Great point! We'd need similar properties for Perps > Predict and vice versa.

So maybe an optional `paymentOverride` enum with `PaymentOverride.Perps`, `PaymentOverride.Predict` and `PaymentOverride.MoneyAccount`.

And then we could have a single generic callback such as `getPaymentOverrideData(paymentOverride, amount)`?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/types.ts
jpuri: That sounds good, let me update the PR.

matthewwalsh0 (approved):

PR #8894, opened 2026-05-26
feat(transaction-pay-controller): add server pay strategy

matthewwalsh0:
## Explanation

Adds a Server strategy to `TransactionPayController` that consumes the intents-API's provider-agnostic `/quote`, `/submit`, and `/status` endpoints.

The strategy supports both gasless flows (where the quote provider handles submission using a signed delegation, removing the need for the user to submit an on-chain transaction) and non-gasless flows (where the quote steps are submitted through `TransactionController` as normal transactions). Both paths converge on a `/status` polling loop to track completion and surface the target hash.

Server strategy flags live in `confirmations_pay_extended` rather than `confirmations_pay`, keeping the rollout gated independently from Relay and Across. The strategy is disabled by default.

**Not yet supported in this initial implementation:**
- Post-quote
- Perps withdrawals
- Predict withdrawals
- Account overrides
- Payment overrides

## References

* https://github.com/consensys-vertical-apps/va-mmcx-intents-api/pull/32

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches payment execution, delegation/EIP-7702 quoting, and on-chain submission paths; rollout is gated by remote flags but misconfiguration or API errors could affect user fund flows.
> 
> **Overview**
> Adds a **Server** pay strategy to `transaction-pay-controller` that talks to the MetaMask intents API (`/quote`, `/submit`, `/status`) instead of provider-specific Relay/Across clients directly.
> 
> **Routing & flags:** `TransactionPayStrategy.Server` is registered in strategy resolution; default strategy order now prefers Server first. Enablement, base URL, and polling settings come from `confirmations_pay_extended.payStrategies.server` (off by default), separate from Relay/Across flags.
> 
> **Quote path:** Builds intents quote requests (exact input vs expected output, ERC-20 recipient decoding, EIP-7702 delegation `calls`/`authorizationList`, gasless eligibility) and normalizes API results into `TransactionPayQuote`, including gas-station / source-network fee handling for non-gasless quotes. Hyperliquid perps deposits are detected via bridge calldata and rewritten to a HyperCore sentinel target.
> 
> **Submit path:** Gasless quotes submit a signed delegation via `/submit`; non-gasless quotes push steps through `TransactionController` (`addTransaction` / `addTransactionBatch`), then both paths poll `/status` until confirmed (with timeouts, source/target hash updates, and `isIntentComplete`). `getGasFee` is exported for reuse in server quote gas math.
> 
> **Tests & changelog:** Broad unit coverage for API client, quotes, submit, perps normalization, strategy wiring, and feature-flag behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e34ddc92da84b00253e4c2814d38376b774fff1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/strategy/server/ServerStrategy.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Pay strategy that routes quote, submit, and status requests through the
+ * MetaMask intents API. Supports both gasless (delegated server-execute) and
+ * non-gasless (TransactionController-submitted) execution paths. Gated by the
+ * `payStrategies.server.enabled` remote feature flag.
+ */
+export class ServerStrategy implements PayStrategy<ServerQuote> {
+  supports(request: PayStrategyGetQuotesRequest): boolean {
+    const config = getPayStrategiesConfig(request.messenger);
+    return config.server.enabled;
+  }
```
OGPoyraz, line 22: Just want to confirm we won't enable this strategy until we fully support the listed use-cases in the PR description right?

As we only gate with FF here, when this flag is eventually turned on, and `Server` becomes first in `DEFAULT_STRATEGY_ORDER` and will be attempted for every transaction - including the not implemented flows listed in the PR description.

The fallback loop only skips to the next strategy if `getQuotes` throws or returns empty. If the intents API happens to return a valid-looking quote for one of these unsupported flows, it'll be accepted and executed without the special handling that Relay/Across provide for those cases.

OGPoyraz (approved):

packages/transaction-pay-controller/src/strategy/server/ServerStrategy.ts
matthewwalsh0: Absolutely, this is just the initial logic.

The endpoints don't even exist in the production API yet so it would throw and try the next strategy in the priority order.

PR #8908, opened 2026-05-27
fix: Normalize remote feature flag value wrappers

pedronfigueiredo:
There is an issue with the current remote feature flag implementation.

As you can see when the feature flags are configured for threshold based config, we define an array of objects that contains name, scope, and value. The problem with this is that there's lots of feature flags that are not defined in this format, and the selectors are unfortunately coupled to this format.

`getIsNotificationEnabledByDefaultFeatureFlag` inside `ui/selectors/metamask-notifications/metamask-notifications.ts` expects a `.value` property for example. But `confirmations_pay` feature flag, we don't expect such .value property to house the configuration JSON, and instead it's available directly at the root level.

## Summary

- Normalize processed remote feature flag wrapper objects that include a top-level `value` field.
- Copy object-valued `value` contents onto the root flag object while preserving the original `.value` field.
- Preserve direct feature flag config objects that do not use wrapper metadata.

## Root Cause

Some consumers read feature flag config directly from the flag object, while threshold or version wrapper shapes expose config under `.value`. This made selectors depend on the rollout format instead of the feature flag config itself.

## Impact

Feature flags can move between direct, threshold, and versioned configurations without requiring selectors to switch between direct config access and `.value` access. Existing consumers that still read `.value` remain compatible.

---

> [!NOTE]
> **Medium Risk**
> Changes processed flag shapes for v2 threshold configs, which can affect UI selectors that expect `.value`; legacy and unversioned thresholds stay wrapped for compatibility.
> 
> **Overview**
> **Remote feature flag threshold processing** now supports a **v2 direct-value shape** via `ThresholdVersion.DirectValue` (`thresholdVersion: 2`): after threshold selection, the controller exposes the chosen config **as the flag value itself** instead of `{ name, value }`.
> 
> **Legacy behavior is unchanged** for entries without that version (and for unknown `thresholdVersion` values), which still emit the `{ name, value }` wrapper. Types add optional `thresholdName` / `thresholdVersion` on threshold scope entries, and **`ThresholdVersion` is exported** from the package. Tests cover direct configs, legacy wrappers, v2 output, and fallback for unrecognized versions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1de751720a2189cc8ba99257df08e60546e4da5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +124,37 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+    return featureFlagValue;
+  }
+
+  return {
+    ...featureFlagValue.value,
+    ...featureFlagValue,
+  };
+}
+
+function normalizeFeatureFlagValue(featureFlagValue: Json): Json {
```
matthewwalsh0: For clarity, should we call this `normalizeThresholdValue`?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -379,7 +416,8 @@ export class RemoteFeatureFlagController extends BaseController<
         }
       }
 
-      processedFlags[remoteFeatureFlagName] = processedValue;
+      processedFlags[remoteFeatureFlagName] =
+        normalizeFeatureFlagValue(processedValue);
```
matthewwalsh0: As we also normalize for versioning, would it be simpler to normalize on line 412 above, so a function specific for getting threshold data?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +124,37 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+
+  return {
+    ...featureFlagValue.value,
+    ...featureFlagValue,
+  };
+}
+
+function normalizeFeatureFlagValue(featureFlagValue: Json): Json {
+  return isFeatureFlagValueWrapper(featureFlagValue)
+    ? spreadFeatureFlagValueWrapper(featureFlagValue)
```
matthewwalsh0: Am I misreading or is this saying if the flag is an object with a `value` property, then return the `value` property directly also?

But won't that mean we duplicate the data in every threshold flag?

Could we check (just for the threshold flow) if a `thresholdVersion` property exists and if set to `2`, then we just return `value` directly? And maybe with a `thresholdName` property instead of `name` for clarity?

matthewwalsh0 (changes requested):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
pedronfigueiredo: Yes we do duplicate because although we know in this function what is the shape of the flag defined on launch darkly, we have no way of knowing at this point what is the shape of the data that the selector needs or expects. In other words, we need this change to be backwards compatible.

We have to support three cases:

1. If a threshold hasn't been defined, then just add the whole flag

2. If a threshold is defined, and the selectors are currently defined without `.value`, we want to spread the .value at the root level so no selector change is required (to support our `confirmations_pay` with thresholds use case)

3. If a threshold is defined, and the selectors already have `.value`, we also need to keep the .value prop that the selector expects.

To support 2 and 3 simultaneously we need to duplicate the data.

I am not sure what you mean by the thresholdVersion prop with value 2, I can't see it in the client config api response. Can you clarify?

matthewwalsh0: I was suggesting we make the distinction explicit with a new property such as `thresholdVersion: 2` that we put in one of the array entries, so the code knows to only spread the value, rather than including it as it does currently.

Both should solve the problem, but didn't know if we wanted to future-proof and make the `RemoteFeatureFlagController` state more readable?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -306,7 +352,15 @@ export class RemoteFeatureFlagController extends BaseController<
...
     }
 
-    return getVersionData(flagValue, this.#clientVersion);
+    const versionData = getVersionData(flagValue, this.#clientVersion);
+
+    if (versionData === null) {
+      return null;
+    }
+
+    return isFeatureFlagValueWrapper(versionData)
```
matthewwalsh0: Doesn't the chosen version data get assigned to `processedValue` and then hit the standard threshold flow so we don't need to touch this?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+  }
+
+  return {
+    ...featureFlagValue,
+    ...featureFlagValue.value,
+  };
+}
+
+function normalizeThresholdValue(featureFlag: FeatureFlagScopeValue): Json {
+  if (featureFlag.thresholdVersion === THRESHOLD_VALUE_VERSION) {
```
matthewwalsh0: Minor, if we add future versions, this constant might get confusing. Maybe an enum?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+    return featureFlag.value;
```
matthewwalsh0, line 124: Minor, could still be useful to also include `thresholdName` property only if an object, for debug in the client?

Not sure how this will impact the new AB testing hooks.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+  }
+
+  const name = featureFlag.thresholdName ?? featureFlag.name;
+
+  return {
+    ...(isJsonObject(featureFlag.value) ? featureFlag.value : {}),
```
matthewwalsh0: If this is the fallback for the existing support, can we just return the same?

```
{
    name: featureFlag.name,
    value: featureFlag.value
}
```

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+
+function isFeatureFlagValueWrapper(
+  value: Json,
+): value is FeatureFlagValueWrapper {
+  return (
+    isJsonObject(value) && Object.prototype.hasOwnProperty.call(value, 'value')
+  );
+}
+
+function spreadFeatureFlagValueWrapper(
```
matthewwalsh0: Not needed?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
   };
 }
 
+function isJsonObject(value: Json): value is JsonObject {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function isFeatureFlagValueWrapper(
```
matthewwalsh0: Not needed?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +125,45 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
   };
 }
 
+function isJsonObject(value: Json): value is JsonObject {
```
matthewwalsh0: Not needed?

matthewwalsh0 (dismissed):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
pedronfigueiredo: good points let's add that if and when it becomes necessary then

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
pedronfigueiredo: good point, I forgot to revert this part

DDDDDanica (approved):

DDDDDanica:
LGTM !

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
```diff
@@ -41,6 +41,8 @@ export type FeatureFlagScope = {
 
 export type FeatureFlagScopeValue = {
   name: string;
+  thresholdName?: string;
+  thresholdVersion?: number;
```
NicolasMassart: **suggestion (blocking):** `thresholdVersion` should use the `ThresholdVersion` enum type, not `number`

The public type accepts any number, while the implementation only handles `ThresholdVersion.DirectValue = 2`. This means callers constructing a `FeatureFlagScopeValue` get no type guidance and must pass the magic number `2`.

Suggest exporting the enum from the types file (or a shared enums file) and using it here:

```suggestion
  thresholdVersion?: ThresholdVersion;
```

(with `ThresholdVersion` exported and imported accordingly)

NicolasMassart (commented):
Nice work overall — the `normalizeThresholdValue` approach is clean and the legacy wrapper fallback is well-preserved. One MEDIUM item to address before merge (type safety), two low-priority notes.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -118,6 +122,17 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
...
+function normalizeThresholdValue(featureFlag: FeatureFlagScopeValue): Json {
+  if (featureFlag.thresholdVersion === ThresholdVersion.DirectValue) {
+    return featureFlag.value;
+  }
+
+  return {
+    name: featureFlag.name,
+    value: featureFlag.value,
+  };
+}
```
NicolasMassart, line 133: **question (non-blocking):** What should happen when `thresholdVersion` is set but unrecognised?

If a future API response sends e.g. `thresholdVersion: 3`, it silently falls through to the legacy `{ name, value }` wrapper. That may be intentional, but it's not documented and there's no test for it. Worth a comment clarifying that unknown versions fall back to the legacy shape, and optionally a test case.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller-types.ts
```diff
@@ -41,6 +41,8 @@ export type FeatureFlagScope = {
 
 export type FeatureFlagScopeValue = {
   name: string;
+  thresholdName?: string;
```
NicolasMassart, line 52: **note (non-blocking):** `thresholdName` is present in the type but `normalizeThresholdValue` never reads it — the v2 path returns `featureFlag.value` directly and the legacy path uses `featureFlag.name`. The deferral is fine, but a short JSDoc comment on this field (e.g. `/** Replaces 'name' for v2 threshold entries; reserved for future debug use */`) would make the intent visible to future maintainers.

NicolasMassart (approved):
Looks good to me.

PR #8914, opened 2026-05-27
change isEnabled to a function

bergarces:
## Explanation

- `AssetsController` now re-evaluates the `isEnabled` callback when handling data-source active chain updates, instead of snapshotting its return value at construction
- `AssetsController` no longer skips constructor initialization when `isEnabled` returns false at construction time, so the controller can handle active chain updates after `isEnabled` later becomes true

Tested changes here: https://github.com/MetaMask/metamask-extension/pull/42990

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Always initializing when disabled at startup may increase background work until isEnabled turns false; behavior change for hosts that relied on the old early-exit constructor path.
> 
> **Overview**
> **`AssetsController`** now keeps the `isEnabled` option as a **callback** instead of freezing its value at construction, so feature gating can flip at runtime.
> 
> Constructor setup **always** runs state init, event subscriptions, and messenger action handlers even when `isEnabled()` is false at first—matching the case where the flag turns true later. **`#handleActiveChainsUpdate`** calls `isEnabled()` on each data-source chain change and skips `getAssets` when disabled.
> 
> Tests drop the “disabled at construction skips init” case, add coverage for re-evaluation on active-chain updates, and register **`NetworkEnablementController:getState`** where constructors are exercised without the full test harness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d1b0e110949a08ca64df06b29680956325250db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bergarces:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #8925, opened 2026-05-28
feat(assets-controller): remove individual RPC calls for token detection fallback

Prithpal-Sooriya:
## Explanation

Single RPC calls for detection is expensive! Racks up to 1000+ RPC calls.
So instead default mode will not use single RPC calls.

Only criteria that does require single RPC calls as fallback is for balance fetcher - since we want to fetch users balances.
- TokenDetector - does not use single calls
- BalanceFetcher - will use single calls as fallback

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes default RPC behavior for token detection and any direct MulticallClient callers; balance paths are preserved via explicit opt-in, but failed multicall may show missing balances where fallback was implicit before.
> 
> **Overview**
> **`MulticallClient.batchBalanceOf`** no longer issues per-token RPC fallbacks by default. Aggregate3 batches are retried (via `createServicePolicy`, `MULTICALL_MAX_RETRIES = 2`); when retries are exhausted or Multicall3 is unsupported, callers get **`success: false`** entries with **no** `provider.call` / `getBalance` unless they pass **`{ fallbackToSingleCalls: true }`**.
> 
> **`BalanceFetcher`** is the only in-package caller that opts into that flag, so user balance fetching can still degrade to single calls. **`TokenDetector`** keeps the default and avoids the large RPC fan-out during detection.
> 
> Changelog documents the behavior change for consumers of `MulticallClient`. Tests were split for default vs opt-in fallback and **`RpcDataSource`** expectations were updated (e.g. empty balances when multicall fails without single-call fallback on native-skip paths).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 241bf6a8b2c6ecf1f25556a2f3357b666f8d60eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -552,59 +565,70 @@ export class MulticallClient {
...
-              });
-            } else {
-              workingResult.push({
-                tokenAddress,
-                accountAddress,
-                success: false,
-              });
+          await createServicePolicy({
+            maxRetries: MULTICALL_MAX_RETRIES,
+          }).execute(async () => {
```
Prithpal-Sooriya, line 571: Looks like a ton of changes, but only because of indentation.

Using a Policy for built in retries and exponential backoff.

packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -552,59 +565,70 @@ export class MulticallClient {
...
+            const fallbackResults = await Promise.all(
+              batch.map((req) => this.#fetchSingleBalance(provider, req)),
+            );
+
+            for (const result of fallbackResults) {
+              workingResult.push(result);
+            }
+          } else {
+            workingResult.push(...this.#createFailedResponses(batch));
           }
```
Prithpal-Sooriya, line 633: Only new behaviour, if multicall fails then we decide if we want to invoke single RPC fallback, or return the failed responses.

packages/assets-controller/src/data-sources/evm-rpc-services/services/BalanceFetcher.ts
```diff
@@ -291,6 +291,7 @@ export class BalanceFetcher extends StaticIntervalPollingControllerOnly<BalanceP
         const responses = await this.#multicallClient.batchBalanceOf(
           chainId,
           batch,
+          { fallbackToSingleCalls: true },
```
salimtb, line 294: this is good , we should keep the fallback for balance fetcher and only stop it for tokenDetector

salimtb (dismissed):

bergarces (dismissed):

juanmigdr (dismissed):

packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -552,59 +565,70 @@ export class MulticallClient {
...
+              } else {
+                workingResult.push({
+                  tokenAddress,
+                  accountAddress,
+                  success: false,
+                });
+              }
             }
-          }
+          });
```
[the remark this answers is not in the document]
Prithpal-Sooriya: I'm pretty sure this is not a real concern. We had this same behaviour even without the retry.

bergarces (approved):

PR #8929, opened 2026-05-29
perf(account-tree-controller): scoped per-wallet sync for `AccountTreeController`

hmalik88:
## Explanation

After auditing performance of the account creation user flows, I observed that calling `AccountTreeController.syncWithUserStorage()` after a single wallet's state changes (e.g., immediately after an SRP import) iterates every local entropy wallet, fetching wallet metadata + the full groups feature list for each one. When only the newly imported wallet has actually changed, that fanout is wasted bandwidth and slows the perceived "import complete" moment, especially for users with multiple existing wallets.

This PR adds a scoped variant that syncs a single wallet by entropy source ID, leaving the existing full-sync API and behavior untouched. Consumers can swap `syncWithUserStorage()` for `syncWalletWithUserStorage(entropySourceId)` at SRP-import call sites to avoid the fanout. The new method is awaitable so sync-then-discover sequencing is preserved at the call site, defers to any in-flight full sync so callers cooperatively wait instead of racing, and does not flip `hasAccountTreeSyncingSyncedAtLeastOnce` (a scoped sync doesn't satisfy the first-full-sync contract).

## References

N/A

## Checklist


hmalik88:
@metamaskbot publish-previews

hmalik88:
@metamaskbot publish-previews

PR #8938, opened 2026-05-29
Remove unnecessary istanbul-ignore in tryToExtractFromCapabilityTypeConstructor

mcmire:
## Explanation

In `tryToExtractFromCapabilityTypeConstructor`, we look up the name of the action/event type that is being declared, and we use a type guard to ensure that it is an `Identifier` before proceeding. But we already know that, because we determined it in `recursivelyFindMessengerCapabilityTypeDeclarations`. Currently we are using `istanbul-ignore` so we don't have to test the redundant check. But a better fix would be to modify `recursivelyFindMessengerCapabilityTypeDeclarations` so that it returns the `typeName` that it's encountered. This way we can use it directly without having to look it up, and we can remove the redundant check along with the `istanbul-ignore`.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Docs extraction-only refactor with added test coverage; behavior for qualified constructor aliases remains skip-without-crash.
> 
> **Overview**
> **platform-api-docs** messenger extraction now records the constructor utility name (`ControllerGetStateAction` / `ControllerStateChangeEvent`) when walking type aliases, but only when the alias body is a **plain identifier** reference—not a namespace-qualified name like `Helpers.ControllerGetStateAction<...>`.
> 
> That lets `tryToExtractFromCapabilityTypeConstructor` use the pre-validated `typeName` instead of re-reading `body.getTypeName()` and dropping an `istanbul-ignore` for unreachable qualified-name handling. Qualified constructor aliases are skipped during the walk (same behavior as before, without redundant checks).
> 
> A regression test covers a messenger union with both a local inline action and a qualified constructor alias: only the local action is documented.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd8e9d7a5c4fa5082f984dfdd30f677299ada798. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/platform-api-docs/src/extraction.ts
+++ b/packages/platform-api-docs/src/extraction.ts
@@ -1,5 +1,6 @@
 import * as path from 'node:path';
 import type {
+  Identifier,
   InterfaceDeclaration,
   JSDocableNode,
   JSDocTag,
@@ -200,7 +201,8 @@ function findClassMethodDeclaration(
   }
   const methodName = indexLiteral.getLiteralValue();
 
-  // Reject qualified-name type references, as we can't follow those.
+  // Reject qualified-name type names, as we need a plain identifier to
+  // resolve the symbol.
   // EXAMPLE:
   //   import * as somePackage from '...';
   //   somePackage.FooController['someMethod']
@@ -285,6 +287,7 @@ type ConstructorMessengerCapabilityTypeDeclaration = {
   kind: 'action' | 'event';
   declaration: TypeAliasDeclaration;
   body: TypeReferenceNode;
+  typeName: Identifier;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/platform-api-docs/src/extraction.ts
```diff
@@ -200,7 +201,8 @@ function findClassMethodDeclaration(
   }
   const methodName = indexLiteral.getLiteralValue();
 
-  // Reject qualified-name type references, as we can't follow those.
+  // Reject qualified-name type names, as we need a plain identifier to
```
mcmire: This comment was inaccurate, so I fixed it as well.

PR #8012, opened 2026-02-22
feat: add @metamask/platform-api-docs package

cryptodev-2s:
## Explanation

Adds `@metamask/platform-api-docs`, a publishable package that generates and serves a searchable site documenting the Platform API — the catalog of actions and events available to clients through the message bus.

The package scans TypeScript source and `.d.cts` declaration files, finds every `*Messenger` type declaration, walks its `Actions` and `Events` type arguments to discover the capability types they reference, extracts JSDoc/handler/payload information for each, and renders the result as a Docusaurus site with per-namespace pages and local search.

### Features

- **Anchored on Messenger declarations.** Extraction discovers `*Messenger` type aliases and walks `Messenger<Namespace, Actions, Events>` to find the capability types — eliminating false positives from unrelated types that happen to share an action/event-like shape.
- **Two capability-type patterns supported.** Inline (`type FooAction = { type: '...'; handler: ... }`) and capability-type constructors (`type FooGetStateAction = ControllerGetStateAction<typeof name, State>`), including their `interface` variants.
- **Union expansion.** When a Messenger references an umbrella union like `FooActions = FooGetAction | FooSetAction`, the walker descends through it without documenting the intermediate alias.
- **Scans three layouts.** Configured `--scan-dir`s, `packages/*/src/` (`.ts`), and `node_modules/@metamask/*/dist/` (`.d.cts`).
- **Source links resolve automatically.** Reads `git remote get-url origin` and `git symbolic-ref refs/remotes/origin/HEAD` to build `https://github.com/<owner>/<repo>/blob/<branch>/<path>#L<line>` URLs; falls back to `main` for shallow clones.
- **Project label and commit SHA stamped on the site.** `--project-label Core` / `--project-label Extension` produces titles like `Platform API (Core)`, and the short Git commit is shown in the intro and navbar so engineers can tell how current the docs are.
- **Dedup across packages.** When a capability is declared in two places (e.g. once as the home definition and once as a re-export), the version with JSDoc and from the matching namespace's package wins.
- **Docusaurus site** with dark/light mode, offline search (no Algolia), and MDX-safe rendering of JSDoc.

### Usage

From the core monorepo:
```bash
yarn docs:platform-api:build   # Generate docs and build static site
yarn docs:platform-api:dev     # Dev server with hot reload
yarn docs:platform-api:serve   # Build and serve
```

From client projects (Extension, Mobile), install `@metamask/platform-api-docs` as a devDependency and add a script:
```json
{
  "scripts": {
[...]

PR #8940, opened 2026-05-29
chore: bump lint plugins and resolve floating-version fallout

cryptodev-2s:
## Explanation

PR #8012 ([feat/messenger-docs-site](https://github.com/MetaMask/core/pull/8012)) originally carried scoped resolutions for `eslint-plugin-import-x`, `eslint-plugin-n`, `eslint-import-resolver-typescript`, and `prettier`. Investigating those (per [r3326274335](https://github.com/MetaMask/core/pull/8012#discussion_r3326274335)) showed they weren't resolving a transitive-dep conflict — `yarn why` confirms each is consumed only by the root workspace. They existed purely to prevent the lockfile from drifting forward to plugin versions that the codebase hadn't been cleaned against. #8012 has since dropped them.

This PR is the follow-up: bump those plugins to the latest versions matching their (newly-widened) semver ranges and fix the two pre-existing lint issues the newer plugins surface, so the lockfile can stay current without the workaround resolutions ever coming back:

- `eslint-plugin-import-x`: `4.6.1` → `^4.16.2`
- `eslint-plugin-n`: `17.15.1` → `^18.0.1`
- `eslint-import-resolver-typescript`: `3.10.1` → `^4.4.4`
- `prettier`: `3.4.2` → `^3.8.3`

### Lint fallout fixed

- **`eslint.config.mjs`**: `import.meta.dirname` is only stable in Node 22.16+/24+, but the repo's `engines` field is `^18.18 || >=20`. `eslint-plugin-n@18`'s `no-unsupported-features` rule catches this. Resolve the config dir through `dirname(fileURLToPath(import.meta.url))` so the config works across the supported Node range.
- **`packages/chain-agnostic-permission/src/scope/constants.ts`**: `eslint-plugin-import-x@4.16` enables `no-named-as-default` by default. `@metamask/api-specs` exports `MetaMaskOpenRPCDocument` as both a named export and the default; switch to the named-import form so the reference is unambiguous.
- **`packages/messenger-cli/package.json`**: `yarn constraints --fix` rolled its `prettier` range forward to match the new root range.

### Shipped separately

Two `import-x/no-extraneous-dependencies` violations that surface under the bumped plugins were extracted into their own PRs and have already merged:

- #8971 declared `@metamask/messenger` on `react-data-query` (devDep)
- #8972 declared `@metamask/keyring-controller` on `transaction-pay-controller` (runtime dep)

## References

- Companion #8012 — already dropped its scoped resolutions in `bb6d436d5`; this PR keeps the lockfile from drifting back to versions that need them
- Companion MetaMask/eslint-config#444 — widens shared-config peer for `eslint-import-resolver-typescript` to `^4.0.0` so consumers using `@metamask/eslint-config-typescript` don't see a peer-dep mismatch

## Checklist

- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md) — N/A, no consumer-facing package surface changed
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them — N/A


matthewwalsh0 (dismissed):

0xEdouardEth (dismissed):
Approve changes on `chain-agnostic-permission`

--

[api-specs/src/index.ts](https://github.com/MetaMask/api-specs/blob/main/src/index.ts)

```
import MetaMaskOpenRPCDocument from "./build/openrpc.json";
import MultiChainOpenRPCDocument from "./build/multichain-openrpc.json";
export { MetaMaskOpenRPCDocument, MultiChainOpenRPCDocument };
export default MetaMaskOpenRPCDocument;
```

adonesky1 (dismissed):
LGTM

package.json
```diff
@@ -69,11 +69,11 @@
     "depcheck": "^1.4.7",
     "eslint": "^9.39.1",
     "eslint-config-prettier": "^9.1.0",
-    "eslint-import-resolver-typescript": "^3.6.3",
-    "eslint-plugin-import-x": "^4.3.0",
+    "eslint-import-resolver-typescript": "^4.4.4",
```
mcmire, line 72: `@metamask/eslint-config-typescript` has a peer dependency on `eslint-import-resolver-typescript` ^3.6.3.

package.json
```diff
@@ -69,11 +69,11 @@
...
+    "eslint-plugin-import-x": "^4.16.2",
     "eslint-plugin-jest": "^28.8.3",
     "eslint-plugin-jsdoc": "^50.2.4",
-    "eslint-plugin-n": "^17.10.3",
+    "eslint-plugin-n": "^18.0.1",
```
mcmire, line 76: `@metamask/eslint-config-nodejs` has a peer dependency on `@metamask/eslint-config-nodejs` ^17.10.3.

mcmire (commented):
I'm not totally sure I understand why this is needed. We seem to be bumping two packages by a major version, which is fine if there are new changes we want to take advantage of. But I am not sure how it would make the `lint` step less deterministic as we are using `^` to indicate ranges, which should tell Yarn to stay within a minor version. Plus we now get two extra peer dependency warnings we didn't get before (I've noted them below).

package.json
cryptodev-2s: Sorry I forget to move this to draft it will require the work here https://github.com/MetaMask/eslint-config/pull/444 before

package.json
cryptodev-2s: Sorry I forget to move this to draft it will require the work here https://github.com/MetaMask/eslint-config/pull/444 before

PR #8945, opened 2026-06-01
feat: disable token list

salimtb:
## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes token list fetch, cache, and StorageService persistence paths; mistakes could leave stale token data or block legitimate refreshes, but behavior is gated behind an opt-in callback with extensive tests.
> 
> **Overview**
> Adds an optional **`isDeprecated`** constructor callback so hosts can turn off token list fetching without tearing down the controller. When it returns true, **`TokenListController`** clears in-memory **`tokensChainsCache`**, skips token API calls on **`initialize`**, **`start`**, **`restart`**, **`_executePoll`**, and **`fetchTokenList`**, and wipes persisted **`tokensChainsCache:*`** keys to **`{ data: {}, timestamp: 0 }`**.
> 
> The flag is **re-checked on every entry point**, so it can be flipped at runtime (including while legacy polling is still scheduled). **`#enforceDisabledState`** also cancels pending debounced persists so stale writes cannot repopulate storage after a disable. Changelog and broad unit coverage document the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c00433f35e9dc2697ba9328b0b53c2ae8fa1546f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


salimtb:
@metamaskbot publish-preview

packages/assets-controllers/src/TokenListController.ts
```diff
@@ -167,6 +167,8 @@ export class TokenListController extends StaticIntervalPollingController<TokenLi
 
   #abortController: AbortController;
 
+  readonly #isDisabled: () => boolean;
```
salimtb: when this is true , tokenList will be an empty object

Prithpal-Sooriya: @salimtb since we are planning to use this pattern for all other old controllers, should we rename this to something else? (I think some old controllers already have an `isDisabled` property)

Prithpal-Sooriya: `isRemoved` / `isDeprecated`?

Prithpal-Sooriya (dismissed):

packages/assets-controllers/src/TokenListController.ts
sahar-fehri: I like this idea ☝️

salimtb:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #8948, opened 2026-06-01
fix(profile-sync-controller): scope storage-key & snap-signature caches by `entropySourceId`

mathieuartu:
## References

Related to: https://consensyssoftware.atlassian.net/browse/MUL-1861
Test drive extension PR: https://github.com/MetaMask/metamask-extension/pull/43110

## Checklist

---

> [!NOTE]
> **High Risk**
> Fixes incorrect cross-SRP cache sharing that could decrypt or write another SRP's backup/sync data when profile IDs align; touches encryption key derivation and snap signing paths.
> 
> **Overview**
> Fixes a **multi-SRP / shared-profile cache collision** where `UserStorageController` keyed storage-key and snap-signature caches only by `metamask:${profileId}`. When two SRPs shared the same `profileId` (e.g. after canonical profile pairing), one SRP could reuse the other’s cached derived key and read or write the wrong encrypted user storage (including symptoms like a second SRP showing the first’s account names).
> 
> **`UserStorageController`** now scopes both caches with `#scopedCacheKey` as `` `${entropySourceId}:${message}` `` (primary uses the first HD keyring metadata id, not a fixed `"primary"` string). Storage callbacks pass `entropySourceId` through; `getStorageKey` returns null while locked; snap signing uses the same scoped cache. **`user-storage` SDK** `StorageOptions` and `getStorageKey` propagate optional `entropySourceId` on get/set cache. Tests cover vault restore (new entropy id → new key), signature reuse after storage-key flush, and distinct API paths per entropy source with identical `profileId`. Changelog documents the fix; test fixtures include a default HD keyring for primary scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1edb9ab0dd14f7cf74cff2f06186b531261f818. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mathieuartu:
@metamaskbot publish-preview

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
```diff
@@ -516,7 +523,23 @@ export class UserStorageController extends BaseController<
...
+   *
+   * @param message - The tagged message used for signing.
+   * @param entropySourceId - The entropy source ID (omitted for the primary).
+   * @returns The scoped cache key.
+   */
+  #scopedCacheKey(
+    message: `metamask:${string}`,
+    entropySourceId?: string,
+  ): string {
+    return `${entropySourceId ?? 'primary'}:${message}`;
```
ccharly: Just wondering, if we somehow start using the primary keyring with its explicit `entropySourceId`, then it would be both under `primary:...` and `<primary-entropy-source-id>:...` scoped keys?

That's probably ok?

Otherwise, we could just fallback to the real primary entropy source ID too?

ccharly (changes requested):

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
ccharly: Another question that came to mind is, during a "reset wallet" flow, are we clearing this cache?

If during that flow we start using a new SRP, this would become the new primary, but I think the Snap message we put in cache here, would change with a different SRP, so we need to invalidate this cache somehow.

(I haven't checked thoroughly this controller, just pointing this out so we can double-check we're not invalid cached value 😅)

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
```diff
@@ -510,13 +527,64 @@ export class UserStorageController extends BaseController<
...
+   */
+  #getPrimaryEntropySourceId(): string {
+    const [primaryEntropySourceId] = this.#getHdKeyringEntropySourceIds();
+    if (!primaryEntropySourceId) {
+      throw new Error('#getPrimaryEntropySourceId - no HD keyring available');
+    }
+    return primaryEntropySourceId;
+  }
+
+  #_snapSignMessageCache: Record<string, string> = {};
```
gantunesr: Can this be moved to the top of the class alongside the rest of the private attrs?

Also, I know this was already here, but why does it starts with `#_`?

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
gantunesr: since this was mentioned, can we double check if `#_snapSignMessageCache` needs to be cleaned/flushed at some point?

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
mathieuartu: Fixed with [a1edb9a](https://github.com/MetaMask/core/pull/8948/commits/a1edb9ab0dd14f7cf74cff2f06186b531261f818)

packages/profile-sync-controller/src/controllers/user-storage/UserStorageController.ts
mathieuartu: Proposition here: [cbcef3d](https://github.com/MetaMask/core/pull/8948/commits/cbcef3d0a4b25691c31eac314a5deceadf0ffd57)

ccharly (approved):
Logic LGTM now, thanks for addressing that!

The only remark I might have is that we have no way to flush the `#snapSignMessageCache` cache. Though, given that we now use truly random ID (entropy source ID) even for the primary, we should never ever have have conflicting keys (+ this will get flushed upon client restart anyway).

PR #8926, opened 2026-05-28
feat: add `RampsController:ensureProviderForAsset` for MM Pay fiat surfaces

OGPoyraz:
## Explanation

When MM Pay surfaces (Perps deposit, Predict deposit) open the confirmation screen, the Ramp bootstrap may have selected a provider that does not support the fiat asset required by the transaction type - or may have failed to select any provider at all (e.g. first-time user in `us-ny` with no Transak Native). This leaves the user with an empty payment methods state and no recovery path.

### Solution

Instead of duplicating provider auto-selection in client-side React hooks (as proposed in metamask-mobile#30719), this PR moves the logic to the controller layer:

1. **New `RampsController.ensureProviderForAsset(assetId)`** - a generic, reusable action that checks if the currently selected provider supports the given CAIP asset. If not (or if no provider is selected), it finds the first available provider that does and auto-selects it.

2. **`TransactionPayController` integration** - when transaction data is first created (`isNewTransaction`), the controller derives the correct fiat asset via `deriveFiatAssetForFiatPayment` (feature-flag-driven, per-transaction-type) and calls `RampsController:ensureProviderForAsset` with the resulting CAIP ID.

3. **`utils/fiat.ts` extraction** - fiat-related utilities (`ensureProviderForFiatAsset`, `updateFiatAssetId`) are extracted from the controller into a dedicated utils file, following the same pattern as `utils/source-amounts.ts`.

### Why controller-side instead of client-side hooks?

- **Correct asset derivation**: The controller uses `deriveFiatAssetForFiatPayment` which resolves the asset from feature flags, not from `requiredTokens`. For `perpsDeposit`, the Ramp should buy ETH on Arbitrum (native token), not USDC - the Relay strategy handles the swap.
- **Single source of truth**: No need to duplicate provider selection logic across BuildQuote, CustomAmountInfo, and future MM Pay surfaces.

## References

- Related to https://github.com/MetaMask/metamask-mobile/pull/30719 (client-side approach for the same problem)

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
[...]

PR #8949, opened 2026-06-01
feat: add provider auto-selection to RampsController.getQuotes

saustrie-consensys:
Jira: https://consensyssoftware.atlassian.net/browse/TRAM-3629

## Summary

Adds an opt-in, **request-scoped provider auto-selection** capability to `RampsController.getQuotes` (and the `RampsController:getQuotes` messenger action). This is a HeadlessBuy platform improvement: a consumer can ask Ramps to pick a provider for a quote without managing provider selection itself, and **without mutating controller state** (`providers.selected` / `providerAutoSelected` are never touched).

### New `getQuotes` options
- `autoSelectProvider?: boolean` — when `true` and `providers` is omitted, Ramps resolves a provider that supports the requested `assetId`, for that request only.
- `preferredProviderIds?: string[]` — optional caller-supplied priority list.
- `restrictToKnownOrNativeProviders?: boolean` — headless-buy gating (see below). Defaults to off; the caller passes the value (e.g. from a remote feature flag).

### Provider type
`Provider` gains an optional `type: 'native' | 'aggregator'` field, mirroring the v2 providers API. Native detection now uses `type === 'native'` instead of matching on the provider id/name. The Transak aggregator is treated like any other provider (no special priority).

### Resolution precedence (request-scoped, no state mutation)
Resolves against the provider list **for the requested region** (fetching when the request targets a non-current region rather than trusting cached state), among those that support the asset:
1. the currently selected provider, when it's in the supporting set;
2. a previously-used provider — from `preferredProviderIds` when supplied, otherwise **derived from the user's completed-order history** (most recent first), read from Ramps' own normalized `state.orders`. This takes priority over native to preserve an existing KYC relationship;
3. a native provider (e.g. Transak Native);
4. the first supporting provider.

When unrestricted and no provider supports the asset, falls back to all known provider IDs (behaves as if no auto-selection occurred).

### `restrictToKnownOrNativeProviders` (gate-both)
For headless-buy v0, where we want to restrict to native (and providers the user already has a relationship with) until more providers are rolled out. When `true`:
- **auto-select path:** only a known (selected / completed-order) or native provider is eligible — no other provider is introduced;
- **explicit `providers` path:** the list is filtered to those supporting the region/asset;
- if nothing qualifies, `getQuotes` returns an **empty response and skips the service call** (passing `[]` would omit the provider filter and quote everyone), so the consumer can surface an "unavailable" state;
- the flag triggers resolution even without `autoSelectProvider`, so it can't be silently ignored.

## Scope / ownership
- **ramps-controller only** (@MetaMask/money-movement). No `transaction-pay-controller` changes — consumer adoption (MM Pay) is intentionally left to @MetaMask/confirmations as a follow-up.
- **Supersedes the approach in #8926**, which added a stateful `RampsController:ensureProviderForAsset`. This PR provides the same capability without the transaction-init side effect or `providers.selected` mutation. #8926 should be closed/abandoned in favor of this.

## Test plan
- [x] `yarn workspace @metamask/ramps-controller run test` — full suite, 100% coverage
- [x] `messenger-action-types:check`, eslint, prettier, and changelog validation pass

---

> [!NOTE]
> **Medium Risk**
> Changes which providers receive quote requests and can block quotes entirely under gating flags, affecting money-movement UX; behavior is opt-in and does not mutate persisted provider selection.
> 
> **Overview**
> Adds **request-scoped provider selection** to `RampsController.getQuotes` (and `RampsController:getQuotes`) so headless buy flows can quote a single provider without updating `providers.selected` or `providerAutoSelected`.
> 
> When `providers` is omitted, **`autoSelectProvider`** narrows the quote to one provider that supports the asset for the requested region (fetching providers for non-current regions instead of using stale cache). Precedence: selected provider → `preferredProviderIds` or completed-order history → `Provider.type === 'native'` → first supporting provider; explicit `providers` still bypasses auto-select.
> 
> **`restrictToKnownOrNativeProviders`** gates headless-buy v0: only selected/history-known or native providers; explicit lists are filtered by region/asset support; if none qualify, returns an empty quotes payload and skips `RampsService:getQuotes` (avoids quoting all providers via an empty filter).
> 
> `Provider` gains optional **`type: 'native' | 'aggregator'`** from the v2 API. Messenger action types, changelog, and extensive unit tests cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c887a43c21c4f528ee9cae878a2595afd1b9d444. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ramps-controller/src/RampsController.ts
```diff
@@ -1836,6 +1857,129 @@ export class RampsController extends BaseController<
...
+    for (const preferredId of preferred) {
+      const match = supportingProviders.find(
+        (provider) => provider.id === preferredId,
+      );
+      if (match) {
+        return [match.id];
+      }
+    }
+
+    const transakProvider = supportingProviders.find(
```
meltingice1337: I'm guessing the point of this is to prioritize Transak native, now Transak has both native and aggregator flows and the api just send back the transak native one instead of the aggregator if it is available. If the point is to prioritize native ramps providers (which I know we are in the process of adding more), there is a config when getting providers that actually tells us if the provider is native or aggregator https://on-ramp-cache.uat-api.cx.metamask.io/v2/regions/dk/providers?sdk=2.1.6&controller=13.2.0&context=mobile-android. `type=aggregator|native`. Maybe it woudl be better to use that, or at least use the cannonical provider name i.e. `transak-native` or `transak-native-staging` instead of the fragile `includes('transak')`. 

LE: Also, I think there are regions that don't have Transak native support but have Transak aggregator support and then we might prioritize Transak aggregator for no reason.

meltingice1337 (commented):
_Redacted_

packages/ramps-controller/src/RampsController.ts
saustrie-consensys: Just pushed a new commit.

Switched native detection to `provider.type === 'native'`. 

Also dropped the aggregator-priority step (after speaking with Lorenzo over Slack Huddle); aggregator's now treated like any other provider. Native is the only special case.

Since we're adding more providers soon, that last tier is behind a restrictToKnownOrNativeProviders flag on getQuotes. When it's on we don't fall through to brand-new providers - if there's no prior relationship and no native, we return empty so the UI shows "not available" instead of quoting someone the user's never used.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -1836,6 +1887,197 @@ export class RampsController extends BaseController<
...
+    if (
+      this.#isRegionCurrent(normalizedRegion) &&
+      this.state.providers.data.length > 0
+    ) {
+      providers = this.state.providers.data;
+    } else {
+      ({ providers } = await this.getProviders(normalizedRegion));
+    }
+
+    const lowerId = assetId.toLowerCase();
```
amitabh94: why do we need to change it lower case? Are we not using checksummed caip ids?

saustrie-consensys: Thanks. Resolved.

packages/ramps-controller/src/RampsService.ts
```diff
@@ -109,6 +109,12 @@ export type ProviderLimits = {
 export type Provider = {
   id: string;
   name: string;
+  /**
+   * Provider classification from the v2 API: 'native' (first-party
+   * integration, e.g. Transak Native) or 'aggregator' (third-party redirect).
+   * May be absent on responses that predate the v2 type field.
+   */
+  type?: 'aggregator' | 'native';
```
amitabh94, line 117: Do we still need this?

saustrie-consensys: Native-provider detection uses the API’s `type` field instead of matching on provider id/name.

`getQuotes` uses `provider.type === 'native'` to prefer native providers and to avoid accidentally prioritizing Transak aggregator in regions where Transak native is unavailable.

amitabh94 (approved):

PR #8951, opened 2026-06-01
fix(keyring-controller): auto-remove keyrings emptied during `withKeyring` callbacks

hmalik88:
## Explanation

When a consumer drains the last account from a non-primary keyring inside a `withKeyring` or `withKeyringV2` callback, the now-empty keyring was left behind in `state.keyrings` and persisted to the vault. This diverges from `removeAccount`, which already removes and destroys a keyring once its final account is removed.

This PR adds a `#cleanUpEmptiedKeyringsAfter` helper that wraps the operation run by `withKeyring`/`withKeyringV2`. It snapshots which keyrings were non-empty before the callback and, afterwards, removes and destroys any non-primary keyring that transitioned from non-empty to empty during the callback, before persistence runs.

To match existing behavior and avoid surprises:

- **Pre-existing empty keyrings are preserved.** Keyrings that were already empty before the callback (e.g. created intentionally via `addNewKeyring` without subsequently adding accounts) are left untouched, mirroring `removeAccount`.
- **The primary keyring (index 0) is never auto-removed**, even if drained, preserving `removeAccount`'s primary-keyring invariant.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes vault/keyring lifecycle during `withKeyring` paths; behavior is scoped and well-tested but affects persisted wallet state.
> 
> **Overview**
> **`withKeyring` and `withKeyringV2` now align with `removeAccount` when a callback drains a keyring.** A new `#cleanUpEmptiedKeyringsAfter` wrapper snapshots which keyrings had accounts before the callback; after it runs, any **non-primary** keyring that went from non-empty to empty is removed from `#keyrings`, destroyed (including V2), and persisted—so empty keyrings no longer stick in `state.keyrings` or the vault.
> 
> **Intentional empty keyrings and the primary HD keyring are unchanged:** pre-existing empty keyrings (e.g. from `addNewKeyring` only) are not removed; the primary keyring is never auto-removed even if emptied. Failed callbacks still roll back, so cleanup does not run on error. Primary-keyring detection is centralized in `#isPrimaryKeyring` and reused by `#assertNotRemovingPrimaryKeyring`.
> 
> Changelog and broad tests cover persistence after lock/unlock, destroy behavior, and both v1 and v2 APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ba48ce772d6bce1cf0f8b04c6ab49269e43b9c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-controller/src/KeyringController.ts
+++ b/packages/keyring-controller/src/KeyringController.ts
@@ -2967,8 +2967,8 @@ export class KeyringController<
    * {@link KeyringController.addNewKeyring} without subsequent account
    * creation) are left alone, as are keyrings created within the operation
    * itself (they are not part of the pre-operation snapshot). The primary
-   * keyring (index 0) is also preserved unconditionally to keep
-   * `removeAccount`'s primary-keyring invariant intact.
+   * keyring (see {@link KeyringController.#isPrimaryKeyring}) is also preserved
+   * unconditionally to keep `removeAccount`'s primary-keyring invariant intact.
    *
    * @param operation - The operation to execute.
    * @returns The result of the operation.
@@ -3002,7 +3002,9 @@ export class KeyringController<
 
     const emptied = this.#keyrings.filter(
       (entry, index) =>
-        index !== 0 && wasNonEmpty.has(entry.keyring) && isNowEmpty[index],
+        !this.#isPrimaryKeyring(entry, this.#keyrings) &&
+        wasNonEmpty.has(entry.keyring) &&
+        isNowEmpty[index],
     );
 
     if (emptied.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2948,6 +2952,72 @@ export class KeyringController<
...
+    const isNowEmpty = await Promise.all(
+      this.#keyrings.map(
+        async ({ keyring }) => (await keyring.getAccounts()).length === 0,
+      ),
+    );
+
+    const emptied = this.#keyrings.filter(
+      (entry, index) =>
+        index !== 0 && wasNonEmpty.has(entry.keyring) && isNowEmpty[index],
+    );
```
gantunesr, line 3008: why do you also need to check if the keyring was not empty before? Should the presence in `isNowEmpty` be enough?

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -2948,6 +2952,72 @@ export class KeyringController<
...
+   * @template Result - The type of the value resolved by the operation.
+   */
+  async #cleanUpEmptiedKeyringsAfter<Result>(
+    operation: () => Promise<Result>,
+  ): Promise<Result> {
+    // Only the primary keyring exists, which is never auto-removed, so there
+    // is nothing to clean up regardless of what the operation does.
+    if (this.#keyrings.length <= 1) {
+      return operation();
+    }
```
gantunesr, line 2984: This makes me think if we should have a more obvious way to identify the "Primary Keyring". This seems like a weak check that could still work in case of a bug

packages/keyring-controller/src/KeyringController.ts
hmalik88: `isNowEmpty` alone would also catch keyrings that were already empty before the callback (e.g. intentionally created via `addNewKeyring`), which we want to preserve to match `removeAccount`. The `wasNonEmpty` snapshot scopes cleanup to keyrings the consumer actually drained during the operation.

packages/keyring-controller/src/KeyringController.ts
hmalik88: Good call, created a helper for this: [7ba48ce](https://github.com/MetaMask/core/pull/8951/commits/7ba48ce772d6bce1cf0f8b04c6ab49269e43b9c8)

gantunesr (approved):

PR #8953, opened 2026-06-01
feat(wallet): wire `ApprovalController` into default initialization

sirtimid:
## Explanation

Wires `ApprovalController` into `@metamask/wallet`'s default initialization, so consumers can use its messenger actions (`ApprovalController:addRequest`, `startFlow`, `acceptRequest`, …).

`@metamask/wallet` is the shared controller-integration layer for the extension, mobile, and wallet-cli, so the two construction values that differ between those clients are exposed as injectable options under `instanceOptions.approvalController` rather than hardcoded. How each environment supplies them:

| Option | Extension | Mobile | Default (e.g. wallet-cli) |
| --- | --- | --- | --- |
| `showApprovalRequest` | its confirmation-UI opener (`showUserConfirmation`) | no-op — drives approvals through state | no-op `() => undefined` |
| `typesExcludedFromRateLimiting` | the 6 default types + smart-tx status page + `snap_dialog` | `Transaction` + `WatchAsset` + smart-tx status + `snap_dialog` | 6 common EVM approval types — signing, transaction, watch-asset, encryption |

Note the two clients don't even share the same string for the smart-transaction status type, so there's no single correct list to hardcode — clients pass their own; consumers that omit it get the baseline above.

`ApprovalController` owns only its own actions/events, so its messenger is a plain namespaced child with no delegation. The wiring is intentionally thin: `ApprovalController` is a dependency-free primitive, so there's little to integrate today — its role grows as later controllers (transaction, signature, user-operation, …) delegate to `ApprovalController:addRequest`/`startFlow` through the wallet's root messenger, defined here once instead of in each client. Adds the `@metamask/approval-controller` and `@metamask/controller-utils` dependencies.

## References

How the clients construct `ApprovalController` today (the source for the table above):

- Extension: [`approval-controller-init.ts`](https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/messenger-client-init/confirmations/approval-controller-init.ts) · [`approval-controller-messenger.ts`](https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/messenger-client-init/messengers/approval-controller-messenger.ts)
- Mobile: [`approval-controller-init.ts`](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine/controllers/approval-controller/approval-controller-init.ts) · [`approval-controller-messenger.ts`](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine/messengers/approval-controller-messenger/approval-controller-messenger.ts)

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking change to shared wallet bootstrap affects all clients’ messenger/controller wiring; mis-migration causes duplicate ApprovalController registration and broken confirmations flows.
> 
> **Overview**
> **BREAKING:** The default `@metamask/wallet` initialization now builds an `ApprovalController` and registers `ApprovalController:*` on the root messenger. Extension/mobile clients that already construct their own must drop that wiring before upgrading or messenger registration will collide.
> 
> A new `approvalController` initialization module wires a namespaced child messenger, forwards persisted state, and exposes optional `instanceOptions.approvalController` hooks: `showApprovalRequest` (defaults to a no-op) and `typesExcludedFromRateLimiting` (defaults to six common EVM approval types). Tests cover defaults, overrides, rate-limit behavior, and root-messenger actions.
> 
> Package metadata adds `@metamask/approval-controller` and `@metamask/controller-utils`, updates the README dependency graph, assigns CODEOWNERS for the new init folder, and documents the change in the wallet changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce8d0619a3bf8babf25649f9f0f8e1f00384a7b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
FrederikBolding, line 1: Please add a CODEOWNER entry for this file and assign it to `@MetaMask/confirmations`

sirtimid: I added `@MetaMask/core-platform` as well following what you did for `keyring-controller`, was this correct?

.github/CODEOWNERS
```diff
@@ -129,6 +129,7 @@
 /packages/profile-metrics-controller            @MetaMask/mobile-platform @MetaMask/extension-platform
 
 ## Initialization
+/packages/wallet/src/initialization/instances/approval-controller.ts @MetaMask/confirmations @MetaMask/core-platform
```
FrederikBolding:
```suggestion
/packages/wallet/src/initialization/instances/approval-controller.ts @MetaMask/confirmations
```
Let's mirror the CODEOWNER rules for the controller itself

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
```diff
@@ -0,0 +1,55 @@
...
+ * signing/transaction types to be queued from the same origin.
+ *
+ * Clients can override this entirely via
+ * `instanceOptions.approvalController.typesExcludedFromRateLimiting`. The set
+ * differs per platform — the extension and mobile each append client-specific
+ * types (e.g. their smart-transaction status page and `snap_dialog`), and even
+ * use different string values for the same concept — so there is no single
+ * correct list to hardcode here.
+ */
+const DEFAULT_TYPES_EXCLUDED_FROM_RATE_LIMITING = [
```
FrederikBolding, line 20: How did you pick these?

.github/CODEOWNERS
sirtimid: Done https://github.com/MetaMask/core/pull/8953/commits/5e1577363bcc5bd9c5402c97c7a35bf95dc59b0e

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
sirtimid: Done https://github.com/MetaMask/core/pull/8953/commits/5e1577363bcc5bd9c5402c97c7a35bf95dc59b0e

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
sirtimid: These are the six EVM `ApprovalType` values carried over from the old-design wallet port (`origin/feat/wallet-library`), which mirrored the extension's exclusion list. They're also the superset of what the clients exclude today: both extension and mobile exempt `Transaction` and `WatchAsset`; the extension additionally exempts the signing (`personal_sign`, `eth_signTypedData`) and encryption (`eth_getEncryptionPublicKey`, `eth_decrypt`) approvals.

I used that as the platform-agnostic baseline and dropped the two client-specific string entries — the smart-transaction status type (whose value even differs between extension and mobile) and `snap_dialog` (which only matters once `SnapController` is wired) — making the whole list overridable via `instanceOptions.approvalController.typesExcludedFromRateLimiting` instead.

Happy to trim the default to just the intersection (`Transaction` + `WatchAsset`) if you'd prefer a more conservative baseline.

FrederikBolding: Let's have someone from the confirmations team sanity check that 👍

packages/wallet/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Wire `ApprovalController` into the default wallet initialization ([#8953](https://github.com/MetaMask/core/pull/8953))
```
FrederikBolding: Technically this will end up being a breaking change because it requires deleting the `ApprovalController` config on the clients

FrederikBolding (commented):
LGTM, let's get someone from @MetaMask/confirmations to review this as well.

.github/CODEOWNERS
```diff
@@ -129,6 +129,7 @@
 /packages/profile-metrics-controller            @MetaMask/mobile-platform @MetaMask/extension-platform
 
 ## Initialization
+/packages/wallet/src/initialization/instances/approval-controller.ts @MetaMask/confirmations
```
matthewwalsh0: Would it be easier to have a directory per controller so we can include other utils or constants etc if needed?

So `/packages/wallet/src/initialization/instances/approval-controller/` ?

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
matthewwalsh0: Looks fine to me, as long as it's the superset of the two, we were meaning to align mobile with extension.

packages/wallet/src/types.ts
```diff
@@ -21,6 +25,12 @@ export type WalletOptions = {
 };
 
 export type InstanceSpecificOptions = {
+  approvalController?: {
```
matthewwalsh0: Could we start the practice of creating a controller specific type for these?

So maybe an `ApprovalControllerInstanceOptions` type defined in `packages/wallet/src/initialization/instances/approval-controller/types.ts`?

matthewwalsh0 (changes requested):

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
FrederikBolding: I would add `snap_dialog` if we are going for a superset default.

FrederikBolding: Though I guess the problem with that would be the client specific smart TX string

packages/wallet/src/types.ts
sirtimid: Done https://github.com/MetaMask/core/pull/8953/commits/343634d9d90337cd70e7b021ae8cbb97d85cef3c

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
sirtimid: Thought about this more and kept the default as the EVM baseline (no `snap_dialog`). The reasoning: the default only applies to consumers that *don't* pass `typesExcludedFromRateLimiting`, and both extension and mobile pass their own full lists — so the default isn't what aligns the clients (that happens via each client's override at adoption time). That means a superset default doesn't really buy us anything, and it can't be a true superset anyway because of the smart-TX string divergence you flagged. So the default stays minimal; `snap_dialog` joins it when we wire `SnapController` (nothing emits it until then), and clients still get exact parity via the override. Happy to add it back if you'd still prefer the superset.

.github/CODEOWNERS
sirtimid: Done https://github.com/MetaMask/core/pull/8953/commits/343634d9d90337cd70e7b021ae8cbb97d85cef3c Since it now makes the layout inconsistent do we need to migrate on a follow-up keyring + storage to the new directory/type convention so the package isn't left in two styles? @FrederikBolding I can do it if you agree as well

packages/wallet/CHANGELOG.md
sirtimid: Fixed https://github.com/MetaMask/core/pull/8953/commits/343634d9d90337cd70e7b021ae8cbb97d85cef3c

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
FrederikBolding: Ideally we didn't have the divergence in the smart TX string, then we could use a superset and not have overrides.

But up to @matthewwalsh0 whether he thinks that is worth the trouble right now.

matthewwalsh0: We can double-check both clients and remove in future once we know the superset is safe?

matthewwalsh0 (dismissed):

.github/CODEOWNERS
FrederikBolding: Sounds good to me, we can move the encryptor out from `keyring-controller`  then!

packages/wallet/src/initialization/instances/approval-controller/approval-controller.ts
```diff
@@ -0,0 +1,53 @@
...
+> = {
+  name: 'ApprovalController',
+  init: ({ state, messenger, options }) =>
+    new ApprovalController({
+      state,
+      messenger,
+      // The consumer supplies the callback that surfaces a pending request to
+      // the user; default to a no-op so the controller works headlessly (this
+      // matches mobile, which drives approvals through state; the extension
+      // injects its own).
```
FrederikBolding:
```suggestion
```
Nit: omit since we have doc string now

FrederikBolding (approved):

PR #8959, opened 2026-06-01
ci: Get access token in each job individually in `update-changelogs` workflow

Mrtenz:
## Explanation

This attempts to fix the `update-changelogs` workflow.

## Checklist

---

> [!NOTE]
> **Low Risk**
> CI-only workflow wiring change; write paths still use the same get-token action and default-branch environment, with read steps on the default token.
> 
> **Overview**
> Refactors the **update-changelogs** workflow so elevated GitHub access is no longer minted once in a shared `get-token` job and passed downstream.
> 
> The **`is-release`** job now depends only on **`is-fork`**, runs when the PR is not from a fork, and uses **`GITHUB_TOKEN`** for read-only steps (`gh pr view`, checkout, merge-base, release detection). The standalone **`get-token`** job is removed.
> 
> **`react-to-comment`** and **`update-changelogs`** each declare **`id-token: write`**, call **`get-token`** in their own steps, and use that token for reactions, checkout with push rights, and PR comments—instead of **`needs.get-token.outputs.token`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28fedae1a65c0c4a02c4a76e05bf3fde49a853ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/update-changelogs.yml
+++ b/.github/workflows/update-changelogs.yml
@@ -32,29 +32,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
 
-  get-token:
-    name: Get access token
-    needs: is-fork
-    if: needs.is-fork.outputs.is-fork == 'false'
-    runs-on: ubuntu-latest
-    environment: default-branch
-    permissions:
-      id-token: write
-    outputs:
-      token: ${{ steps.get-token.outputs.token }}
-    steps:
-      - name: Get access token
-        id: get-token
-        uses: MetaMask/github-tools/.github/actions/get-token@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/update-changelogs.yml
```diff
@@ -67,7 +48,7 @@ jobs:
       - name: Get pull request info
         id: pr-info
         env:
-          GH_TOKEN: ${{ needs.get-token.outputs.token }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Mrtenz, line 51: I don't think we actually need a special token here.

mcmire (approved):
I suspect this whole workflow could probably be optimized better. Looks good regardless.

PR #8960, opened 2026-06-01
feat(multichain-account-service): add `deleteAccount` to providers

hmalik88:
## Explanation

The `removeMultichainAccountWallet` method was updated to properly delete all associated accounts with a wallet. The `deleteAccount` method was added to the providers to facilitate this.

## References

N/A

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking messenger API plus multi-provider account/keyring deletion; partial failures are swallowed after reporting, so consumers must not assume all keyring state was cleaned when errors are logged.
> 
> **Overview**
> Adds **`deleteAccount(id)`** across BIP-44 account providers so wallet teardown can remove accounts in each keyring (EVM via v2 `keyring.deleteAccount`, snap providers via `removeAccount`). **`AccountProviderWrapper`** gains **`unwrap()`** and always forwards **`deleteAccount`**, including when basic functionality has disabled the wrapper.
> 
> **Breaking:** **`removeMultichainAccountWallet`** (and its messenger action) now takes only **`entropySource`**—the **`accountAddress`** argument is gone. Removal walks every registered provider, deletes all accounts for that entropy source, then drops the wallet from the service map. Enumeration uses **`unwrap()`** when wrapped so snap-backed accounts are still found if the wrapper is disabled; per-account and per-provider failures are **best-effort** (cleanup continues) with one aggregated **`reportError`** / Sentry payload listing failures. Typed **`SentryError`** / failure context types support tests and observability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eee68806da85b4de8b07ffe023094bcdca844e1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +534,44 @@ export class MultichainAccountService {
...
     const wallet = this.#getWallet(entropySource);
 
-    await this.#messenger.call(
-      'KeyringController:removeAccount',
-      accountAddress,
-    );
+    for (const provider of this.#providers) {
+      const owned = provider
+        .getAccounts()
+        .filter((account) => account?.options?.entropy?.id === entropySource);
```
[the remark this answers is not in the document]
hmalik88: [3b9422d](https://github.com/MetaMask/core/pull/8960/commits/3b9422df6519a29695a30fdbafe0e5179e5470ea)

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +534,53 @@ export class MultichainAccountService {
...
+        } catch (error) {
+          reportError(
+            this.#messenger,
+            `Unable to delete account during wallet removal`,
+            error,
+            {
+              provider: provider.getName(),
+              accountId: account.id,
+            },
+          );
```
montelaidev: Can we use Promise.allSettled here? We could just aggregate all the failures into one report.

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +534,53 @@ export class MultichainAccountService {
...
+      // functionality is off). The wrapper's `deleteAccount` itself forwards
+      // unconditionally, but its `getAccounts()` returns `[]` when disabled,
+      // which would otherwise leave snap-backed accounts orphaned in their
+      // underlying keyrings.
+      const source = isAccountProviderWrapper(provider)
+        ? provider.wrappedProvider
+        : provider;
+      const owned = source
+        .getAccounts()
+        .filter((account) => account?.options?.entropy?.id === entropySource);
```
montelaidev: Why do we need this filter? All accounts should have an entropy id now.

hmalik88: The filter is needed because we don't want to delete accounts from the provider that aren't associated with that wallet

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Yeah good point, the sentry report should treat the failure as one operation, but we don't need the `Promise.allSettled` is there is no parallelism benefit, the operations are all on the keyring mutex and would get queued up anyway.

hmalik88: [065838b](https://github.com/MetaMask/core/pull/8960/commits/065838bb7430ec69ea71578665fb9ebee6fc97b1)

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -44,6 +44,32 @@ import type {
...
+/**
+ * Aggregated context payload attached to the Sentry report produced by
+ * {@link MultichainAccountService.removeMultichainAccountWallet} when one
+ * or more per-account deletions fail.
+ */
+export type RemoveMultichainAccountWalletFailureContext = {
+  entropySource: EntropySourceId;
+  failureCount: number;
+  failures: RemoveMultichainAccountWalletFailure[];
+};
```
ccharly, line 71: NOTE: Makes me wonder if we should re-apply this pattern for other errors too 🤔

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +560,79 @@ export class MultichainAccountService {
...
+      // functionality is off). The wrapper's `deleteAccount` itself forwards
+      // unconditionally, but its `getAccounts()` returns `[]` when disabled,
+      // which would otherwise leave snap-backed accounts orphaned in their
+      // underlying keyrings.
+      const source = isAccountProviderWrapper(provider)
+        ? provider.wrappedProvider
+        : provider;
+      const owned = source
+        .getAccounts()
+        .filter((account) => account?.options?.entropy?.id === entropySource);
```
ccharly: We shouldn't need any `?` here. We only deal with BIP-44 providers and BIP-44 accounts (that's also enforced type-wise), so `account.options.entropy.id` should always be defined

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +560,79 @@ export class MultichainAccountService {
...
-    );
+    for (const provider of this.#providers) {
+      // For wrapped providers, enumerate via the underlying provider so we
+      // also see accounts when the wrapper has been disabled (i.e. basic
+      // functionality is off). The wrapper's `deleteAccount` itself forwards
+      // unconditionally, but its `getAccounts()` returns `[]` when disabled,
+      // which would otherwise leave snap-backed accounts orphaned in their
+      // underlying keyrings.
+      const source = isAccountProviderWrapper(provider)
+        ? provider.wrappedProvider
```
ccharly: Cosmetic: Thoughts on naming this `unwrap()` instead? (we have this pattern elsewhere for some adapters/wrappers)

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +560,79 @@ export class MultichainAccountService {
...
+      // failure in `context`. The shape is pinned by
+      // `RemoveMultichainAccountWalletFailureContext`.
+      const context: RemoveMultichainAccountWalletFailureContext = {
+        entropySource,
+        failureCount: failures.length,
+        failures: failures.map(({ provider, accountId, error }) => ({
+          provider,
+          accountId,
+          error: toErrorMessage(error),
+          stack: error instanceof Error ? error.stack : undefined,
```
ccharly: I'm not sure we want to send that much information to Sentry though? 🤔

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +560,79 @@ export class MultichainAccountService {
...
   async removeMultichainAccountWallet(
     entropySource: EntropySourceId,
-    accountAddress: string,
   ): Promise<void> {
     const wallet = this.#getWallet(entropySource);
+    const failures: {
+      provider: string;
+      accountId: Bip44Account<KeyringAccount>['id'];
+      error: unknown;
+    }[] = [];
```
ccharly: We could re-use `RemoveMultichainAccountWalletFailure` here

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +560,79 @@ export class MultichainAccountService {
...
+    }
+
+    if (failures.length > 0) {
+      // One aggregated report per wallet-removal action: keeps the Sentry
+      // message stable for grouping while still surfacing every per-account
+      // failure in `context`. The shape is pinned by
+      // `RemoveMultichainAccountWalletFailureContext`.
+      const context: RemoveMultichainAccountWalletFailureContext = {
+        entropySource,
+        failureCount: failures.length,
```
ccharly: Would `failureCount` really be that important? Maybe having the number of success could be better? So we can compute the ratio of success vs failures (with the `failures` array, we can count the number of failures we have).

But that might be too much information 🤷‍♂️ no account removal should really fail anyway

hmalik88: Oo I like that idea, but yeah I don't ever expect this to fail haha.

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: woops, meant to remove that earlier, good catch

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: yeah, I agree we should probably do that for the others too (in a later pr 😄)

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: yeah, that's nicer. i'll make the change

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: [2b4271c](https://github.com/MetaMask/core/pull/8960/commits/2b4271c56052c056eacb8a9af6e0649246319169)

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: [3f52fad](https://github.com/MetaMask/core/pull/8960/commits/3f52fade8f81ae8a5f33d86e2d1314797cf3cf11)

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: [4634931](https://github.com/MetaMask/core/pull/8960/commits/4634931e447609b714d9aeffb45323ffbb21951c)

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Just decided to remove it.

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: Took out `failureCount` and `stack`

packages/multichain-account-service/src/MultichainAccountService.test.ts
```diff
@@ -946,31 +965,237 @@ describe('MultichainAccountService', () => {
...
+      expect(captureExceptionSpy).toHaveBeenCalledTimes(1);
+      const sentryError = captureExceptionSpy.mock
+        .calls[0]?.[0] as SentryError<RemoveMultichainAccountWalletFailureContext>;
+      expect(sentryError.message).toBe(
+        'Failed to delete one or more accounts during wallet removal',
+      );
+      expect(sentryError.context?.entropySource).toBe(
+        MOCK_HD_KEYRING_1.metadata.id,
+      );
+      expect(sentryError.context?.failureCount).toBe(1);
```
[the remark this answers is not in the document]
hmalik88: [5fc8a32](https://github.com/MetaMask/core/pull/8960/commits/5fc8a325e3d351e8c5792a5e6e19469dcebebac8)

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -534,26 +558,73 @@ export class MultichainAccountService {
...
+          await provider.deleteAccount(account.id);
+        } catch (error) {
+          failures.push({
+            provider: provider.getName(),
+            accountId: account.id,
+            error,
+          });
+        }
+      }
+    }
```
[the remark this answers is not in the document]
hmalik88: [fe10e4c](https://github.com/MetaMask/core/pull/8960/commits/fe10e4c2901e9752a540ca47b3fe87077fcb0246)

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -44,6 +44,32 @@ import type {
...
+  error: unknown;
+};
+
+/**
+ * Aggregated context payload attached to the Sentry report produced by
+ * {@link MultichainAccountService.removeMultichainAccountWallet} when one
+ * or more per-account deletions fail.
+ */
+export type RemoveMultichainAccountWalletFailureContext = {
+  entropySource: EntropySourceId;
```
ccharly: I wonder if we should include that 🤔 I tend to think we should not include any entropy related things in logs/reporting, WDYT?

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,8 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - `EvmAccountProvider.deleteAccount(id)` resolves the account's entropy source and forwards to the v2 HD keyring's `deleteAccount(accountId)`. When this drains the last account on a non-primary HD keyring, the keyring controller automatically prunes the empty keyring.
+  - `SnapAccountProvider.deleteAccount(id)` resolves the account's address and forwards to the legacy `SnapKeyring.removeAccount(address)` so that the snap and the host keyring stay in sync.
+  - `AccountProviderWrapper.deleteAccount(id)` forwards unconditionally so that wallet-removal flows can clean up snap-backed accounts even when the wrapper has been disabled.
+  - Added a public `unwrap` method on `AccountProviderWrapper`. This is an escape hatch for cleanup flows (e.g. wallet removal) that need to enumerate the underlying provider's accounts regardless of enabled state, since `getAccounts()` still returns `[]` when the wrapper is disabled.
```
ccharly: I find this slightly "too verbose" 😅 We could have this kind of details in the PR description itself (in case we need to get back to it through the changelog)

WDYT about:
```suggestion
- Added `Bip44AccountProvider.deleteAccount(id)` method ([#8960](https://github.com/MetaMask/core/pull/8960))
  - The `KeyringController` will automatically prunes the non-primary empty keyrings when the last EVM account is getting removed.
  - `AccountProviderWrapper.deleteAccount(id)` always removes the account, even if disabled.
- Added `AccountProviderWrapper.unwrap` method ([#8960](https://github.com/MetaMask/core/pull/8960))
  - Use this if you need to access the inner (wrapped) keyring.
```

packages/multichain-account-service/CHANGELOG.md
```diff
@@ -7,8 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - The method now iterates every account belonging to the wallet and dispatches `provider.deleteAccount(account.id)` to the matching provider, instead of only removing a single EVM address.
+  - For wrapped providers, enumeration uses the underlying provider so snap-backed accounts are still deleted when basic functionality is off (the wrapper's `getAccounts()` returns `[]` when disabled). Without this, snap-backed accounts would be orphaned in their underlying keyrings.
+  - Per-account deletions are best-effort: a single account's failure does not abort cleanup of the remaining accounts. If one or more deletions fail, a single aggregated error is reported via the messenger's `captureException` with the per-account failure details (`provider`, `accountId`, error message, stack) in its `context`. The wallet is always removed from the service's internal map at the end.
```
ccharly: WDYT about?

```suggestion
- **BREAKING:** `MultichainAccountService.removeMultichainAccountWallet` (and messenger action) now takes a single `entropySource` argument ([#8960](https://github.com/MetaMask/core/pull/8960))
  - The previous `accountAddress` parameter has been removed.
  - All accounts are now unconditionally removed from the wallet and providers (even for disabled `AccountProviderWrapper`).
  - Per-account deletions are best-effort: a single account's failure does not abort cleanup of the remaining accounts.
  - Errors are aggregated and reported in case of failure.
```

packages/multichain-account-service/src/MultichainAccountService.ts
hmalik88: hmm yeah, agreed. [939b6fd](https://github.com/MetaMask/core/pull/8960/commits/939b6fd0fc22dcc7f269ca56769df5cfb04fbda8)

packages/multichain-account-service/CHANGELOG.md
hmalik88: [eee6880](https://github.com/MetaMask/core/pull/8960/commits/eee68806da85b4de8b07ffe023094bcdca844e1c)

packages/multichain-account-service/CHANGELOG.md
hmalik88: [eee6880](https://github.com/MetaMask/core/pull/8960/commits/eee68806da85b4de8b07ffe023094bcdca844e1c)

ccharly (approved):
I really liked how you handle the aggregated failures with the context, we should probably re-use that pattern for the other kind of errors (with a context) 💪

PR #8961, opened 2026-06-01
Add basic data service docs: intro + initial tutorial

mcmire:
## Explanation

This commit begins to flesh out the Data Services portion of the Wallet Framework documentation by adding two sections:

- an introduction that briefly explains what data services are, what benefits they provide, and when to use one
- a tutorial that walks through how to initialize a data service and add methods to make read-only requests (roughly porting over the existing tutorial in `docs/code-guidelines/data-services.md`)

To read the docs:

- [**View on GitHub**](https://github.com/MetaMask/core/tree/initialize-data-services-docs/packages/wallet-framework-docs/content/data-services) (note: some things are missing, such as full code examples at the bottom of parts)
- View locally by running `yarn workspace @metamask/wallet-framework-docs run dev`

## Screenshots

Screenshot 2026-06-01 at 1 56 50 PM

Screenshot 2026-06-01 at 1 56 52 PM

## References

https://consensyssoftware.atlassian.net/browse/WPC-1003

## Checklist

---

> [!NOTE]
> **Low Risk**
> Documentation and docs-package tooling only; no changes to production controller or data-service runtime code.
> 
> **Overview**
> Adds **Data Services** content to `wallet-framework-docs`: a short **introduction** (what they are, features, when to use them), a section index, and the first two parts of a **five-part “Writing a data service” tutorial** (getting started with `OrdersService` boilerplate; read-only `fetchOrders` / `fetchOrder` with `fetchQuery`, Superstruct validation, messenger actions, and nock-based tests). The home page now links to the new **Data services** section.
> 
> Tutorial pages use **MDX** with `raw-loader` snapshots under `content/data-services/writing-data-services/` so full examples can be embedded in collapsible code blocks. **`@metamask/wallet-framework-docs`** gains **dependencies** on framework packages used by those examples, **`nock`** for the example tests, **TypeScript project references** and **`content/`** in `tsconfig`, **knip** scanning of `content/**/*`, and **ESLint ignores** for the Docusaurus build output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e2954831d31b54428dc972bdda53e5fe50181a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/orders-service.ts
mcmire, line 1: Each part in the tutorial has a directory, and the state of the project built over the course of the tutorial is kept in each directory. 

I've opted to do this instead of keeping these files in `sample-controllers` because I think the `sample-controllers` package has a different purpose now. Its purpose is to back sample features in the clients, whereas the project being developed in this tutorial is just an example, it's not meant to be used publicly. 

That said, all of the TypeScript files in `packages/wallet-framework-docs` are typechecked, and you can run the tests just like you would any other package.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
```diff
@@ -0,0 +1,820 @@
...
+  CaipAccountIdStruct,
+  CaipAssetIdStruct,
+  CaipAssetTypeStruct,
+} from '@metamask/utils';
+
+/**
+ * A struct that represents a timestamp (number of seconds since the UNIX
+ * epoch).
+ */
+const TimestampStruct = refine(number(), 'timestamp', (value) => {
```
mcmire: I wanted to make this tutorial somewhat realistic and not contrived to demonstrate how `superstruct` and the utilities in `@metamask/utils` can be used, but I'm curious now whether I've made this slightly too complicated. Happy to hear y'all's thoughts.

packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/README.mdx
mcmire, line 1: Goals for this tutorial:

1. It should be easy to follow for native and non-native English speakers. It should be friendly and approachable, and any excessive or confusing language should be cleaned up.
3. The project that the tutorial builds should work — all code should pass lint, pass all tests, etc.
4. The tutorial should address common misunderstandings, caveats, etc. that we've seen from other engineers attempting to use data services.

packages/wallet-framework-docs/content/data-services/introduction.md
mcmire, line 1: This file should accurately communicate the purpose, goals, and features of data services.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
```diff
@@ -0,0 +1,289 @@
...
+  async fetchOrders({
+    sortField = 'createdTime',
+    sortOrder = 'asc',
+  }: {
+    sortField?: 'createdTime' | 'updatedTime';
+    sortOrder?: 'asc' | 'desc';
+  } = {}): Promise<FetchOrdersResponse> {
+    const url = new URL('/v1/orders', BASE_URL);
+    url.searchParams.append('sortField', sortField);
+    url.searchParams.append('sortOrder', sortOrder);
```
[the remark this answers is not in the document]
mcmire: This isn't a contradiction, but I agree it's confusing. In writing this tutorial, I am assuming that if the order is not specified, the _server_ returns results latest-first. But `fetchOrders` always specifies an order. I guess that level of detail isn't necessary though. Updated here: [ce1a155](https://github.com/MetaMask/core/pull/8961/commits/ce1a15513e2f47e12802abc11e5a23873234a893)

packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/orders-service.ts
Mrtenz: > That said, all of the TypeScript files in `packages/wallet-framework-docs` are typechecked, and you can run the tests just like you would any other package.

Great idea! Definitely helps to keep these maintained.

packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/README.mdx
```diff
@@ -0,0 +1,363 @@
...
+
+We want our data service to live in its own package. (In practice, this step may not be necessary — it depends on your use case — but it makes this tutorial simpler.)
+
+Working from this repo, we'll use the `create-package` tool to do just that:
+
+```
+yarn create-package --name orders-service --description "Wraps the Orders API"
+```
+
+That should give us a new directory in `packages/orders-service` which has a `src/` directory. We'll remove `index.ts` and `index.test.ts` so we have a clean slate.
```
Mrtenz: Maybe good to mention here that we prefer a separate pull request for scaffolding, to make reviews easier.

packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/README.mdx
```diff
@@ -0,0 +1,363 @@
...
+  DataServiceInvalidateQueriesAction<typeof DATA_SERVICE_NAME>;
+
+/**
+ * Actions that {@link OrdersService} exposes to other consumers.
+ */
+export type OrdersServiceActions = OrdersServiceInvalidateQueriesAction;
+
+// highlight-start
+/**
+ * ACtions from other messengers that {@link OrdersService} calls.
```
Mrtenz: Nit:
```suggestion
 * Actions from other messengers that {@link OrdersService} calls.
```

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    const responseData = await this.fetchQuery({
+      queryKey: [`${this.name}:fetchOrders`, url.toString()],
```
Mrtenz: Is it a good practice to include the full URL in the `queryKey`? Should it just be the `sortField` and `sortOrder` fields (since everything else is static)?

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
```diff
@@ -0,0 +1,289 @@
...
+   * Retrieves details about an order.
+   *
+   * @param id - The order ID.
+   * @returns The requested order.
+   */
+  async fetchOrder(id: string): Promise<FetchOrderResponse> {
+    const url = new URL(`/v1/orders/${id}`, BASE_URL);
+
+    const responseData = await this.fetchQuery({
+      queryKey: [`${this.name}:fetchOrder`, url.toString()],
```
Mrtenz: Same question here.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
```diff
@@ -0,0 +1,827 @@
...
+    });
+
+    // ...
+  }
+}
+```
+
+Now on to `queryFn`. There is nothing special about making the request itself, but afterward, we carry out two steps:
+
+1. We check the HTTP status of the response and throw error if it is not within the 200-299 range. We use a special error, `HttpError`, which comes from `@metamask/controller-utils`.
```
Mrtenz: Technically, the 300-399 range isn't errors either, but most of the time `fetch` should handle those itself.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
```diff
@@ -0,0 +1,827 @@
...
+Now that we have a method called `fetchOrders`, we need to make sure to add it to `MESSENGER_EXPOSED_METHODS` so that it will become an action on the messenger:
+
+```typescript title="packages/orders-service/src/orders-service.ts"
+// ::diff-deleted-next::
+const MESSENGER_EXPOSED_METHODS = [] as const;
+// ::diff-added-next::
+const MESSENGER_EXPOSED_METHODS = ['fetchOrders'] as const;
+```
+
+Then we'll run `yarn workspace @metamask/orders-service run generate-action-types`. This will generate a new file:
```
Mrtenz:
```suggestion
Then we'll run `yarn workspace @metamask/orders-service run messenger-action-types:generate`. This will generate a new file:
```

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
```diff
@@ -0,0 +1,827 @@
...
+
+```typescript title="packages/orders-service/src/orders-service.ts"
+const MESSENGER_EXPOSED_METHODS = [
+  'fetchOrders',
+  // highlight-next-line
+  'fetchOrder',
+] as const;
+```
+
+We'll run `yarn workspace @metamask/orders-service run generate-action-types` and see that `packages/orders-service/src/orders-service-method-action-types.ts` has this additional content:
```
Mrtenz:
```suggestion
We'll run `yarn workspace @metamask/orders-service run messenger-action-types:generate` and see that `packages/orders-service/src/orders-service-method-action-types.ts` has this additional content:
```

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
FrederikBolding: My understanding is that the best practice is using a name and the parameters, so I would agree.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
```diff
@@ -0,0 +1,827 @@
...
+      jsonResponse,
+      FetchOrdersResponseStruct,
+    );
+    if (error) {
+      throw new Error(
+        `Malformed response received from Orders API (${error.toString()})`,
+      );
+    }
+
+    return validatedJsonResponse;
```
cryptodev-2s: Nit: small typo
```suggestion
    return validJsonResponse;

File: packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx [hunk 11]
```diff
@@ -0,0 +1,827 @@
...
+Finally, we end our method by handling the response.
+
+We want to ensure that the data we get back from the server is in a format we expect. We _could_ simply typecast the response data, but instead, we use the [Superstruct](https://docs.superstructjs.org/) library to define a schema, and we match the data against it. (We assume that there are two kinds of orders, "token" and "asset", and that the `objectId` is different for each kind.)
+
+```typescript title="packages/orders-service/src/orders-service.ts"
+import type { Infer } from '@metamask/superstruct';
+import {
+  array,
+  intersection,
+  literal,
```
cryptodev-2s, line 149: Nit: Missing `number` import
```suggestion
  literal,
  number,

File: packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx [hunk 12]
```diff
@@ -0,0 +1,827 @@
...
+At this point we've implemented `GET /v1/orders` and we know it works.
+
+Now let's implement `GET /v1/orders/:id`. We'll do that by adding a `fetchOrder` method that follows the same steps as we did above. Luckily, we can assume that this endpoint returns data in a similar structure as the `/v1/orders` endpoint, which means we can reuse some types.
+
+```typescript title="packages/orders-service/src/orders-service.ts"
+/**
+ * Struct to validate what `GET /v1/orders/:id` returns.
+ */
+const FetchOrderResponseStruct = type({
+  order: OrderStruct,
```
cryptodev-2s: Nit: should this be ?
```suggestion
  order: ResponseOrderStruct,

File: packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx [hunk 13]
```diff
@@ -0,0 +1,827 @@
...
+### Writing a basic test
+
+Now we can write a couple of tests for the happy path. We define some response data we know will pass validation — it doesn't really matter what this is, but we try to make it realistic — and we make a new service and fetch the orders. We test our method both without any parameters and with all of the parameters, to make sure it uses the correct URL.
+
+:::info
+You may notice that we test the messenger action, not the method. Why? Services are designed to be used in any part of the stack, and the messenger is the primary interface. So we want to test how the service will be used in practice.
+:::
+
+```typescript title="packages/orders-service/src/orders-service.test.ts"
+const MOCK_VALID_RESPONSE_DATA = {
```
cryptodev-2s: Nit: what do you think about renaming this to
```
const MOCK_VALID_ORDERS_RESPONSE_DATA = {

File: packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx [hunk 14]
```diff
@@ -0,0 +1,827 @@
...
+/**
+ * The data that `GET /v1/orders/:id` returns.
+ */
+type FetchOrderResponse = Infer<typeof FetchOrderResponseStruct>;
+
+export class OrdersService extends BaseDataService</* ... */> {
+  // ...
+
+  /**
+   * Uses the API to retrieve details about an order.
```
cryptodev-2s: Just a question here do we really need to each time specify that this would use the API ?

cryptodev-2s (commented):
Some minor suggestions

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
mcmire: I suppose this is a new best practice I am introducing, but there were two reasons I opted to use the whole URL in the cache key instead of just the query parameters:

1. If a maintainer changes the URL being requested or changes any of the query params being passed to the request, then it seems natural to bust the cache.
2. In cases where the request URL varies based on the current environment then it also seems natural to bust the cache if that environment is changed (e.g. if an environment variable controls whether the "dev" variant of the API URL is hit or the "prod" variant and the engineer changes that environment variable and rebuilds the extension or mobile app).
3. When passing an explicit list of query parameters, the engineer may forget to update that list if they add a new query parameter. By passing a stringified version of the URL then all of the query parameters passed in the request will show up in the cache key automatically.
4. When using the whole URL in the cache key the pattern to make a request is more copy-and-pasteable to other data services (there is less code to change in order to adapt to the new data service).

Does that logic make sense? Is it worth deviating from TanStack Query for these reasons?

FrederikBolding: It is baked in as an assumption in the `BaseDataService` that we follow the existing pattern: https://github.com/MetaMask/core/blob/main/packages/base-data-service/src/BaseDataService.ts#L30
https://github.com/MetaMask/core/blob/main/packages/react-data-query/src/createUIQueryClient.ts#L69-L84

I find that it is quite nice that it maps directly to the messenger action + parameters as well.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: I ended up simplifying this.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: Oh, I was referring to our use of `.ok`. Does it only return `true` for 2xx responses? https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: Good catch, fixed here: [0b04f0a](https://github.com/MetaMask/core/pull/8961/commits/0b04f0aae2fde17dd4b8e67f7f3ffbae7b6d07ca)

packages/wallet-framework-docs/content/data-services/writing-data-services/01-getting-started/README.mdx
mcmire: Sure, I've added a note here: [ed5541f](https://github.com/MetaMask/core/pull/8961/commits/ed5541f2ae8816dd55829e69325eab8862943ed8)

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: `number` wasn't missing, but `integer` was. Fixed here: [654c42e](https://github.com/MetaMask/core/pull/8961/commits/654c42e5d8b0f2863d9cb7f6e42d8329b5c5a9e9)

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: Good catch. Fixed here: [b98ff5f](https://github.com/MetaMask/core/pull/8961/commits/b98ff5f170f1bc034a61044dbc0adf1fba62a2a4)

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/README.mdx
mcmire: You're right. Fixed in [da018fc](https://github.com/MetaMask/core/pull/8961/commits/da018fc29600a543e54dfee0fb7fe129cc9e68e0).

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
mcmire: Okay I've updated this section to better follow our current best practices: [bb7155a](https://github.com/MetaMask/core/pull/8961/commits/bb7155ac6f78df99eea45ae7a5555b8e0633d862), [6f87e45](https://github.com/MetaMask/core/pull/8961/commits/6f87e451364bc74106dc5180d30f0675d9f1a5ee).

But I really think we should consider updating these practices, because I think it's not optimal and could lead to over- or under-caching. For instance, in [`getActivity` in ExampleDataService](https://github.com/MetaMask/core/blob/main/packages/base-data-service/tests/ExampleDataService.ts#L111) we're receiving an `address`, then normalizing it before we pass it as a query parameter. However, we are passing the un-normalized `address` to `queryKey`. This means that if our `getActivity` method is called two times with differently cased but equivalent addresses, the resulting requests will be cached under different keys even though the underlying URL is the same. We're also not including the `page` parameter in the cache key, so only the first page will be cached while subsequent pages won't be.

If we used the whole URL in the cache key, however, then these problems would be magically solved.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
mcmire: Fixed in [bb7155a](https://github.com/MetaMask/core/pull/8961/commits/bb7155ac6f78df99eea45ae7a5555b8e0633d862) and [6f87e45](https://github.com/MetaMask/core/pull/8961/commits/6f87e451364bc74106dc5180d30f0675d9f1a5ee), but also see above.

packages/wallet-framework-docs/content/data-services/writing-data-services/02-read-only-requests/orders-service.ts
FrederikBolding: For `fetchInfiniteQuery`, the cache entry contains all pages, newly fetched pages are appended.

Good call on not normalizing `address` though, we probably should for the example.

mcmire: Ah ok, I stand corrected.

Mrtenz (dismissed):

mcmire:
I've addressed all feedback above, this is ready for review again.

Mrtenz (approved):

PR #8967, opened 2026-06-02
feat: adding support for postQuote transactions with paymentOverride defined

jpuri:
## Explanation

Adding support for postQuote transactions with paymentOverride defined. This is done by calling getPaymentOverrideData when and adding additional transactions to quote request.

## References
Related to https://consensyssoftware.atlassian.net/browse/CONF-1407

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how Relay quote bodies are built for Money Account post-quote flows (transfers, delegation, and client-provided override calls), which affects payment execution though behavior is gated and covered by new tests.
> 
> **Overview**
> Adds **Money Account** handling for Relay quotes when `isPostQuote` and `paymentOverride === PaymentOverride.MoneyAccount`. Instead of skipping transaction embedding like other post-quote flows, the Relay strategy now calls `getPaymentOverrideData`, builds an `EXACT_OUTPUT` quote with a funding token transfer plus override batch calls, and optionally applies EIP-7702 `authorizationList` and a custom funding `recipient`.
> 
> `GetPaymentOverrideDataResponse` gains optional **`recipient`** and **`authorizationList`** so clients can steer the transfer target and delegation. Authorization normalization is shared via **`normalizeAuthorizationList`** (also used by the existing delegation path in `processTransactions`).
> 
> Changelog and a dedicated **`processMoneyAccountPostQuote`** test suite cover trade type, `txs` composition, empty override calls, and defaults for amount/value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4311eec654f3a5c418ea74c68c4363635edf308f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Adding processing for postQuote transactions with paymentProvider defined ([#8967](https://github.com/MetaMask/core/pull/8967))
```
matthewwalsh0: Do you mean `paymentOverride`, same for PR description? 😅

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -378,6 +391,59 @@ async function processTransactions(
...
+  requestBody: RelayQuoteRequest,
+  messenger: TransactionPayControllerMessenger,
+): Promise<void> {
+  const { transactionData: transactionDataList } = messenger.call(
+    'TransactionPayController:getState',
+  );
+
+  const transactionData = transactionDataList[transaction.id];
+  const amountHuman = transactionData?.tokens?.[0]?.amountHuman ?? '0';
+  const amountRaw =
```
matthewwalsh0: Why do we have to check the state directly? Shouldn't this always be the `sourceTokenAmount` in `request`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -378,6 +391,59 @@ async function processTransactions(
...
+    recipient,
+    authorizationList,
+  } = await messenger.call('TransactionPayController:getPaymentOverrideData', {
+    amount: amountHuman,
+    transaction,
+    transactionData,
+  });
+
+  if (!overrideCalls.length) {
+    log('No payment override calls for money account post-quote');
```
matthewwalsh0, line 418: Should we throw in this case as something has gone wrong?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: Here it is possible in a future use case the callback conditionally returns calls. Thus we should not throw if there are no calls.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: I changed it for amountRaw.

packages/transaction-pay-controller/CHANGELOG.md
jpuri: My bad, changelog and PR description are updated.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: We do not have amountHuman available, an option is that we send amountHuman to callback rather than amountHuman also. That may be more accurate.

matthewwalsh0 (approved):

PR #8982, opened 2026-06-03
fix: normalize balances

salimtb:
## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how all balance amounts are stored and emitted across controller, websocket, and selectors; truncation rules must stay aligned with extension parsing to avoid subtle balance display errors.
> 
> **Overview**
> Fixes **dust and tiny balances** showing up as scientific-notation strings (e.g. `"1e-18"`) instead of plain decimals, which broke extension code that parses amounts with `BigInt()`.
> 
> Adds **`normalizeAmountString`** and runs it when **`AssetsController`** merges balance updates: exponent form is always rewritten; when **`assetsInfo`** decimals are known, amounts are truncated (round-down) to that precision, otherwise only the notation is fixed so early balances are not stripped.
> 
> **`BackendWebsocketDataSource`** now uses **`toFixed()`** instead of **`toString()`** on human-readable amounts, and the aggregated balance selector emits **`toFixed()`** amounts for the same invariant.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a387ed8bc4533c977e995e97ba167b88ee545ec1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
```diff
@@ -657,10 +657,14 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
         ? BigInt(postBalance.amount).toString()
         : postBalance.amount;
 
-      // Convert to human-readable using asset decimals (match RpcDataSource / pipeline format)
+      // Convert to human-readable using asset decimals (match RpcDataSource / pipeline format).
+      // `.toFixed()` (no args) — not `.toString()` — because BigNumber's default
+      // `EXPONENTIAL_AT: [-7, 21]` makes `.toString()` emit scientific notation
+      // for sub-1e-7 values (e.g. `"1e-18"`), which downstream `BigInt()` parsers
+      // reject and crash on (MMBUGS-772).
```
Prithpal-Sooriya: I don't think we need this comment

```suggestion

```

packages/assets-controller/src/selectors/balance.ts
```diff
@@ -455,7 +455,11 @@ export function getAggregatedBalanceForAccount(
     const { amount } = row;
     const entry: AggregatedBalanceEntry = {
       assetId,
-      amount: amount.toString(),
+      // `.toFixed()` — not `.toString()` — because BigNumber's default
+      // `EXPONENTIAL_AT: [-7, 21]` makes `.toString()` emit scientific notation
+      // for sub-1e-7 values (e.g. `"1e-18"`), which downstream `BigInt()`
+      // parsers reject (MMBUGS-772).
```
Prithpal-Sooriya: Same lets remove this comment


```suggestion
```

packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
salimtb: the toString here is converting into scientific annotation in case of small balances which can cause errors

salimtb: ooh true , removing this comment now

Prithpal-Sooriya (approved):

PR #8987, opened 2026-06-03
fix: MMPay fiat payment fixes 

OGPoyraz:
## Explanation

Fiat payments for `moneyAccountDeposit` transactions fail with multiple sequential errors after the on-ramp order settles:

1. **`isMaxAmount: true` in the fiat re-quote** — after Transak settles, `submitRelayAfterFiatCompletion` re-quotes with `isMaxAmount: true, isPostQuote: false`. This calls `processTransactions`, which throws `"Max amount quotes do not support included transactions"` because moneyAccountDeposit has nested txs (approve + deposit) that require delegation.

2. **`validateRelaySlippage` compares wrong amounts** — it compares `currencyOut.amount` from two relay quotes made with different source amounts (theoretical $5 quoting phase vs actual ~$4.95 post-Transak settlement). This produces ~25% apparent "slippage" that is not real slippage — it is just a smaller input producing a smaller output.

3. **Nested calldata has zero amounts** — the fiat path in `handleDone` only sets `amountFiat` and never calls `updateTokenAmount()`, so `requiredAssets.amount` stays `0x0` and the nested approve + deposit calldata encodes zero amounts.

4. **Wrong wallet address in fiat flow** — `fiat-quotes.ts` and `fiat-submit.ts` used `transaction.txParams.from` as the wallet address. For `moneyAccountDeposit`, `txParams.from` is the money account address on the target chain (Monad), not the user's EOA. This caused Ramps/Transak to receive the wrong deposit address, `resolveSourceAmountRaw` to look for on-chain ETH at the wrong address, and all relay quotes to use the wrong `from`/`user` address — resulting in `TRANSFER_FROM_FAILED` reverts.

5. **Fiat total calculation using wrong amount** — `calculateTotals` derived the payment amount from token amounts or `targetAmount`, which is incorrect for fiat flows where the user enters a specific fiat amount.

### Changes

**`fiat-quotes.ts`** — reads `walletAddress` from the new `from` field on `PayStrategyGetQuotesRequest` instead of re-deriving `accountOverride ?? txParams.from` from state. The `from` field is pre-computed by the quote orchestrator in `quotes.ts`.

**`fiat-submit.ts`** — two wallet address fixes plus split submit paths:
- `submitFiatQuotes`: uses `accountOverride ?? txParams.from` for order polling wallet address
- `submitRelayAfterFiatCompletion`: uses `baseRequest.from` (already accountOverride-aware from the quote) for on-chain amount lookup
- Routes to `submitSimpleRelay` (Perps/Predict — single EXACT_INPUT) or `submitWithTransactionData` (moneyAccountDeposit — three-phase flow)

**`fiat-submit-with-transaction-data.ts`** (new file, renamed from `fiat-submit-with-calldata.ts`) — three-phase relay flow for nested calldata:
- **Phase 1**: Discovery relay quote (`isPostQuote: true`, EXACT_INPUT) with fee reserve subtracted from settled source amount
- **Phase 2**: Calls `getAmountData` via messenger to delegate calldata re-encoding to the client, then patches nested tx data + `requiredAssets[0].amount`
- **Phase 3**: Final relay quote (`isPostQuote: false`, EXACT_OUTPUT) with delegation, followed by `validateRelayRateDrift`
- Fee reserve uses `totalImpact.usd` from the original relay quote directly

**`fiat-submit-simple.ts`** (new file) — single EXACT_INPUT post-quote relay for simple deposits (Perps, Predict) that don't need calldata re-encoding.

**`relay-submit.ts`** — gas price alignment: the prepended original transaction in post-quote batch submissions now carries the relay step's `maxFeePerGas` and `maxPriorityFeePerGas` (including the delegated path via `buildDelegatedOriginalParams`). This ensures the top-level 7702 batch uses the same gas price as the relay quote.

**`validateRelaySlippage` → `validateRelayRateDrift`** — compares USD exchange rate ratios (output_usd / input_usd) between original and final quotes instead of absolute amounts, normalising for different source amounts.

**`TransactionPayController`** — adds optional `getAmountData` callback on the constructor, exposed via messenger as `TransactionPayController:getAmountData`. Keeps ABI knowledge on the client side.

**`types.ts`** — adds `from: Hex` to `PayStrategyGetQuotesRequest` so strategies receive the pre-resolved wallet address (`accountOverride ?? txParams.from`) without re-deriving it.

**`totals.ts`** — adds `fiatPaymentAmount` parameter to `calculateTotals`. When a fiat strategy quote is present, uses the user-entered fiat payment amount directly instead of deriving it from token amounts or relay `targetAmount`. Falls back to `'0'` if fiat amount is unavailable.

**`quotes.ts`** — passes `fiatPayment.amountFiat` from the transaction pay state into `calculateTotals` as `fiatPaymentAmount`. Threads the resolved `from` address through `getQuotes` into `PayStrategyGetQuotesRequest`.

**`relay-quotes.ts` (recipient routing fix)** — adds `skipProcessTransactions` to `QuoteRequest` type. Defaults to `isPostQuote` when not set. The simple fiat path (`fiat-submit-simple.ts`) sets `skipProcessTransactions: false` to force `processTransactions` to run and extract the `transfer(to, amount)` recipient from calldata — fixing Predict/Perps deposits where Relay sent swapped tokens to the user's EOA instead of the target contract.

**`relay-quotes.ts` (7702 batch gas estimation)** — for post-quote flows, includes the original transaction in batch gas estimation alongside relay params. Previously, a single relay step was estimated alone, got `is7702: false`, and the batch fell back to separate type-0x2 transactions — breaking zero-balance fiat-funded accounts that need all native tokens for the swap.

**`relay-quotes.ts` (native gas subtraction)** — extends the phase-2 gas subtraction to trigger when the source token is the native gas token (using `getNativeToken()` — simplified from redundant dual check). Adds a surplus check: skips subtraction when `sourceBalanceRaw - sourceTokenAmount` already covers gas. Early-returns for `isExecute` flows (Relay's relayer handles gas).

**`relay-max-gas-station.ts`** — early-returns the phase-1 quote for `isExecute` flows, skipping the entire gas station estimation since Relay's relayer handles gas.

**`feature-flags.ts`** — `DEFAULT_FEE_RESERVE_MULTIPLIER` (1.2) and `DEFAULT_MAX_RATE_DRIFT_PERCENT` (10) moved to top of file and exported. Removed `getPostQuoteGasBuffer` feature flag function — gas buffer is now a hardcoded constant (`POST_QUOTE_GAS_BUFFER = 1.1`) in `relay-quotes.ts`.

## Feature Flags (LaunchDarkly)

This PR reads optional properties from one feature flag. No new LaunchDarkly flags need to be created — these are properties on an existing flag object.

### `confirmations_pay_fiat`

| Property | Type | Default | Description |
|----------|------|---------|-------------|
| `feeReserveMultiplier` | `number` | `1.2` | Multiplier applied to the original relay fee when reserving source tokens for the discovery quote. Increase (e.g. `1.5`) if EXACT_OUTPUT quotes consistently exceed the settled balance after settlement on cross-chain routes. |
| `maxRateDriftPercent` | `number` | `10` | Maximum allowed relay exchange rate drift (%) between the original fee quote and the post-settlement discovery quote. Increase if legitimate market moves cause false rejections. |

**No action required for launch** — all properties default to safe values when absent. They exist as remote safety valves to tune post-settlement behavior without a code release.

## References

* Consumer PR: MetaMask/metamask-mobile#31009 (branch `ogp/fiat-money-account-deposit-fix`)

## Checklist

---

> [!NOTE]
> **High Risk**
> Touches payment settlement, relay batching, and wallet/accountOverride addressing; incorrect amounts or gas could block deposits or mis-route funds, though behavior is heavily tested and feature-flagged.
> 
> **Overview**
> Fixes **fiat post–on-ramp settlement** for MetaMask Pay, especially `moneyAccountDeposit`, by correcting addresses, totals, relay quoting, and post-settlement submit paths.
> 
> **Wallet & quoting:** Strategies now receive a pre-resolved `from` on `PayStrategyGetQuotesRequest` (`accountOverride ?? txParams.from`). Fiat quote/submit uses that instead of `txParams.from` alone. `calculateTotals` uses the user’s `fiatPayment.amountFiat` when the fiat strategy is active.
> 
> **Post-settlement relay:** After the ramp order completes, submit splits on nested calldata: **simple** deposits (Perps/Predict) use one **EXACT_INPUT** post-quote relay (`submitSimpleRelay`); **nested** flows (e.g. approve + deposit) use a **three-phase** path (`submitWithTransactionData`)—discovery quote with fee reserve, client **`getAmountData`** to re-encode nested calldata, then **EXACT_OUTPUT** delegation quote. Slippage checks move from raw output comparison to **`validateRelayRateDrift`** (USD in/out ratios), with tunables `feeReserveMultiplier` and `maxRateDriftPercent` on `confirmations_pay_fiat`.
> 
> **Relay infrastructure:** Post-quote gas estimation includes the original tx in batch estimates for 7702; native-source post-quote flows subtract buffered gas when needed; `skipProcessTransactions` on quote requests; prepended txs inherit relay gas fees; max-gas-station and post-quote phase-2 skip when `isExecute`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9febec886f7126688958f0a6feb09b37f1e702a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -319,39 +273,77 @@ async function submitRelayAfterFiatCompletion({
...
+    sourceBalanceRaw: sourceAmountRaw,
+    sourceTokenAmount: sourceAmountRaw,
+    targetAmountMinimum: settledTargetRaw,
+  };
+
   const relayQuotes = await getRelayQuotes({
     accountSupports7702: request.accountSupports7702,
     messenger,
     requests: [relayRequest],
-    transaction,
```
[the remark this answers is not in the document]
OGPoyraz: Fixed

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -362,7 +354,7 @@ async function submitRelayAfterFiatCompletion({
     isSmartTransaction: request.isSmartTransaction,
     messenger,
     quotes: relayQuotes,
-    transaction,
+    transaction: updatedTransaction,
```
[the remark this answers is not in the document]
OGPoyraz: Fixed

packages/transaction-pay-controller/src/utils/totals.ts
```diff
@@ -133,6 +146,44 @@ export function calculateTotals({
...
+}: {
+  hasFiatStrategy: boolean;
+  fiatPaymentAmount?: string;
+  isMaxAmount?: boolean;
+  hasQuotes: boolean;
+  targetAmount: string;
+  tokenAmount: string;
+}): string {
+  if (hasFiatStrategy) {
+    return fiatPaymentAmount ?? '0';
```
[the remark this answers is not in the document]
OGPoyraz: That's intentional and won't be an issue as we are using single currency ('USD') for now

packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -9,11 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add optional `getAmountData` callback to `TransactionPayControllerOptions` for client-side nested calldata re-encoding ([#8987](https://github.com/MetaMask/core/pull/8987))
+- Add `TransactionPayController:getAmountData` messenger action ([#8987](https://github.com/MetaMask/core/pull/8987))
```
matthewwalsh0: Can we combine / nest some or all of these?

packages/transaction-pay-controller/src/types.ts
```diff
@@ -177,6 +177,29 @@ export type GetPaymentOverrideDataCallback = (
   request: GetPaymentOverrideDataRequest,
 ) => Promise<GetPaymentOverrideDataResponse>;
 
+export type GetAmountDataRequest = {
+  /** Raw token amount (atomic units) to encode into calldata. */
+  amount: string;
```
matthewwalsh0, line 188: Minor, I think `getPaymentOverrideData` provides the human / formatted amount. Not sure which is easier for the client to handle.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -91,6 +91,7 @@ export async function updateQuotes(
     refundTo,
     sourceAmounts,
     tokens,
+    fiatPayment,
```
matthewwalsh0: Minor, alphabetical.

packages/transaction-pay-controller/src/utils/totals.ts
```diff
@@ -76,26 +79,36 @@ export function calculateTotals({
...
   const totalFiat = new BigNumber(providerFee.fiat)
     .plus(metaMaskFee.fiat)
     .plus(sourceNetworkFeeEstimate.fiat)
     .plus(targetNetworkFee.fiat)
-    .plus(
-      (hasFiatStrategy || isMaxAmount) && hasQuotes
-        ? targetAmount.fiat
-        : amountFiat,
-    )
+    .plus(paymentAmountFiat)
```
matthewwalsh0: For clarity and consistency, could we call this `sourceAmountX` and rename the function to `getSourceAmount`?

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -49,7 +49,8 @@ export async function getFiatQuotes(
   const state = messenger.call('TransactionPayController:getState');
   const transactionData = state.transactionData[transactionId];
   const amountFiat = transactionData?.fiatPayment?.amountFiat;
-  const walletAddress = transaction.txParams.from as Hex;
+  const walletAddress =
+    transactionData?.accountOverride ?? (transaction.txParams.from as Hex);
```
matthewwalsh0: As discussed, this can always use `request.from` as it's the same calculation?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -27,7 +27,7 @@ const log = createModuleLogger(projectLogger, 'fiat-submit');
 
 const ORDER_POLL_INTERVAL_MS = 1000;
 const ORDER_POLL_TIMEOUT_MS = 10 * 60 * 1000;
-const MAX_SLIPPAGE_PERCENT = 5;
+const MAX_RATE_DRIFT_PERCENT = 10;
```
matthewwalsh0: Minor, we discussed a potential feature flag for this in future.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -318,40 +276,90 @@ async function submitRelayAfterFiatCompletion({
...
-  const baseRequest = quotes[0].request;
-  const relayRequest: QuoteRequest = {
+  // Phase 1: Discovery quote with EXACT_INPUT to find the actual target
+  // token output for the settled source amount.
+  const discoveryRequest: QuoteRequest = {
     ...baseRequest,
-    isMaxAmount: true,
-    isPostQuote: false,
+    isMaxAmount: false,
+    isPostQuote: true,
```
matthewwalsh0: This isn't conceptually a post quote, as it's not running after the original intent, it's still trying to achieve the target intent, but this is how we're forcing an `EXACT_INPUT` trade type? But without using the entire balance?

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
matthewwalsh0: On line 95, below, it looks like we were already added the relay fee to the fiat amount?

So was it just the UI that wasn't reflecting that?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -318,40 +276,90 @@ async function submitRelayAfterFiatCompletion({
...
+    );
+  }
+
+  if (updates.length) {
+    updateTransaction(
+      { transactionId, messenger, note: 'Fiat deposit: update settled amount' },
+      (tx) => {
+        for (const { nestedTransactionIndex, data } of updates) {
+          if (tx.nestedTransactions?.[nestedTransactionIndex]) {
+            tx.nestedTransactions[nestedTransactionIndex].data = data;
```
matthewwalsh0: Do we not to update `data` also? In which case should we look at `updateAtomicData` in the transaction controller? Or have the new callback return the full `data` also?

Or are we just updating enough so the Relay quote can generate the delegation and include in the quote?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -318,40 +276,90 @@ async function submitRelayAfterFiatCompletion({
...
+  }
+
+  const updatedTransaction =
+    getTransaction(transactionId, messenger) ?? transaction;
+
+  // Phase 3: Real relay quote with delegation (standard crypto-like flow).
+  const relayRequest: QuoteRequest = {
+    ...baseRequest,
+    isMaxAmount: false,
+    isPostQuote: false,
```
matthewwalsh0: Does this not need to be conditional?

If we're doing a Perps or Predict deposit for example, we won't need transaction data, so we will want to use `EXACT_INPUT` instead of `EXACT_OUTPUT` so we can ensure nothing left over and cheaper Relay fees.

We won't even need the first discovery quote in that case so can avoid an entire additional request?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -318,40 +276,90 @@ async function submitRelayAfterFiatCompletion({
...
+      },
+    );
+  }
+
+  const updatedTransaction =
+    getTransaction(transactionId, messenger) ?? transaction;
+
+  // Phase 3: Real relay quote with delegation (standard crypto-like flow).
+  const relayRequest: QuoteRequest = {
+    ...baseRequest,
```
matthewwalsh0: If we're doing an `EXACT_OUTPUT` for money account deposit, what do we do with any remaining funds on the selected account after the fiat quote?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -318,40 +276,90 @@ async function submitRelayAfterFiatCompletion({
...
+    isMaxAmount: false,
+    isPostQuote: false,
+    sourceBalanceRaw: sourceAmountRaw,
+    sourceTokenAmount: sourceAmountRaw,
+    targetAmountMinimum: settledTargetRaw,
+  };
+
   const relayQuotes = await getRelayQuotes({
```
matthewwalsh0: Suppose we do a Ramps quote that gets $100 into the EOA.

And we get an `EXACT_INPUT` quote that tell us we can get $95.

And then we get our final `EXACT_OUPUT` quote targeting $95... Except the fees are higher in general on `EXACT_OUTPUT`, plus we now have our additional transaction data included. So what if it says the required source amount is $101?

Do we need to leave a buffer for additional cost so intentionally ask for less in the second quote? But that means we always have funds left over.

Or do we retry with repeating lower amounts until it's less than our balance? Meaning we have to repeatedly re-encode the amount.

Or do we just throw if we can't afford it? But that's not the user's fault so no way for them to remedy.

Essentially we have to consider the slippage twice, from original to discovery, and discovery to final.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/CHANGELOG.md
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: Done

OGPoyraz: Yes but this is just a quote, real order flow (headless buy) starts when we click "Add funds" in UI where we need the amount again (in `useFiatConfirm`).

You may ask "why we are not passing the fiat quote amount while starting headless buy", but somehow I never see what amount we requested originally in the quote, either less or more amount saved in fiat quote entity, not sure why. Hence I subtract the fiat provider fee from `totals.total` in client
https://github.com/MetaMask/metamask-mobile/pull/31009/files#diff-aafbaf61a37d99028f87dea7fbe81be4508bc20412d534db932c1a1cd83e59b6

packages/transaction-pay-controller/src/utils/totals.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/types.ts
OGPoyraz: `buildMoneyAccountDepositBatch` in client was expecting raw amount hence I passed as is.
For reference: https://github.com/MetaMask/metamask-mobile/pull/31009/files#diff-c2390f54bc74c73163ad5949d6accb3524100bf2dd425db4dd82c1ed73d7076f

packages/transaction-pay-controller/src/utils/quotes.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Exactly, there are few things may go into FFs, I prefer to handle them at once

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Exactly as you said it's not a post quote at all, just to force `EXACT_INPUT`. 

The intent here is to find out target token output because only at this point we have the settled order amount. This quote basically does nothing else.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Just updating enough parts so only `nestedTransactions` as other parts will be unnecessary for final quote happening in the Phase3 below.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: There is no clean way to avoid it, unfortunately it will leave some dust. 

If we do the `EXACT_INPUT` for everything then the vault deposit calldata amount won't match what Relay delivers, causing reverts

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Very valid point, I will make sure perps predict cases to use `EXACT_OUTPUT`, thanks.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Understood, so depends on the spread between `EXACT_INPUT` and `EXACT_OUTPUT` fees + the delegation gas overhead. I didn't foresee this point as I always did my tests on already delegated account. 

There is no easy solution for that. But we can try adding potential buffer here for the non-delegated cases, which we can assume discovery phase determine partially (for fees) that. But not sure how can we determine the cost of delegation, maybe we can apply some dedicated amount for that. Would it make sense?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-transaction-data.ts
```diff
@@ -0,0 +1,253 @@
...
+        }
+      }
+      if (tx.requiredAssets?.[0]) {
+        tx.requiredAssets[0].amount = `0x${new BigNumber(adjustedTargetRaw).toString(16)}`;
+      }
+    },
+  );
+
+  const updatedTransaction =
+    getTransaction(transactionId, messenger) ?? transaction;
```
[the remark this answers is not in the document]
OGPoyraz: `getAmountData` callback is client-owned - core can't validate which indices need updating without coupling to transaction-type-specific logic. The mobile implementation always returns the correct updates for its deposit type; partial results would be a client bug, not a core concern. The stale snapshot fallback (`?? transaction`) is best-effort for the edge case where a transaction is deleted mid-submission, which doesn't happen in practice

OGPoyraz:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: This is added now

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -326,59 +261,26 @@ async function submitRelayAfterFiatCompletion({
...
+  // the target amount, calldata re-encoding, then a delegation quote.
+  // Simple deposits (Perps, Predict) skip straight to a single EXACT_INPUT
+  // relay quote — cheaper fees, no leftover dust, one fewer request.
+  if (hasNestedCalldata) {
+    return await submitWithCalldataReEncoding({
+      baseRequest,
+      request,
+      sourceAmountRaw,
+      transaction,
+    });
```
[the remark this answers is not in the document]
OGPoyraz: I think this is invalid because `txParams.data` for `moneyAccountDeposit` is the full batch calldata, so singleData is populated, hasNoData is false, and delegation runs correctly.

packages/transaction-controller/src/utils/eip7702.ts
```diff
@@ -205,9 +205,15 @@ export function generateEIP7702BatchTransaction(
...
 
+  const totalValue = transactions.reduce(
+    (sum, tx) => sum + BigInt(tx.value ?? '0x0'),
+    BigInt(0),
+  );
+
   return {
     data,
     to: from,
+    ...(totalValue > BigInt(0) ? { value: toHex(totalValue) } : {}),
```
matthewwalsh0: This isn't how EIP-7702 works. The `to` is the `from` so you'd just be passing value to yourself, plus each call is empowered to pass value as it uses a solidity `call` that includes its own `data`, `to`, and `value`.

What issue were you trying to fix?

packages/transaction-pay-controller/CHANGELOG.md
```diff
@@ -10,15 +10,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Live token balance queries now respect the `confirmations_pay_extended.excludeChainIdsFromInfura` feature flag, skipping the Infura endpoint preference for excluded chains ([#8992](https://github.com/MetaMask/core/pull/8992))
 - Bump `@metamask/assets-controllers` from `^108.3.0` to `^108.5.0` ([#8981](https://github.com/MetaMask/core/pull/8981), [#8999](https://github.com/MetaMask/core/pull/8999))
 - Bump `@metamask/assets-controller` from `^8.0.2` to `^8.3.2` ([#8981](https://github.com/MetaMask/core/pull/8981), [#8985](https://github.com/MetaMask/core/pull/8985), [#8999](https://github.com/MetaMask/core/pull/8999))
+
+### Fixed
+
+- Fix fiat `moneyAccountDeposit` failing after on-ramp settlement by adding `getAmountData` callback for calldata re-encoding and correcting wallet address, quote amount, and slippage validation ([#8987](https://github.com/MetaMask/core/pull/8987))
```
matthewwalsh0: Is this not the same as in `changed`, could we combine into one?

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -49,7 +49,8 @@ export async function getFiatQuotes(
   const state = messenger.call('TransactionPayController:getState');
   const transactionData = state.transactionData[transactionId];
   const amountFiat = transactionData?.fiatPayment?.amountFiat;
-  const walletAddress = transaction.txParams.from as Hex;
+  const walletAddress = (transactionData?.accountOverride ??
```
matthewwalsh0: This is still not needed right as `quotes.ts` does this in the request?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -378,10 +378,17 @@ async function submitTransactions(
...
-  if (!quote.request.isPostQuote && !quote.request.paymentOverride) {
+  // Skip the EOA balance check when the source tokens are not expected to be
+  // in the user's wallet at submit time:
+  // - isPostQuote: tokens are in a Safe/proxy, available after batch execution
+  // - paymentOverride: tokens come from a different source
+  // - isExecute: Relay's relayer handles the source-side transaction
+  const skipBalanceCheck =
+    Boolean(quote.request.isPostQuote) ||
+    Boolean(quote.request.paymentOverride) ||
+    Boolean(quote.original.metamask.isExecute);
```
matthewwalsh0: We still want to validate the balance before doing the execute flow, as it's final validation before the user spends their funds.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -429,6 +436,8 @@ async function submitTransactions(
       : ({
           data: transaction.txParams.data as Hex | undefined,
           from: transaction.txParams.from,
+          maxFeePerGas: normalizedParams[0]?.maxFeePerGas,
```
matthewwalsh0, line 436: Good spot, we use the first params in `addTransaction` later, but if we prepended, that will now be `undefined` so it will just figure it out on the fly during `addTransaction` even though Relay may have provided some good fees.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -143,11 +144,28 @@ async function getQuoteWithPostQuoteGasHandling(
...
 
-  const gasCostRaw = phase1Quote.fees.sourceNetwork.max.raw;
+  // Gas must be subtracted from the source amount when the user's balance
+  // is fully committed to the swap. This applies when gas is paid via an
+  // ERC-20 fee token (isSourceGasFeeToken) OR when the source itself is
+  // the native gas token (gas comes from the same pool as the swap value).
+  const isSourceNative =
+    request.sourceTokenAddress.toLowerCase() ===
+      NATIVE_TOKEN_ADDRESS.toLowerCase() ||
+    request.sourceTokenAddress.toLowerCase() ===
```
matthewwalsh0, line 165: Doesn't `getNativeToken` fallback to `NATIVE_TOKEN_ADDRESS` so we only need the second condition?

amitabh94:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -143,11 +144,28 @@ async function getQuoteWithPostQuoteGasHandling(
 ): Promise<TransactionPayQuote<RelayQuote>> {
   const phase1Quote = await getSingleQuote(request, fullRequest);
 
-  if (!request.isPostQuote || !phase1Quote.fees.isSourceGasFeeToken) {
+  if (!request.isPostQuote) {
```
matthewwalsh0, line 150: Won't the majority of cross-chain withdraws use the execute flow? Meaning we don't need to worry about gas at all, so should early return here if the quote has `origin.metamask.isExecute`?

Also we should probably do the same in `getRelayMaxGasStationQuote`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -143,11 +144,28 @@ async function getQuoteWithPostQuoteGasHandling(
...
+      NATIVE_TOKEN_ADDRESS.toLowerCase() ||
+    request.sourceTokenAddress.toLowerCase() ===
+      getNativeToken(request.sourceChainId).toLowerCase();
+
+  if (!phase1Quote.fees.isSourceGasFeeToken && !isSourceNative) {
+    return phase1Quote;
+  }
+
+  const gasBuffer = getPostQuoteGasBuffer(fullRequest.messenger);
+  const gasCostRaw = new BigNumber(phase1Quote.fees.sourceNetwork.max.raw)
```
matthewwalsh0, line 172: Applies to the existing logic too, but we have no check to see if the user already had sufficient balance for both the withdraw and the native gas / gas fee token already?

We always subtract it even when we might not need to?

And this is assuming we're already not execute flow which should be very rare.

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -821,6 +827,80 @@ export function getFiatAssetPerTransactionType(
   );
 }
 
+const DEFAULT_FEE_RESERVE_MULTIPLIER = 1.2;
```
matthewwalsh0: Top of the file? And exported for unit tests?

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -821,6 +827,80 @@ export function getFiatAssetPerTransactionType(
...
+const DEFAULT_MAX_RATE_DRIFT_PERCENT = 10;
+const DEFAULT_POST_QUOTE_GAS_BUFFER = 1.1;
+
+/**
+ * Returns the fee reserve multiplier for fiat three-phase submit.
+ *
+ * Controls how much of the original relay fee is reserved from the discovery
+ * quote source amount to prevent EXACT_OUTPUT cost overruns.
+ * Defaults to 1 (100% of the original fee).
```
matthewwalsh0: Minor, probably shouldn't put the default in the JSDoc otherwise we have to change it each time?

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -821,6 +827,80 @@ export function getFiatAssetPerTransactionType(
...
+    : DEFAULT_MAX_RATE_DRIFT_PERCENT;
+}
+
+/**
+ * Returns the gas buffer multiplier for post-quote gas reservations.
+ *
+ * @param messenger - Controller messenger.
+ * @returns The gas buffer multiplier.
+ */
+export function getPostQuoteGasBuffer(
```
matthewwalsh0: Do we definitely need a feature flag or buffer for this? This is just to catch drift in the gas fee changing before submit?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
```diff
@@ -0,0 +1,253 @@
...
+ * Phase 3: Delegation quote (EXACT_OUTPUT) with updated nested transaction data.
+ *
+ * @param options - The submission options.
+ * @param options.baseRequest - The base quote request from the original fiat quote.
+ * @param options.request - The original fiat strategy execute request.
+ * @param options.sourceAmountRaw - The settled source amount in atomic units.
+ * @param options.transaction - The transaction metadata.
+ * @returns An object containing the relay transaction hash if available.
+ */
+export async function submitWithCalldataReEncoding({
```
matthewwalsh0: Minor, could just be `submitWithTransactionData` for clarity?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-transaction-data.ts
```diff
@@ -0,0 +1,253 @@
...
+    discoverySourceAmount,
+    sourceAmountRaw,
+    transactionId,
+  });
+
+  const discoveryRequest: QuoteRequest = {
+    ...baseRequest,
+    isMaxAmount: false,
+    isPostQuote: true,
+    sourceBalanceRaw: sourceAmountRaw,
```
matthewwalsh0, line 70: Minor, I don't think the relay strategy uses `sourceBalanceRaw` anywhere, but I see this is pre-existing.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
```diff
@@ -0,0 +1,253 @@
...
+  const sourceRaw = new BigNumber(feeQuote.details.currencyIn.amount);
+  const sourceUsd = new BigNumber(feeQuote.details.currencyIn.amountUsd);
+  const targetUsd = new BigNumber(feeQuote.details.currencyOut.amountUsd);
+
+  if (!sourceUsd.gt(0) || !sourceRaw.gt(0)) {
+    return new BigNumber(0);
+  }
+
+  // Fee in USD = what the relay consumed beyond the target value
+  const feeUsd = sourceUsd.minus(targetUsd);
```
matthewwalsh0: Why are we re-calculating this if this is `fees.provider.usd` on our normalized quote from Relay?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  // Convert fee from USD to raw target token units
+  const usdPerTargetRaw = targetUsd.dividedBy(targetMinRaw);
+  const discoveryFeeInTargetRaw = discoveryFeeUsd
+    .dividedBy(usdPerTargetRaw)
+    .decimalPlaces(0, BigNumber.ROUND_DOWN);
+
+  const adjusted = targetMinRaw.plus(discoveryFeeInTargetRaw);
+
+  log('calculateAdjustedTarget', {
```
matthewwalsh0: Minor, `Adjusted target`?

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-transaction-data.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  // Discovery fee in USD
+  const discoveryFeeUsd = sourceUsd.minus(targetUsd);
+
+  if (!discoveryFeeUsd.gt(0)) {
+    return targetMinRaw.toFixed(0);
+  }
+
+  // Convert fee from USD to raw target token units
+  const usdPerTargetRaw = targetUsd.dividedBy(targetMinRaw);
```
matthewwalsh0, line 233: Might be enough for our purposes, but just a warning we're using the estimated target USD but the minimum target raw, so that may skew the target USD rate.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Done

packages/transaction-controller/src/utils/eip7702.ts
OGPoyraz: Reverted it, this was unintended change.

packages/transaction-pay-controller/CHANGELOG.md
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Reverted - balance check is no longer skipped for execute flows.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: So this is actually fixing the case where user have 0 balance but especially doing `predictDeposit`

`getNativeToken` returns `0x...1010` for Polygon, not `NATIVE_TOKEN_ADDRESS` (`0x...0000`). By this point, normalizeRequest has already converted sourceTokenAddress to Relay convention (`0x...0000`), so neither condition alone covers Polygon. On other chains, `getNativeToken` returns `NATIVE_TOKEN_ADDRESS` directly, so the first condition suffices.

This path is only reached in post-quote fiat flows where `sourceBalanceRaw` equals `sourceTokenAmount` (zero headroom for gas). Regular crypto deposits have existing balance exceeding the swap amount, so the headroom check returns before reaching here.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Added early return in both `getQuoteWithPostQuoteGasHandling` and `getRelayMaxGasStationQuote`

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Removed

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Done, removed ff and hardcoded as `POST_QUOTE_GAS_BUFFER = 1.1` in relay-quotes.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
OGPoyraz: Done, also renamed file according to new name

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
OGPoyraz: I changed it to use `totalImpact` from the quote directly.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit-with-calldata.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: 👍 Also extended the same fix to the delegated path - `buildDelegatedOriginalParams` now copies `maxFeePerGas/maxPriorityFeePerGas` from the relay params too.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Added a remaining balance check - if `sourceBalanceRaw - sourceTokenAmount >= gasCost`, we skip the subtraction. Combined with the `isExecute` early return, execute flows skip gas handling entirely.

matthewwalsh0 (approved):

PR #8975, opened 2026-06-02
feat(wallet): add TransactionController to wallet package

matthewwalsh0:
## Explanation

Adds `TransactionController` to the `@metamask/wallet` package. Wallet initializes the controller during setup; consumers pass platform-specific options via `instanceOptions.transactionController`.

## References

MetaMask/metamask-extension#43182

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [x] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> Expands the default wallet controller graph into transaction submission/signing paths via delegated Keyring and Approval actions; breaking for consumers that assumed TransactionController was not part of default init.
> 
> **Overview**
> **BREAKING:** Default `@metamask/wallet` setup now boots **`TransactionController`** alongside the other core controllers, so consumers get transaction state and messenger actions without wiring it themselves.
> 
> A new initialization module constructs the controller from optional `instanceOptions.transactionController` (including a default `disableSwaps: false`), restores persisted state, and **delegates** parent messenger access for signing, approvals, accounts, network clients, gas estimates, and feature flags. `Wallet` integration tests assert `TransactionController:getState` is available on the root messenger.
> 
> Also adds the `@metamask/transaction-controller` dependency, TypeScript project references, **CODEOWNERS** for the new init path, changelog entry, and focused unit tests for registration, default state, and state hydration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e54b302b4755821e67ceccc967dac5888c7da7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #8988, opened 2026-06-03
feat(wallet): wire TransactionController into default initialization

matthewwalsh0:
## Summary

- Wires `TransactionController` into the wallet package's default initialization
- Uses `TransactionControllerMessenger` directly (no separate init messenger) — same pattern as `ApprovalController` and `KeyringController`
- Delegates all `AllowedActions` and `AllowedEvents` from the root messenger to the controller messenger via `parent.delegate`
- Hardcodes `incomingTransactions: { isEnabled: () => false }` as both extension and mobile share this value
- Exposes `TransactionControllerInstanceOptions` (omits `messenger`, `state`, `incomingTransactions`) in `instanceOptions.transactionController`

### DIFF
--- a/packages/wallet/CHANGELOG.md
+++ b/packages/wallet/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


matthewwalsh0:
Closing — work moved back to #8975

PR #8996, opened 2026-06-04
feat(passkey): reveal SRP and private keys with passkey verification

tanguyenvn:
## **Description**

Passkey-enrolled wallets currently require the wallet password to reveal sensitive key material (Secret Recovery Phrase and account private keys), even when the user has already unlocked MetaMask with a passkey. That creates friction and inconsistency with other passkey-gated flows (e.g. change password).

This PR adds password-less reveal paths for non–social-login wallets with an enrolled passkey:

- **SRP reveal:** `revealSeedWordsWithPasskey` verifies the WebAuthn assertion, retrieves the passkey-wrapped vault key via `retrieveVaultKeyWithPasskey`, and calls `keyringController.exportSeedPhrase({ encryptionKey: vaultKey })`.
- **Private key reveal:** `exportAccountsWithPasskey` uses the same vault-key retrieval, then calls `keyringController.exportAccount({ encryptionKey: vaultKey }, address)` for each address in a multichain account group. The multichain private key list UI uses passkey verification as the initial credential step, with password fallback.
- **Shared component:** Extracts reusable `PasskeyVerification` (and `runPasskeyVerificationCeremony`) from change-password, reducing duplicated WebAuthn ceremony logic across settings and reveal flows.
- **Metrics:** Adds `SrpRevealWithPasskey` MetaMetrics events (started / completed / failed).

In both flows, the passkey assertion is not only a step-up UI gate — the passkey-derived vault key must successfully decrypt the vault via the keyring controller export APIs, cryptographically binding the passkey to this vault.

**Not supported:**
- Social-login (seedless) wallets — still password-only.
- Side panel when the enrolled passkey is incompatible with WebAuthn in the side panel — falls back to password (no full-tab handoff).

**Dependency:** Requires the breaking `@metamask/keyring-controller` change that accepts `VerificationCredentials` (`{ password }` | `{ encryptionKey }`) on `exportSeedPhrase` and `exportAccount` (mm-core PR for TO-737).

## **Changelog**

CHANGELOG entry: Added the ability to reveal your Secret Recovery Phrase and account private keys using passkey verification instead of your wallet password.

## **Related issues**

Fixes: TO-737

### SRP reveal
1. Set up a wallet with passkey enrolled (non–social-login).
2. Go to **Settings → Security & Privacy → Reveal Secret Recovery Phrase**.
3. Complete the SRP quiz (if shown).
4. Confirm the passkey verification step appears instead of the password prompt.
5. Complete the passkey ceremony and verify the SRP is revealed.
6. Click **Use password** and confirm the password flow still works.
7. Cancel or fail the passkey ceremony and confirm fallback to the password prompt.

### Private key reveal (multichain account group)
8. Open the multichain account group private key reveal flow.
9. Confirm passkey verification is offered when passkey is active.
10. Complete the passkey ceremony and verify private keys are shown for exportable addresses.
11. Confirm password fallback works on failure or when choosing **Use password**.

### General
12. Repeat in the side panel with a passkey incompatible with side-panel WebAuthn — confirm password fallback.
13. Confirm social-login wallets still use password-only reveal.
14. Run unit tests: `yarn test:unit ui/pages/keychains/reveal-seed.test.tsx ui/components/app/passkey-verification/passkey-verification.test.tsx ui/store/actions.test.js app/scripts/metamask-controller.actions.test.js`

## **Pre-merge author checklist**

- [ ] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Extension Coding Standards](https://github.com/MetaMask/metamask-extension/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

---

> [!NOTE]
> **High Risk**
> Breaking API on methods that expose seed phrases and private keys, with new encryption-key verification touching vault decryption—high impact for security-sensitive consumers.
> 
> **Overview**
> **Breaking:** `exportSeedPhrase` and `exportAccount` no longer accept a plain password string. Callers must pass a `Credentials` object: `{ password }` or `{ encryptionKey, encryptionSalt? }`.
> 
> Sensitive exports now go through shared `#verifyCredentials`, which either runs existing password decryption or a new `#verifyEncryptionKey` path (vault salt check, `importKey` + `decryptWithKey`). A mismatched optional `encryptionSalt` throws `ExpiredCredentials`. `exportAccount` also enforces the unlocked controller guard before verification.
> 
> The public `Credentials` type is exported from `types.ts`, messenger action JSDoc is updated, and tests cover password and encryption-key success/failure cases. The changelog records the breaking API change for passkey-style vault-key verification (e.g. SRP / private key reveal without re-entering the wallet password).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccbd63131a52e6ae4026eabbda6f08f44a940650. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-controller/src/KeyringController.test.ts
+++ b/packages/keyring-controller/src/KeyringController.test.ts
@@ -605,7 +605,7 @@ describe('KeyringController', () => {
           { type: 'HD Key Tree' },
           async ({ keyring }) => keyring.serialize(),
         );
-        const currentSeedWord = await controller.exportSeedPhrase(password);
+        const currentSeedWord = await controller.exportSeedPhrase({ password });
 
         await controller.createNewVaultAndRestore(password, currentSeedWord);
 
@@ -680,7 +680,7 @@ describe('KeyringController', () => {
             await controller.createNewVaultAndKeychain(password);
 
             const currentSeedPhrase =
-              await controller.exportSeedPhrase(password);
+              await controller.exportSeedPhrase({ password });
 
             expect(currentSeedPhrase.length).toBeGreaterThan(0);
             expect(
@@ -794,13 +794,13 @@ describe('KeyringController', () => {
     describe('when there is an existing vault', () => {
       it('should not create a new vault or keychain', async () => {
         await withController(async ({ controller, initialState }) => {
-          const initialSeedWord = await controller.exportSeedPhrase(password);
+          const initialSeedWord = await controller.exportSeedPhrase({ password });
           expect(initialSeedWord).toBeDefined();
           const initialVault = controller.state.vault;
 
           await controller.createNewVaultAndKeychain(password);
 
-          const currentSeedWord = await controller.exportSeedPhrase(password);
+          const currentSeedWord = await controller.exportSeedPhrase({ password });
           expect(initialState).toStrictEqual(controller.state);
           expect(initialSeedWord).toBe(currentSeedWord);
           expect(initialVault).toStrictEqual(controller.state.vault);
@@ -867,7 +867,7 @@ describe('KeyringController', () => {
 
           primaryKeyring.mnemonic = '';
 
-          await expect(controller.exportSeedPhrase(password)).rejects.toThrow(
+          await expect(controller.exportSeedPhrase({ password })).rejects.toThrow(
             "Can't get mnemonic bytes from keyring",
           );
         });
@@ -878,7 +878,7 @@ describe('KeyringController', () => {
       describe('when correct password is provided', () => {
         it('should export seed phrase without keyringId', async () => {
           await withController(async ({ controller }) => {
-            const seed = await controller.exportSeedPhrase(password);
+            const seed = await controller.exportSeedPhrase({ password });
             expect(seed).not.toBe('');
           });
         });
@@ -886,7 +886,7 @@ describe('KeyringController', () => {
         it('should export seed phrase with valid keyringId', async () => {
           await withController(async ({ controller, initialState }) => {
             const keyringId = initialState.keyrings[0].metadata.id;
-            const seed = await controller.exportSeedPhrase(password, keyringId);
+            const seed = await controller.exportSeedPhrase({ password }, keyringId);
             expect(seed).not.toBe('');
           });
         });
@@ -901,7 +901,7 @@ describe('KeyringController', () => {
         it('should throw error if keyringId is invalid', async () => {
           await withController(async ({ controller }) => {
             await expect(
-              controller.exportSeedPhrase(password, 'invalid-id'),
+              controller.exportSeedPhrase({ password }, 'invalid-id'),
             ).rejects.toThrow('Keyring not found');
           });
         });
@@ -913,7 +913,7 @@ describe('KeyringController', () => {
             jest
               .spyOn(encryptor, 'decrypt')
               .mockRejectedValueOnce(new Error('Invalid password'));
-            await expect(controller.exportSeedPhrase('')).rejects.toThrow(
+            await expect(controller.exportSeedPhrase({ password: '' })).rejects.toThrow(
               'Invalid password',
             );
           });
@@ -927,7 +927,7 @@ describe('KeyringController', () => {
                 .spyOn(encryptor, 'decrypt')
                 .mockRejectedValueOnce(new Error('Invalid password'));
               await expect(
-                controller.exportSeedPhrase('', keyringId),
+                controller.exportSeedPhrase({ password: '' }, keyringId),
               ).rejects.toThrow('Invalid password');
             },
           );
@@ -975,7 +975,7 @@ describe('KeyringController', () => {
       await withController(async ({ controller }) => {
         await controller.setLocked();
 
-        await expect(controller.exportSeedPhrase(password)).rejects.toThrow(
+        await expect(controller.exportSeedPhrase({ password })).rejects.toThrow(
           KeyringControllerErrorMessage.ControllerLocked,
         );
       });
@@ -990,7 +990,7 @@ describe('KeyringController', () => {
             await withController(async ({ controller, initialState }) => {
               const account = initialState.keyrings[0].accounts[0];
               const newPrivateKey = await controller.exportAccount(
-                password,
+                { password },
                 account,
               );
               expect(newPrivateKey).not.toBe('');
@@ -1002,7 +1002,7 @@ describe('KeyringController', () => {
           it('should throw error', async () => {
             await withController(async ({ controller }) => {
               await expect(
-                controller.exportAccount(password, ''),
+                controller.exportAccount({ password }, ''),
               ).rejects.toThrow(KeyringControllerErrorMessage.KeyringNotFound);
             });
           });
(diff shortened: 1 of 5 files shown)


tanguyenvn:
I'll need to support exporting private keys of imported accounts using passkeys

ccharly:
> I'll need to support exporting private keys of imported accounts using passkeys

I was about to review, but IMO we should apply the new logic with the credentials in `exportAccount` too anyway.

The idea being, we now have several "credential kind" to unlock sensitive operations in this controller, passkeys knows the "vault key" (that they use with `submitEncryptionKey`), so they should be able to use this encrytion key (credential) anywhere we need "strong permission" (`exportSeedPhrase` and `exportAccount`).

I would also add this support in this PR!

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1057,16 +1074,29 @@ export class KeyringController<
...
     this.#assertIsUnlocked();
-    await this.verifyPassword(password);
+
+    if (typeof credentials === 'string') {
+      await this.verifyPassword(credentials);
+    } else if (hasProperty(credentials, 'password')) {
+      await this.verifyPassword(credentials.password as string);
+    } else {
+      await this.verifyEncryptionKey(credentials.encryptionKey);
+    }
```
ccharly: Let's add a new `#verifyCredentials` method for this, so we can use it in `exportSeedPhrase` and in `exportAccount` too.

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1057,16 +1074,29 @@ export class KeyringController<
...
-    password: string,
+    credentials: string | { password: string } | { encryptionKey: string },
     keyringId?: string,
   ): Promise<Uint8Array> {
     this.#assertIsUnlocked();
-    await this.verifyPassword(password);
+
+    if (typeof credentials === 'string') {
+      await this.verifyPassword(credentials);
+    } else if (hasProperty(credentials, 'password')) {
```
ccharly: Let's use `in` here, that's the pattern we use in this controller already:
```suggestion
    } else if ('password' in credentials) {
```

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1057,16 +1074,29 @@ export class KeyringController<
...
+      await this.verifyPassword(credentials.password as string);
```
ccharly: We don't need to type-cast here (at least, not when using `if ('password' in credentials)`
```suggestion
      await this.verifyPassword(credentials.password);
```

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1045,6 +1046,41 @@ export class KeyringController<
     await this.#encryptor.decrypt(password, this.state.vault);
   }
 
+  /**
+   * Method to verify a given encryption key validity. Throws an error if the
+   * encryption key is invalid, i.e. it cannot decrypt the vault.
+   *
+   * @param encryptionKey - Serialized vault encryption key.
+   */
+  async verifyEncryptionKey(encryptionKey: string): Promise<void> {
```
ccharly: Maybe we don't need to export it publicly for now, unlike `verifyPassword` that could be used to write "end-user logic" (e.g. to test the password), we should not really need to call `verifyEncryptionKey` explicitely 🤔

We could make this private (e.g. `#verifyEncryptionKey`).

WDYT @danroc?

packages/keyring-controller/src/types.ts
```diff
@@ -72,3 +72,11 @@ export type SignTypedDataMessageV3V4 = {
 export type TypedMessageParams = {
   data: Record<string, unknown>[] | string | SignTypedDataMessageV3V4;
 } & AbstractMessageParams;
+
+/**
+ * Credentials for re-authenticating the keyring during sensitive operations
+ * such as `exportSeedPhrase` and `exportAccount`.
+ */
+export type VerificationCredentials =
```
ccharly: I would go for `Credentials` here!

packages/keyring-controller/src/KeyringController.test.ts
```diff
@@ -3616,6 +3706,44 @@ describe('KeyringController', () => {
     });
   });
 
+  describe('verifyEncryptionKey', () => {
```
ccharly: Since we want to make this private, we can still test those using any methods that uses `Credentials` as parameter, so that should still be fine, maybe we can rename that to:
```suggestion
  describe('#verifyEncryptionKey', () => {
```

tanguyenvn:
@metamaskbot publish-previews

packages/keyring-controller/src/types.ts
```diff
@@ -72,3 +72,9 @@ export type SignTypedDataMessageV3V4 = {
 export type TypedMessageParams = {
   data: Record<string, unknown>[] | string | SignTypedDataMessageV3V4;
 } & AbstractMessageParams;
+
+/**
+ * Credentials for re-authenticating the keyring during sensitive operations
+ * such as `exportSeedPhrase` and `exportAccount`.
+ */
+export type Credentials = { password: string } | { encryptionKey: string };
```
ccharly: We should have have an `encryptionSalt?` (optional yes), so the caller can decide to explicitly send the salt too. If it's not provided, then we would just fallback to the one being serialized alongside the vault (exactly in `#unlockKeyrings`).

Once you made that change, this will also allow us to use this type in `#unlockKeyrings` directly too!

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1045,6 +1046,56 @@ export class KeyringController<
...
+
+  /**
+   * Verifies export credentials by checking either the wallet password or the
+   * vault encryption key.
+   *
+   * @param credentials - Object holding either the `password` or the vault
+   * `encryptionKey`.
+   */
+  async #verifyCredentials(credentials: Credentials): Promise<void> {
+    // eslint-disable-next-line no-restricted-syntax
```
ccharly, line 1088: Nit: Do we really need this? I saw another `'password' in credentials` without this directive.

tanguyenvn: yes, it's needed if we want to use "in" operator

ccharly (approved):
LGTM! Thanks for addressing all this!

PR #9010, opened 2026-06-04
feat: AUS Perps Watchlist

gambinish:
## Explanation

Persists `state.watchlistMarkets` to AUS so the watchlist follows the user across devices and clients, rather than being device-local only.

What changed

- `toggleWatchlistMarket` now performs an optimistic local update first (UI is instant), then syncs to AUS. If the remote write fails, local state is reverted.
- On init(), the controller hydrates state.watchlistMarkets from AUS. AUS is the source of truth; local state acts as an offline cache (persist: true unchanged).
- One-time migration: if a user already has local markets saved but no AUS record yet, those are pushed up automatically on first init.
- Falls back to local-only silently when unauthenticated or when the active provider has no AUS mapping (e.g. aggregated).
- Adds resolveWatchlistExchangeKey() — the single extension point for adding future DEX providers to AUS sync.
Integration (client PRs)

Each client needs to grant the PerpsController messenger two new allowed actions:

- `AuthenticatedUserStorageService:getNotificationPreferences`
- `AuthenticatedUserStorageService:putNotificationPreferences`

No other wiring needed — `AuthenticatedUserStorageService` is already running wherever `NotificationServicesController` is wired, and `AuthenticationController:getBearerToken` was already allowed.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Cross-device user preference writes and a breaking `toggleWatchlistMarket` API change require correct messenger wiring; merge logic and queue ordering reduce race risk but remote/local divergence on auth failures is possible by design.
> 
> **Overview**
> **Perps watchlist favorites** are now synced to **Authenticated UserStorage** (`perps.watchlistMarkets` inside notification preferences) so they follow the signed-in user across devices, while local persisted state remains the offline cache.
> 
> `toggleWatchlistMarket` is now **`async` (`Promise<void>`)** and does an **optimistic** local update, then a **serialized** AUS read-merge-write via `#ausQueue`; failed remote writes **revert** local state. **Init** kicks off non-blocking `#syncWatchlistFromRemote` on the same queue so hydration finishes before toggles race it. Remote is source of truth when an exchange key exists (including **empty** lists); missing exchange keys trigger a **one-time migration** from local favorites. **`resolveWatchlistExchangeKey`** maps `hyperliquid`/`myx`; **`aggregated`** and missing blobs skip AUS without throwing. Favorite-toggle **analytics** are emitted on toggle.
> 
> Clients must allow messenger actions **`AuthenticatedUserStorageService:getNotificationPreferences`** and **`putNotificationPreferences`** (README dependency graph updated). Package adds **`@metamask/authenticated-user-storage`** as a dev dependency and extensive **AUS sync tests**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1d5db8f4b54cb4f9f708539e50562aaaf78e63a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/PerpsController.ts
```diff
@@ -5057,6 +5128,174 @@ export class PerpsController extends BaseController<
...
+      this.#debugLog(
+        'PerpsController: Skipping AUS watchlist sync — provider not mapped',
+        { activeProvider: this.state.activeProvider },
+      );
+      return;
+    }
+
+    const prefs = await this.messenger.call(
+      'AuthenticatedUserStorageService:getNotificationPreferences',
+    );
```
gambinish, line 5249: @aganglada One note on storage approach: why `NotificationPreferences` and not `AssetsWatchlistBlob` or some other purpose built storage blob like `PerpsWatchlistBlob`.

The AUS package actually has two separate storage endpoints:

1. `getAssetsWatchlist` / `setAssetsWatchlist` → `/assets-watchlist` — a dedicated, atomic watchlist blob ({ version: 1, assets: string[] })
2. `getNotificationPreferences` / `putNotificationPreferences` → `/preferences/notifications` — a combined blob covering wallet activity, marketing, perps, and social AI settings

At first glance `assets-watchlist` would be the cleaner fit — or potentially even a new purpose built storage endpoint `perps-watchlist`, but per the ticket spec we are following the existing schema. Using a separate endpoint would arguably be cleaner because we don't need to do this read-merge-write pattern, which runs the risk of accidentally clobbering our existing notification settings.

The tradeoff to be aware of with this approach: every watchlist toggle now does a GET → merge → PUT on the entire notification preferences blob. This is the same pattern `NotificationServicesController` uses for wallet account updates, so it's consistent with the existing codebase — but it does mean a concurrent notification preference change and a watchlist toggle could race. That's worth keeping in mind as the perps notification work continues.

Should we consider storing this under a dedicated endpoint in `@metamask/authenticated-user-storage` instead?

gambinish: If we do want to have this purpose built endpoint (I think we should), the changes needed would roughly follow the pattern of asset-watchlist:

1. New server endpoint in user storage API (likely a GET/PUT)
2. New methods and types on `@metamask/authenticated-user-storage`
3. Simplify this sync code in `perps-controller`, which would make things much cleaner

gambinish: It sounds like `getNotificationPreferences` is actually the correct code path here. The naming seemed confusing to me, as I wasn't sure if this should be considered related to "notifications". Will leave this as is.

gambinish:
@metamaskbot publish-previews

packages/perps-controller/src/PerpsController.ts
```diff
@@ -5113,6 +5210,181 @@ export class PerpsController extends BaseController<
...
+        return;
+      }
+
+      const remoteExchangeWatchlist =
+        prefs.perps.watchlistMarkets?.[exchangeKey];
+
+      // Only treat remote as the source of truth when it has actual symbols.
+      // An empty { testnet: [], mainnet: [] } blob (e.g. created by another
+      // device that had no watchlist) must not silently wipe local favorites.
+      const remoteHasContent =
```
geositta: Can we use field presence instead of symbol count to decide whether AUS has a watchlist for this exchange?

Today remoteHasContent is false when AUS has { testnet: [], mainnet: [] }. That makes an intentionally cleared remote watchlist look the same as "not migrated yet."

Example:

Device A clears the HyperLiquid watchlist, so AUS stores empty arrays.
Device B still has stale local favorites.
Device B initializes, treats the empty AUS watchlist as missing, and writes the stale local favorites back to AUS.

Since the PR defines AUS as the source of truth, should an existing exchange key hydrate local state even when both arrays are empty? Could we reserve migration for when prefs.perps.watchlistMarkets?.[exchangeKey] is absent?

geositta (changes requested):
Requesting changes. The AUS integration follows the expected messenger and read merge write pattern, and the queue addresses concurrent toggle ordering. I found one source of truth issue where an intentionally cleared remote watchlist can be overwritten by stale local cache during init. Changing the init check to use exchange key presence instead of symbol count should preserve remote clears and keep migration limited to users who have not been migrated yet.

packages/perps-controller/src/PerpsController.ts
gambinish: Yes, it should indeed hydrate local state as empty in this case. Resolved and added additional test here: https://github.com/MetaMask/core/pull/9010/commits/f79c1b672d10756208346fe3f4b47995528481e2

geositta (approved):
Approve. Thanks, the AUS watchlist sync now preserves remote source of truth, including empty remote watchlists.

PR #9016, opened 2026-06-05
feat(profile-metrics-service): add proof signing system

mathieuartu:
## References

Related to: https://consensyssoftware.atlassian.net/browse/MUL-1844

## Checklist

---

> [!NOTE]
> **High Risk**
> Introduces cryptographic signing paths (keyring personal_sign and internal snap RPC) that must preserve correct message canonicalization and snap trust boundaries; hosts must delegate KeyringController and SnapController actions on the service messenger.
> 
> **Overview**
> Adds **`ProofOfOwnershipService`** to `@metamask/profile-metrics-controller` so wallets can produce server-verifiable ownership proofs bound to auth API nonces, for use on `AccountWithScopes.proof` when submitting profile metrics.
> 
> **`ProofOfOwnershipService:sign({ account, nonce })`** picks the signer from the **first CAIP-2 scope**: `eip155` uses **`KeyringController:signPersonalMessage`** (EIP-191) on `metamask:proof-of-ownership:<nonce>:<canonical address>`; **Solana, Tron, and Bitcoin** use silent **`SnapController:handleRequest`** (`onClientRequest`, `signProofOfOwnership`) against `account.metadata.snap.id`. Unsupported or missing scopes throw **`ProofUnsupportedNamespaceError`**.
> 
> Package surface: new snaps + `uuid` deps, auto-generated messenger action types, **`proofOfOwnershipServiceName`** export, and **`profileMetricsServiceName`** alias. Import paths for profile-metrics method-action types are tightened to explicit files (no barrel). Broad unit tests cover EVM/snap routing, canonical addresses, and error paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26fc73dd85be02aa1aa7f4ace7396ada9866b2bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,284 @@
...
+ * @returns The namespace portion of the first scope.
+ * @throws {ProofUnsupportedNamespaceError} if the account has no scopes.
+ */
+function getAccountNamespace(account: InternalAccount): string {
+  const [firstScope] = account.scopes;
+  if (!firstScope) {
+    throw new ProofUnsupportedNamespaceError('<no scope>');
+  }
+  // `parseCaipChainId` validates the `<namespace>:<reference>` shape.
+  return parseCaipChainId(firstScope).namespace;
```
ccharly: The `parseCaipChainId` can also throw an `Error` if the CAIP chain ID is malformed. So if you wish to wrap all errors in `ProofUnsupportedNamespaceError` you would need to catch this and re-`throw` it maybe?

Or make the parsing yourself using regex from `@metamask/utils`!

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,284 @@
...
+          },
+        },
+      },
+    );
+
+    if (!SnapSignProofResponseStruct.is(response)) {
+      throw new Error(
+        `ProofOfOwnershipService: snap '${snapId}' returned a malformed response to '${SNAP_SIGN_PROOF_OF_OWNERSHIP_METHOD}'.`,
+      );
+    }
```
ccharly, line 290: Nit: We could use `superstruct.assert` and catch the message (which gives a pretty detailed message) if we wish to know what part was malformed.

I'm not entirely sure how worth that is given that the schema is pretty simple, so I don't expect this error from happening anyway 😅

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,284 @@
...
+    const canonicalAddress = canonicalizeAddress(account.address, namespace);
+    const message = buildProofMessage(nonce, canonicalAddress);
+
+    let signature: string;
+    if (namespace === KnownCaipNamespace.Eip155) {
+      signature = await this.#signEvm(account.address, message);
+    } else {
+      // canonicalizeAddress() already threw for any unsupported namespace,
+      // so anything reaching here is one of solana / tron / bip122.
+      signature = await this.#signViaSnap(account, message);
```
ccharly, line 220: Question: Are we expected to sign this proof before the onboarding actually finishes? All Snap requests are delayed until the onboarding is fully completed. (just trying to make sure we're not causing a "lock" here 😄

mathieuartu: No, the polling only starts on `KeyringController:unlock` 🙂

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
mathieuartu: [8bd1fdb](https://github.com/MetaMask/core/pull/9016/commits/8bd1fdb8754b29354ab88793cf3d3b09e8f0281d)

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
mathieuartu: [8bd1fdb](https://github.com/MetaMask/core/pull/9016/commits/8bd1fdb8754b29354ab88793cf3d3b09e8f0281d)

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
ccharly: Actually, I think I was wrong. The `signature` shouldn't be "that sensitive" but I still feel like we shouldn't report that after second thought.

Sorry for this bad suggestion 😅 Can you re-use your original implementation for this instead? (using `superstruct.is` was fine).

I might need to also re-use that pattern elsewhere in some `accounts` `assert`s we use! 🙊

ccharly (dismissed):

gantunesr (dismissed):

ccharly (approved):

tommasini (approved):

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,294 @@
...
+   */
+  async #signViaSnap(
+    account: InternalAccount,
+    message: string,
+  ): Promise<string> {
+    const snapId = account.metadata.snap?.id;
+    if (!snapId) {
+      throw new Error(
+        `ProofOfOwnershipService: account '${account.id}' has no snap to sign a proof of ownership.`,
+      );
```
NicolasMassart, line 262: The CHANGELOG says callers catch `ProofUnsupportedNamespaceError` to submit accounts without a proof. A non-EVM account that happens to have no `metadata.snap.id` falls into a plain `Error` here, which won't match that catch and will likely propagate as an unhandled failure in the polling pipeline.

Either throw `ProofUnsupportedNamespaceError` (the account is effectively unsignable from this service's perspective) or introduce a sibling typed error so the caller can pattern-match it.

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,294 @@
...
+    const message = buildProofMessage(nonce, canonicalAddress);
+
+    let signature: string;
+    if (namespace === KnownCaipNamespace.Eip155) {
+      signature = await this.#signEvm(account.address, message);
+    } else {
+      // canonicalizeAddress() already threw for any unsupported namespace,
+      // so anything reaching here is one of solana / tron / bip122.
+      signature = await this.#signViaSnap(account, message);
+    }
```
NicolasMassart, line 221: The EVM branch is gated on `KnownCaipNamespace.Eip155`; the `else` covers "anything `canonicalizeAddress` accepted." If a future namespace is added to `canonicalize.ts` (e.g. another EVM-like chain with its own signer), it will land in the snap branch by accident.

Suggest making the snap branch explicit:

```ts
if (namespace === KnownCaipNamespace.Eip155) { ... }
else if (
  namespace === KnownCaipNamespace.Solana ||
  namespace === KnownCaipNamespace.Tron ||
  namespace === KnownCaipNamespace.Bip122
) { ... }
else { throw new ProofUnsupportedNamespaceError(namespace); }
```

This also removes the implicit dependency on `canonicalizeAddress` throwing first.

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -0,0 +1,294 @@
...
+    if (!snapId) {
+      throw new Error(
+        `ProofOfOwnershipService: account '${account.id}' has no snap to sign a proof of ownership.`,
+      );
+    }
+
+    const response: unknown = await this.#messenger.call(
+      'SnapController:handleRequest',
+      {
+        snapId: snapId as SnapId,
```
NicolasMassart, line 268: `account.metadata.snap.id` is plain `string`; this cast asserts a brand we haven't validated. The snap controller will reject malformed IDs at runtime, but a typed parse (e.g. via snaps-utils) would make the boundary explicit and keep malformed values out of the messenger payload.

NicolasMassart (commented):
Nice work — solid test coverage and the silent-snap routing is well documented. A few inline notes below.

One additional item that doesn't have an obvious inline anchor: the service relies on the host wallet delegating both `KeyringController:signPersonalMessage` **and** `SnapController:handleRequest` onto the service messenger. Failing to delegate produces a runtime "Action handler not found" on the first sign attempt. Worth a short integration note in the package README (either in this PR or as a follow-up tracked on MUL-1844).

packages/profile-metrics-controller/src/ProofOfOwnershipService.test.ts
```diff
@@ -0,0 +1,447 @@
...
+        expect(request.request.jsonrpc).toBe('2.0');
+        expect(request.request.method).toBe(
+          SNAP_SIGN_PROOF_OF_OWNERSHIP_METHOD,
+        );
+        expect(request.request.params.accountId).toBe('snap-account-id');
+        // The message payload uses the canonical address — bech32 lowercased
+        // for BIP-122, base58 verbatim for Solana/Tron.
+        expect(request.request.params.message).toMatch(
+          /^metamask:proof-of-ownership:n42:/u,
+        );
```
NicolasMassart, line 189: The bech32 address is supplied uppercase but only the message *prefix* is asserted, so a regression that stops lowercasing `bc1…` would still pass this test. Since this is the only place that exercises the dispatch + canonicalization seam end-to-end, consider replacing the regex with an exact match against the full canonical message — cheap, and it locks in the bech32 lowercasing behavior.

PR #9026, opened 2026-06-05
ci: Add top-level permissions to `publish-release` workflow

Mrtenz:
## Explanation

By default, all jobs in a workflow inherit the permissions provided to it in the caller (`main.yml` in this case). This meant that `publish-release.yml` had `contents: write` and `id-token: write` in all jobs, including the dry run job. This made the dry run job think it was an actual run, and try to publish (which failed). By explicitly specifying permissions, jobs will inherit those permissions instead. We do this in the module template too.

## Checklist

---

> [!NOTE]
> **Low Risk**
> CI-only permission scoping for GitHub Actions; no application or runtime behavior changes.
> 
> **Overview**
> Adds **workflow-level `permissions: contents: read`** to the reusable `publish-release` workflow so jobs no longer inherit **`contents: write`** and **`id-token: write`** from the caller in `main.yml`.
> 
> Without this default, the **`publish-npm-dry-run`** job received publish-level permissions and the npm publish action treated the run as a real release instead of a dry run. Jobs that still need elevated access keep explicit overrides: **`publish-release`** sets `contents: write`, and **`publish-npm`** sets `contents: read` plus `id-token: write`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cbf2ba650f87f0d556ef915b515a0fd29761132. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -8,6 +8,9 @@ on:
       SLACK_WEBHOOK_URL:
         required: true
 
+permissions:
+  contents: read
+
 concurrency:
   group: publish-release
   queue: max


Gudahtt:
Hmm. This is a fairly subtle problem. We should have a more explicit way than this to enable dry runs.

Making the dry-run contingent on the presence of the npm token was already pretty bad IMO, this is worse.

Mrtenz:
> Hmm. This is a fairly subtle problem. We should have a more explicit way than this to enable dry runs.
> 
> Making the dry-run contingent on the presence of the npm token was already pretty bad IMO, this is worse.

Yes, I agree. In hindsight we should've added an explicit `dry-run` option. I don't know if we can do that now without making a breaking change.

mcmire (approved):
Good to know. Knowing when and where to specify permissions is always confusing to me :/

PR #9042, opened 2026-06-08
chore: extend contract utils exchange rates response type

sahar-fehri:
## Explanation

* Adds an optional includeMarketData parameter to fetchTokenContractExchangeRates in assetsUtil.ts

* When includeMarketData is true, returns ContractMarketData (full MarketDataDetails per token — price, percentage changes, market cap, volume, etc.) instead of ContractExchangeRates (price-only map)

* Defaults to false — existing callers are unaffected

## Checklist

---

> [!NOTE]
> **Low Risk**
> Backward-compatible optional parameter on a utility fetch helper; default path unchanged and changes are covered by a new test.
> 
> **Overview**
> **`fetchTokenContractExchangeRates`** gains an optional **`includeMarketData`** flag (default **`false`**) so callers can opt into richer responses without changing existing behavior.
> 
> When **`includeMarketData: true`**, the function returns **`ContractMarketData`**—a map of checksummed contract addresses to full **`MarketDataDetails`** (price, % changes, market cap, volume, etc.)—instead of **`ContractExchangeRates`**, which only maps addresses to numeric prices. TypeScript overloads tie the return type to the flag. A unit test covers the market-data path and address checksumming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1b10e143e322cfc8836fe4256a6474c7a13b2a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sahar-fehri:
@metamaskbot publish-preview

salimtb (dismissed):

Prithpal-Sooriya (dismissed):

sahar-fehri:
@metamaskbot publish-preview

packages/assets-controllers/src/assetsUtil.ts
```diff
@@ -400,6 +407,21 @@ export async function fetchTokenContractExchangeRates({
     initialResult: {},
   });
 
+  if (includeMarketData) {
```
juanmigdr, line 421: Would be nice to have a test case for this

packages/assets-controllers/src/assetsUtil.ts
```diff
@@ -350,19 +353,23 @@ export async function reduceInBatchesSerially<Value, Result>({
...
   chainId,
+  includeMarketData = false,
 }: {
   tokenPricesService: AbstractTokenPricesService;
   nativeCurrency: string;
   tokenAddresses: Hex[];
   chainId: Hex;
-}): Promise<ContractExchangeRates> {
+  includeMarketData?: boolean;
+}): Promise<ContractExchangeRates | ContractMarketData> {
```
juanmigdr, line 383: The return type Promise<ContractExchangeRates | ContractMarketData> does not depend on  includeMarketData. Callers who pass includeMarketData: true still get a union, which forces unsafe casts or manual narrowing.

Could we use something like this?
```typescript
export async function fetchTokenContractExchangeRates(
  args: { ...; includeMarketData?: false }
): Promise<ContractExchangeRates>;
export async function fetchTokenContractExchangeRates(
  args: { ...; includeMarketData: true }
): Promise<ContractMarketData>;
```

packages/assets-controllers/src/assetsUtil.ts
```diff
@@ -400,6 +407,21 @@ export async function fetchTokenContractExchangeRates({
     initialResult: {},
   });
 
+  if (includeMarketData) {
+    return Object.entries(tokenPricesByTokenAddress).reduce<ContractMarketData>(
+      (obj, [tokenAddress, tokenPrice]) => {
+        if (tokenPrice) {
+          return {
+            ...obj,
+            [toChecksumHexAddress(tokenAddress)]: tokenPrice,
```
juanmigdr: This is an issue spotted by cursor, not sure how important it is but worth raising:
```
Map keys are checksummed via toChecksumHexAddress(tokenAddress), but the embedded tokenPrice.tokenAddress inside each MarketDataDetails value is passed through as returned by the prices service (may be lowercase).

Consumers comparing result[address] with value.tokenAddress could hit casing mismatches. Consider normalizing tokenAddress in the returned object to match the checksummed key (similar to how keys are normalized on the price-only path).
```

packages/assets-controllers/src/assetsUtil.ts
sahar-fehri: [8929278](https://github.com/MetaMask/core/pull/9042/commits/8929278cb9a0b36402c41744e7570fa103bfa0ab)

Prithpal-Sooriya (dismissed):

packages/assets-controllers/src/assetsUtil.ts
sahar-fehri: Good point, i think its good to normalize addresses, although mobile does not use it directly but good to have it normalized

sahar-fehri: [f1b10e1](https://github.com/MetaMask/core/pull/9042/commits/f1b10e143e322cfc8836fe4256a6474c7a13b2a5)

packages/assets-controllers/src/assetsUtil.ts
sahar-fehri: [f1b10e1](https://github.com/MetaMask/core/pull/9042/commits/f1b10e143e322cfc8836fe4256a6474c7a13b2a5)

juanmigdr (approved):
LGTM

PR #9047, opened 2026-06-08
chore: replace depcheck with knip

cryptodev-2s:
## Explanation

This repo used [`depcheck`](https://github.com/depcheck/depcheck) to lint for unused dependencies, configured via `.depcheckrc.yml`. Depcheck is largely unmaintained, while [`knip`](https://knip.dev/) is actively developed and produces clearer reports.

This PR replaces depcheck with knip as a near drop-in swap:

- New `knip.config.ts` ports `.depcheckrc.yml`'s ignore list to `ignoreDependencies` (using regex for glob patterns like `@types/*`) and configures the root workspace's `entry`/`project` patterns so knip scans `tests/`, `scripts/`, and root config files the way depcheck implicitly did.
- Invocations use the `--dependencies` shortcut (`dependencies,unlisted,binaries,unresolved,catalog`) so only depcheck-equivalent issue types are reported — files and exports are intentionally not checked, to keep the swap mechanical.
- Root `lint:dependencies` is scoped to `--workspace .` to match depcheck's original root-only coverage.
- The 4 per-package `lint:dependencies` scripts (json-rpc-engine, eth-json-rpc-provider, eth-block-tracker, eth-json-rpc-middleware) are preserved as `knip --workspace <pkg-name> --dependencies`.
- `knip` is pinned to `^6.15.0` — the latest version past this repo's 3-day `npmMinimalAgeGate` in `.yarnrc.yml`.

## References

N/A

## Checklist

---

> [!NOTE]
> **Low Risk**
> Dev-tooling and lint script changes only; no runtime package code. Dependency lint coverage shifts to root-only knip, which may behave slightly differently than the old per-package depcheck runs.
> 
> **Overview**
> Replaces **depcheck** with **knip** for unused-dependency linting at the monorepo root.
> 
> **`.depcheckrc.yml` is removed** and **`knip.config.ts` is added**, carrying over the old ignore rules as `ignoreDependencies` (with regex where depcheck used globs) and defining root `entry` / `project` globs so scripts, tests, and config files are treated like entry points. Root `lint:dependencies` now runs `knip --workspace . --dependencies` (still paired with `yarn dedupe --check`); `depcheck` is dropped from devDependencies and **`knip` ^6.15.0** is added.
> 
> **Four workspace packages** (`json-rpc-engine`, `eth-json-rpc-provider`, `eth-block-tracker`, `eth-json-rpc-middleware`) no longer define **`lint:dependencies` / per-package depcheck** in their `lint` and `lint:fix` scripts—dependency checks are centralized on the root workspace only.
> 
> The **package migration guide** is updated so migrated packages should remove **`knip.config.ts`** (replacing the old `.depcheckrc` reference) when stripping per-repo tooling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc90a6ba6683eafbd8089c33c4101f52a6a381af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/eth-block-tracker/package.json
```diff
@@ -46,10 +46,9 @@
     "build:docs": "typedoc",
     "changelog:update": "../../scripts/update-changelog.sh @metamask/eth-block-tracker",
     "changelog:validate": "../../scripts/validate-changelog.sh @metamask/eth-block-tracker",
-    "lint": "yarn lint:eslint && yarn lint:misc --check && yarn lint:dependencies",
-    "lint:dependencies": "depcheck",
```
cryptodev-2s: these where leftovers when we ported the packages into core

cryptodev-2s:
moving to draft until we merge first https://github.com/MetaMask/core/pull/9048

knip.config.ts
```diff
@@ -0,0 +1,58 @@
...
+        'tests/**/*.ts',
+        '*.config.{js,cjs,mjs,ts}',
+        'jest.config.*.js',
+        'yarn.config.cjs',
+        'eslint.config.mjs',
+        '.prettierrc.js',
+        'release.config.js',
+        'knip.config.ts',
+      ],
+      project: ['scripts/**/*.ts', 'tests/**/*.ts', '*.{js,cjs,mjs,ts}'],
```
Mrtenz, line 17: This sounds like it only works on a limited set of paths, instead of the entire repository?

cryptodev-2s: Yeah, intentional for now: `--workspace .` in the lint script scopes this to the root workspace. depcheck wasn't actually checking `packages/*` either (it only validated root deps), so coverage's the same as before. Happy to expand knip to lint all workspaces in a follow-up PR.
Check the deleted `.depcheckrc.yml`

Mrtenz: I don't think that's intentional 🤔 We should definitely be checking all packages.

Gudahtt: I agree it's a mistake, but it's a pre-existing one. Solving it in a later PR sounds good to me, but we really should fix it.

I had discovered this independently a few of weeks ago but forgot to mention it

Gudahtt (approved):
LGTM!

PR #9049, opened 2026-06-08
chore: expand knip to lint all workspaces

cryptodev-2s:
## Explanation

Until now the root \`lint:dependencies\` script ran \`knip --workspace .\`, scoping checks to the root workspace only — the same blind spot depcheck had before #9047. The ~85 packages under \`packages/*\` were never dep-linted.

This PR drops \`--workspace .\` so knip lints every workspace. The diff is almost entirely in \`knip.config.ts\`, where I've categorized the resulting findings into three groups:

1. **Workspace-level configuration** (real false positives that need explicit handling):
   - \`packages/foundryup\`: \`anvil\` / \`sysctl\` are external system binaries, not npm packages; \`yargs-parser\` is used indirectly.
   - \`packages/perps-controller\`: tests mock a \`core/Engine\` path that only exists in the mobile client repo.

2. **Global \`ignoreDependencies\`** (tooling blind spots that apply across the monorepo):
   - CLI tools invoked through shell wrappers (\`auto-changelog\` via \`scripts/update-changelog.sh\`).
   - Jest test environments referenced as strings (\`jest-environment-jsdom\`, etc.).
   - \`typedoc-plugin-missing-exports\`, \`@metamask/providers\`, \`webextension-polyfill\` declared in most packages but only used in one or two.

3. **Per-workspace \`ignoreDependencies\` snapshots** (~40 packages — real findings deferred for follow-up):
   - Missing direct deps relying on transitive resolution (\`@metamask/keyring-internal-api\`, \`cockatiel\`, etc.).
   - Dead deps that look unused at this scope (\`lodash\`, \`nock\`, \`immer\` etc. in specific packages).
   - The block has a header comment explaining that each entry is a real signal owning teams should triage and remove.

The PR keeps CI green today and turns knip into a forward-looking guardrail: any new unused/unlisted dep in any package will now be caught.

## References

Builds on #9047 (the depcheck → knip swap). See [this discussion comment](https://github.com/MetaMask/core/pull/9047#discussion_r3375611552) for the conversation that motivated this PR.

## Checklist

---

> [!NOTE]
> **Low Risk**
> CI/tooling-only changes to knip config and lint scripts; no runtime application code.
> 
> **Overview**
> **Dependency linting now covers every workspace**, not only the monorepo root. `lint:dependencies` / `lint:dependencies:fix` call `knip --dependencies` without `--workspace .`.
> 
> **`knip.config.ts` is updated** so CI stays green while the broader scope is enforced. The root workspace entry list drops a few self-referential config paths. **Global** `ignoreDependencies` gains documented tooling blind spots (e.g. `typedoc-plugin-missing-exports`, `@metamask/providers`, `webextension-polyfill`) and clearer comments. **Per-package** blocks record current knip findings (~40 packages) as `ignoreDependencies` snapshots for teams to triage later. **Special cases**: `foundryup` ignores system binaries `anvil`/`sysctl`; `perps-controller` ignores an unresolved mobile-only `core/Engine` mock path.
> 
> New unused or mis-declared deps in any package should now surface on the next lint run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e44892ff396fbd637ed10bf42e1d5921f25f5b9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


knip.config.ts
```diff
@@ -21,37 +16,224 @@ const config: KnipConfig = {
...
+      ignoreDependencies: ['@metamask/rpc-errors'],
+    },
+    'packages/ens-controller': {
+      ignoreDependencies: ['punycode'],
+    },
+    'packages/eth-json-rpc-middleware': {
+      ignoreDependencies: ['@metamask/network-controller', 'pify', 'tsd'],
+    },
+    'packages/eth-json-rpc-provider': {
+      ignoreDependencies: ['jest-it-up'],
```
Mrtenz, line 88: You can remove `jest-it-up` from these packages. I think they're leftovers from the migration. Fine to do in a follow-up too!

knip.config.ts
```diff
@@ -21,37 +16,224 @@ const config: KnipConfig = {
       // listed in `devDependencies`.
       ignoreUnresolved: ['ts-jest/jest-preset'],
     },
+    'packages/perps-controller': {
```
Mrtenz, line 19: Are you planning to ask teams to take a look at these?

Mrtenz (approved):

knip.config.ts
```diff
@@ -21,37 +16,224 @@ const config: KnipConfig = {
...
+      // The mobile client provides `core/Engine`; tests mock it via a
+      // relative path that doesn't resolve inside this monorepo.
+      ignoreUnresolved: [/^\.\.\/\.\.\/\.\.\/core\/Engine$/u],
```
cryptodev-2s, line 22: The next follow-up would be to enforce type checking in test files as well maybe having a dedicated tsconfig for tests.... I tried to enable it, but ran into several issues. That said, I think it’s something we should revisit soon.

knip.config.ts
cryptodev-2s: Yes I am planning to do it

PR #9057, opened 2026-06-09
perf: Faster ABI encoding for `MulticallClient`

FrederikBolding:
## Explanation

Improve `MulticallClient` performance by speeding up ABI encoding by introducing `encodeFunctionData` which skips calculating checksums.

```
encodeFunctionData — balanceOf(address) (100,000 iterations)
  controller-utils                 607.31 ms         164,661 ops/s
  ethers                          1160.67 ms          86,157 ops/s
  → controller-utils is 1.91x faster

encodeFunctionData — aggregate3(tuple[]) (100,000 iterations)
  controller-utils               54318.63 ms           1,841 ops/s
  ethers                         98961.99 ms           1,010 ops/s
  → controller-utils is 1.82x faster
```

## References

https://consensyssoftware.atlassian.net/browse/WPC-1063

## Checklist

---

> [!NOTE]
> **Low Risk**
> Encoding output is validated against ethers in tests; risk is mainly assuming inputs are valid addresses without checksum enforcement at encode time.
> 
> **Overview**
> Adds **`encodeFunctionData`** to `@metamask/controller-utils`: ABI call encoding via a custom `FastAbiCoder` that avoids ethers’ address checksumming on encode (documented tradeoff: callers must supply valid addresses). The package now depends on `@ethersproject/abi` and includes unit tests plus export snapshot updates.
> 
> **`MulticallClient`** in assets-controller routes `balanceOf`, `getEthBalance`, and **`aggregate3`** encoding through the new helper instead of `Interface.encodeFunctionData`, with a test asserting **byte-identical** calldata vs the standard ethers path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c24588709b53d17134fe27c59016f496c6eae1c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
+++ b/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
@@ -1,7 +1,6 @@
 import { Interface } from '@ethersproject/abi';
 import {
   createServicePolicy,
-  decodeFunctionResult,
   encodeFunctionData,
 } from '@metamask/controller-utils';
 import type { Hex } from '@metamask/utils';
@@ -433,7 +432,7 @@ export function decodeAggregate3Response(
   data: Hex,
   callCount: number,
 ): { success: boolean; returnData: Hex }[] {
-  const decoded = decodeFunctionResult(multicall3Interface, 'aggregate3', data);
+  const decoded = multicall3Interface.decodeFunctionResult('aggregate3', data);
 
   // decoded[0] is the array of (success, returnData) tuples
   const results = decoded[0] as readonly {
@@ -462,7 +461,7 @@ function decodeUint256(data: Hex): string {
     // Empty or invalid data; treat as 0
     return '0';
   }
-  const decoded = decodeFunctionResult(erc20Interface, 'balanceOf', data);
+  const decoded = erc20Interface.decodeFunctionResult('balanceOf', data);
   return decoded[0].toString();
 }
 
--- a/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
+++ b/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
@@ -1,6 +1,7 @@
 import { Interface } from '@ethersproject/abi';
 import {
   createServicePolicy,
+  decodeFunctionResult,
   encodeFunctionData,
 } from '@metamask/controller-utils';
 import type { Hex } from '@metamask/utils';
@@ -432,7 +433,7 @@ export function decodeAggregate3Response(
   data: Hex,
   callCount: number,
 ): { success: boolean; returnData: Hex }[] {
-  const decoded = multicall3Interface.decodeFunctionResult('aggregate3', data);
+  const decoded = decodeFunctionResult(multicall3Interface, 'aggregate3', data);
 
   // decoded[0] is the array of (success, returnData) tuples
   const results = decoded[0] as readonly {
@@ -461,7 +462,7 @@ function decodeUint256(data: Hex): string {
     // Empty or invalid data; treat as 0
     return '0';
   }
-  const decoded = erc20Interface.decodeFunctionResult('balanceOf', data);
+  const decoded = decodeFunctionResult(erc20Interface, 'balanceOf', data);
   return decoded[0].toString();
 }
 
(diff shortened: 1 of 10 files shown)


packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -377,16 +384,71 @@ const MULTICALL3_ADDRESS_BY_CHAIN: Record<Hex, Hex> = {
...
+      // Casting since we are missing internal unused methods, such _throwError.
+      return new FastAddressCoder(param.name) as unknown as Coder;
+    }
+    return super._getCoder(param);
+  }
+}
+
+const fastAbiCoder = new FastAbiCoder();
+
+function encodeFunctionData(
```
FrederikBolding: This should probably live somewhere else to be re-usable.

packages/controller-utils/src/abi.ts
```diff
@@ -0,0 +1,76 @@
...
+} from '@metamask/utils';
+import type { Hex } from '@metamask/utils';
+
+// Ethers does not export these unfortunately.
+type Coder = ReturnType<AbiCoder['_getCoder']>;
+type Writer = Parameters<Coder['encode']>[0];
+type Reader = Parameters<Coder['decode']>[0];
+
+// FastAddressCoder that skips checksumming addresses when encoding and uses the memoized `getChecksumAddress` for decoding.
+class FastAddressCoder {
```
Mrtenz: Should this extend `Coder`?

packages/controller-utils/src/abi.ts
```diff
@@ -0,0 +1,76 @@
...
+    return super._getCoder(param);
+  }
+}
+
+const fastAbiCoder = new FastAbiCoder();
+
+/**
+ * Encode the data required for a function call.
+ *
+ * Note: This uses `@ethersproject/abi` under the hood, but with some performance optimizations.
```
Mrtenz: Maybe good to note the performance optimisations here to make it clear when we should prefer this over the regular function? Or do you think we should always use this?

packages/controller-utils/src/abi.ts
FrederikBolding: Coder is not exported unfortunately, so we make due with this 😅

Mrtenz: We could at least add `implements Coder` since it's defined at the top right? Or does that introduce too much boilerplate to satisfy the `Coder` requirements?

packages/controller-utils/src/abi.ts
FrederikBolding: How is this?
```suggestion
 * Note: This uses `@ethersproject/abi` under the hood, but does not verify checksums for `address` types, ensure that the addresses passed in are valid.
```

FrederikBolding: I would say in most cases we can use this, except with untrusted input

packages/controller-utils/src/abi.ts
FrederikBolding: Then we would be required to implement `_throwError` too which is not used publicly

packages/controller-utils/src/abi.ts
Mrtenz: Something like this maybe?
```suggestion
 * Note: This uses `@ethersproject/abi` under the hood, but for improved
 * performance, does not verify checksums of addresses. Make sure addresses
 * passed to this function are valid and checksummed if necessary.
```

packages/controller-utils/src/abi.ts
Mrtenz:
```
_throwError() {
  throw new Error('Not implemented.');
}
```

It does what the function name describes. 😅

FrederikBolding: https://github.com/MetaMask/core/pull/9057/commits/65f015af9a615ccfe1c5861d0dfd4a655dfa09a3

packages/controller-utils/src/abi.ts
```diff
@@ -0,0 +1,81 @@
...
+  dynamic = false;
+
+  localName: string;
+
+  constructor(localName: string) {
+    this.localName = localName;
+  }
+
+  encode(writer: Writer, value: string): number {
+    return writer.writeValue(value);
```
Mrtenz, line 31: Should we validate that the value has the expected length at least?

May also need a `padStart` here, not sure if Ethers handles that in `writeValue`.

FrederikBolding: Ethers would just throw if an invalid address is encountered here

Mrtenz (dismissed):

packages/controller-utils/src/abi.ts
FrederikBolding: We could `assert(isHexChecksumAddress(value))`, but if we assume we trust the input it may not be worth it.

Mrtenz (approved):

Prithpal-Sooriya (approved):

PR #9059, opened 2026-06-09
Fix threshold flag fallback without MetaMetrics ID

pedronfigueiredo:
## What changed

Remote feature flag threshold arrays now fall back to the first valid threshold entry when there is no MetaMetrics ID available. This keeps threshold flags in their processed shape instead of exposing the raw array to consumers.

The change preserves the existing behavior for non-threshold arrays, and it supports both legacy threshold wrappers and `ThresholdVersion.DirectValue` entries.

## Why

Users who have not opted into MetaMetrics do not have a MetaMetrics ID, so the controller could not calculate a threshold bucket. The previous behavior passed the raw threshold array through, which caused consumers such as `confirmations_pay_extended` to miss nested values like `payStrategies.relay.gaslessEnabled`.

### DIFF
--- a/packages/remote-feature-flag-controller/CHANGELOG.md
+++ b/packages/remote-feature-flag-controller/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


pedronfigueiredo:
Closing this because both clients always pass an analytics ID, even when the user opts out of metrics collection

PR #8888, opened 2026-05-22
feat: add bearer token authorization to RampsService payments & quotes

meltingice1337:
## Explanation

The ramps API is gating more endpoints behind bearer-token authentication. PR #8843 introduced the auth plumbing on `RampsService` (`#getRequestHeaders`, messenger delegation of `AuthenticationController:getBearerToken`, runtime dependency on `@metamask/profile-sync-controller`) and applied it to `getBuyWidgetUrl`. Two more endpoints — `getPaymentMethods` and `getQuotes` — now also require auth upstream, so without a bearer token these calls will start failing and break the Buy flow.

This PR extends the existing auth pattern to those two endpoints:

- `getPaymentMethods` awaits `#getRequestHeaders()` once before dispatching the request through `#policy.execute`, and passes the resulting `Authorization: Bearer <token>` header to `#fetch`.
- `getQuotes` does the same.
- Token retrieval happens outside the retry/circuit-breaker loop (same shape as `getBuyWidgetUrl`), so a single bearer token is fetched per call and reused across retries. If `getBearerToken` rejects (wallet locked, signed out), no HTTP request is made.

No new dependencies, no messenger contract change — the `AuthenticationController:getBearerToken` action is already delegated by consumers as of #8843.

**Test updates:**

For each of `getPaymentMethods` and `getQuotes`, three new tests mirror the coverage added for `getBuyWidgetUrl` in #8843:

- `sends an Authorization header containing the bearer token` — nock `reqheaders` enforces the header is present on the outgoing request.
- `requests a bearer token exactly once per call` — verifies the token is fetched exactly once per call (i.e. outside the retry loop).
- `rejects without making an HTTP call when the bearer token cannot be retrieved` — verifies the request short-circuits when `AuthenticationController:getBearerToken` rejects.

## References

- Builds on #8843 (which introduced bearer-token auth for `getBuyWidgetUrl` and the supporting `#getRequestHeaders` / messenger delegation).

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

[...]

PR #9060, opened 2026-06-09
fix: remove Authorization header for unauthenticated payment methods …

meltingice1337:
## Explanation

`RampsService.getPaymentMethods` was recently changed (in #8888, still unreleased) to send an `Authorization: Bearer <token>` header, sourcing the token from `AuthenticationController:getBearerToken`. In practice, the payment-methods was changed to be called before the user signed into metamask.

Requiring a bearer token there is problematic because `AuthenticationController.getBearerToken` is not a passive read:

- If the wallet is **locked**, it throws (`getBearerToken - unable to proceed, wallet is locked`).
- If the wallet is **unlocked but not signed in**, it falls through to `getAccessToken`, which performs a full SRP **sign-in** (network round-trip + SIWE/SRP signing) as a side effect.

So fetching a payment-methods catalog could either fail outright or silently sign the user in — neither of which should happen just to render a list of payment options.

This PR scopes bearer-token auth back to the endpoints that genuinely need it. `getPaymentMethods` is now fetched as an unauthenticated request (matching the existing pattern used by `getGeolocation`, `getProviders`, etc.), while `getQuotes` and `getBuyWidgetUrl` keep their `Authorization` headers since they create user-bound orders/widgets.

Changes:

- **`RampsService.getPaymentMethods`** — removed the `#getRequestHeaders()` call and now passes no headers to `fetch`. The private `#getRequestHeaders` helper is unchanged and still used by `getQuotes` and `getBuyWidgetUrl`.
- **Tests** — replaced the three payment-methods auth tests with negative assertions (`does not request a bearer token`, `does not send an Authorization header`) mirroring the existing `getGeolocation` auth-scope tests. The "rejects when the bearer token cannot be retrieved" test was removed as no longer applicable.
- **Changelog** — kept the original #8888 entry intact and added a separate `Changed` entry documenting that `getPaymentMethods` no longer sends an `Authorization` header.

## References

- Related to #8888 (added the bearer-token auth being scoped down here)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes when ramps auth runs and could fail if the API still required a token for payment methods; intended to stop locked-wallet failures and accidental sign-in from catalog loads.
> 
> **Overview**
> Reverts bearer-token auth on **`RampsService.getPaymentMethods`** so the regions/payments catalog can load **before** sign-in. The method no longer calls `#getRequestHeaders()` or sends an `Authorization` header on `fetch`, aligning it with other public catalog calls (e.g. geolocation/providers). **`getQuotes`** and **`getBuyWidgetUrl`** still use authenticated headers.
> 
> Tests now assert payment-methods requests **do not** call `getBearerToken` and **do not** send `Authorization`, and drop the case where a missing token blocked the HTTP call. The ramps-controller changelog records this under **Changed**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 082a6c3ba347655e73bd94f15990d4369a20ca8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amitabh94:
Should we move the payment call after sign in ?

meltingice1337:
> Should we move the payment call after sign in ?

I would assume there is a reason the payments call is made at metamask open. Since it is not part of the abuse flow and doesn't return anything relevant, we can leave it unauthenticated.

amitabh94 (approved):

PR #9065, opened 2026-06-09
feat: add `@metamask/wallet-cli` package scaffold

sirtimid:
## Explanation

This adds `@metamask/wallet-cli` — an [oclif](https://oclif.io)-based `mm` CLI for `@metamask/wallet`, intended to run the wallet as a background daemon and dispatch messenger actions to it.

This first PR is the **package scaffold only**: a buildable, lintable, and testable package with the `mm` binary launcher (`bin/run.mjs`) and a couple of pure daemon path helpers, but no subcommands yet. oclif renders `mm --help` / `mm --version` even with no commands registered, so the package is exercisable end to end. Functionality is added incrementally in follow-up PRs (SQLite persistence, the Unix-socket daemon transport, the `Wallet` factory, and the `daemon` subcommands) so each stays small and self-contained.

Because the package is a CLI tool rather than a published library, it is registered in the `TOOLS` list in `yarn.config.cjs` — which exempts it from the library `exports` / `main` / `types` and `build:docs` constraints — matching how `@metamask/messenger-cli` is handled. The remaining changes outside `packages/wallet-cli/` are the standard new-package registration: ESLint overrides, root `tsconfig` project references, `CODEOWNERS`, `teams.json`, and the generated root README dependency graph.

## References

N/A

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate (the `getDaemonPaths` helper is covered to 100%)
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> New CLI package and lint/yarn config only; no wallet runtime, daemon, or messenger integration in this diff.
> 
> **Overview**
> Introduces **`@metamask/wallet-cli`**, an oclif-based **`mm`** CLI aimed at running `@metamask/wallet` as a background daemon. This PR is **scaffold only**: `bin/run.mjs` / `run.cmd`, oclif metadata pointing at `./dist/commands`, and **`@oclif/core`** as a runtime dependency—no subcommands yet (help/version still work via oclif).
> 
[...]

PR #9067, opened 2026-06-09
feat(wallet-cli): add SQLite-backed controller-state persistence

### LINKED ISSUE #8682: wallet-cli: Move sqlite persistence from `wallet` to `wallet-cli`
Ref: #8446 

We introduced a `better-sqlite3` persistence mechanism to the `wallet` package in #8417. Since that package is intended to be platform-agnostic, the entire `packages/wallet/src/persistence` directory should instead be migrated to `packages/wallet-cli`. There, the persistence mechanism should be integrated with the wallet initialization and tested end-to-end.

sirtimid:
## Explanation

`@metamask/wallet-cli` runs a background daemon that hosts a `@metamask/wallet` instance. For controller state to survive daemon restarts, it needs durable local storage. This PR adds that persistence layer. It is purely additive and not yet wired into a daemon process — the daemon entry point that consumes it lands in a follow-up.

- **`KeyValueStore`** — a synchronous `better-sqlite3`-backed key-value store (a single `kv` table with a TEXT key and a JSON-serialized TEXT value). Corrupt rows fail loud (a descriptive, key-scoped error) rather than returning garbage.
- **`loadState(store)`** — rehydrates persisted state into the `{ [controllerName]: { [property]: value } }` shape accepted by the `Wallet` constructor's `state` option.
- **`subscribeToChanges(messenger, controllerMetadata, store, log?)`** — subscribes to every `<Controller>:stateChanged` event and writes persist-flagged top-level properties through to the store (applying `StateDeriver` persist functions, and deleting a key when its property is removed from state). It returns an unsubscribe handle; teardown is owned by the caller (a `dispose()` handle in a later PR), so there is no package-level teardown event.

### `better-sqlite3` native addon

`better-sqlite3` ships a native C addon. The monorepo runs Yarn with `enableScripts: false`, so the addon is not built during `yarn install`. To handle this:

- a `test:prepare` step (`scripts/install-binaries.sh`) fetches a matching prebuild via `prebuild-install` before tests run;
- `engines.node` is set to `>=20` (better-sqlite3 v12 only ships prebuilt binaries for Node 20+);
- `@metamask/wallet-cli` is excluded from the Node 18.x CI test matrix;
- `yarn.config.cjs` exempts the package from the standard `test`-script and `engines.node` constraints accordingly.

## References

- Closes #8682
- Builds on #9065 (the `@metamask/wallet-cli` package scaffold). This PR targets that branch and should merge after it.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> New durable state path can diverge from in-memory state on disk write failures (logged only today); native addon and Node 20+ requirement affect install and CI.
> 
> **Overview**
> Adds **SQLite-backed persistence** in `@metamask/wallet-cli` so wallet controller state can survive restarts (not yet wired into the daemon).
> 
> A synchronous **`KeyValueStore`** (`better-sqlite3`, JSON values in a `kv` table) is the storage backend. **`loadState`** rebuilds `{ controllerName: { property: value } }` for `Wallet` construction, honoring only properties still marked `persist` in controller metadata so stale rows are not resurrected. **`subscribeToChanges`** listens on `<Controller>:stateChanged`, writes persist-flagged top-level fields from Immer patches (including `StateDeriver` transforms and deletes when properties are removed), logs write failures without stopping the process, and returns an unsubscribe handle.
> 
> **`better-sqlite3`** is added with **`engines.node` `>=20`**, a **`test:prepare`** / **`install-binaries.sh`** prebuild step (monorepo `enableScripts: false`), CI matrix exclusion for Node 18 on this package, and **`yarn.config.cjs`** exceptions for the custom test script and engines field. Dependencies on `@metamask/wallet`, `@metamask/base-controller`, and `immer` are added with matching tsconfig project references.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9854de90c8b0be33482de42b61b0a7377e42ae2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -143,11 +143,6 @@ jobs:
       matrix:
         node-version: [18.x, 20.x, 22.x]
         package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
-        exclude:
-          # @metamask/wallet-cli depends on better-sqlite3, which only ships
-          # prebuilt binaries for Node 20+.
-          - node-version: 18.x
-            package-name: '@metamask/wallet-cli'
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v2
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -143,6 +143,11 @@ jobs:
       matrix:
         node-version: [18.x, 20.x, 22.x]
         package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
+        exclude:
+          # @metamask/wallet-cli depends on better-sqlite3, which only ships
+          # prebuilt binaries for Node 20+.
+          - node-version: 18.x
+            package-name: '@metamask/wallet-cli'
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v2
--- a/packages/wallet-cli/scripts/install-binaries.sh
+++ b/packages/wallet-cli/scripts/install-binaries.sh
@@ -20,7 +20,9 @@ BETTER_SQLITE3_DIR="${MONOREPO_ROOT}/node_modules/better-sqlite3"
 if [ ! -f "${BETTER_SQLITE3_DIR}/build/Release/better_sqlite3.node" ]; then
   (
     cd "${BETTER_SQLITE3_DIR}"
-    "${MONOREPO_ROOT}/node_modules/.bin/prebuild-install" ||
+    if ! "${MONOREPO_ROOT}/node_modules/.bin/prebuild-install"; then
+      echo "wallet-cli: prebuild-install failed (see its output above); compiling better-sqlite3 from source. This needs a C/C++ toolchain and Python." >&2
       "${MONOREPO_ROOT}/node_modules/.bin/node-gyp" rebuild --release
+    fi
   )
 fi
--- a/packages/wallet-cli/scripts/install-binaries.sh
+++ b/packages/wallet-cli/scripts/install-binaries.sh
@@ -11,13 +11,16 @@ cd "${PACKAGE_ROOT}"
 
 # Install the better-sqlite3 native addon if missing. Yarn has
 # `enableScripts: false` globally, so install scripts never run during
-# `yarn install` and the addon may be absent from the filesystem. Invoke the
-# prebuild-install binary directly to fetch a matching prebuild for the active
-# Node version and platform.
+# `yarn install` and the addon may be absent from the filesystem. Reproduce
+# better-sqlite3's own install step (`prebuild-install || node-gyp rebuild
+# --release`): fetch a matching prebuild for the active Node version and
+# platform, and fall back to compiling from source when no prebuild is
+# published for that ABI/libc combination (e.g. some Linux CI runners).
 BETTER_SQLITE3_DIR="${MONOREPO_ROOT}/node_modules/better-sqlite3"
 if [ ! -f "${BETTER_SQLITE3_DIR}/build/Release/better_sqlite3.node" ]; then
   (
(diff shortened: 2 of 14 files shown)


sirtimid:
@SocketSecurity ignore npm/better-sqlite3@12.10.0

sirtimid:
@SocketSecurity ignore npm/detect-libc@2.1.2
Shell access = spawning getconf/ldd to detect glibc vs musl for prebuild selection. Expected & necessary.

@SocketSecurity ignore npm/simple-get@4.0.1
Network access = the HTTP(S) client prebuild-install uses to fetch the binary from GitHub releases.

@SocketSecurity ignore npm/tunnel-agent@0.6.0
Network access = HTTPS-over-proxy tunneling for the prebuild download; "code anomaly" is a false positive.

@SocketSecurity ignore npm/chownr@1.1.4
"Code anomaly" false positive on isaacs' recursive-chown util, used via tar-fs to unpack the prebuild tarball.

@SocketSecurity ignore npm/prebuild-install@7.1.3
Socket flags as deprecated; it's better-sqlite3 v12's bundled prebuild fetcher, runs only at test-prep, not shipped.

All alerts are transitive deps of better-sqlite3's prebuild-install toolchain (better-sqlite3 → prebuild-install → {detect-libc, simple-get, tunnel-agent, tar-fs → chownr}), which downloads/unpacks the prebuilt native SQLite addon. None are in @metamask/wallet-cli's own code or runtime path. The monorepo sets enableScripts: false, so none run on yarn install; prebuild-install runs only via the package's explicit test:prepare script and in CI.

grypez (dismissed):
Familiar! Approved

packages/wallet-cli/scripts/install-binaries.sh
Mrtenz, line 1: Instead of this we can just install `better-sqlite3` into the root package and enable it to run the postinstall script there.

A better alternative would be to find a package that doesn't need postinstall scripts, though.

sirtimid: Thanks! Two parts to this:

On switching to a package without postinstall scripts: we're staying on `better-sqlite3` on purpose. The `ocap-kernel`, that `wallet-cli` will host, already uses it, so picking a different engine here would mean two SQLite implementations in one process.

On installing it at the root + allow-scripts: fair point, that's the idiomatic mechanism (`secp256k1` already goes through it). The reason I kept it as a scoped `test:prepare` step is to avoid running `better-sqlite3` native build on every monorepo `yarn install`. Also v12 only ships prebuilds for Node 22+, so contributors on Node 18 (still supported repo-wide) would fall back to a source compile, and an `allow-scripts` failure exits the whole install, for packages that have nothing to do with `wallet-cli`.

sirtimid: Once we can raise the nodejs floor to >= 22.5 and once it's not anymore experimental we can switch to `node:sqlite` but we researched all the alternatives at the `ocap-kernel` team and concluded that `better-sqlite3` is better, for now.

.github/workflows/lint-build-test.yml
```diff
@@ -143,6 +143,11 @@ jobs:
       matrix:
         node-version: [18.x, 20.x, 22.x]
         package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
+        exclude:
+          # @metamask/wallet-cli depends on better-sqlite3, which only ships
+          # prebuilt binaries for Node 20+.
+          - node-version: 18.x
+            package-name: '@metamask/wallet-cli'
```
Mrtenz: Is this needed after the change to the test script?

sirtimid: Good catch, no, it's not needed anymore. Fixed

Mrtenz (approved):

PR #9080, opened 2026-06-10
feat: add direct mUSD fiat injection flow for Money Account deposits

OGPoyraz:
## Explanation

This PR adds an alternative fiat deposit flow for Money Account behind the `useFiatMUSDQuoteToInjectForMoneyAccount` feature flag (`confirmations_pay_fiat`).

**Current flow:** Fiat provider buys ETH on Mainnet → Relay bridges ETH to Monad, swaps to mUSD, and vaults into the boring vault.

**New flow (flag ON + probe succeeds):** Fiat provider buys mUSD directly on Monad → delivers to the Money Account address → Relay only vaults (approve + deposit) on the same chain. No bridging or swapping needed.

**New flow (flag ON + probe fails):** Falls back silently to the current ETH flow.

### Key changes

**New file: `fiat-direct-musd-quotes-for-money-account.ts`**
- `probeMusdFiatAvailability` — lightweight probe that verifies at least one fiat provider can sell mUSD on Monad before committing to the direct path
- `getDirectMusdToMoneyAccountQuotes` — quote function that uses `MUSD_MONAD_FIAT_ASSET` as the fiat asset and delivers to the Money Account address (`txParams.from`) instead of the user's EVM account

**Modified files:**
- `FiatStrategy.ts` — routes to the new quote function when flag is ON + transaction is `moneyAccountDeposit`, with fallback to `getFiatQuotes` if the probe fails or returns empty
- `fiat-submit.ts` — detects the direct-mUSD path at submit time by inspecting the stored quote's source chain/token (immune to flag changes between quote and submit), overrides `walletAddress` and `fiatAsset` accordingly
- `TransactionPayController.ts` — removes eager `caipAssetId` assignment from the `shouldUpdateFiatToken` block; `caipAssetId` is now set by whichever quote function succeeds, eliminating a race condition between the controller's synchronous write and the async quote phase
- `fiat-quotes.ts` — exports shared helpers (`getRampsQuote`, `buildRelayRequestFromAmountFiat`, `combineQuotes`, `getRelayTotalFeeUsd`, `getRequiredTokens`) for reuse, adds `caipAssetId` to the `updateFiatPayment` call
- `utils.ts` — exports `resolveTransactionType`, adds `isMoneyAccountDepositTransaction`
- `feature-flags.ts` — adds `useFiatMUSDQuoteToInjectForMoneyAccount` to `FiatFlags` type + getter
- `constants.ts` — adds `CHAIN_ID_MONAD`, `MUSD_MONAD_ADDRESS`, `MUSD_MONAD_FIAT_ASSET`, `MUSD_PROBE_AMOUNT_USD`

### Design decisions

- **Provider-agnostic**: The probe and quote flow work with any fiat provider that supports mUSD on Monad — not coupled to Transak
- **Separate file for quotes**: The direct-mUSD quote function has different enough semantics (different fiat asset, different delivery address, probe gate) to warrant its own file rather than conditionals in `getFiatQuotes`
- **Modify existing for submit**: The submit flow is structurally identical — same poll → resolve amount → 3-phase flow — so only two targeted overrides are needed
- **Single `caipAssetId` writer**: Moving `caipAssetId` from the controller to the quote functions eliminates a race condition and ensures it always matches the actual quote returned
- **No mobile changes**: All routing is handled in core. Mobile callbacks and UI are unchanged

## References

- Plan: Obsidian `Plans/transak-musd-inject-money-account.md`

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes fiat quote/submit and relay recipient behavior on Money Account deposit paths; mitigated by feature flag, provider probe, and fallback to the legacy ETH flow.
> 
> **Overview**
> Adds an optional **direct mUSD fiat deposit path** for Money Account deposits when `confirmations_pay_fiat.directMoneyMusdEnabled` is on: probes ramps for mUSD on Monad, then quotes/submits with mUSD delivered to the Money Account (`txParams.from`) and relay overrides (`recipient`, `paymentOverride: moneyAccount`), skipping the ETH→bridge→mUSD path. Probe failure or empty direct quotes **fall back** to the existing fiat+relay flow.
> 
> **`fiatPayment.caipAssetId`** is no longer set on payment-method change in `TransactionPayController`; successful quote paths now set **`caipAssetId` and `rampsQuote` together** in `getFiatQuotes`. Submit detects the direct route from the stored quote’s mUSD-on-Monad source (not the flag) for wallet address, asset validation, and order polling.
> 
> Relay quotes accept an optional **`recipient`**; fiat fee adjustment can **exclude source-network gas** when the source token is not native gas. New Monad/mUSD constants, `isMoneyAccountDepositTransaction`, and expanded tests for standard and direct flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cade68a1810d5205fde3c87f5f564f3bf39fb82c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -894,6 +895,26 @@ export function getFiatMaxRateDriftPercent(
...
+ * @returns `true` if the direct mUSD injection flag is enabled.
+ */
+export function getUseFiatMUSDQuoteToInjectForMoneyAccount(
+  messenger: TransactionPayControllerMessenger,
+): boolean {
+  const state = messenger.call('RemoteFeatureFlagController:getState');
+  const fiatFlags = state.remoteFeatureFlags?.confirmations_pay_fiat as
+    | FiatFlags
+    | undefined;
+  return fiatFlags?.useFiatMUSDQuoteToInjectForMoneyAccount === true;
```
matthewwalsh0: Minor, maybe something simpler like `directToMusdEnabled`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -315,6 +315,13 @@ async function getSingleQuote(
       request.paymentOverride === PaymentOverride.MoneyAccount
     ) {
       await processMoneyAccountPostQuote(transaction, request, body, messenger);
+
+      // For same-chain Money Account flows, Relay rejects requests where
+      // user === recipient. Override recipient to the money account address
+      // so Relay treats it as a transfer rather than a self-send.
+      if (transaction.txParams?.from && body.recipient === from) {
+        body.recipient = transaction.txParams.from as Hex;
```
matthewwalsh0: For this flow, it should always be the EOA / from, since the EOA then transfers the funds to the recipient from the callback, so it can run the delegation from the money account to vault.

So where is it being set to anything different for this flow?

packages/transaction-pay-controller/src/strategy/fiat/constants.ts
```diff
@@ -43,3 +43,19 @@ export const FIAT_ENABLED_TYPES: TransactionType[] = [
   TransactionType.perpsDeposit,
   TransactionType.predictDeposit,
 ];
+
+/** Chain ID for Monad network. */
+export const CHAIN_ID_MONAD = '0x8f' as Hex;
```
matthewwalsh0: Minor, do we not have a central chain ID enum for the entire controller?

And central constants for token addresses like below?

packages/transaction-pay-controller/src/strategy/fiat/FiatStrategy.ts
```diff
@@ -4,14 +4,26 @@ import type {
...
 import { submitFiatQuotes } from './fiat-submit';
 import type { FiatQuote } from './types';
+import { isMoneyAccountDepositTransaction } from './utils';
 
 export class FiatStrategy implements PayStrategy<FiatQuote> {
   async getQuotes(
     request: PayStrategyGetQuotesRequest,
   ): Promise<TransactionPayQuote<FiatQuote>[]> {
+    if (
+      getUseFiatMUSDQuoteToInjectForMoneyAccount(request.messenger) &&
```
matthewwalsh0: For clarity, since this is part of the fiat quotes flow, could we have this inside `getFiatQuotes` but still callout to a different file if needed?

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -152,13 +157,30 @@ export async function getFiatQuotes(
...
 }
 
-function getRequiredTokens(
+/**
+ * Filters required tokens to those that need funding (i.e. not skipped when balance is sufficient).
+ *
+ * @param tokens - The full list of required tokens for a transaction.
+ * @returns Tokens that require funding.
+ */
+export function getRequiredTokens(
```
matthewwalsh0: Minor, this technically isn't needed anymore since we never use `skipIfBalance`.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -194,7 +216,23 @@ async function getRampsQuote({
...
+ * @param options.amountFiat - User-entered fiat amount in USD.
+ * @param options.fiatAsset - The source asset for the relay.
+ * @param options.fiatAsset.address - Token address of the source asset.
+ * @param options.fiatAsset.chainId - Chain ID of the source asset.
+ * @param options.messenger - Controller messenger for token info access.
+ * @param options.requiredToken - The target token the relay should deliver.
+ * @param options.walletAddress - Address of the user's wallet.
+ * @returns A relay quote request, or `undefined` if token info / rates are unavailable.
+ */
+export function buildRelayRequestFromAmountFiat({
```
matthewwalsh0: Another benefit of using the same file is we wouldn't need to export all these previously internal functions.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
```diff
@@ -252,13 +284,19 @@ async function submitRelayAfterFiatCompletion({
   });
 
   const baseRequest = quotes[0].request;
-  const walletAddress = baseRequest.from;
+
+  // For the direct mUSD flow, the fiat provider delivered to the money
+  // account (txParams.from), so on-chain amount resolution must look there.
+  // For the standard flow, it delivered to the user's account (baseRequest.from).
+  const sourceAmountWalletAddress = isDirectMusdToMA
```
matthewwalsh0: Minor, we do this twice so could we have a `getWalletAddress` function in this file to encapsulate the `isDirectMusdToMoneyAccountQuote` call too?

packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd-quotes-for-money-account.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * Probes fiat provider availability first. If no provider supports mUSD
+ * on Monad, returns an empty array so the caller can fall back to the
+ * standard fiat flow.
+ *
+ * @param request - Strategy quotes request.
+ * @returns A single combined fiat strategy quote, or an empty array when
+ * the probe fails or inputs/quotes are unavailable.
+ */
+export async function getDirectMusdToMoneyAccountQuotes(
```
matthewwalsh0: Minor, could we have this at the top as it's the entry point?

matthewwalsh0: Since this is only 200 lines, how much of it overlaps with the existing fiat quotes logic?

Is the main difference just that we're checking support from the Ramps quote first and forcing the fiat asset?

Would we improve readability and maintenance if we had a single fiat quote flow file with two small conditions / branches for conditional probe and conditional fiat asset override?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: For same-chain Money Account flows (direct mUSD on Monad), the Relay request has user = EOA and recipient = EOA (both derived from from).  And then Relay rejects same-address send transactions with `400 - Sender and recipient cannot be the same` (adding screenshot). 

The override sets recipient to the money account address (`txParams.from`) so Relay treats it as a transfer. The delegation callback then handles the vault deposit from the money account. This only activates for same-chain flows where `body.recipient === from`; cross-chain flows are unaffected since the bridge naturally routes to a different recipient.

Screenshot 2026-06-11 at 14 44 19

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Isn't `directToMusdEnabled` is too broad and lacks `MoneyAccount` context? Because we already have `assetPerTransactionType`

packages/transaction-pay-controller/src/strategy/fiat/FiatStrategy.ts
OGPoyraz: Done, moved the routing logic.

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: Done.

packages/transaction-pay-controller/src/strategy/fiat/fiat-submit.ts
OGPoyraz: Good call, extracted this as a tiny helper.

packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd-quotes-for-money-account.ts
OGPoyraz: This file is removed now.

OGPoyraz: Merged this all this in the fiat-quotes.ts now

packages/transaction-pay-controller/src/strategy/fiat/constants.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -159,10 +185,172 @@ export async function getFiatQuotes(
...
+    });
+
+    return isAvailable;
+  } catch (error) {
+    log('mUSD fiat probe failed', { error });
+    return false;
+  }
+}
+
+async function getDirectMusdToMoneyAccountQuotes(
```
matthewwalsh0: This is better, but I was meaning can we not extend the existing logic for fiat quotes since this is still primarily the same flow.

We still call all the following in the same order:

- `buildRelayRequestFromAmountFiat`
- `getRelayQuotes`
- `getRelayTotalFeeUsd` 
- `TransactionPayController:updateFiatPayment`
- `combineQuotes`

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0: But if this is only for the benefit of our new flow, why not just set the `recipient` when we call  `getRelayQuotes` with a new optional property?

packages/transaction-pay-controller/src/utils/feature-flags.ts
matthewwalsh0: `directMoneyMusdEnabled`?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Done 👍

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Done now, added an optional `recipient` parameter

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: Understood, collapsed this into single quote function

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -467,10 +469,13 @@ export function getFeatureFlags(
...
-  const estimate =
-    featureFlags.relayFallbackGas?.estimate ?? DEFAULT_FALLBACK_GAS_ESTIMATE;
+  const estimate = new BigNumber(
+    featureFlags.relayFallbackGas?.estimate ?? DEFAULT_FALLBACK_GAS_ESTIMATE,
+  ).toNumber();
 
-  const max = featureFlags.relayFallbackGas?.max ?? DEFAULT_FALLBACK_GAS_MAX;
+  const max = new BigNumber(
+    featureFlags.relayFallbackGas?.max ?? DEFAULT_FALLBACK_GAS_MAX,
+  ).toNumber();
```
OGPoyraz, line 478: @matthewwalsh0 I think this is a critical find:

`confirmations_pay.relayFallbackGas` returns estimate and max as strings (e.g. "900001") despite the TS type declaring number. When gas simulation fails and the fallback kicks in, these string values flow into `combinePostQuoteGas` and `addPaymentOverrideGas` where + concatenates instead of adding - producing astronomical gas estimates (e.g. "9000017000000075000" instead of 71575001). This inflates `sourceNetwork` fees shows absurd amount of USD in the displayed totals.

This is a pre-existing bug affecting all flows where gas simulation fails, but in practice only surfaced in the direct mUSD same-chain flow because it's the only path where 
- the EOA lacks the token balance for simulation to succeed 
- `combinePrependedGas` runs (account override)
- `addPaymentOverrideGas` further compounds the concatenation.

I am adding the log we added below:

```
{
    "raw": {
        "payStrategies": {
            "relay": {
                "pollingTimeout": 180000,
                "enabled": true,
                "executeEnabled": false,
                "gaslessEnabled": true,
                "originGasOverhead": "300000",
                "pollingInterval": 1500
            },
            "across": {
                "apiBase": "https://intents.uat-api.cx.metamask.io/across",
                "enabled": false,
                "fallbackGas": {
                    "estimate": 900001,
                    "max": 1500001
                }
            }
        },
        "slippageTokens": {
            "0x1": {
                "0x0000000000000000000000000000000000000000": 0.005,
                "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599": 0.005,
                "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": 0.005,
                "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": 0.005,
                "0xacA92E438df0B2401fF60dA7E4337B687a2435DA": 0.005,
                "0xdAC17F958D2ee523a2206206994597C13D831ec7": 0.005
            },
            "0x2105": {
                "0x0000000000000000000000000000000000000000": 0.005,
                "0x4200000000000000000000000000000000000006": 0.005,
                "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913": 0.005,
                "0xfde4C96c8593536E31F229EA8f37b2ADa2699bb2": 0.005
            },
            "0x38": {
                "0x0555E30da8f98308EdB960aa94C0Db47230d2B9c": 0.005,
                "0x2170Ed0880ac9A755fd29B2688956BD959F933F8": 0.005,
                "0x55d398326f99059fF775485246999027B3197955": 0.005,
                "0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d": 0.005,
                "0x0000000000000000000000000000000000000000": 0.005
            },
            "0x89": {
                "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619": 0.005,
                "0xc2132D05D31c914a87C6611C10748AEb04B58e8F": 0.005,
                "0x0000000000000000000000000000000000001010": 0.005,
                "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174": 0.005,
                "0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359": 0.005
            },
            "0xa4b1": {
                "0xaf88d065e77c8cC2239327C5EDb3A432268e5831": 0.005,
                "0x0000000000000000000000000000000000000000": 0.005,
                "0x82aF49447D8a07e3bd95BD0d56f35241523fBab1": 0.005,
                "0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9": 0.005
            },
            "0xe708": {
                "0x0000000000000000000000000000000000000000": 0.005,
                "0x176211869cA2b568f2A7D4EE941E073a821EE1ff": 0.005,
                "0xA219439258ca9da29E9Cc4cE5596924745e12B93": 0.005,
                "0xacA92E438df0B2401fF60dA7E4337B687a2435DA": 0.005,
                "0xe5D7C2a44FfDDf6b295A15c148167daaAf5Cf34f": 0.005
            }
        },
        "strategyOrder": [
            "relay"
        ],
        "bufferSubsequent": 0.05,
        "relayFallbackGas": {
            "estimate": "900001",
            "max": "1500001"
        },
        "allowedPredictWithdrawTokens": {
            "0x38": [
                "0x0000000000000000000000000000000000000000",
                "0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d"
            ],
            "0x89": [
                "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
                "0x0000000000000000000000000000000000000000",
                "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619"
            ],
            "0x1": [
                "0x0000000000000000000000000000000000000000",
                "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
            ]
        },
        "bufferInitial": 0.015,
        "bufferStep": 0.015,
        "slippage": 0.02,
        "relayDisabledGasStationChains": [],
        "relayQuoteUrl": "https://intents.uat-api.cx.metamask.io/relay/quote",
        "stxDisabled": false,
        "predictWithdrawAnyToken": true,
        "perpsWithdrawAnyToken": false,
        "dev": true,
        "attemptsMax": 4,
        "relayExecuteUrl": "https://intents.uat-api.cx.metamask.io/relay/execute"
    },
    "result": {
        "relayDisabledGasStationChains": [],
        "relayExecuteUrl": "https://intents.uat-api.cx.metamask.io/relay/execute",
        "relayFallbackGas": {
            "estimate": "900001",
            "max": "1500001"
        },
        "relayQuoteUrl": "https://intents.uat-api.cx.metamask.io/relay/quote",
        "slippage": 0.02
    }
}
```

matthewwalsh0: Great spot Goktug, though I think this only became an issue when we started adding to the gas limit directly for post quote and now fiat since we had always wrapped in a `BigNumber` when calculating the cost:

https://github.com/MetaMask/core/blob/main/packages/transaction-pay-controller/src/utils/gas.ts#L131

And then passing it through `toHex` during submit:

https://github.com/MetaMask/core/blob/main/packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts#L697

https://github.com/MetaMask/core/blob/main/packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts#L282

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -19,28 +25,79 @@ import {
...
+  });
+}
+
+type FiatQuotePipelineOptions = {
+  fiatAsset: TransactionPayFiatAsset;
+  rampsWalletAddress: Hex;
+  relayRequestOverrides?: Partial<QuoteRequest>;
+};
+
+async function executeFiatQuotePipeline(
```
matthewwalsh0, line 98: ❤️

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
```diff
@@ -95,7 +156,14 @@ export async function getFiatQuotes(
...
 
-    const relayTotalFeeUsd = getRelayTotalFeeUsd(relayQuote);
+    const isSourceNative =
+      fiatAsset.address.toLowerCase() === NATIVE_TOKEN_ADDRESS.toLowerCase();
+    const gasDeductedFromSource =
+      isSourceNative || relayQuote.fees.isSourceGasFeeToken === true;
+
+    const relayTotalFeeUsd = gasDeductedFromSource
+      ? getRelayTotalFeeUsd(relayQuote)
+      : getNonGasRelayFeeUsd(relayQuote);
```
matthewwalsh0, line 166: Minor, if we're using our gas station for the relay quote, won't the user still need to cover the source network cost as it's calculated from the gas fee token so includes the original gas fee plus commission but it will be paid via additional input token?

Similarly, if a native token, there could still be a source network cost?

Do we add this back somewhere else?

matthewwalsh0 (approved):

packages/transaction-pay-controller/src/strategy/fiat/fiat-quotes.ts
OGPoyraz: As this is for `mUSD` flow and Monad transactions are paid by us, this was increasing the total ramps quote unnecessarily, but again as this is just for quoting it's safe for now, I will revisit in case if we need presented fee fixes.

PR #9100, opened 2026-06-11
feat: MUSD-943 money-account-balance-service optimizations

Matt561:
## Explanation

The Money account balance service currently makes multiple sequential RPC calls to fetch a user's Money account  balance (mUSD wallet balance + vault share value).

This PR addresses both issues:

- **Multicall batching:** A new `getMoneyAccountBalance` method fetches the mUSD wallet balance and the vmUSD (Veda vault tokens) in a single multicall. Previously, we needed 3 RPC calls each time we fetched the balance (getUnderlyingToken, getMusdBalance, getVmusdBalance).

- **Configurable stale time:** The balance cache `staleTime` is now driven by a remote feature flag (`moneyAccountBalanceStaletime`). The default balance `staleTime` was bumped from 30s to 60s.

- **Underlying token from config:** The vault config now accepts an optional `underlyingToken` address. When present, the service uses it directly instead of making an extra `Accountant.base()` on-chain read on every mUSD balance fetch. Falls back to the on-chain read when absent (backwards compatible with existing flag payloads).

- **Rename mUSDSHFvd → vmUSD:** The vault share token is called vmUSD everywhere else, so the public API now reflects that (`getMusdSHFvdBalance` → `getVmusdBalance`, etc.). This is a breaking change.

## References

- MUSD-943

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking public API renames require consumer updates; balance and vault config changes affect on-chain reads and caching behavior, though defaults and fallbacks limit rollout risk.
> 
> **Overview**
> Adds **`getMoneyAccountBalance`**, which batches mUSD `balanceOf` and Lens `balanceOfInAssets` in one Multicall3 `aggregate3` call and returns `musdBalance`, `vmusdValueInMusd`, and `totalBalance`.
> 
> **Breaking:** Renames the vault-share API from `musdSHFvd` / `getMusdSHFvdBalance` to **vmUSD** / `getVmusdBalance` (including messenger action types).
> 
> Vault config may include optional **`underlyingToken`** so mUSD reads skip `Accountant.base()` when set; behavior stays backward compatible when omitted.
> 
> On-chain balance cache **`staleTime`** defaults to **60s** (was 30s) and can be overridden at runtime via the **`moneyAccountBalanceStaletime`** remote feature flag (`init` + `RemoteFeatureFlagController:stateChange`, with validation and fallback).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6a3b01f903d2b15834b667132b5f0e5fd0e08ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Matt561:
@metamaskbot publish-previews

packages/money-account-balance-service/src/constants.ts
```diff
@@ -8,11 +8,70 @@ export const VEDA_PERFORMANCE_API_BASE_URL = 'https://api.sevenseas.capital';
  */
 export const VAULT_CONFIG_FEATURE_FLAG_KEY = 'moneyAccountVaultConfig';
 
+/**
+ * The key under which the Money account balance `staleTime` (in milliseconds)
+ * is stored in `RemoteFeatureFlagController` state's `remoteFeatureFlags` map.
+ * Falls back to {@link DEFAULT_BALANCE_STALE_TIME} when absent or malformed.
+ */
+export const MONEY_ACCOUNT_BALANCE_STALETIME_FEATURE_FLAG_KEY =
```
Jwhiles, line 16: I'm not sure it makes sense to store the feature flags in the core library. To me it feels more like a client concern, and in general most of feature flags seem to be defined in the clients - so defining some of them in core seems to make the situation more confusing.

I understand that `VAULT_CONFIG_FEATURE_FLAG_KEY` is already defined in this package, so I’m happy to approve - but IMO we should move away from this pattern

Jwhiles: Also, the changelog says it's `moneyAccountBalanceStaleTime` - note the capital T. Would be good to get them matching ;)

packages/money-account-balance-service/src/money-account-balance-service.test.ts
```diff
@@ -325,6 +329,64 @@ function mockLensBalanceOfInAssets(balanceOfInAssets: string): void {
...
+  };
+} {
+  return {
+    toString: () => value,
+    add: (other) =>
+      makeMockBN((BigInt(value) + BigInt(other.toString())).toString()),
+  };
+}
+
+function mockMoneyAccountBalanceMulticall({
```
Jwhiles, line 348: This mocks out basically everything in `getMoneyAccountBalance` - and I think it takes away a bit of the value of the test.

For example, I think it would be nice to have at least one test actually exercise the decode/encode paths - cos at the moment we only know they work from manual testing.

Jwhiles (dismissed):
Suggested a couple of minor tweaks - but generally happy to merge

packages/money-account-balance-service/src/constants.ts
Matt561: Good catch. I've fixed the typo in the changelog!

As discussed, we'll open a separate PR if we decide to move feature flags to the client.

Jwhiles (approved):

PR #9083, opened 2026-06-10
Release/1032.0.0

Prithpal-Sooriya:
## Explanation

Bumps `@metamask/controller-utils` from `12.1.1` to `12.2.0`
Bumps `@metamask/assets-controller` from `9.0.0` to `9.0.1`

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Dependency and version-metadata only; risk is limited to consumers picking up controller-utils 12.2.0 (e.g. faster `encodeFunctionData`) via semver ranges, not behavioral changes in this PR’s diff.
> 
> **Overview**
> Prepares **monorepo release `1032.0.0`** by aligning package versions, changelogs, and `yarn.lock`—no application source changes in this diff.
> 
> **`@metamask/controller-utils`** is released as **`12.2.0`** (from `12.1.1`); dependents move to **`^12.2.0`**, with Unreleased changelog lines updated to record the bump ([#9083](https://github.com/MetaMask/core/pull/9083)).
> 
> **`@metamask/assets-controller`** is patched to **`9.0.1`** for the same `controller-utils` dependency; **`bridge-controller`** and **`transaction-pay-controller`** pick up **`^9.0.1`** where they depend on assets-controller.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92da19d1f1a9b56f04975fe2d8f31dbc7170642d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9058, opened 2026-06-09
Release/1027.0.0

ccharly:
## Explanation

Major release of the `keyring-controller` (and `wallet` since there was a pending breaking change already for that package).

## References

N/A

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **High Risk**
> Major bumps to keyring-controller and wallet change export/signing credentials and default approval wiring; consumers must update call sites and avoid duplicate ApprovalController registration.
> 
> **Overview**
> Monorepo release **1027.0.0** that cuts new package versions and aligns workspace dependencies so downstream clients can pick up recent controller work in one train.
> 
> The headline breaking changes are **`@metamask/keyring-controller@27.0.0`** (`exportSeedPhrase` / `exportAccount` now take a `Credentials` object instead of a raw password string; non-primary keyrings are auto-removed when drained via `withKeyring` / `withKeyringV2`) and **`@metamask/wallet@3.0.0`** (default `Wallet` now constructs and registers **`ApprovalController`**—apps that already wire their own must drop the duplicate).
> 
> Also shipped in this bump wave: **`@metamask/assets-controllers@108.6.0`** adds runtime **`isDeprecated`** hooks on `TokenRatesController`, `TokenBalancesController`, and `AccountTrackerController`; **`@metamask/controller-utils@12.1.1`** refreshes `NetworkNickname`; **`@metamask/profile-metrics-controller@3.2.0`** adds proof-of-ownership / nonce batch API wiring. Dozens of packages only receive patch/minor version and changelog updates plus dependency bumps to `keyring-controller@^27`, `controller-utils@^12.1.1`, and related packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94aae7e7e8ffa7f82a4135e2d7f3d5bf72228951. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9105, opened 2026-06-12
Release/1038.0.0

meltingice1337:
## Explanation

This is the release PR for `1038.0.0`, generated by `create-release-branch`. It bumps the monorepo version from `1037.0.0` to `1038.0.0` and releases a single package: `@metamask/ramps-controller` `14.1.1` → `14.2.0`. The release moves the entries that had accumulated under `[Unreleased]` into a new `[14.2.0]` section in the package's changelog.

`@metamask/ramps-controller@14.2.0` includes the following:

- **Changed** — `RampsService.getQuotes` now sends an `Authorization: Bearer <token>` header, sourcing the token from `AuthenticationController:getBearerToken` (already a required messenger action since `14.0.0`); the call throws if no token is available (e.g. the wallet is locked or the user is signed out) ([#8888](https://github.com/MetaMask/core/pull/8888)). This is the notable change driving the version bump: `getQuotes` previously made an unauthenticated request and now requires a bearer token, so consumers should review their sign-in state before upgrading.
- **Changed** — Bump `@metamask/controller-utils` from `^12.1.0` to `^12.2.0` ([#9058](https://github.com/MetaMask/core/pull/9058), [#9083](https://github.com/MetaMask/core/pull/9083)).

Note on `getPaymentMethods`: within this unreleased window, auth was added to `getPaymentMethods` (#8888) and then removed again (#9060) so that payment methods can be fetched before sign-in. Because that change was reverted before shipping, it nets to no consumer-facing change and is intentionally omitted from the changelog per the project's "omit reverted changes" guideline.

This is a minor release (`14.2.0`). It contains no breaking changes; the version is bumped to minor — rather than patch — to surface the `getQuotes` authentication behavior change, which alters how an existing method makes its request even though the public interface is unchanged. As part of cutting the release, the in-monorepo consumer `@metamask/transaction-pay-controller` has its dependency on `@metamask/ramps-controller` bumped to `^14.2.0`; this is a mechanical version bump required by the release tooling and does not change `transaction-pay-controller`'s behavior.

## References

Release PR generated by `create-release-branch`; see `docs/processes/releasing.md`. `@metamask/ramps-controller` was bumped to a minor (`14.2.0`) rather than a patch in response to review feedback that the `getQuotes` authentication change is a notable behavior change to an existing method.

## Checklist

### DIFF
--- a/packages/ramps-controller/CHANGELOG.md
+++ b/packages/ramps-controller/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [14.1.2]
+
 ### Added
 
 - Authenticate `RampsService.getPaymentMethods` and `RampsService.getQuotes` by sourcing a bearer token from `AuthenticationController:getBearerToken` and sending it as an `Authorization: Bearer <token>` header ([#8888](https://github.com/MetaMask/core/pull/8888))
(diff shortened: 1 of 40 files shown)


packages/ramps-controller/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [14.1.2]
+
 ### Added
 
 - Authenticate `RampsService.getPaymentMethods` and `RampsService.getQuotes` by sourcing a bearer token from `AuthenticationController:getBearerToken` and sending it as an `Authorization: Bearer <token>` header ([#8888](https://github.com/MetaMask/core/pull/8888))
```
Mrtenz: "Added" implicates that this is a new feature, in which case we should release this with a minor bump. The changelog entry doesn't read as a new feature though. 🤔

packages/ramps-controller/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [14.1.2]
```
FrederikBolding: Should this be a minor bump as new features were added?

Mrtenz (approved):

PR #9108, opened 2026-06-12
feat(wallet-cli): add daemon JSON-RPC socket transport

sirtimid:
## Explanation

`@metamask/wallet-cli` runs a background daemon that hosts a `@metamask/wallet` instance; the CLI talks to it over a Unix socket. This PR adds that transport layer. It is purely additive plumbing — no `mm` subcommand consumes it yet (commands land in a follow-up), so there is no user-visible behavior change.

- **`socket-line`** — `writeLine`/`readLine` for newline-delimited framing over a `net.Socket`, with an optional read timeout and listener cleanup.
- **`daemon-client`** — `sendCommand` opens a connection, writes one JSON-RPC request, reads one response, and closes; it correlates the response `id` with the request and retries once on transient connection errors (`ECONNREFUSED`/`ECONNRESET`). `pingDaemon` is a lightweight `getStatus` health probe that distinguishes "no daemon" (`absent`) from "daemon present but unreachable", classifying the latter by failure mode (`refused` / `timeout` / `permission` / `protocol` / `other`).
- **`rpc-socket-server`** — `startRpcSocketServer` listens on a Unix socket and dispatches one JSON-RPC request per connection to a handler map. It intercepts a built-in `shutdown` method, enforces one-request-per-connection, times out idle connections, and returns a `close()` handle.
- **`daemon-spawn`** — `ensureDaemon` spawns a detached daemon process and polls until the socket is responsive, refusing to take over a wedged or foreign-owned socket. Resolves the entry point from `dist` (prod) or `src` via `tsx` (dev).
- **`stop-daemon`** — `stopDaemon` escalates from a graceful `shutdown` RPC through `SIGTERM` to `SIGKILL`, then removes the PID and socket files best-effort.
- **`prompts`** — `confirmPurge` wraps the ESM-only `@inquirer/confirm` via dynamic import.
- **`types`** — adds `RpcHandler`, `RpcHandlerMap`, `DaemonStatusInfo`, and `DaemonSpawnConfig`.

Adds `@inquirer/confirm` and `@metamask/rpc-errors` dependencies. Every daemon module is covered to the package's 100% coverage thresholds, plus a `socket-integration.test.ts` that exercises the client and server together over a real Unix socket (framing, id correlation, the one-request-per-connection invariant, and real shutdown timing).

### Hardening applied on top of the original branch

A multi-agent review of the ported code surfaced a few issues worth fixing before any command consumes this transport. These are the only deviations from the original `rekm/wallet-cli` modules (otherwise lifted verbatim):

- **Owner-only socket permissions** (`rpc-socket-server.ts`) — `listen()` created the Unix socket with umask-derived (typically world-accessible) permissions. Since the daemon hosts an unlocked wallet, any local user could connect and drive it. The server now `chmod`s the socket to `0600` immediately after binding.
- **Fail fast on a failed spawn** (`daemon-spawn.ts`) — a `ChildProcess` `'error'` (bad interpreter, `EACCES`, `ENOENT`) was logged to stderr but never recorded, so the readiness loop polled for the full 30s and then threw a generic timeout, discarding the real cause. The error is now captured and surfaced immediately.
- **No unhandled `'error'` during a write** (`socket-line.ts`) — `writeLine` relied solely on the write callback; a socket `'error'` arriving mid-write had no listener and would crash the process instead of rejecting. It now attaches a one-shot `'error'` listener for the duration of the write.
- **Test robustness** (`socket-integration.test.ts`) — replaced an `await import('node:net')` with a static import so the file runs standalone without `--experimental-vm-modules`.

## References

- The daemon transport modules originate from Erik Marks's wallet-cli branch (`rekm/wallet-cli`, #8446). They are lifted essentially verbatim from there, split out into this standalone, additive PR and adjusted to apply on the current `main` (see the hardening deltas above).

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches local IPC and process lifecycle around an unlocked-wallet daemon, including owner-only socket permissions and signal-based stop; additive only with no CLI wiring yet, but mistakes could affect security or orphan processes once commands land.
> 
> **Overview**
> Adds **additive daemon IPC plumbing** to `@metamask/wallet-cli` so the CLI can talk to a background wallet daemon over a Unix socket. Nothing in this PR wires it into `mm` subcommands yet.
> 
> The stack is newline-delimited JSON-RPC: **`socket-line`** for framing, **`daemon-client`** (`sendCommand` with request/response id checks and one retry on transient connect errors; **`pingDaemon`** with `absent` / `responsive` / categorized `unreachable`), and **`startRpcSocketServer`** (one request per connection, handler map, built-in `shutdown`, idle timeouts). **`ensureDaemon`** spawns a detached process and polls readiness while refusing to replace a wedged or other-user socket; **`stopDaemon`** escalates shutdown RPC → SIGTERM → SIGKILL and cleans up PID/socket files. Shared **`utils`** cover PID files and signaling; **`confirmPurge`** dynamically imports `@inquirer/confirm`. **`types.ts`** gains RPC and spawn config types.
> 
> **Security / robustness hardening** in this PR: socket files are **`chmod` 0600** after bind (wallet daemon must not be world-connectable), failed child **`spawn` errors fail fast** instead of a 30s timeout, and **`writeLine`** handles mid-write socket errors. Dependencies **`@metamask/rpc-errors`** and **`@inquirer/confirm`** are added; changelog documents the transport layer. Broad unit tests plus **`socket-integration.test.ts`** exercise real sockets end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 556cbd742f22049b0a27c2d7ecf4408b98a65f90. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/wallet-cli/src/daemon/rpc-socket-server.ts
+++ b/packages/wallet-cli/src/daemon/rpc-socket-server.ts
@@ -5,7 +5,7 @@ import type {
   JsonRpcResponse,
 } from '@metamask/utils';
 import { hasProperty, isJsonRpcRequest } from '@metamask/utils';
-import { unlink } from 'node:fs/promises';
+import { chmod, unlink } from 'node:fs/promises';
 import { createServer } from 'node:net';
 import type { Server } from 'node:net';
 
@@ -316,11 +316,16 @@ async function listen(server: Server, socketPath: string): Promise<void> {
     }
   }
 
-  return new Promise((resolve, reject) => {
+  await new Promise<void>((resolve, reject) => {
     server.on('error', reject);
     server.listen(socketPath, () => {
       server.removeListener('error', reject);
       resolve();
     });
   });
+
+  // Restrict the socket to its owner. The daemon hosts an unlocked wallet, so
+  // a world-connectable socket would let any local user drive it. listen()
+  // creates the socket with umask-derived (typically world-accessible) perms.
+  await chmod(socketPath, 0o600);
 }
--- a/packages/wallet-cli/src/daemon/rpc-socket-server.ts
+++ b/packages/wallet-cli/src/daemon/rpc-socket-server.ts
@@ -145,17 +145,7 @@ export async function startRpcSocketServer({
   await listen(server, socketPath);
 
   return {
-    close: async (): Promise<void> => {
-      await new Promise<void>((resolve, reject) => {
-        server.close((error) => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve();
-          }
-        });
-      });
-    },
+    close: async (): Promise<void> => closeServer(server),
   };
 }
 
@@ -215,7 +205,6 @@ async function handleRequest(
   const { id, method, params } = parsed;
 
   try {
-    // Intercept shutdown before handler dispatch.
     if (method === 'shutdown') {
       if (onShutdown) {
         setTimeout(() => {
@@ -327,5 +316,30 @@ async function listen(server: Server, socketPath: string): Promise<void> {
   // Restrict the socket to its owner. The daemon hosts an unlocked wallet, so
   // a world-connectable socket would let any local user drive it. listen()
   // creates the socket with umask-derived (typically world-accessible) perms.
-  await chmod(socketPath, 0o600);
+  try {
+    await chmod(socketPath, 0o600);
+  } catch (error) {
+    // Never leave a possibly world-accessible socket listening with no handle
+    // to close it. Cleanup is best-effort so the chmod failure still surfaces.
+    await closeServer(server).catch(() => undefined);
+    await unlink(socketPath).catch(() => undefined);
+    throw error;
+  }
+}
+
+/**
+ * Close a server, resolving once it stops accepting connections.
+ *
+ * @param server - The net.Server instance.
+ */
+async function closeServer(server: Server): Promise<void> {
+  await new Promise<void>((resolve, reject) => {
+    server.close((error) => {
+      if (error) {
+        reject(error);
+      } else {
+        resolve();
+      }
+    });
+  });
 }
--- a/packages/wallet-cli/src/daemon/rpc-socket-server.ts
+++ b/packages/wallet-cli/src/daemon/rpc-socket-server.ts
@@ -1,0 +1,326 @@
+import { rpcErrors } from '@metamask/rpc-errors';
+import type {
+  JsonRpcId,
+  JsonRpcParams,
+  JsonRpcResponse,
+} from '@metamask/utils';
+import { hasProperty, isJsonRpcRequest } from '@metamask/utils';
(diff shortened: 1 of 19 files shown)


packages/wallet-cli/src/daemon/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+  // Duck-typed rather than `instanceof Error` so error values that crossed a
+  // realm boundary (e.g. from Node built-ins under jest's
+  // `--experimental-vm-modules`) still pass.
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    hasProperty(error, 'code') &&
+    error.code === code
+  );
+}
```
grypez, line 13: Nit: I think `@metamask/utils` already has a duck-typed `isErrorWithCode` implementation, though it doesn't apply the last `error.code === code` check.
```
if isErrorWithCode(error) {
  switch (error.code) {
    case 'ENOENT':
      ...
    default:
      throw error;
  }
}
```

`if isErrorWithCode(error) && error.code === 'ENOENT' { ... }`

packages/wallet-cli/src/daemon/rpc-socket-server.ts
```diff
@@ -0,0 +1,331 @@
...
+    server.listen(socketPath, () => {
+      server.removeListener('error', reject);
+      resolve();
+    });
+  });
+
+  // Restrict the socket to its owner. The daemon hosts an unlocked wallet, so
+  // a world-connectable socket would let any local user drive it. listen()
+  // creates the socket with umask-derived (typically world-accessible) perms.
+  await chmod(socketPath, 0o600);
```
grypez: Flagging, but out of scope for this PR: there is real time between when `server.listen` acquires the socket and when the socket permissions are set to user-read-write. The associated test shows that this line significantly improves the situation. To create the socket with appropriate permissions, we could ensure / rely upon either the directory containing `socketPath` to be `0o700` or `umask` to be `0o077` before socket creation. There may be even better ways.

grypez (dismissed):
A nit and a comment, nothing blocking. Approved!

packages/wallet-cli/src/daemon/utils.ts
sirtimid: Done https://github.com/MetaMask/core/pull/9108/commits/60742bf04b51fd2adce03a121278064d3455bb2e

packages/wallet-cli/src/daemon/rpc-socket-server.ts
sirtimid: Agreed, and good catch. Let's apply the right fix on a follow-up.

grypez (approved):

PR #9122, opened 2026-06-15
chore: add `sha256` hash to `packageManager` field

Mrtenz:
## Explanation

The root `package.json` previously set `packageManager` to `yarn@4.16.0` with no integrity hash. Corepack supports an optional `+sha256.<hash>` suffix that pins the Yarn binary to a known checksum, ensuring that anyone running `corepack` against this repo gets the exact same Yarn binary that was vetted.

This PR adds the sha256 hash to the `packageManager` field. To accommodate it, the Yarn constraint in `yarn.config.cjs` is relaxed from an exact-string match (`yarn@4.16.0`) to a structural check via a new `expectYarnPackageManager` helper that asserts the field starts with `yarn@` and includes a sha256 hash. This lets the hash evolve when Yarn is bumped without requiring a constraint change at the same time.

## References

N/A

## Checklist

- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md) — N/A (no consumer-facing package changes)

### DIFF
--- a/yarn.config.cjs
+++ b/yarn.config.cjs
@@ -258,7 +258,7 @@ module.exports = defineConfig({
       if (isChildWorkspace) {
         workspace.unset('packageManager');
       } else {
-        expectWorkspaceField(workspace, 'packageManager', 'yarn@4.16.0');
+        expectYarnPackageManager(workspace);
       }
 
       // All packages must specify a minimum Node.js version of 18.18.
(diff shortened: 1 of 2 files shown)


FrederikBolding (dismissed):

yarn.config.cjs
```diff
@@ -1020,3 +1020,26 @@ async function expectCodeowner(workspace, workspaceBasename) {
...
 }
+
+/**
+ * Expect that the workspace has a package manager set, and that it is Yarn with
+ * a sha256 hash.
+ *
+ * @param workspace - The workspace to check.
+ */
+function expectYarnPackageManager(workspace) {
+  expectWorkspaceField(workspace, 'packageManager');
```
Mrtenz, line 1031: I don't think checking for a specific version here really makes sense. This field should only ever be set in the root manifest anyway. I've updated this to check for:
- Presence of a "packageManager" field.
- Whether that starts with "yarn@".
- Whether it includes a sha256 checksum.

FrederikBolding (approved):

PR #9137, opened 2026-06-16
feat(perps-controller): integrate Terminal API for market metadata with HyperLiquid fallback

michalconsensys:
## Explanation

The Perps Controller currently relies exclusively on HyperLiquid for market metadata (names, categories, market types). This is limiting because adding or categorising new markets requires code changes to static maps.

This PR integrates the MetaMask Terminal API as an optional market data source, controlled by the caller via `useTerminalApi` parameter on `GetMarketsParams` / `GetMarketDataWithPricesParams`. When enabled:

- **`getMarkets()`** attempts the Terminal API first; on failure or empty response, falls back silently to the HyperLiquid provider. Terminal results respect the same allowlist/blocklist filtering as the provider path via a new `isMarketAllowed` filter built by `PerpsController`.
- **`getMarketDataWithPrices()`** fetches live pricing from HyperLiquid as before, then enriches it with Terminal API metadata (human-readable name, keywords, tags, categories, marketType).
- **Market search** (`getMarketMatchRank`, `rankMarketsByQuery`) now indexes the `keywords` field for richer results.
- **Standalone discovery** (`getMarkets({ standalone: true })`) now routes through `MarketDataService` instead of calling the provider directly, ensuring consistent filtering and Terminal API support.

Key design decisions:

- **`TerminalMarketService`** validates each API response item individually with `@metamask/superstruct`; invalid items are filtered out and logged rather than rejecting the entire response.
- A 5-minute in-memory cache avoids redundant network calls across polling cycles.
- `PerpsPlatformDependencies` gains optional `terminalApiUrl?: string` and `terminalMarketService?: PerpsTerminalMarketService` fields — clients can inject a pre-built service instance or let the controller create one from the URL.
- `PerpsMarketData` gains optional `keywords`, `tags`, and `categories` fields.
- Static fallback maps (`HYPERLIQUID_ASSET_NAMES`, `HIP3_ASSET_MARKET_TYPES`) remain intact for assets absent from the Terminal API.

## References

- Fixes https://github.com/MetaMask/core/pull/9137

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core market discovery and listing behavior when `useTerminalApi` is on, including fallback rules and filter parity; failures are logged but should be validated in client integrations.
> 
> **Overview**
> Adds an optional **MetaMask Terminal API** path for perps market metadata, gated by `useTerminalApi` on `GetMarketsParams` and `GetMarketDataWithPricesParams`.
> 
> When enabled, **`getMarkets`** tries Terminal first (with dex/symbol filtering and the same HIP-3 allowlist/blocklist via a new `isMarketAllowed` filter from `PerpsController`), then **falls back to HyperLiquid** on failure, empty data, or when a constrained query has no Terminal matches. **`getMarketDataWithPrices`** still loads live prices from the provider and **merges** Terminal fields (`name`, `keywords`, `tags`, `categories`, `marketType`). **Standalone** `getMarkets` now goes through `MarketDataService` like the normal path.
> 
> Introduces **`TerminalMarketService`** (5-minute cache, `@metamask/superstruct` per-item validation), optional `terminalApiUrl` / `terminalMarketService` on platform deps, and extends **market search** to rank on `keywords`. Static HyperLiquid name/category maps remain as fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7c4505da74201c6aa30c9c797ff11cd828cbada. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


michalconsensys:
@metamaskbot publish-previews

michalconsensys:
@metamaskbot publish-previews

packages/perps-controller/src/services/MarketDataService.ts
```diff
@@ -711,20 +721,24 @@ export class MarketDataService {
...
    * @param options.params - The operation parameters.
    * @param options.context - The service context for dependencies.
+   * @param options.useTerminalApi - When true, attempt Terminal API before provider.
    * @returns The result of the operation.
    */
   async getMarkets(options: {
     provider: PerpsProvider;
     params?: GetMarketsParams;
     context: ServiceContext;
+    useTerminalApi?: boolean;
```
abretonc7s: this looks duplicated with params

packages/perps-controller/src/services/MarketDataService.ts
```diff
@@ -51,13 +53,21 @@ import type { ServiceContext } from './ServiceContext';
...
    * Create a new MarketDataService instance
    *
    * @param deps - Platform dependencies for logging, metrics, etc.
+   * @param terminalMarketService - Optional terminal market service for
+   * fetching market data from the Terminal API.
    */
-  constructor(deps: PerpsPlatformDependencies) {
+  constructor(
+    deps: PerpsPlatformDependencies,
+    terminalMarketService?: TerminalMarketService,
```
abretonc7s: can we reuse the deps object?

packages/perps-controller/src/services/TerminalMarketService.ts
```diff
@@ -0,0 +1,257 @@
...
+   * Log a Terminal API error to Sentry without surfacing it to the user.
+   *
+   * @param error - The caught error.
+   * @param method - The calling method name for context.
+   */
+  logError(error: unknown, method: string): void {
+    this.#deps.logger.error(
+      ensureError(error, `TerminalMarketService.${method}`),
+      {
+        tags: { feature: 'perps', source: 'terminal-api' },
```
abretonc7s: shouldnt harcode with magic strings for feature

abretonc7s:
## Automated Review — PR #9137

> **BETA** — Automated review from the farmslot pipeline.

| | |
|---|---|
| **Recommendation** | REQUEST_CHANGES |
| **Runner** | claude / opus |
| **Tier** | standard |
| **Cost** | $1.1965 (2.4M tokens) |
| **Recipe** | N/A |

### Summary

- **B1 — required `terminalApiBaseUrl: string`** (`types/index.ts:1671`) breaks both clients' `PerpsPlatformDependencies` factories on upgrade (mobile `mobileInfrastructure.ts:205`, ext `infrastructure.ts:356`). Path is opt-in via `useTerminalApi`, so make it **optional** → PR becomes non-breaking.
- **B2 — changelog** files the breaking field under *Added* without `**BREAKING:**` (`CHANGELOG.md:16`) → wrong semver signal.

<details>
<summary>Full review details</summary>

## Non-blocking
- **N0 (design) — param creep / dual source of truth.** `useTerminalApi` lives on `params` (`types/index.ts:844,859`) **and** as a separate service option (`MarketDataService.ts:739,865`); controller extracts then re-passes it (`PerpsController.ts:2929,2950,2976,2994`) and the service ignores `params.useTerminalApi`. Read it from `params`, delete the option. Also: `transformMarketData` now takes 5 positional args (`marketDataTransform.ts:191`) → use an options object.
- **N1 — dead/divergent path.** `transformMarketData`'s new `terminalMetadata` param is never passed in prod (`HyperLiquidProvider.ts:6812` calls 4-arg); live enrichment is `MarketDataService.#enrichWithTerminalMetadata:1340`, with different precedence. Drop or wire one.
- **N2** — exported `transformMarketData` (`index.ts:473`) references non-exported `TerminalAssetMetadata`.
- **N3** — `item.marketType as MarketType` (`TerminalMarketService.ts:230`) casts unvalidated string into the enum.
- **N4** — `getMarketDataWithPrices` awaits Terminal then provider (`MarketDataService.ts:888-906`); comment claims "parallel".
- **N5** — `TerminalMarketService` uses global `fetch`/`Date.now()`, no injected fetch / `createServicePolicy` / timeout (`:96,105,126`).
- **N6** — changelog "behind feature flag" is imprecise; Core gates only on caller's `useTerminalApi`.

## Downstream (mobile/ext on `^8.1.0`; used `metamask-mobile-1`, `metamask-extension-1`)
- Breaking: B1 only.
- Additive/safe: `PerpsMarketData` += `keywords/tags/categories`; `useTerminalApi` params; `getMarketMatchRank` `Pick` widened; `transformMarketData` 5th param.
- New dep `@metamask/superstruct` → ext needs LavaMoat policy regen at upgrade.
- No removed/renamed exports, no enum/state/action/ctor changes.

## Validation
- Tests **PASS 110/110**: `jest --no-coverage tests/src/{services/TerminalMarketService,services/MarketDataService,utils/marketDataTransform,utils/marketSearch}.test.ts`.
- Build **not completable (environmental)**: `yarn build` fails only on `TS6305` (unbuilt `account-tree-controller` dist, blocked by pre-existing dep cycle `account-tree-controller↔multichain-account-service↔snap-account-service`) + cascading errors in **untouched** files. No error on a PR-changed line. Defer to CI for authoritative type check.

## Fix order
B1 (optional field, kills B2) → N0 → N1. Rest polish.

</details>

---

*No video evidence recorded.*

abretonc7s (changes requested):

packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -328,6 +328,22 @@ export const DATA_LAKE_API_CONFIG = {
...
+ * Terminal API configuration
+ * Endpoints for fetching structured market metadata from the MetaMask Terminal backend.
+ * The active URL at runtime comes from PerpsPlatformDependencies.terminalApiBaseUrl,
+ * not these constants (they are reference-only for each environment).
+ */
+export const TERMINAL_API_CONFIG = {
+  Endpoints: {
+    dev: 'https://terminal.dev-api.cx.metamask.io',
+    uat: 'https://terminal.uat-api.cx.metamask.io',
+    prd: 'https://terminal.api.cx.metamask.io',
```
aganglada: this should be only passed from the client

packages/perps-controller/src/services/TerminalMarketService.ts
michalconsensys: Fixed in https://github.com/MetaMask/core/pull/9137/commits/c761cb25351da653e949c49ad76ab6891a88ef1d

packages/perps-controller/src/services/MarketDataService.ts
michalconsensys: Fixed in https://github.com/MetaMask/core/pull/9137/commits/c761cb25351da653e949c49ad76ab6891a88ef1d

packages/perps-controller/src/services/MarketDataService.ts
michalconsensys: Fixed in https://github.com/MetaMask/core/pull/9137/commits/c761cb25351da653e949c49ad76ab6891a88ef1d

packages/perps-controller/src/constants/perpsConfig.ts
michalconsensys: Fixed in https://github.com/MetaMask/core/pull/9137/commits/a3242fe34d607c664d68f581cbac1b4b21c07575

packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -328,6 +328,16 @@ export const DATA_LAKE_API_CONFIG = {
   OrdersEndpoint: 'https://perps.api.cx.metamask.io/api/v1/orders',
 } as const;
 
+/**
+ * Terminal API configuration.
+ * The base URL is injected at runtime via PerpsPlatformDependencies.terminalApiBaseUrl
+ * from each client build (dev/uat/prd); only path and cache settings live here.
+ */
+export const TERMINAL_API_CONFIG = {
+  PerpetualPath: '/v1/perpetuals',
```
abretonc7s: if we inject the url why not injected the full url and keeping the path?

packages/perps-controller/src/services/MarketDataService.ts
```diff
@@ -51,13 +53,22 @@ import type { ServiceContext } from './ServiceContext';
...
+
   /**
    * Create a new MarketDataService instance
    *
    * @param deps - Platform dependencies for logging, metrics, etc.
+   * @param deps.terminalMarketService - Optional terminal market service for
+   * fetching market data from the Terminal API.
    */
-  constructor(deps: PerpsPlatformDependencies) {
+  constructor(
```
abretonc7s: shouldnt this be part of the platform dependencies directly instead of extending?

abretonc7s (dismissed):
Re-review after the latest push. The earlier blocking compile break is resolved — `terminalApiBaseUrl` is now optional. Remaining issues below; requesting changes primarily on the filter bypass and changelog accuracy.

**Blocking**

- `MarketDataService.ts:764-785` — the Terminal `getMarkets` path filters only by `params.symbols`, then returns early, bypassing the allowlist/blocklist filtering (and `skipFilters`/`dex`) that the provider path enforces via `provider.getMarkets`. With `useTerminalApi` true, blocklisted markets can leak through. Apply the same filtering on the Terminal path.
- `CHANGELOG.md:16` — still says `PerpsPlatformDependencies` gains a **required** `terminalApiBaseUrl: string`; the field is now optional (`types/index.ts:1678`). Update the wording (the change is non-breaking).
- `perpsTerminalApiMarkets` is never read in Core — `getMarkets`/`getMarketDataWithPrices` gate only on `params.useTerminalApi`. Either wire the remote flag or correct the changelog "behind `perpsTerminalApiMarkets` feature flag" claim.

**Non-blocking**

- `marketDataTransform.ts:194` — the new `terminalMetadata` param is never passed in production (`HyperLiquidProvider.ts:6811` calls it with 4 args); live enrichment runs through `MarketDataService.#enrichWithTerminalMetadata` with different name/marketType precedence. Drop the param or wire it — two divergent enrichment paths.
- `index.ts:473` — exported `transformMarketData` references `TerminalAssetMetadata`, which is not exported.
- `TerminalMarketService.ts:230` — `item.marketType as MarketType` casts an unvalidated string straight into the enum; validate against known values first.
- `MarketDataService.ts:~890` — comment says Terminal metadata is fetched "in parallel" with provider data, but the awaits are sequential.

See also my two inline notes (full-URL injection vs base+path; the `deps & { terminalMarketService }` intersection).

Validation: 110/110 targeted tests pass on the pushed state; package build not completable locally due to a pre-existing dependency cycle (unbuilt `account-tree-controller` dist) — no error maps to a changed line.

packages/perps-controller/src/services/MarketDataService.ts
michalconsensys: https://github.com/MetaMask/core/pull/9137/commits/5cf1fd22785dea943540ab12252595b533bf58fe

packages/perps-controller/src/constants/perpsConfig.ts
michalconsensys: https://github.com/MetaMask/core/pull/9137/commits/5cf1fd22785dea943540ab12252595b533bf58fe

packages/perps-controller/src/services/MarketDataService.ts
```diff
@@ -740,12 +749,65 @@ export class MarketDataService {
...
+            }
+
+            if (context.stateManager) {
+              context.stateManager.update((state) => {
+                state.lastError = null;
+                state.lastUpdateTimestamp = Date.now();
+              });
+            }
+            traceData = { success: true };
+            return filtered;
```
geositta: `getMarkets()` currently falls back only when the Terminal request fails or returns an empty response. The PR description also says static fallback data remains available for assets absent from Terminal, but when Terminal returns a non-empty list that lacks the requested symbol or DEX, the filtered result is returned as [] and the HyperLiquid fallback is skipped.

Please fall back to the provider when a constrained Terminal query (symbols or dex) produces no matches, so Terminal partial coverage does not make provider-backed markets undiscoverable

geositta (changes requested):
Requesting changes. Please preserve provider fallback when Terminal returns a non-empty response that does not contain the requested symbol or DEX, so partial Terminal coverage does not hide valid HyperLiquid markets.

packages/perps-controller/src/services/MarketDataService.ts
michalconsensys: Hello @geositta could you please share more details what you're expecting? Do you suggest we should fallback to HL if there are no results from our BE matching the search?

geositta: @michalconsensys I would expect HL fallback only for symbol- or DEX specific `getMarkets()` lookups, not for every empty search result.

I checked mobile and this flow is possible: token details uses `usePerpsMarketForAsset`, which calls `getMarkets({ symbols: [lookupSymbol], standalone: true })` to decide whether a spot asset has a perps market. If mobile enables `useTerminalApi` on that path and Terminal returns a nonempty list that does not include that symbol yet, the current code returns `[]` and skips HL, so the token would look like it has no perps market even if HL supports it.

For broad `getMarkets({ useTerminalApi: true })`, returning the Terminal list directly makes sense. The fallback that I am suggesting is only when the caller asks for a specific `symbols` or `dex` value and Terminal's filtered result is empty.

geositta (approved):
Approve. Thanks for addressing the `getMarkets()` fallback.

abretonc7s:
## Automated Review — PR #9137

> **BETA** — Automated review from the farmslot pipeline.

| | |
|---|---|
| **Recommendation** | COMMENT |
| **Runner** | claude / opus |
| **Tier** | standard |
| **Cost** | N/A (0.0M tokens) |
| **Recipe** | N/A |

### Summary


No blocking issues. Local correctness is solid, tests are comprehensive, and the change is fully additive and opt-in, so it is backward-compatible for MetaMask Mobile and Extension. A few non-blocking improvements are listed below — chiefly a missing network timeout on the Terminal API fetch and an unconditional `name` override during enrichment.

<details>
<summary>Full review details</summary>

## What the PR does

Adds an optional MetaMask Terminal API market-metadata source, gated per-call by `useTerminalApi` on `GetMarketsParams` / `GetMarketDataWithPricesParams`:

- `getMarkets()` tries Terminal first; on failure, empty response, or a constrained query (symbols/dex) with no Terminal matches, falls back silently to the provider. Terminal results are filtered by the same allowlist/blocklist via a new `PerpsController.#buildMarketAllowedFilter()` predicate.
- `getMarketDataWithPrices()` keeps live pricing from the provider and merges Terminal metadata (`name`, `marketType`, `keywords`, `tags`, `categories`).
- New `TerminalMarketService` with per-item `@metamask/superstruct` validation and a 5-minute in-memory cache.
- `getMarkets({ standalone: true })` now routes through `MarketDataService` instead of calling the provider directly.
- Market search (`getMarketMatchRank`/`rankMarketsByQuery`) now also ranks on `keywords`.

## Local correctness

Implementation reviewed in full (not just diff hunks). Key points verified:

- **Filter parity** — Terminal results in `getMarkets` reuse the provider's allowlist/blocklist via `shouldIncludeMarket` (`MarketDataService.ts:763-267`), then dex/symbol filtering. DEX filter correctly handles main-DEX (`dex: ''` → reject names containing `':'`) vs HIP-3 (`dex: 'xyz'` → require `xyz:` prefix). ✔
- **Fallback semantics** — constrained query (symbols or dex set) with zero Terminal matches falls through to provider; broad query returns Terminal result as-is (`MarketDataService.ts:291-304`). Covered by tests. ✔
- **Enrichment is metadata-only** — `#enrichWithTerminalMetadata` only overrides `name`/`marketType` and attaches `keywords`/`tags`/`categories`; live price/volume/funding from the provider are preserved. Unmatched symbols pass through untouched (`MarketDataService.ts:400-419`). ✔
- **Cache safety** — `fetchMarkets` returns cached arrays/maps by reference, but consumers only `.filter()` (new array) or read via `.get()`; no mutation of cached data. ✔
- **`transformMarketData.ts` refactor** is behavior-preserving (renamed local, identical logic). ✔
- **`marketType` validation** — only values in `MarketCategory` are accepted from Terminal (`TerminalMarketService.ts:654-660`); invalid/empty rejected. ✔
- **Schema validation** — loose `type()` accepts backend extra fields; invalid items are dropped and logged per-item, never failing the whole response (`TerminalMarketService.ts:567-599`). ✔

## Non-blocking findings

1. **[Medium] No timeout/abort on Terminal API fetch** — `TerminalMarketService.ts:527` uses bare `fetch` with no `AbortController`/timeout. In `getMarketDataWithPrices`, the Terminal fetch is awaited *before* the provider call (`MarketDataService.ts:918-932`), so a hung/slow Terminal endpoint delays live price loading on each cache-miss window. The try/catch only helps once the request settles. Recommend an explicit timeout so a stalled Terminal API degrades to the provider promptly.

2. **[Low-Medium] Enrichment can downgrade a display name** — `#enrichWithTerminalMetadata` sets `name: meta.name` unconditionally (`MarketDataService.ts:411`), and `meta.name = item.name ?? item.symbol` (`TerminalMarketService.ts:642`). If Terminal returns a symbol with a null/missing `name` (schema allows `nullable`), a good provider name (e.g. `Bitcoin`) is replaced by the raw symbol (`BTC`). Mobile/Extension render `name` for display. Consider overriding `name` only when Terminal supplies a non-null value.

3. **[Low] Data-service guideline deviation** — per repo `AGENTS.md`, data services should wrap requests in `createServicePolicy` and expose `onRetry`/`onBreak`/`onDegraded`. `TerminalMarketService` does neither (it does validate per-item and throw on non-2xx, which is good). Acceptable given the silent-fallback design, but worth a consistency note. Combined with finding (1), a policy would also supply timeout/retry behavior.

4. **[Low] Standalone `getMarkets` now mutates controller state** — previously `getMarkets({ standalone: true })` returned `provider.getMarkets(params)` with no side effects; it now routes through `MarketDataService.getMarkets`, which on success runs `context.stateManager.update()` setting `lastError=null` / `lastUpdateTimestamp` (`MarketDataService.ts:296-301`). Subtle behavior change for pre-initialization discovery callers; likely intentional for consistency, flagged for awareness.

5. **[Nit] Changelog "Changed" entry is intra-PR** — `CHANGELOG.md:22` ("Replace unsafe `as` type cast with runtime schema validation…") describes a change between commits within this same PR; `TerminalMarketService` is new in this Unreleased cycle, so no released `as` cast ever existed. `AGENTS.md` says to omit non-consumer-facing changes. The "Added" section already documents the validation behavior.

6. **[Nit] Param doc references a non-existent flag** — `useTerminalApi` is documented as "force Terminal API … regardless of remote feature flag" (`types/index.ts:733,741`), but no remote feature flag gating exists in this code; the path is purely caller-controlled. Doc wording could mislead.

## Downstream compatibility (Mobile / Extension)

Canonical clones (`../metamask-mobile`, `../metamask-extension`) are absent; read-only reference clones on `main` were used instead (`../metamask-mobile-ref`, `../metamask-extension-ref`) — 479 and 78 `@metamask/perps-controller` reference lines respectively. See `artifacts/downstream-usage.log`.

**Assessment: backward-compatible, no breaking changes.**

- **No removed/renamed exports.** Only additions to the public surface: `PerpsTerminalMarketService`, `TerminalAssetMetadata` (`index.ts`), optional `terminalApiUrl?` / `terminalMarketService?` on `PerpsPlatformDependencies`, optional `keywords`/`tags`/`categories` on `PerpsMarketData`, optional `useTerminalApi?` on both market params.
- **`getMarketMatchRank` signature widened** from `Pick<PerpsMarketData,'symbol'|'name'>` to `Pick<…,'symbol'|'name'|'keywords'>`. Since `keywords` is optional, existing `{symbol,name}` callers still satisfy it — backward-compatible.
- **Feature is dormant until clients opt in.** Mobile builds deps via `createMobileInfrastructure()`; neither client sets `terminalApiUrl`/`terminalMarketService` today, and no client uses `useTerminalApi`, `getMarketMatchRank`, or `rankMarketsByQuery` directly (grep over both ref clones returned nothing). New behavior cannot trigger without explicit client wiring.
- **Behavior nuance when enabled:** with `useTerminalApi: true`, Terminal-supplied `marketType` overrides the provider value before `applyMarketFilters` (which filters on `marketType`), so category filtering/bucketing results can differ. This is opt-in and by design (Terminal is treated as authoritative), but clients enabling it should validate category UI. Findings (1) and (2) are the concrete risks to watch in client integration.
- **Semver:** minor (`feat`, additive). CHANGELOG records changes under `[Unreleased] → Added` (+ a `Changed` entry, see nit 5). New runtime dep `@metamask/superstruct@^3.1.0` added to `package.json` + `yarn.lock`.

## Validation evidence

All commands run from repo root on branch `feat/perps-terminal-api-markets`.

| Check | Command | Result |
|---|---|---|
| Targeted tests | `yarn workspace @metamask/perps-controller run jest --no-coverage tests/src/services/TerminalMarketService.test.ts tests/src/services/MarketDataService.test.ts tests/src/utils/marketSearch.test.ts tests/src/utils/marketDataTransform.test.ts` | **4 suites, 108 tests passed** |
| Controller test | `… jest --no-coverage tests/src/PerpsController.trading.test.ts` | **1 suite, 15 tests passed** |
| Package build (exports changed) | `yarn workspace @metamask/perps-controller run build` | **Project built successfully** |
| Lint (changed files) | `yarn eslint <8 changed src files>` | **exit 0, no violations** |
| Changelog | `yarn workspace @metamask/perps-controller run changelog:validate` | **exit 0, valid** |

Note: running tests via root `yarn jest <path>` produced a Babel parse error (wrong transform for TS inline-type imports); use the workspace jest config as above.

</details>

---

*No video evidence recorded.*

packages/perps-controller/src/services/MarketDataService.ts
michalconsensys: https://github.com/MetaMask/core/pull/9137/commits/c7c4505da74201c6aa30c9c797ff11cd828cbada

PR #9148, opened 2026-06-16
fix(transaction-controller): set isExternalSign when isGasFeeSponsored is confirmed by simulation

OGPoyraz:
## Summary

When `isGasFeeSponsored` is confirmed `true` by the simulation response, `isExternalSign` is now also set to `true` on the transaction meta.

## Fix

Accounts using the Money Account keyring cannot execute `KeyringController:signTransaction` — the keyring only supports EIP-7702 authorization and message signing. Transactions from these accounts that are gas-sponsored (e.g. the Card link ERC-20 approval on Monad) would crash with:

```
KeyringController - The keyring for the current address does not support the method signTransaction.
```

Setting `isExternalSign = true` alongside `isGasFeeSponsored = true` skips the local signing step, allowing Sentinel Relay to handle submission via the sponsorship path.

https://github.com/user-attachments/assets/37e22c82-0724-4204-ac4e-6d5b45998d7a

## Changes

- `TransactionController`: after persisting `isGasFeeSponsored` from simulation, set `isExternalSign = true` when sponsored

### DIFF
--- a/packages/transaction-controller/src/TransactionController.ts
+++ b/packages/transaction-controller/src/TransactionController.ts
@@ -4078,6 +4078,11 @@ export class TransactionController extends BaseController<
       (txMeta) => {
         txMeta.gasFeeTokens = gasFeeTokens;
         txMeta.isGasFeeSponsored = isGasFeeSponsored;
+
+        if (isGasFeeSponsored) {
+          txMeta.isExternalSign = true;
+        }
+
         txMeta.gasUsed = gasUsed;
 
         if (!this.#isBalanceChangesSkipped(txMeta)) {
--- a/packages/transaction-controller/src/TransactionController.ts
+++ b/packages/transaction-controller/src/TransactionController.ts
@@ -1129,6 +1129,7 @@ export class TransactionController extends BaseController<
       isGasFeeTokenIgnoredIfBalance,
       isGasFeeIncluded,
       isGasFeeSponsored,
+      ...(isGasFeeSponsored ? { isExternalSign: true } : {}),
       // To avoid the property to be set as undefined.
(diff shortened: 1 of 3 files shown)


OGPoyraz:
@metamaskbot publish-preview

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -4078,6 +4078,11 @@ export class TransactionController extends BaseController<
       (txMeta) => {
         txMeta.gasFeeTokens = gasFeeTokens;
         txMeta.isGasFeeSponsored = isGasFeeSponsored;
+
+        if (isGasFeeSponsored) {
```
matthewwalsh0: What if `isGasFeeSponsored` is passed in transaction or batch options?

What if simulation fails?

Should we also set when adding transaction?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/TransactionController.ts
matthewwalsh0: Should we ignore simulation result if passed in as transaction option? To force the gasless route?

OGPoyraz: I will add it through `addTransaction` options. I think it make sense to preserve it it in case simulation fails.

Adding fix for both soon

OGPoyraz: Done herE: https://github.com/MetaMask/core/pull/9148/commits/6e8cbc0b069016cc4c5a0e57a43ce168c32d3618

OGPoyraz:
@metamaskbot publish-preview

matthewwalsh0 (approved):

PR #9150, opened 2026-06-16
fix: MM Pay transaction with isQuoteRequired that have same source and destination chain and token

jpuri:
## Explanation

Fix MM Pay transaction with isQuoteRequired that have same source and destination chain and token

## References
Related to https://consensyssoftware.atlassian.net/browse/CONF-1548

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches MM Pay quote eligibility and Relay request identity for accountOverride post-quote paths; scoped logic with tests, but incorrect user/source handling could break pay execution.
> 
> **Overview**
> Fixes MM Pay **post-quote** flows where source and destination are the **same chain and token** but **`isQuoteRequired`** still needs a Relay quote (previously those legs were dropped or quoted with the wrong `user`).
> 
> **Source amounts:** Post-quote `calculatePostQuoteSourceAmounts` no longer skips identical source/target tokens when `isQuoteRequired` is set, so quote retrieval can run instead of ending with empty `sourceAmounts`.
> 
> **Relay quotes:** Quote bodies now set `user` via **`getQuoteUser`**. For same-chain/same-token with an active **`accountOverride`** and recipient equal to the override wallet, **`user`** is **`txParams.from`** (the real sender); external recipients or cross-chain/token cases keep using the resolved `from` address.
> 
> Changelog and unit tests cover both behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35456e2b9e92d87bb7fc6de82551068d04f4f5e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vinistevam (dismissed):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -1181,6 +1183,41 @@ function getTransferRecipient(data: Hex): Hex {
...
+
+  const isSameSourceAndTarget =
+    sourceChainId === targetChainId &&
+    sourceTokenAddress.toLowerCase() === targetTokenAddress.toLowerCase();
+
+  const txParamsFrom = transaction.txParams?.from as Hex | undefined;
+  const hasAccountOverride =
+    txParamsFrom && from.toLowerCase() !== txParamsFrom.toLowerCase();
+
+  return isSameSourceAndTarget && hasAccountOverride ? txParamsFrom : from;
```
matthewwalsh0: Is the condition also that the recipient matches the account override?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
jpuri: Good point, I updated PR to include the condition.

matthewwalsh0 (approved):

PR #9164, opened 2026-06-17
refactor: Use permission decoders from `@metamask/7715-permission-types`

jeffsmale90:
## Explanation

https://github.com/MetaMask/smart-accounts-kit/pull/259 adds permission decoder definitions to @metamask/7715-permission-types released in 0.8.0.

This PR updates @metamask/gator-permissions-controller to consume these permission definitions instead of defining the permission decoders directly in the controller.

Note: `erc20-token-revocation` permission type is no longer supported.

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking removal of a permission type and relocation of delegation decode logic to an external package affects how permissions are interpreted; misconfiguration or version skew could mis-decode or reject delegations.
> 
> **Overview**
> Moves Gator permission decoding onto **`@metamask/7715-permission-types` ^1.0.0** by building decoders from `makePermissionDecoderConfigs` instead of local per-type decoder modules (stream/periodic/allowance, payee/expiry/redeemer rules, etc.). Chain enforcer addresses are mapped through new **`toEnforcerAddressesByName`** / **`delegationContractsByChainId`** before decode and status sync.
> 
> **Breaking:** **`erc20-token-revocation`** is removed from default `supportedPermissionTypes`, README examples, and tests. Decode failures from missing contracts or checksum errors are wrapped as **`PermissionDecodingError`** with the underlying cause preserved.
> 
> Tests shift to controller-level decode coverage for supported types (including allowance and **`token-approval-revocation`**) and slimmer unit tests around decoder selection; large integration-style decoder test suites in this package are dropped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ba2af84d8240b929410a33e077ed12de5d818bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/gator-permissions-controller/package.json
```diff
@@ -53,7 +53,7 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
   "dependencies": {
-    "@metamask/7715-permission-types": "^0.7.1",
+    "@metamask/7715-permission-types": "^0.8.0",
```
mj-kiwi: Should we pump the version to 0.9.0?

packages/gator-permissions-controller/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
```
mj-kiwi, line 11: Small suggestion — the Unreleased entries currently just say "Changed" for switching to package decoders / bumping the dependency, but don't mention that `erc20-token-revocation` support seems to have been dropped along the way (see the comment on types.ts). If that drop is intentional, might be worth calling it out explicitly (possibly with a **BREAKING:** prefix per our changelog convention), so consumers relying on that permission type aren't caught off guard after upgrading.

packages/gator-permissions-controller/README.md
```diff
@@ -26,7 +26,6 @@ const gatorPermissionsController = new GatorPermissionsController({
       'native-token-periodic',
```
mj-kiwi, line 26: Minor doc nit — the example `supportedPermissionTypes` array dropped `erc20-token-revocation` but doesn't seem to have added `token-approval-revocation` in its place, even though that type still looks supported. Might just need one line added back so the example stays accurate.

packages/gator-permissions-controller/src/decodePermission/utils.ts
```diff
@@ -1,11 +1,9 @@
+import type { DeployedContractsByName } from '@metamask/7715-permission-types';
```
mj-kiwi: Noticed `DeployedContractsByName` moved from the precise local alias `(typeof DELEGATOR_CONTRACTS)[number][number]` to the package's `Record<string, Hex>`. Not a big deal, but it does mean a typo'd contract name would only surface at runtime (via the `Contract not found` throw) rather than at compile time like before. Wanted to flag in case that tradeoff wasn't deliberate.

packages/gator-permissions-controller/src/decodePermission/types.ts
```diff
@@ -9,12 +9,8 @@ import type {
   ApprovalRevocationTerms,
   Caveat,
```
mj-kiwi: While in here — it looks like `Erc20TokenAllowancePermission`, `NativeTokenAllowancePermission`, and `TokenApprovalRevocationPermission` (lines 23, 40, 53 below) are still declared locally with a "not yet defined in the package" comment, but `@metamask/7715-permission-types@^0.8.0` seems to export all three now (worth double-checking, since the local `TokenApprovalRevocationPermission` shape looks slightly different from the package's version). If the package really does cover these now, importing them directly would avoid future drift.

packages/gator-permissions-controller/src/decodePermission/decoders/index.ts
```diff
@@ -1,14 +1,8 @@
-import type { DeployedContractsByName, PermissionDecoder } from '../types';
-import { getChecksumEnforcersByChainId } from '../utils';
-import { makeErc20TokenAllowanceDecoderConfig } from './erc20TokenAllowance';
-import { makeErc20TokenPeriodicDecoderConfig } from './erc20TokenPeriodic';
-import { makeErc20TokenRevocationDecoderConfig } from './erc20TokenRevocation';
-import { makeErc20TokenStreamDecoderConfig } from './erc20TokenStream';
+import { makePermissionDecoderConfigs } from '@metamask/7715-permission-types';
```
mj-kiwi, line 1: Just flagging for awareness — with per-type validation logic now living in `@metamask/7715-permission-types`, this package's tests only exercise a synthetic `specified-permission-type` rather than the real permission types end-to-end. Might be worth a lightweight integration test here (or confirming coverage exists upstream) so a future bump of the dependency that loosens a validation rule wouldn't slip by unnoticed.

packages/gator-permissions-controller/src/decodePermission/decoders/makePermissionDecoder.ts
```diff
@@ -1,45 +1,22 @@
-import type { Rule } from '@metamask/7715-permission-types';
+import type {
```
mj-kiwi, line 1: Small one — enforcer addresses look like they get checksummed once for matching, then re-checksummed again per candidate decoder inside `validateAndDecodePermission` on the ambiguous-type path. Probably not costly today, but could be worth reusing the already-checksummed set if this path gets exercised more.

packages/gator-permissions-controller/src/decodePermission/types.ts
```diff
@@ -9,12 +9,8 @@ import type {
   ApprovalRevocationTerms,
```
mj-kiwi: `RuleDecoder` looks like it's only referenced from a test file now, not from any production decoder — the real shape seems to live in the package's `PermissionDecoderConfig`. If that's right, might be simpler to move it into the test file (or drop it) so it doesn't read as something production code still relies on.

packages/gator-permissions-controller/src/decodePermission/utils.ts
```diff
@@ -1,11 +1,9 @@
...
 import type { Hex } from '@metamask/utils';
 
-import type {
-  ChecksumEnforcersByChainId,
-  DeployedContractsByName,
-} from './types';
+import type { ChecksumEnforcersByChainId } from './types';
 
 /**
  * The names of the enforcer contracts for each permission type.
```
mj-kiwi: Nit — `ENFORCER_CONTRACT_NAMES` maps each key to an identical string value, so `getChecksumEnforcersByChainId` ends up repeating the same 3-line pattern 12 times. Could probably just use the string literals directly and drop the map, but totally optional.

mj-kiwi (commented):
Solid refactor, just a few minor notes.

packages/gator-permissions-controller/src/decodePermission/types.ts
jeffsmale90: Good call - I've removed the 3 locally defined permission types, along with `ExtendedPermissionTypes`

packages/gator-permissions-controller/README.md
jeffsmale90: I don't think this example needs to include _all_ implemented permission types - this is showing how the controller may be instantiated with a list of permission types that should be supported.

This list can be incomplete - it's actually highlighting the capability to restrict supported permissions, but also we probably don't want to update this every time we add permission types to the dependency.

packages/gator-permissions-controller/package.json
jeffsmale90: 1.0.0 even!

packages/gator-permissions-controller/CHANGELOG.md
jeffsmale90: I've added:
```
  - **BREAKING:** Drop support for `erc20-token-revocation` permission type
```
to the changelog :)

packages/gator-permissions-controller/src/decodePermission/utils.ts
jeffsmale90: I've improved @metamask/7715-permission-types in https://github.com/MetaMask/smart-accounts-kit/pull/278 to define `EnforcerAddressesByName` as the API type for contract addresses (keyed contract names, with camelCase). I've updated this PR to assert the type of the @metamask/delegation-deployments contract addresses (exported as `Record<string, Hex>`) to `DelegationDeploymentsEnforcerAddressesByName` (keyed contract names, with PascalCase) and converts them to `EnforcerAddressesByName` in `toEnforcerAddressesByName`.

packages/gator-permissions-controller/src/decodePermission/decoders/makePermissionDecoder.ts
jeffsmale90: This is still the case.

My concern here is that it's difficult to ensure that addresses are checksummed (at the API interface) without checksumming them 😓  - we can't type them in such a way to enforce this (unless we switch to byteslike types where case mismatching isn't possible 🤔). So it's a balance between tight coupling, and more checksumming.

Because this only ever happens _once_ per permission, and the number of candidates that pass the `caveatAddressesMatch` check is normally going to be only 1, I think for now it's a low cost.

I'm going to add a comment explaining it though - let me know if you think that we should actually address this now though.

packages/gator-permissions-controller/src/decodePermission/types.ts
jeffsmale90: Moved it into the test file. It does give us a little utility within the tests.

packages/gator-permissions-controller/src/GatorPermissionsController.ts
```diff
@@ -584,12 +583,14 @@ export class GatorPermissionsController extends BaseController<
...
 
-    const contracts = contractsByChainId[chainId];
+    const deploymentContracts = delegationContractsByChainId[chainId];
 
-    if (!contracts) {
+    if (!deploymentContracts) {
       throw new Error(`Contracts not found for chainId: ${chainId}`);
     }
 
+    const contracts = toEnforcerAddressesByName(deploymentContracts);
```
mj-kiwi: `toEnforcerAddressesByName(deploymentContracts)` is now called before the `try` block in `decodePermissionFromPermissionContextForOrigin`. This means a "Contract not found" error would escape unwrapped instead of being
  caught and rethrown as `PermissionDecodingError`, which the method's `@throws` doc promises. Could we move this call back inside the `try` block, similar to how `resolveGrantedPermissionOnChainStatus` in `permissionOnChainStatus.ts` handles it?

mj-kiwi (dismissed):
LGTM

packages/gator-permissions-controller/src/GatorPermissionsController.ts
jeffsmale90: Good call!

mj-kiwi (approved):
LGTM

PR #9165, opened 2026-06-17
feat(social-controllers): expose optional 7-day per-chain breakdown

joaosantos15:
## Explanation

The Social Leaderboard surfaces in MetaMask Mobile are switching their PnL displays from the 30-day window to the 7-day window (TSA-766). The trader-profile "Return" headline sums the **per-chain** PnL breakdown (rather than the global stat) so it includes Hyperliquid/perps and stays consistent with the position list. `TraderStats` already exposed scalar 7-day fields (`pnl7d`, `winRate7d`, `roiPercent7d`), but `PerChainBreakdown` only had the 30-day per-chain maps — so a 7-day, Hyperliquid-inclusive headline wasn't expressible.

This adds the 7-day per-chain breakdown to `PerChainBreakdown` (and its struct), alongside the matching social-api change that populates it (Clicker already returns these windows).

## References

- Mobile ticket: TSA-766 (Use 7d PnL)
- social-api counterpart: `va-mmcx-social-api` — exposes `perChainPnl7d` / `perChainRoi7d` / `perChainVolume7d` on the profile response and chain-scopes the leaderboard 7-day fields.

## Changes

- **Added**: optional `perChainPnl7d` (`Record<string, number>`), `perChainRoi7d` (`Record<string, number | null>`), and `perChainVolume7d` (`Record<string, number>`) to the `PerChainBreakdown` type and `PerChainBreakdownStruct`.
- The unsuffixed fields (`perChainPnl`, `perChainRoi`, `perChainVolume`) remain the **30-day** window.
- New fields are **optional** so the controller stays backward compatible with social-api versions that only return the 30-day breakdown.

## Non-breaking

Purely additive. The response struct uses `superstruct`'s non-exhaustive `type()`, and the new struct fields are `optional()`, so existing 30-day-only responses continue to validate unchanged. No public export was removed, renamed, or had its signature changed.

## Test plan

- `yarn workspace @metamask/social-controllers run jest --no-coverage SocialService.test.ts` — added cases asserting the 7-day per-chain breakdown round-trips, and that a profile omitting it still validates (`perChainPnl7d` is `undefined`). 45/45 pass.
- `yarn workspace @metamask/social-controllers run build` — type check passes.
- Changelog updated + `changelog:validate` passes.

### DIFF
--- a/packages/social-controllers/CHANGELOG.md
+++ b/packages/social-controllers/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add optional 7-day per-chain fields to the `PerChainBreakdown` type (and `PerChainBreakdownStruct`): `perChainPnl7d` (`Record<string, number>`), `perChainRoi7d` (`Record<string, number | null>`), and `perChainVolume7d` (`Record<string, number>`) — exposing the 7-day Hyperliquid/per-chain breakdown alongside the existing 30-day fields. The unsuffixed fields (`perChainPnl`, `perChainRoi`, `perChainVolume`) remain the 30-day window; the new fields are optional for backward compatibility with social-api versions that only return the 30-day breakdown ([#9999](https://github.com/MetaMask/core/pull/9999))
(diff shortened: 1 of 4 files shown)


joaosantos15:
@metamaskbot publish-previews

xavier-brochard (approved):

PR #9189, opened 2026-06-18
fix(assets-controller): Dedup assets price requests

Kriys94:
### Problem

`PriceDataSource` fetches spot prices from multiple, independently-triggered
paths: enrichment middleware (on balance/metadata updates), the per-subscription
polling loop, currency changes, and manual refreshes. These triggers frequently
overlap, so the same asset's price was requested from the Price API multiple
times within a few seconds, producing redundant network traffic and load.

### Solution

This PR introduces a small, reusable `DedupingBatchFetcher` utility and routes
all `PriceDataSource` price fetches through it. The fetcher deduplicates by key
across two dimensions:

- **Freshness TTL** — a key fetched within `freshnessTtlMs` is skipped entirely
  (the absence of a value is also cached, so we don't re-ask for assets the API
  had no price for until the TTL expires).
- **In-flight coalescing** — concurrent callers requesting a key that already has
  a fetch in progress join the existing promise instead of issuing a new request.

Stale, not-in-flight keys from a single call are batched into one `fetchBatch`
invocation and then split back into per-key results so each key resolves
independently.

Wiring in `PriceDataSource`:

- New `priceFreshnessTtlMs` option on `PriceDataSourceConfig` (defaults to the
  poll interval) controls how long a fetched price is considered fresh.
- New `PriceDataSource.invalidatePriceCache()` clears the freshness cache;
  `AssetsController` calls it when the selected currency changes, since cached
  prices are denominated in the previous currency and must be refreshed.

### Non-obvious detail: freshness TTL must stay *below* the poll interval

The subscription poll fires on a fixed `setInterval(pollInterval)`, but a key's
`fetchedAt` is stamped when the fetch *completes* — slightly after the tick that
triggered it. If the TTL equals the poll interval, then at the next tick
`now - fetchedAt` is `pollInterval - latency`, which is still within the TTL, so
the asset reads as "fresh" and the poll is skipped. The net effect is the
subscription only issues a real request every *other* poll (e.g. an observed
~2-minute cadence for a 1-minute poll interval).

To fix this, the TTL is capped at `0.9 * pollInterval` (via
`FRESHNESS_TTL_POLL_RATIO`) rather than the previous `Math.min(ttl, pollInterval)`,
giving enough headroom to absorb network latency and timer jitter. A regression
test reproduces the every-other-poll skip by simulating fetch latency with
`jest.setSystemTime` (it fails with the old cap and passes with the new one).

### Also in this PR

- `AssetsController` no longer triggers a redundant one-time balance fetch when
  the enabled-network list changes; subscription refresh and the
  enabled-networks handler already cover those fetches.

## References

- Related to the assets-controller price-fetch deduplication effort.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core price-fetch timing and caching; incorrect TTL or invalidation could delay price updates or leave stale currency-denominated prices, though behavior is heavily covered by new tests.
> 
> **Overview**
> Introduces **`DedupingBatchFetcher`** and routes all **`PriceDataSource`** spot-price fetches through it so middleware, subscription polls, currency changes, and manual refreshes no longer hammer the Price API for the same asset.
> 
> **Freshness TTL** (`priceFreshnessTtlMs`, default poll interval) skips re-fetching recently completed prices; **in-flight coalescing** joins parallel callers on one batch. On subscribe, TTL is capped at **0.9 × poll interval** so completion latency does not make every other poll a no-op.
> 
> Adds **`invalidatePriceCache()`**; **`AssetsController`** calls it on selected-currency change before the price-only **`getAssets`** refresh.
> 
> **`handleActiveChainsUpdate`** drops the extra one-time **`getAssets`** when chains are added; it only re-subscribes (tests updated accordingly).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8633f7f2a6f709f62bc2aaae78357b1d05cfe5b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `priceFreshnessTtlMs` option to `PriceDataSourceConfig` controlling how long a fetched price is considered fresh before it is re-fetched (defaults to the poll interval, 60 000 ms) ([#9189](https://github.com/MetaMask/core/pull/9189))
+- Add `PriceDataSource.invalidatePriceCache()` to force the next fetch to bypass the freshness cache; `AssetsController` now calls it when the selected currency changes so cached prices in the previous currency are refreshed ([#9189](https://github.com/MetaMask/core/pull/9189))
+
+### Fixed
+
+- Deduplicate price fetches in `PriceDataSource` so overlapping triggers (enrichment middleware, subscription polls, and manual refreshes) no longer issue duplicate price API requests for the same asset; assets fetched within the freshness TTL are skipped and concurrent in-flight fetches are joined ([#9189](https://github.com/MetaMask/core/pull/9189))
+- Remove a redundant one-time balance fetch in `AssetsController` when the enabled network list changes, since subscription refresh and the enabled-networks handler already cover those fetches ([#9189](https://github.com/MetaMask/core/pull/9189))
+
 ## [9.0.2]
 
 ### Changed
--- a/packages/assets-controller/src/AssetsController.test.ts
+++ b/packages/assets-controller/src/AssetsController.test.ts
@@ -1267,45 +1267,48 @@ describe('AssetsController', () => {
   });
 
   describe('handleActiveChainsUpdate', () => {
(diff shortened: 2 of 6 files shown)


salimtb:
@metamaskbot publish-preview

salimtb (approved):

PR #9196, opened 2026-06-18
fix: ARC native to point to ERC20 + format CAIP ERC20

maxime-oe:
## Explanation

**Context:**

- On Arc, there are two USDC:
  - Native one, 18 decimals.
  - ERC20 one, 6 decimals.

After some back-and-forth, the deduplication approach is now the following:
- For Asset list / Send features: We show the native USDC.
- For Bridge/Swap: We show the ERC20 USDC.

For Swap/Bridge the "quick approach" taken was to:
- Declare the actual native USDC as native in `packages/bridge-controller/src/constants/tokens.ts`.
- But make changes in the UI to ensure that USDC native is never actually presented to the user.
This is because setting USDC to the ERC20 version in the `bridge-controller` led to unexpected behavior that we didn't have the bandwidth to address or to de-risk. For example, this was causing the Bridge API to be called with the `zeroAddress` and never the ERC20 address one - which we'd want.

**New approach goals:** 

The goal now is to make in sort that it becomes possible to properly set an ERC20 as "native" token through the `bridge-controller`, making acceptable adjustments in the logic.
Goals:
- Solve remaining issues with Swap/Bridge on ARC: Most importantly the "max swap" behavior (since the current approach doesn't consider USDC as native, which leaves "max" button visible).
- Remove as many "quick fixes" as possible from the codebase.

**Changes:**

It looks like one of the important changes would be in `formatAddressToCaipReference` to treat `erc20:0x3600000000000000000000000000000000000000` as a ERC20 address.
Previous logic of the function:
- If address matches what is defined in `SYMBOL_TO_SLIP44_MAP`, overwrite to `zeroAddress` --> This allows to have formats like `slip44:60` to be converted in a ERC20 address but the problem is that even `erc20:0x3600000000000000000000000000000000000000` gets taken by that part of the logic and is turned to the `zeroAddress` before asking for a quote to the Bridge API.
New logic of the function:
- Check first if the part after the `:` is an ERC20 address, if yes use that. Otherwise proceed with the previous logic of the function. This way, when asking for a quote on Arc, `srcToken` will be set to `0x3600000000000000000000000000000000000000` instead of `0x0000000000000000000000000000000000000000`.

If accepted, QA and non-regression on existing chains would be done to ensure this doesn't break anything.
Also open to other ways to rewrite this function.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core token identity, native detection, and Bridge API address formatting across all chains; incorrect `isNativeAddress` or CAIP parsing could mis-route quotes or balances outside Arc.
> 
> **Overview**
> **Arc swap/bridge** now treats the chain default as **USDC at `0x3600…`** (6 decimals) instead of a hidden zero-address “USDC-native,” and maps it in `SYMBOL_TO_SLIP44_MAP` as `erc20:0x3600…`.
> 
> **Address handling** is updated so Arc’s “native” USDC is not collapsed to `0x000…` when building Bridge API requests or exchange-rate lookups: `formatAddressToCaipReference` resolves a trailing hex segment (including CAIP-19 `erc20:…`) before the native-token path, and `isNativeAddress` matches SLIP44/erc20 references on the asset suffix. **Balances** use on-chain ETH balance only for literal `AddressZero`; Arc USDC uses ERC20 `balanceOf`.
> 
> `selectExchangeRateByAssetId` catches failures from `formatAddressToCaipReference` instead of relying on the removed try/catch in the EVM rate helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c09cc9479848d577a26d42c2c5c95af413f34a0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bridge-controller/CHANGELOG.md
+++ b/packages/bridge-controller/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add optional `environment_type` property to the `ButtonClicked` unified swap/bridge event context ([#9121](https://github.com/MetaMask/core/pull/9121))
+- Add support for ERC20 as native token, starting with Arc USDC ([#9196](https://github.com/MetaMask/core/pull/9196))
 
 ### Changed
 
--- a/packages/bridge-controller/src/constants/tokens.ts
+++ b/packages/bridge-controller/src/constants/tokens.ts
@@ -192,10 +192,10 @@ const MEGAETH_SWAPS_TOKEN_OBJECT = {
 
 // Leaving for code consistency but we won't display it in the asset picker
 const ARC_SWAPS_TOKEN_OBJECT = {
-  symbol: 'USDC-native',
-  name: 'USDC-native',
-  address: '0x0000000000000000000000000000000000000000',
-  decimals: 18,
+  symbol: 'USDC',
+  name: 'USDC',
+  address: '0x3600000000000000000000000000000000000000',
+  decimals: 6,
   iconUrl: '',
 } as const;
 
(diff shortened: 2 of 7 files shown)


maxime-oe:
@metamaskbot publish-preview

maxime-oe:
Deprecated

PR #9226, opened 2026-06-23
feat(wallet-cli): add wallet factory and daemon entry point

### LINKED ISSUE #8779: wallet-cli: Replace { wallet, store } with { wallet, dispose }

sirtimid:
## Explanation

Adds the wallet factory and daemon entry point for `@metamask/wallet-cli`, after the scaffold (#9065), persistence (#9067), and transport (#9108) slices:

- **`wallet-factory.ts`** — `createWallet()` builds a `@metamask/wallet` `Wallet` over the SQLite `KeyValueStore`, hydrates it from persisted state, imports the SRP on first run, and returns `{ wallet, dispose }`. `dispose` (idempotent, resilient) runs unsubscribe → `wallet.destroy()` → `store.close()`.
- **`daemon-entry.ts`** — the daemon process: claims the PID slot, builds the wallet, serves the messenger over the owner-only Unix socket, and tears down via `dispose` on shutdown.

It uses the current `@metamask/wallet` `instanceOptions` API — `storageService` (in-memory adapter), `connectivityController` (`AlwaysOnlineAdapter`), and `remoteFeatureFlagController`. `NetworkController` and `TransactionController` slots are left as TODOs until they're wired on `main`.

Supersedes draft #8847 (the `{ wallet, dispose }` refactor), folded in here on top of `main`.

Closes #8779.

## Checklist

- [x] `build`, `test` (100% coverage on both new files), `lint`, `changelog:validate`, and `constraints` pass.

---

> [!NOTE]
> **High Risk**
> Touches keyring/SRP import, encrypted vault persistence, and exposes the full wallet messenger over a Unix socket with only filesystem permissions as the access boundary.
> 
> **Overview**
> Adds **`createWallet`** and **`daemon-entry`** so the CLI can run a long-lived background wallet process backed by SQLite persistence.
> 
> **`createWallet`** opens a `KeyValueStore`, hydrates controller state (via a short-lived metadata probe `Wallet`), wires headless `instanceOptions` (in-memory `storageService`, `AlwaysOnlineAdapter`, Infura + remote feature flags), subscribes persistence, runs **`wallet.init()`** (startup aborts if any step fails), imports the SRP on first run when no vault exists, and returns **`{ wallet, dispose }`** with ordered, idempotent teardown. First-run failures can delete on-disk DB files.
> 
> **`daemon-entry`** validates required env vars, **removes password/SRP from `process.env`** after capture, locks the data dir to **`0o700`**, runs **`claimDaemonSlot`** (stale socket/PID cleanup, refuse live siblings), claims the daemon with an **exclusive PID write before opening the DB**, builds the wallet, and serves **`getStatus`** plus a **`call`** RPC that forwards arbitrary messenger actions. Shutdown/SIG handlers coalesce teardown and only remove PID files the process owns.
> 
> Also adds **`remote-feature-flag-controller`** and **`storage-service`** dependencies and a small **`Readonly`** typing tweak in persistence subscriptions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 180ffc15dc64c7e290d6aa42b6dba2ed57671b2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet-cli/src/daemon/daemon-entry.ts
```diff
@@ -0,0 +1,299 @@
...
+    log(`Removing stale socket at ${socketPath} (${ping.error.message}).`);
+  }
+  // Always clear both files before claiming the slot. The PID file may be
+  // corrupt (truncated, partial write from a crashed run); without this, the
+  // exclusive `wx` write below would fail with EEXIST and the daemon could
+  // not start until a human manually deleted the file.
+  await Promise.all([
+    rm(pidPath, { force: true }),
+    rm(socketPath, { force: true }),
+  ]);
```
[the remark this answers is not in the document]
sirtimid: It's technically correct, but narrow in practice. Will defer this as a tracked daemon-robustness follow-up when the rest of the work for the daemon land.

packages/wallet-cli/src/daemon/wallet-factory.ts
```diff
@@ -0,0 +1,266 @@
...
+    wasFirstRun = !hasPersistedKeyring(state);
+
+    wallet = new Wallet({ state, instanceOptions: buildInstanceOptions() });
+    // `wallet.messenger` is typed `Readonly`, but persistence must register
+    // (and later remove) subscriptions on it.
+    unsubscribe = subscribeToChanges(
+      wallet.messenger as RootMessenger<DefaultActions, DefaultEvents>,
+      wallet.controllerMetadata,
+      store,
+      log,
```
grypez:
```suggestion
      log: logFn,
```
I think you meant this?

packages/wallet-cli/src/daemon/daemon-entry.ts
```diff
@@ -0,0 +1,299 @@
...
+    throw new Error('MM_WALLET_SRP environment variable is required');
+  }
+
+  // 0o700: owner-only. The daemon exposes the full wallet messenger over
+  // the socket inside this directory, so anyone who can traverse the dir
+  // can also `connect()` to the socket. Restricting to the owning user is
+  // the only access-control boundary. We chmod after mkdir because the
+  // `mode` option is ignored when the directory already exists.
+  mkdirSync(dataDir, { recursive: true, mode: 0o700 });
+  await chmod(dataDir, 0o700);
```
grypez: Awesome, this addresses the comment I raised in your last PR.

The associated unit test `daemon-entry > creates data dir, wallet, server, and writes PID exclusively on successful startup` asserts that `mkdirSync` is called with `mode: 0o700`, while this comment states this is not sufficient. That conflict makes this a good import boundary. Recommend moving these lines to a new function in `data-dir.ts`, with the comment in the jsdoc.

e.g.

```suggestion
  await ensureOwnerOnlyDirectory(dataDir);
```

and have the test assert that `ensureOwnerOnlyDirectory` was called.

grypez (changes requested):

packages/wallet-cli/src/daemon/wallet-factory.ts
sirtimid: Yep — good catch. Switched to `logFn` so the resolved logger (the `console.error` fallback when no `log` is supplied) is always forwarded, instead of the raw optional `log`. Done in 7854c96.

packages/wallet-cli/src/daemon/daemon-entry.ts
sirtimid: Done in 7854c96. Extracted into `ensureOwnerOnlyDirectory(dataDir)` in a new `data-dir.ts`, with the owner-only rationale moved into its JSDoc. The daemon-entry test now mocks `./data-dir` and asserts `ensureOwnerOnlyDirectory` was called with the data dir, and `data-dir.test.ts` covers the `mkdir` + `chmod` behavior directly.

grypez (approved):

PR #9238, opened 2026-06-23
feat: reserve HyperLiquid activation fee for unactivated Pay withdrawals

dan437:
## Explanation

MetaMask Pay "Perps Withdraw" (HyperLiquid → any token) fails for some users with `Insufficient USDC balance for token transfer gas` (and the sibling `Insufficient balance for token transfer`). HyperLiquid charges a one-time ~1 USDC activation fee on an account's first outbound send (the `sendAsset` step Relay uses), so a **Max** withdraw — which submits the full withdrawable amount with no headroom — is rejected. This is independent of withdrawal size and accounts for the majority of the observed failures.

This change handles the fee on our side (a temporary measure while Relay/HyperLiquid evaluate doing it at quote time), **opt-in behind a feature flag and disabled by default**:

- When the source is HyperLiquid (`isHyperliquidSource`) and the feature is enabled, the account's activation is checked via HyperLiquid's `userNonFundingLedgerUpdates` info endpoint — a prior outbound send/withdrawal by the account means it is already activated.
- If **not** activated, the activation fee is reserved from the amount sent to HyperLiquid (so the `sendAsset` step keeps enough balance to cover it) and surfaced as part of the **provider fee**. The displayed withdrawal amount is unchanged — mirroring how the deposit fee is presented (no separate UI line, no UX change).
- If already activated — or if the HyperLiquid info request errors — no reserve is applied, so the common path is never penalised by a transient outage.

Resolution / non-obvious bits:

- Feature flag `confirmations_pay_post_quote`, resolved from `overrides.<transactionType>.hyperliquidActivationFee` and falling back to `default.hyperliquidActivationFee`; shape `{ enabled?: boolean; amountUsd?: number }` with `amountUsd` defaulting to `1`.
- The reserved fee is always surfaced in the provider fee, including when the relay provider fee is subsidized to zero, since it is withheld from the source send regardless.

## References

- Follow-up to the MetaMask Pay Perps Withdraw work; adopted by metamask-extension and metamask-mobile once they bump `@metamask/transaction-pay-controller`.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes quoted amounts and fee breakdown for HyperLiquid-source withdrawals when the flag is enabled; mis-activation detection or fee math could cause failed sends or wrong fees, though the flag defaults off and API errors fail open to prior behavior.
> 
> **Overview**
> Adds **opt-in** handling for HyperLiquid’s one-time ~$1 USDC activation fee on Pay **Perps Withdraw** flows when HyperCore is the source, behind `confirmations_pay_post_quote` → `hyperliquidActivationFee` (disabled by default).
> 
> Before Relay quotes, the Relay path now calls **`applyHyperliquidActivationFee`**: if the flag is on, it POSTs to HyperLiquid **`/info`** (`userNonFundingLedgerUpdates`) to see whether the account has already made an outbound send/withdraw; on failure it assumes activated so max withdraws are not blocked. For unactivated accounts it **lowers `sourceTokenAmount`** sent to Relay by the configured fee (default $1) and sets **`hyperliquidActivationFeeUsd`** on the quote request, while the **shown withdrawal amount stays the same**. Quote fee math **always adds** that reserved amount to **provider fee**, including when the relay fee is subsidized to zero.
> 
> Also adds **`getHyperliquidActivationFeeConfig`**, **`HYPERLIQUID_INFO_URL`**, and the optional **`hyperliquidActivationFeeUsd`** field on **`QuoteRequest`**, with unit and integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 726a8b589f2a528fceac328a61b6e1bcd8e0ebc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -90,19 +91,29 @@ export async function getRelayQuotes(
...
+          const isExactInputRequest =
+            Boolean(singleRequest.isMaxAmount) &&
+            new BigNumber(singleRequest.sourceTokenAmount).gt(0);
+
+          return hasTargetMinimum || isPostQuote || isExactInputRequest;
+        })
+        .map((singleRequest) =>
+          normalizeRequest(singleRequest, request.transaction),
+        )
+        .map((singleRequest) =>
```
OGPoyraz:
```suggestion
        .map((normalizedRequest) =>
```

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -90,19 +91,29 @@ export async function getRelayQuotes(
...
+          applyHyperliquidActivationFee(
+            singleRequest,
```
OGPoyraz:
```suggestion
            normalizedRequest,
```

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -972,6 +973,68 @@ export function getFiatOrderPollTimeoutMs(
...
+
+type PostQuoteFeatureFlags = {
+  default?: PostQuoteConfig;
+  overrides?: Record<string, PostQuoteConfig>;
+};
+
+export type HyperliquidActivationFeeConfig = {
+  enabled: boolean;
+  amountUsd: number;
+};
```
OGPoyraz: Can we pull these types above the file like others do?

packages/transaction-pay-controller/src/strategy/relay/constants.ts
```diff
@@ -9,6 +9,7 @@ export const RELAY_EXECUTE_URL = `${RELAY_URL_BASE}/execute`;
 export const RELAY_QUOTE_URL = `${RELAY_URL_BASE}/quote`;
 export const RELAY_STATUS_URL = `${RELAY_URL_BASE}/intents/status/v3`;
 export const HYPERLIQUID_EXCHANGE_URL = 'https://api.hyperliquid.xyz/exchange';
+export const HYPERLIQUID_INFO_URL = 'https://api.hyperliquid.xyz/info';
```
OGPoyraz: Minor:
```suggestion
export const HYPERLIQUID_URL_BASE = 'https://api.hyperliquid.xyz';
export const HYPERLIQUID_INFO_URL = `${HYPERLIQUID_URL_BASE}/info`;
export const HYPERLIQUID_EXCHANGE_URL = `${HYPERLIQUID_URL_BASE}/exchange`;
```

OGPoyraz (dismissed):

OGPoyraz (approved):

PR #9247, opened 2026-06-24
fix(seedless-onboarding-controller): fix InvalidPrimarySecretDataType for legacy accounts with client clock skew

huggingbot:
## Explanation

`#fetchAllSecretDataFromMetadataStore` sorted results then assumed `results[0]` was the primary SRP. This assumption breaks for legacy accounts where:

1. Items pre-dating the `dataType` migration have no `createdAt` (server-assigned TIMEUUID), so ordering falls back to the client-side `timestamp`.
2. Client clocks can skew — historically the server timestamp was also derived non-uniquely (`toUnixTimestamp(NOW())`, millisecond resolution), so collisions were possible.
3. A private key with a skewed-earlier `timestamp` sorts into `results[0]`, failing the `isMnemonic` check and throwing `InvalidPrimarySecretDataType`.

## Changes

Instead of hard-failing on `results[0]`, scan the full sorted list for the first item whose `dataType` is `PrimarySrp` or unset AND whose `SecretType` is `Mnemonic`. Promote that item to index 0 so callers continue to work correctly.

**Limitation:** `SecretType.Mnemonic` covers both primary and imported SRPs — the `dataType` plaintext server field is the only discriminator, and legacy items lack it. If multiple legacy untagged mnemonics exist, the true primary is indistinguishable; we take the oldest by sort order (same heuristic the `dataType` migration uses). For users already tagged (`dataType = PrimarySrp`), the sort puts them first anyway so the `findIndex` short-circuits immediately.

Only throws `InvalidPrimarySecretDataType` when no mnemonic candidate exists at all — a genuinely unrecoverable state.

## References
- Regression test: `should promote the primary SRP to first when a legacy non-mnemonic sorts ahead of it (clock skew)`

## Checklist

- [x] Tests written and passing (`yarn jest SeedlessOnboardingController`)
- [x] ESLint clean
- [x] No breaking changes to public API


grvgoel81:
@metamaskbot publish-preview

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1646,22 +1646,38 @@ export class SeedlessOnboardingController<
...
+      // items have no `createdAt`, so ordering falls back to the client
+      // `timestamp`, which clock skew can sort a private key ahead of.
+      // A candidate is a mnemonic whose `dataType` is `PrimarySrp` or unset.
+      // `dataType` (a plaintext server field) is the only thing separating
+      // primary from imported SRP; legacy items lack it and the encrypted
+      // `SecretType` groups both, so among legacy mnemonics the primary is
+      // indistinguishable — we take the oldest (best-effort).
+      const primaryIndex = results.findIndex((item) => {
+        const isDataTypePrimary =
+          item.dataType === undefined ||
```
lwin-kyaw: Legacy data has `type` field, `privateKey` data could have the `item.type="privatekey"`, so we should add one more condition here `item.type !== 'privatekey'`.

lwin-kyaw: Sorry, my bad. Wasn't aware of this, `SecretMetadata.matchesType(item, SecretType.Mnemonic)`.

lwin-kyaw (dismissed):

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1717,6 +1734,7 @@ export class SeedlessOnboardingController<
       oldAuthKeyPair: authKeyPair,
       newKeyShareIndex: globalKeyIndex,
       newPassword,
+      existingDataItems,
```
lwin-kyaw: There's a slim chance of race condition here.

After we fetched and sorted the existing data items and before we acquire the metadata lock, user could add new secret data and the `existingDataItems` isn't updated anymore.

This could cause the data loss or crash (if we don't handle decryption very well).

lwin-kyaw (approved):

PR #9259, opened 2026-06-24
feat(remote-feature-flag-controller): merge localOverrides into remoteFeatureFlags

asalsys:
## Summary

- `remoteFeatureFlags` now exposes the **effective** flag values consumers should read: processed remote flags merged with `localOverrides`
- Override mutations (`setFlagOverride`, `removeFlagOverride`, `clearAllFlagOverrides`) update `remoteFeatureFlags` directly instead of requiring callers to merge state themselves
- Remote flag fetches re-apply active overrides after processing so cached overrides survive refresh

## Behavior

Previously, `remoteFeatureFlags` contained only processed remote values and `localOverrides` was a separate map. Consumers that needed the effective value had to merge both manually, e.g. `{ ...remoteFeatureFlags, ...localOverrides }`.

With this change, the controller owns that merge:

| State field | Purpose |
|-------------|---------|
| `remoteFeatureFlags` | Effective values (`{ ...processedRemote, ...localOverrides }`) — **read this** |
| `localOverrides` | Which flags are locally overridden and their override values |
| `rawRemoteFeatureFlags` | Unprocessed API response |

### When merging happens

- **`updateRemoteFeatureFlags()`** — after version/threshold processing, overrides are spread onto the processed result
- **`setFlagOverride()`** — writes the override value directly into `remoteFeatureFlags`
- **`removeFlagOverride()` / `clearAllFlagOverrides()`** — restores processed remote values (tracked in a private in-memory cache refreshed on fetch)

### Consumer impact

- **Before:** merge `remoteFeatureFlags` + `localOverrides` at the call site
- **After:** read `remoteFeatureFlags` directly for effective values
- Call sites that already merge manually should drop the extra merge to avoid double-applying overrides

## Test plan

- [x] `yarn workspace @metamask/remote-feature-flag-controller run jest --no-coverage remote-feature-flag-controller.test.ts`
- [ ] Verify extension/mobile consumers that read `remoteFeatureFlags` no longer need a manual merge with `localOverrides`

### DIFF
--- a/packages/remote-feature-flag-controller/CHANGELOG.md
+++ b/packages/remote-feature-flag-controller/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Merge `localOverrides` into `remoteFeatureFlags` at the controller level so consumers receive effective flag values directly
 - Bump `@metamask/utils` from `^11.9.0` to `^11.11.0` ([#9074](https://github.com/MetaMask/core/pull/9074))
 - Bump `@metamask/controller-utils` from `^12.1.0` to `^12.3.0` ([#9058](https://github.com/MetaMask/core/pull/9058), [#9083](https://github.com/MetaMask/core/pull/9083), [#9218](https://github.com/MetaMask/core/pull/9218))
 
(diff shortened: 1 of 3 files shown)


Cal-L (approved):
LGTM

PR #9265, opened 2026-06-25
fix: add BackendWebsocketDataSource tests for Arbitrum USDC balance update

salimtb:
## Summary

Fixes stale or missing token balances when websocket live updates compete with Accounts API / RPC polling, and when websocket subscriptions are torn down and recreated during account switches or reconnects.

**Manual verification (Arbitrum USDC):**
- USDC on Arbitrum (`eip155:42161/erc20:…`) notification processed correctly after send/receive
- EVM addresses matched case-insensitively (checksummed vs lowercase) so notifications are not dropped for the wrong account
- Balance recovers correctly after websocket disconnect → reconnect → resubscribe (no silently dropped notification during the gap)

### What was broken?

`AssetsController` merges balances from several sources: websocket push (`BackendWebsocketDataSource`), Accounts API polling (`AccountsApiDataSource`), RPC, Snap, etc. Several races caused the UI to show stale values or never recover after a transaction:

1. **Websocket vs polling** — A live websocket balance could be overwritten seconds later by a stale Accounts API poll (TanStack Query cache, ~60s `staleTime`), or polling could “win” right after a websocket update.
2. **Account switch ordering** — On account group change we re-subscribed before forcing a fresh fetch, so websocket notifications could arrive before authoritative API data was applied.
3. **Websocket subscription races** — Concurrent subscribe/unsubscribe during account switches could interleave; EVM address comparison was case-sensitive, so a checksummed vs lowercase address looked like a “new” account and triggered unnecessary rebinds or missed matches.
4. **Disconnect/reconnect gap** — Notifications routed by `subscriptionId` could be dropped when the server-side subscription ID changed after reconnect; there was no per-channel callback fallback.
5. **`forceUpdate` still stale** — `getAssets({ forceUpdate: true })` did not bypass the Accounts API TanStack cache, and websocket freshness guards could block the forced fetch from applying (e.g. receiver switches to account 2 after a send but state still shows the pre-receive balance).

### What does this PR do?

**`AssetsController`**
- Tags applied updates with an internal `sourceId` on `DataResponse`.
- After a `BackendWebsocketDataSource` balance push, marks those `accountId:assetId` entries fresh for **120s**; passive polling sources (`AccountsApiDataSource`, `RpcDataSource`, `SnapDataSource`, `StakedBalanceDataSource`) **cannot overwrite** them during that window.
- **`getAssets({ forceUpdate: true })`** is tagged `getAssets:forceUpdate`, **clears** freshness locks for applied balances, and is **not** subject to the websocket freshness filter — authoritative for account switch / manual refresh.
- Account group and account-tree refreshes run under a **mutex**: **`forceUpdate` fetch first**, then **subscribe**, so API balances land before websocket recovery.
- Account-tree handler skips tree updates with **no overlapping account IDs** (full group switches still go through `#handleAccountGroupChanged`).

**`AccountsApiDataSource`**
- When `request.forceUpdate` is true, passes `{ staleTime: 0, gcTime: 0 }` to `fetchV5MultiAccountBalances` so forced refreshes bypass the TanStack cache.

**`BackendWebsocketDataSource`**
- Serializes subscribe/unsubscribe with a lock so account switches cannot interleave.
- Treats EVM addresses as unchanged when only checksumming differs; re-subscribes when the account set actually changes.
- Registers optional **per-channel callbacks** when server `subscriptionId` routing is unreliable after reconnect.
- Tears down subscriptions inline (without re-acquiring the subscribe lock) and forwards the original `DataRequest` with balance updates.

### Non-obvious details

- **`sourceId` is internal** — added to `DataResponse` for merge policy only; not a new public messenger action.
- **Freshness guard is polling-only** — websocket updates are still applied immediately; the guard only blocks *stale* poll merges, not other websocket pushes or `forceUpdate`.
- **Channel callbacks are best-effort** — if `BackendWebSocketService:addChannelCallback` is not delegated in the client messenger, the primary websocket subscription path still works; callbacks are a fallback for reconnect routing mismatches.

### Scope

All changes are in `@metamask/assets-controller` only. No dependency upgrades and no breaking public API changes.

## Test plan

- [x] `yarn workspace @metamask/assets-controller test` — new/updated unit tests pass
- [x] `yarn workspace @metamask/assets-controller run changelog:validate`
- [x] Manual: send USDC on Arbitrum, confirm balance updates via websocket
- [x] Manual: disconnect/reconnect websocket, confirm balance recovers after resubscribe
- [x] Manual: switch to receiver account after inbound transfer, confirm `forceUpdate` shows correct balance

**New unit test coverage:**
- `AssetsController`: polling does not overwrite recent websocket balance; `getAssets` `forceUpdate` wins over websocket freshness
- `AccountsApiDataSource`: `forceUpdate` bypasses TanStack cache
- `BackendWebsocketDataSource`: concurrent subscribe serialization, checksummed vs lowercase EVM addresses, channel-callback fallback, disconnect/reconnect chain reclaim

## References

- Fixes stale balance races after transactions, account switches, and websocket reconnects ([#9265](https://github.com/MetaMask/core/pull/9265))

## Checklist

- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate *(changelog only; behavior is internal merge policy)*
- [x] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them *(N/A — no breaking changes)*

---

> [!NOTE]
> **Medium Risk**
> Changes how displayed balances are merged across websocket and polling paths and reorders fetch/subscribe on account changes; incorrect freshness or locking could show wrong balances until the next `forceUpdate` fetch.

> **Overview**
> Fixes **stale or flickering balances** when backend websocket updates compete with Accounts API / RPC polling, and tightens **account-switch and websocket subscription** ordering so live updates are not dropped after reconnects or rapid resubscribes.
>
> - **`AssetsController`**: internal `sourceId` tagging; 120s websocket freshness guard for polling sources; `getAssets:forceUpdate` clears locks and wins; mutex-serialized fetch-then-subscribe on account change
> - **`AccountsApiDataSource`**: `staleTime: 0` / `gcTime: 0` on `forceUpdate`
> - **`BackendWebsocketDataSource`**: subscribe lock, case-insensitive EVM address matching, per-channel callback fallback, clean teardown on resubscribe

---

> [!NOTE]
> **Medium Risk**
> Changes how displayed balances are merged across websocket and polling paths and reorders fetch/subscribe on account changes; incorrect freshness or locking could show wrong balances until the next forceUpdate fetch.
> 
> **Overview**
> Fixes **stale or flickering token balances** when websocket pushes compete with Accounts API / RPC polling, and when subscriptions are recreated on account switch or reconnect.
> 
> **`AssetsController`** tags applied updates with internal `sourceId` on `DataResponse`. After websocket balance pushes, per-asset freshness locks block passive polling sources for **120s**; **`getAssets({ forceUpdate: true })`** is tagged `getAssets:forceUpdate`, skips that filter, and clears locks for applied balances. Account group and account-tree refreshes use a **mutex** and run **force fetch before subscribe**; account-tree updates with no overlapping account IDs are ignored.
> 
> **`AccountsApiDataSource`** passes `{ staleTime: 0, gcTime: 0 }` to multi-account balance fetches when `forceUpdate` is set.
> 
> **`BackendWebsocketDataSource`** serializes subscribe/unsubscribe, treats EVM addresses as unchanged when only checksumming differs, registers optional **per-channel callbacks** for reconnect routing, and passes the original `DataRequest` with balance updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 773e4e83d65cffac096ecddc3b4e6925952b88ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
+++ b/packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
@@ -160,6 +160,43 @@ function buildAccountActivityChannel(
 }
 
 /**
+ * Normalize addresses for stable comparison when detecting account changes.
+ *
+ * @param address - Account address (hex or base58).
+ * @returns Normalized address for comparison.
+ */
+function normalizeAddressForComparison(address: string): string {
+  return address.startsWith('0x') ? address.toLowerCase() : address;
+}
+
+/**
+ * Check whether subscribed account addresses changed (case-insensitive for EVM).
+ *
+ * @param nextAddresses - Addresses from the incoming subscribe request.
+ * @param existingAddresses - Addresses from the active subscription.
+ * @returns True when the address sets differ.
+ */
+function haveAddressesChanged(
+  nextAddresses: string[],
+  existingAddresses: string[],
+): boolean {
+  if (nextAddresses.length !== existingAddresses.length) {
+    return true;
+  }
+
+  const normalizedNext = nextAddresses
+    .map(normalizeAddressForComparison)
+    .sort();
+  const normalizedExisting = existingAddresses
+    .map(normalizeAddressForComparison)
+    .sort();
+
+  return normalizedNext.some(
+    (address, index) => address !== normalizedExisting[index],
+  );
+}
+
+/**
  * Normalize API chain identifier to CAIP-2 ChainId.
  * Passes through strings already in CAIP-2 form (e.g. eip155:1, solana:5eykt...).
  * Converts bare decimals to eip155:decimal.
@@ -209,6 +246,8 @@ function toChainId(chainIdOrDecimal: number | string): ChainId {
  * - BackendWebSocketService:subscribe
  * - BackendWebSocketService:getConnectionInfo
  * - BackendWebSocketService:findSubscriptionsByChannelPrefix
+ * - BackendWebSocketService:addChannelCallback
+ * - BackendWebSocketService:removeChannelCallback
  */
 const DEFAULT_CHAINS_REFRESH_INTERVAL_MS = 20 * 60 * 1000; // 20 minutes
 
@@ -247,6 +286,12 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
 
   /** Store original subscription requests for reconnection */
   readonly #subscriptionRequests: Map<string, SubscriptionRequest> = new Map();
+
+  /** Channels with registered BackendWebSocketService channel callbacks */
+  readonly #registeredChannelCallbacks: Set<string> = new Set();
+
+  /** Serializes subscribe/unsubscribe so account switches cannot interleave. */
+  #subscribeLock: Promise<void> = Promise.resolve();
 
   constructor(options: BackendWebsocketDataSourceOptions) {
     super(CONTROLLER_NAME, {
@@ -448,6 +493,23 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
   // ============================================================================
 
   async subscribe(subscriptionRequest: SubscriptionRequest): Promise<void> {
+    const previousLock = this.#subscribeLock;
+    let releaseLock: () => void = () => undefined;
+    this.#subscribeLock = new Promise<void>((resolve) => {
+      releaseLock = resolve;
+    });
+
+    await previousLock;
+    try {
+      await this.#subscribeInternal(subscriptionRequest);
+    } finally {
+      releaseLock();
+    }
+  }
+
+  async #subscribeInternal(
+    subscriptionRequest: SubscriptionRequest,
+  ): Promise<void> {
     const { request, subscriptionId, isUpdate } = subscriptionRequest;
 
     // Filter to active chains only
@@ -473,7 +535,7 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
         this.#pendingSubscriptions.set(subscriptionId, subscriptionRequest);
         return;
       }
-    } catch {
+    } catch (error) {
       // Store anyway - will be processed when we can connect
       this.#pendingSubscriptions.set(subscriptionId, subscriptionRequest);
       return;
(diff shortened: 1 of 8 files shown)


salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
```diff
@@ -554,6 +555,17 @@ export class BackendWebsocketDataSource extends AbstractDataSource<
...
       this.#subscriptionRequests.set(subscriptionId, subscriptionRequest);
+
+      // Catch-up fetch: recover any balance updates missed during the disconnect
+      // window. The WS is purely event-driven and has no state for the gap period,
+      // so we fire one API fetch immediately after re-subscribing.
+      this.#fetchCatchUpBalances(
+        request,
+        chainsToSubscribe,
+        subscriptionRequest.onAssetsUpdate,
+      ).catch((catchUpError: unknown) => {
```
Prithpal-Sooriya: wrap in lodash throttle?

downside is users would wait a little for updates to appear.

```
_.throttle(this.#fetchCatchUpBalances, 300ms)(...)
```

Prithpal-Sooriya: lets test first to see how often this happens.

Prithpal-Sooriya (dismissed):

packages/assets-controller/src/data-sources/BackendWebsocketDataSource.ts
Prithpal-Sooriya: just wanna make sure so we don't raise any incidents of API degradation/getting hit

salimtb:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #9270, opened 2026-06-25
feat(network-controller): emit RPC service analytics via AnalyticsController

cryptodev-2s:
## Explanation

Today `NetworkController` emits no analytics. It publishes `rpcEndpointUnavailable` and `rpcEndpointDegraded` events, and **each client app independently** subscribes to them and turns them into the `RPC Service Unavailable` / `RPC Service Degraded` Segment events. The extension still routes those through `MetaMetricsController`; mobile already routes them through `AnalyticsController`. The handlers are near identical duplication across both repos.

This moves that translation into `NetworkController` so it is written once and delivered through the `AnalyticsController:trackEvent` action over the messenger (which is the whole reason `AnalyticsController` exists). Clients can then delete their duplicated handlers and just wire the new option.

The feature is fully opt in via a new optional `analytics` constructor option. When omitted, no analytics are emitted and `AnalyticsController` is never called, so existing consumers are unaffected. The two genuinely client specific pieces are injected because core cannot know them:

- `isRpcEndpointUrlPublic(url)`: whether an endpoint URL is safe to report verbatim (depends on the client network lists, Quicknode URLs, Infura key)
- `rpcServiceEventsSampleRate`: the proportion of events to emit (depends on the build environment, typically 1% in production and 100% in dev)

Core reads `analyticsId` via `AnalyticsController:getState`, applies the sample rate with `generateDeterministicRandomNumber`, skips local connection errors, builds the properties, and delivers the event. `AnalyticsController` keeps owning consent gating.

## What changed

- New `src/rpc-service-events.ts`: pure helpers (`sanitizeRpcEndpointUrl`, `buildRpcServiceEventProperties`, `toAnalyticsTrackingEvent`) plus the `NetworkControllerAnalyticsOptions` and `RpcServiceEventName` types
- `src/NetworkController.ts`: new optional `analytics` option, widened `AllowedActions` with `AnalyticsController:getState` + `AnalyticsController:trackEvent`, two guarded self subscriptions, and a private `#trackRpcServiceEvent`
- Added `@metamask/analytics-controller` dependency and tsconfig references (no dependency cycle)
- `index.ts`, `tests/helpers.ts`, `CHANGELOG.md`, and a new `tests/NetworkController.analytics.test.ts`

## Open question

`AnalyticsTrackingEvent` has no `category` field, so the old `category: 'Network'` is not carried for now. Confirming with the analytics owners how they want category mapped before relying on it downstream.

## Follow ups (separate repos)

- metamask-extension: delete `messenger-action-handlers.ts` + `utils.ts`, pass the `analytics` option, delegate the two AnalyticsController actions. This is where the extension stops using `MetaMetricsController` for these events.
- metamask-mobile: same deletion and wiring (it already delegates AnalyticsController).

## References

N/A

## Checklist

- [x] I've manually tested code paths covered by the new tests

---

> [!NOTE]
> **Medium Risk**
> Breaking messenger requirements affect all NetworkController consumers; analytics touches RPC failure paths but defaults to off and failures are swallowed via captureException.
> 
> **Overview**
> **Centralizes RPC health analytics in `NetworkController`** so clients no longer need duplicate subscribers on `rpcEndpointUnavailable` / `rpcEndpointDegraded`. The controller subscribes to those events and, when configured, emits **RPC Service Unavailable** and **RPC Service Degraded** via `AnalyticsController:trackEvent`.
> 
> New optional constructor/instance option **`analyticsOptions`** (`isRpcEndpointUrlPublic`, `rpcServiceEventsSampleRate`) defaults to no tracking (`sample rate 0`, URLs reported as `'custom'`). Tracking skips local connection errors, requires a non-empty `analyticsId`, applies deterministic sampling, and sanitizes endpoint URLs in new `rpc-service-analytics` helpers.
> 
> **Breaking messenger wiring:** `NetworkControllerMessenger` must delegate **`AnalyticsController:getState`** and **`AnalyticsController:trackEvent`**. `@metamask/wallet` passes `analyticsOptions` through and delegates those actions; **`wallet-cli`** registers no-op Analytics handlers (empty `analyticsId`) so daemon startup does not throw. Adds `@metamask/analytics-controller` dependency and tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10a3f4c8a73b3f4d4c031ac6408e68bb9323e62f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/network-controller/src/NetworkController.ts
```diff
@@ -762,6 +778,15 @@ export type NetworkControllerOptions = {
...
   ) => Omit<PollingBlockTrackerOptions, 'provider'>;
+  /**
+   * Optional configuration that, when provided, makes the controller emit
+   * "RPC Service Unavailable" and "RPC Service Degraded" analytics events via
+   * the `AnalyticsController:trackEvent` action whenever an RPC endpoint becomes
+   * unavailable or degraded. When omitted, no analytics are emitted and the
+   * `AnalyticsController` actions are never called. When provided, the messenger
+   * must allow `AnalyticsController:getState` and `AnalyticsController:trackEvent`.
+   */
+  analytics?: NetworkControllerAnalyticsOptions;
```
mcmire: Do we need to make this optional? I see that [we already follow this pattern for AccountTreeController](https://github.com/MetaMask/core/blob/7cc2df6f0f75c95b9a9f869d5805cb7e9eabd116/packages/account-tree-controller/src/backup-and-sync/syncing/metadata.ts#L38) but I'm wondering if that's a good idea. We don't typically make interactions with other controllers optional. 

Maybe we can this `analyticsOptions` to match `getRpcServiceOptions` and `getBlockTrackerOptions`?

packages/network-controller/src/NetworkController.ts
```diff
@@ -1374,6 +1445,80 @@ export class NetworkController extends BaseController<
...
+        duration: payload.duration,
+        retryReason: payload.retryReason,
+        rpcMethodName: payload.rpcMethodName,
+        traceId: payload.traceId,
+        type: payload.type,
+      });
+
+      this.messenger.call(
+        'AnalyticsController:trackEvent',
+        toAnalyticsTrackingEvent(name, properties),
```
mcmire: It seems like we are building event properties only to then convert them to a tracking event. Maybe we can combine these steps by folding `buildRpcServiceEventProperties` into `toAnalyticsTrackingEvent`?

packages/network-controller/src/NetworkController.ts
```diff
@@ -1374,6 +1445,80 @@ export class NetworkController extends BaseController<
...
+      );
+    } catch (error) {
+      this.messenger.captureException?.(error as Error);
```
mcmire: Can we convert this to an Error instead of typecasting? `@metamask/utils` has a `wrapError` function so we can use:

``` suggestion
      this.messenger.captureException?.(
        wrapError(error, "Could not create analytics event"),
      );

File: packages/network-controller/src/NetworkController.ts [hunk 4]
```diff
@@ -1364,6 +1393,48 @@ export class NetworkController extends BaseController<
...
+            duration,
+            endpointUrl,
+            error,
+            retryReason,
+            rpcMethodName,
+            traceId,
+            type,
+          });
+        },
+      );
```
mcmire: I feel like the way that I wrote this before — having one function/method that's responsible for creating two kinds of events — makes this more complicated than it seems.

In fact since we already have a file for storing analytics-related things why don't we move the private method(s) there?

Also do we need to spell out the payload here? And do we really need to pass the analytics options if we can get them through `this.#analytics`?

Putting this together, what do you think about:

``` suggestion
      this.messenger.subscribe(
        `${this.name}:rpcEndpointUnavailable`,
        (payload) => trackRpcServiceUnavailable(this.messenger, payload),
      );
      this.messenger.subscribe(`${this.name}:rpcEndpointDegraded`, (payload) =>
        trackRpcServiceDegraded(this.messenger, payload),
      );
```

packages/network-controller/src/rpc-service-analytics.ts
mcmire, line 1: What do you think about renaming this file to `rpc-service-analytics.ts`? "Events" could mean messenger events, not analytics events.

packages/network-controller/src/rpc-service-analytics.ts
```diff
@@ -0,0 +1,154 @@
...
+ * client's lists of known networks, and the sample rate depends on the client's
+ * build environment.
+ */
+export type NetworkControllerAnalyticsOptions = {
+  /**
+   * Returns `true` if the given RPC endpoint URL is safe to report verbatim (a
+   * "public" endpoint), or `false` if it must be reported as the literal string
+   * `'custom'` to avoid leaking private servers.
+   */
+  isRpcEndpointUrlPublic: (endpointUrl: string) => boolean;
```
mcmire: Do we need this to be an option? Can we move the `isRpcEndpointUrlPublic` logic into this package?

packages/network-controller/src/rpc-service-analytics.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Wraps an event name and properties into the shape expected by the
+ * `AnalyticsController:trackEvent` action.
+ *
+ * @param name - The analytics event name.
+ * @param properties - The analytics event properties.
+ * @returns The analytics tracking event.
+ */
+export function toAnalyticsTrackingEvent(
```
mcmire: See note in `NetworkController.ts`. Curious if we should have a `buildRpcServiceUnavailableAnalyticsTrackingEvent` and a `buildRpcServiceDegradedAnalyticsTrackingEvent`. We could move common event properties into a separate constant or function or something.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Done in 57081f2989.

packages/network-controller/src/rpc-service-analytics.ts
cryptodev-2s: Renamed in cddfca6bc7.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Renamed to `analyticsOptions` and made it required in 016f403b22, since both clients wire analytics today.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Combined in b9b6236a2c. The builders for each event now produce the tracking event directly.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Done in b9b6236a2c. Moved to trackRpcServiceUnavailable/trackRpcServiceDegraded in the analytics file, subscribing with just the payload. Kept analyticsOptions as a param since it's still injected, pending the isRpcEndpointUrlPublic question below.

packages/network-controller/src/rpc-service-analytics.ts
cryptodev-2s: Split in b9b6236a2c into buildRpcServiceUnavailableAnalyticsTrackingEvent/buildRpcServiceDegradedAnalyticsTrackingEvent, with shared props in buildCommonRpcServiceEventProperties.

packages/network-controller/src/rpc-service-analytics.ts
cryptodev-2s: Keeping this injected on purpose. Deciding which endpoint URLs are safe to send to analytics is a privacy policy each client owns, and the inputs already diverge between extension and mobile (checked against origin/main):

- The Quicknode set differs: mobile recognizes `bsc-mainnet` (`app/util/networks/customNetworks.tsx`), the extension map does not (`shared/constants/network.ts`). The URLs themselves are secrets injected at build time via `process.env.QUICKNODE_*_URL`.
- The known custom list differs: extension uses `FEATURED_RPCS` + `BUILT_IN_CUSTOM_NETWORKS_RPC` (`shared/lib/network-utils.ts`), mobile uses `PopularList` + `BUILT_IN_CUSTOM_NETWORKS_RPC` (`customNetworks.tsx`). Mobile's `PopularList` embeds `require('../../images/*.png')` assets, so it cannot exist outside the app.
- The public domain check is a different module on each: extension goes through `isKnownDomain` backed by a chainlist cache (`app/scripts/lib/util.ts`), mobile uses `isPublicRpcDomain` with a hardcoded `ALLOWED_PROVIDER_DOMAINS = { infura.io, alchemyapi.io }` (`app/util/rpc-domain-utils.ts`). So a given URL is not guaranteed to classify the same on both.
- `infuraProjectId` is also per client.

`@metamask/network-controller` and `@metamask/controller-utils` have none of this (only `BUILT_IN_CUSTOM_NETWORKS_RPC` is shared). Moving the classifier in would force the package to bundle or fetch curated lists, build secrets, and a domain registry, and it still could not reconcile the divergence, so it would end up needing per client config or injection anyway. It also matches the controller's existing injected policy hooks (`getRpcServiceOptions`, `getBlockTrackerOptions`, `failoverUrls`), so I would keep `isRpcEndpointUrlPublic` as a caller supplied option.

cryptodev-2s:
@metamaskbot publish-preview

jpuri (dismissed):

cryptodev-2s:
Clients integration blocked by https://github.com/MetaMask/metamask-extension/pull/44527 and https://github.com/MetaMask/metamask-mobile/pull/33430

packages/wallet/src/initialization/instances/network-controller/network-controller.ts
```diff
@@ -28,6 +29,8 @@ export const networkController: InitializationConfiguration<
     parent.delegate({
       messenger: networkControllerMessenger,
       actions: [
+        'AnalyticsController:getState',
+        'AnalyticsController:trackEvent',
```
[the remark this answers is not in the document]
cryptodev-2s: @mcmire circling back on making `analyticsOptions` required. Bugbot flagged a real consequence: the wallet delegates `AnalyticsController:getState` and `AnalyticsController:trackEvent` to `NetworkController`, but consumers that do not run analytics (the default `Wallet` and `wallet-cli`) register no handler for them. So on every `rpcEndpointUnavailable` / `rpcEndpointDegraded` event the controller calls `getState`, the messenger throws, and `trackRpcServiceEvent` swallows it.

I looked at registering the handlers on the wallet so we could keep it required, but it does not work cleanly:

- The wallet cannot be the `AnalyticsController` itself. It needs a platform adapter to deliver events (Segment on extension, native on mobile), which is client specific, and the wallet is a generic framework with no adapter. It is also not a wallet default instance, and the wallet does not depend on `@metamask/analytics-controller`.
- Registering it there would collide with the client's own `AnalyticsController`. The extension and mobile already register it on the root messenger they hand to the wallet, and messenger actions register once per tree, so a wallet owned one would clash. The delegation works for the clients precisely because the handler lives on their root and the wallet just delegates access to it.

So I would like to make `analyticsOptions` optional. The clients that wire analytics (extension, mobile) still pass it and register `AnalyticsController` exactly as before, so nothing changes for them. Consumers that do not wire analytics simply omit it, so the controller never subscribes and never calls `AnalyticsController`, which removes the throw. It also lets us drop the placeholder options we added in `wallet-cli`.

This stays compatible with the wallet adopting `AnalyticsController` as a default instance later (like gasFee/seedless/passkey), at which point we could move back to required. Let me know if you disagree.

packages/network-controller/src/rpc-service-analytics.ts
mcmire: Oh right, I forgot we have a bunch of logic to determine whether something is public in each client. Hmm. Just feels a bit dirty to have this here but I can understand why. We can think about this more later, maybe we can consolidate some of that logic in `core` at some point in the future. Or maybe moving the network config registry will make this easier to determine. We'll have to see.

Fine to leave this here for now.

packages/wallet/src/initialization/instances/network-controller/network-controller.ts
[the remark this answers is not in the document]
mcmire: The problem with using an option to conditionally call actions on another controller  is that it doesn't match what the messenger type says should be possible. We don't have any concept of conditional actions in the messenger type. If an action is present at the type level, then engineers ought to be able to assume it is callable at runtime. If it is not, that should be a bug, not a "feature".

Another problem: if we're allowing for the possibility that `AnalyticsController` actions may not have registered here, then what about other controllers? Are we saying that any time anyone wants to use `AnalyticsController` in a controller they have to add this option? If so, that seems like a footgun to me — it seems like something people are likely to forget.

So that's why I'm hesitant to make this conditional. It fits our current design for the messenger and it feels easier to understand if we can trust `AnalyticsController:getState` and `AnalyticsController:trackEvent` are callable at all times (because they're present in `NetworkControllerMessenger`).

---

As for how we would do this, I don't think it would make sense for `Wallet` to register these handlers. `Wallet` is just like any other service, it does not need to pre-register handlers for actions it (or any of its "child" controller/services) uses.

However, what we could do is add `AnalyticsController:getState` and `AnalyticsController:trackEvent` to `Wallet`'s "root messenger" type, copying the types from `analytics-controller` itself.

If we went this route, anything that wants to use `Wallet` would need to register `AnalyticsController:getState` and `AnalyticsController:trackEvent` _somewhere_ (the change to `Wallet` would be breaking). This already happens for extension and mobile, so there would be no changes needed there. But as you point out, `wallet-cli` would be a special case, as it would need to register action handlers that end up doing nothing (considering it does not actually use `AnalyticsController`). This may sound strange to do, but we would only need to do it once, and we could add a comment to explain why we're doing that.

---

What do you think about that approach?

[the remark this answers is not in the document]
cryptodev-2s: Agreed, the type contract argument makes sense. I'll go with that approach.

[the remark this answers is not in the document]
cryptodev-2s: Done in 3e2e73bc51. Added the two AnalyticsController actions to the Wallet root messenger type and registered stub handlers in wallet-cli.

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/network-controller/src/rpc-service-analytics.ts
```diff
@@ -0,0 +1,277 @@
...
+ * @param messenger - The controller messenger.
+ * @param analyticsOptions - The analytics configuration.
+ * @param payload - The `rpcEndpointUnavailable` event payload.
+ */
+export function trackRpcServiceUnavailable(
+  messenger: NetworkControllerMessenger,
+  analyticsOptions: NetworkControllerAnalyticsOptions,
+  payload: RpcEndpointUnavailablePayload,
+): void {
+  trackRpcServiceEvent(messenger, analyticsOptions, payload.error, () =>
```
mcmire: Nit: Do we need to pass a function here? And since we're passing more than 3 arguments would it be worth it for `trackRpcServiceEvent` to take an options bag?

packages/network-controller/tests/NetworkController.analytics.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  });
+  return { controller, rootMessenger, networkControllerMessenger };
+}
+
+describe('NetworkController analytics', () => {
+  it('emits "RPC Service Unavailable" when an endpoint becomes unavailable', () => {
+    const { networkControllerMessenger } = buildController({
+      analyticsOptions: DEFAULT_ANALYTICS_OPTIONS,
+    });
+    const callSpy = jest.spyOn(networkControllerMessenger, 'call');
```
mcmire: Why are we spying on `call`? Since we have a messenger and we control both subscribing and publishing, can we configure `buildController` so that we can pass a Jest mock function so that it can use it to register `AnalyticsController:trackEvent`? Then we can assert that the mock function is called.

packages/wallet-cli/src/daemon/wallet-factory.test.ts
```diff
@@ -49,6 +49,8 @@ function makeMockWallet(): Wallet {
       call: jest.fn(),
       subscribe: jest.fn(),
       unsubscribe: jest.fn(),
+      // Called when the analytics stub messenger delegates its handlers up.
+      _internalRegisterDelegatedActionHandler: jest.fn(),
```
mcmire, line 53: Yikes. I don't love this. This is why we don't encourage mocking the messenger in tests. Oh well :(

packages/network-controller/CHANGELOG.md
```diff
@@ -7,8 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Add required `analyticsOptions` constructor option that makes `NetworkController` emit `RPC Service Unavailable` and `RPC Service Degraded` analytics events via the `AnalyticsController:trackEvent` action when an RPC endpoint becomes unavailable or degraded ([#9270](https://github.com/MetaMask/core/pull/9270))
+  - The option takes `isRpcEndpointUrlPublic` (decides whether an endpoint URL is safe to report verbatim or as `'custom'`) and `rpcServiceEventsSampleRate` (the proportion of events to emit, between `0` and `1`).
+  - Adds the `NetworkControllerAnalyticsOptions` and `RpcServiceEventName` types.
+
 ### Changed
 
+- **BREAKING:** `NetworkControllerMessenger` now requires the `AnalyticsController:getState` and `AnalyticsController:trackEvent` actions ([#9270](https://github.com/MetaMask/core/pull/9270))
```
mcmire: Nit: What do you think about grouping this entry with the other one you added for this PR?

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
```diff
@@ -60,6 +60,22 @@ describe('createWallet (real Wallet, in-memory)', () => {
...
+      ).toBe('');
+      expect(
+        wallet.messenger.call('AnalyticsController:trackEvent', {
+          name: 'test',
+          properties: {},
+          sensitiveProperties: {},
+          saveDataRecording: false,
+          hasProperties: false,
+        }),
+      ).toBeUndefined();
```
mcmire, line 92: Should we add this? It seems that the goal of this test is to confirm that the wallet is unlocked, not to confirm that the wallet is fully initialized.

``` suggestion
```

packages/network-controller/src/NetworkController.ts
```diff
@@ -765,6 +776,13 @@ export type NetworkControllerOptions = {
   getBlockTrackerOptions?: (
     rpcEndpointUrl: string,
   ) => Omit<PollingBlockTrackerOptions, 'provider'>;
+  /**
+   * Configuration that makes the controller emit "RPC Service Unavailable" and
+   * "RPC Service Degraded" analytics events via the `AnalyticsController:trackEvent`
+   * action whenever an RPC endpoint becomes unavailable or degraded. The messenger
+   * must allow `AnalyticsController:getState` and `AnalyticsController:trackEvent`.
+   */
+  analyticsOptions: NetworkControllerAnalyticsOptions;
```
mcmire: What do you think about making `analyticsOptions` and both of its properties optional? This way in e.g. `gas-fee-controller.test.ts` (below) we don't need to make any changes at all. The default for `isRpcEndpointUrlPublic`, for instance, could be `() => false` and the default for the sample rate could be 0.

packages/wallet/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Wire analytics into the default `NetworkController` initialization ([#9270](https://github.com/MetaMask/core/pull/9270))
+  - Adds a required `instanceOptions.networkController.analyticsOptions` option (`isRpcEndpointUrlPublic` and `rpcServiceEventsSampleRate`) that the controller uses to emit `RPC Service Unavailable` and `RPC Service Degraded` events.
```
mcmire: See note above, maybe we don't need to make this required?

packages/wallet/src/initialization/defaults.ts
```diff
@@ -55,7 +59,18 @@ export type DefaultInstances = {
   >]: ExtractInstance<DefaultConfigs[Key]>;
 };
 
-export type DefaultActions = MessengerActions<AllDefaultMessengers>;
+/**
+ * `NetworkController` calls `AnalyticsController:getState` and
+ * `AnalyticsController:trackEvent` to emit RPC service analytics. `Wallet` does
```
mcmire: Hmm, is it necessary to explicitly add the AnalyticsController actions here? `NetworkControllerMessenger` declares these actions, so `MessengerActions<AllDefaultMessengers>` should include them automatically.

packages/network-controller/src/rpc-service-analytics.ts
cryptodev-2s: Done in 6bbf34cf30. Moved to an options bag. Kept the builder as a function so the event build (which runs the client's `isRpcEndpointUrlPublic`) stays inside the try, so a throw is captured instead of propagated, and so we skip building when the event will not be sent.

packages/network-controller/CHANGELOG.md
cryptodev-2s: Done in 3f2da06a5f. Merged the messenger requirement into the analytics entry.

packages/wallet/src/initialization/defaults.ts
cryptodev-2s: Done in 5c4917f1b1. Removed them; they were already included via `NetworkControllerMessenger`, and dropped the now unused analytics-controller dep.

packages/wallet-cli/src/daemon/wallet-factory.integration.test.ts
cryptodev-2s: Done in 0bfb8f3c8e. Moved them out of the unlock test into a dedicated test, since they cover the stub handlers.

packages/network-controller/tests/NetworkController.analytics.test.ts
cryptodev-2s: Done in f82403dbb8. `buildController` now registers a `trackEvent` mock and the tests assert on it.

packages/network-controller/src/NetworkController.ts
cryptodev-2s: Done in 17480a8082. Made `analyticsOptions` and both properties optional (defaults `isRpcEndpointUrlPublic: () => false`, sample rate `0`), and dropped it from the wallet and wallet-cli fixtures. The stubs stay, so the messenger calls still resolve.

packages/network-controller/tests/helpers.ts
```diff
@@ -666,6 +698,7 @@ export async function withController<ReturnValue>(
   const controller = new NetworkController({
     messenger: networkControllerMessenger,
     infuraProjectId: 'infura-project-id',
+    analyticsOptions: MOCK_ANALYTICS_OPTIONS,
```
mcmire: We can drop this now, right? 

``` suggestion
```

packages/network-controller/tests/NetworkController.test.ts
```diff
@@ -174,6 +175,7 @@ describe('NetworkController', () => {
               networkConfigurationsByChainId: {},
             },
             infuraProjectId: 'infura-project-id',
+            analyticsOptions: MOCK_ANALYTICS_OPTIONS,
```
mcmire: Can we drop these now that this is optional?

mcmire (commented):
A couple more things I noticed.

packages/network-controller/tests/helpers.ts
cryptodev-2s: Done in b01be2d597.

packages/network-controller/tests/NetworkController.test.ts
cryptodev-2s: Done in b01be2d597.

mcmire (approved):
LGTM!

PR #9273, opened 2026-06-25
fix: remove ws balance freshness guard

salimtb:
## Explanation

UI PR: https://github.com/MetaMask/metamask-extension/pull/43829

Token balances in `@metamask/assets-controller` could stay stale after transactions, account switches, or WebSocket reconnects. Several independent issues contributed:

1. **WebSocket notifications were not reaching `BackendWebsocketDataSource`** — Clients subscribe on wildcard channels (`account-activity.v1.eip155:0:0x…`) but the server sends notifications on specific chains (`…eip155:42161:0x…`). `BackendWebSocketService` matched channels exactly, so post-confirm balance pushes were silently dropped even though frames appeared in the Network tab.

2. **Stale `subscriptionId` after reconnect/resubscribe** — Notifications could arrive with a server subscription id that no longer matched the local map. The subscription callback was skipped and channel fallback was unreliable.

3. **Race conditions on account switch** — Overlapping subscribe/unsubscribe and fetch work could interleave, dropping notifications or applying responses out of order.

4. **Stale balances from cached/partial fetches** — `AccountsApiDataSource` used TanStack Query’s 60s balance cache even on `forceUpdate`, and force-refresh pipelines used merge semantics that never removed tokens absent from partial API responses (e.g. USDC still shown after a swap when the API only returned ETH).

5. **Websocket freshness guard blocked corrections** — A 120s guard prevented polling/API updates from overwriting recent WS balances. That helped in some send scenarios but also blocked legitimate corrections (e.g. receiver account showing wrong balance after account switch).

**`@metamask/core-backend` — `BackendWebSocketService`**

- **Wildcard channel matching** — `eip155:0:0xaddr` matches `eip155:42161:0xaddr` for both subscription callbacks and channel callbacks.
- **Subscription fallback** — If `subscriptionId` lookup fails, resolve the handler by matching the notification channel against registered subscription channels.
- **Nested payload normalization** — Promote `channel` / `subscriptionId` from nested `data` when the server wraps notifications.
- **No-callback fallthrough** — If a subscription entry exists but has no callback, fall through to channel routing instead of swallowing the message.

**`@metamask/assets-controller` — `BackendWebsocketDataSource`**

- Serializes subscribe/unsubscribe via a lock to prevent account-switch races.
- Registers subscription state and handlers **before** awaiting the server subscribe handshake.
- Registers `addChannelCallback` per channel as a fallback when subscription-id routing fails.
- Case-insensitive EVM address matching for notifications.
- Resolves `onAssetsUpdate` from stored subscription state when notifications arrive with stale ids.

**`@metamask/assets-controller` — `AssetsController`**

- Fetches before re-subscribing on account switch; `#accountRefreshMutex` serializes overlapping refresh work.
- Deduplicates account-tree refresh when selected accounts have not changed.
- Removes the 120s websocket balance freshness guard so force-refresh and polling can correct stale WS values.

**`@metamask/assets-controller` — balance update modes**

- Adds **`update`** mode: patches balance amounts only; does not remove tokens or overwrite metadata/prices. Used for `getAssets({ forceUpdate: true })` and Accounts API / Snap fetch responses when the API returns a partial snapshot (e.g. only ETH after a swap).
- **`merge`** remains the default for event-driven updates (WS, polling).
- **`full`** remains for responses that should be authoritative for a chain scope.

**`AccountsApiDataSource`**

- Passes `{ staleTime: 0, gcTime: 0 }` to the API client when `request.forceUpdate` is true.

End-to-end flow after a confirmed swap:

---

> [!NOTE]
> **High Risk**
> Touches core balance state, websocket subscription routing, and introduces a breaking AssetsController messenger event; incorrect merge or routing could show wrong balances after txs or network switches.
> 
> **Overview**
> Fixes stale unified asset balances after transactions, account/network switches, and websocket reconnects by tightening how updates are routed, merged, and refreshed.
> 
> **`BackendWebSocketService`** now matches account-activity notifications to wildcard subscribe channels (`eip155:0` vs specific chain), falls back to channel-based subscription lookup when `subscriptionId` is stale, normalizes nested notification payloads, and re-enables previously skipped connection-timeout tests.
> 
> **`AssetsController`** drops the 120s websocket freshness guard and `sourceId`-based polling filters; adds **`replaceCoveredChainBalances`** on merge so `getAssets({ forceUpdate: true })` can drop stale tokens on API-covered chains without a full replace. It subscribes to **`AccountActivityService:balanceUpdated`** (breaking messenger allowlist), **`TransactionController:transactionConfirmed`**, and **`NetworkController:networkDidChange`** (refresh `activeChains`, re-subscribe, force-fetch selected chain + missing prices). Startup uses fetch-then-subscribe via `#runStartupRefresh`; the slow Snap/RPC pipeline skips chains the fast Accounts API path already satisfied.
> 
> **Data sources / pipeline:** shared **`processAccountActivityBalanceUpdates`**; websocket subscribe registers handlers before the handshake; **`DetectionMiddleware`** queues detected assets missing prices; **`PriceDataSource`** uses merge on poll and fetches only missing detected prices; **`AccountsApiDataSource`** / **`BackendWebsocketDataSource`** / **`RpcDataSource`** expose active-chain refresh on network switch.
> 
> Tests and changelogs updated across `assets-controller` and `core-backend`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd7b53f49c8819e9eb525531b677460fff4a2813. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/AssetsController.ts
Prithpal-Sooriya, line 1: 3.5K Controller 💀 
FUTURE - A lot of then internals can be gutted and moved into separate files (for better testability).

E.g:
- Move pipelines/middleware construction into separate file
- Move state update logic (merge vs update vs full) into a separate file
- Subscription handlers can also be moved into a separate file (you can pass in messenger as inputs to functions).
- Traversing or shape manipulation can also be decoupled.

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

packages/assets-controller/src/AssetsController.ts
```diff
@@ -326,31 +321,50 @@ type AllowedActions =
...
+
+type ClientControllerStateChangedEvent = ControllerStateChangedEvent<
+  'ClientController',
+  ClientControllerState
+>;
+
+type NetworkEnablementControllerStateChangedEvent = ControllerStateChangedEvent<
+  'NetworkEnablementController',
+  NetworkEnablementControllerState
+>;
```
Prithpal-Sooriya, line 337: This needs to be refactored and moved out of this controller.

Gudahtt: Deleted actually, this is a bug that is causing type errors in mobile. These events being constructed here for other controllers _do not exist_.

PR #9289, opened 2026-06-26
feat(remote-feature-flag-controller): return threshold values directly and track threshold groups

asalsys:
## Summary

- **BREAKING:** Threshold feature flags now return the selected `value` directly from `remoteFeatureFlags` instead of a `{ name, value }` wrapper object.
- Add optional `featureFlagThresholdGroups` state field (`Record<string, string>`) that maps feature flag names to their selected threshold group name when the selected threshold entry includes `name`.
- Remove `normalizeThresholdValue` and the legacy `{ name, value }` wrapper fallback.

## References

- Fixes [MCWP-655](https://consensyssoftware.atlassian.net/browse/MCWP-655)

## Motivation

Separating the processed flag value from threshold group metadata makes it easier for consumers to use threshold flags as plain values (booleans, objects, etc.) without unwrapping. The new `featureFlagThresholdGroups` field provides explicit access to A/B group assignment when the selected threshold entry includes `name`.

Entries that only define `thresholdName` (for example in `thresholdVersion: 2` configs) still resolve to the selected `value`, but do not populate `featureFlagThresholdGroups` because that field is driven by `name`.

## Migration

Consumers that previously read threshold group names from `remoteFeatureFlags[flagName].name` should instead:

1. Read the flag value directly from `remoteFeatureFlags[flagName]`.
2. Read the threshold group name from `featureFlagThresholdGroups[flagName]` when available (populated when the selected threshold entry includes `name`).

**Before:**
```ts
const flag = remoteFeatureFlags.myThresholdFlag;
// { name: 'groupB', value: { enabled: true } }
const groupName = flag.name;
const value = flag.value;
```

**After:**
```ts
const value = remoteFeatureFlags.myThresholdFlag;
// { enabled: true }

const groupName = featureFlagThresholdGroups.myThresholdFlag;
// 'groupB'
```

## Related

- Documentation: [MetaMask/contributor-docs#176](https://github.com/MetaMask/contributor-docs/pull/176)

## Test plan

- [x] `yarn workspace @metamask/remote-feature-flag-controller run test` — all tests pass with 100% coverage
- [x] Threshold flags return the selected `value` directly from `remoteFeatureFlags`
- [x] Threshold flags populate `featureFlagThresholdGroups` when the selected entry includes `name`
- [x] Threshold flags with only `thresholdName` return the value directly and leave `featureFlagThresholdGroups` empty
- [x] Stale `featureFlagThresholdGroups` entries are cleaned up when flags are removed from the server response

[MCWP-655]: https://consensyssoftware.atlassian.net/browse/MCWP-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Breaking change to the public shape of threshold flag values affects all consumers of `remoteFeatureFlags`; persisted state gains a new field but assignment logic is otherwise unchanged.
> 
> **Overview**
> **BREAKING:** Threshold/A/B flags in `remoteFeatureFlags` now expose the selected entry’s **`value` only** (string, boolean, object, etc.) instead of a `{ name, value }` wrapper. Group assignment moves to a new optional persisted state field **`featureFlagThresholdGroups`** (`flagName → group name`) when the winning threshold item has a `name`.
> 
> Processing drops **`normalizeThresholdValue`** and **`ThresholdVersion`** branching—legacy wrappers and “direct value vs wrapper” version handling are gone; threshold arrays are resolved the same way regardless of `thresholdVersion`. On each fetch, **`featureFlagThresholdGroups` is rebuilt** from the current server payload so entries for removed flags are cleared (covered by tests).
> 
> **Migration:** read values from `remoteFeatureFlags[flag]`; read group names from `featureFlagThresholdGroups[flag]` when present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98b74a20f2a3b89034e63d967c05e173ff8bead0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -67,6 +67,12 @@ const remoteFeatureFlagControllerMetadata = {
     includeInDebugSnapshot: false,
     usedInUi: false,
   },
+  featureFlagThresholdGroups: {
+    includeInStateLogs: true,
+    persist: true,
+    includeInDebugSnapshot: true,
+    usedInUi: true,
```
Gudahtt: Nit: I believe this would only be used by the metrics utility, which is not used as part of the UI?

```suggestion
    usedInUi: false,
```

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -314,6 +310,17 @@ export class RemoteFeatureFlagController extends BaseController<
       }
     }
 
+    const updatedFeatureFlagThresholdGroups = {
+      ...(this.state.featureFlagThresholdGroups ?? {}),
```
Gudahtt: This might result in a bug if a threshold is _removed_ on the LaunchDarkly side. We need to completely overwrite this value each time rather than merging it in, so that removed thresholds don't stick around.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -409,14 +419,22 @@ export class RemoteFeatureFlagController extends BaseController<
         );
 
         if (selectedGroup) {
-          processedValue = normalizeThresholdValue(selectedGroup);
+          processedValue = selectedGroup.value;
+          if (selectedGroup.thresholdName) {
```
Gudahtt: Where is `thresholdName` coming from? Currently our flags are using `name`, not `thresholdName`

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -409,14 +426,22 @@ export class RemoteFeatureFlagController extends BaseController<
         );
 
         if (selectedGroup) {
-          processedValue = normalizeThresholdValue(selectedGroup);
+          processedValue = selectedGroup.value;
+          if (selectedGroup.name) {
+            featureFlagThresholdGroupUpdates[remoteFeatureFlagName] =
+              selectedGroup.name;
```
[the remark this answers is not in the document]
asalsys: we are moving away from feature flag v2

DDDDDanica:
@asalsys Thanks for the change, could you also update the doc to reflect new API？
https://github.com/MetaMask/contributor-docs/blob/main/docs/remote-feature-flags.md?plain=1#L97

asalsys:
@DDDDDanica I updated the docs in this pr: https://github.com/MetaMask/contributor-docs/pull/177

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -314,6 +310,24 @@ export class RemoteFeatureFlagController extends BaseController<
...
+    }
+
+    for (const flagName of Object.keys(updatedFeatureFlagThresholdGroups)) {
+      if (!currentFlagNames.includes(flagName)) {
+        delete updatedFeatureFlagThresholdGroups[flagName];
+      }
+    }
+
     // Single state update with all changes batched together
     this.#processedRemoteFeatureFlags = processedFlags;
```
Gudahtt, line 315: I noticed that the way we're updating state here is suboptimal. We're replacing the entire state each time, rather than updating just the properties that have changed.

But since it's a pre-existing problem, better not to solve it in this PR. Maybe we can address it in a later PR.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -327,6 +341,7 @@ export class RemoteFeatureFlagController extends BaseController<
         rawRemoteFeatureFlags: remoteFeatureFlags,
         cacheTimestamp: Date.now(),
         thresholdCache: updatedThresholdCache,
+        featureFlagThresholdGroups: updatedFeatureFlagThresholdGroups,
```
Gudahtt: Nit: The `updatedFeatureFlagThresholdGroups` intermediate variable we're creating here is identical to `featureFlagThresholdGroupUpdates`. We can just use `featureFlagThresholdGroupUpdates`
```suggestion
        featureFlagThresholdGroups: featureFlagThresholdGroupUpdates,
```

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -478,15 +478,15 @@ describe('RemoteFeatureFlagController', () => {
...
+      expect(controller.state.remoteFeatureFlags.testFlagForThreshold).toBe(
+        'valueB',
+      );
+      expect(controller.state.featureFlagThresholdGroups).toStrictEqual({
+        testFlagForThreshold: 'groupB',
       });
     });
 
-    it('preserves selected legacy threshold object value wrappers', async () => {
+    it('stores selected threshold group name from name field when thresholdName is absent', async () => {
```
Gudahtt: Nit: This test case seems redundant now that we've removed the "threshold v2" feature

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -515,13 +515,13 @@ describe('RemoteFeatureFlagController', () => {
...
-        name: 'enabled',
-        value: thresholdFlagValue,
+      ).toStrictEqual(thresholdFlagValue);
+      expect(controller.state.featureFlagThresholdGroups).toStrictEqual({
+        thresholdObjectFlag: 'enabled',
       });
     });
 
-    it('returns selected threshold version 2 values without wrapper metadata', async () => {
+    it('does not map threshold group when only thresholdName is provided', async () => {
```
Gudahtt: Nit: This test case seems redundant now that we've removed the "threshold v2" feature

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -552,9 +552,10 @@ describe('RemoteFeatureFlagController', () => {
       expect(
         controller.state.remoteFeatureFlags.thresholdObjectFlag,
       ).toStrictEqual(thresholdFlagValue);
+      expect(controller.state.featureFlagThresholdGroups).toStrictEqual({});
     });
 
-    it('falls back to legacy threshold wrappers for unrecognized threshold versions', async () => {
+    it('returns selected threshold values for unrecognized threshold versions', async () => {
```
Gudahtt: Nit: This test case seems redundant now that we've removed the "threshold v2" feature

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
```diff
@@ -1406,6 +1419,72 @@ describe('RemoteFeatureFlagController', () => {
...
+      await messenger.call(
+        'RemoteFeatureFlagController:updateRemoteFeatureFlags',
+      );
+      await flushPromises();
+
+      expect(controller.state.featureFlagThresholdGroups).toStrictEqual({
+        flagB: 'groupB',
+      });
+
+      jest.useRealTimers();
```
Gudahtt: It would be safer to put this in an `afterEach` hook or in a `try.. finally` block, so that it still runs even if there is a crash.

if this test crashes _after_ file timers are enabled but _before_ restoring real timers, the fake timers could remain in-place for later tests as well, causing confusing cascading test failures.

Gudahtt (changes requested):
Generally looks good! At least one change needed for the test suite, but the rest of my comments are nits

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.test.ts
Gudahtt: Actually nevermind, maybe we keep this one so that we ensure the `thresholdVersion` flag is ignored and does not interfere with feature flag processing? We'd want to re-word the decsription though.

Really _all_ `thresholdVersion` values are "unrecognized" now. So we could say, for example:

```suggestion
    it('ignores threshold version field, processing threshold values normally', async () => {
```

Gudahtt (approved):
LGTM!

PR #9290, opened 2026-06-26
feat(wallet): add opt-in ramps initialization configs

georgeweiler:
## Summary
- Add opt-in `@metamask/wallet` initialization configs for `RampsService`, `RampsController`, and `TransakService` under `@metamask/wallet/initialization/instances/ramps`.
- Bump `@metamask/wallet` to 4.1.0 and add `@metamask/ramps-controller` as a dependency.
- Include unit tests for all three configs; no changes to the `@metamask/ramps-controller` package itself.

## Test plan
- [x] `yarn workspace @metamask/wallet run jest --no-coverage --testPathPattern=ramps --watchman=false`
- [ ] `yarn workspace @metamask/wallet run build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "1077.0.0",
+  "version": "1078.0.0",
   "private": true,
   "description": "Monorepo for packages shared between MetaMask clients",
   "repository": {
--- a/packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
+++ b/packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
@@ -6,6 +6,14 @@ import {
 
 import { InitializationConfiguration } from '../../../types';
 
+/**
+ * Opt-in initialization config for `RampsService`.
+ *
+ * **Prerequisite:** `AuthenticationController` must be registered on the root
+ * messenger before this config is used — `RampsService` delegates
+ * `AuthenticationController:getBearerToken` and will throw at runtime if that
+ * action is not available.
+ */
 export const rampsService: InitializationConfiguration<
   RampsService,
   RampsServiceMessenger
--- a/packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
+++ b/packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
@@ -29,14 +29,18 @@ export const rampsService: InitializationConfiguration<
       baseUrlOverride: options.baseUrlOverride,
     }),
   getMessenger: (parent) => {
+    // Type cast required: RampsServiceMessenger's parent constraint includes
+    // AuthenticationController:getBearerToken which is not in DefaultActions
+    // (these are opt-in configs, not default wallet instances).
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const messenger: RampsServiceMessenger = new Messenger({
       namespace: 'RampsService',
-      parent,
+      parent: parent as never,
     });
 
     parent.delegate({
       messenger,
-      actions: ['AuthenticationController:getBearerToken'],
+      actions: ['AuthenticationController:getBearerToken'] as never[],
       events: [],
     });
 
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/core-monorepo",
-  "version": "1077.0.0",
+  "version": "1078.0.0",
```
sirtimid, line 3:
```suggestion
  "version": "1077.0.0",
```

Revert the root + wallet version bumps, move the two entries back under ## [Unreleased] (keep 5.0.0 section verbatim, no [5.1.0] link/section), and drop the wallet-cli dep bump. Touching the root package.json version is exclusively `create-release-branch`'s job, no feature PR should do it.

packages/wallet/src/types.ts
```diff
@@ -28,6 +31,9 @@ export type InstanceSpecificOptions = {
   connectivityController: ConnectivityControllerInstanceOptions;
   keyringController?: KeyringControllerInstanceOptions;
   networkController: NetworkControllerInstanceOptions;
+  rampsController?: RampsControllerInstanceOptions;
+  rampsService?: RampsServiceInstanceOptions;
   remoteFeatureFlagController: RemoteFeatureFlagControllerInstanceOptions;
   storageService: StorageServiceInstanceOptions;
+  transakService?: TransakServiceInstanceOptions;
```
sirtimid, line 38: Here these are typed as optional but `RampsServiceInstanceOptions.context/fetch` and `TransakServiceInstanceOptions.context/fetch` are required. Consider validating the required options in init and throwing a clear error.

packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
```diff
@@ -0,0 +1,49 @@
...
+    // (these are opt-in configs, not default wallet instances).
+
+    const messenger: RampsServiceMessenger = new Messenger({
+      namespace: 'RampsService',
+      parent: parent as never,
+    });
+
+    parent.delegate({
+      messenger,
+      actions: ['AuthenticationController:getBearerToken'] as never[],
```
sirtimid, line 43: The `as never[]` casts seem like an acceptable unavoidable workaround, but please contain them in a shared `createOptInMessenger` helper so the delegated action names stay type-checked, as written, a typo in an action string compiles and only fails at runtime (`AuthenticationController:getBererToken` typo now compiles and only fails at runtime)

sirtimid (changes requested):

packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
```diff
@@ -0,0 +1,49 @@
...
+> = {
+  name: 'RampsService',
+  init: ({ messenger, options }) =>
+    new RampsService({
+      messenger,
+      environment: options.environment,
+      context: options.context,
+      fetch: options.fetch,
+      policyOptions: options.policyOptions,
+      baseUrlOverride: options.baseUrlOverride,
```
FrederikBolding, line 29: Are none of these able to be standardized across clients?

packages/wallet/src/initialization/instances/ramps/transak-service/transak-service.ts
```diff
@@ -0,0 +1,32 @@
...
+  name: 'TransakService',
+  init: ({ messenger, options }) =>
+    new TransakService({
+      messenger,
+      environment: options.environment,
+      context: options.context,
+      fetch: options.fetch,
+      apiKey: options.apiKey,
+      policyOptions: options.policyOptions,
+      orderRetryDelayMs: options.orderRetryDelayMs,
```
FrederikBolding, line 22: Same question here, can none of these be standardized?

packages/wallet/src/types.ts
FrederikBolding: If any of the underlying instance options are required, please mark them as required here too.

packages/wallet/src/initialization/instances/ramps/ramps-service/ramps-service.ts
```diff
@@ -0,0 +1,49 @@
+import { Messenger } from '@metamask/messenger';
+import {
+  RampsService,
+  RampsServiceMessenger,
+} from '@metamask/ramps-controller';
+
+import { InitializationConfiguration } from '../../../types';
+
+/**
+ * Opt-in initialization config for `RampsService`.
```
FrederikBolding, line 10: Why would we not want all of these initialization configs to be included by default?

FrederikBolding (changes requested):

package.json
FrederikBolding: ++

PR #9294, opened 2026-06-26
feat: add getSessionProperties to `chain-agnostic-permissions`. Use it in `multichain-api-middleware`

jiexi:
## Explanation

Adds `getSessionProperties` to `chain-agnostic-permissions` and then uses it in `wallet_getSession` and `wallet_createSession` handlers in `multichain-api-middleware`. Effectively this change makes it so that CAIP session `result.sessionProperties` now include a `eip155Capabilities` field which contains the `wallet_getCapabilities` result for each `eip155` account in the permission.

This is needed to help make the CAIP Multichain API handshake truly a one request handshake as without this, many dapps need to make separate `wallet_getCapabilities` calls. This is problematic for the MWP flow in MetaMask Connect mainly where we want to avoid making unnecessary calls as each call that must resolve in the wallet forces the user out of the dapp and into the wallet causing poor UX.

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1560

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking middleware hook contract and changes session payload shape dapps consume; capability lookup failures are non-fatal per address, which may surprise callers expecting full maps.
> 
> **Overview**
> Adds **`getSessionProperties`** in `chain-agnostic-permission`, which merges stored CAIP-25 `sessionProperties` with an **`eip155Capabilities`** map: for each unique permitted EVM address (via `getEthAccounts`), it calls an injected **`getCapabilities`** hook and keys results by address. Failures for a single address are logged and that address is omitted from the map.
> 
> **`multichain-api-middleware`** now uses this for **`wallet_getSession`** and **`wallet_createSession`** instead of returning only persisted caveat properties. Both handlers require a new **`getCapabilities`** hook (**breaking** for integrators). **`wallet_getSession`** always includes **`sessionProperties`** in the result (including `{}` when there is no active session).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43e6d6596159241642c8615fbae0b89f214c4600. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jiexi:
@metamaskbot publish-previews

jiexi:
@metamaskbot publish-preview

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
```diff
@@ -143,6 +144,43 @@ export const getSessionScopes = (
...
+ */
+export const getSessionProperties = (
+  caip25CaveatValue: Pick<
+    Caip25CaveatValue,
+    'requiredScopes' | 'optionalScopes' | 'sessionProperties'
+  >,
+  {
+    getCapabilities,
+  }: {
+    getCapabilities: (params: { address: string }) => unknown;
```
adonesky1: is the sync signature on purpose? the handlers here already await async hooks like [requestPermissionsForOrigin](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/multichain-api-middleware/src/handlers/wallet-createSession.ts#L258) and [handleNonEvmRequestForOrigin](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/multichain-api-middleware/src/handlers/wallet-invokeMethod.ts#L154), so going async would be pretty low-friction, you'd just have [getSessionProperties](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts#L157) return a Promise. are we planning a separate sync cache-backed thing, or should this just match the async [getCapabilities](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/eip-5792-middleware/src/hooks/getCapabilities.ts#L41) over in eip-5792-middleware?

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
```diff
@@ -107,7 +108,7 @@ const getNormalizedScopesObject = (
 export const getSessionScopes = (
   caip25CaveatValue: Pick<
     Caip25CaveatValue,
-    'requiredScopes' | 'optionalScopes'
+    'requiredScopes' | 'optionalScopes' | 'sessionProperties'
```
adonesky1: this seems like a left over mistake?
```suggestion
    'requiredScopes' | 'optionalScopes'
```

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
```diff
@@ -143,6 +144,43 @@ export const getSessionScopes = (
...
+  caip25CaveatValue: Pick<
+    Caip25CaveatValue,
+    'requiredScopes' | 'optionalScopes' | 'sessionProperties'
+  >,
+  {
+    getCapabilities,
+  }: {
+    getCapabilities: (params: { address: string }) => unknown;
+  },
+): Record<string, unknown> => {
```
adonesky1: typing nit: this returns `Record<string, unknown>` and types `getCapabilities` as `unknown`, but session props are already `Record<string, Json>` everywhere in this package (e.g. `Caip25CaveatValue.sessionProperties`), which you're spreading in right below. why not type `getCapabilities` as `=> Json` and return `Record<string, Json>`? stays consistent with the rest of the module and kills the `as Record<string, Json>` casts at both call sites ([getSession](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/multichain-api-middleware/src/handlers/wallet-getSession.ts#L84), [createSession](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/multichain-api-middleware/src/handlers/wallet-createSession.ts#L293)). `Json` also would've caught the [`jest.fn()` returning `undefined`](https://github.com/MetaMask/core/blob/e1c7d2838f7a61c26abd727ec984dd9fb086010e/packages/multichain-api-middleware/src/handlers/wallet-createSession.test.ts#L111) case, which silently drops on serialization today.

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
jiexi: should be async. good catch

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
jiexi: fair point!

jiexi:
@metamaskbot publish-previews

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
```diff
@@ -143,6 +144,45 @@ export const getSessionScopes = (
...
+): Promise<Record<string, Json>> => {
+  const addresses = getEthAccounts(caip25CaveatValue);
+
+  const eip155Capabilities: Record<Hex, Record<string, Json>> = {};
+
+  await Promise.all(
+    addresses.map(async (address) => {
+      eip155Capabilities[address] = await getCapabilities({ address });
+    }),
+  );
```
adonesky1, line 184: This is unguarded, so if `getCapabilities` rejects for any single address the whole `Promise.all` rejects, `getSessionProperties` throws, and both `wallet_getSession` and `wallet_createSession` fail. Should we make this best-effort, e.g. try/catch per address and default to `{}` on failure?

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
```diff
@@ -143,6 +144,45 @@ export const getSessionScopes = (
...
+  >,
+  {
+    getCapabilities,
+  }: {
+    getCapabilities: (params: {
+      address: string;
+    }) => Promise<Record<Hex, Record<string, Json>>>;
+  },
+): Promise<Record<string, Json>> => {
+  const addresses = getEthAccounts(caip25CaveatValue);
```
adonesky1, line 170: How are you feeling about making returning capabilities conditional on callers passing a new sessionProperty param in the request?

adonesky1:
Do we need to include these capabilties in `wallet_sessionChanged` returns as well?

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
adonesky1: Also Should we lowercase the addresses here? The real `wallet_getCapabilities` already [normalizes to lowercase](https://github.com/MetaMask/core/blob/bc74c062037d07e882aa922ef677fab6af66898d/packages/eip-5792-middleware/src/utils.ts#L71) via [validateAndNormalizeKeyholder](https://github.com/MetaMask/core/blob/bc74c062037d07e882aa922ef677fab6af66898d/packages/eip-5792-middleware/src/utils.ts#L52), but [getEthAccounts](https://github.com/MetaMask/core/blob/bc74c062037d07e882aa922ef677fab6af66898d/packages/chain-agnostic-permission/src/operators/caip-permission-operator-accounts.ts#L77) returns them in the caveat's casing (likely checksummed), so we'd be keying eip155Capabilities differently than how that method treats the same addresses. Seems worth standardizing on the lowercase form here (key + the address passed to `getCapabilities`) so both paths agree, otherwise anything resolving these caps by address has to know which casing to expect.

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
jiexi: yeah, i thought about if it was worth a try catch or not. I suppose we shouldn't cause the whole request to fail on this

packages/chain-agnostic-permission/src/operators/caip-permission-operator-session-scopes.ts
jiexi: > How are you feeling about making returning capabilities conditional on callers passing a new sessionProperty param in the request?

Still feel like it isn't worth the trouble, but if you feel strongly still then we can choose a new property name and check it here

> Also Should we lowercase the addresses here?

`wallet_getCapabilities` doesn't return the account address in its result, so I think this is up to us to interpret

jiexi:
yeah, `wallet_sessionChanged` implementation is on the client side though

jiexi:
@metamaskbot publish-previews

adonesky1 (approved):
LGTM!

mcmire (approved):
LGTM.

PR #9306, opened 2026-06-29
test(wallet-cli): add subprocess daemon lifecycle e2e

sirtimid:
## What

Adds a subprocess-level end-to-end test for the `mm daemon` command suite: it spawns the **built** `mm` CLI as real child processes against a throwaway data directory and drives the actual `start → call → status/stop/purge` lifecycle over the Unix socket.

This closes the one gap the existing tests can't reach. `socket-integration.test.ts` exercises the transport in the test realm, and `wallet-factory.e2e.test.ts` constructs a real `Wallet` in-process — but nothing tested the daemon **as a running process**. Until now that flow was only a manual smoke check.

## What it covers

- **`start`** brings the daemon up and reports its socket; a **second `start`** finds the responsive daemon and reports "already running" (the live-sibling guard).
- **`call KeyringController:getState`** over the socket returns the SRP-derived account (`isUnlocked: true`, `0x…` address).
- **`status`** reports `PID + Uptime`; **`stop`** tears it down; a follow-up `status` reports not-running.
- **Persistence across restart** — after `stop` + `start`, the wallet resumes from the on-disk vault instead of re-importing the SRP. Proven by the observable signature of the resume path: first run is **unlocked**, the restart comes back **locked** with the vault still on disk (a re-import would have left it unlocked).
- **`purge --force`** deletes the daemon state files.
- The socket is owner-only (`0600`) and the data dir is `0700`.

Each test uses a fresh temp data dir and `afterEach` guarantees teardown (SIGKILL by the recorded PID + `rm -rf`), so a failed run can't leak a daemon or socket.

Offline-safe: daemon startup neither fetches feature flags nor looks up the network, and the only action called is local — so it runs without network access.

## How it's isolated

It needs the built `dist/` and the native `better-sqlite3` addon, so it must stay out of the fast unit run:

- Lives in `*.daemon-e2e.test.ts` with its own `jest.config.e2e.js` (no coverage threshold), run via a new `yarn test:e2e` script (`build` → `test:prepare` → jest).
- Excluded from `jest.config.js`'s default run **and** from coverage collection, so the unit suite stays at 100%.

## CI

The existing per-package `test-*` matrix runs each workspace's `test` against source with no build, so it can't host this suite. A dedicated `test-wallet-cli-e2e` job (Node 20.x + 24.x) builds the wallet-cli dependency subtree, then runs `yarn workspace @metamask/wallet-cli test:e2e`.

> Whether this job should be a **required** check is a branch-protection setting (outside this file) — worth deciding before merge.

## Notes

- **No changelog entry** — this is test infrastructure only, nothing consumer-facing.
- The tests that want a live transport (full `SIGTERM→SIGKILL` ordering, chunked-payload framing, id-correlation under out-of-order completion) are intentionally **not** in scope here — lower value, flakier; better as a focused follow-up.

---

---

> [!NOTE]
> **Medium Risk**
> Changes daemon spawn, stop, and socket I/O behavior used in production CLI flows, not only tests; incorrect stale-socket or signal logic could affect local daemon state.
> 
> **Overview**
> Adds a **subprocess e2e** path for `wallet-cli`: `tests/lifecycle.e2e.test.ts` drives the built `mm` CLI through `start → call → status/stop/purge` (including vault resume and `0600`/`0700` checks), with a separate `jest.config.e2e.js`, `yarn test:e2e`, unit config ignoring `tests/`, and a CI job that builds the package subtree then runs e2e on Node 20/22/24.
> 
> **Daemon runtime fixes** shipped alongside the suite: `ensureDaemon` creates the data dir before opening the log, redirects detached stderr to the log file, and prefers spawn `error` over `exit` during startup; `stopDaemon` removes stale socket + PID when the socket is refused and the PID is dead (but not on permission errors), and removes the socket after a successful stop; `writeLine` keeps an error listener after a failed write callback so trailing `EPIPE` does not crash the process.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56d3cf9e86438c933ec2ccceb08c20bd6231139b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/lint-build-test.yml
```diff
@@ -203,3 +203,33 @@
...
+  test-wallet-cli-e2e:
+    name: Test wallet-cli daemon e2e (${{ matrix.node-version }})
+    runs-on: ubuntu-latest
+    needs: prepare
+    strategy:
+      matrix:
+        node-version: [20.x, 24.x]
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v2
```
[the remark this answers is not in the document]
grypez: This change is aligned with existing convention for the file

.github/workflows/lint-build-test.yml
```diff
@@ -203,3 +203,33 @@
...
+      - name: Build wallet-cli and its dependencies
+        run: yarn workspaces foreach --topological-dev --recursive --from '@metamask/wallet-cli' run build
+      - run: yarn workspace @metamask/wallet-cli run test:e2e
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty at end of job"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
grypez: Out of scope for this PR

packages/wallet-cli/src/daemon/daemon-spawn.test.ts
```diff
@@ -1,21 +1,30 @@
...
 
 const mockSpawn = jest.mocked(spawn);
 const mockExistsSync = jest.mocked(existsSync);
+const mockOpenSync = jest.mocked(openSync);
+const mockCloseSync = jest.mocked(closeSync);
+const mockEnsureOwnerOnlyDirectory = jest.mocked(ensureOwnerOnlyDirectory);
 const mockPingDaemon = jest.mocked(pingDaemon);
 const mockGetDaemonPaths = jest.mocked(getDaemonPaths);
 
+// Arbitrary fd handed back by the mocked `openSync` so tests can assert it is
```
grypez: Nit: can we write file descriptor instead of fd throughout?

grypez (dismissed):
Looks good! Ignore the bots

grypez (dismissed):

.github/workflows/lint-build-test.yml
```diff
@@ -207,14 +191,38 @@ jobs:
...
+  # The wallet-cli daemon e2e spawns the BUILT `mm` CLI and the native
+  # better-sqlite3 addon as real child processes, so it needs its dependency
+  # subtree built first and cannot run in the per-package `test-*` matrix above.
+  test-wallet-cli-e2e:
+    name: Test wallet-cli daemon e2e (${{ matrix.node-version }})
+    runs-on: ubuntu-latest
+    needs: prepare
+    strategy:
+      matrix:
+        node-version: [20.x, 24.x]
```
Mrtenz: Why are we not testing on 22.x?

.github/workflows/lint-build-test.yml
```diff
@@ -207,14 +191,38 @@ jobs:
...
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v2
```
Mrtenz:
```suggestion
        uses: MetaMask/action-checkout-and-setup@v3
```

.github/workflows/lint-build-test.yml
```diff
@@ -207,14 +191,38 @@ jobs:
...
+        with:
+          is-high-risk-environment: false
+          node-version: ${{ matrix.node-version }}
```
Mrtenz, line 242:
```suggestion
          node-version: ${{ matrix.node-version }}
          persist-credentials: false
```

Mrtenz (changes requested):

.github/workflows/lint-build-test.yml
sirtimid: Good catch! Done https://github.com/MetaMask/core/pull/9306/commits/9b0489caed797b1f0454710e25aca3c099b7d409

.github/workflows/lint-build-test.yml
sirtimid: Done https://github.com/MetaMask/core/pull/9306/commits/9b0489caed797b1f0454710e25aca3c099b7d409

.github/workflows/lint-build-test.yml
sirtimid: Done https://github.com/MetaMask/core/pull/9306/commits/9b0489caed797b1f0454710e25aca3c099b7d409

packages/wallet-cli/package.json
```diff
@@ -39,6 +39,7 @@
     "test:prepare": "./scripts/install-binaries.sh",
     "test": "yarn test:prepare && NODE_OPTIONS=--experimental-vm-modules jest --reporters=jest-silent-reporter",
     "test:clean": "NODE_OPTIONS=--experimental-vm-modules jest --clearCache",
+    "test:e2e": "yarn build && yarn test:prepare && NODE_OPTIONS=--experimental-vm-modules jest --config jest.config.e2e.js",
```
Mrtenz: Do we need to build again here? It's building the package twice now in CI.

.github/workflows/lint-build-test.yml
```diff
@@ -222,3 +222,31 @@ jobs:
...
+        node-version: [20.x, 22.x, 24.x]
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: false
+          persist-credentials: false
+          node-version: ${{ matrix.node-version }}
+      - name: Build wallet-cli and its dependencies
+        run: yarn workspaces foreach --topological-dev --recursive --from '@metamask/wallet-cli' run build
```
Mrtenz, line 244: Wondering if instead of building here, we should make this job depend on the build step, and use artifacts to upload/download the build files.

This is maybe faster though. 🤔

packages/wallet-cli/package.json
sirtimid: Hmm yeah, I did this to be easier for devs, but one would know that one needs to build first for e2e tests, I ll change it

.github/workflows/lint-build-test.yml
sirtimid: I would be surprised if this wasn't faster.

packages/wallet-cli/package.json
sirtimid: Done https://github.com/MetaMask/core/pull/9306/commits/56d3cf9e86438c933ec2ccceb08c20bd6231139b

Mrtenz (approved):

PR #9328, opened 2026-07-01
ci: build platform-api-docs in parallel, deploy only on main

cryptodev-2s:
## Explanation

Follow up to the platform-api-docs CI wiring. Addresses review feedback that the docs build added several minutes to total CI time.

Before, `deploy-platform-api-docs` was its own job that `needs: lint-build-test`. Its `build-docs` step ran on every PR but could only start after the whole lint/build/test matrix finished, so it was matrix time plus docs build time in series.

## Changes

- `lint-build-test.yml`: added a `build-platform-api-docs` job that runs in parallel with lint/build/test and uploads the site artifact. On PRs this adds no serial time, and it still gates PR success (a broken docs build fails `lint-build-test`).
- `deploy-platform-api-docs.yml`: dropped the `build-docs` job. It now keeps only `get-token` and `deploy`, which consume the artifact produced by the build job in the same run. Removed the per job `if` guards since the caller gates the whole workflow.
- `main.yml`: gated the `deploy-platform-api-docs` call on `main` pushes, and removed it from `all-jobs-complete` (same treatment as the other main only jobs like `publish-release`).

Net effect: PRs build the docs in parallel and skip deploy entirely. On `main` pushes the build runs in parallel too, then deploy publishes after the full suite passes.

## References

- Follow up to #8012

---

> [!NOTE]
> **Low Risk**
> Workflow-only changes with no application code; deploy still uses the same build command and gh-pages settings, with slightly different job structure and environments.
> 
> **Overview**
> **CI timing:** Platform API docs are built in a new **`build-platform-api-docs`** job inside `lint-build-test.yml`, so that work runs **in parallel** with lint/build/test instead of waiting for the full matrix. PRs still fail if the docs build breaks, without adding serial minutes after the matrix.
> 
> **Deploy workflow:** `deploy-platform-api-docs.yml` is reduced to a **single job** that checks out, runs `yarn docs:platform-api:build` (with GitHub Pages site URL/base), fetches a publish token, and deploys from `.platform-api-docs/build`. The old **artifact upload/download** and separate **`get-token`** job are removed; token and deploy run in the same job.
> 
> **Main pipeline:** In `main.yml`, **`deploy-platform-api-docs`** is gated with **`if: main` push** only (after `lint-build-test`), and it is **dropped from `all-jobs-complete`** so PR/merge-queue completion no longer waits on Pages deploy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7a2d74883ae59e712d63fc0b5a534071b2d9b66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/deploy-platform-api-docs.yml
```diff
@@ -4,22 +4,35 @@ on:
...
+      # Token Exchange Service; `contents: read` is for the checkout. The push
+      # to `gh-pages` uses the exchanged PAT, not GITHUB_TOKEN.
+      id-token: write
       contents: read
     steps:
+      # Build the site fresh here rather than reusing the artifact from the
+      # `build-platform-api-docs` job in `lint-build-test`: that job also runs
+      # for PRs, so trusting its output (or the caches behind it) for a
+      # production deploy would risk publishing poisoned content.
+      # `is-high-risk-environment` isolates the build.
```
Mrtenz: These comments seem redundant.

Mrtenz (approved):

PR #9344, opened 2026-07-01
feat(assets-controller): use AccountsAPI v6 balances

Kriys94:
## Explanation

The `AccountsApiDataSource` currently fetches account balances from the Accounts API **v5** multi-account endpoint. This PR adds support for the **v6** endpoint and lets it run **at parity with v5**, gated behind the `assetsAccountsApiV6` remote feature flag (default off → v5).

What changed:

- **Flag-gated endpoint selection.** When the `assetsAccountsApiV6` remote flag is `true`, balances are fetched via `fetchV6MultiAccountBalances`; otherwise the legacy v5 endpoint is used. The flag is read from `RemoteFeatureFlagController` **on every fetch (not cached)**, so v6 can be rolled out — and instantly rolled back to v5 — remotely without a client release.
- **Parity with v5.** Only `category: 'token'` rows from the v6 response are consumed (DeFi positions are ignored), so token-balance behavior is identical to v5. Unprocessed networks and fetch timeouts still surface as per-chain errors for RPC fallback, exactly as before.
- **Wiring.** `AccountsApiDataSource` now takes the shared `AssetsController` messenger and calls `RemoteFeatureFlagController:getState`; `AssetsController` passes the messenger through and extends `AllowedActions` accordingly. Adds `@metamask/remote-feature-flag-controller` as a dependency.
- **Test override.** An optional `useBalanceV6` getter on `AccountsApiDataSourceConfig` overrides the remote-flag-driven selection (mainly for tests); when omitted, the remote flag is used.

Because the default is v5 and v6 produces the same token balances, there is no user-facing behavior change until the flag is enabled.

## References

- Consumer integration (extension): wires `RemoteFeatureFlagController:getState` into the Assets controller messenger and bumps `@metamask/core-backend` / `@metamask/assets-controller`.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes which backend balances API runs and how responses are normalized; incorrect flag wiring or v6 parsing could affect displayed token balances, though rollout is flag-gated and defaults to v5.
> 
> **Overview**
> **Accounts API balance fetches can use the v6 multi-account endpoint** when the remote flag `assetsAccountsApiV6` is true; otherwise behavior stays on v5. The choice is evaluated on each fetch (not cached) so clients can roll out or roll back without redeploying, with an optional `useBalanceV6` getter on `AccountsApiDataSourceConfig` to override the flag (mainly for tests).
> 
> `AccountsApiDataSource` now **requires the shared `AssetsController` messenger** and calls `RemoteFeatureFlagController:getState`; `AssetsController` passes that messenger through and extends its allowed actions accordingly. The package adds **`@metamask/remote-feature-flag-controller`** as a dependency.
> 
> On the v6 path, balances are loaded via `fetchV6MultiAccountBalances`, mapped from the per-account response shape, and **only `category: 'token'` rows are applied** (DeFi positions are skipped to match v5). Custom assets are forwarded as **`includeAssetIds`** for ERC-20 tokens on chains in the current request (native and off-chain assets filtered out). Unprocessed networks and timeouts still surface as chain errors for RPC fallback, same as v5.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 766168bac7620399bc34e6ad92ac385e02dbce59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

> [!NOTE]
> **Medium Risk**
> Switches the balances API and response normalization when the flag is on, which can affect displayed token balances; risk is mitigated because v5 remains the default and rollout is flag-gated.
> 
> **Overview**
> **`AccountsApiDataSource` can call the Accounts API v6 multi-account balances endpoint** when the remote flag `assetsAccountsApiV6` is enabled (`{ value: true }` from `RemoteFeatureFlagController`); otherwise it keeps using v5. The choice is evaluated on **each fetch** (not cached), so rollout and rollback are remote-controlled with default behavior unchanged (v5).
> 
> Wiring adds a **required shared `AssetsController` messenger** on `AccountsApiDataSource`, delegates **`RemoteFeatureFlagController:getState`** through `AssetsController` allowed actions, and adds **`@metamask/remote-feature-flag-controller`** (plus README/tsconfig/yarn updates). V6 responses are normalized via **`#processV6Balances`**, applying only **`category: 'token'`** rows so DeFi positions are skipped for v5 parity; unprocessed networks and timeouts still become per-chain errors for RPC fallback.
> 
> Tests cover flag on/off/runtime toggle, v6 parsing, and an **`AssetsController` integration** case that asserts v6 is used when the flag is set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69c686b744f5044d4aff66fa550614d472f81f8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
+++ b/packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
@@ -7,7 +7,6 @@ import type { RemoteFeatureFlagControllerGetStateAction } from '@metamask/remote
 import {
   isCaipChainId,
   KnownCaipNamespace,
-  parseCaipAssetType,
   toCaipChainId,
 } from '@metamask/utils';
 
@@ -79,17 +78,6 @@ export type AccountsApiDataSourceConfig = {
    * middleware hands them off to the next data source (e.g. RPC fallback).
    */
   fetchTimeoutMs?: number;
-  /**
-   * Optional override for the Accounts API balances endpoint version.
-   *
-   * When omitted, the data source reads the `assetsAccountsApiV6` remote feature
-   * flag from the RemoteFeatureFlagController on every fetch (defaulting to v5
-   * when the flag is unset or unavailable). Provide a getter to bypass the flag
-   * (e.g. in tests). When it returns true the v6 endpoint is used; otherwise the
-   * legacy v5 endpoint is used. Kept as a getter so the value can be toggled at
-   * runtime without re-instantiating the data source.
-   */
-  useBalanceV6?: () => boolean;
 };
 
 export type AccountsApiDataSourceOptions = AccountsApiDataSourceConfig & {
@@ -216,13 +204,6 @@ export class AccountsApiDataSource extends AbstractDataSource<
   /** Getter avoids stale value when user toggles token detection at runtime. */
   readonly #tokenDetectionEnabled: () => boolean;
 
-  /**
-   * Feature flag getter selecting the balances endpoint version. Kept as a
-   * getter so it can be flipped at runtime (e.g. to revert v6 -> v5). Defaults
-   * to reading the `assetsAccountsApiV6` remote feature flag off the messenger.
-   */
-  readonly #useBalanceV6: () => boolean;
-
   /** Shared AssetsController messenger, used to read remote feature flags. */
   readonly #messenger: AssetsControllerMessenger;
 
@@ -247,8 +228,6 @@ export class AccountsApiDataSource extends AbstractDataSource<
     this.#tokenDetectionEnabled =
       options.tokenDetectionEnabled ?? ((): boolean => true);
     this.#messenger = options.messenger;
-    this.#useBalanceV6 =
-      options.useBalanceV6 ?? ((): boolean => this.#isBalanceV6Enabled());
     this.#apiClient = options.queryApiClient;
 
     this.#initializeActiveChains().catch(console.error);
@@ -261,6 +240,10 @@ export class AccountsApiDataSource extends AbstractDataSource<
    * the same across clients (extension, mobile). Defaults to `false` when the
    * flag is unset or the controller is unavailable.
    *
+   * The flag is a LaunchDarkly JSON variation shaped `{ value: boolean }` (same
+   * shape as `backendWebSocketConnection`), so the nested `value` is read rather
+   * than treating the flag as a plain boolean.
+   *
    * @returns `true` when the v6 balances endpoint should be used.
    */
   #isBalanceV6Enabled(): boolean {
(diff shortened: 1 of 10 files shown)


packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -212,11 +245,43 @@ export class AccountsApiDataSource extends AbstractDataSource<
     this.#fetchTimeoutMs = options.fetchTimeoutMs ?? DEFAULT_FETCH_TIMEOUT_MS;
     this.#tokenDetectionEnabled =
       options.tokenDetectionEnabled ?? ((): boolean => true);
+    this.#messenger = options.messenger;
+    this.#useBalanceV6 =
+      options.useBalanceV6 ?? ((): boolean => this.#isBalanceV6Enabled());
```
bergarces: Why have both a parameter (that is not being used) and a default? It's adding code surface that makes it more difficult to find the source and needs to be tested.

If we pass it down as a method there are certain advantages:
- It can be tested in isolation
- It stops requiring passing down the messenger.

Kriys94: Agree!

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -449,6 +590,73 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+        const normalizedAssetId = normalizeAssetId(
+          item.assetId as Caip19AssetId,
+        );
+
+        // Store balance as returned by API
+        assetsBalance[accountId][normalizedAssetId] = {
+          amount: item.balance,
+        };
+      }
+    }
```
bergarces, line 635: Not blocking for this PR, but I'm noticing that most of the tests for the logic-heavy `processV6Balances` and `processV5Balances` are just happy paths with a single account and a single balance.

Maybe it would make sense in a follow-up ticket to move this logic outside and test it against some responses with more edge cases (for v6 only, not point doing it for v5 if we are going to drop it soon). Ideally, we would want to take the opportunity to see if there is a suitable parser in core that we can use to generate a schema for the api response and validate it beforehand to protect ourselves from runtime errors (this could even be done at core-backend level).

That would leave this class just responsible of dealing with a well-known and safe piece of data.

But again, this is more of an afterthought to reduce the cognitive load when trying to work with this data source middleware.

bergarces (dismissed):

bergarces (dismissed):

bergarces (approved):

PR #9350, opened 2026-07-01
Copy createServicePolicy to base-data-service

mcmire:
## Explanation

Currently `createServicePolicy` lives in `controller-utils`, as originally we wrote it as an abstract utility that could theoretically be used for any kind of data access. However, there are a few reasons why this is suboptimal now:

1. `controller-utils` is too large and serves as a dependent of almost every package in the monorepo, so extreme care must be taken when changing this package to avoid breaking changes.
2. `createServicePolicy` was added before we formalized data services. Now we have a `BaseDataService` class, and this class sets up `createServicePolicy` by default, so it's highly unlikely that `createServicePolicy` will be used independently going forward.
3. Also, we only use `createServicePolicy` for HTTP requests, so making it abstract is kind of unnecessary.

This commit deprecates the existing exports in `controller-utils`, copies them to `base-data-service`, and changes `BaseDataService` to use the copied version instead.

In the future we will remove the symbols from `controller-utils` and modify `createServicePolicy` further so it works better for HTTP requests.

## References

No ticket yet, but this is related to https://consensyssoftware.atlassian.net/browse/WPC-1037 in that BaseDataService needs to be more fully-fledged to be as easy to use.

## Checklist

### DIFF
--- a/packages/base-data-service/package.json
+++ b/packages/base-data-service/package.json
@@ -58,6 +58,7 @@
     "@metamask/messenger": "^1.2.0",
     "@metamask/utils": "^11.11.0",
     "@tanstack/query-core": "^4.43.0",
+    "cockatiel": "^3.1.2",
     "fast-deep-equal": "^3.1.3"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


mcmire:
Superseded by #9418.

PR #9352, opened 2026-07-01
Add mixins for using polling in data services

mcmire:
## Explanation

The `polling-controller` package contains mixins for integrating the polling pattern into a class. Using these mixins on their own is somewhat cumbersome, so the package also contains utility functions which can be used in place of a superclass.

Currently, there are two styles of polling supported — one that subscribes to the block tracker, the other that runs on a cadence — and then past this, there are variants of these utility functions for controllers and non-controllers.

However, there is no variant for data services. That means if engineers want to integrate polling into data services, they are out of luck. This commit corrects that by adding two new mixins: `BlockTrackerPollingDataService` and `StaticIntervalPollingDataService`.

## References

Closes https://consensyssoftware.atlassian.net/browse/WPC-1038.

## Checklist

### DIFF
--- a/packages/polling-controller/src/BlockTrackerPollingController.ts
+++ b/packages/polling-controller/src/BlockTrackerPollingController.ts
@@ -1,4 +1,5 @@
 import { BaseController } from '@metamask/base-controller';
+import { BaseDataService } from '@metamask/base-data-service';
 import type {
   NetworkClientId,
   NetworkClient,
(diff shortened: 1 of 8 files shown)


packages/polling-controller/src/BlockTrackerPollingController.ts
```diff
@@ -103,3 +104,13 @@ export const BlockTrackerPollingController = <
   BlockTrackerPollingControllerMixin<typeof BaseController, PollingInput>(
     BaseController,
   );
+
+export const BlockTrackerPollingDataService = <
```
mcmire, line 108: I thought about placing these in `base-data-service` instead of this package, since we want to encourage engineers to use polling in data services and discourage them from using it in controllers. But then `base-data-service` would depend on `polling-controller`, and that seemed a little strange.

PR #9373, opened 2026-07-02
ci: build, test, lint only changed packages and dependants

Mrtenz:
## Explanation

Currently, every CI run rebuilds all packages from scratch and runs tests, changelog validation, and ESLint across the entire monorepo, regardless of how many packages actually changed.

This PR scopes TypeScript builds, tests, changelog validation, and ESLint to only the packages that changed plus their transitive dependants, by:

- Using the GitHub Compare API to find the exact merge base between the PR head and the target branch.
- Running `git diff --name-only` against that merge base to find changed files.
- Expanding the changed set to transitive dependants (so type-correctness across package boundaries is preserved), and also to transitive dependencies when building (so `ts-bridge` has all referenced `dist/` outputs available).
- Falling back to a full run when any file outside a package directory changes (e.g., root config files or workflow files), or when there is no merge base (push to `main`).

A new `scripts/get-changed-workspaces.mts` script computes the changed package set and is called from the `prepare` job (24.x matrix run only). Downstream jobs read the outputs from `prepare` directly — the separate `get-changed-packages` job has been removed, saving one sequential job hop.

The `changed-paths` output gates both ESLint scope and TypeScript build scope: when it is `"full"`, both run against all packages; when it is a JSON array of paths, each runs only against that subset.

### Benchmark PRs

Four benchmark PRs target this branch to verify the expected behaviour:

| PR | Change | Expected |
|---|---|---|
| [#9486](https://github.com/MetaMask/core/pull/9486) | CI-only change (workflow file comment) | Full run — workflow files are not in `IGNORED_ROOT_FILES` |
| [#9487](https://github.com/MetaMask/core/pull/9487) | Single package (`@metamask/logging-controller`) | 3 packages (`logging-controller` + 2 transitive dependants) |
| [#9488](https://github.com/MetaMask/core/pull/9488) | Three packages (`accounts-`, `gas-fee-`, `network-controller`) | 38 packages (those 3 + transitive dependants) |
| [#9489](https://github.com/MetaMask/core/pull/9489) | `README.md` (ignored) + `logging-controller` | 3 packages — `README.md` does not trigger a full run |

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how every PR is validated; incorrect workspace expansion or root-change detection could skip builds or tests, though impact is limited to CI, not shipped product code.
> 
> **Overview**
> **CI runs only what the PR touches** instead of rebuilding and testing the whole monorepo on every run.
> 
> The `prepare` job (Node 24.x) resolves the merge base via the GitHub Compare API, runs `get-changed-workspaces.mts`, and exports `package-names`, `changed-paths`, and `merge-base`. Downstream jobs use those outputs: changelog validation and per-Node test matrices iterate only affected workspace names; `wallet-cli` e2e runs only when that package is in the set.
> 
> **ESLint** moves to a dedicated `lint-eslint` job that runs `yarn lint:eslint` on all packages when `changed-paths` is `full`, otherwise only on the listed workspace paths.
> 
> **Build** uses `generate-partial-build-tsconfig.mts` plus `ts-bridge` for a partial TypeScript reference graph (changed packages plus transitive dependants and dependencies) when a merge base exists and there is no root-level change; otherwise it keeps `yarn build`. Root or workflow changes outside ignored files (e.g. not `README.md`) still force a full run; pushes without a merge base fall back to all packages.
> 
> New shared logic lives in `scripts/lib/workspaces.mts` (git diff, dependency graph, root-change detection); `tsconfig.scripts.json` now includes `*.mts` for those scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b910c315f2667f29cca128d10fd37f2615d8b6c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -27,6 +27,62 @@ jobs:
           echo "child-workspace-package-names=$(yarn workspaces list --no-private --json | jq --slurp --raw-output 'map(.name) | @json')" >> "$GITHUB_OUTPUT"
         shell: bash
 
+
+  get-changed-packages:
+    name: Get changed packages
+    runs-on: ubuntu-latest
+    needs: prepare
+    outputs:
+      merge-base: ${{ steps.fetch-merge-base.outputs.merge-base }}
+      package-names: ${{ steps.packages.outputs.package-names }}
+      eslint-paths: ${{ steps.packages.outputs.eslint-paths }}
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: false
+          persist-credentials: false
+          node-version: 24.x
+      - name: Fetch merge base
+        id: fetch-merge-base
+        if: github.base_ref != '' || github.event.merge_group.base_ref != ''
+        run: |
+          set -euo pipefail
+
+          PREFIXED_REF_REGEX='refs/heads/(.+)'
+          if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+            BASE_REF="${BASH_REMATCH[1]}"
+          fi
+
+          MERGE_BASE=$(gh api "repos/$GITHUB_REPOSITORY/compare/$BASE_REF...$HEAD_SHA" --jq '.merge_base_commit.sha')
+          git fetch --unshallow --filter=blob:none --no-tags origin HEAD
+
+          echo "merge-base=$MERGE_BASE" >> "$GITHUB_OUTPUT"
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.ref || github.event.merge_group.base_ref }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha }}
+          GH_TOKEN: ${{ github.token }}
+      - name: Get changed package names
+        id: packages
+        run: |
+          if [[ -n "$MERGE_BASE" ]]; then
+            OUTPUT=$(yarn tsx scripts/get-changed-workspaces.mts "$MERGE_BASE" "$HEAD_SHA")
+            PACKAGES=$(echo "$OUTPUT" | jq '.names')
+            if [[ $(echo "$OUTPUT" | jq '.hasRootChange') == "true" ]]; then
+              ESLINT_PATHS="full"
+            else
+              ESLINT_PATHS=$(echo "$OUTPUT" | jq '.locations')
+            fi
+          else
+            PACKAGES=$(yarn workspaces list --no-private --json | jq --slurp --raw-output 'map(.name) | @json')
+            ESLINT_PATHS="full"
+          fi
+          echo "package-names=$PACKAGES" >> "$GITHUB_OUTPUT"
+          echo "eslint-paths=$ESLINT_PATHS" >> "$GITHUB_OUTPUT"
+        env:
+          MERGE_BASE: ${{ steps.fetch-merge-base.outputs.merge-base }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha }}
+
   lint:
     name: Lint (${{ matrix.script }})
     runs-on: ubuntu-latest
@@ -38,7 +94,6 @@ jobs:
           - codeowners:check
           - constraints
           - lint:dependencies
-          - lint:eslint
           - lint:misc:check
           - lint:teams
           - lint:tsconfigs:all
@@ -55,6 +110,40 @@ jobs:
         run: yarn "$SCRIPT"
         env:
           SCRIPT: ${{ matrix.script }}
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty at end of job"
+            exit 1
+          fi
+
+  lint-eslint:
+    name: Lint (lint:eslint)
+    runs-on: ubuntu-latest
+    if: needs.get-changed-packages.outputs.eslint-paths != '[]'
+    needs:
+      - prepare
+      - get-changed-packages
+    strategy:
+      matrix:
+        node-version: [24.x]
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: false
+          persist-credentials: false
+          node-version: ${{ matrix.node-version }}
+      - name: Lint
+        run: |
+          if [[ "$ESLINT_PATHS" == "full" ]]; then
+            yarn lint:eslint
+          else
+            yarn lint:eslint $(echo "$ESLINT_PATHS" | jq -r '.[]' | tr '\n' ' ')
+          fi
+        env:
+          ESLINT_PATHS: ${{ needs.get-changed-packages.outputs.eslint-paths }}
       - name: Require clean working directory
         shell: bash
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-build-test.yml
```diff
@@ -252,6 +367,7 @@ jobs:
   test-wallet-cli-e2e:
     name: Test wallet-cli daemon e2e (${{ matrix.node-version }})
     runs-on: ubuntu-latest
+    if: needs.prepare.outputs.package-names == '[]' || contains(fromJson(needs.prepare.outputs.package-names), '@metamask/wallet-cli')
```
Gudahtt: I'm not following the intent behind this first condition - why would we want this to run if no packages were changed?

Mrtenz: Good catch, it's a leftover from an earlier iteration of this change, and not needed anymore. Removed in 14ce87c35c0dea6bb6d4aed3a31abb9db935b5fe.

scripts/lib/workspaces.mts
```diff
@@ -0,0 +1,212 @@
...
+ * @param includeDependencies - Whether to also expand to transitive dependencies.
+ * @param changedFiles - Pre-fetched list of changed files; fetched if omitted.
+ * @returns The set of workspace names to check.
+ */
+export async function computeChangedWorkspaces(
+  workspaces: Workspace[],
+  mergeBase: string,
+  headRef: string,
+  includeDependencies: boolean,
+  changedFiles?: string[],
```
Gudahtt: Nit: It's a little odd for this parameter to exist, because the required `mergeBase` and `headRef` parameters aren't actually used when this is passed in.

It seems clearer to choose one other the other: either require changed packages to be passed in, or require `mergeBase` and `headRef`.

Given how this is used in `get-changed-workspaces`, probably requiring `changedPackages` would be better

scripts/lib/workspaces.mts
```diff
@@ -0,0 +1,212 @@
...
+export const ROOT_WORKSPACE = new URL('../..', import.meta.url).pathname;
+
+// Files that can change without requiring a full rebuild/test run.
+const IGNORED_ROOT_FILES = new Set([
+  '.gitignore',
+  'AGENTS.md',
+  'CLAUDE.md',
+  'README.md',
+  'teams.json',
+  'yarn.lock',
```
Gudahtt: Not sure about this one. A lockfile change could definitely break something

scripts/lib/workspaces.mts
Gudahtt: Another nit: I prefer to use option objects for cases like this so that each parameter can be identified at the callsite. I had to scroll back and forth to ensure they were passed in the correct order, and to identify what the `true` 4th param represented.

I'm tempted to propose this as a lint rule, but it feels a bit too opinionated 🤔 Not sure.

scripts/get-changed-workspaces.mts
```diff
@@ -0,0 +1,64 @@
...
+    type: 'boolean',
+    default: false,
+    describe:
+      'Also expand to transitive dependencies (needed for TypeScript builds)',
+  })
+  .demandCommand(1, 'merge-base is required')
+  .help()
+  .parseAsync();
+
+const mergeBase = argv._[0] as string;
```
Gudahtt: Nit: I found that we can get rid of these type assertions by getting the named parameters from `argv`, e.g. `argv.mergeBase`, `argv.headSha`, etc.).

I had to switch `demandCommand` to `demandOption` though; apparently `demandCommand` only works for sub-commands not positional parameters (which their docs seem to contradict, but 🤷 I tested it).

Gudahtt: Here's the full snippet I tested:

```
const argv = await yargs(hideBin(process.argv))
  .usage('$0 <merge-base> [head-sha] [options]')
  .positional('merge-base', {
    type: 'string',
    describe: 'Merge base SHA',
  })
  .positional('head-sha', {
    type: 'string',
    describe: 'PR branch tip SHA (defaults to HEAD)',
    default: 'HEAD',
  })
  .option('include-dependencies', {
    type: 'boolean',
    default: false,
    describe:
      'Also expand to transitive dependencies (needed for TypeScript builds)',
  })
  .demandOption('merge-base')
  .help()
  .parseAsync();

const { mergeBase, headSha: headRef, includeDependencies } = argv;
const workspaces = await getAllWorkspaces();

const changedFiles = await getChangedFiles(mergeBase, headRef);
const hasRootChange = checkRootChange(workspaces, changedFiles);

const changed = await computeChangedWorkspaces(
  workspaces,
  mergeBase,
  headRef,
  includeDependencies,
  changedFiles,
);
```

Gudahtt (dismissed):
LGTM!

Mostly had nits, though I was tempted to block on the yarn.lock ignore issue. I am on the fence about that.

scripts/lib/workspaces.mts
Mrtenz: I'm still on the fence about this too.

If we do consider changes to `yarn.lock` as needing to run everything, I feel like we limit the number of PRs where these changes are useful a lot.

I was thinking we could try to parse the changes to `yarn.lock` to make it more granular, but that's probably not worth the effort.

I'm open to removing this from `IGNORED_ROOT_FILES` for now.

scripts/get-changed-workspaces.mts
Mrtenz: Ah, nice. That's definitely a lot better.

Done in c5ad03cceed6ef7528d0ac0b706287a3675ce691.

scripts/lib/workspaces.mts
Gudahtt: Maybe as a compromise we could ignore `yarn.lock` changes that are just workspace package bumps (e.g. only run the workspaces that were bumped, rather than all). But if _any_ external dependency changes, re-run everything.

That would be far more achievable than trying to calculate exactly which packages were impacted by a given lockfile change.

We already have some code for parsing lockfile diffs here: https://github.com/MetaMask/metamask-extension/blob/446c5a25b56d6521ad963e7503ce337ad975a8fb/.github/scripts/yarn-audit-diff.mts#L233
(though I didn't review it carefully)

Gudahtt: For now I'm leaning towards removing it from the ignore list. We can always add it later.

Mrtenz: Sounds good. Will remove it for now, and create a ticket for us to consider this in the future.

scripts/get-changed-workspaces.mts
Gudahtt: Whoops, I guess I didn't test it properly. I see the CI failure.

Mrtenz: Hmm, I tried the `demandOption` parameter in the `positional` options too, but then it's not added to the `argv` object as parameter, only to `argv._`. I'll just make them regular parameters.

scripts/lib/workspaces.mts
Mrtenz: Good point. Removed `mergeBase` and `headRef` from this function. I would change it to an option object, but it probably makes less sense now that it only has three parameters?

f7f904f2059933c512047d2b0a2de05410089259

scripts/lib/workspaces.mts
Mrtenz: Done: b910c315f2667f29cca128d10fd37f2615d8b6c5.

scripts/get-changed-workspaces.mts
Mrtenz: Fixed: 5a90382e26b66226cf66aa5cf45ba50af198bb96.

Gudahtt: I think I found the problem. Our `.usage` call has no description. If I pass a description, then `.positional` works as expected, and `demandOption` gives it the proper type.

Why the usage description matters here, I have no idea

Gudahtt (approved):
LGTM!

scripts/get-changed-workspaces.mts
Gudahtt: Oh, making it an option works too

Mrtenz: Interesting. 😅 I guess it doesn't really matter if we use an option compared to a positional, so will just keep it as-is.

PR #9388, opened 2026-07-04
fix(assets-controller): seed native balances after account refresh

salimtb:
Zero-balancing for new accounts ran before getAssets on account-tree changes, so assetsBalance was often never populated for the new account. Seed native zero balances using the refreshed account list after the fetch.

- Fetch balances via RpcDataSource when a user switches EVM accounts, enables RPC-only networks, switches account groups, or adds a new account to the account tree
- Fixes missing `assetsBalance` for newly added accounts on custom/RPC networks (e.g. DXC / `eip155:50`) where `assetsInfo` was present but balances were not fetched

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes balance refresh timing in `AssetsController` for account-tree and subscription flows; localized but affects wallet balance display on custom/RPC networks.
> 
> **Overview**
> Fixes cases where **`assetsBalance` stayed empty** for new or RPC-only-network accounts even when metadata was present, by tightening when `AssetsController` force-refreshes balances.
> 
> On **account tree updates**, the controller now tracks **newly added accounts** and, after the usual group-wide `getAssets` and re-subscribe, runs a **second forced fetch scoped to those new accounts** so the slow/RPC pipeline can populate balances (e.g. custom chains like `eip155:50`).
> 
> Documents the same class of fix for **account group switches** and **newly enabled RPC-only networks** in the changelog. `#handleEnabledNetworksChanged` only refactors to reuse a local `accounts` variable when fetching added chains—behavior unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e8793363de057d48dff138524aced4fe1d3f6c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -1370,42 +1370,6 @@ export class AssetsController extends BaseController<
   }
 
   /**
-   * Fetch balances via RpcDataSource and merge into state.
-   *
-   * @param accounts - Accounts to fetch for.
-   * @param chainIds - Chains to fetch on.
-   */
-  async #fetchAccountBalancesViaRpc(
-    accounts: InternalAccount[],
-    chainIds: ChainId[],
-  ): Promise<void> {
-    if (accounts.length === 0 || chainIds.length === 0) {
-      return;
-    }
-
-    const rpcChains = chainIds.filter((chainId) =>
-      this.#rpcDataSource.getActiveChainsSync().includes(chainId),
-    );
-    if (rpcChains.length === 0) {
-      return;
-    }
-
-    const request = this.#buildDataRequest(accounts, rpcChains, {
-      dataTypes: ['balance'],
-      forceUpdate: true,
-    });
-    const { response } = await this.#executeMiddlewares(
-      [
-        createParallelBalanceMiddleware([this.#rpcDataSource]),
-        this.#detectionMiddleware,
-      ],
-      request,
-    );
-
-    await this.#updateState({ ...response, updateMode: 'merge' });
-  }
-
-  /**
    * Force-fetch balances then subscribe. Used on unlock / first startup.
    *
    * @param accounts - Selected accounts to refresh.
@@ -3369,10 +3333,6 @@ export class AssetsController extends BaseController<
       // Subscribe after fetch so WS notifications can recover state
       this.#subscribeAssets();
 
-      await this.#fetchAccountBalancesViaRpc(accounts, [
-        ...this.#enabledChains,
-      ]);
-
       this.#ensureNativeBalancesDefaultZero();
       this.#ensureDefaultTrackedAssetsSeeded();
     } finally {
@@ -3424,8 +3384,6 @@ export class AssetsController extends BaseController<
         forceUpdate: true,
         updateMode: 'merge',
       });
-
-      await this.#fetchAccountBalancesViaRpc(accounts, addedChains);
     }
 
     this.#ensureNativeBalancesDefaultZero();
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -2,6 +2,7 @@ import type {
   AccountTreeControllerGetAccountsFromSelectedAccountGroupAction,
   AccountTreeControllerSelectedAccountGroupChangeEvent,
   AccountTreeControllerState,
+  AccountTreeControllerStateChangeEvent,
 } from '@metamask/account-tree-controller';
 import type {
   AccountsControllerGetSelectedAccountAction,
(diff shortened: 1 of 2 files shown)


packages/assets-controller/src/AssetsController.ts
```diff
@@ -321,26 +328,16 @@ type AllowedActions =
...
-  NetworkEnablementControllerState
->;
-
 type AllowedEvents =
   // AssetsController
   | AccountTreeControllerSelectedAccountGroupChangeEvent
-  | AccountTreeControllerStateChangedEvent
-  | ClientControllerStateChangedEvent
+  | AccountTreeControllerStateChangeEvent
+  | ControllerStateChangedEvent<
```
Prithpal-Sooriya: Can we use the imported version?
Less type issues then

Prithpal-Sooriya: Same for the other state change events

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1079,6 +1080,14 @@ export class AssetsController extends BaseController<
       },
     );
 
+    // Switching EVM address within the same account group (e.g. Account 1 → Account 2).
+    this.messenger.subscribe(
+      'AccountsController:selectedEvmAccountChange',
```
Prithpal-Sooriya: How come this and not the account tree change?

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1329,25 +1338,37 @@ export class AssetsController extends BaseController<
...
+  ): Promise<void> {
     const releaseLock = await this.#accountRefreshMutex.acquire();
     try {
       await this.getAssets(accounts, {
         chainIds: [...this.#enabledChains],
         forceUpdate: true,
       });
       this.#subscribeAssets();
+      if (newAccounts.length > 0) {
+        await this.#fetchAccountBalancesViaRpc(newAccounts, [
```
Prithpal-Sooriya: Thoughts on calling the above getAssets with newAccounts?

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1356,6 +1377,42 @@ export class AssetsController extends BaseController<
...
   }
 
+  /**
+   * Fetch balances via RpcDataSource and merge into state.
+   *
+   * @param accounts - Accounts to fetch for.
+   * @param chainIds - Chains to fetch on.
+   */
+  async #fetchAccountBalancesViaRpc(
+    accounts: InternalAccount[],
```
Prithpal-Sooriya: See prev comment, then we can potentially remove this?

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3320,13 +3377,51 @@ export class AssetsController extends BaseController<
       // Subscribe after fetch so WS notifications can recover state
       this.#subscribeAssets();
 
+      await this.#fetchAccountBalancesViaRpc(accounts, [
```
Prithpal-Sooriya: Need to call around this. I'm not a fan of even more complexity

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3320,13 +3377,51 @@ export class AssetsController extends BaseController<
...
+    const releaseLock = await this.#accountRefreshMutex.acquire();
+    try {
+      await this.getAssets([account], {
+        chainIds: [...this.#enabledChains],
+        forceUpdate: true,
+      });
+
+      this.#subscribeAssets();
+
+      await this.#fetchAccountBalancesViaRpc(
```
Prithpal-Sooriya: Not sure I understand why this is needee

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3364,12 +3459,15 @@ export class AssetsController extends BaseController<
...
-      await this.getAssets(this.#getSelectedAccounts(), {
+    const accounts = this.#getSelectedAccounts();
+    if (addedChains.length > 0 && accounts.length > 0) {
+      await this.getAssets(accounts, {
         chainIds: addedChains,
         forceUpdate: true,
         updateMode: 'merge',
       });
+
+      await this.#fetchAccountBalancesViaRpc(accounts, addedChains);
```
Prithpal-Sooriya: See prev comment

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3487,6 +3585,8 @@ export class AssetsController extends BaseController<
         dataTypes: ['balance', 'metadata', 'price'],
       });
 
+      await this.#fetchAccountBalancesViaRpc(accounts, [selectedChainId]);
```
Prithpal-Sooriya: See prev comment

packages/assets-controller/src/AssetsController.ts
salimtb: accountTree change is not fired , that was the issue , now i need to relay on this event

packages/assets-controller/src/AssetsController.ts
salimtb: hmmmm because getAssets will fetch also quick pipeline and restart all subscription , here i wanted to scope it only on rpc scope to make it better for perf

Prithpal-Sooriya (approved):

PR #9415, opened 2026-07-07
fix(geolocation-controller): point to API Platform geolocation-api instead of legacy Ramps endpoint

amitabh94:
## Explanation

`GeolocationApiService` (used by Ramp, Perps, Rewards, Card, mUSD, and Money Account across extension and mobile) still fetches from the legacy Ramps-owned `on-ramp.{env}api.cx.metamask.io/geolocation` endpoint.

That endpoint is slated for deprecation now that geolocation is a platform-owned concern — see internal Slack context. API Platform's replacement, `geolocation.{env}api.cx.metamask.io/v1/geolocation`, has been live since March and returns the same plain-text ISO 3166-2 response format, so this is a same-contract URL swap with no consumer-facing changes.

## Open item before merge

API Platform has **not yet provisioned a dedicated UAT deployment** for `geolocation-api` (confirmed: `geolocation.uat-api.cx.metamask.io` does not resolve, and the UAT workload config in `va-mmcx-geolocation-api-workload` is empty). This PR temporarily maps `Env.UAT` to the production URL so UAT testing isn't broken by this change. That mapping should be revisited once a real UAT deployment exists.

Opening as **draft** pending confirmation from API Platform (`@api-platform-devs`) that:
- the PROD/DEV URLs are correct and stable
- the UAT-points-at-PROD interim approach is acceptable on their end

## References

- `packages/geolocation-controller` is owned by `@metamask-mobile-platform` per CODEOWNERS
- API Platform docs: https://docs.cx.metamask.io/api/geolocation/

---

> [!NOTE]
> **Medium Risk**
> Geolocation is shared across multiple products (Ramp, Perps, Rewards, etc.); a wrong host or UAT→prod mapping could affect regional behavior, though the response contract is intended to stay the same.
> 
> **Overview**
> **`GeolocationApiService` now calls API Platform’s `geolocation-api` instead of the legacy Ramps `on-ramp` geolocation host**, ahead of that endpoint’s deprecation. The path changes from `/geolocation` to `/v1/geolocation`, and hosts move from `on-ramp.{env}api.cx.metamask.io` to `geolocation.{env}api.cx.metamask.io` (DEV still uses the `dev-` prefix). Response handling and caching are unchanged.
> 
> **UAT no longer uses a separate UAT URL** because API Platform has not deployed `geolocation-api` there yet; `Env.UAT` temporarily uses the **production** geolocation URL. Tests and the changelog document that interim mapping.
> 
[...]

PR #9417, opened 2026-07-07
fix(geolocation-controller): point to API Platform geolocation-api instead of legacy Ramps endpoint

amitabh94:
## Explanation

`GeolocationApiService` (used by Ramp, Perps, Rewards, Card, mUSD, and Money Account across extension and mobile) still fetches from the legacy Ramps-owned `on-ramp.{env}api.cx.metamask.io/geolocation` endpoint.

That endpoint is slated for deprecation now that geolocation is a platform-owned service. API Platform's replacement, `geolocation.{env}api.cx.metamask.io/v1/geolocation`, has been live since March and returns the same plain-text ISO 3166-2 response format, so this is a same-contract URL swap with no consumer-facing changes.

## Open item before merge

API Platform has **not yet provisioned a dedicated UAT deployment** for `geolocation-api` (confirmed: `geolocation.uat-api.cx.metamask.io` does not resolve, and the UAT workload config in `va-mmcx-geolocation-api-workload` is empty). This PR temporarily maps `Env.UAT` to the production URL so UAT testing isn't broken by this change. That mapping should be revisited once a real UAT deployment exists.

Opening as **draft** pending confirmation from API Platform (`@api-platform-devs`) that:
- the PROD/DEV URLs are correct and stable
- the UAT-points-at-PROD interim approach is acceptable on their end

## References

- `packages/geolocation-controller` is owned by `@metamask-mobile-platform` per CODEOWNERS
- API Platform docs: https://docs.cx.metamask.io/api/geolocation/

Supersedes #9415 (opened from a fork by mistake, which broke the changelog-diff CI check).

---

> [!NOTE]
> **Medium Risk**
> Wide downstream use (Ramp, Perps, Rewards, etc.) makes any geolocation outage impactful, though the API contract is unchanged; UAT hitting production is an interim operational caveat.
> 
> **Overview**
> **`GeolocationApiService`** now calls API Platform’s **`geolocation-api`** host and **`/v1/geolocation`** path instead of the legacy Ramps **`on-ramp.*`** **`/geolocation`** URLs that are being deprecated. The response contract is unchanged (plain-text ISO 3166-2).
> 
> **`getGeolocationUrl`** builds **`https://geolocation.api.cx.metamask.io`** for production and UAT, and **`https://geolocation.dev-api.cx.metamask.io`** for DEV. **UAT** no longer uses a separate **`uat-`** host and temporarily shares the production URL until API Platform provisions a UAT deployment. Tests and the changelog reflect the new URLs and UAT behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 411259b25db2c234a296e15faf46ee41269758cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/geolocation-controller/src/geolocation-api-service/geolocation-api-service.ts
+++ b/packages/geolocation-controller/src/geolocation-api-service/geolocation-api-service.ts
@@ -11,7 +11,7 @@ import type { GeolocationApiServiceMethodActions } from './geolocation-api-servi
 
 const DEFAULT_TTL_MS = 5 * 60 * 1000;
 
-const ENDPOINT_PATH = '/geolocation';
+const ENDPOINT_PATH = '/v1/geolocation';
 
 // === GENERAL ===
 
@@ -68,12 +68,17 @@ export type GeolocationApiServiceMessenger = Messenger<
 /**
  * Returns the base URL for the geolocation API for the given environment.
  *
+ * Served by API Platform's `geolocation-api` service, not the legacy
+ * Ramps-owned `on-ramp` endpoint this previously pointed to. API Platform has
+ * not yet provisioned a dedicated UAT deployment for this service, so UAT
+ * temporarily resolves to the production URL until one exists.
+ *
  * @param env - The environment to get the URL for.
  * @returns The full URL for the geolocation endpoint.
  */
 function getGeolocationUrl(env: Env): string {
-  const envPrefix = env === Env.PRD ? '' : `${env}-`;
-  return `https://on-ramp.${envPrefix}api.cx.metamask.io${ENDPOINT_PATH}`;
+  const envPrefix = env === Env.DEV ? 'dev-' : '';
+  return `https://geolocation.${envPrefix}api.cx.metamask.io${ENDPOINT_PATH}`;
 }
 
 /**
--- a/packages/geolocation-controller/CHANGELOG.md
+++ b/packages/geolocation-controller/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Point `GeolocationApiService` at API Platform's `geolocation-api` service instead of the legacy Ramps-owned `on-ramp` geolocation endpoint, which is slated for deprecation ([#TBD](https://github.com/MetaMask/core/pull/TBD))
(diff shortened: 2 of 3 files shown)


packages/geolocation-controller/src/geolocation-api-service/geolocation-api-service.ts
```diff
@@ -68,12 +68,17 @@ export type GeolocationApiServiceMessenger = Messenger<
...
+ * not yet provisioned a dedicated UAT deployment for this service, so UAT
+ * temporarily resolves to the production URL until one exists.
+ *
  * @param env - The environment to get the URL for.
  * @returns The full URL for the geolocation endpoint.
  */
 function getGeolocationUrl(env: Env): string {
-  const envPrefix = env === Env.PRD ? '' : `${env}-`;
-  return `https://on-ramp.${envPrefix}api.cx.metamask.io${ENDPOINT_PATH}`;
+  const envPrefix = env === Env.DEV ? 'dev-' : '';
```
mcmire, line 80: Why not map UAT to dev? This way we are free to make changes to the API if need be without needing to deploy to prod.

amitabh94: We can. I was waiting for someone from the @MetaMask/api-platform to comment if they are going to provision UAT endpoints. 
Usually UAT envs point to prod in case of fallbacks to get the closest feel of user testing.

Why do you think we cannot make changes without deploying to prod? We have the dev env to test right?

mcmire: Yeah, to be clear, by "we" I meant the API Platform team. My thought was that if a bug was encountered in the API during UAT then it would make sense to test that fix in a dev environment, but then maybe the API Platform team wouldn't want to deploy the change to prod until it was tested again in the context of UAT? But then I see your point — UAT should be as close to prod as possible. 

I guess it depends on the API Platform team wants to do here. I didn't see that there is essentially a blocker on this PR until that point is resolved, so that's fine.

Cal-L (approved):
LGTM

packages/geolocation-controller/src/geolocation-api-service/geolocation-api-service.ts
amitabh94: Thank you for taking the time to review this @mcmire !

PR #9418, opened 2026-07-07
Copy createServicePolicy to base-data-service

mcmire:
## Explanation

Currently `createServicePolicy` lives in `controller-utils`, as originally we wrote it as an abstract utility that could theoretically be used for any kind of data access. However, there are a few reasons why this is suboptimal now:

1. `controller-utils` is too large and serves as a dependent of almost every package in the monorepo, so extreme care must be taken when changing this package to avoid breaking changes.
2. `createServicePolicy` was added before we formalized data services. Now we have a `BaseDataService` class, and this class sets up `createServicePolicy` by default, so it's highly unlikely that `createServicePolicy` will be used independently going forward.
3. Also, we only use `createServicePolicy` for HTTP requests, so making it abstract is kind of unnecessary.

This commit deprecates the existing exports in `controller-utils`, copies them to `base-data-service`, and changes `BaseDataService` to use the copied version instead.

In the future we will remove the symbols from `controller-utils` and modify `createServicePolicy` further so it works better for HTTP requests.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1130

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches shared HTTP retry/circuit-breaker behavior for data services; behavior is intended to be a straight copy, but any drift would affect outage handling. Deprecations are non-breaking for now.
> 
> **Overview**
> Moves **`createServicePolicy`** (retry/circuit-breaker wiring around Cockatiel) from `@metamask/controller-utils` into `@metamask/base-data-service`, so data-layer resilience lives next to `BaseDataService` instead of a widely depended-on utils package.
> 
> **`@metamask/base-data-service`** adds `createServicePolicy.ts`, exports the same API (including Cockatiel re-exports), drops the **`@metamask/controller-utils`** dependency in favor of direct **`cockatiel`**, and points **`BaseDataService`** at the local module. Tests for the policy are added under base-data-service; example/tests import `BrokenCircuitError` / `ConstantBackoff` from `cockatiel` where appropriate.
> 
> **`@metamask/controller-utils`** keeps the old implementations but marks **`createServicePolicy`** and related types, constants, and Cockatiel re-exports as **deprecated**, with changelogs directing callers to **`@metamask/base-data-service`**. The README dependency graph no longer shows `base-data-service` → `controller-utils` for this path.
> 
> No in-repo consumer migrations beyond `BaseDataService` in this PR; other packages can keep importing from controller-utils until a follow-up removes the deprecated surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2e3704584971984aa9cb00c2404d9e7464def76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/base-data-service/package.json
+++ b/packages/base-data-service/package.json
@@ -54,10 +54,10 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
   "dependencies": {
-    "@metamask/controller-utils": "^12.3.0",
     "@metamask/messenger": "^2.0.0",
     "@metamask/utils": "^11.11.0",
     "@tanstack/query-core": "^4.43.0",
+    "cockatiel": "^3.1.2",
     "fast-deep-equal": "^3.1.3"
   },
   "devDependencies": {
--- a/packages/base-data-service/src/createServicePolicy.test.ts
+++ b/packages/base-data-service/src/createServicePolicy.test.ts
@@ -1,0 +1,964 @@
+import { CircuitState, ConstantBackoff, handleWhen } from 'cockatiel';
+
+import {
+  createServicePolicy,
+  DEFAULT_CIRCUIT_BREAK_DURATION,
+  DEFAULT_DEGRADED_THRESHOLD,
+  DEFAULT_MAX_CONSECUTIVE_FAILURES,
+  DEFAULT_MAX_RETRIES,
+  ServicePolicy,
+} from './createServicePolicy';
+
+describe('createServicePolicy', () => {
+  beforeEach(() => {
+    jest.useFakeTimers({ doNotFake: ['nextTick', 'queueMicrotask'] });
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  describe('execute', () => {
+    describe('when the service succeeds at least on the first attempt', () => {
+      it('returns what the service returns', async () => {
+        const policy = createServicePolicy();
+        const result = await policy.execute(() => ({ some: 'data' }));
+        expect(result).toStrictEqual({ some: 'data' });
+      });
+
+      it('fires onAvailable on the first successful execution and not again on subsequent successful executions', async () => {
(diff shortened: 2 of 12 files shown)


packages/base-data-service/src/createServicePolicy.test.ts
mcmire, line 1: This should be equivalent to `packages/controller-utils/src/create-service-policy.test.ts`. (I decided to use camelCase for the file name because I have found recently that camelCased file names are more searchable than snake-cased filenames.)

packages/base-data-service/src/createServicePolicy.ts
mcmire, line 1: This should be essentially equivalent to `packages/controller-utils/src/create-service-policy.ts`, except that the `cockatiel` re-exports have been moved to `index.ts`. (I decided to use camelCase for the file name because I have found recently that camelCased file names are more searchable than snake-cased filenames.)

packages/base-data-service/package.json
```diff
@@ -54,10 +54,10 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
   "dependencies": {
-    "@metamask/controller-utils": "^12.3.0",
```
mcmire, line 57: We don't need `controller-utils` anymore because we were only using it for `createServicePolicy`.

packages/base-data-service/src/index.ts
```diff
@@ -1,3 +1,18 @@
+export type {
```
mcmire, line 1: In `controller-utils`, these re-exports were located in `create-service-policy.ts`. However, in moving them to this package, I found that Istanbul mistakenly tried to check that these exports are exercised in tests. Moving them to `index.ts` sidesteps this issue because `index.ts` files are automatically excluded from coverage checks.

cryptodev-2s (approved):

PR #9425, opened 2026-07-08
Assets 3365 perf optimise multicall client to calculate address specific data once

salimtb:
## Checklist

---

> [!NOTE]
> **Low Risk**
> Pure encoding memoization with unchanged multicall targets and calldata semantics; covered by existing and new unit tests.
> 
> **Overview**
> **`MulticallClient`** now **caches** ABI-encoded `balanceOf` and `getEthBalance` calldata **per account address** (case-normalized), because that payload only depends on the account, not the token contract target.
> 
> Batch **`aggregate3`** construction moves into **`buildAggregate3BalanceCalls`**, which pulls cached encodings instead of re-encoding for every token row. The **single-call ERC-20 fallback** path uses the same cache.
> 
> A test asserts **`encodeFunctionData`** runs once per distinct account for each method type in a batch, and that counts stay flat on a **second** batch with the same accounts (cache reuse). Changelog updated under **Unreleased**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe4349ab3ae7bca1ffac66e285852cbf764e127a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
+++ b/packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
@@ -402,6 +402,72 @@ function encodeGetEthBalance(accountAddress: Address): Hex {
   ]);
 }
 
+type BalanceCallDataCache = {
+  getErc20BalanceCallData: (accountAddress: Address) => Hex;
+  getNativeBalanceCallData: (accountAddress: Address) => Hex;
+};
+
+/**
+ * Cache balance call encodings per account address for a single batch request.
+ * ERC-20 `balanceOf` and native `getEthBalance` call data depend only on the
+ * account being queried, not the token contract (which is the multicall target).
+ *
+ * @returns A cache scoped to one `batchBalanceOf` invocation.
+ */
+function createBalanceCallDataCache(): BalanceCallDataCache {
+  const erc20ByAccount = new Map<string, Hex>();
+  const nativeByAccount = new Map<string, Hex>();
+
+  return {
+    getErc20BalanceCallData(accountAddress: Address): Hex {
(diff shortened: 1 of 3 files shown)


packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -527,10 +593,11 @@ export class MulticallClient {
 
     const multicallAddress = MULTICALL3_ADDRESS_BY_CHAIN[chainId];
     const provider = this.#getProvider(chainId);
+    const callDataCache = createBalanceCallDataCache();
```
Prithpal-Sooriya: Hmm new cache created every time you call batchBalanceOf.

Thoughts on making the cache live outside and is always reused?

```
// in root of module
const callDataCache = createBalanceCallDataCache();

// now we don't need prop drilling, places can use directly.
function buildAggregate3BalanceCalls(
  batch: BalanceOfRequest[],
  multicallAddress: Hex,
) {
  // internals stay the same, it now uses the cache from outside (not injected in).
  // pro - no prop drilling, con - not a "pure function", but this tradeoff is fine...
}

```

salimtb: done here [1c80599](https://github.com/MetaMask/core/pull/9425/commits/1c80599eb1e2b6166af20eb0118485c9df2e2d67)

Prithpal-Sooriya (dismissed):

Prithpal-Sooriya (approved):

PR #9430, opened 2026-07-08
feat(assets-controller): gate non-EVM account-activity websocket behind a namespace feature flag

salimtb:
Add a generic `webSocketEnabledNamespaces` constructor option so non-EVM account-activity balances (starting with Solana, extensible to e.g. Tron) are served over WebSocket only when their CAIP namespace is enabled. EVM (`eip155`) is always served over WebSocket.

- BackendWebsocketDataSource claims an enabled namespace's chains (from a namespace -> chains map, since the supported-networks API is EVM-only) while the WebSocket is connected, and subscribes their account-activity channels. Disabled namespaces are left unclaimed and unsubscribed so the SnapDataSource serves them.
- When a chain (or a whole non-EVM namespace) is reported down via `system-notifications`, it is excluded from the claimable set and automatically falls back to the SnapDataSource until it recovers.
- AccountsApiDataSource is documented as EVM-only (removing the stale "restore solana" TODO); non-EVM is served by WebSocket or Snap, keeping Snap as the sole non-EVM fallback.

## Checklist

### DIFF
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


salimtb:
@metamaskbot publish-preview

PR #9432, opened 2026-07-08
feat!: allow dynamic infura networks

mikesposito:
## Explanation

Currently, `NetworkController` manages two types of network config: `Infura` and `Custom`. The `Infura` type is dedicated to configurations with RPC endpoints with an Infura URL that uses a template string for the project ID, while `Custom` configurations are generic network configurations that can be also added by users.

In order to add a network configuration of Infura type to NetworkController, the network must first be added to `@metamask/controller-utils` to satisfy `NetworkController` validations, or else it won’t be recognized as a known network, and the operation will fail with an error. This currently blocks us from leveraging mm-config-registry (Dynamic registry) from being used to add Infura network configurations remotely.

This PR relaxes `NetworkController` validations to accept Infura configurations that are not included in `@metamask/controller-utils`, nor in any other bundled known list.

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WPN-1554

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking changes to core RPC/network resolution and chain ID scaffolding affect every consumer of getNetworkClientById; misconfigured dynamic Infura networks could report wrong chain IDs or route IDs to unexpected client types.
> 
> **Overview**
> **NetworkController** relaxes Infura-specific constraints so Infura-backed networks can be added or updated without appearing in the bundled `@metamask/controller-utils` network list (e.g. for remote/dynamic registry config).
> 
> **Breaking behavior:** `BuiltInNetworkClientId` and `InfuraNetworkClientConfiguration.network` are plain `string` instead of `InfuraNetworkType`. **`getNetworkClientById`** resolves clients by checking the Infura registry first, then custom—so an ID like `mainnet` can bind to an Infura client even when a custom endpoint reused that ID. Missing clients now throw a single message: `No network client was found with ID "…"`.
> 
> **Removed validations:** custom endpoints may use Infura-style `networkClientId` values; Infura RPC URLs are no longer required to imply a chain ID that matches the network configuration’s `chainId`. Infura client middleware now scaffolds **`eth_chainId`** from the configured **`chainId`** instead of `ChainId[knownNetworkName]`.
> 
> Downstream test expectations in **assets-controllers** and **network-controller** are updated for the unified error text; coverage thresholds in **jest** are raised after deleting obsolete validation tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4be7c05db1f09f45ccbebf8ea104085b581af70. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mikesposito:
@metamaskbot publish-preview

packages/network-controller/src/types.ts
```diff
@@ -47,7 +46,7 @@ export type CustomNetworkClientConfiguration =
  */
 export type InfuraNetworkClientConfiguration =
   CommonNetworkClientConfiguration & {
-    network: InfuraNetworkType;
+    network: string;
```
mcmire, line 49: The only drawback in changing this is that we will no longer be using a type to represent the list of possible Infura networks, and therefore we will no longer be using a type to ensure that the URL for an Infura network is valid (e.g., `network` can now be set to anything, so `https://asdfasdfsdf.infura.io/v3/${configuration.infuraProjectId}` would be a valid URL, at least from a NetworkController / RPC middleware perspective). I think that is an acceptable tradeoff for the benefit we are getting, but it does mean that `eth-json-rpc-infura` can probably be simplified, and we no longer need this type, since we don't use it anymore: https://github.com/MetaMask/eth-json-rpc-infura/blob/81e1c2295d5e7b1c6d5e6af66fa715bd6c941a0f/src/types.ts#L19. Just noting this in case it comes up again :)

mikesposito:
@metamaskbot publish-preview

mikesposito:
@metamaskbot publish-preview

packages/network-controller/CHANGELOG.md
```diff
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - If you have code that narrows on `BuiltInNetworkClientId`, you will need to remove the narrowing or check the network client type via `NetworkClientType` instead.
+- **BREAKING:** `InfuraNetworkClientConfiguration.network` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))
+  - Previously only known Infura network names were accepted. Any valid Infura subdomain string is now accepted.
+- **BREAKING:** `NetworkController.getNetworkClientById` no longer distinguishes between Infura and custom network clients when the provided ID is not found ([#9432](https://github.com/MetaMask/core/pull/9432))
+  - The method now checks the Infura registry first and falls back to the custom registry, without using `isInfuraNetworkType` to decide which registry to consult.
+  - The error message when no client is found has changed from `"No Infura network client was found with the ID \"<id>\"."` / `"No custom network client was found with the ID \"<id>\"."` to `"No network client was found with ID \"<id>\"."`.
```
mikesposito:
```suggestion
```

packages/network-controller/src/NetworkController.ts
mcmire: `getNetworkClientById` already looks up the network client in the registry. So now we would be doing the lookup twice. 

I wonder if we can replace all of this with:

``` typescript
    this.#autoManagedNetworkClient = this.getNetworkClientById(networkClientId);

File: packages/network-controller/CHANGELOG.md [hunk 3]
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** `BuiltInNetworkClientId` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))
+  - This type was previously constrained to known Infura network names (e.g. `"mainnet"`, `"sepolia"`). It is now `string` to allow dynamically configured Infura networks whose names are not bundled into the package.
+  - If you have code that narrows on `BuiltInNetworkClientId`, you will need to remove the narrowing or check the network client type via `NetworkClientType` instead.
+- **BREAKING:** `InfuraNetworkClientConfiguration.network` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))
```
mcmire: Do we think that this particular change is breaking? It seems like this would be backward-compatible.

``` suggestion
- `InfuraNetworkClientConfiguration.network` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))

File: packages/network-controller/CHANGELOG.md [hunk 4]
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** `BuiltInNetworkClientId` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))
+  - This type was previously constrained to known Infura network names (e.g. `"mainnet"`, `"sepolia"`). It is now `string` to allow dynamically configured Infura networks whose names are not bundled into the package.
+  - If you have code that narrows on `BuiltInNetworkClientId`, you will need to remove the narrowing or check the network client type via `NetworkClientType` instead.
```
mcmire, line 14: Maybe worth noting another breaking change for `getNetworkClientById`? Something like this (if this doesn't make sense, feel free to modify):

``` suggestion
- **BREAKING:** `BuiltInNetworkClientId` is now `string` instead of `InfuraNetworkType` ([#9432](https://github.com/MetaMask/core/pull/9432))
  - This type was previously constrained to known Infura network names (e.g. `"mainnet"`, `"sepolia"`). It is now `string` to allow dynamically configured Infura networks whose names are not bundled into the package.
  - If you have code that narrows on `BuiltInNetworkClientId`, you will need to remove the narrowing or check the network client type via `NetworkClientType` instead.
- **BREAKING:** `getNetworkClientById` no longer uses the given network client ID  to determine RPC endpoint type, prioritizing Infura RPC endpoints over custom RPC endpoints ([#9432](https://github.com/MetaMask/core/pull/9432))
  - If you have an RPC endpoint with a `type` of `custom` but with a `networkClientId` that previously matched a known Infura network name (e.g. `mainnet`), this will now be treated as an Infura network rather than a custom network.
  - This method is not only public but is also used internally to resolve network client IDs, so this is a change in behavior across the whole controller.

mcmire (commented):
The approach overall looks good. I just left a few comments below.

packages/network-controller/src/NetworkController.ts
mikesposito: Nice! done in [275c142](https://github.com/MetaMask/core/pull/9432/commits/275c142211a71911477c7b16e60a2e90e31b291b)

mikesposito:
@metamaskbot publish-preview

mcmire (approved):
LGTM!

Prithpal-Sooriya (approved):

PR #9445, opened 2026-07-09
feat(base-data-service): Support persisting the cache

FrederikBolding:
## Explanation

Add support for cache persistence, which can be enabled by configuring `persistenceConfig`. Persistence is achieved using `StorageService` and debounced to not cause too many writes. Rehydration of the cache happens on `init`, while expired or invalid data is discarded. Any errors with persistence are logged, but ultimately ignored, due to the reduced importance of the data persisted.

This approach of storing the dehydrated state is similar to how `@tanstack/query-persist-client-core` works.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1128

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches shared data-layer caching and storage I/O; stale or rehydrated cache could affect offline/API behavior, though persistence is opt-in and errors are non-fatal.
> 
> **Overview**
> Adds **optional TanStack Query cache persistence** to `BaseDataService` via a new constructor `persistenceConfig` (`maxAge`, optional `writeDelay` / `maxWriteDelay`). When enabled, successful query cache entries are **dehydrated** and written through messenger actions to `@metamask/storage-service` under key `cache`, with **lodash debounced** writes after cache add/update/remove and **`removeItem`** when the cache is empty.
> 
> Introduces **`init()`** to **rehydrate** from storage on startup; entries older than `maxAge` are dropped and removed. Persistence is **off unless `persistenceConfig` is set**; storage failures are swallowed (logged via `captureException`). **`destroy()`** now cancels pending debounced writes.
> 
> The package gains dependencies on `@metamask/storage-service` and `lodash`, exports `PersistenceConfiguration`, and includes broad persistence tests. README dependency graph and changelog are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79adfe5d3bba8aebdea131fb6aa5726f04130685. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/base-data-service/src/BaseDataService.ts
```diff
@@ -138,8 +178,27 @@ export class BaseDataService<
...
+    this.#debouncedPersist =
+      this.#persistConfig &&
+      debounce(
+        () => {
+          this.#persistCache().catch(
+            /* istanbul ignore next */
+            (error) => this.#messenger.captureException?.(error),
+          );
+        },
+        this.#persistConfig.debounce ?? inMilliseconds(10, Duration.Second),
```
FrederikBolding: Open to suggestions for defaults here

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -265,11 +326,22 @@ export class BaseDataService<
...
   }
 
+  /**
+   * Initialize the service, rehydrating the cache with persisted data if possible.
+   */
+  init(): void {
+    this.#rehydrateCache().catch(
+      /* istanbul ignore next */
+      (error) => this.#messenger.captureException?.(error),
+    );
```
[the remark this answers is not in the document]
FrederikBolding: I don't think we want to wait for this anywhere, so this was intended 🤔

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -77,6 +87,19 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
   },
 };
 
+const STORAGE_SERVICE_KEY = 'cache';
+
+type PersistConfiguration = {
```
Mrtenz: Maybe export this, so we can use it in the example data service test as well?

Some docs for what the options do could be useful too.

```suggestion
export type PersistenceConfiguration = {
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -307,4 +379,59 @@ export class BaseDataService<
...
+   * Persist the query client cache using the StorageService, if the cache is not empty.
+   *
+   * @returns Nothing.
+   */
+  async #persistCache(): Promise<void> {
+    const state = dehydrate(this.#queryClient);
+
+    if (state.queries.length === 0 && state.mutations.length === 0) {
+      return;
+    }
```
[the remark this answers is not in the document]
FrederikBolding: In theory we could empty out the cache if this happens, or always allow writes, but I don't know if it is worth it. This was added as an optimization to reduce the number of writes 🤔

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Done!

packages/base-data-service/tests/ExampleDataService.ts
```diff
@@ -60,7 +60,12 @@ export class ExampleDataService extends BaseDataService<
 
   readonly #tokensBaseUrl = 'https://tokens.api.cx.metamask.io';
 
-  constructor(messenger: ExampleMessenger) {
+  constructor(
+    messenger: ExampleMessenger,
+    { persistConfig }: { persistConfig?: { maxAge: number } } = {
+      persistConfig: { maxAge: inMilliseconds(1, Duration.Day) },
+    },
```
Mrtenz: Can we use `PersistenceConfig` here?

FrederikBolding: Sure: https://github.com/MetaMask/core/pull/9445/commits/477b945bb1cb3f09ef238ae53085450e606ed629

Mrtenz: Is it intentional that it's not optional anymore?

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -77,6 +87,33 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
   },
 };
 
+export const STORAGE_SERVICE_KEY = 'cache';
+
+/**
+ * Options for persistence configuration.
+ */
+export type PersistenceConfiguration = {
```
Mrtenz, line 97: Should these be exported from `index.ts`?

packages/base-data-service/tests/ExampleDataService.ts
FrederikBolding: Nope, typoed

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: At least added `PersistenceConfiguration`, I think the key is an implementation detail 🤔

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -307,4 +395,69 @@ export class BaseDataService<
...
+    const cache: PersistedCache = {
+      timestamp: Date.now(),
+      state,
+    };
+
+    await this.#externalMessenger.call(
+      'StorageService:setItem',
+      this.name,
+      STORAGE_SERVICE_KEY,
+      cache as unknown as Json,
```
Mrtenz, line 429: `cache` could contain non-JSON values. Does `StorageService:setItem` simply strip these?

FrederikBolding: For mobile it does `JSON.stringify`, extension does `browser.storage.local.set`, which I would have to check how it works.

Though it is somewhat of an assumption in the `BaseDataService` that query keys and state are JSON serializable, since otherwise the UI syncing doesn't work either (even if the TanStack types are less narrow).

Mrtenz: > Though it is somewhat of an assumption in the `BaseDataService` that query keys and state are JSON serializable, since otherwise the UI syncing doesn't work either (even if the TanStack types are less narrow).

I mean the `DehydratedState` can contain `DehydratedQuery`s which can have a `promise` (`Promise<unknown>`) option. Don't know if this is the case in practice, but we should verify that this doesn't break anything.

FrederikBolding: Where are you seeing `Promise<unknown>`? 🤔

```ts
export interface QueryState<TData = unknown, TError = unknown> {
  data: TData | undefined
  dataUpdateCount: number
  dataUpdatedAt: number
  error: TError | null
  errorUpdateCount: number
  errorUpdatedAt: number
  fetchFailureCount: number
  fetchFailureReason: TError | null
  fetchMeta: any
  isInvalidated: boolean
  status: QueryStatus
  fetchStatus: FetchStatus
}
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -105,26 +149,38 @@ export class BaseDataService<
...
+  readonly #debouncedPersist?: DebouncedFunc<() => void>;
+
+  readonly #persistConfig?: PersistenceConfiguration;
+
   constructor({
     name,
     messenger,
     queryClientConfig = {},
     policyOptions,
+    persistConfig,
```
mcmire: What do you think about matching the name of the type and calling this `persistenceConfig`? 

```suggestion
    persistenceConfig,
```

(I guess this option is sort of a boolean in the sense that if you specify it, it you are enabling persistence. But I don't know if the name is the right place to communicate that fact.)

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -77,6 +89,33 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
...
+  debounce?: number;
+  /**
+   * The maximum number of milliseconds to wait between persistence writes.
+   */
+  debounceMaxWait?: number;
+};
+
+type PersistedCache = {
+  state: DehydratedState;
+  timestamp: number;
```
mcmire, line 116: Nit: I'm curious if the name could be ambiguous. Do you think it's worth it to be explicit?

``` suggestion
  persistedTimestamp: number;
}

File: packages/base-data-service/src/BaseDataService.ts [hunk 10]
```diff
@@ -77,6 +89,33 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
...
+  /**
+   * The maximum age before the cache is treated as expired in milliseconds.
+   * This is relevant for rehydrating the state during initialization,
+   * if the cached state is too old it will be discarded.
+   */
+  maxAge: number;
+  /**
+   * The number of milliseconds to wait before triggering persistence following a cache update.
+   */
+  debounce?: number;
```
mcmire: This name sounds like this could be a boolean. Maybe:

``` suggestion
  debounceInterval?: number;

Comment (mcmire) [reply packages/base-data-service/src/BaseDataService.ts]: Or maybe consumers shouldn't need to think about the fact that we are using debounce under the hood?

``` suggestion
  writeDelay?: number;

File: packages/base-data-service/src/BaseDataService.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (mcmire) [reply packages/base-data-service/src/BaseDataService.ts:353]: Why don't we want to wait?

File: packages/base-data-service/src/BaseDataService.ts [hunk 11]
```diff
@@ -77,6 +89,33 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
...
+  /**
+   * The maximum number of milliseconds to wait between persistence writes.
+   */
+  debounceMaxWait?: number;
```
mcmire: Similar as above, would this name make more sense?

``` suggestion
  maxWriteDelay?: number;

File: packages/base-data-service/src/BaseDataService.ts [hunk 12]
```diff
@@ -105,26 +149,38 @@ export class BaseDataService<
...
+    persistConfig?: PersistenceConfiguration;
   }) {
     this.name = name;
 
-    // We are storing a separately typed messenger for known actions and events provided by data services
-    // and a generic public one that is typed using the generic parameters and accessible to implementations.
+    // We store two narrowly-typed messengers alongside the generic public one:
+    // - #messenger handles the service's own action registration and event publishing
+    // - #externalMessenger handles calls to external actions
+    // Splitting them avoids TypeScript issues with mixing template-literals with regular strings
```
mcmire, line 174: Interesting. We've never had to do this before. Would be curious to understand the issues here, but I can look at this later.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -307,4 +395,72 @@ export class BaseDataService<
...
+      'StorageService:getItem',
+      this.name,
+      STORAGE_SERVICE_KEY,
+    );
+
+    if (!persisted) {
+      return;
+    }
+
+    const cache = persisted as unknown as PersistedCache;
```
mcmire: Nit: Naming

``` suggestion
    const { result: untypedCache } = await this.#externalMessenger.call(
      'StorageService:getItem',
      this.name,
      STORAGE_SERVICE_KEY,
    );

    if (!untypedCache) {
      return;
    }

    const cache = untypedCache as unknown as PersistedCache;

File: packages/base-data-service/src/BaseDataService.ts [hunk 14]
```diff
@@ -307,4 +395,72 @@ export class BaseDataService<
...
+    );
+  }
+
+  /**
+   * Rehydrate the query client cache using the StorageService, if persistence is configured
+   * and the persisted cache is not expired.
+   *
+   * @returns Nothing.
+   */
+  async #rehydrateCache(): Promise<void> {
```
mcmire: It seems like we are mixing terminology a bit. Is there a name we can choose that's the opposite of `persistCache`?

``` suggestion
  async #loadCache(): Promise<void> {

File: packages/base-data-service/CHANGELOG.md [hunk 15]
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for cache persistence ([#9445](https://github.com/MetaMask/core/pull/9445))
+  - Persistence can be configured by passing a `persistConfig` to the constructor.
```
mcmire: Nit:

``` suggestion
  - Persistence can be configured by passing a `persistConfig` option to the constructor.

File: packages/base-data-service/src/BaseDataService.ts [hunk 16]
```diff
@@ -77,6 +89,33 @@ const QUERY_CLIENT_DEFAULTS: DefaultOptions = {
   },
 };
 
+export const STORAGE_SERVICE_KEY = 'cache';
```
mcmire, line 92: Are we using one storage service key for all data services? Should we namespace this key (i.e. move this to a private property and call this `${this.name}:cache`)?

FrederikBolding: The `StorageService` is already namespaced, that's why the first argument in all the calls is `this.name`

packages/base-data-service/src/BaseDataService.ts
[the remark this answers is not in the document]
FrederikBolding: I was thinking this would be similar to controllers providing a fire and forget `init` function to do setup.

I don't think we would want to block anywhere in the clients to wait for loading these persisted caches 🤔

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Yeah, the problem seems to be the pattern of passing in template-literals like the `ServiceName`, e.g.
```ts
export type DataServiceInvalidateQueriesAction<ServiceName extends string> = {
  type: `${ServiceName}:invalidateQueries`;
  handler: (
    filters?: InvalidateQueryFilters<Json>,
    options?: InvalidateOptions,
  ) => Promise<void>;
};
```

packages/base-data-service/src/BaseDataService.ts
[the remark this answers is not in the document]
mcmire: Yeah I guess the question is, if cache persistence is not fully initialized will this prevent the data service from functioning correctly or would it cause adverse effects? It sounds like it would not, but just want to make sure.

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: It seems a bit excessive to me to be explicit here, but if you feel strongly, I am happy to make the change.

packages/base-data-service/src/BaseDataService.ts
mcmire: Ah. Right :)

packages/base-data-service/src/BaseDataService.ts
mcmire: Seems fine. Could consider shortening to 5 seconds if this is too slow in practice. But even 10 seconds doesn't seem bad.

packages/base-data-service/src/BaseDataService.ts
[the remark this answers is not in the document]
FrederikBolding: It shouldn't, it just means that we may have to refetch queries. Similarly to if fetching the cache fails for whatever reason.

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Since they are non-critical writes, my thinking was starting higher and tweaking later, but 🤷‍♂️

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -265,11 +343,22 @@ export class BaseDataService<
...
+      (error) => this.#messenger.captureException?.(error),
+    );
+  }
+
   /**
    * Prepares the service for garbage collection. This should be extended
    * by any subclasses to clean up any additional connections or events.
    */
   destroy(): void {
+    this.#debouncedPersist?.cancel();
```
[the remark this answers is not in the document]
FrederikBolding: This is not a bug, this is the whole feature.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -307,4 +396,72 @@ export class BaseDataService<
...
+    if (Date.now() - cache.timestamp >= this.#persistenceConfig.maxAge) {
+      await this.#externalMessenger.call(
+        'StorageService:removeItem',
+        this.name,
+        STORAGE_SERVICE_KEY,
+      );
+      return;
+    }
+
+    hydrate(this.#queryClient, cache.state);
```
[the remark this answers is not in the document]
FrederikBolding: TanStack does not overwrite fresher queries.

packages/base-data-service/src/BaseDataService.ts
mcmire: I'm on the fence. Let's stick with `timestamp`.

packages/base-data-service/src/BaseDataService.ts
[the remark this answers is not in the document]
mcmire: Okay, that's what I thought. Yeah I guess that's not so bad.

mcmire (approved):
LGTM!

PR #9460, opened 2026-07-10
refactor(multichain-account-service): remove v1 `createAccount` entirely

ccharly:
## Explanation

All Snaps properly implements `createAccounts` (even if they are not truly v2-compatible yet).

So we can safely remove the old `createAccount` flow for those.

NOTE: The current `SnapKeyring` (v2) knows how to properly call `createAccounts` v1 or v2, see:
- https://github.com/MetaMask/accounts/blob/7ecc4248f8a80a06ca77f43e1f7345da2e478bbb/packages/keyring-snap-bridge/src/v2/SnapKeyring.ts#L326-L334

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core Snap account creation behavior across chains; relies on all BIP-44 Snaps supporting `createAccounts`, though v1 discovery remains a separate path.
> 
> **Overview**
> Snap account providers no longer branch on capabilities to call singular **`createAccount`** (v1) versus batched **`createAccounts`**. **All** derive/discover/range creation now goes through **`SnapKeyringProxy.createAccounts`** under the **`Provider Create Accounts (v2 - batched)`** trace; the v1 create trace and **`RestrictedSnapKeyring` / `createAccountV1`** hooks are removed from **`SnapAccountProvider`** and chain providers (BTC, Solana, Tron, Stellar).
> 
> **`withSnap`** now builds a mutex-wrapped **`SnapKeyringProxy`** instead of exposing optional **`v1.createAccount`**. v1 **discovery** still uses the client’s **`discoverAccounts`**, but persisting discovered accounts uses the same **`createAccounts`** path as explicit creation. Tests and changelog document the removal and the prior v1 side effects (e.g. unwanted auto-selection).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae00c4c29c0947daa3f850b4556067c15cb28cc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/multichain-account-service/src/analytics/traces.ts
+++ b/packages/multichain-account-service/src/analytics/traces.ts
@@ -75,8 +75,7 @@ export function toCreateAccountsV2DataTraces(
 export enum TraceName {
   SnapDiscoverAccounts = 'Snap Discover Accounts',
   EvmDiscoverAccounts = 'EVM Discover Accounts',
-  ProviderCreateAccountV1 = 'Provider Create Account (v1)',
-  ProviderCreateAccounts = 'Provider Create Accounts (v2 - batched)',
+  ProviderCreateAccounts = 'Provider Create Accounts',
   WalletAlignment = 'Wallet Alignment',
   WalletCreateMultichainAccountGroup = 'Wallet Create Multichain Account Group',
   WalletCreateMultichainAccountGroups = 'Wallet Create Multichain Account Groups',
--- a/packages/multichain-account-service/src/analytics/traces.ts
+++ b/packages/multichain-account-service/src/analytics/traces.ts
@@ -75,7 +75,7 @@ export function toCreateAccountsV2DataTraces(
 export enum TraceName {
   SnapDiscoverAccounts = 'Snap Discover Accounts',
   EvmDiscoverAccounts = 'EVM Discover Accounts',
-  ProviderCreateAccounts = 'Provider Create Accounts',
+  ProviderCreateAccounts = 'Provider Create Accounts (v2 - batched)',
   WalletAlignment = 'Wallet Alignment',
   WalletCreateMultichainAccountGroup = 'Wallet Create Multichain Account Group',
   WalletCreateMultichainAccountGroups = 'Wallet Create Multichain Account Groups',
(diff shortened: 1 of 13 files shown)


ccharly:
@metamaskbot publish-preview

packages/multichain-account-service/src/analytics/traces.ts
```diff
@@ -75,7 +75,6 @@ export function toCreateAccountsV2DataTraces(
 export enum TraceName {
   SnapDiscoverAccounts = 'Snap Discover Accounts',
   EvmDiscoverAccounts = 'EVM Discover Accounts',
-  ProviderCreateAccountV1 = 'Provider Create Account (v1)',
   ProviderCreateAccounts = 'Provider Create Accounts (v2 - batched)',
```
ccharly, line 78: Decided to keep this name, as changing trace name could have an impact our metrics. For now, I'll keep it that way to minimize the number of changes.

ccharly:
@metamaskbot publish-preview

packages/multichain-account-service/src/providers/SnapAccountProvider.ts
```diff
@@ -328,7 +277,9 @@ export abstract class SnapAccountProvider extends BaseBip44AccountProvider {
                 // We still need to remove the accounts from the Snap keyring since we're
                 // about to create the same account again, which will use a new ID, but will
                 // keep using the same address, and the Snap keyring does not allow this.
-                await keyring.deleteAccount(account.id);
+                await this.#withSnapKeyring(async ({ keyring }) => {
```
hmalik88: why did we have to make this change? i.e. why can't we still use the keyring from the `withSnap` call?

ccharly: We could, but I feel like the `RestrictedSnapKeyring` was mostly here because of v1.

So I preferred to get rid of it entirely in favoir of the traditional "withKeyring" -> operation.

We could keep it and wrap every call with `withSnapKeyring` though if you think that's a better/simpler pattern?

ccharly: Re-introduced it with a different name: https://github.com/MetaMask/core/pull/9460/commits/5af7740f94df85cd2baa1aeaece119e2906a2c2c

ccharly:
@metamaskbot publish-preview

hmalik88 (approved):

PR #9471, opened 2026-07-10
fix(perps-controller): repair CommonJS build and add close leverage + order latency

abretonc7s:
## Explanation

perps-controller@9.2.1 (#9311) has a broken CommonJS build: dist/services/HyperLiquidClientService.cjs and dist/utils/standaloneInfoClient.cjs emit require("file:///home/runner/work/hyperliquid/hyperliquid/src/mod.ts") — an absolute build-machine path — so any CJS/bundler/Jest consumer throws Cannot find module. Root cause is upstream: @nktkas/hyperliquid (0.33.0+) ships .d.ts with a /// <amd-module name="file:///home/runner/.../mod.ts" /> directive (a Deno->npm dnt build artifact); ts-bridge uses that amd-module name as the CJS require() target, leaking the absolute path. The ESM build is unaffected.

This PR:
1. Fixes the CJS build with a yarn patch on @nktkas/hyperliquid that strips the amd-module directives, applied via the root monorepo resolutions (build-time only). perps-controller's own dependency range stays ^0.33.1 — the patch is deliberately NOT in the package's dependencies, since a published patch: specifier would break every downstream consumer. After rebuild both .cjs files emit require("@nktkas/hyperliquid") with zero file: leaks. Bumping the SDK does not fix this (0.33.1 is latest and still ships the directive; 0.32.2 is the last clean version but predates breaking API changes this package depends on).
2. — PERPS_POSITION_CLOSE_TRANSACTION.leverage is now the computed effective leverage abs(positionValue)/marginUsed rounded to 1 decimal (guarded for finite values and marginUsed > 0), populated for every close including TP/SL.
3. — adds optional TrackingData.orderExecutionLatencyMs, emitted as order_execution_latency_ms on the trade terminal event only (scoped to #trackOrderResult, not the shared attribution builder), with a !== undefined guard.

All changes are additive and recorded under `## [Unreleased]` in the CHANGELOG (version stays 9.2.1 in this PR per release-cut convention; the next perps-controller release picks them up).

## References

- Related to #9311 (original perps analytics contract)
- Upstream bug report: https://github.com/nktkas/hyperliquid/issues/165
- Consumer PRs: MetaMask/metamask-extension#44324, MetaMask/metamask-mobile#33095 (extension carries a temporary yarn patch on 9.2.1; removed once the next perps-controller release built with this resolutions patch is published)

## Checklist

- [ ] I've introduced breaking changes in this PR and have prepared draft pull requests for clients and consumer packages to resolve them — N/A, additive/non-breaking

---

> [!NOTE]
> **Medium Risk**
> Touches live trading analytics classification and provider order results alongside a monorepo-wide dependency patch; changes are additive and well-tested but affect how partial fills and margin failures are reported.
> 
> **Overview**
> Repairs the **broken CommonJS build** for `@metamask/perps-controller` consumers by adding a **Yarn patch** on `@nktkas/hyperliquid@0.33.1` (wired through root `resolutions`) that strips `/// <amd-module name="file:///...">` lines from shipped `.d.ts` files, so `ts-bridge` no longer emits `require("file:///home/runner/...")` in `.cjs` artifacts.
> 
> **Perps analytics and trading telemetry** are extended in `TradingService` and `HyperLiquidProvider`: close events report **effective leverage** (`abs(positionValue) / marginUsed`, 1 dp) instead of configured leverage; successful opens can emit an extra **`partially_filled`** `PERPS_TRADE_TRANSACTION` when `filledSize` is below the exchange **submitted** size (new `OrderResult.submittedSize` from the provider, including post-retry normalization); batch closes add **`number_positions_closed`**; **`updateMargin`** now emits a failed **Risk Management** event when the provider returns `{ success: false }` without throwing. **`TradeAction`** types now include flip variants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d71e795d0de96fe6e8d5a822ceb0b22787ea8f97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/perps-controller/src/services/TradingService.ts
+++ b/packages/perps-controller/src/services/TradingService.ts
@@ -2225,6 +2225,19 @@ export class TradingService {
         // Invalidate standalone caches so external hooks refresh
         this.#deps.cacheInvalidator.invalidate({ cacheType: 'positions' });
         this.#deps.cacheInvalidator.invalidate({ cacheType: 'accountState' });
+      } else {
+        // Track failure analytics for a non-throwing provider failure so the
+        // terminal Risk Management event is emitted exactly once here (the
+        // thrown path below handles exceptions).
+        this.#deps.metrics.trackPerpsEvent(PerpsAnalyticsEvent.RiskManagement, {
+          [PERPS_EVENT_PROPERTY.STATUS]: PERPS_EVENT_VALUE.STATUS.FAILED,
+          [PERPS_EVENT_PROPERTY.ASSET]: symbol,
+          [PERPS_EVENT_PROPERTY.ACTION]:
+            parseFloat(amount) > 0 ? 'add_margin' : 'remove_margin',
+          [PERPS_EVENT_PROPERTY.MARGIN_USED]: Math.abs(parseFloat(amount)),
+          [PERPS_EVENT_PROPERTY.COMPLETION_DURATION]: completionDuration,
+          [PERPS_EVENT_PROPERTY.ERROR_MESSAGE]: result.error ?? 'Unknown error',
+        });
       }
 
       this.#deps.tracer.endTrace({
--- a/packages/perps-controller/src/services/TradingService.ts
+++ b/packages/perps-controller/src/services/TradingService.ts
@@ -1,3 +1,4 @@
+import { BigNumber } from 'bignumber.js';
 import { v4 as uuidv4 } from 'uuid';
 
 import {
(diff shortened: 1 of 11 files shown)


packages/perps-controller/src/services/TradingService.ts
```diff
@@ -299,12 +299,39 @@ export class TradingService {
...
     // Propagate discovery attribution + hl_fee_rate (TAT-3080, TAT-3149)
     Object.assign(
       properties,
       this.#buildAttributionProperties(params.trackingData),
     );
 
+    // Emit an additional partially filled trade event when the fill is partial,
+    // mirroring the close path so the fill's partiality is visible in analytics
+    // rather than hidden behind a status=executed event.
+    const requestedSize = parseFloat(params.size);
```
geositta: Could we classify partial fills using the exact size submitted by the provider? This comparison uses params.size, while HyperLiquidProvider.placeOrder can normalize that value to asset precision, recalculate it from usdAmount, or adjust it during the minimum order retry. A complete fill of the normalized size can therefore emit partially_filled with a small incorrect remaining_amount. Returning the submitted size in OrderResult, or having the provider return the partial-fill classification, would keep the analytics status aligned with the executed order.

packages/perps-controller/src/services/TradingService.ts
```diff
@@ -299,12 +299,39 @@ export class TradingService {
       properties[PERPS_EVENT_PROPERTY.AB_TESTS] = params.trackingData.abTests;
     }
 
+    // Order execution latency on the terminal trade event
+    if (params.trackingData?.orderExecutionLatencyMs !== undefined) {
+      properties[PERPS_EVENT_PROPERTY.ORDER_EXECUTION_LATENCY_MS] =
```
geositta: What distinction should order_execution_latency_ms have from the existing completion_duration on this terminal event? The new field accepts a completed elapsed duration in the request data, although callers construct that request before the operation completes. If a client specific start boundary is required, accepting a start timestamp and calculating the elapsed value here would make that contract easier to use. If the controller boundary is sufficient, the existing duration already provides it.

packages/perps-controller/src/services/TradingService.ts
```diff
@@ -299,12 +299,39 @@ export class TradingService {
...
+    const filledSize =
+      result?.filledSize === undefined ? NaN : parseFloat(result.filledSize);
+    if (
+      result?.success === true &&
+      Number.isFinite(filledSize) &&
+      filledSize > 0 &&
+      filledSize < requestedSize
+    ) {
+      this.#deps.metrics.trackPerpsEvent(PerpsAnalyticsEvent.TradeTransaction, {
+        ...properties,
```
geositta, line 342: Nit: Could we confirm the intended schema for the supplemental partial-fill event? Spreading properties gives it order_size = filledSize and order_execution_latency_ms, while the close partial-fill path keeps order_size as the requested size and reports the fill separately through amount_filled. If dashboard parity is intended, we should override order_size with the requested size and decide explicitly whether latency belongs on both outcomes.

geositta (changes requested):
I am requesting one change before approval: partial-fill classification should use the exact size submitted to the provider rather than the earlier request size. Provider normalization or retry adjustment can otherwise report a complete fill as partial and emit an incorrect remainder.

I also left two nonblocking contract questions concerning how order_execution_latency_ms differs from the existing completion_duration, and whether the supplemental partial-fill event should mirror the close-event schema.

packages/perps-controller/src/services/TradingService.ts
abretonc7s: Fixed in 7a0e3f347. `HyperLiquidProvider.placeOrder` now returns the final normalized size it actually submits to the exchange as `OrderResult.submittedSize` (post precision rounding, USD recalculation, and the $10-minimum retry), and `TradingService` classifies partial fills against that value (`filledSize < submittedSize`) rather than the pre-normalization `params.size`. When the provider does not report a submitted size we no longer classify at all, so a complete fill of the normalized size is never misreported as partially filled.

packages/perps-controller/src/services/TradingService.ts
abretonc7s: Agreed — removed entirely. An input tracking param cannot carry a post-completion measurement, and `completion_duration` already measures submission through terminal result at the controller boundary. Dropped the field, its emission, tests, and the CHANGELOG entry in 7a0e3f347.

packages/perps-controller/src/services/TradingService.ts
abretonc7s: Aligned with the close-event schema in 7a0e3f347: `order_size` is now the submitted size, the fill is reported separately via `amount_filled`, and `remaining_amount = submittedSize - filledSize`. The open-trade partial event now mirrors the close path exactly.

geositta (approved):
Approved. Thanks for addressing the review feedback. I verified that the provider now returns the final normalized size submitted to the exchange, including the minimum order retry path, and that TradingService uses that value for partial fill classification.

The updated implementation also aligns the partial-fill event schema with the close path, uses decimal arithmetic for comparison and remaining_amount, and safely skips classification when either size is unavailable or invalid. Removing the client supplied latency field resolves the contract ambiguity with the existing controller measured completion_duration.

PR #9480, opened 2026-07-12
feat: add restoreEnabledNetworkMap action

salimtb:
Expose a public messenger action to restore a previously snapshotted enabledNetworkMap, so callers can add networks without switching the active network filter.

UI PR: https://github.com/MetaMask/metamask-extension/pull/44371

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes persisted user network filter state; incorrect snapshots could disable networks users expect enabled, though scope is limited to extension-style direct calls with an intentional undo API.
> 
> **Overview**
> Adds **`restoreEnabledNetworkMap`** on `NetworkEnablementController` so callers with direct controller access (e.g. extension) can revert `enabledNetworkMap` to a prior snapshot after adding a custom network.
> 
> The method walks **only keys already in the current map**; snapshot values are applied per key and anything missing in the snapshot becomes **`false`**. That supports undoing the exclusive filter flip from **`#onAddNetwork`** when a network is added without making it the active selection—tests cover snapshot restore, empty snapshot disabling everything, and the `NetworkController:networkAdded` flow.
> 
> **Not** registered as a messenger action (`MESSENGER_EXPOSED_METHODS` unchanged); changelog documents that explicitly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 005dc8706df6e62ec7fce5e77b670bf19d72fd33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/network-enablement-controller/src/NetworkEnablementController-method-action-types.ts
+++ b/packages/network-enablement-controller/src/NetworkEnablementController-method-action-types.ts
@@ -128,10 +128,6 @@ export type NetworkEnablementControllerDisableNetworkAction = {
 /**
  * Restores the enabled network map to a previously snapshotted state.
  *
- * This is used when adding a network without switching the active network
- * filter. Callers should snapshot `enabledNetworkMap` before adding a
- * network, then restore it after the controller reacts to `networkAdded`.
- *
  * @param enabledNetworkMap - Previously snapshotted enabledNetworkMap.
  */
 export type NetworkEnablementControllerRestoreEnabledNetworkMapAction = {
--- a/packages/network-enablement-controller/src/NetworkEnablementController-method-action-types.ts
+++ b/packages/network-enablement-controller/src/NetworkEnablementController-method-action-types.ts
@@ -126,16 +126,6 @@ export type NetworkEnablementControllerDisableNetworkAction = {
 };
 
 /**
- * Restores the enabled network map to a previously snapshotted state.
- *
- * @param enabledNetworkMap - Previously snapshotted enabledNetworkMap.
- */
-export type NetworkEnablementControllerRestoreEnabledNetworkMapAction = {
-  type: `NetworkEnablementController:restoreEnabledNetworkMap`;
-  handler: NetworkEnablementController['restoreEnabledNetworkMap'];
-};
-
-/**
  * Checks if a network is enabled.
  *
  * @param chainId - The chain ID of the network to check. Can be either:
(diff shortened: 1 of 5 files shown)


packages/network-enablement-controller/src/NetworkEnablementController-method-action-types.ts
```diff
@@ -125,6 +125,20 @@ export type NetworkEnablementControllerDisableNetworkAction = {
...
+/**
+ * Restores the enabled network map to a previously snapshotted state.
+ *
+ * This is used when adding a network without switching the active network
+ * filter. Callers should snapshot `enabledNetworkMap` before adding a
+ * network, then restore it after the controller reacts to `networkAdded`.
+ *
+ * @param enabledNetworkMap - Previously snapshotted enabledNetworkMap.
+ */
+export type NetworkEnablementControllerRestoreEnabledNetworkMapAction = {
```
Prithpal-Sooriya: I need to digest this a little. It seems similar to an update/merge.

A little worried about exposing this to clients. The consumer needs to be diligent when using this method to not create a bad state. 

Maybe instead of a controller method, we should be exposed as a util?

GuillaumeRx: This can't really be extracted outside of the controller as this needs to call `this.update()` which is a protected method, inaccessible outside of the controller

GuillaumeRx: But indeed this is a band-aid over a eager default behavior of the controller, it by default will disable other networks and enable the new one when in some cases we don't want to do it. 

The best way of fixing this would be to introduce a breaking change and remove this behavior from this controller but for now we need this fixed ASAP to unblock other work on the core-platform side :)

salimtb: @Prithpal-Sooriya  I’m just moving the logic we already had in the extension into the controller since `this.update` is a protected method.

If we move this into a utility function, it won’t be able to call it through the messenger. Since it’s updating controller state, I think it makes more sense for this to be implemented as a controller action rather than a utility.

utility will be called through an instance of this controller , do we want that to avoid breaking change ? is this the reason ?

salimtb: done [fdc478f](https://github.com/MetaMask/core/pull/9480/commits/fdc478fd8788a6eb4f103abbea12cec8d483d0f2)

Prithpal-Sooriya: Ah thx both!

Yeah happy to keep this as a bandaid or quick fix for now. We can iterate :)

Prithpal-Sooriya (approved):

PR #8526, opened 2026-04-20
feat: adding more validations for types sign v4 messages

jpuri:
## Explanation

Adding more validations for typed sign v4 messaged.

## Checklist

- [X] I've updated the test suite for new or updated code as appropriate
- [X] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [X] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [X] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> Tightens validation on `eth_signTypedData_v4` inputs, which may break dapps that previously sent additional top-level fields in typed data. Change is localized but affects a commonly used signing path.
> 
> **Overview**
> **Stricter `eth_signTypedData_v4` input validation:** the wallet middleware now rejects EIP-712 typed-data JSON that includes any *extraneous top-level keys* beyond those defined by `@metamask/eth-sig-util`’s `TYPED_MESSAGE_SCHEMA`.
> 
> This introduces a new `validateTypedMessageKeys` utility (and tests) and wires it into the `eth_signTypedData_v4` flow, plus adds coverage ensuring requests with unexpected keys fail with `Invalid input.`; the changelog is updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c2ae43efc033a93c32c2ea0baa57856f8be285e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9482, opened 2026-07-13
feat: validate `eth_sendTransaction` / `eth_signTransaction` params

OGPoyraz:
## Explanation

Malicious dapps can send `eth_sendTransaction` with a valid-looking payload plus a deeply-nested junk field (e.g. `test.b.b.b...` ~1200 levels). Downstream normalization in `@metamask/transaction-controller` recurses into these params and throws `RangeError: Maximum call stack size exceeded`. When this crash happens inside the extension/mobile PPOM middleware it short-circuits the security-scan pipeline: the request never reaches the Security Alerts API, and the user sees the confirmation without a Blockaid scan.

The same class of attack was fixed for `eth_signTypedData_v4` in #8526 by rejecting requests with extraneous top-level keys. This PR applies the equivalent guardrail to transaction methods.

### Changes

- New `validateTransactionParams(params)` in `src/utils/validation.ts`, exported from the package. Rejects with `rpcErrors.invalidInput()` when:
  - `params` is not a plain object, or
  - contains a top-level key outside the dapp-facing subset of `TransactionParams` (`accessList`, `authorizationList`, `chainId`, `data`, `from`, `gas`, `gasLimit`, `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, `nonce`, `to`, `type`, `value`), or
  - nests beyond depth 10 anywhere inside the params tree (legitimate params, including `accessList` / `authorizationList`, are ≤4 levels deep).
- Wired into the `eth_sendTransaction` and `eth_signTransaction` handlers in `createWalletMiddleware`, before `validateAndNormalizeKeyholder`.
- Also exported from `index.ts` so extension/mobile PPOM middleware can gate params before their own `normalizeTransactionParams` call (which is where the WASM crash actually originates in clients).

### Design notes

- **Reject vs. strip:** matches #8526 — reject the whole request with `invalidInput`. Stripping silently would hide the attack pattern from telemetry.
- **Allowlist scope:** intentionally omits internal-only fields on `TransactionParams` (`estimateGasError`, `estimatedBaseFee`, `estimateSuggested`, `estimateUsed`, `gasUsed`) — dapps should not be able to inject those.
- **Depth of 10:** ~2-3× headroom over legitimate use. Change if any real dapp turns out to nest deeper.

## References

- Mirrors #8526 (EIP-712 equivalent fix in this same package)
- Related client-side counterpart: MetaMask/metamask-extension#42080
- Internal: CONF-1662

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking RPC validation on security-critical transaction paths may reject previously accepted dapp payloads; changes directly target bypass of Blockaid/PPOM scanning.
> 
> **Overview**
> **BREAKING:** `eth_sendTransaction` and `eth_signTransaction` now run new **`validateTransactionParams`** before address checks. Requests must match a fixed transaction field schema (no extra top-level keys, typed `accessList` / `authorizationList`, quantity fields as hex string or number) and stay under **200 KB** serialized size or they fail with invalid params.
> 
> The helper is **exported** from the package so clients can reuse the same guard ahead of normalization/PPOM. This mirrors the earlier `eth_signTypedData_v4` hardening and blocks junk nested fields and padded payloads that could crash downstream normalization and skip security scans.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4e603a9c171a1b62dbbddc491ec42d634808486. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/eth-json-rpc-middleware/src/index.ts
```diff
@@ -35,4 +35,5 @@ export {
 } from './methods/wallet-get-supported-execution-permissions';
 export * from './providerAsMiddleware';
 export * from './retryOnEmpty';
+export { validateTransactionParams } from './utils/validation';
```
OGPoyraz: Extension/mobile PPOM middleware can gate params before their own `normalizeTransactionParams` call (which is where the WASM crash actually originates in clients).

OGPoyraz:
@metamaskbot publish-preview

jpuri (dismissed):

OGPoyraz:
@metamaskbot publish-preview

hilljessica8787-dot (approved):

jpuri (dismissed):

packages/eth-json-rpc-middleware/src/wallet.test.ts
```diff
@@ -150,6 +149,64 @@ describe('wallet', () => {
...
+          },
+        ],
+      };
+
+      await expect(
+        engine.handle(...createHandleParams(payload)),
+      ).rejects.toThrow('Invalid input.');
+    });
+
+    it('throws for the incident repro payload with deeply-nested junk', async () => {
```
mcmire: Similar as elsewhere:

``` suggestion
    it('throws when params contain deeply nested invalid data', async () => {

File: packages/eth-json-rpc-middleware/src/wallet.test.ts [hunk 3]
```diff
@@ -287,6 +342,64 @@ describe('wallet', () => {
...
```
mcmire: Similar as elsewhere:

``` suggestion
    it('throws when params contain deeply nested invalid data', async () => {

File: packages/eth-json-rpc-middleware/CHANGELOG.md [hunk 4]
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add strict validation for `eth_sendTransaction` and `eth_signTransaction` params — reject requests whose params do not match the transaction schema (extraneous top-level keys, ill-typed fields such as non-hex `to`/`data`, malformed `accessList` / `authorizationList` entries) or exceed `MAX_TRANSACTION_PARAMS_SIZE_BYTES` when serialized. Prevents downstream normalization / PPOM WASM from crashing on deeply-nested junk fields or padded payloads and silently bypassing security scans. ([#9482](https://github.com/MetaMask/core/pull/9482))
```
mcmire: Would this behavior be surprising for consumers, and if so should we mark it as breaking? We didn't do this when we added strict validation for `eth_signTypedData_v4` (https://github.com/MetaMask/core/pull/8526) so I guess you could make the case we don't have to do that here. But I don't know.

``` suggestion
- **BREAKING:** Add strict validation for `eth_sendTransaction` and `eth_signTransaction` params ([#9482](https://github.com/MetaMask/core/pull/9482))
  - Reject requests whose params do not match the transaction schema (extraneous top-level keys, ill-typed fields such as non-hex `to`/`data`, malformed `accessList` / `authorizationList` entries) or exceed `MAX_TRANSACTION_PARAMS_SIZE_BYTES` when serialized
  - Prevents downstream normalization / PPOM WASM from crashing on deeply-nested junk fields or padded payloads and silently bypassing security scans
```

In any case, instead of placing this summary into one entry can we at least use a nested list to break it up?

``` suggestion
- Add strict validation for `eth_sendTransaction` and `eth_signTransaction` params ([#9482](https://github.com/MetaMask/core/pull/9482))
  - Reject requests whose params do not match the transaction schema (extraneous top-level keys, ill-typed fields such as non-hex `to`/`data`, malformed `accessList` / `authorizationList` entries) or exceed `MAX_TRANSACTION_PARAMS_SIZE_BYTES` when serialized
  - Prevents downstream normalization / PPOM WASM from crashing on deeply-nested junk fields or padded payloads and silently bypassing security scans
```

packages/eth-json-rpc-middleware/src/utils/validation.test.ts
```diff
@@ -278,4 +280,205 @@ describe('Validation Utils', () => {
       });
     });
   });
+
+  describe('validateTransactionParams', () => {
+    const VALID_FROM = '0xbe93f9bacbcffc8ee6663f2647917ed7a20a57bb';
+    const VALID_TO = '0xdac17f958d2ee523a2206206994597c13d831ec7';
+
+    beforeEach(() => {
+      const actual = jest.requireActual<{
```
mcmire, line 289: It looks like we are unmocking superstruct. Why are we mocking superstruct in the first place? That seems very weird to me.

packages/eth-json-rpc-middleware/src/utils/validation.test.ts
```diff
@@ -278,4 +280,205 @@ describe('Validation Utils', () => {
...
+      expect(() =>
+        validateTransactionParams({
+          from: VALID_FROM,
+          to: VALID_TO,
+          extraKey: 'unexpected',
+        }),
+      ).toThrow('Invalid input.');
+    });
+
+    it('throws for the incident repro payload (deeply-nested junk field)', () => {
```
mcmire: What incident? We don't make any reference to an incident elsewhere in this code, so do we need to be that specific? 

``` suggestion
    it('throws given deeply-nested invalid data', () => {

File: packages/eth-json-rpc-middleware/src/utils/validation.ts [hunk 7]
```diff
@@ -234,3 +235,99 @@ export function validateTypedMessageKeys(data: string): void {
...
+ * @throws rpcErrors.invalidParams() with a Superstruct failure summary if
+ * the schema mismatch is on a typed field.
+ */
+export function validateTransactionParams(params: unknown): void {
+  if (params === null || typeof params !== 'object' || Array.isArray(params)) {
+    throw rpcErrors.invalidInput();
+  }
+
+  for (const key of Object.keys(params)) {
+    if (!ALLOWED_TRANSACTION_PARAM_KEYS.has(key)) {
```
mcmire: Would this check not already be covered by Superstruct? Since we are using `object` to define `TransactionParamsStruct` I think this already makes that object strict in terms of its properties.

packages/eth-json-rpc-middleware/src/utils/validation.test.ts
OGPoyraz: I agree, superstruct doesn't have to be mocked at first place. But removing that will reflect the size of the PR and unnecessary test changes. 

If we feel comfortable adding those changes I can do it in this PR.

packages/eth-json-rpc-middleware/CHANGELOG.md
OGPoyraz: I think it make sense to mark it breaking change as it's including the validation behaviour even though we are assuming benign RPC consumers won't get any hit.

packages/eth-json-rpc-middleware/src/utils/validation.test.ts
OGPoyraz: This is totally on me, I think my local AI tooling got confused and missing context while I made the changes. Reverted all that.

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Totally right, this was a previous approach that I implemented but then switched to superstruct, all removed.

OGPoyraz:
@metamaskbot publish-preview

packages/eth-json-rpc-middleware/src/utils/validation.test.ts
mcmire: Hmm, you're right. Okay, I guess we'll have to clean this up later. No worries about this then.

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+ * contains an extraneous top-level key, or exceeds the size limit.
+ * @throws rpcErrors.invalidParams() with a Superstruct failure summary if
+ * the schema mismatch is on a typed field.
+ */
+export function validateTransactionParams(params: unknown): void {
+  if (params === null || typeof params !== 'object' || Array.isArray(params)) {
+    throw rpcErrors.invalidInput();
+  }
+
+  validateParams(params, TransactionParamsStruct);
```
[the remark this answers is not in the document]
mcmire: `invalidParams` produces an official JSON-RPC error (code -32602), whereas `invalidInput` produces a non-standard error (code -32000). `invalidParams` seems appropriate if indeed any part of params are invalid. But maybe we should strive for consistency? I'm not sure. This is probably fine.

mcmire (commented):
This looks good to me, but I would like another review from my team just to be safe.

packages/eth-json-rpc-middleware/src/utils/validation.ts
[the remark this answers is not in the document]
OGPoyraz: I think this will be fine, we don't really need to adjust if we are returning correct error code. Because benign dApp requests will never going to hit this anyway, they will operate as before. 

The main case we were trying to fix here was; scammers were bloating the RPC params with unnecessary junk and letting PPOM & security alerts to crash, I believe they won't care about the return code 😅

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+ *
+ * @param params - The transaction params object supplied by the dapp.
+ * @throws rpcErrors.invalidInput() if params is not a plain object,
+ * contains an extraneous top-level key, or exceeds the size limit.
+ * @throws rpcErrors.invalidParams() with a Superstruct failure summary if
+ * the schema mismatch is on a typed field.
+ */
+export function validateTransactionParams(params: unknown): void {
+  if (Array.isArray(params)) {
+    throw rpcErrors.invalidInput();
```
FrederikBolding: I believe we use invalidParams elsewhere when parameters have an unexpected shape

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+  }
+
+  validateParams(params, TransactionParamsStruct);
+
+  if (JSON.stringify(params).length > MAX_TRANSACTION_PARAMS_SIZE_BYTES) {
```
FrederikBolding:
```suggestion
  if (new TextEncoder().encode(JSON.stringify(params)).byteLength > MAX_TRANSACTION_PARAMS_SIZE_BYTES) {
```
Technically this is more accurate (and also what we use in `getJsonSize` from utils). Though it is slower on mobile 😞

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+
+/**
+ * Validates `eth_sendTransaction` / `eth_signTransaction` params against the
+ * standard transaction schema and rejects payloads whose serialized size
+ * exceeds `MAX_TRANSACTION_PARAMS_SIZE_BYTES`.
+ *
+ * Checks run in this order to guarantee we never recurse into hostile
+ * subtrees:
+ *
+ * 1. Top-level shape: params must be a plain object whose top-level keys
```
FrederikBolding: Makes a lot of sense to be more cautions about validation here. Not necessarily a blocker, but I wonder if we've done any benchmarking at the cost (especially on mobile) of running x number of regexes on mobile?

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+  chainId: StrictHexStruct,
+  nonce: StrictHexStruct,
+  r: optional(StrictHexStruct),
+  s: optional(StrictHexStruct),
+  yParity: optional(StrictHexStruct),
+});
+
+export const TransactionParamsStruct = object({
+  accessList: optional(array(AccessListEntryStruct)),
+  authorizationList: optional(array(AuthorizationListEntryStruct)),
```
FrederikBolding: Should we size limit these, just to some reasonable cap that cannot be exceeded in practice?

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
     }
   }
 }
+
+const AccessListEntryStruct = object({
+  address: HexAddressStruct,
+  storageKeys: array(StrictHexStruct),
```
FrederikBolding: Same question as https://github.com/MetaMask/core/pull/9482/files#r3587566153

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+  from: HexAddressStruct,
+  gas: optional(StrictHexStruct),
+  gasLimit: optional(StrictHexStruct),
+  gasPrice: optional(StrictHexStruct),
+  maxFeePerGas: optional(StrictHexStruct),
+  maxPriorityFeePerGas: optional(StrictHexStruct),
+  nonce: optional(StrictHexStruct),
+  to: optional(HexAddressStruct),
+  type: optional(StrictHexStruct),
+  value: optional(StrictHexStruct),
```
FrederikBolding: Do we not have coercion somewhere on fields like `value`, `gasPrice`, `gasLimit` etc somewhere? I thought dapps could get away with passing in numbers as well 🤔

OGPoyraz: Good point, but with this PR we're only focused on the shape of the request and its size - we're not aiming to fully implement strict validations yet. More rules will likely follow in future work.

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Unfortunately we haven't done any benchmarking because no data on Mixpanel of average size of benign dApp requests.

One plan in our mind is to check Sentinel state logs for submitted transactions but it will take time and we want to implement this as soon as possible to stop attack vectors.

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +235,89 @@ export function validateTypedMessageKeys(data: string): void {
...
+  yParity: optional(StrictHexStruct),
+});
+
+export const TransactionParamsStruct = object({
+  accessList: optional(array(AccessListEntryStruct)),
+  authorizationList: optional(array(AuthorizationListEntryStruct)),
+  chainId: optional(StrictHexStruct),
+  data: optional(StrictHexStruct),
+  from: HexAddressStruct,
+  gas: optional(StrictHexStruct),
```
matthewwalsh0: Sorry for the confusion Goktug, I was suggesting we add a superstruct schema as a mechanism to define valid fields.

But I definitely **would not** recommend we add any additional validations in this PR, in the form of specific property types, as the scope of the change is specifically for total request size.

So should all these just be loose properties as they were before, at least for now?

So the only net change is the size validation?

matthewwalsh0 (changes requested):

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: In fact, we don't need this at all as we are validating vs `TransactionParamsStruct` so removed this.

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Thanks for the ping changed this to use `getJsonSize`. 
About performance concerns, I think this will be fine for benign requests as they are not massive.

packages/eth-json-rpc-middleware/src/utils/validation.ts
matthewwalsh0: Also here, I wouldn't recommend we change any errors or messages at this stage if possible, since the validation we'd like to add is very limited.

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: But size check may not enough alone?

Attackers are can bloat the request with tiny addition of object extending like `{"a":{"a":{"a":{"a"....` and even small size of the request can still overwhelm PPOM. Hence I was thinking implementing loose validation of shape. 

But if the intent was just implementing size validation, I am totally fine to remove all.

packages/eth-json-rpc-middleware/src/utils/validation.ts
FrederikBolding: `getJsonSize` has another performance penalty as it also validates that the input is JSON. We can skip that in this case. Maybe time to upstream: https://github.com/MetaMask/snaps/blob/a2ea600bc3cfa66a5e476fed4335d557c9cb2a4b/packages/snaps-utils/src/json.ts#L33-L37 😅

packages/eth-json-rpc-middleware/src/utils/validation.ts
matthewwalsh0: No I think the schema is a good idea to define what fields are allowed, but I'd be nervous to define strict types for those fields in this PR, as the risk to dApps is big given normalization etc.

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Applied the suggestion 😅

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +240,74 @@ export function validateTypedMessageKeys(data: string): void {
...
+ * serialized size limit.
+ * @throws rpcErrors.invalidInput() if params fails schema validation
+ * (wrong type, extraneous top-level key, or malformed nested field).
+ */
+export function validateTransactionParams(params: unknown): void {
+  if (
+    new TextEncoder().encode(JSON.stringify(params)).byteLength >
+    MAX_TRANSACTION_PARAMS_SIZE_BYTES
+  ) {
+    throw rpcErrors.invalidInput();
```
matthewwalsh0: Should we specify a useful message like `Request too large`?

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +240,74 @@ export function validateTypedMessageKeys(data: string): void {
...
+        r: optional(string()),
+        s: optional(string()),
+        yParity: optional(string()),
+      }),
+    ),
+  ),
+  chainId: optional(string()),
+  data: optional(string()),
+  from: string(),
+  gas: optional(string()),
```
matthewwalsh0: I think it's gas specifically which can accept numbers, but wondering if safest to support `number` or `string` for any numerical values for now, as `TransactionController` already normalises?

matthewwalsh0 (changes requested):

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Done 👍

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: Thanks Matthew, it make sense, using a union type here to accept both

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +242,74 @@ export function validateTypedMessageKeys(data: string): void {
...
+        r: optional(string()),
+        s: optional(string()),
+        yParity: optional(string()),
+      }),
+    ),
+  ),
+  chainId: optional(string()),
+  data: optional(string()),
+  from: string(),
+  gas: optional(union([string(), number()])),
```
matthewwalsh0: I meant for all the numerical values? Have you checked what transaction controller normalizes?

matthewwalsh0 (changes requested):

packages/eth-json-rpc-middleware/src/utils/validation.ts
OGPoyraz: I've extended all of those to `string | number`. I left `type` as string-only since `validateTxParams` validates it against the `TransactionEnvelopeType` string enum, where a number would always be rejected.

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +242,78 @@ export function validateTypedMessageKeys(data: string): void {
...
+
+// Numerical fields accept both hex strings and numbers, as some dapps send
+// numbers and `TransactionController` normalizes them downstream.
+const QuantityStruct = union([string(), number()]);
+
+export const TransactionParamsStruct = object({
+  accessList: optional(
+    array(object({ address: string(), storageKeys: array(string()) })),
+  ),
+  authorizationList: optional(
```
matthewwalsh0, line 254: Minor, we don't allow external authorization lists for security, but that will be caught in `TransactionController` for now.

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +242,78 @@ export function validateTypedMessageKeys(data: string): void {
     }
   }
 }
+
+// Numerical fields accept both hex strings and numbers, as some dapps send
+// numbers and `TransactionController` normalizes them downstream.
+const QuantityStruct = union([string(), number()]);
```
matthewwalsh0, line 248: ❤️

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +242,78 @@ export function validateTypedMessageKeys(data: string): void {
...
+
+export const TransactionParamsStruct = object({
+  accessList: optional(
```
matthewwalsh0, line 251: I'm sure we'll hit issues with some dApps as we can't predict all inputs, but we're long overdue schema validation here, so this should be a safe start 👍

matthewwalsh0 (dismissed):

packages/eth-json-rpc-middleware/src/utils/validation.ts
```diff
@@ -234,3 +242,78 @@ export function validateTypedMessageKeys(data: string): void {
...
+ * serialized size limit.
+ * @throws rpcErrors.invalidInput() if params fails schema validation
+ * (wrong type, extraneous top-level key, or malformed nested field).
+ */
+export function validateTransactionParams(params: unknown): void {
+  if (
+    new TextEncoder().encode(JSON.stringify(params)).byteLength >
+    MAX_TRANSACTION_PARAMS_SIZE_BYTES
+  ) {
+    throw rpcErrors.invalidInput('Request too large');
```
FrederikBolding:
```suggestion
    throw rpcErrors.invalidParams('Request too large');
```
Should this be invalidParams?

OGPoyraz: 👍 resolved in https://github.com/MetaMask/core/pull/9482/commits/a4e603a9c171a1b62dbbddc491ec42d634808486

FrederikBolding (approved):

PR #9484, opened 2026-07-13
feat(transaction-pay-controller): store no-op quotes for direct routes

dan437:
## Explanation

`TransactionPayController` stores an empty quotes array both when no conversion is needed and when a quote could not be fetched, so at publish time the two cases cannot be told apart.

This PR makes the distinction explicit:

- Store a no-op quote (`TransactionPayStrategy.None`) when a payment token is selected but the route needs no conversion. The no-op quote is built inside `getQuotes` when there are no quote requests, and is skipped when `isQuoteRequired` is set, since an empty request list then means source amounts could not be calculated.
- No-op quotes are not routable: the quote refresher skips them, they are excluded from totals and the external-sign flag, and `TransactionPayPublishHook` filters them out before selecting a strategy.
- Clients can rely on `quotes.length > 0` whenever MetaMask Pay is engaged, and enforce quote presence at publish time on their side.

## References

- Mobile adoption: MetaMask/metamask-mobile#33194

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes quote shape and publish-time behavior (breaking for quote consumers) in the transaction pay path, but logic is guarded with filters and broad tests rather than touching auth or funds directly.
> 
> **Overview**
> Direct MetaMask Pay routes (payment token selected, no conversion) now persist a **no-op quote** with strategy `none` instead of an empty `quotes` array, so **“no conversion needed”** is distinguishable from **“quote still missing.”**
> 
> `getQuotes` builds that marker via `buildNoOpQuote` when there are no quote requests and `isQuoteRequired` is false; fiat-only empty cases and quote-required flows still clear or omit it. **`TransactionPayStrategy.None` is not routable**—`isTransactionPayStrategy` excludes it, the publish hook skips execution, quote refresh/polling ignores only-no-op state, and fee totals / `hasQuotes` sync use **executable** quotes only.
> 
> **Breaking:** downstream clients that treat any entry in `quotes` as a real pay quote must ignore or handle `strategy: 'none'`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30646ebe39da15db7e655566c0f7ffcc526648c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/index.ts
```diff
@@ -40,3 +40,4 @@ export type {
 export { PaymentOverride, TransactionPayStrategy } from './constants';
 export { TransactionPayController } from './TransactionPayController';
 export { TransactionPayPublishHook } from './helpers/TransactionPayPublishHook';
+export { isNoOpQuote } from './utils/no-op-quote';
```
matthewwalsh0: This just encapsulates checking a single enum value so is it really warranted?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -172,8 +173,24 @@ export async function updateQuotes(
       transactionId,
     });
 
+    // A selected payment token with no conversion requests means the
+    // controller deemed the route direct. Store an explicit no-op quote so
+    // clients and the publish hook can distinguish "no conversion needed"
+    // from "quote needed but missing".
+    let finalQuotes = quotes;
+
+    if (
```
matthewwalsh0: Why are we doing it here after we've called `getQuotes`?

Shouldn't we only set this on line 623 inside `getQuotes` if there are no requests?

packages/transaction-pay-controller/src/helpers/TransactionPayPublishHook.ts
```diff
@@ -101,4 +112,40 @@ export class TransactionPayPublishHook {
...
+   * Validate that submitting the transaction without MetaMask Pay is safe.
+   *
+   * Direct submission is allowed only when MetaMask Pay was never engaged for
+   * the transaction, or when the controller explicitly marked the route as
+   * direct via a no-op quote and no conversion is currently required.
+   *
+   * @param transactionData - Pay state for the transaction.
+   * @param hasNoOpQuote - Whether a no-op quote is present.
+   */
+  #validateDirectSubmit(
```
matthewwalsh0: Do we need any changes in this file?

The `source-amounts.ts` and `quotes.ts` files already only generate quotes if there are source amounts, so this code should never be hit.

I assumed we'd just want to add the checks in the clients, to throw if the transaction type is in `[METAMASK_PAY_TYPES]` and no quotes?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/index.ts
dan437: Removed, will just compare `quote.strategy !== TransactionPayStrategy.None` directly.

packages/transaction-pay-controller/src/utils/quotes.ts
dan437: Agreed, moved it into getQuotes in the no-requests branch.

packages/transaction-pay-controller/src/helpers/TransactionPayPublishHook.ts
dan437: Agreed, all the throwing logic from the hook was removed. The only change left in this file is filtering out no-op quotes before picking a strategy, which is needed because otherwise the hook would call getStrategyByName('none') and fail instead of submitting directly.

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

PR #9496, opened 2026-07-14
test(base-data-service): Use fake timers

FrederikBolding:
## Explanation

Separated out for review from https://github.com/MetaMask/core/pull/9445, since it is needed for multiple PRs. Makes it easier to write tests for `BaseDataService` going forward.

## References

N/A

## Checklist

### DIFF
--- a/packages/base-data-service/src/BaseDataService.ts
+++ b/packages/base-data-service/src/BaseDataService.ts
@@ -271,6 +271,7 @@ export class BaseDataService<
    */
   destroy(): void {
     this.#queryCacheUnsubscribe();
+    this.#queryClient.clear();
     this.messenger.clearSubscriptions();
     this.messenger.clearActions();
   }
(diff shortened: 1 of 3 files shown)


packages/base-data-service/src/BaseDataService.ts
```diff
@@ -271,6 +271,7 @@ export class BaseDataService<
    */
   destroy(): void {
     this.#queryCacheUnsubscribe();
+    this.#queryClient.clear();
```
Mrtenz, line 274: Is this related to the tests or does this fix another bug? 🤔

FrederikBolding: The tests would hang waiting for a GC timer without this

Mrtenz (approved):

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Though it makes sense to clear it out when destroying regardless!

PR #9497, opened 2026-07-14
feat(transaction-pay-controller): add generic non-atomic post-Relay flow via `atomic` and `recipient`

OGPoyraz:
## Explanation

Introduces a **generic non-atomic post-Relay flow** on `transaction-pay-controller`, driven by a new field on `TransactionConfig` / `TransactionData`:

- **`atomic?: boolean`** — when `false`, TPC skips atomic quote embedding (no `processTransactions`, no `processMoneyAccountPostQuote`). The Relay quote bridges/swaps only, then a second leg runs after completion.

This unblocks two previously-blocked flows without adding another bespoke MA-specific module:

- **Perps/Predict withdraws to Money Account** — post-quote flows whose destination is mUSD on Monad. Settled amount is only known after Relay completes.
- **Max-amount Money Account deposits** — `isMaxAmount` forces EXACT_INPUT, so vault calls cannot be pre-encoded at quote time.

For both, Relay settles mUSD directly to a recipient **derived internally at quote time** (no client-configurable `recipient` field); once complete, a sponsored `addTransactionBatch` runs the second leg (approve + teller deposit for post-quote flows; the transaction's own nested calls for non-post-quote flows).

### Recipient derivation (quote time)

`resolveNonAtomicRecipient` in `relay-quotes.ts` derives where the Relay output must settle for non-atomic flows:

- **Post-quote** (Perps/Predict withdraw → MA): calls the client `getPaymentOverrideData` callback — the same source of truth the atomic path already uses via `processMoneyAccountPostQuote` — and uses its returned `recipient` (the Money Account address, which cannot be derived from the request). Falls back to `from` if the callback returns no recipient.
- **Non-post-quote** (max-amount MA deposit): uses the parent transaction's own `txParams.from`, which is the Money Account rather than the `accountOverride`-resolved funding EOA. Without this, max deposits settled mUSD on the EOA where the sponsored vault batch and CHOMP backup could not reach it.

The derived recipient is persisted on `quote.request.recipient` (internal-only) so the post-completion leg reads/settles against the correct account.

### Changes

**`types.ts`** — Adds `atomic` to `TransactionConfig` and `TransactionData`. `QuoteRequest.recipient` is now internal-only (derived, not client-set).

**`TransactionPayController.ts`** — Propagates `atomic` through `transactionData` ↔ `config` synchronisation (via spread + `Object.assign`).

**`utils/quotes.ts`** — Threads `atomic` through `buildQuoteRequests` / `buildPostQuoteRequests`.

**`strategy/relay/relay-quotes.ts`**
- New `resolveNonAtomicRecipient` (see above); the derived recipient feeds the quote body and the normalized quote request.
- `processTransactions` now returns a boolean and encapsulates all its skip conditions (skipProcessTransactions/isPostQuote, Polymarket deposit wallet, non-atomic) in a single early-return, so `getSingleQuote` routes to the alternate handlers when it declines.

**`strategy/relay/relay-submit.ts`** — After successful Relay completion, when `atomic === false`, runs `submitPostNonAtomic`:
- Resolves the actually-settled amount via `resolveSettledAmount`:
  - Cross-chain: reads the Transfer log on the polled target hash; throws on read failure rather than guessing (using the minimum would knowingly strand dust and defeat EXACT_INPUT).
  - Same-chain: polling is skipped (placeholder target hash), so the hash returned by `submitTransactions` — the transaction that actually moved the funds — is read instead. Relay execute submissions return the `FALLBACK_HASH` placeholder rather than a real hash, so only in that case does the quote's minimum output remain as the last available source.
- For `isPostQuote === true`, calls `getPaymentOverrideData` with the settled amount to build the deposit batch; prefers the callback's returned `recipient` as source of truth for the second-leg account. **Throws `Missing post-quote deposit calls` if the callback returns empty** — post-quote parent metas have no vault-side nested calls, so falling through to `getAmountData` would strand funds on the recipient.
- For `isPostQuote === false`, falls through to the transaction's own nested calls re-encoded via `getAmountData`.
- Delegates to `submitMoneyAccountVaultDeposit`.

**`utils/ma-vault-deposit.ts`** — Extends `submitMoneyAccountVaultDeposit` with two new optional params:
- `moneyAccountAddress` — explicit MA address override (used when `txParams.from` isn't the MA, e.g. Perps/Predict withdraws).
- `depositCalls` — pre-built batch. When provided, `resolveVaultDepositBatch` returns it directly, skipping the `getAmountData` re-encoding path.

**Removed** — `strategy/relay/relay-post-ma-vault.ts` and its tests. The dedicated MA post-Relay module is replaced by the generic `submitPostNonAtomic` in `relay-submit.ts`.

### Same-chain payment-override prepend gate

Also fixes a pre-existing bug in `relay-submit.ts`: the `paymentOverride` prepend onto the source execute batch is skipped for non-atomic flows — their second leg is submitted separately after completion, so prepending would double-embed the vault deposit. Cross-chain atomic flows (e.g. `moneyAccountWithdraw` on Monad settling USDC on Arbitrum for Perps) still prepend the source-side vault withdraw + transfer so Relay has funds to bridge.

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes Relay quote construction, post-completion settlement, and Money Account vault submission paths where incorrect amounts or recipients could strand user funds.
> 
> **Overview**
> Adds **`atomic?: boolean`** on `TransactionConfig` / `TransactionData` / `QuoteRequest` (default **true**). When **`atomic` is `false`**, Relay quotes **only bridge** to an internally derived **`recipient`**—no embedded `txs`—and the **second leg** (vault deposit) runs **after** Relay succeeds via new **`submitPostNonAtomic`** in `relay-submit.ts`.
> 
> **Quote time:** `resolveNonAtomicRecipient` sets the bridge recipient (post-quote: `getPaymentOverrideData`; max-deposit: parent tx `from`). **`processTransactions`** returns a boolean and skips embedding for non-atomic flows.
> 
> **Submit time:** Settled amount comes from on-chain Transfer logs when possible (stricter than the removed module); post-quote legs use **`getPaymentOverrideData`** with the settled human amount and optional **`depositCalls`** / **`moneyAccountAddress`** on **`submitMoneyAccountVaultDeposit`**. Non-atomic flows **skip** `paymentOverride` prepend to avoid double-embedding the vault step.
> 
> **Removed:** `relay-post-ma-vault.ts` and its tests—replaced by the generic path above.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbc766e54abc1b7cd4ba9719f164ae58ebb33a51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,6 +186,24 @@ async function executeSingleQuote(
...
+  ) {
+    const { transactionHash } = await submitPostRelayVaultDeposit({
+      completion,
+      messenger,
+      quote,
+      transaction,
+    });
+
+    return { transactionHash: transactionHash ?? completion.targetHash };
+  }
```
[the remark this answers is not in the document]
OGPoyraz: This is intended because the post vault deposit is sponsored and also we have CHOMP for backup.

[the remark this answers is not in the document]
OGPoyraz: Just wondering if this will cause any activity issues if we turn `isIntentComplete` early, will update here

packages/transaction-pay-controller/src/strategy/relay/relay-post-ma-vault.ts
```diff
@@ -134,3 +212,38 @@ async function resolvePostRelayAmount({
...
+    Boolean(request.isMaxAmount) &&
+    isMoneyAccountDepositTransaction(transaction);
+
+  const isWithdrawToMoneyAccount =
+    request.paymentOverride === PaymentOverride.MoneyAccount &&
+    (isPerpsWithdrawTransaction(transaction) ||
+      isPredictWithdrawTransaction(transaction)) &&
+    request.targetChainId === MUSD_MONAD_FIAT_ASSET.chainId &&
+    request.targetTokenAddress.toLowerCase() ===
+      MUSD_MONAD_FIAT_ASSET.address.toLowerCase();
```
OGPoyraz: Please refer the comment here: https://github.com/MetaMask/metamask-mobile/pull/33489/files#diff-8f54478706d449175887e700ecfc1431678458c2a75aa9bbb3a055c74f8a1f61R270

`isMaxAmount` can't be settled to `true` for predict and perps withdraw. Hence all perps/predict withdraw to `MoneyAccount` will use the post-Relay vaulting regardless of "max".

Obviously we can have a separate flag but I didn't to not bloat it. I am open for any suggestion to fix it cleanly.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -290,6 +291,20 @@ async function getSingleQuote(
   );
 
   try {
+    const usePostRelayVaultDeposit = isPostRelayMoneyAccountVaultDeposit(
+      request,
+      transaction,
+    );
```
OGPoyraz: It's important to predetermine if we will use post-Relay vaulting so that we can settle `recipient` as MA and skip atomic vault embedding because we will let MA to vault itself as it's sponsored and CHOMP as backup.

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -157,6 +157,7 @@ export class TransactionPayController extends BaseController<
         refundTo: transactionData.refundTo,
         accountOverride: transactionData.accountOverride,
         paymentOverride: transactionData.paymentOverride,
+        moneyAccountAddress: transactionData.moneyAccountAddress,
```
OGPoyraz: We will either have to add it like this: https://github.com/MetaMask/metamask-mobile/pull/33489/files#diff-7e3ad9d255d73f5dbc48f989c8991864aba5e51edb98ce587f309bc4fb496609R334

Or have a callback that returns `MoneyAccount` address, because for predict/perps withdraw to MA flows, it's not possible to derive MA address from the `request` at the quoting stage.

matthewwalsh0: Everything we've done so far is to try decouple from specific flows, can we trigger this same logic in a generic way?

matthewwalsh0: In the privacy work, I've added a flag here called `atomic` which when false, forces the transaction data to be executed manually after the quote is completed and funds are in place. Would that work here to abstract the entire second leg in a generic client triggered way?

packages/transaction-pay-controller/src/strategy/relay/relay-post-ma-vault.ts
matthewwalsh0, line 1: Can this entire file be abstracted through the transaction data itself?

packages/transaction-pay-controller/src/utils/transaction.ts
```diff
@@ -333,6 +333,27 @@ export function isPredictWithdrawTransaction(
...
+/**
+ * Check whether a transaction is a Perps withdrawal.
+ *
+ * Returns `true` when the transaction's own type is `perpsWithdraw`, or
+ * when any of its nested transactions has that type.
+ *
+ * @param transaction - Transaction metadata.
+ * @returns `true` when the transaction is a Perps withdrawal.
+ */
+export function isPerpsWithdrawTransaction(
```
matthewwalsh0: I think you suggested a while back that we export a generic `hasTransactionType` util from `TransactionController`? Could we use that here?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/utils/transaction.ts
OGPoyraz: Sure thing that's done now

OGPoyraz: Resolving this as I've done it separately

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -149,29 +149,33 @@ export class TransactionPayController extends BaseController<
...
         paymentOverride: transactionData.paymentOverride,
+        recipient: transactionData.recipient,
+        refundTo: transactionData.refundTo,
       };
 
       const previousAccountOverride = config.accountOverride;
 
       callback(config);
 
       transactionData.accountOverride = config.accountOverride;
```
matthewwalsh0: Minor, can we simplify both of these and just spread to set `config` then `Object.apply` after the callback?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -326,18 +326,28 @@ async function getSingleQuote(
...
     // Skip transaction processing when skipProcessTransactions (defaulting to
     // isPostQuote) is true — the original transaction will be included in the
     // batch separately, not as part of the quote.
     // Skip for Polymarket deposit wallet flows — the source is already a
     // bridged token transfer, not a contract call to embed.
+    // Skip for non-atomic flows — the second leg is submitted after settlement.
     const shouldProcessTransactions =
       !(request.skipProcessTransactions ?? request.isPostQuote) &&
-      !request.isPolymarketDepositWallet;
+      !request.isPolymarketDepositWallet &&
```
matthewwalsh0: Minor, since we already pass the `request` to `processTransactions`, would it be cleaner to include all those checks in there and early return `false` so it encapsulates that logic?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -326,18 +326,28 @@ async function getSingleQuote(
...
+      !isNonAtomic;
```
matthewwalsh0: Should we invert this to `isAtomic` so we're not checking if it's not not atomic? 😅

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -457,34 +472,40 @@ function buildQuoteRequests({
...
+  atomic?: boolean;
+  destinationToken: TransactionPaymentToken;
   from: Hex;
-  isMaxAmount: boolean;
   isHyperliquidSource?: boolean;
+  isMaxAmount: boolean;
   isPolymarketDepositWallet?: boolean;
   paymentOverride?: PaymentOverride;
-  destinationToken: TransactionPaymentToken;
+  recipient?: Hex;
```
matthewwalsh0: Minor, I think I regret decoupling this request type, maybe easier to just pass transaction ID,  source amounts, and transaction data in future.

Out of scope now.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+ * re-encoded via `getAmountData`.
+ *
+ * @param options - Submit options.
+ * @param options.completion - Outcome of `waitForRelayCompletion`.
+ * @param options.messenger - Controller messenger.
+ * @param options.quote - The Relay quote that was submitted.
+ * @param options.transaction - Original transaction meta.
+ * @returns Hash of the final submitted child transaction, if available.
+ */
+async function submitPostCompletionBatch({
```
matthewwalsh0: Minor, `submitPostNonAtomic` to keep it generic?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+      );
+
+      if (onChainAmount) {
+        log('Resolved settled amount from on-chain transaction', {
+          targetHash: completion.targetHash,
+          onChainAmount,
+        });
+        return onChainAmount;
+      }
+    } catch (error) {
```
matthewwalsh0: This seems bad, if we can't even work out how much was just sent, should we throw?

Using the minimum knowingly leaves dust which defeats the benefit of an `EXACT_INPUT` quote?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+  quote: TransactionPayQuote<RelayQuote>;
+  transaction: TransactionMeta;
+}): Promise<{ transactionHash?: Hex }> {
+  const sourceAmountRaw = await resolveSettledAmount({
+    completion,
+    messenger,
+    quote,
+  });
+
+  const recipient = quote.request.recipient ?? quote.request.from;
```
matthewwalsh0: The `getPaymentOverrideData` callback returns an optional `recipient` already, so do we even need `recipient` in that withdraw scenario?

For the deposit flow, the original `from` of the transaction should already be the money account so the recipient we need?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+  const { transactionData } = messenger.call(
+    'TransactionPayController:getState',
+  );
+
+  const { decimals } = quote.original.details.currencyOut.currency;
+  const amountHuman = new BigNumber(sourceAmountRaw)
+    .shiftedBy(-decimals)
+    .toFixed();
+
+  const { calls } = await messenger.call(
```
matthewwalsh0: This also returns the `recipient` which we can use as a better source of truth?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+    {
+      amount: amountHuman,
+      transaction,
+      transactionData: transactionData[transaction.id],
+    },
+  );
+
+  if (!calls.length) {
+    throw new Error(
+      'Missing post-quote deposit calls from getPaymentOverrideData',
```
matthewwalsh0: Minor, assuming this could be user facing, should we drop the callback name?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
matthewwalsh0: As mentioned in other comment, can get `recipient` here.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -182,9 +185,205 @@ async function executeSingleQuote(
...
+    .isPostQuote
+    ? await buildPostQuoteDepositCalls({
+        messenger,
+        sourceAmountRaw,
+        transaction,
+        quote,
+      })
+    : undefined;
+
+  return submitMoneyAccountVaultDeposit({
```
matthewwalsh0, line 285: I appreciate the intent for `atomic` is to make the flow generic, but I can see this function is essentially the generic logic we want, plus the CHOMP checks we need 👍 

So we can make this more generic in future as discussed, to apply to the fiat flow also.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/TransactionPayController.ts
OGPoyraz: Done 👍

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Moved the checks into `processTransactions`, now returns a boolean and early-returns false when skipping.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: 😄 done

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Good point, now throws when a cross-chain read fails/returns nothing. Same-chain (`FALLBACK_HASH`) still uses the quote minimum since there's no settlement tx to read.

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Now prefer the callback's recipient as source of truth but also kept `recipient` on the request since Relay still needs it at quote time

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Done

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: 👍 Agreed - good candidate to share with the fiat flow later, I will do a follow up here.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -392,13 +392,33 @@ function normalizeAuthorizationList(
...
+  const isAtomic = request.atomic !== false;
+
+  // Skip when skipProcessTransactions (defaulting to isPostQuote) is set — the
+  // original transaction is submitted separately, not embedded in the quote.
+  // Skip Polymarket deposit wallet flows — the source is already a bridged
+  // token transfer, not a contract call to embed. Skip non-atomic flows — the
+  // second leg is submitted after Relay settlement.
+  const skipProcessTransactions =
+    request.skipProcessTransactions ?? request.isPostQuote;
+  const shouldSkip =
```
matthewwalsh0: Minor, could both of these be one condition?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
```diff
@@ -208,9 +211,208 @@ async function executeSingleQuote(
...
+}: {
+  completion: RelayCompletionOutcome;
+  messenger: TransactionPayControllerMessenger;
+  quote: TransactionPayQuote<RelayQuote>;
+}): Promise<string> {
+  const recipient = (quote.request.recipient ?? quote.request.from) as
+    | Hex
+    | undefined;
+
+  const hasSettlementHash = Boolean(
```
matthewwalsh0: We will always have a hash, since the a transaction moved the funds, but maybe we need to get it from `submitTransactions` instead since we skip polling?

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
matthewwalsh0: But we already call `getPaymentOverrideData` at quote time too, for the atomic withdraw flow to get the transactions to put in the quote, so can't we just call it for non-atomic too?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-submit.ts
OGPoyraz: Removed `recipient` from quoting as well, it derives it from `resolveNonAtomicRecipient` => `getPaymentOverrideData `

matthewwalsh0 (approved):

PR #9502, opened 2026-07-14
fix(react-data-query): Retain queries in cache until GC

FrederikBolding:
## Explanation

Following this PR, instead of automatically removing queries from the UI query client when a `BaseDataService` instance evicts it from the cache, we retain the query in cache until the `QueryClient` itself cleans it up (via `cacheTime`).

This prevents issues with observers in the UI having their cache wiped just because a refetch hasn't happened in a while. Since the UI query client considers the data stale, it may also choose to refetch and repopulate the query in the `BaseDataService` instance before eviction anyways.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1168

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes cache lifecycle for all data-service-backed queries and could leave stale UI data visible longer if service eviction was relied on for freshness, though invalidation/hydration paths are unchanged.
> 
> **Overview**
> **UI query cache no longer mirrors `BaseDataService` evictions.** When the data service publishes a granular `removed` cache update, `createUIQueryClient` now ignores that event instead of calling `cache.remove`, so subscribed UI observers keep their data until TanStack Query’s own `cacheTime` garbage collection runs.
> 
> Tests replace the old “sync removal on remove event” case with fake-timer coverage: cache stays populated while observers are active even after the service’s long TTL, and entries are cleared only after observers are destroyed and the client `cacheTime` elapses. Changelog records the fix under Unreleased.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2a68488ba5db0892f7cb4d0e4e06209bac24e5b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-data-query/src/createUIQueryClient.test.ts
+++ b/packages/react-data-query/src/createUIQueryClient.test.ts
@@ -1,9 +1,10 @@
 import { Messenger } from '@metamask/messenger';
+import { Duration, inMilliseconds } from '@metamask/utils';
 import {
-  hashQueryKey,
   InfiniteData,
   InfiniteQueryObserver,
   QueryClient,
+  QueryClientConfig,
   QueryObserver,
 } from '@tanstack/query-core';
 
@@ -23,7 +24,7 @@ import { createUIQueryClient } from './createUIQueryClient';
 
 const DATA_SERVICES = ['ExampleDataService'];
 
-function createClients(): {
+function createClients(config?: QueryClientConfig): {
   service: ExampleDataService;
   clientA: QueryClient;
   clientB: QueryClient;
@@ -40,8 +41,8 @@ function createClients(): {
   >({ namespace: 'ExampleDataService' });
   const service = new ExampleDataService(serviceMessenger);
 
-  const clientA = createUIQueryClient(DATA_SERVICES, serviceMessenger);
-  const clientB = createUIQueryClient(DATA_SERVICES, serviceMessenger);
+  const clientA = createUIQueryClient(DATA_SERVICES, serviceMessenger, config);
+  const clientB = createUIQueryClient(DATA_SERVICES, serviceMessenger, config);
 
   return { service, clientA, clientB, messenger: serviceMessenger };
 }
@@ -65,6 +66,10 @@ describe('createUIQueryClient', () => {
     mockAssets();
     mockTransactionsPage1();
     mockTransactionsPage2();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
   });
 
   it('proxies requests to the underlying service', async () => {
@@ -218,8 +223,10 @@ describe('createUIQueryClient', () => {
     observerB.destroy();
   });
 
-  it('synchronizes cache removal after remove event', async () => {
-    const { messenger, clientA, clientB } = createClients();
+  it('does not remove from the cache if observers still are subscribed', async () => {
+    jest.useFakeTimers();
+
+    const { clientA, clientB } = createClients();
 
     const observerA = new QueryObserver(clientA, {
       queryKey: getAssetsQueryKey,
@@ -231,7 +238,7 @@ describe('createUIQueryClient', () => {
 
     const promiseA = new Promise((resolve) => {
       observerA.subscribe((event) => {
-        if (event.status === 'success') {
+        if (event.status === 'success' && !event.isFetching) {
           resolve(event.data);
         }
       });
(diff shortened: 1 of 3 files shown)


packages/react-data-query/src/createUIQueryClient.test.ts
```diff
@@ -67,8 +68,12 @@ describe('createUIQueryClient', () => {
     mockTransactionsPage2();
   });
 
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   it('proxies requests to the underlying service', async () => {
-    const { clientA: client } = createClients();
+    const { clientA: client, service } = createClients();
```
mcmire, line 76: Nit: Maybe we should change `createClients` to `withClients` and have it automatically call `service.destroy`? Seems like this would make writing tests easier in the future (but of course also adds to the size of this PR).

packages/react-data-query/src/createUIQueryClient.test.ts
```diff
@@ -231,40 +243,91 @@ describe('createUIQueryClient', () => {
...
         }
       });
     });
 
+    jest.advanceTimersByTime(0);
+
     await Promise.all([promiseA, promiseB]);
 
-    const hash = hashQueryKey(getAssetsQueryKey);
+    jest.advanceTimersByTime(inMilliseconds(1, Duration.Day));
```
mcmire, line 265: Where does this duration come from? Does it make sense for this duration to be set explicitly in the test somehow? Otherwise it makes this test a bit magical.

packages/react-data-query/src/createUIQueryClient.test.ts
```diff
@@ -231,40 +243,91 @@ describe('createUIQueryClient', () => {
...
     const promiseB = new Promise((resolve) => {
       observerB.subscribe((event) => {
-        if (event.status === 'success') {
+        if (event.status === 'success' && !event.isFetching) {
           resolve(event.data);
         }
       });
     });
 
+    jest.advanceTimersByTime(0);
```
mcmire, line 260: Why is this necessary? I've seen this in other tests but can't remember why. Is GC occurring at this step or no?

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding: Agreed that these tests will get hard to maintain long term, we can resolve this separately though

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding: Perhaps a comment would help? This is `cacheTime` for `ExampleDataService`: https://github.com/MetaMask/core/blob/dce810b8efeb95ffab054a2210f58d40a1e435e5/packages/base-data-service/tests/ExampleDataService.ts#L97

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding: All observer notifications from TanStack look something like this: `setTimeout(() => notify(), 0);`. So it is required in order for the subscription promise to resolve when using fake timers.

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding:
```suggestion
    // Advance the full cacheTime of ExampleDataService
    jest.advanceTimersByTime(inMilliseconds(1, Duration.Day));
```

mcmire: Ah. Gotcha. Yes, that would help.

packages/react-data-query/src/createUIQueryClient.test.ts
mcmire: Gotcha, okay. Curious if extracting this to a function that mentions TanStack in the name somewhere would help us remember this. We don't use fake timers much in this file so it's fine for now but maybe something to think about in the future.

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding: Added earlier today 👍

packages/react-data-query/src/createUIQueryClient.test.ts
FrederikBolding: Agreed, I think we should revisit the tests for `BaseDataService` and `createUIQueryClient` soon.

mcmire (approved):
LGTM.

PR #9505, opened 2026-07-14
fix: query layer 1 gas fee oracles via direct eth_call instead of ethers Web3Provider

cryptotavares:
## Explanation

**Current state:** `OracleLayer1GasFeeFlow` and `MantleLayer1GasFeeFlow` call OP-stack gas price oracle contracts using an ethers `Contract` backed by `Web3Provider`. Every read through this path (`contract.getL1Fee` → `BaseProvider.call`) first awaits `getNetwork()` → `detectNetwork()`, which internally depends on zero-delay `setTimeout` calls: the constructor's network-detection `setTimeout(…, 0)`, an `await timer(0)` inside `_uncachedDetectNetwork`, and the detect-network cache-clear `setTimeout(…, 0)`.

**Why it needs to change:** On React Native iOS, the JS timer pump is driven by a single `CADisplayLink` registered in `RCTDisplayLink`. When the display link stops receiving ticks from the render server (backboardd), all `setTimeout` callbacks silently stop firing — while the app remains visually foreground-active. This happens reproducibly when a WKWebView (in-app browser) becomes the active surface. Bridgeless RN has no zero-delay fast path, so even `setTimeout(…, 0)` waits for a display frame that never comes. As a result, ethers' `getNetwork()`/`detectNetwork()` timers never fire, the oracle `eth_call` is never dispatched, `updateTransactionLayer1GasFee` blocks indefinitely inside `addTransaction`, no approval is ever created, and the dapp confirmation modal never appears. This was reported as a Sev1 in MetaMask Mobile (MetaMask/metamask-mobile#32863) and reproduced on both iOS Simulator and physical iPhone.

**The fix:** Replace the `Contract`/`Web3Provider` dispatch path with:
1. `Interface.encodeFunctionData` (from `@ethersproject/abi`, already an indirect dependency) to ABI-encode the calldata for `getL1Fee`, `getOperatorFee`, and `tokenRatio`.
2. The controller's existing `rpcRequest` utility, which calls `provider.request` directly — a plain EIP-1193 call with no timer dependency.

`Layer1GasFeeFlowRequest` gains a `messenger` field so the flows can reach `rpcRequest`. The fix covers both the base oracle flow (`OracleLayer1GasFeeFlow`: `getL1Fee` + `getOperatorFee`) and `MantleLayer1GasFeeFlow` (`tokenRatio` in `transformOracleFee`). The fix was validated end-to-end on a live mobile repro: after the patch, the confirmation modal appears correctly while the RN timer pump is frozen.

No new npm dependencies are introduced — `@ethersproject/abi` was already transitively present; `@ethersproject/contracts` and `@ethersproject/providers` are no longer imported.

## References

* MetaMask/metamask-mobile#32863 — Sev1: dapp eth_sendTransaction confirmation modal never appears on iOS in-app browser

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches the `addTransaction` L1 fee path and introduces a breaking shape change on `Layer1GasFeeFlowRequest` for any custom flows, though behavior is covered by updated tests and targets a known mobile deadlock.
> 
> **Overview**
> Layer 1 gas fee flows no longer use ethers `Contract` + `Web3Provider` to read OP-stack and Mantle oracles. They ABI-encode `getL1Fee`, `getOperatorFee`, and `tokenRatio` with `@ethersproject/abi` and call the existing `rpcRequest` helper (`eth_call` on `provider.request`), avoiding ethers’ `setTimeout`-based network detection that can hang indefinitely on React Native iOS and block `addTransaction` / dapp confirmations.
> 
> **`Layer1GasFeeFlowRequest`** now carries **`messenger`** instead of **`provider`**; `updateTransactionLayer1GasFee` passes the messenger through and no longer resolves a provider up front. Oracle responses treat empty `0x` / non-string results as errors for L1 fee and Mantle `tokenRatio` (operator fee still defaults to zero on failure).
> 
> `@ethersproject/providers` is dropped from dependencies; tests are updated to mock `rpcRequest` and assert `eth_call` payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f191b5fc426e46a97ea64ce69accfdd3622e42a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add `messenger` property to `Layer1GasFeeFlowRequest` ([#9502](https://github.com/MetaMask/core/pull/9502))
-- Query layer 1 gas fee oracles via direct `eth_call` RPC requests instead of an ethers `Contract` backed by `Web3Provider` ([#9502](https://github.com/MetaMask/core/pull/9502))
+- Add `messenger` property to `Layer1GasFeeFlowRequest` ([#9505](https://github.com/MetaMask/core/pull/9505))
+- Query layer 1 gas fee oracles via direct `eth_call` RPC requests instead of an ethers `Contract` backed by `Web3Provider` ([#9505](https://github.com/MetaMask/core/pull/9505))
   - `Web3Provider` schedules its JSON-RPC dispatch with `setTimeout`, which never fires on React Native iOS when the timer pump is starved, blocking `addTransaction` indefinitely and preventing dapp confirmations from appearing ([MetaMask/metamask-mobile#32863](https://github.com/MetaMask/metamask-mobile/issues/32863))
 
 ## [69.0.0]
--- a/packages/transaction-controller/CHANGELOG.md
+++ b/packages/transaction-controller/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add `messenger` property to `Layer1GasFeeFlowRequest` ([#9502](https://github.com/MetaMask/core/pull/9502))
+- Query layer 1 gas fee oracles via direct `eth_call` RPC requests instead of an ethers `Contract` backed by `Web3Provider` ([#9502](https://github.com/MetaMask/core/pull/9502))
+  - `Web3Provider` schedules its JSON-RPC dispatch with `setTimeout`, which never fires on React Native iOS when the timer pump is starved, blocking `addTransaction` indefinitely and preventing dapp confirmations from appearing ([MetaMask/metamask-mobile#32863](https://github.com/MetaMask/metamask-mobile/issues/32863))
+
 ## [69.0.0]
 
 ### Changed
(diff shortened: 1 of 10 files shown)


packages/transaction-controller/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add `messenger` property to `Layer1GasFeeFlowRequest` ([#9505](https://github.com/MetaMask/core/pull/9505))
```
matthewwalsh0: Minor, is this unnecessary in changelog since `Layer1GasFeeFlowRequest` isn't exposed by the package?

packages/transaction-controller/src/types.ts
```diff
@@ -1527,6 +1527,9 @@ export type GasFeeFlow = {
 
 /** Request to a layer 1 gas fee flow to obtain layer 1 fee estimate. */
 export type Layer1GasFeeFlowRequest = {
+  /** The messenger instance. */
+  messenger: TransactionControllerMessenger;
+
   /** RPC Provider instance. */
   provider: Provider;
```
matthewwalsh0: Can we remove this at the same time since it's no longer used and encourages use of `rpcRequest` instead?

matthewwalsh0 (changes requested):

matthewwalsh0 (approved):

PR #9521, opened 2026-07-15
feat(assets-controller): add getAsset

Kriys94:
## Explanation

WPN-1472 centralizes the Assets domain in `@metamask/assets-controller` so consumers (including Snaps) can read controller-owned balances, metadata, prices, and fiat values instead of tracking them independently. Today the controller exposes `getAssets` (bulk, per-account map), `getAssetMetadata` (metadata only), and `getAssetsPrice`, but there is no way to read a single combined asset for a specific account/asset pair.

This PR adds a `getAsset` read API to `AssetsController`:

- **`getAsset(accountId, assetId): Asset | undefined`** returns the combined `Asset` (`balance` + `metadata` + `price` + computed `fiatValue`) for a single account/asset pair from current controller state, or `undefined` when a complete renderable asset is not available.
- Exposed as the **`AssetsController:getAsset`** messenger action so it can be called through the messenger (e.g. by preinstalled Snaps once WPC-1017 lands).

Behavior details:

- Reuses the same state-composition and filtering logic as `getAssets` (via the private `#getAssetsFromState`) so the returned shape never drifts: `balance` and `metadata` are required; a missing price falls back to `{ price: 0, lastUpdated: 0 }` with `fiatValue: 0`; hidden/filtered assets resolve to `undefined`.
- Reads from current state only — it does **not** trigger a data-source refresh.
- Validates inputs and throws (rather than silently returning `undefined`) on an empty `accountId` or a malformed CAIP-19 `assetId`, so consumers can distinguish bad requests from absent data.
- Normalizes the `assetId` (checksums EVM ERC-20 addresses) before lookup, consistent with how state is keyed.
- `#getAssetsFromState`'s `accounts` parameter was widened to `Pick<InternalAccount, 'id'>[]` (it only reads `account.id`) so `getAsset` can reuse it without constructing a full account object.

The generated messenger action types (`AssetsControllerGetAssetAction`) were regenerated and exported from `index.ts`, and a `### Added` changelog entry was added.

## References

- Jira: [WPN-1473](https://consensyssoftware.atlassian.net/browse/WPN-1473) — [AssetsController] Add `getAsset` read API for Snap consumers
- Part of the Assets migration epic (WPN-1472); sibling controller-side stories are WPN-1489 (stage-gated ingestion) and WPN-1490 (stage-gated ignoring of Snap events)
- Consumer-side exposure to Snaps is blocked by WPC-1017 (messenger API exposure); the controller-side method in this PR is independent of that work

## Checklist

[WPN-1473]: https://consensyssoftware.atlassian.net/browse/WPN-1473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Additive read-only API with shared composition logic and broad unit tests; no fetch or persistence behavior changes beyond internal refactor of `getAssets` state assembly.
> 
> **Overview**
> Adds **`getAsset(accountId, assetId)`** and the **`AssetsController:getAsset`** messenger action so callers can read one combined **`Asset`** (balance, metadata, price, **`fiatValue`**) from current state without running **`getAssets`** or triggering a fetch. Behavior matches **`getAssets`** for composition and filtering (balance + metadata required, zero price fallback, hidden assets excluded); invalid **`accountId`** / **`assetId`** throw, and EVM IDs are normalized before lookup. **`getAsset`** is limited to fungible assets via the shared helper.
> 
> **`#getAssetsFromState`** now delegates per-asset work to a new **`#getAssetFromState`** helper (also used by **`getAsset`**), takes **`Pick<InternalAccount, 'id'>[]`**, and builds account/asset maps with null-prototype objects for prototype-pollution hardening; tests assert empty maps via **`Object.keys`** instead of **`{}`** equality.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9832d81e5191d42f356fbad957a8e7f546d1ed17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.test.ts
+++ b/packages/assets-controller/src/AssetsController.test.ts
@@ -875,6 +875,154 @@ describe('AssetsController', () => {
     });
   });
 
+  describe('getAsset', () => {
+    const metadata = {
+      type: 'erc20' as const,
+      symbol: 'USDC',
+      name: 'USD Coin',
+      decimals: 6,
+    };
+
+    it('returns the combined asset with computed fiatValue', async () => {
+      const initialState: Partial<AssetsControllerState> = {
+        assetsInfo: { [MOCK_ASSET_ID]: metadata },
+        assetsBalance: {
+          [MOCK_ACCOUNT_ID]: { [MOCK_ASSET_ID]: { amount: '100' } },
+        },
+        assetsPrice: { [MOCK_ASSET_ID]: { price: 2, lastUpdated: 123 } },
+      };
+
+      await withController({ state: initialState }, ({ controller }) => {
+        const asset = controller.getAsset(MOCK_ACCOUNT_ID, MOCK_ASSET_ID);
+
+        expect(asset).toStrictEqual({
+          id: MOCK_ASSET_ID,
+          chainId: 'eip155:1',
+          balance: { amount: '100' },
+          metadata,
+          price: { price: 2, lastUpdated: 123 },
+          fiatValue: 200,
+        });
+      });
+    });
+
+    it('normalizes a lowercase EVM asset ID before lookup', async () => {
+      const initialState: Partial<AssetsControllerState> = {
+        assetsInfo: { [MOCK_ASSET_ID]: metadata },
+        assetsBalance: {
+          [MOCK_ACCOUNT_ID]: { [MOCK_ASSET_ID]: { amount: '1' } },
+        },
+        assetsPrice: { [MOCK_ASSET_ID]: { price: 1, lastUpdated: 1 } },
+      };
+
+      await withController({ state: initialState }, ({ controller }) => {
+        const asset = controller.getAsset(
+          MOCK_ACCOUNT_ID,
+          MOCK_ASSET_ID_LOWERCASE,
+        );
+
+        expect(asset?.id).toBe(MOCK_ASSET_ID);
+      });
+    });
+
+    it('returns undefined when the balance is missing', async () => {
+      const initialState: Partial<AssetsControllerState> = {
+        assetsInfo: { [MOCK_ASSET_ID]: metadata },
+      };
+
+      await withController({ state: initialState }, ({ controller }) => {
+        expect(
+          controller.getAsset(MOCK_ACCOUNT_ID, MOCK_ASSET_ID),
+        ).toBeUndefined();
+      });
+    });
+
+    it('returns undefined when the metadata is missing', async () => {
+      const initialState: Partial<AssetsControllerState> = {
+        assetsBalance: {
+          [MOCK_ACCOUNT_ID]: { [MOCK_ASSET_ID]: { amount: '100' } },
+        },
+      };
+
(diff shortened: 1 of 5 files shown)


packages/assets-controller/src/AssetsController.ts
```diff
@@ -1775,6 +1776,51 @@ export class AssetsController extends BaseController<
...
+      );
+    }
+
+    const chainId = extractChainId(normalizedAssetId);
+
+    const assets = this.#getAssetsFromState(
+      [{ id: accountId }],
+      [chainId],
+      ['fungible'],
+    );
```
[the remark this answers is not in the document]
Kriys94: Well this function will be used by Snaps independently from the enabledChains

Prithpal-Sooriya: nit - A tad nuts that we are calculating all assets, only to return a single asset here.

This is okay for now, we can raise/improve moving forward.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -2698,7 +2744,7 @@ export class AssetsController extends BaseController<
   }
 
   #getAssetsFromState(
-    accounts: InternalAccount[],
+    accounts: Pick<InternalAccount, 'id'>[],
```
Prithpal-Sooriya, line 2745: Very good! Keep interface minimal!
Easier to test and use.

packages/assets-controller/src/AssetsController.test.ts
```diff
@@ -856,6 +856,154 @@ describe('AssetsController', () => {
     });
   });
 
+  describe('getAsset', () => {
```
Prithpal-Sooriya, line 878: 🌶️  this test file is too big, impossible to read!

Thoughts on:
1. Making this spec more dry. I think we can leverage test tables (`it.each`)
2. spicer -- separate test for each method, truly isolated, easy to read 😈

Prithpal-Sooriya: I'm fine keeping this as is... but I'd like to pencil in test cleanup.

Prithpal-Sooriya (dismissed):
Approving with comments

packages/assets-controller/src/AssetsController.ts
Kriys94: Great comment, I've refactored a bit to have `#getAssetFromState` and   `#getAssetsFromState` is using `#getAssetFromState`

packages/assets-controller/src/AssetsController.test.ts
Kriys94: YEah probably the best time will be after some cleaning

Prithpal-Sooriya (approved):

PR #9409, opened 2026-07-07
feat(ramps-controller): consume moneyHeadlessAllProviders flag controller-side; share provider/quote/error helpers

saustrie-consensys:
## Explanation

Headless Buy provider expansion previously reached `RampsController` through an injected `getProviderScope` callback returning a multivariate scope (`off` | `in-app` | `all`), and `metamask-mobile` re-derived gating logic the controller already owns privately. This PR replaces the scope concept with a single boolean remote feature flag that the controller reads itself, and lifts the shared logic into pure, tested, exported helpers so consumers cannot drift from the controller's own selection.

Flag semantics (`moneyHeadlessAllProviders`, boolean; this is the client-side key, which the ClientConfig API camelizes from the kebab-case LaunchDarkly dashboard key `money-headless-all-providers`):

- `true`: the auto-selection quote path (`autoSelectProvider` / `restrictToKnownOrNativeProviders`) widens to every supporting provider class: native, in-app WebView aggregator, and external-browser / custom-action. The best quote is returned at `success[0]`, enforcing per-provider fiat limits.
- `false`, missing, or any non-boolean value: today's native-only behavior. The intermediate `in-app` scope is removed, along with the exported `ProviderScope` type and the `getProviderScope` constructor option (breaking).

Controller-side consumption:

- `RemoteFeatureFlagController:getState` is added to the controller's allowed messenger actions (same pattern as `ConfigRegistryController` / `bridge-controller`), and a private `#isAllProvidersEnabled()` resolves the flag per `getQuotes` call, so a remote fetch or a local dev override takes effect at runtime. The read fails closed: if the action is not delegated, quoting stays native-only.
- New `featureFlags` module exports `MONEY_HEADLESS_ALL_PROVIDERS_FLAG_KEY` and the pure `isHeadlessAllProvidersEnabled(remoteFeatureFlagState)` helper (plus a `HeadlessFeatureFlagsLookup` type). The helper owns the key lookup, the `localOverrides`-over-`remoteFeatureFlags` merge (mobile's Settings > Feature flag override screen writes `localOverrides`, and published `RemoteFeatureFlagController` versions such as `4.2.2` do not fold overrides into `remoteFeatureFlags` state), and defensive coercion: only the literal boolean `true` enables. Mobile UI hooks call the same helper, so the UI gate and the controller resolve the flag identically.

Shared pure helpers (consumed by the controller internally so they stay behaviourally identical to its own selection):

[...]

PR #9524, opened 2026-07-15
feat(ramps-controller): accept a JSON `moneyHeadlessAllProviders` payload carrying provider allowlists

saustrie-consensys:
## Explanation

`moneyHeadlessAllProviders` is currently a boolean remote flag: the literal `true` widens headless fiat auto-selection quoting to every provider class (#9409). Phase 2 of the Headless Buy all-providers work needs QA to force a specific provider mix (for example external-browser-only) without a client change per test configuration.

This PR widens the flag's value contract so the same LaunchDarkly key can carry provider allowlists:

- The flag now also accepts an object payload `{ "enabled": true, "featureVersion": "1", "providerIds": ["/providers/x"] }`. `isHeadlessAllProvidersEnabled` returns `true` for a payload whose `enabled` is the literal `true` and `featureVersion` is `"1"`; every other value resolves to `false`, and the boolean forms behave exactly as before. In LaunchDarkly the variation is delivered inside a `versions` envelope (e.g. `{ "versions": { "0.0.0": { "enabled": false, "featureVersion": "1" }, "8.4.0": { "enabled": true, "featureVersion": "1" } } }`); `RemoteFeatureFlagController` resolves that to the flat payload before writing to state, so ramps code always evaluates the flat form.
- New pure `getHeadlessProviderAllowlist(remoteFeatureFlagState)` resolves the allowlist from the payload's top-level `providerIds`; empty or malformed lists resolve to `undefined` (no restriction). Unknown keys and non-string entries are ignored.
- When the resolved allowlist is non-empty, the widened pick drops candidates whose provider is not listed, before the existing fiat-limit filter and reliability-then-price ranking. Ids match in prefixed (`/providers/x`) or bare form, case-insensitively, via a module-local normalizer used only for allowlist matching (provider ids still match as-is everywhere else; the v16 removal of id normalization is not reverted). If nothing survives, `getQuotes` returns an empty `success[]` with `sorted` / `error` / `customActions` preserved.
- The controller resolves the enabled bit and the allowlist from a single `RemoteFeatureFlagController:getState` read per `getQuotes` call, so a flag edit during the awaited fetch cannot produce a mixed read.

Backwards compatibility, and why this is a minor (not breaking) change: clients on earlier versions coerce any non-boolean value to `false` (native-only), so serving the object payload can never turn widening on for a client that cannot parse it, and production serves `false` everywhere today. No consumer needs code changes to keep working; boolean-only configurations see identical behavior. The documented "any non-boolean resolves to false" contract is widened, not changed, for the values it previously rejected.

## References

- Implements the flag-expansion milestone (P2.M5, "Granular LaunchDarkly provider control") of the Headless Buy all-providers plan in `MetaMask/metamask-mobile` (`app/components/UI/Ramp/headless/PLAN_-_ALL_PROVIDERS_SUPPORT.md`), revised to evolve the single existing flag instead of adding a companion flag.
- Builds on #9409 (boolean flag consumed controller-side).
- Consumer PR (version bump): MetaMask/metamask-mobile#33362

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes which provider wins headless fiat auto-select when the flag carries allowlists; gated by remote flags and backward-compatible for boolean `true`, but misconfigured LD payloads could yield empty quotes in production.
> 
> **Overview**
> Extends **`moneyHeadlessAllProviders`** so LaunchDarkly can send an object payload (`enabled`, **`featureVersion: '1'`**, optional **`providerIds`**, optional **`minimumVersion`**) in addition to boolean `true`. **`isHeadlessAllProvidersEnabled`** treats only literal `true` or a version-matched enabled object as on; other shapes (including `{ enabled: true }` without `featureVersion`) stay off, while boolean-only configs behave as before.
> 
> New helpers **`getHeadlessProviderAllowlist`**, **`getHeadlessAllProvidersMinimumVersion`**, and **`HEADLESS_ALL_PROVIDERS_FEATURE_VERSION`** centralize parsing with the same remote/local override merge. **`RampsController.getQuotes`** reads enabled + allowlist in one flag snapshot; on the widened auto-select path, **`#pickWidenedQuote`** drops quotes from unlisted providers (case-insensitive match after normalizing `/providers/` vs bare ids) before fiat limits and reliability/price ranking. If every candidate is filtered out, **`success[]`** is empty while **`sorted`**, **`error`**, and **`customActions`** are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7737b6defffe88384ff226a494fcf2f7041030dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saustrie-consensys:
@metamaskbot publish-preview

saustrie-consensys:
@metamaskbot publish-preview

saustrie-consensys:
@metamaskbot publish-preview

packages/ramps-controller/src/featureFlags.ts
```diff
@@ -22,17 +58,102 @@ export type HeadlessFeatureFlagsLookup = {
...
+): string | undefined {
+  const value = resolveFlagValue(remoteFeatureFlagState);
+  if (!isEnabledPayload(value)) {
+    return undefined;
+  }
+  const { minimumVersion } = value;
+  return typeof minimumVersion === 'string' && minimumVersion.trim() !== ''
+    ? minimumVersion
+    : undefined;
+}
```
amitabh94, line 108: Do we not get these utils from feature flag controller itself? Looks like we are reinventing the wheel for flag mechanism.

amitabh94:
Is the flag shape mentioned in PR description matching the actual LD flag shape?
https://app.launchdarkly.com/projects/metamask-client-config-api-mobile/flags/money-headless-all-providers/variations?env=main-exp&env=qa-prod&env=test&env=production&selected-env=main-exp

saustrie-consensys:
@amitabh94 I just modified the PR body to match the new changes within LaunchDarkly

packages/ramps-controller/src/featureFlags.ts
saustrie-consensys: These aren't functions that are already exported in remote-feature-flag-controller's index.ts. I could export it (which would also meaning version bumping `remote-feature-flag-controller` too instead of just `ramps-controller`). 

`transaction-pay-controller` faced a similar problem before, and they chose to do it in-line instead of modifying the remote-feature-flag-controller to export a shared utility: https://github.com/MetaMask/core/blob/505289fe98/packages/transaction-pay-controller/src/utils/feature-flags.ts#L791-L802

I'm pro keeping things DRY, but I didn't want to bump `remote-feature-flag-controller`. I can make the change if you agree with my initial thought process of bumping `remote-feature-flag-controller`. @amitabh94

amitabh94: We can keep it inline but this should be solved by the @MetaMask/mobile-platform team since multiple teams are running into this. 

Maybe we can raise a PR for the platform team to review but I won't block this PR for this then.

packages/ramps-controller/src/featureFlags.ts
```diff
@@ -43,9 +164,80 @@ export type HeadlessFeatureFlagsLookup = {
...
+ * relevant subset of it). May be `null`/`undefined` before the controller is
+ * initialized.
+ * @param surface - Optional consumer surface key; canonical values are the
+ * {@link HEADLESS_ALLOWLIST_SURFACES} members.
+ * @returns The provider ids the widened pick is restricted to, or `undefined`
+ * for no restriction.
+ */
+export function getHeadlessProviderAllowlist(
+  remoteFeatureFlagState: HeadlessFeatureFlagsLookup | null | undefined,
+  surface?: string,
```
amitabh94: Do we need surface here?

amitabh94 (dismissed):

packages/ramps-controller/src/featureFlags.ts
saustrie-consensys: Removed it. https://github.com/MetaMask/metamask-mobile/pull/33362#discussion_r3646723093

amitabh94 (approved):

PR #9526, opened 2026-07-15
chore(base-data-service): simplify `invalidateQueries` type

Gudahtt:
## Explanation

The generic parameter has been removed from the `invalidateQueries` method, simplifying the type signature. In practice this was never used. It also created an odd inconsistency between the method type and the action type, which would be nice to avoid.

## References

N/A

## Checklist

  - I didn't test with a preview build, but I did search the entire organization for any references where the generic type param was used, and found none

---

> [!NOTE]
> **Low Risk**
> Narrow API/type-signature change with identical runtime behavior; breaking only for external callers that explicitly parameterized `invalidateQueries`, which the team reports were none.
> 
> **Overview**
> **Breaking (documented):** `BaseDataService.invalidateQueries` no longer accepts a `TPageData` type parameter; optional filters are now typed as `InvalidateQueryFilters<Json>` and still delegate to the query client unchanged.
> 
> This aligns the public method signature with `DataServiceInvalidateQueriesAction`, which already used `Json` for filters. The unreleased changelog notes the break even though no in-org callers used the generic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2faae3512a4e6a025d91c2dfee0cff6ca55b29af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/base-data-service/src/BaseDataService.ts
+++ b/packages/base-data-service/src/BaseDataService.ts
@@ -336,8 +336,8 @@ export class BaseDataService<
    * @param options - Additional optional options for query invalidations.
    * @returns Nothing.
    */
-  async invalidateQueries<TPageData extends Json>(
-    filters?: InvalidateQueryFilters<TPageData>,
+  async invalidateQueries(
+    filters?: InvalidateQueryFilters<Json>,
     options?: InvalidateOptions,
   ): Promise<void> {
     return this.#queryClient.invalidateQueries(filters, options);
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **BREAKING:** Remove `TPageData` type parameter from `invalidateQueries` method ([#9526](https://github.com/MetaMask/core/pull/9526))
+  - This is technically a breaking change, but this was not used in any of our codebases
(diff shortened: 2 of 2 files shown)


packages/base-data-service/src/BaseDataService.ts
```diff
@@ -336,8 +336,8 @@ export class BaseDataService<
    * @param options - Additional optional options for query invalidations.
    * @returns Nothing.
    */
-  async invalidateQueries<TPageData extends Json>(
```
mcmire, line 339: Interesting 🤔 I was going to say that this type parameter was probably added to match, or at least mimic, the [type signature of `invalidateQueries` in `QueryClient`](https://github.com/TanStack/query/blob/%40tanstack/query-core%404.43.0/packages/query-core/src/queryClient.ts#L402). But the type of the handler in `DataServiceInvalidateQueriesAction` doesn't include this type parameter. So your change makes sense, if at least to match the action.

By the way I realized that while we've upgraded to `@tanstack/query-core` ^5.62.16 in `react-data-query`, we are still using ^4.43.0 in this package. I think that was an oversight in a previous PR. That may end up changing this type signature again, just a heads up.

mcmire (approved):
LGTM.

packages/base-data-service/src/BaseDataService.ts
Gudahtt: Oh good catch. We should update query-core before we release

PR #9536, opened 2026-07-16
feat!: replace hybrid CJS/ESM build with ESM-only

Mrtenz:
## Explanation

Currently, packages in this monorepo are built as both CommonJS and ESM using `@ts-bridge/cli`. This PR replaces that with ESM-only output built with standard `tsc` and TypeScript project references.

The motivation is to simplify the build toolchain by removing the `ts-bridge` dependency in preparation for the TypeScript 6+ upgrade.

## Checklist

---

> [!NOTE]
> **High Risk**
> Every published package drops CommonJS and changes export surfaces, so MetaMask clients and other consumers need coordinated upgrades. The breadth of build, test, and import changes increases regression risk across the monorepo.
> 
> **Overview**
> Replaces **`@ts-bridge/cli`** with **`tsc --build`** for monorepo and per-package builds. CI partial builds use `yarn tsc --build` instead of `ts-bridge`. Clean steps also remove `tsconfig.build.tsbuildinfo`. Root and every workspace package set **`"type": "module"`** and publish a single ESM entry (`dist/index.js` + `dist/index.d.ts`) instead of dual `import`/`require` export maps.
> 
> **Breaking:** package changelogs document **ESM-only** output (no published CommonJS). Downstream apps must use native ESM or Node 22+ `require(esm)` / dynamic import.
> 
> Tooling follows the module split: Jest configs move to **`.cjs`**, package tests compile sources to CommonJS via **ts-jest** (with **`lodash-es` → `lodash`** mapping), and script tests run as ESM. ESLint/knip/docs reference **`jest.config.cjs`** and updated file globs.
> 
> Source fixes for ESM: widespread **`lodash` → `lodash-es`**, explicit **`.js` import extensions**, **`bignumber.js` named imports**, small type/import workarounds (`uuid`, `@metamask/api-specs`, selector typing), and removal of **`@ts-bridge/*`** from Yarn preapproved packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0ebb9fb16052a01f2c4893c51bd7b2bcbf43d3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/account-tree-controller/package.json
```diff
@@ -18,19 +18,12 @@
...
-        "types": "./dist/index.d.mts",
-        "default": "./dist/index.mjs"
-      },
-      "require": {
-        "types": "./dist/index.d.cts",
-        "default": "./dist/index.cjs"
-      }
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
     },
```
Mrtenz, line 27: Removed `import`/`require` distinction to have Node.js resolve files regardless of whether it's running in ESM or CommonJS mode. This should work most of the time since `require(esm)` is possible now.

packages/accounts-controller/src/AccountsController.ts
```diff
@@ -34,7 +34,7 @@ import type { NetworkClientId } from '@metamask/network-controller';
 import { isCaipChainId } from '@metamask/utils';
 import type { CaipChainId } from '@metamask/utils';
 import type { WritableDraft } from 'immer/dist/internal.js';
-import { cloneDeep } from 'lodash';
+import { cloneDeep } from 'lodash-es';
```
Mrtenz, line 37: `lodash` does not work in ESM.

packages/analytics-controller/src/AnalyticsController.ts
```diff
@@ -24,7 +24,7 @@ import type {
   AnalyticsLocationContext,
   AnalyticsUserTraits,
   AnalyticsTrackingEvent,
-} from './AnalyticsPlatformAdapter.types';
+} from './AnalyticsPlatformAdapter.types.js';
```
Mrtenz, line 27: Missed a couple file extension refactors in my previous PRs, since these imports already have ".[something]". 😅

packages/assets-controller/src/data-sources/evm-rpc-services/utils/parsing.ts
```diff
@@ -5,7 +5,7 @@ import {
   parseCaipChainId,
 } from '@metamask/utils';
 import type { CaipChainId, Hex } from '@metamask/utils';
-import BigNumberJS from 'bignumber.js';
+import { BigNumber as BigNumberJS } from 'bignumber.js';
```
Mrtenz, line 8: `bignumber.js` does not have a default import in ESM.

packages/assets-controllers/src/selectors/token-selectors.ts
```diff
@@ -461,7 +461,11 @@ const selectAllMultichainAssets = createAssetListSelector(
   },
 );
 
-export const selectAllAssets = createAssetListSelector(
+// TS2742: The inferred type of 'selectAllAssets' cannot be named without a
+// reference to '../../../../node_modules/@metamask/account-tree-controller/dist/wallet.js'.
+// This is likely not portable. A type annotation is necessary.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const selectAllAssets: any = createAssetListSelector(
```
Mrtenz, line 468: Unsure why this is a problem now. `@ts-expect-error` does not work in this specific instance, so I solved it by adding an `any` annotation for now.

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -53,6 +53,26 @@ import type {
...
+const walletModule = (typeof ethereumjsWallet.default === 'object' &&
+ethereumjsWallet.default
+  ? ethereumjsWallet.default
+  : ethereumjsWallet) as unknown as {
+  default: typeof ethereumjsWallet.default;
+  thirdparty: typeof ethereumjsWallet.thirdparty;
+};
+
+const Wallet = walletModule.default;
+const importers = walletModule.thirdparty;
```
Mrtenz, line 74: This may be a bug in Node.js 22 as it works without this in Node.js 24. It can't detect the named imports, so importing `thirdparty` fails.

This boilerplate is necessary to make it work in Jest (transpiling to CommonJS) and ESM.

packages/kyc-controller/src/KycController.ts
```diff
@@ -1461,6 +1461,8 @@ export class KycController extends BaseController<
       // `@ts-expect-error` cannot be used: ts-bridge does not surface
       // TS2589, so the directive is unused and fails the build.
       // type issue only happens at the IDE level.
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore: TODO: Investigate issue.
```
Mrtenz: The comment above indicates that there may have been a `@ts-ignore` here before. I just added it back as it seems to be a problem again.

packages/messenger-cli/src/check.test.ts
```diff
@@ -1,3 +1,4 @@
+import { jest } from '@jest/globals';
```
Mrtenz, line 1: `messenger-cli` runs Jest in ESM mode, which requires some additional changes such as importing the `jest` global where needed. The types of the `jest` global imported from `@jest/globals` are a bit more strict, hence the changes below.

packages/platform-api-docs/src/cli.test.ts
```diff
@@ -149,9 +149,9 @@ export type QuxMessenger = Messenger<'Qux', QuxAction, never>;
     const result = await runCLI(['--help']);
 
     expect(result.exitCode).toBe(0);
-    expect(result.all).toContain(
-      'Produces documentation for the platform API, the set of actions and events\navailable in clients through the message bus',
-    );
+    // Keep this short: yargs wraps the help text to the terminal width, so a
+    // longer excerpt would depend on where the tests run.
+    expect(result.all).toContain('Produces documentation for the platform API');
```
Mrtenz, line 154: This was breaking now in CI, unsure what determines the line length, but manually splitting with a newline seems brittle anyway.

packages/smart-transactions-controller/src/utils.ts
```diff
@@ -7,7 +7,7 @@ import type {
 } from '@metamask/transaction-controller';
 import { TransactionStatus } from '@metamask/transaction-controller';
 import { BigNumber } from 'bignumber.js';
-import _ from 'lodash';
+import { camelCase, isArray, isObject, mapKeys, mapValues } from 'lodash-es';
```
Mrtenz, line 10: `lodash-es` doesn't have a default export.

Gudahtt:
I've assigned the `DO-NOT-MERGE` label temporarily until we've discussed how to roll this out to clients, and have prepared for it adequately. As this will be a breaking change for all packages, we'll need to be prepared to roll this out quickly to avoid disruption.

Mrtenz:
> I've assigned the `DO-NOT-MERGE` label temporarily until we've discussed how to roll this out to clients, and have prepared for it adequately. As this will be a breaking change for all packages, we'll need to be prepared to roll this out quickly to avoid disruption.

@Gudahtt This PR targets `mrtenz/drop-node-20`. I've applied the label to that PR instead.

packages/account-tree-controller/package.json
```diff
@@ -39,8 +32,8 @@
     "registry": "https://registry.npmjs.org/"
   },
   "scripts": {
-    "build": "ts-bridge --project tsconfig.build.json --verbose --clean --no-references",
-    "build:all": "ts-bridge --project tsconfig.build.json --verbose --clean",
+    "build": "tsc --project tsconfig.build.json",
```
mcmire, line 35: Previously `yarn build` would automatically clean `dist`. With these changes it appears that is no longer the case. Is it worth adding a `build:clean` script to all packages in case that is needed? Or should we drop it for now?

packages/assets-controllers/src/selectors/token-selectors.ts
mcmire: Yeah that's definitely interesting...

packages/authenticated-user-storage/package.json
```diff
@@ -39,15 +32,15 @@
...
+    "build": "tsc --project tsconfig.build.json",
+    "build:all": "tsc --build tsconfig.build.json --verbose",
     "build:docs": "typedoc",
     "changelog:update": "../../scripts/update-changelog.sh @metamask/authenticated-user-storage",
     "changelog:validate": "../../scripts/validate-changelog.sh @metamask/authenticated-user-storage",
     "lint:tsconfigs": "tsx ../../scripts/lint-tsconfigs/lint-tsconfigs.mts",
     "lint:tsconfigs:fix": "tsx ../../scripts/lint-tsconfigs/lint-tsconfigs.mts --fix",
-    "messenger-action-types:check": "tsx ../../packages/messenger-cli/src/cli.ts --esm --check",
-    "messenger-action-types:generate": "tsx ../../packages/messenger-cli/src/cli.ts --esm --generate",
+    "messenger-action-types:check": "tsx ../../packages/messenger-cli/src/cli.ts --formatter oxfmt --esm --check",
```
mcmire, line 44: Are these scripts not enforced via `yarn.config.cjs`?

packages/java-tron-up/package.json
```diff
@@ -15,23 +15,16 @@
     "type": "git",
     "url": "https://github.com/MetaMask/core.git"
   },
-  "bin": "./dist/bin/java-tron-up.mjs",
+  "bin": "./dist/bin/java-tron-up.js",
```
mcmire, line 18: ~Interesting, is this required?~

Never mind, I understand why this change is needed now.

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -1376,6 +1396,7 @@ export class KeyringController<
           try {
             wallet = importers.fromEtherWallet(input, password);
           } catch {
+            // @ts-expect-error: Wallet.fromV3 does not exist?
```
mcmire, line 1399: That's weird. I guess this line still works at runtime?

packages/messenger-cli/jest.config.cjs
```diff
@@ -0,0 +1,42 @@
...
+    global: {
+      branches: 90.65,
+      functions: 100,
+      lines: 97.39,
+      statements: 97.4,
+    },
+  },
+});
+
+// `deepmerge` concatenates arrays, so the CommonJS transform inherited from the
```
mcmire, line 36: Oof. Maybe it was a bad idea to use `deepmerge`...

packages/platform-api-docs/src/cli.test.ts
mcmire: Yeah, that's fine.

packages/smart-transactions-controller/src/SmartTransactionsController.ts
mcmire, line 496: Feels like there ought to be a better way to do this than deep cloning. Oh well...

In any case, why not use `cloneDeep`? Other packages seem to be using this.

scripts/create-package/cli.test.ts
```diff
@@ -1,8 +1,22 @@
-import cli from './cli.js';
-import { commands, commandMap } from './commands.js';
-import * as utils from './utils.js';
-
-jest.mock('./utils');
+import { jest } from '@jest/globals';
+
+// `jest.mock` does not apply to ES modules, so the module registry is stubbed
```
mcmire, line 3: Oof. Seems strange that we have to dynamically import the modules.

Is there something we're doing wrong here? Or is this one of those instances where Vitest provides a much better experience?

jest.config.packages.cjs
```diff
@@ -94,6 +94,11 @@ module.exports = {
       '<rootDir>/../../node_modules/@metamask/$1',
     ],
     '^uuid$': require.resolve('uuid'),
+    // The sources import `lodash-es`, because `lodash` is CommonJS and its
```
mcmire, line 97: Good work figuring this out 👏🏻

jest.config.scripts.cjs
```diff
@@ -42,6 +42,15 @@ module.exports = {
 
   preset: 'ts-jest',
 
+  // Unlike the packages, the scripts use `import.meta`, which has no CommonJS
```
mcmire, line 45: Interesting. I wonder if we can clean this up in the future so we don't have to make these overrides. Or maybe this is just how it has to work from now on until we switch to Vitest...

package.json
```diff
@@ -12,11 +12,12 @@
...
   ],
   "files": [],
+  "type": "module",
   "scripts": {
-    "build": "yarn ts-bridge --project tsconfig.build.json --verbose",
+    "build": "yarn tsc --build tsconfig.build.json --verbose",
     "build:clean": "yarn build:only-clean && yarn build",
     "build:docs": "yarn workspaces foreach --all --no-private --parallel --interlaced --verbose run build:docs",
-    "build:only-clean": "rimraf -g 'packages/*/dist'",
+    "build:only-clean": "rimraf -g 'packages/*/dist' 'packages/*/tsconfig.build.tsbuildinfo'",
```
mcmire, line 20: Thank you, I had always meant to add this but never did 🤦🏻

Mrtenz: It would be useless anyway for `ts-bridge` 😅.

scripts/create-package/cli.test.ts
Mrtenz: This is just how Jest works with ESM unfortunately, and why I've been pushing back on the full Jest to ESM migration. With Vitest we shouldn't need to dynamically import anything.

packages/smart-transactions-controller/src/SmartTransactionsController.ts
Mrtenz: Replaced this with `klona` but then found many other instances of `lodash` after so ended up replacing that with `lodash-es`. Will do the same here.

packages/keyring-controller/src/KeyringController.ts
Mrtenz: It does, not sure why the types are broken. 😕

packages/authenticated-user-storage/package.json
Mrtenz: Apparently not. 😅

packages/account-tree-controller/package.json
Mrtenz: Good point, we can add a `build:clean` script to all packages. If we make it clean by default we don't get the benefits of incremental builds.

jest.config.scripts.cjs
Mrtenz: I think we should just start migrating to Vitest at some point. Cleaning this up otherwise would likely mean migrating all Jest tests to ESM.

packages/account-tree-controller/package.json
mcmire: Yup, that sounds like a good plan.

packages/authenticated-user-storage/package.json
mcmire: A problem for another day, I suppose 😅

jest.config.scripts.cjs
mcmire: Agreed, maybe that is the next big task.

package.json
mcmire: Ah yeah that's fair!

packages/account-tree-controller/package.json
mcmire: Should we document this in the changelog as a breaking change?

packages/accounts-controller/package.json
```diff
@@ -18,19 +18,12 @@
...
     ".": {
-      "import": {
-        "types": "./dist/index.d.mts",
-        "default": "./dist/index.mjs"
-      },
-      "require": {
-        "types": "./dist/index.d.cts",
-        "default": "./dist/index.cjs"
-      }
+      "types": "./dist/index.d.ts",
```
mcmire, line 25: Should we document this in the changelog?

(and so on for other packages)

packages/account-tree-controller/package.json
Mrtenz: Done!

yarn.config.cjs
```diff
@@ -145,21 +148,33 @@ module.exports = defineConfig({
...
+            workspace,
+            'scripts.build:clean',
+            'yarn build:only-clean && yarn build',
+          );
+
+          expectWorkspaceField(
+            workspace,
+            'scripts.build:only-clean',
+            `rimraf './dist' './tsconfig.build.tsbuildinfo'`,
           );
```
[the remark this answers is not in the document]
mcmire: Ugh. We should really just make that template a private package.

Mrtenz:
Adding `no-changelog` because the `cryptography` package doesn't have any changelog entries as it has not been released yet.

Mrtenz:
@SocketSecurity ignore npm/lodash-es@4.18.1

Seems to be intentional.

scripts/create-package/package-template/package.json
FrederikBolding, line 1: Missing `type: 'module'` and other package.json updates 👀

packages/assets-controller/src/utils/normalizeAssetId.ts
```diff
@@ -1,10 +1,12 @@
 import { toChecksumAddress } from '@ethereumjs/util';
 import { parseCaipAssetType, parseCaipChainId } from '@metamask/utils';
-import type { MemoizedFunction } from 'lodash';
-import memoize from 'lodash/memoize';
+import { memoize } from 'lodash-es';
```
FrederikBolding, line 3: Will we still be able to tree shake parts of lodash like this?

packages/chain-agnostic-permission/src/scope/constants.ts
```diff
@@ -54,6 +54,7 @@ export const Eip1193OnlyMethods = [
 /**
  * All MetaMask methods, except for ones we have specified in the constants above.
  */
+// @ts-expect-error: TODO: `api-specs` is not supported in ESM.
```
FrederikBolding: How does this package work then?

packages/keyring-controller/src/KeyringController.ts
```diff
@@ -53,6 +53,26 @@ import type {
...
+ * here so the imports work whichever applies.
+ */
+/* istanbul ignore next: only one branch is reachable per module system */
+const walletModule = (typeof ethereumjsWallet.default === 'object' &&
+ethereumjsWallet.default
+  ? ethereumjsWallet.default
+  : ethereumjsWallet) as unknown as {
+  default: typeof ethereumjsWallet.default;
+  thirdparty: typeof ethereumjsWallet.thirdparty;
+};
```
FrederikBolding, line 71: sigh

packages/messenger-cli/src/parse-source.ts
```diff
@@ -10,7 +10,12 @@ import type {
   SourceFile,
   Type,
 } from 'typescript';
-import {
+import typescript from 'typescript';
+
+// Older versions of TypeScript break when importing directly in ESM. This is a
```
FrederikBolding, line 15: Does it matter if we bump to TS6/TS7 anyways?

packages/multichain-api-middleware/src/middlewares/multichainMethodCallValidatorMiddleware.ts
```diff
@@ -9,8 +9,8 @@ import type {
   OpenrpcDocument,
   ReferenceObject,
 } from '@open-rpc/meta-schema';
-import dereferenceDocument from '@open-rpc/schema-utils-js/build/dereference-document';
-import { makeCustomResolver } from '@open-rpc/schema-utils-js/build/parse-open-rpc-document';
+import { dereferenceDocument } from '@open-rpc/schema-utils-js';
```
FrederikBolding, line 12: Any reason to not keep this as-is and just add `.js`

packages/network-controller/src/create-network-client.ts
```diff
@@ -173,6 +173,7 @@ export function createNetworkClient({
   if (configuration.type === NetworkClientType.Infura) {
     rpcApiMiddleware = asV2Middleware(
       createInfuraMiddleware({
+        // @ts-expect-error: Type mismatch.
```
FrederikBolding, line 178: What is causing this?

packages/wallet-framework-docs/package.json
```diff
@@ -62,6 +63,7 @@
     "raw-loader": "^4.0.2",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
+    "rimraf": "^5.0.5",
```
FrederikBolding: Not actually used here?

packages/assets-controller/src/utils/normalizeAssetId.ts
Mrtenz: Yes, tree shaking support is much better for ESM.

packages/chain-agnostic-permission/src/scope/constants.ts
Mrtenz: Does not seem to work under ESM actually. The default export is now an object containing `MetaMaskOpenRPCDocument`, so to access the methods we'd need `MetaMaskOpenRPCDocument.MetaMaskOpenRPCDocument.methods`. 😕  Jest is still running in CJS mode so didn't catch this issue.

May need a similar fix as `ethereumjs-wallet`.

packages/messenger-cli/src/parse-source.ts
Mrtenz: Haven't tried if this works in TS6 actually, but the TS6 bump is after this PR anyway.

packages/keyring-controller/src/KeyringController.ts
Mrtenz: I didn't find a better solution to this unfortunately. This package just seems to be broken in ESM.

packages/multichain-api-middleware/src/middlewares/multichainMethodCallValidatorMiddleware.ts
Mrtenz: `TS2614: Module '@open-rpc/schema-utils-js/build/dereference-document.js' has no exported member dereferenceDocument.`

packages/network-controller/src/create-network-client.ts
Mrtenz: Long error, but this is the last part. Didn't look into it much further.
```
Type JsonRpcRequest<any> is not assignable to type { method: 'eth_getBlockByNumber'; }
  Types of property method are incompatible.
    Type string is not assignable to type 'eth_getBlockByNumber'
```

packages/messenger-cli/src/parse-source.ts
FrederikBolding: We could consider reverting it in the next stacked PR then

packages/chain-agnostic-permission/src/scope/constants.ts
Mrtenz: Found a better fix that seems to work actually.

packages/multichain-api-middleware/src/middlewares/multichainMethodCallValidatorMiddleware.ts
FrederikBolding: Strange...

packages/network-controller/src/create-network-client.ts
FrederikBolding: Maybe add a note to look into this later post migration?

FrederikBolding (dismissed):

mcmire (dismissed):

HowardBraham (dismissed):

ffmcgee725 (approved):

ffmcgee725 (dismissed):

geositta (dismissed):

ulissesferreira (dismissed):

packages/assets-controllers/src/selectors/token-selectors.ts
Prithpal-Sooriya: Interesting. Not a dealbreaker for assets team, this whole package is actively being deprecated.

Prithpal-Sooriya (dismissed):

zone-live (dismissed):

jeffsmale90 (dismissed):

adonesky1 (dismissed):

roz0n (dismissed):

lwin-kyaw (dismissed):

abretonc7s (dismissed):

baptiste-marchand (dismissed):
LGTM for engagement files

packages/keyring-controller/src/KeyringController.ts
ccharly: Nit: Then, I wonder if we can keep the `@ts-expect-error` and just add another comment saying it's still ok runtime-wise?

I mean, it's not new, only the directive is, but to someone just reading the code, it's a bit confusing 😅

ccharly (dismissed):
LGTM for accounts (left 1 nit, but the changes are still ok)

zelkibuilds (dismissed):

chakra-guy (dismissed):

dovydas55 (dismissed):

n3ps (dismissed):

Akaryatrh (dismissed):

tommasini (dismissed):

GeorgeGkas (dismissed):

OGPoyraz (dismissed):

rarquevaux (approved):

rarquevaux (dismissed):

eslint-suppressions.json
```diff
@@ -500,7 +500,7 @@
       "count": 25
     },
     "@typescript-eslint/prefer-nullish-coalescing": {
-      "count": 1
+      "count": 2
```
mcmire, line 503: :( Oh well, not a big increase.

mcmire (dismissed):
LGTM!

mcmire (approved):

PR #9540, opened 2026-07-16
feat(base-data-service): Add struct option for validation

FrederikBolding:
## Explanation

Since most data services use `fetch` and `response.json()`, they can end up returning `any` without much type safety or runtime validation assuming they don't deal with this inside the data service themselves. This PR aims to improve that by introducing an optional `struct` field, when this is defined the response is validated using Superstruct (and optionally coerced) before being returned back to the caller. Additionally `TData` (and the return value) is adjusted to match the inferred type from the struct.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1039

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches the shared query execution path for all data services; adopters that pass `struct` will start failing queries at runtime when APIs return unexpected shapes.
> 
> **Overview**
> Adds an optional **`struct`** argument to **`fetchQuery`** and **`fetchInfiniteQuery`** so subclasses can validate (and coerce) JSON from **`queryFn`** via **`@metamask/superstruct`** before results are cached or returned. When **`struct`** is set, **`TData`** / the method return type is inferred from the struct instead of the raw query function type.
> 
> A new **`processQueryResponse`** helper runs after the service policy executes **`queryFn`**; invalid payloads throw an error that includes the query key prefix (e.g. **`ExampleDataService:getAssets returned an invalid response:`**). **`ExampleDataService.getAssets`** is updated to pass **`GetAssetsResponseStruct`**, and tests cover a failed validation path.
> 
> **`@metamask/superstruct`** is added as a runtime dependency; the changelog documents the feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51ea322aa68f3b291e8cdff2e96b694fdf5290a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING:** Bump `@tanstack/query-core` from `^4.43.0` to `^5.62.16` ([#9712](https://github.com/MetaMask/core/pull/9712))
   - The option types accepted by `fetchQuery`, `fetchInfiniteQuery`, and `invalidateQueries` now follow the query-core v5 API. Subclasses need to rename `cacheTime` to `gcTime`.
   - `fetchInfiniteQuery` now requires `initialPageParam` and `getNextPageParam`, matching query-core's options for infinite queries. This also lets `TPageParam` be inferred instead of spelled out in the type parameters.
+- **BREAKING:** `fetchQuery` and `fetchInfiniteQuery` now take one more type parameter: `TDataStruct` is the third parameter and the others have been shifted up ([#9540](https://github.com/MetaMask/core/pull/9540))
 - Bump `@metamask/utils` from `^11.9.0` to `^11.11.0` ([#9074](https://github.com/MetaMask/core/pull/9074))
 - Bump `@metamask/controller-utils` from `^12.1.0` to `^12.3.0` ([#9058](https://github.com/MetaMask/core/pull/9058), [#9083](https://github.com/MetaMask/core/pull/9083), [#9218](https://github.com/MetaMask/core/pull/9218))
 - Bump `@metamask/messenger` from `^1.2.0` to `^2.0.0` ([#9392](https://github.com/MetaMask/core/pull/9392))
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `handleWhen`
 - Export types `DataServiceActions` and `DataServiceEvents` ([#9475](https://github.com/MetaMask/core/pull/9475))
 - Add `struct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))
-  - When provided, the struct validates the response and its type is used to infer the return type of the query method
+  - When provided, the struct is used to validate the response and for inferring the return type of the query
 
 ### Changed
 
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `handleAll`
     - `handleWhen`
 - Export types `DataServiceActions` and `DataServiceEvents` ([#9475](https://github.com/MetaMask/core/pull/9475))
-- Add `struct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))
+- Add `responseStruct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))
   - When provided, the struct is used to validate the response and for inferring the return type of the query
 
 ### Changed
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `handleAll`
     - `handleWhen`
 - Export types `DataServiceActions` and `DataServiceEvents` ([#9475](https://github.com/MetaMask/core/pull/9475))
+- Add `struct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))
+  - When provided, the struct validates the response and its type is used to infer the return type of the query method
 
 ### Changed
 
--- a/packages/base-data-service/src/BaseDataService.test.ts
+++ b/packages/base-data-service/src/BaseDataService.test.ts
@@ -222,6 +222,35 @@ describe('BaseDataService', () => {
     expect(publishSpy).toHaveBeenCalledTimes(8);
   });
 
+  describe('validation', () => {
+    beforeAll(() => {
+      jest.useRealTimers();
+    });
+
+    afterAll(() => {
+      jest.useFakeTimers({ doNotFake: ['nextTick', 'setImmediate'] });
+    });
+
+    beforeEach(() => {
+      cleanAll();
+    });
+
+    it('throws when fetchQuery response fails struct validation', async () => {
+      const messenger = new Messenger({ namespace: serviceName });
+      const service = new ExampleDataService(messenger);
+
+      mockAssets({ status: 200, body: { foo: 'bar' } });
+      mockAssets({ status: 200, body: { foo: 'bar' } });
+      mockAssets({ status: 200, body: { foo: 'bar' } });
+
+      await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
+        /ExampleDataService:getAssets.*returned an invalid response:/u,
+      );
+
+      service.destroy();
+    });
+  });
+
   describe('service policy', () => {
     beforeAll(() => {
       jest.useRealTimers();
--- a/packages/base-data-service/src/BaseDataService.test.ts
+++ b/packages/base-data-service/src/BaseDataService.test.ts
@@ -245,7 +245,7 @@ describe('BaseDataService', () => {
       mockAssets({ status: 200, body: { foo: 'bar' } });
 
       await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
-        'ExampleDataService:getAssets returned an invalid response: Expected an array value, but received: [object Object].',
+        'Query function for "ExampleDataService:getAssets" returned an unexpected response: Expected an array value, but received: [object Object].',
       );
 
       service.destroy();
--- a/packages/base-data-service/src/BaseDataService.test.ts
+++ b/packages/base-data-service/src/BaseDataService.test.ts
@@ -245,7 +245,7 @@ describe('BaseDataService', () => {
       mockAssets({ status: 200, body: { foo: 'bar' } });
 
       await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
-        /ExampleDataService:getAssets.*returned an invalid response:/u,
+        'ExampleDataService:getAssets returned an invalid response: Expected an array value, but received: [object Object].',
       );
 
       service.destroy();
--- a/packages/base-data-service/src/BaseDataService.test.ts
+++ b/packages/base-data-service/src/BaseDataService.test.ts
@@ -225,6 +225,35 @@ describe('BaseDataService', () => {
     expect(publishSpy).toHaveBeenCalledTimes(8);
   });
 
+  describe('validation', () => {
+    beforeAll(() => {
+      jest.useRealTimers();
+    });
+
+    afterAll(() => {
+      jest.useFakeTimers({ doNotFake: ['nextTick', 'setImmediate'] });
+    });
+
+    beforeEach(() => {
+      cleanAll();
+    });
+
+    it('throws when fetchQuery response fails struct validation', async () => {
+      const messenger = new Messenger({ namespace: serviceName });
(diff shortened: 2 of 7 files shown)


packages/base-data-service/src/BaseDataService.ts
```diff
@@ -250,26 +252,38 @@ export class BaseDataService<
...
+    'queryKey'
+  > & {
+    queryFn: QueryFunction<TQueryFnData, TQueryKey>;
+    struct?: TStruct;
+  }): Promise<TData> {
     return this.#queryClient.fetchQuery({
       ...options,
       queryFn: (context) =>
-        this.#policy.execute(() => options.queryFn(context)),
+        // TODO: Consider if the validation should happen outside the policy executor
```
FrederikBolding: Open to opinions on this, unsure if we want to treat validation errors as needing a retry

Mrtenz: Does `queryFn` handle request errors separately? If so, how likely is it that we get a non-error response (e.g., HTTP 200) that does not pass validation which passes validation on retry?

FrederikBolding: > Does queryFn handle request errors separately?

Most data services follow the pattern of checking `response.ok` as the example does: https://github.com/MetaMask/core/blob/fb/base-data-service-structs/packages/base-data-service/tests/ExampleDataService.ts#L114-L116

> If so, how likely is it that we get a non-error response (e.g., HTTP 200) that does not pass validation which passes validation on retry?

That is my question as well. My intuition is that this would very rarely happen and mostly during development as the struct not matching the API response would be unexpected and a bug to fix before shipping 🤔

Mrtenz: Agreed, I don't think we need retries for validation failures.

FrederikBolding: Moved https://github.com/MetaMask/core/pull/9540/commits/51ea322aa68f3b291e8cdff2e96b694fdf5290a7

Mrtenz (dismissed):

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -250,26 +252,38 @@ export class BaseDataService<
...
+    ...options
+  }: WithRequired<
+    OmitKeyof<
+      FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+      'retry' | 'retryDelay' | 'queryFn'
+    >,
+    'queryKey'
+  > & {
+    queryFn: QueryFunction<TQueryFnData, TQueryKey>;
+    struct?: TStruct;
```
mcmire: What do you think about calling this `expectedResponseStruct` just to be clear? Otherwise people might interpret this as validating the request params, or something. 

``` suggestion
    expectedResponseStruct,
    ...options
  }: WithRequired<
    OmitKeyof<
      FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
      'retry' | 'retryDelay' | 'queryFn'
    >,
    'queryKey'
  > & {
    queryFn: QueryFunction<TQueryFnData, TQueryKey>;
    expectedResponseStruct?: TStruct;

File: packages/base-data-service/src/BaseDataService.ts [hunk 3]
```diff
@@ -278,17 +292,24 @@ export class BaseDataService<
...
+    TStruct extends Struct<TQueryFnData> | undefined = undefined,
+    TData extends TQueryFnData = TStruct extends Struct<infer StructType>
+      ? StructType
+      : TQueryFnData,
     TQueryKey extends QueryKey = QueryKey,
     TPageParam extends Json = Json,
   >(
-    options: WithRequired<
+    {
+      struct,
```
mcmire: Same as above.

packages/base-data-service/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+): Response {
+  if (!struct) {
+    return response;
+  }
+
+  const [error, result] = validate(response, struct);
+
+  if (error) {
+    throw new Error(
+      `${queryKey[0]} returned an invalid response: ${error.message}.`,
```
mcmire: Is the error here that the action returned an invalid response? Ideally we would say that the URL returned an unexpected response, but we don't have access to that here, so perhaps just

``` suggestion
      `The request for ${queryKey[0]} returned an unexpected response: ${error.message}.`,

File: packages/base-data-service/src/BaseDataService.ts [hunk 5]
```diff
@@ -250,26 +252,38 @@ export class BaseDataService<
...
    */
   protected async fetchQuery<
     TQueryFnData extends Json,
     TError = DefaultError,
-    TData = TQueryFnData,
+    TStruct extends Struct<TQueryFnData> | undefined = undefined,
+    TData = TStruct extends Struct<infer StructType>
+      ? StructType
+      : TQueryFnData,
     TQueryKey extends QueryKey = QueryKey,
```
mcmire, line 265: To prevent a breaking change, would it make more sense to have `TData` remain the same as `TQueryFnData` and then define `TQueryFnData` in terms of `TStruct`?

``` suggestion
    TQueryFnData extends Json,
    TError = DefaultError,
    TData = TQueryFnData,
    TQueryKey extends QueryKey = QueryKey,
    TStruct extends Struct<TQueryFnData> | undefined = undefined,
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -278,17 +292,24 @@ export class BaseDataService<
...
   protected async fetchInfiniteQuery<
     TQueryFnData extends Json,
     TError = DefaultError,
-    TData extends TQueryFnData = TQueryFnData,
+    TStruct extends Struct<TQueryFnData> | undefined = undefined,
+    TData extends TQueryFnData = TStruct extends Struct<infer StructType>
+      ? StructType
+      : TQueryFnData,
     TQueryKey extends QueryKey = QueryKey,
     TPageParam extends Json = Json,
```
mcmire, line 307: Similar as above:

``` suggestion
    TQueryFnData extends Json,
    TError = DefaultError,
    TData extends TQueryFnData = TQueryFnData,
    TQueryKey extends QueryKey = QueryKey,
    TPageParam extends Json = Json,
    TStruct extends Struct<TQueryFnData> | undefined = undefined,
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -303,6 +324,7 @@ export class BaseDataService<
     > &
       InfiniteQueryPageParamsOptions<TQueryFnData, TPageParam> & {
         queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam>;
+        struct?: TStruct;
```
mcmire: Same as above, what about `expectedResponseStruct`?

``` suggestion
      expectedResponseStruct,
      ...options
    }: WithRequired<
      OmitKeyof<
        FetchQueryOptions<
          TQueryFnData,
          TError,
          InfiniteData<TData, TPageParam>,
          TQueryKey,
          TPageParam
        >,
        'retry' | 'retryDelay' | 'queryFn' | 'initialPageParam'
      >,
      'queryKey'
    > &
      InfiniteQueryPageParamsOptions<TQueryFnData, TPageParam> & {
        queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam>;
        expectedResponseStruct?: TStruct;
```

packages/base-data-service/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    );
```
mcmire, line 28: Is it worth including all of the struct validation errors? Also should we clarify that it's the query function that returned an invalid response, and that the response is "unexpected" rather than "invalid"?

``` suggestion
    const failureMessages = error
      .failures()
      .map((failure) =>
        failure.key
          ? `- At \`${failure.key}\`: ${failure.message}`
          : `- At root: ${failure.message}`,
      );
    throw new Error(
      `Query function for ${queryKey[0]} returned an unexpected response:\n${failureMessages.join('\n')}`,
    );
```

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -225,6 +225,33 @@ describe('BaseDataService', () => {
...
+
+    it('throws when fetchQuery response fails struct validation', async () => {
+      const messenger = new Messenger({ namespace: serviceName });
+      const service = new ExampleDataService(messenger);
+
+      mockAssets({ status: 200, body: { foo: 'bar' } });
+
+      await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
+        /ExampleDataService:getAssets.*returned an invalid response:/u,
+      );
```
mcmire, line 249: Nit: Is it worth testing the whole error message so we can make sure we know what it would say in practice?

``` suggestion
      mockAssets({ status: 200, body: { foo: 'bar' } });

      await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
        'ExampleDataService:getAssets returned an invalid response: Expected an array value, but received: [object Object].',
      );
```

Alternatively, is it worth having the response be a value such that multiple validation errors would be produced? Coupled with the suggestion I made below you could then test for it like so:

``` suggestion
      mockAssets({
        status: 200,
        body: [
          {
            assetId:
              'eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f',
            decimals: 18,
            name: 'Dai Stablecoin',
            XXXsymbol: 'DAI',
          },
        ],
      });

      await expect(service.getAssets(MOCK_ASSETS)).rejects.toThrow(
        `Query function for ExampleDataService:getAssets returned an unexpected response:
- At \`symbol\`: Expected a string, but received: undefined
- At \`XXXsymbol\`: Expected a value of type \`never\`, but received: \`"DAI"\``,
      );
```

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Unfortunately not, this is what I first tried, but `TQueryFnData` is most often inferred as the result of `response.json()` which is `any`. When that happens we don't get any proper type inference from the struct itself.

packages/base-data-service/src/utils.ts
FrederikBolding: Sure! I updated error message: https://github.com/MetaMask/core/pull/9540/commits/6327eeb30136ab7f332c5b8eaa179bb4ddee0f39

`.failures()` will iterate and validate every single key recursively whereas `validate` just returns the first validation error that occurred. Unsure we want to pay the performance penalty of listing out every single wrong field if the blob is considered invalid in general? 🤔

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: https://github.com/MetaMask/core/pull/9540#discussion_r3820872632

packages/base-data-service/src/BaseDataService.ts
mcmire: Ah, right...

I guess this has to be a breaking change then 😓

packages/base-data-service/src/utils.ts
mcmire: Sure, makes sense. Including *any* details about the validation error is better what we have now in any case.

packages/base-data-service/src/BaseDataService.ts
mcmire: (I forgot we've already introduced breaking changes so this is okay)

packages/base-data-service/CHANGELOG.md
mcmire, line 47: Should we mention that the type parameters changed on `fetchQuery` and `fetchInfiniteQuery`?

``` suggestion
- **BREAKING:** `fetchQuery` and `fetchInfiniteQuery` now take one more type parameter: `TDataStruct` is the third parameter and the others have been shifted up ([#9540](https://github.com/MetaMask/core/pull/9540))
- Bump `@metamask/utils` from `^11.9.0` to `^11.11.0` ([#9074](https://github.com/MetaMask/core/pull/9074))
```

You could also include it as a sub-bullet point above if you like.

packages/base-data-service/src/BaseDataService.ts
FrederikBolding: Yeah, shouldn't be a big deal! There was no required changes in the monorepo either!

packages/base-data-service/CHANGELOG.md
mcmire, line 50: Definitely minor but perhaps worth mentioning that we've added a dependency:

``` suggestion
- Add dependency `superstruct` (`^3.4.1`) ([#9418](https://github.com/MetaMask/core/pull/9418))

**Review by mcmire — COMMENTED** [review]
A couple more things and I'm good.

File: packages/base-data-service/CHANGELOG.md
Comment (FrederikBolding) [reply packages/base-data-service/CHANGELOG.md:50]: Do we usually document added dependencies? 🤔

File: packages/base-data-service/CHANGELOG.md [hunk 10]
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `handleAll`
     - `handleWhen`
 - Export types `DataServiceActions` and `DataServiceEvents` ([#9475](https://github.com/MetaMask/core/pull/9475))
+- Add `struct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))
```
mcmire: Noticed this as well:

``` suggestion
- Add `responseStruct` option to `fetchQuery` and `fetchInfiniteQuery` for validating query responses using Superstruct ([#9540](https://github.com/MetaMask/core/pull/9540))

FrederikBolding: Ah, thanks!

packages/base-data-service/CHANGELOG.md
mcmire: We document when they're bumped, so I've always considered documenting them when they're added as an extension of that. I also know that we've also never totally agreed on that to begin with 😅

mcmire: I'll treat this as a nitpick. It's up to you.

mcmire (approved):
LGTM.

PR #9543, opened 2026-07-16
feat: optimize Transaction Pay amount quote latency

pedronfigueiredo:
### Context

Updating the amount of a Money Account deposit currently fans one logical user action out through multiple nested transaction updates, parent calldata rebuilds, controller publications, gas estimations, and quote generations. In the reference iOS simulator trace, Update/Done to an executable quote took **8.503 seconds**:

| Segment | Baseline |
| --- | ---: |
| Update/Done → Relay request | 3,393.546 ms |
| Relay request | 4,107.940 ms |
| Relay response → executable quote | 1,001.045 ms |
| **Total** | **8,502.531 ms** |

Relay is a third-party request whose latency varies independently. This proof of concept targets client-controlled work before and after that request.

### What this PR changes

This PR adds two generic Transaction Controller primitives while preserving Transaction Pay's existing reactive architecture. The final Core diff is limited to `packages/transaction-controller`; Money Account-specific orchestration lives in the companion Mobile PR.

#### TransactionController

- Adds `updateTransactionMetadata`, a synchronous messenger action that atomically mutates the latest transaction metadata, with an optional `skipResimulate` flag that defaults to normal automatic re-simulation.
  - Callers opt out explicitly with `skipResimulate: true`; normal callers retain automatic re-simulation by default.
  - Normal transaction update APIs retain their existing re-simulation behavior.
- Exports `updateEIP7702BatchData`, a pure synchronous utility that validates indexed nested-transaction calldata updates, returns a new nested-transaction list, and rebuilds parent EIP-7702 batch calldata once.
- Keeps `updateAtomicBatchData` as the existing asynchronous API for consumers that require gas estimation, while synchronously clearing stale gas, gas-limit, gas-used, simulation, security-alert, and gas-revert metadata when new batch calldata is published.

#### Transaction Pay architecture

Transaction Pay continues to use its existing state-driven quote lifecycle:

1. The client asynchronously prepares transaction-specific calldata.
2. The client commits required assets, all nested calldata, and parent EIP-7702 calldata in one `updateTransactionMetadata` call.
3. `TransactionPayController` observes the Transaction Controller state change.
4. It derives required payment tokens from the updated transaction.
5. The existing latest-wins quote path starts one quote request and aborts any older quote request for the same transaction.

The earlier proof-of-concept-specific `TransactionPayController.updateAmount` action, constructor preparation callback, explicit quote invocation, quote suppression set, and duplicate quote-cancellation API have been removed. This leaves transaction metadata as the integration boundary and the existing listener as the single quote trigger.

### Flow

```text
Before
======
Update/Done
    |
    +--> update approval call --> rebuild parent --> publish
    |                                      `--> quote generation A (superseded)
    |
    `--> update deposit call  --> rebuild parent --> publish
                                           `--> quote generation B --> Relay

After
=====
Update/Done
    |
    `--> client prepares complete approval + deposit calldata
             |
             `--> one synchronous metadata update
                     - required assets
                     - all nested calldata
                     - parent EIP-7702 calldata
                     - clear stale derived metadata
                     - skip automatic re-simulation
                          |
                          `--> existing state listener
                                   |
                                   `--> one latest-wins quote --> Relay
```

### Mobile first adopter

The companion Mobile PR owns the Money Account-specific asynchronous preparation because it has the required vault configuration, provider access, and transaction builders.

Its coordinator:

- shares identical in-flight amount intents;
- prevents superseded asynchronous preparations from committing stale results;
- validates the approval/deposit template again against current metadata at commit time;
- derives `requiredAssets` from current metadata rather than replacing it from a stale pre-request snapshot;
- applies both indexed calldata updates in one `updateEIP7702BatchData` call;
- clears stale gas, simulation, security-alert, and simulation-revert metadata;
- relies on the existing Transaction Pay listener to launch the quote.

The optimized path remains feature-gated, and unsupported Money Account intents retain the legacy path.

### Configurable automatic re-simulation

Changing `txParams.data` through the normal transaction update path triggers an asynchronous re-simulation check. For this MetaMask Pay flow, calldata is produced by controlled builders using known providers, and re-simulation is not required to construct the vendor quote. `updateTransactionMetadata` lets controlled callers make this performance and security policy explicit while leaving all other update paths unchanged.

### Benchmark result

A follow-up React Native DevTools trace used the same inferred Update/Done and executable-commit anchors:

| Segment | Baseline | PoC | Change |
| --- | ---: | ---: | ---: |
| Update/Done → Relay request | 3,393.546 ms | **1,512.891 ms** | **-55.4%** |
| Relay request | 4,107.940 ms | 5,118.617 ms | +24.6% |
| Relay response → executable quote | 1,001.045 ms | **805.860 ms** | **-19.5%** |
| **Observed total** | **8,502.531 ms** | **7,437.368 ms** | **-12.5%** |

The Relay request regressed by 1.011 seconds in the second sample even though this PR does not optimize Relay. Treating Relay as the same 4.108-second contribution in both traces gives a normalized PoC total of **6,426.691 ms**, a **2,075.840 ms / 24.4% end-to-end improvement**. Client-controlled time fell from **4,394.591 ms to 2,318.751 ms**, a **47.2% reduction**.

Structural trace evidence:

- 29 → 19 requests in the measured window
- 18 → 11 requests before Relay
- two → one `eth_estimateGas` calls
- 1,640 ms → 483 ms of pre-request React scheduler time
- 1,091 ms → 313 ms of pre-request render time

These are two simulator traces, not production p50/p95 measurements. Relay latency should continue to be reported separately, and old/new runs should be alternated over multiple samples. The trace predates the final review-driven architecture, so it is directional evidence rather than a benchmark of the exact final code.

### Correctness and rollout constraints

- Indexed nested updates are validated before transaction state is published.
- Required assets, nested calldata, and parent calldata are committed together or not at all.
- The EIP-7702 utility is applied against current transaction metadata.
- Superseded Mobile preparations cannot commit stale transaction data.
- Transaction Pay's existing state listener remains the only quote trigger.
- Existing quote generation remains latest-wins per transaction.
- Normal transaction update flows continue to use automatic re-simulation.
- The Mobile integration is feature-gated and retains the legacy path.
- Security/privacy review is still required before rollout to confirm whether Sentinel gates execution only or must also gate vendor disclosure.

### Validation

Core:

- Focused Transaction Controller tests pass with coverage disabled; Transaction Pay listener and quote integration tests also pass.
- `yarn build` passes for the complete monorepo.
- Transaction Controller messenger action-type checks pass.
- Transaction Controller changelog validation passes.
- Targeted ESLint and `git diff --check` pass.
- `yarn lint:misc:check` passes, including the final EIP-7702 utility formatting fix.
- `yarn workspace @metamask/transaction-controller run test --runTestsByPath src/TransactionController.test.ts --coverage=false` passes.

Mobile companion:

- 33 focused tests pass for amount coordination, controller initialization, and the confirmation hook.
- Targeted ESLint, Prettier, messenger action-type checks, and `git diff --check` pass.
- TypeScript validation passes when checked against the locally built Transaction Controller from this PR; the Mobile PR must consume the released package version before merge.

## References

- Mobile first-adopter PR: [MetaMask/metamask-mobile#33433](https://github.com/MetaMask/metamask-mobile/pull/33433)
- Baseline trace SHA-256: `28e517a73d33db3aa63470f9bc55bdbb47fc61060b09227dce3c5df2d4ae13a9`
- PoC trace SHA-256: `e80c86da8913924a228627d851ab75965174a30c7607cee1416917aa2c34f`

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core transaction state, gas estimation, and re-simulation policy; incorrect metadata clearing or `skipResimulate` misuse could affect quotes and safety checks for EIP-7702 batch flows.
> 
> **Overview**
> Adds **`updateTransactionMetadata`** and the messenger action **`TransactionController:updateTransactionMetadata`**, so callers can mutate the latest transaction metadata in one synchronous update via a callback, with optional **`skipResimulate`** to bypass automatic re-simulation when calldata changes.
> 
> Exports pure **`updateEIP7702BatchData`** (and **`NestedTransactionUpdate`**) to apply indexed nested calldata updates, rebuild parent EIP-7702 batch data without mutating the input, and validate duplicate or missing indices.
> 
> **`updateAtomicBatchData`** now uses that helper, validates the transaction exists up front, clears stale gas, simulation, security-alert, and gas-revert metadata when batch calldata changes (while preserving receipt/simulation reverts where applicable), and returns the regenerated batch calldata.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1344bed4ac65506b6ab59e55a8913e171419a900. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2552,6 +2556,119 @@ export class TransactionController extends BaseController<
...
+          `Nested transaction not found with index - ${transactionIndex}`,
+        );
+      }
+
+      updateIndexes.add(transactionIndex);
+    }
+
+    log('Beginning atomic batch update', request);
+
+    const updatedTransactionMeta = this.#updateTransactionInternal(
```
matthewwalsh0, line 2624: Is this duplicating lots of the `updateAtomicBatchData` logic?

Should we export a synchronous util from `eip7702` such as `updateEIP7702BatchData` that accepts existing data and multiple new calls with an index to encapsulate?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2552,6 +2556,119 @@ export class TransactionController extends BaseController<
...
+          nestedTransaction.data = transactionData;
+        }
+
+        const batchTransaction = generateEIP7702BatchTransaction(
+          from,
+          nestedTransactions ?? [],
+        );
+
+        revision = (transactionMeta.transactionRevision ?? 0) + 1;
+        transactionMeta.requiredAssets = requiredAssets;
```
matthewwalsh0: I'm nervous to couple lots of updates here as it's very context specific.

Something i've wanted to do for a long time is expose a new `updateTransactionCallback` action that essentially exposes `updateTransactionInternal` so callers can update multiple properties at once without passing the full metadata, so avoids race conditions.

Then maybe some of this could be client logic to update multiple values at once?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2576,46 +2693,50 @@ export class TransactionController extends BaseController<
...
-    });
+    return transaction.txParams.data as Hex;
+  }
 
-    await this.#updateGasEstimate(draftTransaction);
+  async #prepareAtomicBatchUpdate(
+    draftTransaction: TransactionMeta,
+    revision: number,
+  ): Promise<AtomicBatchPreparationResult> {
+    await Promise.all([
```
matthewwalsh0: Do we even need either of these, could our update be entirely synchronous?

The target gas shouldn't matter as the quote dictates the source gas and the provider owns the target gas?

Similarly, the target simulation isn't used or rendered in MetaMask Pay?

For any non-MetaMask Pay scenarios, we could let the caller explicitly update gas and pass it to our new `updateTransactionCallback`? To avoid adding that overhead implicitly?

I've been meaning to go one step further and disable simulation and gas polling for all MetaMask Pay transactions too. That may not help performance, but would minimise unnecessary network overhead.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2552,6 +2556,119 @@ export class TransactionController extends BaseController<
...
   }
 
+  /**
+   * Atomically updates all amount-dependent data in an atomic batch and starts
+   * revision-bound local preparation.
+   *
+   * @param request - Complete atomic batch update.
+   * @returns The synchronously published revision and its preparation handle.
+   */
+  beginAtomicBatchUpdate(
```
matthewwalsh0: Following my other comment about a generic update method that accepts a callback, do we want to encourage single updates, or should we expose the synchronous `updateEIP7702BatchData` util I mentioned below, so callers could update data themselves, then pass to `updateTransactionCallback`?

So total flexibility to empower the caller to update efficiently in a single call, but generically so not limited to single flows?

packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -182,6 +205,152 @@ export class TransactionPayController extends BaseController<
...
+  /**
+   * Prepares and atomically commits an exact transaction amount, then launches
+   * one quote generation joined to revision-bound local preparation.
+   * Identical in-flight intents share the same promise; different intents
+   * supersede and abort earlier work.
+   *
+   * @param request - Exact amount and transaction ID.
+   * @returns Whether the matching quote generation was published.
+   */
+  updateAmount(request: UpdateAmountRequest): Promise<boolean> {
```
matthewwalsh0: This is a cool mechanism to cancel wasted requests, but as per my other comment, do we even care about gas and simulation so is there anything async we need at all?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2552,6 +2556,119 @@ export class TransactionController extends BaseController<
...
+        ...transaction.txParams,
+        // Clear existing gas to force estimation.
+        gas: undefined,
+      },
+    });
+
+    return {
+      revision,
+      transaction,
+      preparation: this.#prepareAtomicBatchUpdate(draftTransaction, revision),
```
matthewwalsh0: This is clever, but we do similar for transaction lifecycle, returning promise properties, and it introduces some complexity around dangling promises, plus I fear it adds confusion to the interface.

So ideally, would it be safer to stick with explicit synchronous actions and full promise returns?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2576,46 +2693,50 @@ export class TransactionController extends BaseController<
...
+  async #prepareAtomicBatchUpdate(
+    draftTransaction: TransactionMeta,
+    revision: number,
+  ): Promise<AtomicBatchPreparationResult> {
+    await Promise.all([
+      this.#updateGasEstimate(draftTransaction),
+      this.#isSimulationEnabled()
+        ? this.#updateSimulationData(draftTransaction)
+        : Promise.resolve(),
+    ]);
```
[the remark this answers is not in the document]
pedronfigueiredo: Addressed in `670fa768e` ([commit](https://github.com/MetaMask/core/commit/670fa768ecf0f0a78deb67ffc50f049357f816d5)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Good suggestion. I’ll extract a synchronous, exported `updateEIP7702BatchData` utility that validates indexed updates and returns the updated nested transactions plus encoded batch data, then use it from the controller.

Addressed in `a5d05c5d2` ([commit](https://github.com/MetaMask/core/commit/a5d05c5d21e5e119ecd15e8ead05b9cdcd04a840)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed. I’ll add a generic `updateTransactionCallback` messenger action that performs a callback-based atomic update with the controller’s normal validation and re-simulation behavior. I’ll keep this change to the reusable primitive and move the flow-specific updates in the follow-up changes.

Addressed in `836ecc496` ([commit](https://github.com/MetaMask/core/commit/836ecc49657b61c84b78525f4b7dbca87eaadae7)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed. I’ll make the amount commit synchronous in Transaction Pay using `updateEIP7702BatchData` and `updateTransactionCallback`, remove the revision-bound gas/simulation preparation from the quote path, and leave explicit gas estimation to the existing non-Pay update API. I’ll keep the broader global simulation and gas-polling disablement separate.

Addressed in `411a67f55` ([commit](https://github.com/MetaMask/core/commit/411a67f5560f13f3cf528598167009f0e4f650e8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Addressed in `411a67f55` ([commit](https://github.com/MetaMask/core/commit/411a67f5560f13f3cf528598167009f0e4f650e8)).

packages/transaction-pay-controller/src/TransactionPayController.ts
pedronfigueiredo: Gas and simulation are no longer part of this path. The cancellation remains because amount preparation and vendor quote requests are still asynchronous, so it prevents stale intent and quote results from being published.

Addressed in `411a67f55` ([commit](https://github.com/MetaMask/core/commit/411a67f5560f13f3cf528598167009f0e4f650e8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed. The preparation promise property and `beginAtomicBatchUpdate` action have been removed. Transaction updates now use the explicit synchronous `updateTransactionCallback` action, while `updateAmount` returns the full quote-pipeline promise.

Addressed in `411a67f55` ([commit](https://github.com/MetaMask/core/commit/411a67f5560f13f3cf528598167009f0e4f650e8)).

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2576,63 +2591,53 @@ export class TransactionController extends BaseController<
...
+      transactionMeta.gasLimitNoBuffer = draftTransaction.gasLimitNoBuffer;
+
+      const draftGasRevert = draftTransaction.revert?.gas;
+      if (draftGasRevert) {
+        transactionMeta.revert = {
+          ...transactionMeta.revert,
+          gas: draftGasRevert,
+        };
+      }
+    });
```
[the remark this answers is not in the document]
pedronfigueiredo: Addressed in `366f7b73d` ([commit](https://github.com/MetaMask/core/commit/366f7b73d3215a8e065f1f841ce8ee82f47cd235)).

pedronfigueiredo:
I’ll clear the stale gas, simulation, security-alert, and revert metadata when committing regenerated batch data from updateAmount, and added coverage for the reset behavior.

Addressed in `90e4879d8` ([commit](https://github.com/MetaMask/core/commit/90e4879d8c3f112bab5ae8b66fd91756fd7bdced)).

packages/transaction-controller/CHANGELOG.md
```diff
@@ -25,11 +30,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [69.1.0]
 
+### Added
```
matthewwalsh0: Bad changelog in older release.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1622,6 +1623,20 @@ export class TransactionController extends BaseController<
...
+  /**
+   * Updates an existing transaction using a callback.
+   *
+   * @param transactionId - ID of the transaction to update.
+   * @param callback - Function that updates the transaction metadata.
+   * @returns The updated transaction metadata.
+   */
+  updateTransactionCallback(
+    transactionId: string,
+    callback: (transactionMeta: TransactionMeta) => TransactionMeta | void,
```
matthewwalsh0: If we ever add more optionality here, should we make the entire thing an object?

Also should we scrap the return value on the callback to discourage callers from changing the entire thing and instead just mutating values?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2580,63 +2595,70 @@ export class TransactionController extends BaseController<
...
     const updatedTransactionMeta = this.#updateTransactionInternal(
-      {
-        transactionId,
-      },
+      { transactionId },
       (transactionMeta) => {
-        const { nestedTransactions, txParams } = transactionMeta;
-        const from = txParams.from as Hex;
-        const nestedTransaction = nestedTransactions?.[transactionIndex];
+        transactionMeta.nestedTransactions = nestedTransactions;
```
matthewwalsh0, line 2627: Why are we doing all this here? Shouldn't this be identical but now it generates the nested transactions and data from our new `updateEIP7702BatchData` util?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1622,6 +1623,20 @@ export class TransactionController extends BaseController<
...
   }
 
+  /**
+   * Updates an existing transaction using a callback.
+   *
+   * @param transactionId - ID of the transaction to update.
+   * @param callback - Function that updates the transaction metadata.
+   * @returns The updated transaction metadata.
+   */
+  updateTransactionCallback(
```
matthewwalsh0: Minor, I've always been torn on the name, as `updateTransaction` is taken, but `Callback` feels too explicit.

Maybe `updateTransactionMetadata` is more accurate?

packages/transaction-controller/src/utils/eip7702.ts
```diff
@@ -164,6 +165,59 @@ export async function isAccountUpgradedToEIP7702(
...
+/**
+ * Update indexed transactions in an EIP-7702 batch and regenerate its calldata.
+ *
+ * @param from - The sender address.
+ * @param transactions - The existing nested transactions.
+ * @param updates - Indexed calldata updates.
+ * @returns Updated nested transactions and regenerated batch calldata.
+ */
+export function updateEIP7702BatchData(
+  from: Hex,
```
matthewwalsh0: Also here, worth doing as an object to be future-proof?

packages/transaction-pay-controller/src/TransactionPayController.ts
matthewwalsh0: Apologies, I'm not sure why we need this.

A core aspect of the pay controller architecture is that it automatically monitors transaction state changes to generate new required tokens and trigger new quote requests. It generates those required tokens by decoding token transfers or looking at the required assets state, so we break that pattern and abstraction if we instead rely on a constructor callback to build the call-data.

The quote requests are already single-threaded per transaction, so we automatically abort if a new quote begins on the same transaction.

So I assumed the performance gain was the MetaMask Pay UI would call our new `updateTransactionCallback` to now update **all** the transaction data in one **synchronous** request that doesn't trigger any gas or simulation requests.

In the case of money account, it's an asynchronous process to generate the new nested transaction call data, but we then update synchronously, and the quote request is automatically triggered asynchronously.

matthewwalsh0 (changes requested):

packages/transaction-controller/CHANGELOG.md
pedronfigueiredo: Removed the PR entries from the older 69.1.0 release section and kept them under Unreleased.

Addressed in `c63e19bc7` ([commit](https://github.com/MetaMask/core/commit/c63e19bc73e84e3b5ac4ff92d4db5fa8ea45fb9c)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: I agree that the callback should be mutation-only to discourage replacing the entire transaction metadata object. I updated its return type to `void` while keeping the current positional signature and the method’s return value. I’m leaving the options-object refactor for later, since we don’t have additional parameters to justify the larger API change yet.

Addressed in `94dfcc2f0` ([commit](https://github.com/MetaMask/core/commit/94dfcc2f02a4eda1caf3a9693daf7db61a744a17)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: This was already addressed in `472db143a` ([commit](https://github.com/MetaMask/core/commit/472db143a)), which extracted `updateEIP7702BatchData` and updated `updateAtomicBatchData` to use it for generating both `nestedTransactions` and the batch transaction data. The implementation at the commit referenced by this comment already uses that utility, so I don’t think any further change is needed here.

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: I went with `updateTransactionMetadata` since it describes the purpose of the action more accurately than exposing the callback mechanism in the public name. I updated the core controller/messenger API and the companion mobile delegation accordingly.

Core: [`103f5b364`](https://github.com/MetaMask/core/commit/103f5b364e79775f21c1d1b530a437a7c9f83f88)
Mobile: [`66def19a34`](https://github.com/MetaMask/metamask-mobile/commit/66def19a3455574e6edf67d69bb4a31caabfb45a).

packages/transaction-controller/src/utils/eip7702.ts
pedronfigueiredo: I went ahead with the options-object form for this utility. `updateEIP7702BatchData` now accepts `{ from, transactions, updates }`, and I updated its callers and tests. The companion mobile worktree has no direct usage of this utility and required no additional source change; its existing messenger delegation remains covered by [`66def19a34`](https://github.com/MetaMask/metamask-mobile/commit/66def19a3455574e6edf67d69bb4a31caabfb45a).

Core: [`56a765d5c`](https://github.com/MetaMask/core/commit/56a765d5cb2d96a03b74d7d0eb863968d7eb5b85).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Following up on my earlier response: I reconsidered the options-object tradeoff and migrated `updateTransactionMetadata` to accept `{ transactionId, callback }` as well. This keeps both related APIs consistent and gives us room to add options without another positional-argument migration.

Core: [`56a765d5c`](https://github.com/MetaMask/core/commit/56a765d5cb2d96a03b74d7d0eb863968d7eb5b85). The companion mobile delegation required no additional source change beyond the existing rename in [`66def19a34`](https://github.com/MetaMask/metamask-mobile/commit/66def19a3455574e6edf67d69bb4a31caabfb45a).

packages/transaction-pay-controller/src/TransactionPayController.ts
pedronfigueiredo: I agree with the architecture you described. I moved the asynchronous Money Account preparation and stale-intent coordination back to Mobile, reduced Core to one atomic transaction metadata commit, and removed the constructor preparation callback, explicit `updateAmount` orchestration, quote suppression, and manual quote request. The existing transaction-state listener now derives the required tokens and remains the single trigger for the latest-wins quote lifecycle.

One detail that was not guaranteed out of the box was the ‘doesn’t trigger simulation’ part. A generic metadata update that changes `txParams.data` still passes through the normal re-simulation check. I made that contract explicit with `updateTransactionMetadataWithoutResimulation`, which is used only for this coherent MetaMask Pay commit; ordinary transaction update paths retain their existing re-simulation behavior. We’re comfortable with that tradeoff here because these Pay transactions are produced by controlled builders and known providers.

The Mobile coordinator also deduplicates identical intents, prevents superseded asynchronous preparations from committing, validates the current approval/deposit template at commit time, derives `requiredAssets` from the latest metadata, and clears stale gas, simulation, security-alert, and simulation-revert metadata before the reactive quote begins.

Core: [`bd972a763`](https://github.com/MetaMask/core/commit/bd972a763206ac8f680a5bf19a01b51154be6e44)
Mobile: [`9aef7f5def`](https://github.com/MetaMask/metamask-mobile/commit/9aef7f5def8879c9d1bc609f3c1734036276d2fa).

pedronfigueiredo: Follow-up cleanup: with the explicit amount pipeline removed, I also removed its now-unused external quote abort signal and public `abortQuotes` plumbing. The existing per-transaction latest-wins cancellation remains unchanged.

Addressed in `affdc8ef7` ([commit](https://github.com/MetaMask/core/commit/affdc8ef7c8669572983f17e4fdaec83f38b16cb)).

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1622,6 +1623,29 @@ export class TransactionController extends BaseController<
...
 
+  /**
+   * Updates transaction metadata without triggering automatic re-simulation.
+   *
+   * @param options - Update options.
+   * @param options.transactionId - ID of the transaction to update.
+   * @param options.callback - Function that mutates the transaction metadata.
+   * @returns The updated transaction metadata.
+   */
+  updateTransactionMetadataWithoutResimulation({
```
matthewwalsh0: Why not keep the name `updateTransactionMetadata` and have `skipResimulate` as one of the optional properties now we have an object?

Ideally every caller would use this, so the automatic simulation is usually a good mechanic, but not for MetaMask Pay. So we empower the caller to make that decision, but default on.

matthewwalsh0 (changes requested):

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Good point. I’ll rename this to `updateTransactionMetadata` and add an optional `skipResimulate` option that defaults to `false`, so callers get the existing simulation behavior unless they explicitly opt out. I’ll update the Pay usage and related tests accordingly.

Addressed in `9995de8eb9` ([commit](https://github.com/MetaMask/core/commit/9995de8eb988299434c3859b8757f414c2e7437f)).

matthewwalsh0 (approved):

PR #9546, opened 2026-07-17
fix(assets-controller): avoid recomputing getStateForTransactionPay on every transaction state change

### LINKED ISSUE #9545: formatStateForTransactionPay runs keccak256 per asset on every TransactionController:stateChange
### Describe the bug

`formatStateForTransactionPay` (`packages/assets-controller/src/utils/formatStateForTransactionPay.ts`) rebuilds the full legacy state shape — including `toChecksumAddress` (keccak256) for every asset and `parseCaipAssetType` for every asset ID — on every invocation, with no memoization.

It is exposed as the `AssetsController:getStateForTransactionPay` messenger action, which transaction-pay-controller calls on every `TransactionController:stateChange`. Those events fire many times during a single transaction approval, and messenger delegation layers can double the invocations, while the function's actual inputs (`assetsBalance` / `assetsInfo` / `assetsPrice` state slices, selected currency, selected accounts, network configurations) only change when the assets pipeline updates.

### Expected behavior

Repeated calls with unchanged inputs should not re-run keccak256 over the entire asset list. A single-slot memo on input identity (state slices and network configurations by reference — BaseController state updates are immutable — currency by value, accounts/native-asset map element-wise) makes repeat calls O(size of the comparison) instead of O(assets × keccak256).

### Observed behavior

In a 7.8 s CPU profile of MetaMask Mobile (RC, iPhone Pro Max, multi-chain portfolio) captured while triggering a transaction approval in the MetaMask Pay deposit flow:

- keccak256 via `toChecksumAddress` accounted for **10.4% of all samples (~23% of active CPU)** — the single largest active bucket; the dominant call chain is `_TransactionController_addMetadata` → messenger publish → `onTransactionChange` → `parseRequiredTokens` → `getStateForTransactionPay` → `formatStateForTransactionPay`, with `delegatedActionHandler` two levels deep in every stack (the computation runs at least twice per event).
- GC pauses accounted for a further 4.1%, consistent with the object churn of rebuilding the full state shape per event.
- A wallet with ~50 tokens across 3 chains pays 50+ keccak256 hashes per state-change event.

### Steps to reproduce

1. Import a wallet with a large multi-chain token portfolio.
2. Start a transaction-pay flow (e.g. MetaMask Pay deposit) so transaction-pay-controller subscribes to transaction state.
3. Profile the JS thread while a transaction is created/approved.
4. Observe `formatStateForTransactionPay` → `toChecksumAddress` → keccak256 dominating active CPU on every `TransactionController:stateChange`.

MetaMask Mobile is currently mitigating this with a Yarn patch on `@metamask/assets-controller` that adds the single-slot memoization described above: https://github.com/MetaMask/metamask-mobile/pull/33466. Upstreaming that (or caching checksummed addresses) would let the patch be removed.

Related, from the same profile: ethers.js ABI encoding during gas estimation re-derives checksums (`pack` → `getAddress` → keccak256) with no caching (6.5% of samples), so checksum caching at the address level may be worth considering beyond this one function.

Kureev:
## Explanation

`formatStateForTransactionPay` converts the full AssetsController state into the legacy five-controller shape for transaction-pay-controller, running `toChecksumAddress` (keccak256) for every asset and `parseCaipAssetType` for every asset ID on every call. It is exposed as the `AssetsController:getStateForTransactionPay` messenger action, which transaction-pay-controller invokes on every `TransactionController:stateChange` — many times during a single transaction approval, and at least twice per event through messenger delegation — while its inputs only change when the assets pipeline updates.

In a CPU profile of MetaMask Mobile (RC) captured while triggering a transaction approval in the MetaMask Pay deposit flow, keccak256 via `toChecksumAddress` accounted for 10.4% of all samples (~23% of active CPU), the single largest active bucket, with GC pauses (4.1%) as a symptom of the per-event object churn. See #9545 for the full breakdown.

This change memoizes the last result on input identity: the `assetsBalance` / `assetsInfo` / `assetsPrice` state slices and `networkConfigurationsByChainId` are compared by reference (BaseController state updates are immutable, so unchanged slices retain identity), `selectedCurrency` by value, and the per-call-rebuilt `accounts` array and `nativeAssetIdentifiers` record element-wise. Repeat calls with unchanged inputs return the cached object; any real input change recomputes. The conversion body is unchanged, moved to an internal `computeStateForTransactionPay`.

MetaMask Mobile currently carries this fix as a Yarn patch (MetaMask/metamask-mobile#33466), which can be dropped once this lands and a release is picked up.

## References

* Fixes #9545
* Mobile-side mitigation: MetaMask/metamask-mobile#33466

## Checklist

---

> [!NOTE]
> **Low Risk**
> Performance-only change on a read path; freezing the legacy state object could affect consumers that mutate the return value, though transaction-pay is expected to treat it as read-only.
> 
> **Overview**
> **Reduces CPU during transaction approval** by keeping the last `formatStateForTransactionPay` result when inputs are unchanged, so `AssetsController:getStateForTransactionPay` does not repeat expensive `toChecksumAddress` / CAIP work on every `TransactionController:stateChange`.
> 
> Memoization compares immutable controller slices (`assetsBalance`, `assetsInfo`, `assetsPrice`, `networkConfigurationsByChainId`) by reference and rebuilt `accounts` / `nativeAssetIdentifiers` with lodash `isEqual`. Parameter equality is centralized in `isTransactionPayParamsIdentical`. **Cached outputs are `Object.freeze`d** so in-place mutations cannot poison the cache for later callers.
> 
> Tests expand the memoization suite (cache hits, invalidation on balance/accounts/currency changes, frozen results, cache vs recompute parity). Changelog documents the freeze behavior alongside the broader memoization work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 186e8f393bef316ecaef449bc74235fd496b3131. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/utils/formatStateForTransactionPay.ts
```diff
@@ -75,15 +132,40 @@ function getAmountFromBalance(balance: AssetBalance): string {
...
+    lastCall.params.selectedCurrency === params.selectedCurrency &&
+    lastCall.params.networkConfigurationsByChainId ===
+      params.networkConfigurationsByChainId &&
+    areAccountsEqual(lastCall.params.accounts, params.accounts) &&
+    areRecordsShallowEqual(
+      lastCall.params.nativeAssetIdentifiers,
+      params.nativeAssetIdentifiers,
+    )
+  ) {
+    return lastCall.result;
```
salimtb, line 128: nit / optional : Cache hits now return the *same* object reference across calls (previously fresh each time). Since this is a public export, a consumer mutating the result would poison the cache. Worth freezing it before caching:

```ts
lastCall = { params, result: Object.freeze(result) };
```

packages/assets-controller/src/utils/formatStateForTransactionPay.ts
```diff
@@ -49,22 +49,79 @@ export type TransactionPayLegacyFormat = {
...
+  if (!a || !b) {
+    return false;
+  }
+  const aKeys = Object.keys(a);
+  const bKeys = Object.keys(b);
+  if (aKeys.length !== bKeys.length) {
+    return false;
+  }
+  return aKeys.every((key) => a[key] === b[key]);
+}
```
salimtb: Optional: `lodash` is already a dep, so both helpers could just be `isEqual` (deep, but inputs are tiny/flat so equivalent):

```ts
isEqual(lastCall.params.accounts, params.accounts)
isEqual(lastCall.params.nativeAssetIdentifiers, params.nativeAssetIdentifiers)
```

Only for these rebuilt inputs though — keep the big state slices reference-compared, and avoid `_.memoize` (unbounded cache).

salimtb:
nice one , this is a good perf improvement

packages/assets-controller/src/utils/formatStateForTransactionPay.ts
```diff
@@ -75,15 +132,40 @@ function getAmountFromBalance(balance: AssetBalance): string {
...
+    lastCall.params.assetsInfo === params.assetsInfo &&
+    lastCall.params.assetsPrice === params.assetsPrice &&
+    lastCall.params.selectedCurrency === params.selectedCurrency &&
+    lastCall.params.networkConfigurationsByChainId ===
+      params.networkConfigurationsByChainId &&
+    areAccountsEqual(lastCall.params.accounts, params.accounts) &&
+    areRecordsShallowEqual(
+      lastCall.params.nativeAssetIdentifiers,
+      params.nativeAssetIdentifiers,
+    )
```
Prithpal-Sooriya: Lets move this into a small util: `isTransactionPayParamsIdentical()`

salimtb (dismissed):

Cal-L (dismissed):
Platform LGTM

salimtb (approved):

PR #9549, opened 2026-07-20
feat(perps-controller): add dedicated aggregated order book connection

michalconsensys:
## Explanation

The order-book panel needs a server-aggregated (`nSigFigs`) `l2Book` stream in addition to the raw, full-precision stream used for spread/slippage. The subscription service multiplexes every subscription onto a single Hyperliquid WebSocket, and the SDK dispatches `l2Book` events by `coin` only. Running both the raw and the aggregated subscription for the same coin on that shared socket cross-contaminates them, so the coarse ladder and the precise spread clobber each other.

This PR adds `AggregatedOrderBookConnection`, a service that owns a dedicated Hyperliquid WebSocket used solely for the panel's aggregated `l2Book` subscription. Because that socket only ever carries a single `l2Book` stream, the collision is removed entirely and the main socket is never touched by the panel's grouping.

Key details:

- The socket is created lazily on the first subscription and torn down once the last subscription is removed, so it exists only while an order-book panel is open.
- Network is a global setting, so the transport is recreated when `isTestnet` changes between (re)subscriptions.
- Connection health is surfaced via `OrderBookConnectionStatus` (`connecting` / `connected` / `error`) so the UI can show a manual reconnect affordance. `error` is reported only once the SDK's automatic reconnection is exhausted (`terminate`), which is unrecoverable and requires a fresh transport on the next subscribe.
- The subscription is sent through the transport's raw payload so `fast: true` reaches the server (the SDK's typed `l2Book` method drops unknown fields and cannot request fast mode).
- `processAggregatedOrderBook` transforms the raw snapshot into the `OrderBookData` shape the UI consumes, mirroring the subscription service's internal processing so this connection is a drop-in replacement on the aggregated channel.
- Unsubscribe follows the subscription service's synchronous-unsubscribe contract: the returned function can be called before the async subscribe resolves and will cancel the pending subscription.

The service and its `processAggregatedOrderBook` helper plus public types are exported from the package index.

## References

Fixes https://consensyssoftware.atlassian.net/browse/TAT-3309

## Checklist

---

> [!NOTE]
> **Medium Risk**
> New real-time WebSocket lifecycle code with many edge cases (network flips, stale subscriptions, reconnection); additive public API only—no changes to existing `subscribeToOrderBook` paths in this diff.
> 
> **Overview**
> Adds **`AggregatedOrderBookConnection`**, a new exported service so the Pro order-book panel can subscribe to server-aggregated Hyperliquid `l2Book` (`nSigFigs`, fast mode) on its **own WebSocket**, instead of sharing the multiplexed subscription socket where raw and aggregated streams for the same coin would collide.
> 
> The service **reference-counts** subscriptions, opens the socket lazily and closes it when the last subscriber leaves, recreates the transport on testnet/mainnet changes or after reconnection exhaustion, and exposes **`OrderBookConnectionStatus`** (`connecting` / `connected` / `error`) for manual reconnect UI. **`processAggregatedOrderBook`** maps raw snapshots to **`OrderBookData`** for drop-in use alongside existing book processing.
> 
> Public API surface is documented in the changelog and re-exported from the package **`index`**; behavior is covered by a large unit test suite.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62067a4ec4f1ee463c58ad77223e95c90c003b95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


michalconsensys:
Currently if you subscribe twice to l2Book with 2 different parameters, then you don't know from which subscription you are getting from the response. This can affect how we for example compute mid price, if there are 2 concurrent subscriptions. This PR addresses this issue by creating a new WS connection for the orderbook

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,320 @@
...
+    // the UI surfaces with a manual reconnect button.
+    const handleOpen = (): void => reportStatus('connected');
+    const handleClose = (): void => reportStatus('connecting');
+    const handleTerminate = (): void => {
+      this.#terminated = true;
+      reportStatus('error');
+    };
+    socket.addEventListener('open', handleOpen);
+    socket.addEventListener('close', handleClose);
+    socket.addEventListener('terminate', handleTerminate);
```
geositta: WebSocketTransport.socket is an @nktkas/rews@4.0.1 ReconnectingWebSocket, reached transitively through @nktkas/hyperliquid. That socket emits only open, close, error, and message; permanent termination aborts socket.terminationSignal instead of dispatching terminate.

addEventListener accepts arbitrary event names, so this compiles without warning, but handleTerminate is dead code. During a sustained outage, rews emits a final close, the panel reports connecting indefinitely, #terminated remains false, and #ensureTransport can reuse the dead transport. The documented error state and manual reconnect affordance therefore cannot execute.

Please listen for terminationSignal’s abort event or inspect terminationSignal.aborted in handleClose, and update the tests to exercise that production signal rather than manually dispatching a nonexistent terminate event.

HyperLiquidClientService.ts:266 contains the same pre-existing pattern, which explains how this was copied but should be tracked separately from this PR.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,320 @@
...
+   * Mirrors the subscription service's synchronous-unsubscribe contract: the
+   * returned function can be called before the async subscribe resolves and
+   * will cancel the pending subscription.
+   *
+   * @param params - Subscription parameters.
+   * @returns An unsubscribe function.
+   */
+  subscribe(params: SubscribeAggregatedOrderBookParams): () => void {
+    const levels = params.levels ?? DEFAULT_LEVELS;
+    const transport = this.#ensureTransport(this.#isTestnet());
```
geositta, line 220: The dedicated socket fixes the current raw versus aggregated collision, but this service still reuses that socket for every active subscription. If the same symbol is subscribed with different nSigFigs or mantissa values, every l2Book listener receives both responses, and the payload contains no aggregation identifier that can demultiplex them.

This leaves the concurrent different-parameters case unresolved within the new public API. Please enforce one payload per transport, use separate transports for distinct same-symbol payloads, or reject unsupported concurrent subscriptions. Add a same-symbol/different-parameters test; the current BTC/ETH test cannot reproduce the ambiguity.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,320 @@
...
+      this.#transport &&
+      this.#transportIsTestnet === isTestnet &&
+      !this.#terminated
+    ) {
+      return this.#transport;
+    }
+    // First use, the network changed, or the previous socket was terminated —
+    // (re)create the dedicated transport.
+    this.#closeTransport();
+    const transport = new WebSocketTransport({ isTestnet });
```
geositta: Please configure finite reconnection attempts here. WebSocketTransport defaults maxRetries to Infinity, so an ordinary sustained outage never exhausts reconnection and never reaches the documented manual-reconnect state. Reusing HYPERLIQUID_TRANSPORT_CONFIG would apply the package’s existing five-attempt policy and keep both transports consistent.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,320 @@
...
+    // transport directly (this is exactly what the typed method does, minus the
+    // validation) so `fast: true` reaches the server. The listener receives the
+    // SDK's `CustomEvent`, whose `detail` is the `l2Book` snapshot.
+    transport
+      .subscribe<HyperliquidL2BookEvent>(
+        'l2Book',
+        {
+          type: 'l2Book',
+          coin: params.symbol,
+          nSigFigs: params.nSigFigs ?? null,
```
geositta: Omitting nSigFigs sends null, which requests the raw full precision book rather than an aggregated book. That makes the default behavior contradict the service contract. Please require nSigFigs or provide an aggregation default such as the existing service's value of 5, then cover omission in a test.

geositta (changes requested):
Requesting changes. Creating a separate socket isolates the panel's aggregated book from the raw book on the main connection, but the implementation does not yet enforce the intended guarantee for concurrent l2Book subscriptions with different parameters. The new service reuses its transport, so two same symbol aggregation configurations remain indistinguishable.

The outage recovery contract is also nonfunctional: the transport retries indefinitely and listens for a terminate event that its underlying ReconnectingWebSocket never emits. During a sustained outage, the panel remains connecting, never exposes manual reconnect, and may reuse the terminated transport.

Please enforce subscription isolation for distinct same symbol payloads, configure finite retries, observe terminationSignal, and test these production behaviors. Aligning the optional nSigFigs default with the service’s aggregated contract remains a recommended nonblocking improvement.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/220195cfa83d92203c7162071562d4d4952957bd

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/358ddeb51b402c275625c8c974d8f0e9752ff2e0

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/d82a3e1654d7ab4d843a519dd3ff146dff8869e1

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/67b3b13127eb2e1c9d45f002f98a51f14bd3ed45

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,386 @@
...
+
+    reportStatus('connecting');
+
+    // The SDK's typed `l2Book` subscription drops unknown fields, so it can't
+    // request `fast` mode. Send the raw subscription payload through the
+    // transport directly (this is exactly what the typed method does, minus the
+    // validation) so `fast: true` reaches the server. The listener receives the
+    // SDK's `CustomEvent`, whose `detail` is the `l2Book` snapshot.
+    transport
+      .subscribe<HyperliquidL2BookEvent>(
```
geositta: The SDK distinguishes initial subscription failures from post-confirmation failures. The existing .catch handles the initial request, but after reconnecting, a rejected resubscription removes the listener and reports the failure only through options.onError.

Without that callback, the service reports connected on socket open and then silently stops receiving updates. The panel displays a frozen order book as connected, while #activeCount and #payloads still represent a live subscription.

Please pass an idempotent onError handler that reports error and calls teardown, reuse it from the existing .catch, and test an initially confirmed subscription followed by a rejected resubscription.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,386 @@
...
+        }
+        if (this.#activeCount === 0) {
+          this.#closeTransport();
+        }
+      }
+    };
+
+    reportStatus('connecting');
+
+    // The SDK's typed `l2Book` subscription drops unknown fields, so it can't
```
geositta: Nonblocking:
@nktkas/hyperliquid@0.33.1 includes fast in its typed l2Book schema and forwards subscription options, including onError. Please correct this outdated comment. Switching to the typed API would restore payload validation, but that refactor does not need to block the lifecycle fix above.

geositta (changes requested):
equesting changes. The latest commits resolve the previous isolation, retry, termination, and aggregation contract findings. One post-reconnect failure path remains: if the SDK rejects a confirmed subscription’s resubscription, it removes the listener and reports the failure only through onError. The service currently remains connected while displaying a frozen order book.

Please handle that callback with the existing error-and-teardown behavior and add coverage for the reconnect/rejection sequence. The outdated typed-API comment is non-blocking.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,404 @@
...
+          try {
+            await sub.unsubscribe();
+          } catch {
+            // Ignore cleanup errors on an already-cancelled subscription.
+          }
+          return undefined;
+        }
+        subscription = sub;
+        reportStatus('connected');
+        return undefined;
```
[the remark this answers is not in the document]
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/9b33f370f9ec4a7562e6dfe1f05a9f6f536de572

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/57b79b354edafd92b22591a98d6275a542db9aa6

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,417 @@
...
+ * Owns a dedicated Hyperliquid WebSocket connection used solely for the
+ * order-book panel's server-aggregated `l2Book` subscription.
+ *
+ * The main connection (managed by the subscription service) multiplexes every
+ * subscription onto a single socket. The Hyperliquid SDK dispatches `l2Book`
+ * events by `coin` only, so running the raw (full-precision) and the aggregated
+ * (`nSigFigs`) subscriptions for the same coin on that shared socket
+ * cross-contaminates them — the coarse ladder and the precise spread/slippage
+ * clobber each other. Giving the aggregated subscription its own socket removes
+ * the collision entirely: this socket only ever carries a single `l2Book`
```
geositta, line 149: Nonblocking: Could we clarify this as "at most one l2Book payload per asset"? The service intentionally supports different assets on the same transport, while preventing ambiguous same asset configurations. "A single stream" does not match that behavior.

geositta (dismissed):
Approved with one nonblocking request to tweak a code comment to be more accurate.

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  #closeTransport(): void {
+    const transport = this.#transport;
+    this.#transport = null;
+    this.#activeCount = 0;
+    this.#payloads.clear();
+    this.#terminated = false;
+    if (transport) {
+      transport.close();
+    }
```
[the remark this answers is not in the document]
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/1cfd42cb1458c7e9703caf4f98ad09d0d937bb2a

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
```diff
@@ -0,0 +1,479 @@
...
+    for (const forceTerminate of subscriptions) {
+      forceTerminate();
+    }
```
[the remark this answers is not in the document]
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/6951101dcfba1e07d99d5a4aa68d4d99d9eb2ad6

packages/perps-controller/src/services/AggregatedOrderBookConnection.ts
michalconsensys: https://github.com/MetaMask/core/pull/9549/commits/0dd848349bc1e020728af4b409e464015a94fa9e

geositta (approved):

PR #9555, opened 2026-07-20
perf(assets-controller): memoize hot-path asset ID and legacy format converters

salimtb:
Avoid repeated keccak256 checksumming and CAIP parsing on read paths that run far more often than assets state changes.

## Explanation

`AssetsController`'s read paths recompute pure, deterministic conversions on
every call, even when the underlying assets state hasn't changed:

- `normalizeAssetId` runs `parseCaipAssetType` + `parseCaipChainId` and, for EVM
  ERC-20 tokens, `toChecksumAddress` (keccak256) on every call — including the
  common case where the same already-checksummed ID is passed repeatedly across
  the pipeline.
- `formatExchangeRatesForBridge` (via `getExchangeRatesForBridge`) and
  `formatStateForTransactionPay` (via `getStateForTransactionPay`) re-run
  keccak256 checksumming and CAIP parsing for every priced/held asset. These are
  invoked on hot paths — e.g. bridge quote/rate refreshes and every
  `TransactionController:stateChange` — far more often than the assets state that
  feeds them actually changes.

This PR memoizes those conversions so unchanged inputs skip the redundant work:

- **`normalizeAssetId`** is wrapped in lodash `memoize`, keyed on the asset ID.
  Repeated calls with the same ID (the norm once IDs are checksummed) return the
  cached result without re-parsing or re-checksumming.
- **`formatExchangeRatesForBridge`** and **`formatStateForTransactionPay`** cache
  their last result. Cache validity is checked by identity (`===`) for
  BaseController state slices — which keep stable references across updates thanks
  to immer — and by lodash `isEqual` for maps/arrays that are rebuilt each call
  (`nativeAssetIdentifiers`, and `accounts` for transaction-pay). The compute
  logic is split into internal `computeExchangeRatesForBridge` /
  `computeStateForTransactionPay` helpers, with the exported functions handling
  the cache check.
- **`#getNativeAssetMap`** now returns a stable empty-object constant
  (`EMPTY_NATIVE_ASSET_MAP`) when the query cache is empty, so `?? {}` identity
  churn doesn't defeat downstream memoization.

Each memoized module exports a `clear*CacheForTesting` helper (from the internal
`utils` barrel) so tests can reset cache state between cases. The
`FormatExchangeRatesForBridgeParams` and `FormatStateForTransactionPayParams`
parameter types are now named and exported for reuse.

No behavior changes for consumers: the functions are pure and the memoized
results are value-equal to the previously recomputed ones. Note that cached
results are returned by reference and should be treated as read-only.

## References

N/A

## Checklist

---

> [!NOTE]
> **Low Risk**
> Performance-only caching of pure converters with no API or behavior contract changes; main caveat is consumers must not mutate cached return values.
> 
> **Overview**
> Adds **memoization** on read paths that repeatedly run CAIP parsing and EVM address checksumming (`toChecksumAddress` / keccak256) even when underlying assets state is unchanged.
> 
> **`normalizeAssetId`** is wrapped in lodash `memoize` so repeated calls with the same CAIP-19 ID return a cached checksummed result.
> 
> **`formatExchangeRatesForBridge`** and **`formatStateForTransactionPay`** now keep a single-entry cache: controller state slices are compared by reference (`===`), while rebuilt `nativeAssetIdentifiers` / `accounts` use lodash `isEqual`. Heavy work lives in internal `compute*` helpers. Named param types **`FormatExchangeRatesForBridgeParams`** and **`FormatStateForTransactionPayParams`** are exported; test-only **`clear*CacheForTesting`** helpers reset caches between cases.
> 
> **`#getNativeAssetMap`** returns a module-level **`EMPTY_NATIVE_ASSET_MAP`** instead of `?? {}`, so empty native maps keep stable identity and do not invalidate the formatter caches.
> 
> Outputs are intended to be value-equal to before; cached objects are returned by reference and should be treated as read-only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1266836163271002efe520ab4a7eca56847eafdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Memoize hot-path asset ID / legacy-format conversions to cut repeated keccak256 (`toChecksumAddress`) and CAIP parsing
+  - `normalizeAssetId` uses lodash `memoize` so already-normalized IDs skip re-checksumming across the pipeline
+  - `formatExchangeRatesForBridge` caches the last result on input identity (`===` for BaseController slices, lodash `isEqual` for rebuilt native maps); exports `FormatExchangeRatesForBridgeParams`
+  - `formatStateForTransactionPay` caches the last result the same way (`isEqual` for rebuilt `accounts` / `nativeAssetIdentifiers`); exports `FormatStateForTransactionPayParams`
+  - `#getNativeAssetMap` returns a stable empty object when uncached so memoized formatters are not busted by `?? {}` identity churn
 - `TokenDataSource` EVM spam filtering now uses per-chain floors from Token API `GET /v1/suggestedOccurrenceFloors` (`queryApiClient.token.fetchV1SuggestedOccurrenceFloors`) instead of a hardcoded minimum of 3 occurrences. Chains missing from the response (or a failed floors fetch) still fall back to 3 ([#9537](https://github.com/MetaMask/core/pull/9537))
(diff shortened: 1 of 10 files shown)


juanmigdr (approved):
Very impressed with the changes! LGTM 🚀

PR #9556, opened 2026-07-20
fix: do not default to zero if exchange rate is not available

GeorgeGkas:
## Explanation

`selectExchangeRateByAssetId` fell back to a price of `0` whenever an EVM token's market data entry had a missing or zero `price`. That zero exchange rate had two harmful side effects:

- It made affected tokens (e.g. mUSD) display a `$0.00` fiat value wherever this selector's output was used.
- Because the resulting `"0"` rate string is truthy, `selectIsAssetExchangeRateInState` treated the token as already having a known rate, which caused the bridge controller to skip fetching the token's real price from the price API — so the `$0.00` display never self-corrected.

This PR changes `selectExchangeRateByAssetId` to return `{}` when the market data price is missing or `0`, instead of computing a rate from a `0` price. This makes `selectIsAssetExchangeRateInState` correctly report "no rate known" for these tokens, so the controller fetches their real exchange rate instead of stopping at a stale/absent `$0.00` value.

No other packages needed changes; this is isolated to `bridge-controller`'s selector logic.

## References

https://consensyssoftware.atlassian.net/browse/SWAPS-4797

## Checklist

---

> [!NOTE]
> **Low Risk**
> Small, targeted selector change in bridge-controller with new unit tests; affects fiat display and rate-fetch gating only, not transaction submission.
> 
> **Overview**
> Fixes bridge/swap quotes showing **$0.00** fiat for some EVM tokens (e.g. mUSD) when market data had a missing or zero `price`.
> 
> **`selectExchangeRateByAssetId`** no longer coerces a missing price to `0` or builds an exchange rate from a zero price. It returns `{}` instead, so **`selectIsAssetExchangeRateInState`** does not treat a truthy `"0"` string as a known rate and the controller can fetch the real price from the API.
> 
> Tests cover zero/missing market `price` and the “not in state” path; changelog updated under `[Unreleased]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3014f4f88ec30b3485912a2d2cd7a0fcd3957418. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bridge-controller/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = merge(baseConfig, {
       statements: 98.79,
     },
     './src/selectors.ts': {
-      branches: 91.58,
+      branches: 91.42,
```
GeorgeGkas, line 26: I lowered the branch coverage threshold for selectors.ts because the fix removed the `?? 0` fallback on `evmTokenExchangeRateForAddress?.price` (now just `evmTokenExchangeRateForAddress?.price`). Removing a 100%-covered branch pair from the denominator pulls the average down slightly even though no actual test coverage regressed.

infiniteflower (dismissed):

micaelae (approved):

PR #9588, opened 2026-07-22
feat: wire ClaimsService and ClaimsController initialization

tuna1207:
## Explanation

Adds `ClaimsService` and `ClaimsController` as default initialized units in the `@metamask/wallet` package, following the existing `InitializationConfiguration` pattern used by `TransactionController`, `StorageService`, and other wallet instances.

`@metamask/wallet` is the shared controller-integration layer for extension, mobile, and other clients, so build-specific and runtime values are injected via `instanceOptions` rather than hardcoded.

**ClaimsService** — selects the Claims API environment and platform `fetch` implementation. `instanceOptions.claimsService.env` and `instanceOptions.claimsService.fetchFunction` are required. The wallet init delegates `AuthenticationController:getBearerToken` to the service messenger; hosts must register `AuthenticationController` on the supplied root messenger before authenticated Claims API calls succeed.

**ClaimsController** — stateful controller with no additional constructor options beyond optional persisted `state`. The wallet init delegates only the ClaimsService actions the controller calls (`fetchClaimsConfigurations`, `getRequestHeaders`, `getClaimsApiUrl`, `generateMessageForClaimSignature`, `getClaims`) plus `KeyringController:signPersonalMessage`.

Wallet release work is intentionally deferred; this PR lands the integration and documents consumer options only.

### Per-environment options

| Option | Extension | Mobile | Default (Node / tests / wallet-cli) |
| --- | --- | --- | --- |
| `claimsService.env` | Build flavor (`Env.DEV` / `Env.UAT` / `Env.PRD`) | Build flavor (`Env.DEV` / `Env.UAT` / `Env.PRD`) | `Env.PRD` in wallet-cli; `Env.DEV` in wallet unit tests |
| `claimsService.fetchFunction` | Platform `fetch` | Platform `fetch` | `globalThis.fetch` |
| `claimsController` state | Persisted via wallet `state` blob when supplied | Persisted via wallet `state` blob when supplied | Optional; defaults apply when omitted |
| `AuthenticationController:getBearerToken` | Host must register `AuthenticationController` on the wallet root messenger (not an `instanceOptions` slot) | Host must register `AuthenticationController` on the wallet root messenger (not an `instanceOptions` slot) | Not wired in wallet-cli; authenticated Claims calls require host auth wiring |

### Verification

- `yarn workspace @metamask/wallet run test`
- `yarn workspace @metamask/wallet run jest --no-coverage src/initialization/instances/claims-service/claims-service.test.ts src/initialization/instances/claims-controller/claims-controller.test.ts`
- `yarn workspace @metamask/wallet run messenger-action-types:check`
- `yarn workspace @metamask/wallet run lint:tsconfigs`
- `yarn workspace @metamask/wallet-cli run lint:tsconfigs`
- `yarn readme-content:check`
- `yarn codeowners:check`
- `yarn lint:teams`

## Checklist

- [x] I've introduced breaking changes in this PR and documented required consumer options in the table above
- [ ] Wallet release deferred to a follow-up request


packages/wallet/src/initialization/instances/claims-controller/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { Env } from '@metamask/claims-controller';
+
+export type ClaimsControllerInstanceOptions = Record<string, never>;
```
lwin-kyaw: Please use strict types.
```suggestion
export type ClaimsControllerInstanceOptions = 
  ConstructorParameters<typeof ClaimsController>[0];
```

packages/wallet/src/initialization/instances/claims-controller/types.ts
```diff
@@ -0,0 +1,16 @@
...
+   * Claims API environment. Supplied by the consumer per build flavor
+   * (dev, uat, production).
+   */
+  env: Env;
+
+  /**
+   * Platform fetch implementation used for Claims API requests.
+   */
+  fetchFunction: typeof fetch;
+};
```
lwin-kyaw: Same here. We can re-use the construct args type.
```suggestion
export type ClaimsServiceInstanceOptions = 
  ConstructorParameters<typeof ClaimsService>[0];
```

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
```diff
@@ -125,6 +126,10 @@ function getInstanceOptions(): WalletOptions['instanceOptions'] {
       storage: new InMemoryStorageAdapter(),
     },
     remoteFeatureFlagController: REMOTE_FEATURE_FLAG_OPTIONS,
+    claimsService: {
```
lwin-kyaw: Why? TransactionController doesn't interact with claims though.

packages/wallet/src/index.ts
```diff
@@ -1,6 +1,7 @@
 export { Wallet } from './Wallet.js';
 export { AlwaysOnlineAdapter } from './initialization/instances/connectivity-controller/always-online-adapter.js';
 export { importSecretRecoveryPhrase } from './utilities.js';
+export { Env as ClaimsEnv } from '@metamask/claims-controller';
```
lwin-kyaw: I don't think we need to export this. 
Clients can just import from `@metamask/claims-controller`.

packages/wallet/src/types.ts
```diff
@@ -29,6 +33,8 @@ export type WalletOptions = {
 
 export type InstanceSpecificOptions = {
   approvalController?: ApprovalControllerInstanceOptions;
+  claimsService: ClaimsServiceInstanceOptions;
+  claimsController?: ClaimsControllerInstanceOptions;
```
lwin-kyaw: May I know why is `claimsController` optional but `claimsService` is required?

packages/wallet/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Wire `ClaimsService` and `ClaimsController` into the default wallet initialization ([#9588](https://github.com/MetaMask/core/pull/9588))
+  - Passing `instanceOptions.claimsService.env` is now required.
```
lwin-kyaw: I think just one entry (above **BREAKING**) is sufficient enough, to keep the CHANGELOG clean. These are not really necessary.

lwin-kyaw:
One more thing I noticed is that shield/subscriptions/claims controllers have the different constructor interfaces.

Some accept `ENV` and some accept `URL`. I think they should be consistent. 
And what's your thought on replacing `ClaimsEnv` `SubscriptionsEnv` with `ENV` from the `AuthenticationController`?

pedronfigueiredo (changes requested):
I see there's a bunch of comments and merge conflicts, can you address them and then re-request my review here on github once it's ready for another look?

packages/wallet/src/types.ts
tuna1207: The controller only needs state and messenger, which initialization already provides. There's nothing for the consumer to configure, so the slot is optional.
Claim Service need required fetchFunction option

tuna1207:
> One more thing I noticed is that shield/subscriptions/claims controllers have the different constructor interfaces.
> 
> Some accept `ENV` and some accept `URL`. I think they should be consistent. And what's your thought on replacing `ClaimsEnv` `SubscriptionsEnv` with `ENV` from the `AuthenticationController`?

Env is a generic infra concept, not an auth concept. Making claims/subscription/shield depend on profile-sync for a 3-value enum couples unrelated packages - every profile-sync major bump (it's at v28, moves fast) forces churn in dependents. claims/subscription currently only use it type-only; adding a runtime import makes it a real dependency.

I think we can move Env into a neutral shared package - @metamask/controller-utils is the natural home since everyone already depends on it. Then profile-sync, claims, subscription, geolocation, money-account, sentinel all import from there. Each package keeps its own service URL map.

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
tuna1207: similar to https://github.com/MetaMask/core/pull/9588#discussion_r3682323926

pedronfigueiredo: Is this really required for the test file only, since no change has been made to the corresponding controller initialization file? If not, please remove

pedronfigueiredo (changes requested):

packages/wallet/src/initialization/instances/claims-service/claims-service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const claimsService: InitializationConfiguration<
+  ClaimsService,
+  ClaimsServiceMessenger
+> = {
+  name: 'ClaimsService',
+  init: ({ messenger, options }) =>
+    new ClaimsService({
+      messenger,
+      env: options.env ?? Env.PRD,
```
mcmire, line 17: Is this safe, do we want this to be development by default instead? I don't know what convention we've been using for other services, but it seems like we wouldn't want to hit production unnecessarily.

Alternatively should we make this option required and have clients always pass this?

mcmire (commented):
Nice work, I just have one comment.

packages/wallet/src/initialization/instances/claims-service/claims-service.ts
tuna1207: This one follow other controllers like profile sync authentication controller, money account api service all default to prd, since for all services with auth we need to follow authentication controller i think we can keep this for now and refactor all controllers to default to dev in the future if needed

tuna1207:
@mcmire @pedronfigueiredo i have refactored claim service to its own separate data service and resolve all the comments, please help me review again, thank you

pedronfigueiredo (dismissed):

tuna1207:
@metamaskbot publish-previews

packages/claims-controller/src/ClaimsService.ts
```diff
@@ -119,29 +165,37 @@ export class ClaimsService {
...
   async fetchClaimsConfigurations(): Promise<ClaimsConfigurationsResponse> {
     try {
-      const headers = await this.getRequestHeaders();
-      const url = `${this.getClaimsApiUrl()}/configurations`;
-      const response = await this.#fetch(url, {
-        headers,
+      const bearerToken = await this.messenger.call(
+        'AuthenticationController:getBearerToken',
+      );
+      const profileKey = bytesToHex(await sha256(stringToBytes(bearerToken)));
```
chaitanyapotti: can u just not use the sub inside the token + service name + method name as query key?

packages/claims-controller/src/ClaimsService.ts
```diff
@@ -153,28 +207,41 @@ export class ClaimsService {
...
+
+      return await this.fetchQuery({
+        queryKey: [`${this.name}:getClaims`, profileKey],
+        // TODO: Restore default staleTime once claim reads are invalidated
```
chaitanyapotti, line 216: todo here

packages/claims-controller/src/ClaimsService.ts
```diff
@@ -186,28 +253,41 @@ export class ClaimsService {
...
-      const response = await this.#fetch(url, {
-        headers,
+      const bearerToken = await this.messenger.call(
+        'AuthenticationController:getBearerToken',
+      );
+      const profileKey = bytesToHex(await sha256(stringToBytes(bearerToken)));
+
+      return await this.fetchQuery({
+        queryKey: [`${this.name}:getClaimById`, id, profileKey],
+        // TODO: Restore default staleTime once claim reads are invalidated
```
chaitanyapotti, line 259: todo here

packages/claims-controller/src/ClaimsService.ts
tuna1207: this is because in client (ext) we are currently calling submit api outside of background controller to avoid file serialization https://github.com/MetaMask/metamask-extension/blob/9def44e4a61641fb2c34010a2b4da05f0d3b31cb/ui/store/actions.ts#L7966
we can update this to have separate invalidation call for client to trigger but that would require client update

packages/claims-controller/src/ClaimsService.ts
tuna1207: same as https://github.com/MetaMask/core/pull/9588#discussion_r3710573802

packages/claims-controller/src/ClaimsService.ts
tuna1207: updated to use profile id as key https://github.com/MetaMask/core/pull/9588/commits/6063e211876dd6d0cfe01a6126268f0ff865662d

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

sirtimid (dismissed):

packages/wallet/src/initialization/instances/claims-service/claims-service.ts
Gudahtt: The antipattern here is that `env` exists as an option in the first place. We should be passing in config, rather than hard-coding it into the library and using `env` to choose which to use at runtime. This puts it under the control of the clients, which gives us more confidence that the clients are correctly configured, and more flexibility in updating config on short notice if we need to.

Agreed that it can wait for another PR, but please make a ticket to track getting rid of this variable, this is very much contrary to our application architecture.

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
Gudahtt: It's strange that this didn't require changes to `packages/wallet/src/initialization/instances/gas-fee-controller/gas-fee-controller.test.ts` as well 🤔 That file also uses the wallet `InstanceSpecificOptions` type, which has the new required property that necessitated this change.

Gudahtt: The fact that we're embedding the options for the _entire wallet library_ in these test suites is a bit odd. It might be better to test initialization in the wallet test suite, rather than in the test suites for the initialization options for each individual controller. Or _at least_ we should be sharing a "default options" fixture.

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

sirtimid (dismissed):

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

sirtimid (dismissed):

packages/wallet/src/initialization/instances/claims-service/claims-service.ts
tuna1207: Agree, i have created a [ticket](https://consensyssoftware.atlassian.net/browse/SUB-986?atlOrigin=eyJpIjoiNzM2MzZhM2I1ODEzNDVlYmJhYWZjNWM3MTJjYWMyYmQiLCJwIjoiaiJ9) for this and will update in a subsequent PR.

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

chaitanyapotti (dismissed):

packages/wallet/src/initialization/instances/claims-controller/types.ts
```diff
@@ -0,0 +1,6 @@
+import type { ClaimsControllerOptions } from '@metamask/claims-controller';
+
+export type ClaimsControllerInstanceOptions = Omit<
```
FrederikBolding: There seems to be no options consumed by `claims-controller`. In that case I would just omit this file.

packages/wallet/src/initialization/instances/claims-service/claims-service.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import type {
+  DefaultActions,
+  DefaultEvents,
+  RootMessenger,
+} from '../../defaults.js';
+import { claimsService } from './claims-service.js';
+import type { ClaimsServiceInstanceOptions } from './types.js';
+
+const REQUIRED_OPTIONS: ClaimsServiceInstanceOptions = {
+  env: Env.DEV,
```
FrederikBolding: Is it intentional that we are testing against dev but defaulting to prod?

packages/wallet/src/initialization/instances/claims-controller/types.ts
tuna1207: removed in https://github.com/MetaMask/core/pull/9588/commits/1b7eef84238e37072d2b4ae292cc7c1da41e108c

packages/wallet/src/initialization/instances/claims-service/claims-service.test.ts
tuna1207: i updated the test to make the test case correct default PRD and DEV env for specific provided case only https://github.com/MetaMask/core/pull/9588/commits/4d3a86516c57d37c4836762cb759154af616dd4a

FrederikBolding (approved):

chaitanyapotti (approved):

PR #9598, opened 2026-07-22
feat: Integrate SubscriptionController into wallet initialization

tuna1207:
## Explanation

Adds SubscriptionController as a default initialized unit in @metamask/wallet, with injectable or default SubscriptionService construction and AuthenticationController messenger delegation.

@metamask/wallet is the shared controller-integration layer for extension, mobile, and other clients. Build-specific and runtime values are injected via instanceOptions rather than hardcoded.

Breaking changes
- BREAKING: instanceOptions.subscriptionController is now required with at least env and fetchFunction.
- When passing a full subscriptionService override, env and fetchFunction remain required at the type level but are unused for service construction.

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking constructor and integration contracts for subscription billing/auth flows, plus a large architectural shift to messenger-delegated HTTP and shared wallet initialization that all clients must adopt correctly.
> 
> **Overview**
> **BREAKING:** Subscription API access moves behind a messenger-backed **`SubscriptionService`** that extends **`BaseDataService`**, authenticates via **`AuthenticationController:getBearerToken`** / **`getSessionProfile`**, validates responses with **superstruct**, and exposes endpoints as **`SubscriptionService:*`** actions (with retries and profile-scoped query keys). **`SubscriptionController`** no longer accepts or constructs a service instance; it only calls those messenger actions. Removed types/options include **`SubscriptionServiceConfig`**, **`ISubscriptionService`**, and the controller’s **`subscriptionService`** constructor option, plus exported **`AllowedEvents`** / **`AuthenticationController:stateChange`** on the controller messenger.
> 
> **`@metamask/wallet`** now initializes **`SubscriptionService`** and **`SubscriptionController`** by default, with optional **`instanceOptions.subscriptionService`** / **`subscriptionController`** and messenger delegation to auth and service actions. Tests, exports, CODEOWNERS, and dependency graphs are updated accordingly; wallet-cli changelog documents daemon **`subscriptionService`** options (production env + **`fetch`**), with authenticated calls still depending on host **`AuthenticationController`** wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e9301bcce99374880f3401fb8b0aea78a7b8382. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pedronfigueiredo (dismissed):

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
```diff
@@ -125,6 +126,10 @@ function getInstanceOptions(): WalletOptions['instanceOptions'] {
       storage: new InMemoryStorageAdapter(),
     },
     remoteFeatureFlagController: REMOTE_FEATURE_FLAG_OPTIONS,
+    subscriptionController: {
```
pedronfigueiredo: clean up?

packages/wallet/src/initialization/instances/gas-fee-controller/gas-fee-controller.test.ts
```diff
@@ -217,6 +218,10 @@ function getInstanceOptions(): WalletOptions['instanceOptions'] {
       storage: new InMemoryStorageAdapter(),
     },
     remoteFeatureFlagController: REMOTE_FEATURE_FLAG_OPTIONS,
+    subscriptionController: {
```
pedronfigueiredo: clean up?

tuna1207: subscriptionController option is still required because of required `fetchFunction: globalThis.fetch,`, do you think we should default this option ?

pedronfigueiredo: this is required for the test to pass, even though there's no corresponding change to the controller init file?

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -93,16 +95,54 @@ export type SubscriptionControllerStateChangeEvent = ControllerStateChangeEvent<
...
 
-export type AllowedEvents =
-  AuthenticationController.AuthenticationControllerStateChangeEvent;
-
 // Messenger
 export type SubscriptionControllerMessenger = Messenger<
   typeof controllerName,
   SubscriptionControllerActions | AllowedActions,
-  SubscriptionControllerEvents | AllowedEvents
+  SubscriptionControllerEvents
```
mcmire: Can we keep this type but set it to `never`? This way in the future if we need to specify external actions there is a "slot" for it.

``` suggestion
type AllowedEvents = never;

// Messenger
export type SubscriptionControllerMessenger = Messenger<
  typeof controllerName,
  SubscriptionControllerActions | AllowedActions,
  SubscriptionControllerEvents | AllowedEvents

File: packages/subscription-controller/src/SubscriptionController.ts [hunk 4]
```diff
@@ -251,13 +287,40 @@ export class SubscriptionController extends StaticIntervalPollingController()<
     });
 
     this.setIntervalLength(pollingInterval);
-    this.#subscriptionService = subscriptionService;
+    this.#subscriptionService = this.#resolveSubscriptionService(options);
```
mcmire: FYI, the way that we designed services, they are not supposed to be instantiated inside of controllers, they are supposed to be instantiated separately and then consumed by the controller via the messenger (the service gets its own messenger). This is okay for now to get this PR through, but we should consider making a change here in the future.

packages/subscription-controller/src/SubscriptionController.ts
```diff
@@ -251,13 +287,40 @@ export class SubscriptionController extends StaticIntervalPollingController()<
...
+    options: SubscriptionControllerServiceOptions,
+  ): ISubscriptionService {
+    if (options.subscriptionService) {
+      return options.subscriptionService;
+    }
+
+    const getAccessToken =
+      options.getAccessToken ??
+      (async (): Promise<string> =>
+        await this.messenger.call('AuthenticationController:getBearerToken'));
```
mcmire: Similar as above, it would be better if the service made this call itself rather than taking an `accessToken`. This is the pattern that we use in other services, e.g., [ChompApiService](https://github.com/MetaMask/core/blob/62ddf4015039b4d34e3bb3273e2eeb3f2fa00433/packages/chomp-api-service/src/chomp-api-service.ts#L351). Again perhaps something we could take care of in another PR.

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
mcmire: This is here because this test file seems to test the Wallet constructor. I believe this is the wrong place for this test, it should be in `Wallet.test.ts` instead. Moving that test means we could remove this test helper. I have a task to clean up all test files, so I'll make sure to take care of this then.

eslint-suppressions.json
```diff
@@ -2187,19 +2187,14 @@
   },
   "packages/subscription-controller/src/SubscriptionController.test.ts": {
     "no-restricted-syntax": {
-      "count": 2
+      "count": 1
```
mcmire, line 2187: Nice! :tada:

packages/subscription-controller/CHANGELOG.md
```diff
@@ -7,10 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - The new options shape is exported as `SubscriptionControllerServiceOptions`.
+
 ### Changed
 
 - Bump `@metamask/transaction-controller` from `^69.0.0` to `^69.3.0` ([#9568](https://github.com/MetaMask/core/pull/9568), [#9589](https://github.com/MetaMask/core/pull/9589), [#9593](https://github.com/MetaMask/core/pull/9593), [#9693](https://github.com/MetaMask/core/pull/9693))
 
+### Removed
+
+- **BREAKING:** Remove the `AuthenticationController:stateChange` event from `SubscriptionControllerMessenger`, along with the `AllowedEvents` type export ([#9598](https://github.com/MetaMask/core/pull/9598))
+  - The controller never subscribed to this event, so it no longer needs to be delegated to the controller's messenger.
```
mcmire: Good call on removing `AllowedActions` and `AllowedEvents`. Per convention those shouldn't have been exported anyway. We might want to move that fact to its own entry:

``` suggestion
- **BREAKING:** Remove the `AuthenticationController:stateChange` event from `SubscriptionControllerMessenger` ([#9598](https://github.com/MetaMask/core/pull/9598))
  - The controller never subscribed to this event, so it no longer needs to be delegated to the controller's messenger.
- **BREAKING:** Remove `AllowedActions` and `AllowedEvents` exports ([#9598](https://github.com/MetaMask/core/pull/9598))

File: packages/wallet/src/initialization/instances/gas-fee-controller/gas-fee-controller.test.ts (hunk 2 above)
Comment (mcmire) [reply packages/wallet/src/initialization/instances/gas-fee-controller/gas-fee-controller.test.ts]: This is similar to the tests for TransactionController — this test file mistakenly verifies that Wallet initializes GasFeeController. This should be done in `Wallet.test.ts` instead. I can clean this up later.

**Review by mcmire — COMMENTED** [review]
Nicely done, I just had a suggestion about the changelog, the rest are notes / suggestions for the future.

**Comment by tuna1207** [comment]
@mcmire @pedronfigueiredo i have refactored subscription service to its own separate data service and resolve all the comments, please help me review again, thank you

**Review by pedronfigueiredo — DISMISSED** [review]

**Comment by tuna1207** [comment]
@metamaskbot publish-previews

File: packages/subscription-controller/src/SubscriptionService.ts [hunk 8]
```diff
@@ -350,11 +606,17 @@ export class SubscriptionService implements ISubscriptionService {
...
+    profileKey: string;
+    bearerToken: string;
+  }> {
     try {
-      const accessToken = await this.authUtils.getAccessToken();
-      return { Authorization: `Bearer ${accessToken}` };
+      const bearerToken = await this.messenger.call(
+        'AuthenticationController:getBearerToken',
+      );
+      const profileKey = bytesToHex(await sha256(stringToBytes(bearerToken)));
```
chaitanyapotti: pls do the same as claims pr

pedronfigueiredo (dismissed):

packages/subscription-controller/src/SubscriptionService.ts
tuna1207: updated in https://github.com/MetaMask/core/pull/9598/commits/4995403946557c654f11415b35c616fbee4256e6

pedronfigueiredo (dismissed):

chaitanyapotti (dismissed):

sirtimid (dismissed):

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

sirtimid (dismissed):

packages/wallet/src/types.ts
```diff
@@ -35,6 +37,8 @@ export type InstanceSpecificOptions = {
   networkController: NetworkControllerInstanceOptions;
   remoteFeatureFlagController: RemoteFeatureFlagControllerInstanceOptions;
   storageService: StorageServiceInstanceOptions;
+  subscriptionService: SubscriptionServiceInstanceOptions;
```
FrederikBolding: Could `subscriptionService` have a default for `fetchFunction` set to `globalThis.fetch` so we don't need to require it?

packages/wallet/src/initialization/instances/subscription-service/subscription-service.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const subscriptionService: InitializationConfiguration<
+  SubscriptionService,
+  SubscriptionServiceMessenger
+> = {
+  name: 'SubscriptionService',
+  init: ({ messenger, options }) =>
+    new SubscriptionService({
+      messenger,
+      ...options,
```
FrederikBolding:
```suggestion
      ...options,
      messenger,
```
Makes sure we cannot overwrite the `messenger`.

packages/wallet/src/initialization/instances/subscription-controller/subscription-controller.ts
```diff
@@ -0,0 +1,53 @@
...
+export const subscriptionController: InitializationConfiguration<
+  SubscriptionController,
+  SubscriptionControllerMessenger
+> = {
+  name: 'SubscriptionController',
+  init: ({ state, messenger, options }) =>
+    new SubscriptionController({
+      messenger,
+      state,
+      ...options,
```
FrederikBolding:
```suggestion
      ...options,
      messenger,
      state,
```

packages/wallet/src/types.ts
tuna1207: It's a convention for our core service, fetch function should be required `so that we don't have to rely on a particular JavaScript runtime or environment where a global `fetch` function may not exist`
https://github.com/MetaMask/core/blob/9f0c1a4cbfd5016518128e89b1f2e8dcb8048301/AGENTS.md#L212
https://github.com/MetaMask/core/blob/9f0c1a4cbfd5016518128e89b1f2e8dcb8048301/docs/code-guidelines/data-services.md#L27

packages/wallet/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING:** Wire subscription into the default wallet initialization ([#9598](https://github.com/MetaMask/core/pull/9598))
```
FrederikBolding:
```suggestion
- **BREAKING:** Wire `SubscriptionController` and `SubscriptionService` into the default wallet initialization ([#9598](https://github.com/MetaMask/core/pull/9598))
```
nit

packages/wallet/src/types.ts
FrederikBolding: It makes perfect sense to allow overriding it, I'm just curious if it couldn't default to `globalThis.fetch` for environments where that is supported (which is most environments at this point).

tuna1207: Agree, i think that make sense since most env support it nowadays

packages/wallet-cli/src/daemon/wallet-factory.ts
```diff
@@ -129,6 +132,9 @@ function buildInstanceOptions(
     storageService: {
       storage: new InMemoryStorageAdapter(),
     },
+    subscriptionService: {
+      fetchFunction: globalThis.fetch,
+    },
```
FrederikBolding: If this is now the default it can be omitted

FrederikBolding (dismissed):

chaitanyapotti (approved):

FrederikBolding (approved):

pedronfigueiredo (approved):

PR #9602, opened 2026-07-22
ci: scope changelog RC validation to released packages

cryptodev-2s:
## Explanation

#9559 fixed release branch detection so `--rc` actually runs, but that surfaced a second problem and the PR was reverted in #9590.

A release only bumps a subset of packages, while `changelog:validate` runs across every workspace (the CI matrix validates all package names). The `--rc` check requires an empty `[Unreleased]` section, so applying it to every package on a release branch made all the packages that legitimately carry unreleased entries fail, even though they are not part of the release. On a real release branch only a couple of packages get a new version, so almost everything failed.

This scopes `--rc` per package: it runs only when the package version was bumped on this branch. "Bumped on this branch" is decided by comparing the current `package.json` version against the version at the merge base with the base branch, so it stays correct even as the base branch moves on. Packages that are not part of the release validate without `--rc`, exactly as before.

Other details:

* The corrected release detection from #9559 (`GITHUB_HEAD_REF` / `GITHUB_REF_NAME` plus an unquoted `^release/` pattern) is kept.
* When the base version cannot be determined, we default to skipping `--rc` so an undetermined base never blocks a release PR.
* `0.0.0` (unpublished) packages never get `--rc`, since there is no released version to match a heading against.

## Verification

Checked against a real release branch (`release/1136.0.0`, which bumps only `transaction-controller` and `transaction-pay-controller`):

* Before: `yarn changelog:validate` failed for 72 packages with `Unreleased changes present in the changelog`.
* After: passes. Only the two bumped packages get `--rc`; injecting an `[Unreleased]` entry into a bumped package still fails validation, so RC enforcement is intact.

## References

* Reverted PR: #9559
* Revert: #9590

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate (tests added in a stacked PR)
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate (inline comments)
- [ ] I've communicated my changes to consumers by updating changelogs (n/a, internal CI script, not consumer facing)
- [ ] I've introduced breaking changes (n/a)

---

> [!NOTE]
> **Low Risk**
> CI and internal release tooling only; behavior is narrower and safer for release PRs, with tests covering the decision logic.
> 
> **Overview**
> **Changelog `--rc` validation** on release branches now runs only for packages whose version changed on that branch (current `package.json` vs version at the merge base with the base ref), instead of every workspace when the branch looks like a release. Non-release packages keep normal validation so legitimate `[Unreleased]` entries no longer fail release CI.
> 
> `validate-changelog.sh` is refactored into testable helpers (`resolve_branch_name`, `resolve_base_version`, `should_validate_as_release_candidate`) with CI branch detection via `GITHUB_HEAD_REF` / `GITHUB_REF_NAME`. The validate-changelog workflow uses **`fetch-depth: 0`** so merge-base resolution works.
> 
> **Shell tests** are added with **bats** (`scripts/validate-changelog.bats`), wired through `test:scripts:shell` in root `test` and the Test Scripts CI job; knip ignores the non-standard bats binary path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15dec5b1b2a42d2aa2da7de43f9bc7e2ff74063a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


package.json
```diff
@@ -49,10 +49,11 @@
     "setup": "yarn install",
     "skills": "metamask-skills sync",
     "skills:postinstall": "metamask-skills postinstall",
-    "test": "yarn test:scripts --silent --collectCoverage=false --reporters=jest-silent-reporter && yarn test:packages",
+    "test": "yarn test:scripts --silent --collectCoverage=false --reporters=jest-silent-reporter && yarn test:scripts:shell && yarn test:packages",
     "test:clean": "yarn workspaces foreach --all --parallel --verbose run test:clean && yarn test",
     "test:packages": "yarn test:verbose --silent --collectCoverage=false --reporters=jest-silent-reporter",
     "test:scripts": "NODE_OPTIONS=--experimental-vm-modules yarn jest --config ./jest.config.scripts.js --silent",
+    "test:scripts:shell": "node_modules/bats/bin/bats scripts/*.bats",
```
Gudahtt, line 56: Hmm. It's unfortunate that we need to hard-code the path like this. I didn't realize Yarn didn't symlink bash script bin entries. Oh well.

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+  fi
+}
 
-if [[ "${GITHUB_REF:-}" =~ '^release/' ]]; then
-  yarn auto-changelog validate --formatter oxfmt --tag-prefix "${package_name}@" --rc "$@"
-else
-  yarn auto-changelog validate --formatter oxfmt --tag-prefix "${package_name}@" "$@"
+# Only run when executed directly, so tests can source the functions above.
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+  set -euo pipefail
```
Gudahtt, line 103: Nit: Why move the `set` call below as well?

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+# the base cannot be determined (e.g. an unrelated history or a base branch that
+# isn't available).
+#
+# Comparing against the merge base rather than the base branch tip keeps the
+# result correct even as the base branch moves on.
+resolve_base_version() {
+  local base_ref="${CHANGELOG_BASE_REF:-}"
+  if [[ -z "${base_ref}" ]]; then
+    local candidate
+    for candidate in "origin/main" "main"; do
```
Gudahtt: Why the fallback to `main` here? I thought at first that this was meant for cases where the remove isn't named `origin`, but we assume that later on in the script, so I'm not sure.

If the user's remote isn't called `origin`, then using this fallback might actually make things worse (by using an outdated `main`, and giving the wrong answer rather than failing).

Maybe it would be simpler to remove this loop, and just check `origin/main`? Unless there is some other purpose I am missing.

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+        break
+      fi
+    done
+    if [[ -z "${base_ref}" ]] && git fetch --quiet --depth=1 origin main 2>/dev/null; then
+      base_ref="FETCH_HEAD"
+    fi
+  fi
+
+  local base_commit=""
+  if [[ -n "${base_ref}" ]]; then
```
Gudahtt: What happens when this isn't set? It looks like we'd just emit nothing from this function.

I don't think that's what we want here. Maybe better to return an error or something? Or if it is what we want, and you'd prefer this error handled later for some reason, we could at least make this a bit more readable by explicitly doing an early exit here

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+    base_commit="$(git merge-base "${base_ref}" HEAD 2>/dev/null || true)"
+    # Deepen the history if a shallow CI checkout hides the common ancestor.
```
Gudahtt: It looks like this case is guaranteed to be hit in ci, because we only checkout with a depth of 1. Perhaps we should increase the fetch depth for this job in `lint-build-test`? Otherwise it will fail once first each time

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+    if [[ -z "${base_commit}" ]]; then
+      git fetch --quiet --deepen=100 origin main 2>/dev/null || git fetch --quiet --unshallow origin 2>/dev/null || true
```
Gudahtt: What is the purpose of the second `git fetch` here? i.e. when is the first expected to fail?

scripts/validate-changelog.sh
```diff
@@ -1,17 +1,114 @@
...
+      base_commit="$(git merge-base "${base_ref}" HEAD 2>/dev/null || true)"
+    fi
+  fi
+
+  if [[ -n "${base_commit}" ]]; then
+    git show "${base_commit}:./package.json" 2>/dev/null | node -e "const fs=require('fs');try{process.stdout.write(String(JSON.parse(fs.readFileSync(0,'utf8')).version||''))}catch{process.stdout.write('')}" || true
```
Gudahtt: There is no need for us to run minified inline scripts. This would be more readable using `jq`, e.g. `jq -r .version`

package.json
cryptodev-2s: Yeah. The `bats` npm package ships a directory style `bin`, so Yarn creates no `.bin/bats` shim and it has to be invoked by path. Left a knip note so the dependency checks stay happy.

scripts/validate-changelog.sh
cryptodev-2s: It's so the bats tests can `source` this file to call the functions without turning on errexit in the test shell. If `set` ran at the top of the file, sourcing would flip it on globally and interfere with bats. Added a comment saying so in 15dec5b1b2.

scripts/validate-changelog.sh
cryptodev-2s: Good call. Dropped the loop, it uses `origin/main` directly now (still overridable via `CHANGELOG_BASE_REF` for the tests). 15dec5b1b2

scripts/validate-changelog.sh
cryptodev-2s: Reworked it: it now fails loudly with a message when no merge base is found, instead of emitting nothing. An empty result now only means the package did not exist at the base (a newly added package), which is handled explicitly. 15dec5b1b2

scripts/validate-changelog.sh
cryptodev-2s: Right. Set `fetch-depth: 0` on the Validate changelog job so the merge base is always present, and removed the fallback. 15dec5b1b2

scripts/validate-changelog.sh
cryptodev-2s: That was the deepen fallback for the shallow checkout, which is exactly the fragile bit you flagged just above. It is gone now that the job fetches full history. 15dec5b1b2

scripts/validate-changelog.sh
cryptodev-2s: Switched to `jq --raw-output '.version'` here, and for the current version read too. 15dec5b1b2

Gudahtt (approved):
LGTM!

PR #9615, opened 2026-07-22
feat: KYC controller

Akaryatrh:
# `@metamask/kyc-controller`

## Overview

`@metamask/kyc-controller` is a shared, **platform-agnostic KYC / identity-verification** package for MetaMask clients (mobile, extension, web). It centralizes all KYC orchestration logic in the `core` monorepo so every client consumes one implementation instead of maintaining its own.

### Core modules

- **`KycController`** (`src/KycController.ts`) — A `BaseController` that owns the entire identity flow:
  - The **flow state machine**: `idle → terms → session → check → auth → form → submit → done / error`.
  - The **Check/Auth frame message protocol** for communicating with MoonPay's hosted frames (`blocks.moonpay.com`) over dedicated channels (`ch_1`, `ch_2`, `ch_reset`).
  - **X25519 credential decryption** and key-pair generation.
  - **SumSub orchestration** through an injected `KycSumSubLauncher` adapter, keeping the controller SDK-free while each client injects its own launcher.
  - **Product-scoped auto-continuation**: an optional `product` (`ramps` | `card`) passed to `initialize` / `acceptTermsAndStartSession` is stored as `activeProduct`. Once authentication reaches the `form` phase, the controller **automatically** runs the KYC-required check and, when KYC is required, chains into the SumSub sub-flow — with no extra consumer calls. Without a `product`, the flow stops at `form` for manual control (`checkKycRequired` / `startSumSub`).
  - A per-product **"is KYC required" cache** and security-aware state metadata (session/access tokens are marked non-persisted and non-logged).

- **`KycService`** (`src/KycService.ts`) — A data service that performs the **Universal KYC (UKYC) HTTP calls** via an injected `fetch`. It wraps requests with `createServicePolicy` for resilience, validates responses with `superstruct`, and sources the auth bearer token (from `profile-sync-controller`) and geolocation (from `geolocation-controller`) through the messenger. It also accepts an optional `baseUrl` to override the `env`-derived URL for local/staging APIs.

### Supporting modules

- **`crypto.ts`** — X25519 key-pair generation and encrypted-credential decryption.
- **`countryCodes.ts`** — ISO 3166-1 alpha-2 → alpha-3 country-code mapping.
- **`selectors.ts`** — Memoized state selectors (`selectKycPhase`, `selectKycSumSub`, `selectIsKycRequiredForProduct`).
- **`types.ts`** — Shared vendor-neutral types (`KycPhase`, `KycProduct`, `KycVendor`, `KycSumSubLauncher`, `KycDisclaimer`, etc.).
- **`KycController-method-action-types.ts` / `KycService-method-action-types.ts`** — Messenger action type definitions for the controller and service.
- **`index.ts`** — Explicit (non-barrel) exports defining the package's public API.

### Tests

Comprehensive Jest test suites covering the controller, service, crypto, country codes, selectors, and the index/public API (`KycController.ts` at 100% coverage, including the new auto-continuation paths).

### Package scaffolding & monorepo wiring

- Standard package files: `package.json`, `tsconfig.json`, `tsconfig.build.json`, `jest.config.js`, `typedoc.json`, `CHANGELOG.md`, `README.md`, and the `LICENSE`, `LICENSE.APACHE2`, and `LICENSE.MIT` files.
- Registered in the root `tsconfig.json` / `tsconfig.build.json` project references, added to the root `README.md` package list, and reflected in `yarn.lock`.

## Public API

The package explicitly exports:

- **Controller:** `KycController`, `getDefaultKycControllerState`, `controllerName`, plus its messenger, options, state, action, and event types. `initialize` / `acceptTermsAndStartSession` accept an optional `product` to enable the automatic post-authentication continuation.
- **Service:** `KycService`, `serviceName`, plus its messenger, options, params, response, action, and event types.
- **Selectors:** `selectIsKycRequiredForProduct`, `selectKycPhase`, `selectKycSumSub`.
- **Utilities:** `alpha2ToAlpha3`, `ALPHA2_TO_ALPHA3`, `decryptCredentials`, `generateKeyPair`, and related crypto/domain types.

## Note

The SumSub sub-flow currently uses placeholders (`MOCK_JWT_TOKEN` and a hardcoded `vendorUserId`), which must be replaced with real UKYC-issued credentials before production use.

## Checklist

---

> [!NOTE]
> **High Risk**
> New identity/KYC surface handling access tokens, encryption keys, JWT attestation, and user storage; incorrect flow guards or crypto could block verification or mishandle secrets, and mock JWT remains for production.
> 
> **Overview**
> Adds **`@metamask/kyc-controller`**, a new shared package that orchestrates vendor-neutral KYC for **`ramps`** and **`card`**: **`KycController`** runs the phase machine (terms → MoonPay Check/Auth frames → KYC-required → SumSub), decrypts frame credentials (X25519/AES-GCM), and **`KycService`** calls the Universal KYC API with bearer auth, geolocation, retries, and **superstruct** validation.
> 
> **Product-scoped flows** store `activeProduct` and, after authentication, automatically run the KYC-required check and chain into document verification when needed; **`reset()`** and generation/phase guards prevent stale async work and late frame messages from corrupting state. The SumSub path adds UKYC wrapping-key verification (JWKS **`jwtChain`**), **`local_user_secret`** via **User Storage**, read-only **`ukyc_capability_token`**, SDK launch through an injected **`KycSumSubLauncher`**, and post-SDK session-status polling.
> 
> Monorepo updates assign **Universal KYC** ownership, wire the package into the dependency graph, relax ESLint for the dev **`mint:ukyc-token`** script, and document architecture in **`ARCHITECTURE.md`**. **Note:** **`startSumSub`** still uses a **`MOCK_JWT_TOKEN`** placeholder until real UKYC credentials are wired.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a29cbb5de404eaac3371494b4bc3bae0cd78431. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,418 @@
...
+  ): Promise<UkycSessionResponse> {
+    const url = new URL('/sessions', this.#baseUrl);
+    const data = await this.#request(url, {
+      method: 'POST',
+      body: JSON.stringify({
+        vendorId: 'moonpay',
+        vendorUserId: 'mockedId',
+        jwtToken: params.jwtToken,
+        vendorMetadata: params.vendorMetadata,
+      }),
```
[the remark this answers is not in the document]
Akaryatrh: On purpose (temporary)

packages/kyc-controller/src/countryCodes.ts
```diff
@@ -0,0 +1,270 @@
+/**
+ * ISO 3166-1 alpha-2 to alpha-3 country code mapping.
+ *
+ * The geolocation source returns ISO 3166-2 codes whose leading segment is an
+ * alpha-2 country code (e.g. "US", "US-NY"). The identity vendor APIs
+ * (disclaimers, kyc-required) expect alpha-3 codes (e.g. "USA"). This map
+ * bridges the two.
+ */
+export const ALPHA2_TO_ALPHA3: Record<string, string> = {
```
jiexi, line 9: thoughts on this living in a different package or in the client for now?

I think the KYC Controller interface should expect the country code to be in 3 character format

Akaryatrh:
@SocketSecurity ignore npm/yargs@17.7.3
@SocketSecurity ignore npm/chokidar-cli@3.0.0
@SocketSecurity ignore npm/babel-plugin-istanbul@6.1.1
@SocketSecurity ignore npm/create-jest@29.7.0
@SocketSecurity ignore npm/istanbul-lib-source-maps@4.0.1

Akaryatrh:
@metamaskbot publish-preview

jiexi (dismissed):

packages/kyc-controller/src/ukyc/index.ts
mcmire, line 1: Do we need this file? We generally discourage using "barrel" files as it creates more ways that a file can be found and thus has the potential to slow down static analysis tools like ESLint. We have a guideline about this for packages here: https://github.com/MetaMask/core/blob/main/docs/code-guidelines/package-guidelines.md#list-exports-explicitly

packages/kyc-controller/src/crypto.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+/**
+ * Decode a base64 / base64url string to bytes without relying on `atob` or
+ * `Buffer`.
+ *
+ * @param value - The (possibly url-safe, possibly unpadded) base64 string.
+ * @returns The decoded bytes.
+ */
+function base64ToBytes(value: string): Uint8Array {
```
mcmire: Have you seen `@metamask/utils`? We have a lot of utility functions related to base64, binary, etc.: https://metamask.github.io/utils/latest/. For instance here's its version of `base64ToBytes`: https://github.com/MetaMask/utils/blob/6b14e89/src/bytes.ts#L341

packages/kyc-controller/src/index.test.ts
mcmire, line 1: Some packages have this file, but it's not needed anymore and you should be able to remove it. We should be able to expect `index.ts` in the root lists _all_ of the package's exports. Again see https://github.com/MetaMask/core/blob/main/docs/code-guidelines/package-guidelines.md#list-exports-explicitly.

packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,613 @@
...
+
+// === SERVICE DEFINITION ===
+
+/**
+ * `KycService` communicates with the Universal KYC (UKYC) backend to drive the
+ * identity + document-verification flow. It is stateless and platform-agnostic:
+ * HTTP is performed through an injected `fetch`, and the auth bearer token and
+ * geolocation come from other controllers via the messenger.
+ */
+export class KycService {
```
mcmire: Have you seen `BaseDataService`? This is a relatively new thing and I realize that we need to update our [tutorial on how to write a data service](https://github.com/MetaMask/core/blob/66c3cff4fcbd85634676027b9630700ef326b45f/docs/code-guidelines/data-services.md), but it does a lot of what you're doing already. You can read some documentation on this here: https://app.notion.com/p/metamask-consensys/Data-Services-Pattern-330f86d67d688059b5dfe2eccd98a1aa. And there are some examples linked in that document as well.

packages/kyc-controller/src/ukyc/storageAccessToken.ts
```diff
@@ -0,0 +1,247 @@
...
+  presenter?: UkycTokenPresenter;
+  /** UKYC session id. Required when presenter is `idos-relay`. */
+  sessionId?: string;
+  /** Token issue time. Defaults to now. */
+  issuedAt?: Date;
+  /** Token expiry. Must be strictly after `issuedAt`. */
+  expiresAt: Date;
+};
+
+type JsonValue =
```
mcmire, line 82: Have you seen the `Json` type from `@metamask/utils`? I think what you're doing is the same thing (except it doesn't support `undefined`, which isn't valid JSON anyway): https://github.com/MetaMask/utils/blob/6b14e89/src/json.ts#L36

packages/kyc-controller/src/KycController.ts
```diff
@@ -0,0 +1,1433 @@
...
+   * typing the callback parameter as the plain {@link KycControllerState}
+   * instead of Immer's `Draft`, we avoid expanding the draft type while keeping
+   * the same mutate-in-place semantics (the underlying value is still the Immer
+   * draft at runtime).
+   *
+   * @param updater - The state mutation to apply.
+   */
+  #applyUpdate(updater: (state: KycControllerState) => void): void {
+    this.update((state) => {
+      updater(state as unknown as KycControllerState);
```
mcmire: Hmm, this is too bad. We've definitely encountered this before with other controllers but never worked out a solution. 

Instead of using a type assertion here though can we use `// @ts-expect-error`? Then if this problem mysteriously disappears (which sometimes happens) we know why.

``` suggestion
      // @ts-expect-error Avoid "type instantiation is excessively deep".
      updater(state);

File: packages/kyc-controller/src/KycService.ts [hunk 7]
```diff
@@ -0,0 +1,613 @@
...
+
+/**
+ * Options for constructing a {@link KycService}.
+ */
+export type KycServiceOptions = {
+  messenger: KycServiceMessenger;
+  fetch: typeof fetch;
+  env: KycServiceEnvironment;
+  /**
+   * Overrides the base URL derived from `env`. When provided, this value is
```
mcmire: What are your thoughts on always accepting a `baseUrl` instead of `env` and defining the API URLs at the client level? Both clients have a `builds.yml` file which allows us to define which URL maps to which environment. 

(This is a recent rule that we've added which we haven't applied everywhere yet.)

packages/kyc-controller/README.md
mcmire, line 1: FYI, it's best to keep this README short and sweet for people to read. Minimally the README should explain the purpose of this controller and offer guidance on how it's intended to be used. If you need something for agents, maybe try adding it to AGENTS.md? If you need to explain the architecture maybe try ARCHITECTURE.md?

mcmire (commented):
Hey there, I know you are going to split this PR up into two PRs, but while this was on my mind I decided to take a look at this and had some suggestions. Let me know if you have any questions.

Akaryatrh:
Thanks a lot for review @mcmire!

packages/kyc-controller/src/KycController.ts
Akaryatrh: Apparently the issue only happens at the IDE level. So i removed the type cast and build is not failing 🤷 
And i can't use @ts-expect-error as the build is then failing, complaining that the rule is usuless.

I added a comment that explains this issue.

packages/kyc-controller/src/ukyc/index.ts
Akaryatrh: @mcmire Removed barrel file and also remove useless export, this should be cleaner now 🙂

packages/kyc-controller/src/crypto.ts
Akaryatrh: Function removed! Instead we now use another util (encoding.ts file) that is using Metamask utils package.

packages/kyc-controller/src/KycService.ts
Akaryatrh: KYC service refactored to match this new pattern 👍

packages/kyc-controller/src/ukyc/storageAccessToken.ts
Akaryatrh: Fixed 👍

packages/kyc-controller/src/KycService.ts
Akaryatrh: Done 👍

packages/kyc-controller/README.md
Akaryatrh: Done, I created a dedicated architecture doc, thx for proposal 👍

packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,704 @@
...
+    queryClientConfig = {},
+    policyOptions = {},
+  }: KycServiceOptions) {
+    super({
+      name: serviceName,
+      messenger,
+      queryClientConfig,
+      policyOptions,
+    });
+    this.#fetch = fetchFunction;
```
mcmire: FYI, since introducing `BaseDataService`, we have dropped `fetch` as a parameter, since it is available natively on all modern runtimes (Node 18 included). What do you think about removing this? 

``` suggestion
```

packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,704 @@
...
+    if (!baseUrl) {
+      throw new Error('KycService: baseUrl is required');
+    }
```
mcmire: Do you need this runtime check? We tend to rely on TypeScript to ensure that required arguments are given.

``` suggestion
```

packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,704 @@
...
+  messenger: KycServiceMessenger;
+  fetch: typeof fetch;
+  /**
+   * Mandatory value that sets the base url to KYC api
+   */
+  baseUrl: string;
+  /**
+   * Base URL of the Fractal encryption service, from which the JWKS used to
+   * verify the `jwtChain` returned by {@link KycService.getWrappingKey} is
+   * fetched. Required to run the wrapping-key exchange in
```
mcmire: Should this argument always be required? Otherwise it sounds like leaving it out would make the service halfway functional.

packages/kyc-controller/src/KycService.ts
```diff
@@ -0,0 +1,704 @@
...
+    if (authenticated) {
+      const bearerToken = await this.messenger.call(
+        'AuthenticationController:getBearerToken',
+      );
+      assert(bearerToken, string());
+      if (!bearerToken) {
+        throw new Error(
+          'Unable to obtain an authentication bearer token — is the wallet signed in?',
+        );
+      }
```
mcmire: Is this necessary? The type of `AuthenticationController:getBearerToken` ensures that a string is always returned. You shouldn't have to mistrust what `AuthenticationController:getBearerToken` is doing.

``` suggestion
```

mcmire (commented):
A few more minor things I noticed.

Akaryatrh:
@metamaskbot publish-preview

amitabh94:
@metamaskbot publish-preview

packages/kyc-controller/src/KycService.ts
jiexi: https://github.com/MetaMask/core/pull/9615/commits/7a443d138387cec0c5385fd3b07e7515e73b2f80

packages/kyc-controller/src/KycService.ts
jiexi: yes https://github.com/MetaMask/core/pull/9615/commits/2c8b829c806fbbba24c0a2eb8b80ce5c16b2631c

packages/kyc-controller/src/KycService.ts
jiexi: makes sense https://github.com/MetaMask/core/pull/9615/commits/853970cb608bc048bb14df8d3a0c07d7f3250a77

packages/kyc-controller/src/KycService.ts
jiexi: https://github.com/MetaMask/core/pull/9615/commits/2bb59a35110dfa3cc7747b7b110fc4a65bbb162a

packages/kyc-controller/CHANGELOG.md
```diff
@@ -9,6 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `KycController.getCustomerIdentity()` method and the `KycController:getCustomerIdentity` messenger action (plus the exported `KycControllerGetCustomerIdentityAction` and `KycCustomerIdentity` types). Returns the vendor-scoped `{ vendor, id }` for the currently authenticated customer, or `null` before authentication and after `reset()`. Lets consumers (e.g. ramps autoramp creation) attach the vendor customer id to downstream calls without reading the full KYC state, which also holds session/access tokens. The id is session-scoped and never persisted. ([#9852](https://github.com/MetaMask/core/pull/9852))
```
mcmire: Has `kyc-controller` been released yet? We tend to treat the changelog like release notes, so each section lists differences from the previous version. For the first version of a package there isn't a previous version — everything is new — so it's often enough to have a single bullet point in `### Added` that says something like "Initial release of the `@metamask/kyc-controller`" and maybe lists some highlights of the controller (interesting features, etc.) below it. 

This isn't something you have to change now but make sure to change it when you make a release.

mcmire (dismissed):
One nitpick, but it can be addressed later. LGTM.

packages/kyc-controller/CHANGELOG.md
jiexi: sorry for the sloppiness here https://github.com/MetaMask/core/pull/9615/commits/f4afa86066ddfd0b9bf15185c0c0fae52f9d2155

mcmire (approved):
LGTM.

amitabh94 (approved):

amitabh94 (approved):

PR #9616, opened 2026-07-23
feat: shield controller wire in wallet package

tuna1207:
## Explanation

Adds ShieldController as a default initialized unit in @metamask/wallet.

@metamask/wallet is the shared controller-integration layer for extension, mobile, and other clients. Build-specific and runtime values are injected via instanceOptions rather than hardcoded.

Breaking changes
- BREAKING: instanceOptions.shieldController is now required with at least baseUrl and fetchFunction.
- When passing a full backend override, baseUrl and fetchFunction remain required at the type level but are unused for backend construction.
- Hosts must register AuthenticationController and SignatureController on the wallet root messenger.
- Hosts must explicitly call ShieldController:start after wiring - wallet initialization constructs the controller but does not auto-start it.

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking changes to Shield integration touch transaction/signature coverage and logging with bearer-token auth; incorrect messenger wiring or migration from the removed backend API could break protection flows in clients.
> 
> **Overview**
> **Breaking refactor in `@metamask/shield-controller`:** removes `ShieldRemoteBackend`, `ShieldBackend`, and the `backend` constructor option. Remote Shield work moves to a new **`ShieldApiService`** (`BaseDataService`) that exposes `checkCoverage`, `checkSignatureCoverage`, `logSignature`, and `logTransaction` as messenger actions and authenticates via **`AuthenticationController:getBearerToken`**. **`ShieldController`** now invokes those `ShieldApiService:*` actions instead of an injected backend. API base URLs are resolved from a new **`Env`** map (`getShieldApiBaseUrl`).
> 
> **`@metamask/wallet`:** registers default initialization for **`ShieldApiService`** and **`ShieldController`**, with optional `instanceOptions` (`shieldApiService`: `env`, `fetchFunction`, polling/policy; `shieldController`: history limits and `normalizeSignatureRequest`). Messengers delegate auth to the service and Shield API actions plus transaction/signature state-change events to the controller. **Hosts must register `AuthenticationController` (and `SignatureController` for reactive coverage) on the root messenger and call `ShieldController:start` explicitly**—initialization does not auto-start Shield.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23dfba3bbb3a4665337422f8831760375121bacb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet/src/initialization/instances/shield-controller/shield-controller.ts
```diff
@@ -0,0 +1,88 @@
...
+    state,
+    messenger,
+    options,
+  }: {
+    state: Partial<ShieldControllerState> | undefined;
+    messenger: ShieldControllerInitializationMessenger;
+    options: ShieldControllerInstanceOptions;
+  }) => {
+    return new ShieldController({
+      messenger: messenger as unknown as ShieldControllerMessenger,
```
lwin-kyaw: Why need casting here?

packages/wallet/src/initialization/instances/shield-controller/shield-controller.ts
```diff
@@ -0,0 +1,88 @@
...
+      state,
+      backend: resolveShieldBackend(messenger, options),
+      transactionHistoryLimit: options.transactionHistoryLimit,
+      coverageHistoryLimit: options.coverageHistoryLimit,
+      normalizeSignatureRequest: options.normalizeSignatureRequest,
```
lwin-kyaw, line 20: Let's just forward instead of hard assigning each.
```suggestion
      ...options,
```

packages/wallet/src/initialization/instances/shield-controller/types.ts
```diff
@@ -0,0 +1,45 @@
...
+import type {
+  NormalizeSignatureRequestFn,
+  ShieldControllerActions,
+  ShieldControllerEvents,
+  ShieldRemoteBackend,
+} from '@metamask/shield-controller';
+import type { SignatureStateChange } from '@metamask/signature-controller';
+import type { TransactionControllerStateChangeEvent } from '@metamask/transaction-controller';
+
+export type ShieldBackend = Pick<
```
lwin-kyaw: Why do we need to modify this?

I've noticed that this is being used in `ShieldControllerInstanceOptions`. I don't think we will ever provide the custom backend instance here.

packages/wallet/src/initialization/instances/shield-controller/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  handler: (entropySourceId?: string) => Promise<string>;
+};
+
+export type ShieldControllerInitializationMessenger = Messenger<
+  'ShieldController',
+  ShieldControllerActions | AuthenticationControllerGetBearerTokenAction,
+  | ShieldControllerEvents
+  | SignatureStateChange
+  | TransactionControllerStateChangeEvent
+>;
```
lwin-kyaw: Messenger should come from the controller package right? Why?

packages/wallet/src/initialization/instances/shield-controller/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  baseUrl: string;
+  fetchFunction: typeof fetch;
+  getAccessToken?: () => Promise<string>;
+  captureException?: (error: Error) => void;
+  getCoverageResultTimeout?: number;
+  getCoverageResultPollInterval?: number;
+  transactionHistoryLimit?: number;
+  coverageHistoryLimit?: number;
+  normalizeSignatureRequest?: NormalizeSignatureRequestFn;
+};
```
lwin-kyaw: Please use direct type reference from controller, instead of redefining or overrides.
```suggestion
export type ShieldControllerInstanceOptions =
  ConstructorParameters<typeof ShieldController>[0];
```

packages/wallet/src/types.ts
```diff
@@ -38,4 +39,5 @@ export type InstanceSpecificOptions = {
   transactionController?: TransactionControllerInstanceOptions;
   passkeyController?: PasskeyControllerInstanceOptions;
   seedlessOnboardingController?: SeedlessOnboardingControllerInstanceOptions;
+  shieldController: ShieldControllerInstanceOptions;
```
lwin-kyaw: Why `shieldController` is required type here?

lwin-kyaw: Most controllers/services are optional here.

pedronfigueiredo (changes requested):
I see there's a bunch of comments and merge conflicts, can you address them and then re-request my review here on github once it's ready for another look?

packages/wallet/src/types.ts
tuna1207: shieldController is required because it follows the same rule as the other required slots.
Needs baseUrl + fetchFunction, or an injected backend.

initialize() always creates ShieldController (it's in defaultConfigurations). Making the slot required forces every Wallet consumer to explicitly provide Shield connectivity config, similar to gasFeeController.clientId.

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
```diff
@@ -125,6 +125,15 @@ function getInstanceOptions(): WalletOptions['instanceOptions'] {
       storage: new InMemoryStorageAdapter(),
     },
     remoteFeatureFlagController: REMOTE_FEATURE_FLAG_OPTIONS,
+    shieldController: {
```
pedronfigueiredo: this should no longer be needed

tuna1207: shieldController option is still required because of required fetchFunction: globalThis.fetch, do you think we should default this option ?

pedronfigueiredo: Not quite true: fetchFunction is not required when a backend is supplied. The type is a union: either provide backend, or provide fetchFunction so a backend can be created. The test currently provides both, but fetchFunction is ignored because initialization uses:
   ```ts
     options.backend ?? createShieldRemoteBackend(...)
   ```

 So this would be sufficient:

 ```ts
   shieldController: {
     backend: {
       checkCoverage: jest.fn(),
       checkSignatureCoverage: jest.fn(),
       logSignature: jest.fn(),
       logTransaction: jest.fn(),
     },
   },
 ```

packages/shield-controller/src/backend.ts
```diff
@@ -327,6 +328,55 @@ export class ShieldRemoteBackend implements ShieldBackend {
...
+ * Defaults to `Env.PRD`.
+ * @returns The created backend.
+ */
+export function createShieldRemoteBackend({
+  messenger,
+  getAccessToken,
+  env = Env.PRD,
+  ...options
+}: CreateShieldRemoteBackendOptions): ShieldRemoteBackend {
+  return new ShieldRemoteBackend({
```
mcmire: Hmm, why does this package need to export a convenience function to export this class? If we are initializing this class in `@metamask/wallet` then do we need this?

packages/wallet/src/initialization/instances/shield-controller/shield-controller.ts
```diff
@@ -0,0 +1,57 @@
...
+  ShieldControllerMessenger
+> = {
+  name: 'ShieldController',
+  init: ({ state, messenger, options }) =>
+    new ShieldController({
+      messenger,
+      state,
+      backend:
+        options.backend ??
+        createShieldRemoteBackend({
```
mcmire: Similar to the [comment I posted about SubscriptionController](https://github.com/MetaMask/core/pull/9598#discussion_r3685419907), it would be better if ShieldRemoteBackend were a data service class, so that it could access `AuthenticationController:getBearerToken` itself through the messenger (see [ChompApiService](https://github.com/MetaMask/core/blob/62ddf4015039b4d34e3bb3273e2eeb3f2fa00433/packages/chomp-api-service/src/chomp-api-service.ts#L351) for an example. 

Also it looks like `ShieldController` takes `ShieldRemoteBackend` as an argument. If the service had its own messenger, then it would be better if the controller could use it to access the service itself (see [`ClaimsController`](https://github.com/MetaMask/core/blob/62ddf4015039b4d34e3bb3273e2eeb3f2fa00433/packages/claims-controller/src/ClaimsController.ts#L142) and [`ClaimsService`](https://github.com/MetaMask/core/blob/62ddf4015039b4d34e3bb3273e2eeb3f2fa00433/packages/claims-controller/src/ClaimsService.ts).

Again this could be a future change, but I wanted to point this out.

packages/wallet/src/Wallet.test.ts
```diff
@@ -23,23 +28,38 @@ const REMOTE_FEATURE_FLAG_OPTIONS = {
...
-      },
-      networkController: {
-        infuraProjectId: 'fake-infura-project-id',
-      },
-      storageService: {
-        storage: new InMemoryStorageAdapter(),
-      },
-      remoteFeatureFlagController: REMOTE_FEATURE_FLAG_OPTIONS,
-    },
+    instanceOptions: getInstanceOptions(),
```
mcmire, line 51: Instead of extracting `getInstanceOptions`, what do you think about having `setupWallet` take an option that allows for overriding options? Then the tests that use `new Wallet` could use `setupWallet` instead, and pass whatever is relevant to them (e.g. `initializationConfigurations`)?

mcmire (commented):
Nice, just had one immediate comment and one future comment.

packages/wallet/src/initialization/instances/transaction-controller/transaction-controller.test.ts
tuna1207: oh yes, you're correct, that can be removed, updated in https://github.com/MetaMask/core/pull/9616/changes/f1bd0790fa378af31d156d77d05fc68d54d51ae4

packages/shield-controller/src/backend.ts
tuna1207: Oh initially it's intended to keeps the "Shield auth comes from AuthenticationController:getBearerToken" coupling inside the shield package. But i also agree with making ShieldRemoteBackend a data service, will update it in this PR too.

packages/wallet/src/initialization/instances/shield-controller/shield-controller.ts
tuna1207: I have updated to use `ShieldApiService`

tuna1207:
@mcmire @pedronfigueiredo i have refactored shield api service to its own separate data service and resolve all the comments, please help me review again, thank you

pedronfigueiredo (dismissed):

tuna1207:
@metamaskbot publish-previews

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

sirtimid (dismissed):

chaitanyapotti (dismissed):

pedronfigueiredo (dismissed):

pedronfigueiredo (approved):

chaitanyapotti (approved):

FrederikBolding (approved):

PR #9650, opened 2026-07-24
feat(client-utils): map ramps orders into the shared activity item shape

wenfix:
## Explanation

Ramps (buy/sell) orders currently have no representation in the shared, cross-client `ActivityItem` model that Send/Swap/Bridge/etc. already use, so consumers (extension, mobile) have had to maintain bespoke, order-specific UI to show them instead of the generic activity list and details pipeline.

This PR:

- Adds `'rampBuy' | 'rampSell'` to `ActivityKind` and a matching arm to the `ActivityItem` union, covering the order's fiat/token amounts, fee, provider info (id/name/link), status description, and manual-payment details.
- Adds an optional `id` field to the shared `ActivityData` base type. A pending ramp order has an empty `txHash` (no on-chain tx exists yet while awaiting fiat settlement), so consumers need a stable identifier to key list rows / open details before settlement. `id` is populated only for `rampBuy`/`rampSell` (from the provider's order id) and left `undefined` for every other kind — no behavior change for existing consumers.
- Adds `mapRampsOrder`, a pure mapper (following the existing `mapLocalTransaction`/`mapApiTransaction`/`mapKeyringTransaction` pattern) that normalizes a ramps order into an `ActivityItem`, mapping the 8-value `RampsOrderStatus` down to the existing 4-value `Status`.
- The mapper's input type (`RampsOrderLike`) redeclares the narrow subset of `RampsOrder` (from `@metamask/ramps-controller`) it needs, rather than importing it — `client-utils` and `ramps-controller` don't depend on each other today, and this keeps it that way.
- Chain id normalization is chain-agnostic: it reuses the existing `formatChainIdToCaip` helper (already used elsewhere in this package for Solana/Bitcoin/Stellar-aware mappers) rather than hardcoding an EVM namespace, so it passes through an already-CAIP-formatted chain id unchanged and only assumes `eip155` for a bare numeric/hex reference (today's only observed format from the ramps API).

## References

Extension Client PR: https://github.com/MetaMask/metamask-extension/pull/44689

## Checklist

---

> [!NOTE]
> **Low Risk**
> Additive types and a pure mapper in client-utils; no auth or transaction submission changes, though consumers must handle new `ActivityKind` variants and optional ramp `chainId`.
> 
> **Overview**
> Adds **`rampBuy` / `rampSell`** to the shared activity model so ramps orders can use the same list/details pipeline as other activity kinds. Ramp items carry fiat/token amounts, fees, provider metadata, payment instructions, and an optional **`data.id`** when there is no real **`hash`** yet (pending fiat settlement).
> 
> Introduces **`mapRampsOrder`** (plus exported **`RampsOrderLike`**) to normalize ramps controller–shaped orders: status → shared **`Status`**, buy/sell/deposit typing, chain id resolution (network → crypto chain/asset fallbacks), and filtering that returns **`null`** for hidden statuses or **`excludeFromPurchases`**. Placeholder **`txHash`** values are dropped so list keys do not collide.
> 
> **`formatChainIdToCaip`** now treats an empty chain id as **`undefined`** (avoids **`eip155:0`** on precreated stub orders). Changelog updated; tests cover the mapper and CAIP edge case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68ca1fb163f9c60ecaeaa17c9ab99fc0586975d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wenfix:
@metamaskbot publish-previews

packages/client-utils/src/mappers/ramps-order-mapper.ts
```diff
@@ -0,0 +1,120 @@
...
+    },
+  ];
+
+  // `network.chainId` may already be CAIP-2 (non-EVM orders) or a bare
+  // numeric/hex reference (today's only observed EVM format) — normalize
+  // without assuming a namespace.
+  const chainId = formatChainIdToCaip(order.network.chainId) as CaipChainId;
+
+  return {
+    type: order.orderType === 'buy' ? 'rampBuy' : 'rampSell',
```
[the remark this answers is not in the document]
wenfix: Confirmed — checked against `RampsService.test.ts` and `RampsController.test.ts`, every fetched-order fixture uses uppercase `orderType: 'BUY'`/`'SELL'`; the lowercase `'buy'` I copied only appears in the client-side stub order (`RampsController.ts` `addPrecreatedOrder`). Fixed in b2bfbe44b by normalizing with `.toUpperCase()`, with tests covering both the real API's uppercase shape and the stub order's lowercase shape.

georgeweiler (dismissed):

packages/client-utils/src/types.ts
```diff
@@ -79,9 +81,23 @@ type ActivityData<Type extends ActivityKind, Data> = {
   status: Status;
   timestamp: number;
   hash?: string;
+  // Stable identifier for items that may not have a hash yet (e.g. a ramp
+  // order pending fiat settlement, where `hash` is empty until it settles
+  // on-chain).
+  id?: string;
```
n3ps: @wenfix this is the base type common to all activity rows so we want to avoid putting fields here that are not applicable to every kind of activity 

The union below where you added `rampBuy` and `rampSell` is where to put all fields specific to that type

n3ps (changes requested):
Can we move `id` specific to the ramps type?

packages/client-utils/src/mappers/ramps-order-mapper.ts
```diff
@@ -0,0 +1,122 @@
...
+  ];
+
+  // `network.chainId` may already be CAIP-2 (non-EVM orders) or a bare
+  // numeric/hex reference (today's only observed EVM format) — normalize
+  // without assuming a namespace.
+  const chainId = formatChainIdToCaip(order.network.chainId) as CaipChainId;
+
+  return {
+    type: isBuy ? 'rampBuy' : 'rampSell',
+    chainId,
```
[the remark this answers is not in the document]
wenfix: addressed here: d5358537c

packages/client-utils/src/types.ts
wenfix: Great call, addressed here: d5358537c

Sorry there's another change in that commit tackling a bug Cursor had flagged.

packages/client-utils/src/mappers/ramps-order-mapper.ts
```diff
@@ -0,0 +1,121 @@
...
+      token,
+      fees,
+      provider: {
+        id: order.provider?.id,
+        name: order.provider?.name,
+        orderLink: order.providerOrderLink,
+      },
+      statusDescription: order.statusDescription,
+      paymentDetails: order.paymentDetails,
+      id: order.providerOrderId,
```
wachunei: Once again I see this pattern, is this the direction Ramps team is going? is very unlikely for orders to clash ids but using the `order.id` reduces that chance to 0.

packages/client-utils/src/mappers/ramps-order-mapper.ts
```diff
@@ -0,0 +1,121 @@
...
+      return 'success';
+    case 'FAILED':
+    case 'ID_EXPIRED':
+      return 'failed';
+    case 'CANCELLED':
+      return 'cancelled';
+    default:
+      return 'pending';
+  }
+}
```
wachunei, line 86: UNKNOWN and ID_EXPIRED are meant to be invisible in the UI because they represent unsuccessful attempts the user never get to see, is this going to show them in the list?

If user tries to checkout with a provider like paypal or similar, each time they click the button a `PRECREATED` order is generated, imagine user goes back and forth 5 times (editing the amount, for example), then 5 orders are going to be tracked in the background and attempt to match an actual order (status created, pending or superior) — if not, they will be ID_EXPIRED, should get deleted. 

Is this still the pattern the controller uses? will this mapStatus surface them to the UI?

packages/client-utils/src/mappers/ramps-order-mapper.ts
wenfix: Changed it to `order.id`: 228d9f6898ce996191c0f9391aaf5466b959ea7e

packages/client-utils/src/mappers/ramps-order-mapper.ts
wenfix: You're right, good catch — nothing was filtering these out. I checked `RampsController`: `ID_EXPIRED` is in `TERMINAL_ORDER_STATUSES` but that only stops polling, it doesn't remove the order from state, and `removeOrder()` is only ever called externally, never automatically by the controller. So `PRECREATED` stubs that never match, and orders that transition to `ID_EXPIRED`, were staying in state indefinitely with nothing downstream filtering them before this mapper.

Pushed a fix in [642341c](https://github.com/MetaMask/core/commit/642341ca16f39c563b0d833b66c7504813355bcc): `mapRampsOrder` now returns `null` for `UNKNOWN` and `ID_EXPIRED` orders ([`ramps-order-mapper.ts#L61-L73`](https://github.com/MetaMask/core/blob/642341ca16f39c563b0d833b66c7504813355bcc/packages/client-utils/src/mappers/ramps-order-mapper.ts#L61-L73)) so callers can skip them instead of surfacing them as pending/failed activity items. Left `PRECREATED` mapping to `'pending'` since that's the live in-flight attempt itself, not the noise — let me know if you think that one should be hidden too.

wenfix:
@metamaskbot publish-previews

georgeweiler (dismissed):

packages/client-utils/src/types.ts
```diff
@@ -77,15 +79,25 @@ export type Fee = {
   assetType?: AssetType;
 };
 
-type ActivityData<Type extends ActivityKind, Data> = {
+type ActivityData<Type extends ActivityKind, Data, ChainId = CaipChainId> = {
   type: Type;
-  chainId: CaipChainId;
+  chainId: ChainId;
```
n3ps: @wenfix why was this changed? If a ramp type doesn't guarantee a network you can make this optional but you do not need to downgrade to just chainId

wenfix: good call, changed here: 6eb79db79

n3ps:
@wenfix can you scope the changes that are specific to the ramps type within the ramps types?

wenfix:
> @wenfix can you scope the changes that are specific to the ramps type within the ramps types?

Changed here: 4928a522b

wachunei (dismissed):

packages/client-utils/src/mappers/helpers/caip.ts
```diff
@@ -32,6 +32,10 @@ export function formatChainIdToCaip(
     return Number.isNaN(reference) ? undefined : `eip155:${reference}`;
   }
 
+  if (!chainId) {
```
n3ps: chainId is string | number, therefore 0 which is "falsey" is allowed

although there is no chainId 0, it's best to check the actual value instead

something like
```
if (chainId === undefined)
```

packages/client-utils/src/types.ts
```diff
@@ -1,6 +1,8 @@
 import type { ValueTransfer as _ValueTransfer } from '@metamask/core-backend';
 import type { CaipChainId } from '@metamask/utils';
 
+import type { RampOrderPaymentDetail } from './mappers/ramps-order-mapper.js';
```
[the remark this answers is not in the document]
wenfix: addressed here: d01c08674

packages/client-utils/src/mappers/helpers/caip.ts
wenfix: this is fixing the case where, if `chainId` is an empty string, `reference` below would be `0` and the return would be `eip155:0` instead of undefined.

we could change the check to be explicit like: 
`if(chainId === '') return undefined`

would you prefer that?

n3ps: yes checking against empty string makes sense

georgeweiler: [635dd2e](https://github.com/MetaMask/core/pull/9650/commits/635dd2e5f5be3fd2770ea97a2987f5e09e519f5a)

wachunei (dismissed):

packages/client-utils/src/types.ts
```diff
@@ -163,7 +175,37 @@ export type ActivityItem =
...
+      // Precreated stub orders (see `RampsController.addPrecreatedOrder`) may
+      // not have an assigned network yet, so unlike every other activity
+      // kind, a ramp order's chain id isn't guaranteed.
+      chainId?: CaipChainId;
+      // Stable identifier for orders that may not have a hash yet (e.g. a
+      // ramp order pending fiat settlement, where `hash` is empty until it
+      // settles on-chain). Sits next to `hash` since, like `hash`, it's a
+      // cross-kind identity concept — scoped to this union arm only, since no
+      // other activity kind needs it today.
+      id?: string;
```
n3ps: can we move this inside data? since this only applies to ramps

n3ps (approved):

PR #9651, opened 2026-07-24
feat(assets-controller): resolve customAssets in AccountsAPI

Kriys94:
## Explanation

Mobile preview: https://github.com/MetaMask/metamask-mobile/pull/36604
Extension preview: https://github.com/MetaMask/metamask-extension/pull/46451

Architecture walkthrough: [`packages/assets-controller/docs/architecture/v1/CHANGES.md`](packages/assets-controller/docs/architecture/v1/CHANGES.md)

**Onboarding**
- Current logic + v5: https://www.loom.com/share/c55eed16333749e29d99cbda87ca4d9e
- After PR + v6: https://www.loom.com/share/6f443eda5f344d7aa1733c3b5bf7ea4e

**Start / basic flows (change account, network, add/remove custom tokens, add non-supported network)**
- Current logic + v5: https://www.loom.com/share/d5fdd69c1f384c0a97574729773d578f
- After PR + v6: https://www.loom.com/share/4753a4fd68c54a58abed48e728836760

**Transactions**
- Current logic + v5: https://www.loom.com/share/fbc0e58acc9e435694dc6f2f2dfb6d6f
- After PR + v6: https://www.loom.com/share/5b3b1cf5f58f4e6fa31de34f30cc01a9

**Fallback**
- Current logic + v5: https://www.loom.com/share/16f95d2b237e4b63ab0f68e5d2d18cb1
- After PR + v6: https://www.loom.com/share/d34beee512d041b19ccb1869cef5c327
- After PR + v6 + custom tokens: https://www.loom.com/share/11f342e1ac1e4644a695403a536b9780

### Why

Accounts API **v6** can take pinned custom tokens (`includeAssetIds`) and return them in the balance snapshot even at zero balance, and can drop hidden tokens (`excludeAssetIds`). Once the API owns that, the client does not need a parallel custom-token path on the v6 flow.

v5 stays intact. Which path runs is decided only in `AssetsController.#isBalanceV6Enabled()` (`assetsAccountsApiV6 === true`).

### What v6 does

- Subscribe and force refresh assign **accounts + chains only**. Pins and hides are **not** copied onto `DataRequest`.
- `AccountsApiDataSource` reads controller state via a required `getAssetsState` getter:
  - visible pins on fetched chains → `includeAssetIds`
  - hidden assets on fetched chains → `excludeAssetIds`
  - a hide wins over a pin
- `request.customAssets` is only a **scoped override** (`getAssets({ customAssets })`, `addCustomAsset`, RPC fallback).
- Accounts API reports `updateMode: 'full'`. `mergeAccountBalancesV6` replaces the covered chain slice, keeps `unprocessedCustomAssets` and staked vault balances, and re-asserts default tracked assets (mUSD) when the snapshot omits a zero balance.
- RPC fallback recovers errored chains and pins listed in `unprocessedCustomAssets`.
- Account Activity ignores pins/hides.
- Token-detection filtering is **not** applied; the v6 snapshot is kept in full.
- `CustomAssetGraduationMiddleware` does **not** run on this path. There is **no** `customAssetsOnly` RPC supplement.

### What v5 still does (flag off)

Unchanged from production:

- Force update attaches every pin (including hidden) on `DataRequest.customAssets`.
- Extra RPC poll (`customAssetsOnly`) for pins on chains another source already owns.
- `updateMode: 'merge'` with `replaceCoveredChainBalances`.
- `CustomAssetGraduationMiddleware` on Accounts API and Account Activity updates.

### Breaking

- Remove `excludeAssetIds` from `DataRequest`. v6 builds the query param from `assetPreferences`.
- `AccountsApiDataSource` now requires `getAssetsState`. `AssetsController` injects `() => this.state`.

`CustomAssetGraduationMiddleware`, `customAssetsOnly`, and `replaceCoveredChainBalances` stay on the v5 path and are **not** removed from the public API.

## References

- [APIPLAT-2499](https://consensys.atlassian.net/browse/APIPLAT-2499) — v6 `includeAssetIds` for pinned / zero-balance custom tokens
- Architecture: [`packages/assets-controller/docs/architecture/v1/CHANGES.md`](packages/assets-controller/docs/architecture/v1/CHANGES.md)
- Clients: [metamask-mobile#36604](https://github.com/MetaMask/metamask-mobile/pull/36604), [metamask-extension#46451](https://github.com/MetaMask/metamask-extension/pull/46451)

## Checklist

[APIPLAT-2499]: https://consensyssoftware.atlassian.net/browse/APIPLAT-2499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Large, breaking changes to core asset balance orchestration, subscriptions, and merge semantics behind a feature flag, with separate v5/v6 paths that must stay behaviorally correct at rollout.
> 
> **Overview**
> Adds a **runtime-split** balance pipeline for Accounts API **v6** when `assetsAccountsApiV6` is true, while keeping the existing **v5** behavior when the flag is off. The flag is read once in `AssetsController` and injected as `isBalanceV6Enabled` / `getAssetsState`.
> 
> On **v6**, pins and hides come from controller state (`includeAssetIds` / `excludeAssetIds` on the API), responses use **`updateMode: 'full'`** with `mergeAccountBalancesV6`, and **`getAssets({ customAssets })`** scopes one-shot fetches (e.g. new pins). **`unprocessedCustomAssets`** lets `RpcFallbackMiddleware` recover pins the API could not resolve; hide/unhide and pin changes **re-run subscriptions**. v6 skips custom-asset graduation and the v5-only RPC `customAssetsOnly` supplement.
> 
> **Breaking:** drop `'update'` from `AssetsUpdateMode`, remove `updateMode` from `getAssets`, and require **`getAssetsState`** on `AccountsApiDataSourceOptions`. Also fixes the remote flag to treat `assetsAccountsApiV6 === true` as enabled, and keeps default tracked assets (e.g. mUSD) at zero when a v6 full snapshot omits them.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4b51f14541cc53e89f5a40fe41ba4a8aa874f6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -292,19 +293,20 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+    //   );
+    //   const flag = remoteFeatureFlags?.assetsAccountsApiV6;
+    //   return (
+    //     typeof flag === 'object' &&
+    //     flag !== null &&
+    //     Boolean((flag as { value?: unknown }).value)
+    //   );
+    // } catch {
+    //   return false;
+    // }
```
Kriys94: Temp, revert before merge

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
Once happy with the v6 path, we expect to delete 2000 lines of code

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

packages/assets-controller/docs/architecture/v1/CHANGES.md
Prithpal-Sooriya, line 1: nit - future.

I think decisions like this should be captured as local doc/decisions folder & also prefixed with a timestamp.
`docs/decisions/2026-09-23-accounts-api-v6-integration.md`

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -252,15 +264,17 @@ export class AccountsApiDataSource extends AbstractDataSource<
   /** Shared AssetsController messenger, used to read remote feature flags. */
   readonly #messenger: AssetsControllerMessenger;
 
+  /** Injected by AssetsController; `true` when the v6 balances endpoint should be used. */
+  readonly #isBalanceV6Enabled: () => boolean;
+
+  readonly #getAssetsState: () => AssetsControllerStateInternal;
```
Prithpal-Sooriya, line 270: Nit - I really hate the types we have on this controller why are we trying to maintain `AssetsControllerStateInternal` and `AssetsControllerStateInternal`?

Pencil this for future papercuts

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -468,11 +456,61 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
-        ? await this.#fetchV6Balances(accountIds, fetchOptions, request)
-        : await this.#fetchV5Balances(accountIds, fetchOptions, request);
+      const isV6 = this.#isBalanceV6Enabled();
+      let fetchResult: {
+        unprocessedNetworks: string[];
+        unprocessedIncludeAssetIds: string[];
+        assetsBalance: Record<string, Record<Caip19AssetId, AssetBalance>>;
+      };
+
+      if (isV6) {
```
Prithpal-Sooriya, line 466: Instead of an if-else, could we not have returned early on one of the branches?

Not a deal breaker right now if we promise that this is a very short lived feature flag.

Prithpal-Sooriya: On this, I would actually like to add some stricter cyclomatic complexity rules to this controller - lets pencil this.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -507,14 +544,112 @@ export class AccountsApiDataSource extends AbstractDataSource<
       }
     }
 
-    // When token detection is disabled, filter out tokens not already in state
-    if (!this.#tokenDetectionEnabled() && this.#getAssetsState) {
+    // v5: when token detection is off, drop tokens not already in state.
+    // v6: the snapshot is authoritative (`updateMode: 'full'`); do not filter.
```
Prithpal-Sooriya, line 548: 1. is this comment necessary.
2. The negative conditions confused me.. essentially we are saying that this is a v5-only path?

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -507,14 +544,112 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+   */
+  #getIncludeAssetIds(
+    request: DataRequest,
+    chainsToFetch: ChainId[],
+    excludeAssetIds: Caip19AssetId[] | undefined,
+  ): Caip19AssetId[] | undefined {
+    const candidates =
+      request.customAssets && request.customAssets.length > 0
+        ? request.customAssets
+        : this.#getVisibleCustomAssetsFromState(request);
```
Prithpal-Sooriya, line 577: nit/long term architecture discussion.

What is the standardised pattern we want to centralise for data sources to get data that they need? I see areas were we inject/provide data, and maintaining both paths seem problematic in the long term. Is this something that we can decide and centralise on.
1. DataRequest fields?
2. DataSource constructor / callbacks?
3. Messenger calls

IMO fully remove the `customAssets` and other stateful properties from our requests... Or remove the controller state access from data sources.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -468,11 +456,61 @@ export class AccountsApiDataSource extends AbstractDataSource<
             }
           : undefined;
 
-      // Feature-flagged: v6 endpoint with a fallback to legacy v5. The flag is
-      // read here (not cached) so a runtime toggle can revert v6 -> v5.
-      const { unprocessedNetworks, assetsBalance } = this.#isBalanceV6Enabled()
-        ? await this.#fetchV6Balances(accountIds, fetchOptions, request)
-        : await this.#fetchV5Balances(accountIds, fetchOptions, request);
```
Prithpal-Sooriya, line 475: nit - This approach was much cleaner - a better abstraction.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -468,11 +456,61 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+        );
+
+        const validUnprocessedAssetIds =
+          fetchResult.unprocessedIncludeAssetIds.filter((assetId) => {
+            try {
+              parseCaipAssetType(assetId as Caip19AssetId);
+              return true;
+            } catch {
+              return false;
+            }
```
Prithpal-Sooriya, line 495: Can we use `isCaipAssetType` instead of this parsing try/catch?

```ts
fetchResult.unprocessedIncludeAssetIds.filter(
  (assetId) => isCaipAssetType(assetId)
)
```

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -468,11 +456,61 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+          });
+        if (validUnprocessedAssetIds.length > 0) {
+          response.unprocessedCustomAssets = [
+            ...(response.unprocessedCustomAssets ?? []),
```
Prithpal-Sooriya, line 499: Realistically no other pipeline/datasource will fill this response... we don't need to spread in anything here.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -507,14 +544,112 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+
+  #getVisibleCustomAssetsFromState(request: DataRequest): Caip19AssetId[] {
+    const state = this.#getAssetsState();
+    const visible: Caip19AssetId[] = [];
+    for (const { account } of request.accountsWithSupportedChains) {
+      for (const assetId of state.customAssets[account.id] ?? []) {
+        if (!state.assetPreferences[assetId]?.hidden) {
+          visible.push(assetId);
+        }
+      }
```
Prithpal-Sooriya, line 607: Okay, I think this is fine. We are getting customAssets that are not hidden.
Seems okay.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -507,14 +544,112 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+        continue;
+      }
+      let chainId: ChainId;
+      try {
+        chainId = parseCaipAssetType(assetId).chainId;
+      } catch {
+        continue;
+      }
+      if (
+        chainId.startsWith(`${KnownCaipNamespace.Eip155}:`) &&
```
Prithpal-Sooriya, line 643: nit (ignore not a dealbreaker) Above you perform `chainId = parseCaipAssetType(assetId).chainId`
Since you are parsing you could grab the namespace?

```ts
namespace === KnownCaipNamespace.Eip155
```

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -507,14 +544,112 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+        chainsToFetchSet.has(chainId)
+      ) {
+        filtered.add(assetId);
+      }
+    }
+
+    return filtered.size > 0 ? [...filtered] : undefined;
```
Prithpal-Sooriya, line 650: is spread needed?

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -790,16 +943,43 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
         }
 
+        if (response.updateMode === 'full') {
+          context.response = {
+            ...context.response,
+            updateMode: 'full',
+          };
+        } else if (
+          response.updateMode === 'merge' &&
+          context.response.updateMode !== 'full'
```
Prithpal-Sooriya, line 953: Can you walk through with me the difference between the response and the context.response?

packages/assets-controller/src/middlewares/ParallelMiddleware.ts
```diff
@@ -53,6 +53,17 @@ export function mergeDataResponses(responses: DataResponse[]): DataResponse {
...
+    if (
+      response.unprocessedCustomAssets &&
+      response.unprocessedCustomAssets.length > 0
+    ) {
+      merged.unprocessedCustomAssets = [
+        ...new Set([
+          ...(merged.unprocessedCustomAssets ?? []),
+          ...response.unprocessedCustomAssets,
+        ]),
+      ];
```
Prithpal-Sooriya, line 65: nice

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -1515,20 +1501,68 @@ export class RpcDataSource extends AbstractDataSource<
...
+      assetsInfo = {},
+      customAssets = {},
+      assetPreferences = {},
+    } = this.#getAssetsState();
+    const candidates =
+      requestCustomAssets && requestCustomAssets.length > 0
+        ? requestCustomAssets
+        : (customAssets[accountId] ?? []).filter(
+            (assetId) => !assetPreferences[normalizeAssetId(assetId)]?.hidden,
+          );
```
Prithpal-Sooriya, line 1535: Mentioned before - feels odd that we are trying to get customAssets from either request or from state - lets centralise on 1 flow.

packages/assets-controller/src/pipeline/buildFastFetchSources.ts
```diff
@@ -65,7 +72,9 @@ export function buildFastFetchSources(
       accountsApiDataSource,
       stakedBalanceDataSource,
     ]),
-    customAssetGraduationMiddleware,
+    ...(options.includeCustomAssetGraduation
+      ? [customAssetGraduationMiddleware]
+      : []),
```
Prithpal-Sooriya, line 77: nice!

packages/assets-controller/src/pipeline/buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -239,3 +255,82 @@ describe('assets pipeline: BNB Chain spam token (CDOGE)', () => {
     });
   });
 });
+
+describe('assets pipeline: BNB Chain spam token (CDOGE) imported as a custom asset', () => {
```
Prithpal-Sooriya, line 259: Yes love it! Proves that custom assets remain in the wallet.

If a user legitimately added a spam token, we shouldn't remove it.

packages/assets-controller/src/pipeline/buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -239,3 +255,82 @@ describe('assets pipeline: BNB Chain spam token (CDOGE)', () => {
...
+        rpcDataSource: source,
+        includeCustomAssetGraduation: false,
+      },
+    );
+
+    expect(requests).toHaveLength(0);
+    expect(BALANCES.lookUp(response, CDOGE_ASSET_ID_LOWERCASE)).toBeDefined();
+  });
+
+  it('re-reads the custom asset on RPC when the Accounts API omits it', async () => {
```
Prithpal-Sooriya, line 314: very good - this proves that RPC fallback is called for unsupported tokens!

packages/assets-controller/src/pipeline/buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya, line 1: Can we replicate this test (we can copy the test file) to prove testing with v6?

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya, line 1: Very nice test additions!

Lets copy this integration test file to prove v6 calls.

Prithpal-Sooriya (commented):
Changes look good. Main areas if we can polish:

1. DataSource state retrieval - centralise `customAssets` fetch to be from state instead of request.
2. Some of the shared DataSource logic (fetching/filtering custom assets) can be extracted - reduce responsibilities.
3. Lets copy/paste the integrate test files to ensure we test the v6 path integration.
  - new integration test files so we have a clear separation.
  - I can help with this.

Kriys94:
@metamaskbot publish-preview

Kriys94:
@metamaskbot publish-preview

packages/assets-controller/src/AssetsController.ts
```diff
@@ -637,6 +619,53 @@ function mergeAccountBalances(
...
+  accountBalances: Record<string, AssetBalance>,
+  skipDelete: (assetId: Caip19AssetId) => boolean,
+  updateMode: AssetsUpdateMode,
+): Record<string, AssetBalance> {
+  if (updateMode === 'merge') {
+    return { ...previousBalances, ...accountBalances };
+  }
+
+  const coveredChains = new Set(
+    Object.keys(accountBalances).map((assetId) => assetId.split('/')[0]),
```
Kriys94: Parse caip19 -> caip2

packages/assets-controller/src/AssetsController.ts
```diff
@@ -637,6 +619,53 @@ function mergeAccountBalances(
...
+  );
+  const next: Record<string, AssetBalance> = {};
+  for (const [assetId, balance] of Object.entries(previousBalances)) {
+    if (
+      !coveredChains.has(assetId.split('/')[0]) ||
```
Kriys94: Check against caip2

packages/assets-controller/src/AssetsController.ts
```diff
@@ -637,6 +619,53 @@ function mergeAccountBalances(
...
+      skipDelete(assetId as Caip19AssetId)
+    ) {
+      next[assetId] = balance;
+    }
+  }
+  Object.assign(next, accountBalances);
```
Kriys94: Change naming and logic

packages/assets-controller/docs/decisions/README.md
Prithpal-Sooriya: TY

packages/assets-controller/src/__fixtures__/bsc-spam-token/bscSpamWallet.ts
```diff
@@ -1,6 +1,6 @@
 import type { InternalAccount } from '@metamask/keyring-internal-api';
 
-import type { AssetsControllerStateInternal } from '../../types.js';
+import type { AssetsControllerState } from '../../types.js';
```
Prithpal-Sooriya: type change, this is acceptable.

packages/assets-controller/src/__fixtures__/scam-token-cleanup/scamWalletState.ts
```diff
@@ -1,6 +1,6 @@
 import type {
   AssetMetadata,
-  AssetsControllerStateInternal,
+  AssetsControllerState,
```
Prithpal-Sooriya: type change, this is acceptable.

packages/assets-controller/src/__fixtures__/spamWalletState.ts
```diff
@@ -4,7 +4,7 @@ import { getDefaultTrackedAssetsForChain } from '../defaults.js';
 import type {
   AssetMetadata,
   AssetPrice,
-  AssetsControllerStateInternal,
+  AssetsControllerState,
```
Prithpal-Sooriya: Type change, this is acceptable.

packages/assets-controller/src/data-sources/evm-rpc-services/services/BalanceFetcher.ts
```diff
@@ -200,6 +224,65 @@ export class BalanceFetcher extends StaticIntervalPollingControllerOnly<BalanceP
...
+   * visible set (native, pins, default tracked) plus anything already tracked
+   * with a balance entry, minus hidden assets and staking vault share tokens.
+   * Share-token `balanceOf` is not the converted staked amount; publishing it
+   * as `updateMode: 'full'` would overwrite the value from StakedBalanceFetcher.
+   *
+   * @param chainId - Hex chain ID (e.g. "0x1").
+   * @param accountId - Account UUID.
+   * @returns Array of asset fetch entries for the requested chain.
+   */
+  #getAssetsToFetchV6(
```
Prithpal-Sooriya: From our discussions, we are treating each balance fetching logic to be an "all or nothing".

For example - if AccountsAPI is unable to process a single token, then we will perform fallback for all tokens on that chain that had the failing token.

Lets keep a note of this, see if we see increase RPC usage through this.
I'm okay with this and we can optimise later if needed.

packages/assets-controller/src/data-sources/evm-rpc-services/services/BalanceFetcher.ts
```diff
@@ -200,6 +224,65 @@ export class BalanceFetcher extends StaticIntervalPollingControllerOnly<BalanceP
...
+    chainId: ChainId,
+    accountId: AccountId,
+  ): AssetFetchEntry[] {
+    const state = this.#messenger.call('AssetsController:getState');
```
Prithpal-Sooriya: Interesting that DataSources uses a constructor callback, whereas this uses a messenger.

I would prefer if we centralised all of our behaviour here... but not a deal breaker.

packages/assets-controller/src/data-sources/AbstractDataSource.ts
```diff
@@ -1,9 +1,4 @@
-import type {
-  ChainId,
-  DataRequest,
-  DataResponse,
-  AssetsControllerStateInternal,
-} from '../types.js';
+import type { ChainId, DataRequest, DataResponse } from '../types.js';
```
Prithpal-Sooriya: Acceptable - type change and constructor cleanup

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -489,30 +471,191 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+
+      const { unprocessedNetworks, unprocessedIncludeAssetIds, assetsBalance } =
+        await this.#fetchV6Balances(
+          accountIds,
+          this.#buildFetchOptions(request),
+          request,
+          visibleAssetIds.length > 0 ? visibleAssetIds : undefined,
+          hiddenAssetIds.length > 0 ? hiddenAssetIds : undefined,
+        );
+      response.updateMode = 'full';
```
Prithpal-Sooriya: Not for myself - I always was confused where the `updateMode` was applied.

After discussions it is confirmed that `updateMode` is **only applied to the `assetsBalance` property**. All other properties do not have update modes.

Prithpal-Sooriya: This is also an important distinction. V5 uses merge, whereas V6 is a full replace!

packages/assets-controller/src/data-sources/PriceDataSource.ts
Prithpal-Sooriya: This is an interface change. Rather than using request ctx for getting state, we instead inject via constructor callback.

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -561,6 +605,94 @@ export class RpcDataSource extends AbstractDataSource<
...
+      newBalances[balance.assetId] = {
+        amount: humanReadableAmount,
+      };
+    }
+
+    const response: DataResponse = {
+      assetsBalance: {
+        [result.accountId]: newBalances,
+      },
+      assetsInfo,
```
Prithpal-Sooriya: Hmmm just noting, no changes needed.

So the RPCDataSource also has responsibility for updating metadata??
Is this because if BF is disabled, we need metadata coming from RPC?

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -561,6 +605,94 @@ export class RpcDataSource extends AbstractDataSource<
...
+      updateMode: 'full',
```
Prithpal-Sooriya: Legit question - should RPC use a full update Mode?
Especially since RPCs can require making multiple calls. What happens if a call fails? Do the tokens vanish from the UI?

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -561,6 +605,94 @@ export class RpcDataSource extends AbstractDataSource<
...
+   * of the v6 path if `assetsAccountsApiV6` is rolled back.
+   *
+   * @param result - The balance fetch result.
+   */
+  async #handleBalanceUpdateV6(result: BalanceFetchResult): Promise<void> {
+    if (result.failedAddresses.length > 0) {
+      log('Skipping incomplete v6 poll snapshot', {
+        accountId: result.accountId,
+        chainId: result.chainId,
+        failedAddresses: result.failedAddresses,
```
Prithpal-Sooriya: Wow so if RPC (the last mechanism for balance fetching) fails, we just fail to show balances or anything to the User?

Can users see this as a regression, would we get incident reports or support tickets on this?

Prithpal-Sooriya: Consider scenario:
- I am a user who adds a custom network + RPC.
- my RPC is flakey
- my wallet shows a "RPC connectivity" banner, this is nice!
- BUT my wallet removes ALL my custom network tokens. FUD!

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -519,6 +546,140 @@ export class SnapDataSource extends AbstractDataSource<
...
+
+        // Step 3: Guard against an incomplete snap response. A `full` replace
+        // would drop any expected visible asset that was requested but omitted.
+        for (const assetId of visibleAssetIds) {
+          accountBalances[assetId] ??= { amount: '0' };
+        }
+
+        if (results.assetsBalance) {
+          results.assetsBalance[accountId] = accountBalances;
+        }
```
Prithpal-Sooriya: Hmm I think this is okay if we are happy with the user perspective.

As a user, should we accept that if the snap fails you will see `0` in your token balance instead of a stale balance?

packages/assets-controller/CHANGELOG.md
```diff
@@ -19,6 +38,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- On the Accounts API v6 RPC poll path, skip staking vault share tokens so a `full` snapshot cannot overwrite the converted staked amount with a raw `balanceOf`
```
Prithpal-Sooriya: Thoughts on shortening the changelog? 😅

Prithpal-Sooriya (dismissed):

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -1032,136 +1218,234 @@ export class RpcDataSource extends AbstractDataSource<
...
+    return this.#completeFetchResponse({
+      // A fully failed chain contributes nothing, so its balances stay as they are.
+      assetsBalance: filterFailedChainBalances(
+        assetsBalance,
+        new Set(failedChains),
+      ),
+      assetsInfo,
+      failedChains,
+      chainsToFetch,
+      updateMode: hasPartialSuccess ? 'merge' : 'full',
```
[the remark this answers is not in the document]
Kriys94: Fixed that. RPC is still updating in Full, In case of error for a balance, we just carry the old balance

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Fixed that. Same as RPC, we will is still updating in Full, In case of an error for a balance, we just carry the old balance or put 0

packages/assets-controller/src/data-sources/RpcDataSource.ts
Kriys94: So if we completely fail -> do nothing
If we partially fail -> keep old value and keep full update (potentially stale balance, but that is supposed to be a better UX than doing nothing for all balances of the chain)

packages/assets-controller/CHANGELOG.md
Kriys94: I've simplified it. WDYT?

packages/assets-controller/src/data-sources/evm-rpc-services/services/BalanceFetcher.ts
Kriys94: NIce finding. I removed that

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
Kriys94: I've documented the full update. https://github.com/MetaMask/core/blob/33429d645e8fa314a45363d96b6a2e5c6df4677a/packages/assets-controller/docs/decisions/2026-09-23-accounts-api-v6-integration.md

I've relaxed it a bit in case of a partial failure for RPC (carry old values) and Snaps (carry old values only for visible assets). For complete failures for RPC or Snaps, then we do nothing, even if that means stale balances.

packages/assets-controller/src/data-sources/RpcDataSource.ts
Kriys94: Yes — but only a narrow slice of metadata, not the Token API job.

`#collectMetadataForBalances` only emits a **native stub** (symbol/name from chain config, `decimals: 18`) when state does not already have valid decimals. ERC-20s are not written here; we only read existing decimals to convert the raw RPC amount.

That stub exists because:
1. **Basic functionality off** — Token + Price are skipped, so this is the only way a first-seen native gets symbol/decimals.
2. **Even with BF on** — a live RPC poll can land before TokenDataSource has filled `assetsInfo`, and we still need decimals to convert the balance.

Richer ERC-20 metadata still comes from TokenDataSource when BF is on, or from token detection on the RPC path when a new token is found. We also avoid re-emitting existing native metadata so we don't overwrite Accounts API fields (image, description) with the stub.

Kriys94: Yeah no change because that is currently in the (main) codebase

Kriys94: But yeah I dont think that is necessary

packages/assets-controller/src/data-sources/RpcDataSource.ts
Kriys94: I've documented the full update. https://github.com/MetaMask/core/blob/33429d645e8fa314a45363d96b6a2e5c6df4677a/packages/assets-controller/docs/decisions/2026-09-23-accounts-api-v6-integration.md

I've relaxed it a bit in case of a partial failure for RPC (carry old values) and Snaps (carry old values only for visible assets). For complete failures for RPC or Snaps, then we do nothing, even if that means stale balances.

Kriys94:
@metamaskbot publish-preview

PR #9652, opened 2026-07-24
Add lint:tsc package script

mcmire:
## Explanation

Currently, the monorepo is not being fully typechecked in CI. This is a problem because type errors have sneaked into test files, and the only way to see them is in an editor context (they are not always caught by Jest).

To fix this, this commit adds another category of tsconfig files (`tsconfig.lint.json`) which are like the development-only versions, except they emit type declarations to a temporary directory. This allows packages which have dependencies on other packages to resolve types correctly. These tsconfigs are linted by the `lint:tsconfigs` scripts. This commit also adds a `lint:tsc` package script which makes use of `tsconfig.lint.json`.

Note that if we were to run this across the monorepo we would have to address many type errors. To allow us to address these incrementally, this commit only enables linting for a subset of packages. We will increase this set in future PRs.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1174

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes are limited to tooling, CI, and TypeScript config; no runtime product code.
> 
> **Overview**
> Adds **repository-wide TypeScript typechecking in CI** via a new `lint:tsc` script (`tsc --build tsconfig.lint.json`), wired into the root `lint` command and the lint workflow matrix.
> 
> Introduces a **`tsconfig.lint.json` layer** (root + shared `tsconfig.packages.lint.json`) that mirrors dev configs but **emits declarations into `.tsc-lint-cache/`** so project references resolve across workspace packages. Only **four packages** are enabled initially (`base-controller`, `messenger`, `messenger-cli`, `platform-api-docs`) so the rest of the monorepo can opt in later.
> 
> Extends **`lint:tsconfigs`** to validate and fix references on `tsconfig.lint.json` separately from `tsconfig.json` / `tsconfig.build.json`, using workspaces that actually have a lint config. **`lint:dependencies`** now points knip at `knip.config.mts` explicitly; root `tsconfig.json` includes that file instead of `knip.config.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a50e40f309144e3e0f37594784f7ecc1fb27fff7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/base-controller/tsconfig.lint.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": ["./tsconfig.json", "../../tsconfig.packages.lint.json"],
```
mcmire, line 2: These per-package config files work a bit differently than the per-package `tsconfig.build.json` files, because they extend _two_ files rather than one. I chose to do this because these files are very similar to `tsconfig.json` in concept but they just need to be tweaked a little bit (whereas `tsconfig.build.json` is almost entirely different from `tsconfig.json`, so it only makes sense to extend `tsconfig.packages.build.json`).

scripts/create-package/cli.test.ts
```diff
@@ -44,7 +44,7 @@ describe('create-package/cli', () => {
   beforeEach(() => {
     // yargs calls process.exit() with 1 on failure and sometimes 0 on success.
     // We have to intercept it.
-    jest.spyOn(process, 'exit').mockImplementation((code?: number) => {
```
mcmire, line 47: Fixing existing type error in this file, which only showed up in my editor.

knip.config.mts
mcmire, line 1: In typechecking this file `tsc` told me that it interpreted `knip` as an ESM-only module but it interpreted this file as CommonJS. Renaming this file cleared the type error (although it did mean I had to update how we call `knip` in `package.json`, because `knip` won't look for `knip.config.mts` by default).

tsconfig.lint.json
```diff
@@ -0,0 +1,24 @@
+{
+  /**
+   * This configuration incrementally enables repository-wide type checking.
+   */
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "skipLibCheck": true,
```
mcmire, line 7: We enable this setting for builds, but we don't enable for development (i.e. `tsconfig.json` or `tsconfig.packages.json`). However, I found it to be necessary in this case. We are getting a large number of errors from `@types/node` and I wasn't quite sure how to resolve them. I figured we could investigate that problem in a different PR.

cryptodev-2s (commented):
I had started similar work in a local branch this can be extended easily to cover other packages if you want we can either merge here https://github.com/MetaMask/core/pull/9664 or merge after this one is merged

mcmire:
@cryptodev-2s Oh nice! I would merge it into this PR but then we'd need codeowner approvals from a bunch of teams. I still haven't updated the codeowners for the tsconfig files yet so maybe we can merge your PR once I've done that.

cryptodev-2s (approved):
LGTM!

PR #9655, opened 2026-07-24
chore: auto-close stale release PRs after 3 hours of inactivity

cryptodev-2s:
## Explanation

Abandoned `release/*` PRs can sit open and block other engineers from starting a new release. There is no automated cleanup today, so stale release candidates have to be closed by hand.

This PR adds a scheduled GitHub Actions workflow (every 30 minutes, plus `workflow_dispatch`) that:

- Finds open same-repo PRs whose head branch starts with `release/`
- Closes them after **3 hours of inactivity** (`updated_at`)
- Leaves a comment explaining why
- Deletes the release branch **only when its tip SHA is unchanged**

Fork heads, PRs in the merge queue, PRs with auto-merge enabled, and PRs labeled `release:keep-open` are skipped. The releasing docs are updated to describe this policy.

Before destructive actions, the workflow:

- Fetches a fresh GraphQL snapshot and skips if the PR is no longer eligible (open / skip label / fork / 3h / merge-queue or auto-merge)
- Closes before commenting so a failed close does not reset the stale timer
- Re-fetches the branch tip immediately before `deleteRef` and skips deletion unless the SHA still matches
- Comments only after a successful close and the delete attempt; tip-move skips and real delete failures get distinct note wording (successful deletes rely on GitHub’s normal closed-PR UI)
- Continues past per-PR API failures so one error does not abort the rest of the run

## Flow

```mermaid
flowchart TD
  trigger(["cron every 30m / workflow_dispatch"]) --> checkout[Checkout and setup]
  checkout --> listOpen[List open PRs via REST paginate]
  listOpen --> candidates[Filter candidates:<br/>same-repo release/*,<br/>no release:keep-open]

  candidates --> more{"More candidates?"}
  more -->|no| done([Done])
  more -->|yes| snap[GraphQL snapshot]

  snap --> elig{"Eligible?<br/>OPEN, not fork, not skip label,<br/>updatedAt older than 3h,<br/>not merge-queue / auto-merge"}
  elig -->|no| more
  elig -->|yes| closePR[Close PR]

  closePR --> closeOk{Close succeeded?}
  closeOk -->|no| more
  closeOk -->|yes| getRef[git.getRef for head branch]

  getRef --> tipOk{"Tip SHA still matches<br/>snapshot headRefOid?"}
  tipOk -->|refresh failed| keepRefresh[Keep branch:<br/>kept-refresh-failed]
  tipOk -->|moved| keepMoved[Keep branch:<br/>kept-head-moved]
  tipOk -->|unchanged| deleteRef[git.deleteRef]

  deleteRef --> delOk{Delete succeeded?}
  delOk -->|yes| deleted[Branch deleted]
  delOk -->|no| keepFail[Keep branch:<br/>kept-delete-failed]

  keepRefresh --> comment[Comment with outcome]
  keepMoved --> comment
  deleted --> comment
  keepFail --> comment
  comment --> more
```

## References

N/A

## Checklist

## Test plan

- [ ] Confirm workflow checks / actionlint pass on this PR
- [ ] Optionally run via `workflow_dispatch` against a throwaway stale `release/*` PR and verify comment, close, and branch delete
- [ ] Confirm a recently updated release PR is left alone
- [ ] Confirm a PR with `release:keep-open` is left alone
- [ ] Confirm a late push (head SHA change) causes branch deletion to be skipped and the comment says the branch was kept

---

> [!NOTE]
> **Medium Risk**
> The workflow can close PRs and delete `release/*` branches using `GITHUB_TOKEN`; safeguards (fresh snapshot, SHA check, skip label) limit mistakes but mistaken closure would disrupt in-flight releases.
> 
> **Overview**
> Adds automated cleanup for abandoned **release** PRs so they no longer block new releases.
> 
> A new scheduled GitHub Actions workflow runs every 30 minutes (and on `workflow_dispatch`) and executes `scripts/close-stale-release-prs.mts`. That script finds open same-repo PRs on `release/*` heads, closes them after **3 hours** without activity (`updatedAt`), posts an explanatory comment, and deletes the branch only if the tip SHA is unchanged. It skips fork heads, merge-queue/auto-merge PRs, and anything labeled `release:keep-open`. Per-PR failures are logged and do not stop the rest of the run.
> 
> **Releasing** docs now describe the 3-hour policy and the opt-out label. Root `package.json` gains `@actions/core` and `@actions/github` for the script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 528c27ce0da60b61f547178147c9ec77a86283aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,150 @@
...
+                issue_number: pr.number,
+                body,
+              });
+
+              await github.rest.pulls.update({
+                owner,
+                repo,
+                pull_number: pr.number,
+                state: 'closed',
+              });
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed in `0fd851f210`: re-fetch each PR before close/delete.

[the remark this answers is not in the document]
cryptodev-2s: Fixed in `0fd851f210`: close before commenting so a failed close doesn’t bump `updated_at`.

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            for (const candidate of releasePrs) {
+              // Re-fetch before acting: the initial list is a snapshot and the
+              // PR may have merged/closed (or become active) since then.
+              const { data: pr } = await github.rest.pulls.get({
+                owner,
+                repo,
+                pull_number: candidate.number,
+              });
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed in `d9cf15124c`: wrap `pulls.get` and merge-state checks in try/catch and continue.

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,324 @@
...
+                  `Closed #${latestPr.number} but failed to refresh ${expectedHeadRef} before delete: ${error.message}`,
+                );
+                continue;
+              }
+
+              if (branchSha !== expectedHeadSha) {
+                core.warning(
+                  `Closed #${latestPr.number} but skipped deleting ${expectedHeadRef}: head moved from ${expectedHeadSha} to ${branchSha}`,
+                );
+                continue;
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed in `$(git rev-parse --short HEAD)`: comment is posted after the delete attempt and reports whether the branch was deleted or kept.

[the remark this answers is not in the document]
cryptodev-2s: Fixed in `1538aed358`: comment is posted after the delete attempt and reports whether the branch was deleted or kept.

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,353 @@
...
+            const { owner, repo } = context.repo;
+
+            const pulls = await github.paginate(github.rest.pulls.list, {
+              owner,
+              repo,
+              state: 'open',
+              per_page: 100,
+            });
+
+            const releasePrs = pulls.filter((pr) => {
```
Mrtenz: I believe `github.rest.pulls.list` has options for filtering, so we can use that to search for specific branches, labels, even updated time, AFAIK. That should let us simplify this script a bunch.

Mrtenz: Hmm, not supported in `github.rest.pulls.list` directly actually, but you can do it using `github.request`(?): https://docs.github.com/en/rest/search/search?apiVersion=2026-03-10#search-issues-and-pull-requests

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,353 @@
...
+    name: Close stale release PRs
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Close inactive release PRs
+        uses: actions/github-script@v8
+        env:
+          STALE_HOURS: '3'
```
mcmire: Why are these environment variables? Why not set them as variables inside the script? This way you don't have to parse them as numbers.

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,353 @@
...
+          # Optional escape hatch for long-running releases that must stay open.
+          EXEMPT_LABEL: 'release:keep-open'
+        with:
+          script: |
+            const staleHours = Number(process.env.STALE_HOURS);
```
mcmire: This script is pretty long for being within a workflow file. Perhaps we should extract it?

mcmire (commented):
I made some comments below, but they assume that adding this workflow makes sense. That said, maybe there is a preexisting action we can use? This feels like something we shouldn't have to reinvent. For instance if we automatically assign a `release` label to release PRs then perhaps we can use something like this with the `only-pr-labels` option: https://github.com/marketplace/actions/close-stale-issues. There are also quite a few actions in the GitHub marketplace that do a similar thing.

.github/workflows/close-stale-release-prs.yml
cryptodev-2s: Done in `$(git rev-parse --short HEAD)`: these are now script constants (`STALE_HOURS` / `EXEMPT_LABEL`) instead of workflow env vars.

cryptodev-2s: Done in `0c750f5ab5`: these are now script constants (`STALE_HOURS` / `EXEMPT_LABEL`) instead of workflow env vars.

.github/workflows/close-stale-release-prs.yml
cryptodev-2s: Done in `0c750f5ab5`: logic moved to `.github/scripts/close-stale-release-prs.cjs`; the workflow now just checks out and invokes it.

cryptodev-2s:
> I made some comments below, but they assume that adding this workflow makes sense. That said, maybe there is a preexisting action we can use? This feels like something we shouldn't have to reinvent. For instance if we automatically assign a `release` label to release PRs then perhaps we can use something like this with the `only-pr-labels` option: https://github.com/marketplace/actions/close-stale-issues. There are also quite a few actions in the GitHub marketplace that do a similar thing.

I evaluated actions/stale. It handles general day-based staleness and label filtering, but this workflow needs a three-hour window, release/* branch targeting, merge-queue/auto-merge exemptions, and head-SHA verification before deletion. Supporting those guarantees would still require custom logic, so the extracted script remains the safer fit.

.github/workflows/close-stale-release-prs.yml
cryptodev-2s: I investigated the search endpoint, but it doesn’t support the required `release/*` head-prefix and precise three-hour filtering in a way that removes our validation logic. We would still need per-PR refreshes and merge-state checks, so search would add another API shape without materially simplifying the script.

mcmire:
@cryptodev-2s Ah I didn't realize that was the official `stale` action. Yeah, I looked through some other (non-official) stale actions and didn't really find anything. I'll continue looking at this PR.

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+      - name: Close inactive release PRs
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const closeStaleReleasePrs = require('./.github/scripts/close-stale-release-prs.cjs');
+            await closeStaleReleasePrs({ github, context, core });
```
mcmire: All of our scripts so far have been written in TypeScript, so this would introduce a new convention. What do you think about making `close-stale-release-prs.cjs` executable? Then we can make it a TypeScript file and run it here via `tsx`.

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+ * delete the branch when the tip is unchanged.
+ *
+ * Intended to be invoked from `actions/github-script`:
+ * `await require('./.github/scripts/close-stale-release-prs.cjs')({ github, context, core });`
+ */
+
+// Optional escape hatch for long-running releases that must stay open.
+const STALE_HOURS = 3;
+const EXEMPT_LABEL = 'release:keep-open';
+const STALE_MS = STALE_HOURS * 60 * 60 * 1000;
```
mcmire: Nit:

- Why declare two versions of the "stale time"?
- Instead of using "exempt" what about "skip"?
- Should we add JSDoc?

``` suggestion
/**
 * The label users can use to prevent stale release PRs from being auto-closed.
 */
const SKIP_LABEL = 'release:keep-open';

/**
 * The amount of time to keep stale release PRs open (in milliseconds).
 */
const STALE_HOURS = 3 * 60 * 60 * 1000;
```

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+ */
+module.exports = async function closeStaleReleasePrs({
+  github,
+  context,
+  core,
+}) {
+  const now = Date.now();
+  const { owner, repo } = context.repo;
+
+  const pulls = await github.paginate(github.rest.pulls.list, {
```
mcmire: Nit: Should we be more accurate about what these are?

``` suggestion
  const pullRequests = await github.paginate(github.rest.pulls.list, {

File: .github/scripts/close-stale-release-prs.cjs [hunk 10]
```diff
@@ -0,0 +1,549 @@
...
+ */
+function isReleasePrCandidate(pr, owner, repo, core) {
+  if (!pr.head.ref.startsWith('release/')) {
+    return false;
+  }
+
+  // Only manage same-repo release branches (never forks).
+  if (pr.head.repo?.full_name !== `${owner}/${repo}`) {
+    return false;
+  }
```
mcmire: Looks like PRs in the GitHub API have a `repo.fork` property you can use instead of checking the full name of the repo, maybe we can use that? 

``` suggestion
  if (pr.repo.fork) {
    return false;
  }

File: .github/scripts/close-stale-release-prs.cjs [hunk 11]
```diff
@@ -0,0 +1,549 @@
...
+ * Close inactive same-repo `release/*` PRs.
+ *
+ * @param {object} params - `actions/github-script` runtime bindings.
+ * @param {object} params.github - Octokit client from `actions/github-script`.
+ * @param {object} params.context - GitHub Actions context.
+ * @param {object} params.core - `@actions/core` helpers.
+ * @returns {Promise<void>}
+ */
+module.exports = async function closeStaleReleasePrs({
+  github,
```
mcmire: What do you think about using the `@actions/github` and `@actions/core` packages to get these variables instead of passing them from the workflow file?

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+/**
+ * Fetch a pull request by number.
+ *
+ * @param {object} github - Octokit client from `actions/github-script`.
+ * @param {string} owner - Repository owner.
+ * @param {string} repo - Repository name.
+ * @param {number} pullNumber - Pull request number.
+ * @returns {Promise<object>} Pull request payload.
+ */
+async function getPull(github, owner, repo, pullNumber) {
```
mcmire: Nit: Should we be more accurate about the naming here?

``` suggestion
async function getPullRequest(github, owner, repo, pullNumber) {

File: .github/scripts/close-stale-release-prs.cjs [hunk 13]
```diff
@@ -0,0 +1,549 @@
...
+  const suffix = phase ? ` ${phase}` : '';
+
+  if (pr.state !== 'open') {
+    core.info(
+      `Skipping #${pr.number} (${pr.head.ref}): no longer open${suffix}`,
+    );
+    return { eligible: false };
+  }
+
+  if (pr.labels.some((label) => label.name === EXEMPT_LABEL)) {
```
mcmire: Didn't we already exclude these kinds of PRs in `isReleasePrCandidate`?

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+ * @param {object} params.context - GitHub Actions context.
+ * @param {object} params.core - `@actions/core` helpers.
+ * @returns {Promise<void>}
+ */
+module.exports = async function closeStaleReleasePrs({
+  github,
+  context,
+  core,
+}) {
+  const now = Date.now();
```
mcmire: Since this never changes during the lifetime of the script, can this be a constant? This way we don't need to pass it around.

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+ *
+ * @param {object} github - Octokit client from `actions/github-script`.
+ * @param {string} owner - Repository owner.
+ * @param {string} repo - Repository name.
+ * @param {number} pullNumber - Pull request number.
+ * @returns {Promise<{ isInMergeQueue: boolean, autoMergeRequest: object | null }>}
+ * Merge state for the pull request.
+ */
+async function getMergeState(github, owner, repo, pullNumber) {
+  const mergeQueue = await github.graphql(MERGE_STATE_QUERY, {
```
mcmire: Instead of doing another query to get data about the PR, can we use GraphQL to fetch the PR info in the first place?

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+  try {
+    latestPr = await getPull(github, owner, repo, pr.number);
+  } catch (error) {
+    core.warning(
+      `Failed final refresh for #${pr.number} (${expectedHeadRef}): ${error.message}`,
+    );
+    return;
+  }
+
+  const finalEligibility = evaluateStaleEligibility({
```
mcmire: Why do we have to run through these checks a second time? I get that we need to be absolutely sure that the PR is safe to close, but I wonder if we can do that by reducing the number of asynchronous operations we are performing, e.g., performing _one_ GraphQL query instead of a few requests.

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+  } else {
+    branchStatusLines = [
+      `The release branch \`${expectedHeadRef}\` was **not** deleted (${detail}).`,
+      '',
+      'Delete the branch manually if it is no longer needed, or start a fresh release with `yarn create-release-branch`.',
+    ];
+  }
+
+  return [
+    '## Stale release PR closed',
```
mcmire: Nit: 

``` suggestion
    '## This pull request has been closed',

File: .github/scripts/close-stale-release-prs.cjs [hunk 18]
```diff
@@ -0,0 +1,549 @@
...
+    '',
+    `This release PR was automatically closed because it had no activity for ${STALE_HOURS} hours (last updated ${inactiveHours}h ago).`,
+    '',
+    'Open release PRs on `release/*` branches are expected to merge promptly so they do not block others from starting a new release.',
```
mcmire: Maybe we can omit the implementation detail:

``` suggestion
    'Open release PRs are expected to merge promptly so they do not block others from starting a new release.',

File: .github/scripts/close-stale-release-prs.cjs [hunk 19]
```diff
@@ -0,0 +1,549 @@
...
+ */
+function buildCloseComment({
+  inactiveHours,
+  expectedHeadRef,
+  expectedHeadSha,
+  outcome,
+  detail,
+  branchSha,
+}) {
+  let branchStatusLines;
```
mcmire: Hmm, do we need to print the status of whether the branch was deleted? If a pull request is closed, GitHub will already report whether the branch was deleted.

If a branch _fails_ to be deleted, I can see why we might want to include a little note to ourselves in the comment, but maybe we can kind of hide it? Something like:

> (A failed attempt was made to delete this branch. See more details here: \<link to workflow>)

.github/scripts/close-stale-release-prs.cjs
```diff
@@ -0,0 +1,549 @@
...
+ *
+ * @param {object} github - Octokit client from `actions/github-script`.
+ * @param {string} owner - Repository owner.
+ * @param {string} repo - Repository name.
+ * @param {number} pullNumber - Pull request number.
+ * @param {string} body - Markdown comment body.
+ * @param {object} core - `@actions/core` helpers.
+ * @returns {Promise<void>}
+ */
+async function commentOnPull(github, owner, repo, pullNumber, body, core) {
```
mcmire: Nit: Naming:

``` suggestion
async function commentOnPullRequest(github, owner, repo, pullNumber, body, core) {

File: .github/scripts/close-stale-release-prs.cjs (hunk 8 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`: renamed to `SKIP_LABEL`, collapsed to a single `STALE_DURATION_MS` with JSDoc.

File: .github/scripts/close-stale-release-prs.cjs (hunk 9 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 10 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`: using `head.repo.fork` (with a missing-`head.repo` guard). Note: REST exposes this on `head.repo`, not `pr.repo`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 12 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 17 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`: now `## This pull request has been closed`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 18 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 19 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `365060a387`: success no longer narrates deletion; failures get a short note with a link to the workflow run.

File: .github/workflows/close-stale-release-prs.yml (hunk 7 above)
Comment (cryptodev-2s) [reply .github/workflows/close-stale-release-prs.yml]: Done in `624f8146db`: moved to executable `scripts/close-stale-release-prs.ts` and run via `yarn tsx`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 11 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `624f8146db`: the script now uses `@actions/github` and `@actions/core` directly instead of workflow injection.

File: .github/scripts/close-stale-release-prs.cjs (hunk 13 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Kept on purpose for refreshed snapshots so a `release:keep-open` label added after the initial list still blocks close. Documented in `evaluateStaleEligibility`.

File: .github/scripts/close-stale-release-prs.cjs (hunk 14 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `624f8146db`: use a `staleBefore` cutoff for the first pass; the final check before close still uses a live cutoff.

File: .github/scripts/close-stale-release-prs.cjs (hunk 15 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `624f8146db`: one GraphQL snapshot query loads state, labels, head tip, and merge queue or auto merge status.

File: .github/scripts/close-stale-release-prs.cjs (hunk 16 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `624f8146db`: collapsed the REST get and separate merge state query into GraphQL snapshots. We still do a final snapshot plus tip `getRef` before delete for race safety.

File: .github/scripts/close-stale-release-prs.cjs (hunk 20 above)
Comment (cryptodev-2s) [reply .github/scripts/close-stale-release-prs.cjs]: Done in `364d420f96`: renamed to `commentOnPullRequest`.

File: .github/scripts/close-stale-release-prs.cjs [hunk 21]
```diff
@@ -0,0 +1,549 @@
...
+    `To keep a release PR open longer in exceptional cases, add the \`${SKIP_LABEL}\` label.`,
+  ];
+
+  // GitHub already surfaces successful branch deletion on the closed PR.
+  // Only call out failures / intentional skips.
+  if (outcome !== 'deleted') {
+    lines.push(
+      '',
+      `> (A failed attempt was made to delete this branch. See more details here: ${workflowRunUrl})`,
+    );
```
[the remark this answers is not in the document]
cryptodev-2s: Done in `364d420f96`: `kept-head-moved` now says the branch was left in place because the tip changed; only real delete failures use the failed attempt wording.

scripts/close-stale-release-prs.mts
```diff
@@ -0,0 +1,560 @@
...
+    `To keep a release PR open longer in exceptional cases, add the \`${SKIP_LABEL}\` label.`,
+  ];
+
+  // GitHub already surfaces successful branch deletion on the closed PR.
+  // Only call out failures / intentional skips.
+  if (outcome !== 'deleted') {
+    lines.push(
+      '',
+      `> (A failed attempt was made to delete this branch. See more details here: ${getWorkflowRunUrl()})`,
+    );
```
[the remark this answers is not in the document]
cryptodev-2s: Done in `364d420f96`: `kept-head-moved` now says the branch was left in place because the tip changed; only real delete failures use the failed attempt wording.

scripts/close-stale-release-prs.mts
```diff
@@ -0,0 +1,482 @@
...
+      owner,
+      repo,
+      state: 'open',
+      per_page: 100,
+    },
+  );
+
+  const releasePrs = pullRequests.filter((pullRequest) =>
+    isReleasePrCandidate(pullRequest),
+  );
```
mcmire: Nit: Can this be shortened?

``` suggestion
  const releasePrs = pullRequests.filter(isReleasePrCandidate);

File: scripts/close-stale-release-prs.mts [hunk 24]
```diff
@@ -0,0 +1,482 @@
...
+      octokit,
+      candidate,
+      staleBefore,
+    });
+  }
+}
+
+main().catch((error: unknown) => {
+  const message = error instanceof Error ? error.message : String(error);
+  core.setFailed(message);
```
mcmire: Nit: Can we use [`getErrorMessage` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89/src/errors.ts#L65)?

``` suggestion
  core.setFailed(getErrorMessage(error));

File: scripts/close-stale-release-prs.mts [hunk 25]
```diff
@@ -0,0 +1,482 @@
...
+  try {
+    await octokit.rest.issues.createComment({
+      owner,
+      repo,
+      issue_number: pullNumber,
+      body,
+    });
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    core.warning(`Closed #${pullNumber} but failed to comment: ${message}`);
```
mcmire: Nit: Can we use [`getErrorMessage` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89/src/errors.ts#L65)?

``` suggestion
    core.warning(`Closed #${pullNumber} but failed to comment: ${getErrorMessage(error)}`);

File: scripts/close-stale-release-prs.mts [hunk 26]
```diff
@@ -0,0 +1,482 @@
...
+  staleBefore: number;
+}): Promise<void> {
+  let snapshot: PullRequestSnapshot;
+  try {
+    snapshot = await getPullRequestSnapshot(octokit, candidate.number);
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    core.warning(
+      `Failed to refresh #${candidate.number} (${candidate.head.ref}): ${message}`,
+    );
```
mcmire: Nit: Can we use [`getErrorMessage` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89/src/errors.ts#L65)?

``` suggestion
    core.warning(
      `Failed to refresh #${candidate.number} (${candidate.head.ref}): ${getErrorMessage(error)}`,
    );

File: scripts/close-stale-release-prs.mts [hunk 27]
```diff
@@ -0,0 +1,482 @@
...
+    return;
+  }
+
+  const eligibility = evaluateStaleEligibility({ snapshot, staleBefore });
+  if (!eligibility.eligible || eligibility.ageMs === undefined) {
```
mcmire: If `eligible` is true, isn't `ageMs` present?

``` suggestion
  if (!eligibility.eligible) {

File: scripts/close-stale-release-prs.mts [hunk 28]
```diff
@@ -0,0 +1,482 @@
...
+ * @returns Markdown comment body.
+ */
+function buildCloseComment({
+  inactiveHours,
+  outcome,
+}: {
+  inactiveHours: string;
+  outcome: BranchDeleteOutcome['outcome'];
+}): string {
+  const staleHours = STALE_DURATION_MS / (60 * 60 * 1000);
```
mcmire: Nit: Could we use [`Duration.Hour` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89518ce3316ad940e35a1ef1a27a75cad24/src/time.ts)?

``` suggestion
  const staleHours = STALE_DURATION_MS / Duration.Hour;
```

(Also, sorry, I didn't realize this was used before. Maybe it _does_ make sense to make this a separate constant, like `STALE_DURATION_HOURS` or something.)

scripts/close-stale-release-prs.mts
```diff
@@ -0,0 +1,482 @@
...
+
+/**
+ * The label users can use to prevent stale release PRs from being auto-closed.
+ */
+const SKIP_LABEL = 'release:keep-open';
+
+/**
+ * How long inactive release PRs stay open before auto-close (milliseconds).
+ */
+const STALE_DURATION_MS = 3 * 60 * 60 * 1000;
```
mcmire: Nit: Could use use [`inMilliseconds` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89518ce3316ad940e35a1ef1a27a75cad24/src/time.ts)?

``` suggestion
const STALE_DURATION_MS = inMilliseconds(3, Duration.Hour);

File: scripts/close-stale-release-prs.mts [hunk 30]
```diff
@@ -0,0 +1,482 @@
...
+    );
+    return;
+  }
+
+  const eligibility = evaluateStaleEligibility({ snapshot, staleBefore });
+  if (!eligibility.eligible || eligibility.ageMs === undefined) {
+    return;
+  }
+
+  const inactiveHours = (eligibility.ageMs / (60 * 60 * 1000)).toFixed(1);
```
mcmire: Nit:

``` suggestion
  const inactiveHours = (eligibility.ageMs / Duration.Hour).toFixed(1);

File: scripts/close-stale-release-prs.mts [hunk 31]
```diff
@@ -0,0 +1,482 @@
...
+    await octokit.rest.pulls.update({
+      owner,
+      repo,
+      pull_number: pullNumber,
+      state: 'closed',
+    });
+    return true;
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    core.warning(`Failed to close #${pullNumber} (${headRef}): ${message}`);
```
mcmire: Nit: Can we use [`getErrorMessage` from `@metamask/utils`](https://github.com/MetaMask/utils/blob/6b14e89/src/errors.ts#L65)?

``` suggestion
    core.warning(`Failed to close #${pullNumber} (${headRef}): ${getErrorMessage(error)}`);

File: scripts/close-stale-release-prs.mts [hunk 32]
```diff
@@ -0,0 +1,482 @@
...
+ * @param options - Evaluation inputs.
+ * @param options.snapshot - Fresh GraphQL pull request snapshot.
+ * @param options.staleBefore - Epoch ms; PRs updated at/after this are kept.
+ * @returns Eligibility result.
+ */
+function evaluateStaleEligibility({
+  snapshot,
+  staleBefore,
+}: {
+  snapshot: PullRequestSnapshot;
```
mcmire: Nit: Might be good to call this `pullRequestSnapshot` so it's easier to tell at a glance what this refers to when skimming through this function.

``` suggestion
  pullRequestSnapshot,
  staleBefore,
}: {
  pullRequestSnapshot: PullRequestSnapshot;
```

or:

``` suggestion
  pullRequest,
  staleBefore,
}: {
  pullRequest: PullRequestSnapshot;
```

scripts/close-stale-release-prs.mts
```diff
@@ -0,0 +1,482 @@
...
+
+  if (snapshot.headRepository?.isFork) {
+    core.info(`Skipping #${snapshot.number} (${ref}): fork head`);
+    return { eligible: false };
+  }
+
+  const updatedAtMs = Date.parse(snapshot.updatedAt);
+  if (updatedAtMs >= staleBefore) {
+    core.info(
+      `Skipping #${snapshot.number} (${ref}): updated ${Math.round((Date.now() - updatedAtMs) / 60000)}m ago`,
```
mcmire: Nit: 

``` suggestion
      `Skipping #${snapshot.number} (${ref}): updated ${Math.round((Date.now() - updatedAtMs) / Duration.Hour)}m ago`,

**Review by mcmire — COMMENTED** [review]
Looks better. Small suggestions for now but I will do another higher-level review a bit later.

File: scripts/close-stale-release-prs.mts (hunk 23 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`.

File: scripts/close-stale-release-prs.mts (hunk 24 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`: now using `getErrorMessage` from `@metamask/utils`.

File: scripts/close-stale-release-prs.mts (hunk 25 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`: now using `getErrorMessage` from `@metamask/utils`.

File: scripts/close-stale-release-prs.mts (hunk 26 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`: now using `getErrorMessage` from `@metamask/utils`.

File: scripts/close-stale-release-prs.mts (hunk 27 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Good catch. Done in `3f8f445cc6`: return type is now a discriminated union so `ageMs` is required when `eligible` is true.

File: scripts/close-stale-release-prs.mts (hunk 28 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`.

File: scripts/close-stale-release-prs.mts (hunk 29 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`.

File: scripts/close-stale-release-prs.mts (hunk 30 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`.

File: scripts/close-stale-release-prs.mts (hunk 31 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`: now using `getErrorMessage` from `@metamask/utils`.

File: scripts/close-stale-release-prs.mts (hunk 32 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`: renamed to `pullRequest`.

File: scripts/close-stale-release-prs.mts (hunk 33 above)
Comment (cryptodev-2s) [reply scripts/close-stale-release-prs.mts]: Done in `3f8f445cc6`, using `Duration.Minute` so the `m ago` wording stays correct (Hour would understate the age).

File: scripts/close-stale-release-prs.mts [hunk 34]
```diff
@@ -0,0 +1,482 @@
...
+    return;
+  }
+
+  const inactiveHours = (eligibility.ageMs / (60 * 60 * 1000)).toFixed(1);
+
+  // Close before commenting so a failed close does not bump updatedAt.
+  const closed = await closePullRequest(
+    octokit,
+    snapshot.number,
+    snapshot.headRefName,
```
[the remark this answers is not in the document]
cryptodev-2s: Intentional after we simplified the closer: one GraphQL eligibility check plus tip verification before delete is enough for this cron. The tip check still protects branch deletion if the head moves after close.

scripts/close-stale-release-prs.mts
```diff
@@ -0,0 +1,487 @@
...
+
+  if (pullRequest.headRepository?.isFork) {
+    core.info(`Skipping #${pullRequest.number} (${ref}): fork head`);
+    return { eligible: false };
+  }
+
+  const updatedAtMs = Date.parse(pullRequest.updatedAt);
+  if (updatedAtMs >= staleBefore) {
+    core.info(
+      `Skipping #${pullRequest.number} (${ref}): updated ${Math.round((Date.now() - updatedAtMs) / Duration.Minute)}m ago`,
```
mcmire: Nit: Maybe we want to remind ourselves what the stale age of a release PR is:

``` suggestion
      `Skipping #${pullRequest.number} (${ref}): has not reached stale age of ${STALE_DURATION_HOURS}h yet (updated ${Math.round((Date.now() - updatedAtMs) / Duration.Minute)}m ago)`

File: scripts/close-stale-release-prs.mts [hunk 36]
```diff
@@ -0,0 +1,487 @@
...
+}
+
+/**
+ * Whether a listed PR head is a same-repo `release/*` branch that is not skipped.
+ *
+ * @param pullRequest - Pull request from `pulls.list`.
+ * @returns True when the PR is a candidate for stale close.
+ */
+function isReleasePrCandidate(pullRequest: ListedPullRequest): boolean {
+  if (!pullRequest.head.ref.startsWith('release/')) {
```
mcmire, line 128: Do you think it's worth it to reuse the [same check we use in `update-changelogs.yml`](https://github.com/MetaMask/core/blob/1a012c4c77be6a260735cea6ee33d4a556fdb2a4/.github/workflows/update-changelogs.yml#L36-L90) to determine whether this is a release branch? I realize that means part of this logic will live in the action and not the script. Alternatively we could not worry about it for now and refactor it in another PR.

mcmire (commented):
A couple more minor suggestions, but overall this looks pretty good.

scripts/close-stale-release-prs.mts
cryptodev-2s: Done in `7a05448336`.

scripts/close-stale-release-prs.mts
cryptodev-2s: Opened https://github.com/MetaMask/core/pull/9685 against this branch with a shared `.github/actions/is-release-pr` used by both `update-changelogs` and the stale closer (`release/*` prefilter, then the same merge-base + `action-is-release` check). Is that what you had in mind?

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,36 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: false
+          persist-credentials: false
+          cache-node-modules: true
```
mcmire: Do we need these two options? [`persist-credentials` no longer stores the token in the repo in `actions/checkout@v6`](https://github.com/actions/checkout#whats-new-1) ([we are using v7 in `action-checkout-and-setup`](https://github.com/MetaMask/action-checkout-and-setup/blob/0543b5929698c71e3ccc6ed24eac87825669b5de/action.yml#L310)). And since we aren't running the checkout step in one job and reusing the result in another job we don't need to cache Node modules.

``` suggestion
```

.github/workflows/close-stale-release-prs.yml
```diff
@@ -0,0 +1,36 @@
...
+    # Check twice an hour so the 3h window is reasonably precise.
+    - cron: '*/30 * * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: close-stale-release-prs
+  cancel-in-progress: false
```
mcmire: Nit: I think this is the default: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency

``` suggestion
```

mcmire (commented):
A couple more things.

scripts/close-stale-release-prs.mts
mcmire: Hmm, that's a bit more complex than I imagined. Maybe we can merge this PR and then we can discuss further in that PR.

cryptodev-2s: Happy to do a follow up after the merge

mcmire: Sounds good!

.github/workflows/close-stale-release-prs.yml
cryptodev-2s: Done in `68f2bf600a`: removed `cancel-in-progress: false`.

.github/workflows/close-stale-release-prs.yml
cryptodev-2s: Follow-up: dropped `cache-node-modules: true` in `` as well. You were right. With `is-high-risk-environment: false` the action already restores a shared `node_modules` cache when one exists; `cache-node-modules` is for the prepare-job write path (and wants `force-setup` when the same job needs a workspace). Not needed for this single cron job.

cryptodev-2s: Done in `68f2bf600a` and `528c27ce0d`: dropped both `persist-credentials: false` and `cache-node-modules: true`.

mcmire (approved):
LGTM!

PR #9656, opened 2026-07-24
fix(assets-controller): consume Snaps Events in all cases

Kriys94:
## Explanation

`SnapDataSource` previously delivered snap-sourced balance updates (from `AccountsController:accountBalancesUpdated`) by iterating `activeSubscriptions` and calling each subscription's own `onAssetsUpdate`. This meant updates were silently dropped whenever no active subscription was tracked for the chain at the moment the event arrived — which happens for networks like Tron, where a snap can report asset types (e.g. energy/bandwidth) that another data source (e.g. the WebSocket data source) owns the base chain for, but doesn't itself support. Since that other data source, not `SnapDataSource`, held the "active" subscription, the snap's updates never reached `AssetsController`.

The fix has `SnapDataSource` accept a single `onAssetsUpdate` callback in its constructor and report all snap balance updates directly to `AssetsController` through that callback, instead of fanning them out through per-subscription state. This decouples delivery from subscription bookkeeping: since snaps push real-time updates and don't need polling, `subscribe()`/`activeSubscriptions` no longer need to store or invoke their own `onAssetsUpdate`, which also let the associated subscribe/unsubscribe/destroy logic be simplified (no more `isUpdate` branching or per-subscription cleanup for snap chains).

`AssetsController` now passes `onAssetsUpdate: (response) => this.handleAssetsUpdate(response, 'SnapDataSource')` when constructing `SnapDataSource`, mirroring how the other data sources are wired up.

## References

* Fixes an issue where Tron (and similar snap-sourced) asset updates could be silently dropped when no active `SnapDataSource` subscription existed for the chain.

## Checklist

EOF

---

> [!NOTE]
> **Medium Risk**
> Changes how snap balance updates reach unified `assetsBalance` state; behavior is narrower but touches the multichain assets pipeline where incorrect delivery affects displayed balances.
> 
> **Overview**
> Fixes snap-sourced balances (e.g. Tron energy/bandwidth) being **silently dropped** when `SnapDataSource` had no matching entry in `activeSubscriptions`—common when another data source owns the chain subscription.
> 
> **`SnapDataSource`** now takes a required **`onAssetsUpdate`** at construction. `AccountsController:accountBalancesUpdated` and subscribe-time initial fetches call that callback directly instead of iterating `activeSubscriptions` or invoking per-subscription `onAssetsUpdate`. Subscribe logic is slimmed down (no `isUpdate` / subscription bookkeeping for delivery; less destroy cleanup).
> 
> **`AssetsController`** passes `onAssetsUpdate: (response) => this.handleAssetsUpdate(response, 'SnapDataSource')` when creating `SnapDataSource`. Tests and changelog updated; Jest coverage thresholds adjusted slightly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0745a3b082d0bc9bb0575b4d5c95f55a1fbb18b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -921,6 +921,8 @@ export class AssetsController extends BaseController<
     this.#snapDataSource = new SnapDataSource({
       messenger: this.messenger,
       onActiveChainsUpdated: this.#onActiveChainsUpdated,
+      onAssetsUpdate: (response): Promise<void> =>
+        this.handleAssetsUpdate(response, 'SnapDataSource'),
     });
     this.#rpcDataSource = new RpcDataSource({
       messenger: this.messenger,
--- a/packages/assets-controller/src/data-sources/SnapDataSource.ts
+++ b/packages/assets-controller/src/data-sources/SnapDataSource.ts
@@ -166,6 +166,12 @@ export type SnapDataSourceOptions = {
     chains: ChainId[],
     previousChains: ChainId[],
   ) => void;
+  /**
+   * Called directly with snap-sourced balance updates (from
+   * AccountsController:accountBalancesUpdated), independent of whether the
+   * controller currently has an active subscription for the chain.
+   */
+  onAssetsUpdate: (response: DataResponse) => void | Promise<void>;
   /** Configured networks to support (defaults to all snap networks) */
   configuredNetworks?: ChainId[];
   /** Default polling interval in ms for subscriptions */
@@ -208,6 +214,8 @@ export class SnapDataSource extends AbstractDataSource<
     previousChains: ChainId[],
   ) => void;
 
+  readonly #onAssetsUpdate: (response: DataResponse) => void | Promise<void>;
+
   /** Bound handler for snap keyring balance updates, stored for cleanup */
   readonly #handleSnapBalancesUpdatedBound: (
     payload: AccountBalancesUpdatedEventPayload,
@@ -226,6 +234,7 @@ export class SnapDataSource extends AbstractDataSource<
 
     this.#messenger = options.messenger;
     this.#onActiveChainsUpdated = options.onActiveChainsUpdated;
+    this.#onAssetsUpdate = options.onAssetsUpdate;
 
     // Bind handlers for cleanup in destroy()
     this.#handleSnapBalancesUpdatedBound = this.#handleSnapBalancesUpdated.bind(
@@ -305,9 +314,7 @@ export class SnapDataSource extends AbstractDataSource<
     // Only report if we have snap-related updates
     if (assetsBalance) {
       const response: DataResponse = { assetsBalance, updateMode: 'merge' };
-      for (const subscription of this.activeSubscriptions.values()) {
-        subscription.onAssetsUpdate(response)?.catch(console.error);
-      }
+      Promise.resolve(this.#onAssetsUpdate(response)).catch(console.error);
     }
   }
 
@@ -607,7 +614,7 @@ export class SnapDataSource extends AbstractDataSource<
   // ============================================================================
 
   async subscribe(subscriptionRequest: SubscriptionRequest): Promise<void> {
-    const { request, subscriptionId, isUpdate } = subscriptionRequest;
+    const { request, subscriptionId } = subscriptionRequest;
 
     // Guard against undefined request or chainIds
     if (!request?.chainIds) {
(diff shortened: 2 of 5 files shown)


packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -607,7 +622,7 @@ export class SnapDataSource extends AbstractDataSource<
   // ============================================================================
 
   async subscribe(subscriptionRequest: SubscriptionRequest): Promise<void> {
-    const { request, subscriptionId, isUpdate } = subscriptionRequest;
+    const { request, subscriptionId, onAssetsUpdate } = subscriptionRequest;
```
Prithpal-Sooriya: This makes sense, a `SubscriptionRequest` already has onAssetsUpdate.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -623,58 +638,22 @@ export class SnapDataSource extends AbstractDataSource<
...
-      chains: supportedChains,
-      onAssetsUpdate: subscriptionRequest.onAssetsUpdate,
-    });
-
-    // Initial fetch to get current balances
+    // Snaps provide real-time updates via AccountsController:accountBalancesUpdated,
+    // delivered directly to the controller via the `onAssetsUpdate` passed to
+    // the constructor (see #handleSnapBalancesUpdated). No polling, and no
+    // per-subscription state to maintain here — a fetch reports the current
+    // snapshot whether this is the initial subscribe or a later update.
```
Prithpal-Sooriya: Do we need this comment after merging? The comment is not pointing to "missing code", so very confusing to read.

This seems more of a comment you can add in this github PR.

Lets remove, and keep this comment in GH.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -623,58 +638,22 @@ export class SnapDataSource extends AbstractDataSource<
       return;
     }
 
-    if (isUpdate) {
```
Prithpal-Sooriya, line 629: Below was removed because:

Snaps provide real-time updates via AccountsController:accountBalancesUpdated, delivered directly to the controller via the `onAssetsUpdate`. No polling, and no per-subscription state to maintain here — a fetch reports the current snapshot whether this is the initial subscribe or a later update.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -208,6 +217,8 @@ export class SnapDataSource extends AbstractDataSource<
     previousChains: ChainId[],
   ) => void;
 
+  readonly #onAssetsUpdate: (response: DataResponse) => void | Promise<void>;
```
Prithpal-Sooriya, line 217: Not sure I entirely buy needed this when we already pass this `onAssetsUpdate` as part of each subscription?

packages/assets-controller/src/AssetsController.ts
```diff
@@ -921,6 +921,8 @@ export class AssetsController extends BaseController<
     this.#snapDataSource = new SnapDataSource({
       messenger: this.messenger,
       onActiveChainsUpdated: this.#onActiveChainsUpdated,
+      onAssetsUpdate: (response): Promise<void> =>
```
Prithpal-Sooriya, line 924: All other datasources do not pass in the `onAssetUpdate`, they instead use the subscription shape to perform updates.

Are we positive this is worth the interface change?

packages/assets-controller/src/data-sources/SnapDataSource.ts
Prithpal-Sooriya, line 629: This is the actual chunk we are removing -- the main area for this PR.

The rest feel like interface changes, that I think we can avoid.

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -305,9 +322,7 @@ export class SnapDataSource extends AbstractDataSource<
     // Only report if we have snap-related updates
     if (assetsBalance) {
       const response: DataResponse = { assetsBalance, updateMode: 'merge' };
-      for (const subscription of this.activeSubscriptions.values()) {
-        subscription.onAssetsUpdate(response)?.catch(console.error);
-      }
+      Promise.resolve(this.#onAssetsUpdate(response)).catch(console.error);
```
Kriys94, line 317: @Prithpal-Sooriya this is actually the main point of this PR

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: Good catch. This is actually not necessary

packages/assets-controller/src/data-sources/SnapDataSource.ts
```diff
@@ -623,58 +638,22 @@ export class SnapDataSource extends AbstractDataSource<
...
-    });
-
-    // Initial fetch to get current balances
+    // Snaps provide real-time updates via AccountsController:accountBalancesUpdated,
+    // delivered directly to the controller via the `onAssetsUpdate` passed to
+    // the constructor (see #handleSnapBalancesUpdated). No polling, and no
+    // per-subscription state to maintain here — a fetch reports the current
+    // snapshot whether this is the initial subscribe or a later update.
     try {
       const fetchResponse = await this.fetch({
```
Kriys94, line 634: Here, the fetch should return only the "time-based" assets, such as energy and bandwidth.

packages/assets-controller/src/data-sources/SnapDataSource.ts
Kriys94: This is exactly the purpose, i.e., pass all Snaps events to the AssetsController, even if there is no subscription. That will be the case with TRON, where we will not have explicit Snaps Subscription because it will be assigned to Websocket. Without subscription we still want to process `bandwidth` and `energy` updates

Kriys94:
@metamaskbot publish-preview

Prithpal-Sooriya (approved):

PR #9678, opened 2026-07-28
chore: improve unsub failure and retry logic

sahar-fehri:
## Summary

- Fix `OHLCVService` unsubscribe cleanup so failed WebSocket unsubs no longer leak gateway subscription slots and block future OHLCV streams.
- Flush grace-period channels immediately when subscribing to a different asset/interval, while keeping same-channel grace reuse for rapid back navigation.
- Retry failed unsubs with backoff (`1s → 2s → 4s`) and call `forceReconnection` only after retries are exhausted.

## Explanation

**Current state:** When `#performUnsubscribe` failed, `OHLCVService` removed the channel from local tracking before the gateway unsub succeeded. That left orphaned server-side subscriptions, which could exhaust the gateway's 2 `market-data` subscription cap and cause later subscribes to fail (clients fall back to REST `/latest` polling).

**Solution:**
1. **Delete-on-success** — keep channel entries until WS unsub succeeds.
2. **Grace-period flush** — when subscribing to a different channel, immediately unsub other channels in grace/failed-cleanup instead of holding two live subs for up to 3s. Same-channel grace reuse also flushes other channels first (Bugbot fix).
3. **Retry + last-resort reconnect** — on unsub failure, retry with backoff; after 3 retries, call `BackendWebSocketService:forceReconnection` to reset server subscription state.

Same-channel grace (3s) is unchanged so users returning to the same token/interval within 3s can reuse the existing subscription.

## Important: `forceReconnection` affects the shared WebSocket

`BackendWebSocketService` is a **single shared connection** on mobile, used by multiple services (not OHLCV-only):

| Consumer | Purpose |
|----------|---------|
| `AccountActivityService` | Real-time account balance / activity push |
| `OHLCVService` | Chart OHLCV streaming |
| `AssetsController` (indirect) | Forwards `AccountActivityService:balanceUpdated` |

Calling `forceReconnection()` closes and reconnects **that entire WebSocket**. On disconnect it clears all local subscription state; on reconnect each service resubscribes what it needs:

- **`AccountActivityService`** — marks chains down on `DISCONNECTED`, then resubscribes the selected account on `CONNECTED` (this service already uses `forceReconnection` for its own cleanup today).
- **`OHLCVService`** — only resubscribes channels with `refCount > 0` (active chart consumers). Failed-cleanup / grace entries (`refCount === 0`) are dropped on reconnect and are **not** resubscribed.

### When does OHLCV trigger it?

Only as a **last resort**, after unsub has failed **4 times total** (initial grace-period attempt + 3 retries at 1s / 2s / 4s). Typical navigation (leave chart, switch token, return within grace) uses delete-on-success, flush, or grace reuse — **not** force reconnect.

Rough timeline after user leaves a chart with a stuck unsub: ~**10s** of background retries before reconnect fires.

### User-visible impact

| Scenario | Impact |
|----------|--------|
| User **left chart**, unsub stuck → force reconnect (Test 5) | Brief (~1–2s) WS gap for **account activity** push; chart unaffected (hook unmounted). OHLCV does not resubscribe. |
| User **has chart open** when reconnect fires (unlikely — retries run after `refCount === 0`) | Brief WS gap; OHLCV resubscribes active channel; mobile hook may poll REST briefly during staleness window. |
| Normal chart navigation | **No** force reconnect |

### Why we still use it

- A permanently leaked `market-data` sub hits the gateway **2-sub cap** and can break **all** OHLCV streaming on that connection — worse than a one-time reconnect blip.
- `forceReconnection` is an existing, shared primitive (introduced in #6861; `AccountActivityService` already relies on it).
- Retries are attempted first; reconnect is the escalation path only when repeated unsub attempts fail.

**Open question for reviewers:** If the cross-service reconnect blast radius is unacceptable, alternatives are (a) gateway-side sub cleanup API, or (b) OHLCV-only reconnect/isolation — neither exists today on the shared `BackendWebSocketService`.

## Fixes

- [ASSETS-3763](https://consensyssoftware.atlassian.net/browse/ASSETS-3763)

## References

- Related mobile investigation: `UNSUBSCRIBE_LEAK_FIX_PLAN.md` in metamask-mobile
- Gateway behavior confirmed with backend: steady state is 1 live OHLCV sub per connection; cap of 2 is a race buffer for unsub→sub overlap
- `BackendWebSocketService.forceReconnection()` — #6861

## Test plan

- [x] `yarn workspace @metamask/core-backend run jest --no-coverage packages/core-backend/src/ws/ohlcv/OHLCVService.test.ts` (43 tests)
- [x] Manual (mobile, debug build): unsub retry — fail 4×, succeed on 5th
- [x] Manual: same-token return within grace — reuse WS sub, no new subscribe
- [x] Manual: token switch within grace — flush old channel immediately
- [x] Manual: grace reuse flushes other channels first (Bugbot scenario)
- [x] Manual: exhausted retries → `forceReconnection` → WS disconnected / connected cycle
- [ ] Manual: rapid token hopping on TDP — verify third subscribe succeeds (no `limit would be exceeded`)

## Checklist

- [ ] I've documented the shared-WebSocket impact of `forceReconnection` in code (JSDoc) — **PR description above; happy to add JSDoc in follow-up if reviewers want it in-tree**

[ASSETS-3763]: https://consensyssoftware.atlassian.net/browse/ASSETS-3763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes real-time OHLCV subscription lifecycle and can trigger shared WebSocket `forceReconnection`, affecting other consumers (e.g. account activity) after repeated unsub failures.
> 
> **Overview**
> Fixes **OHLCV WebSocket unsubscribe leaks** that could leave orphaned gateway `market-data` subscriptions and hit the **2-sub cap**, breaking later chart streams.
> 
> **`OHLCVService`** now keeps channel entries until the server unsub succeeds (**delete-on-success**). When subscribing to a **different** asset/interval, it **flushes** other channels still in grace or failed cleanup instead of holding two live subs for up to 3s; same-channel grace reuse is unchanged but also flushes other pending channels first. Failed unsubs are retried via **cockatiel** with **1s → 2s → 4s** backoff; only after retries are exhausted does it call **`BackendWebSocketService:forceReconnection`** on the shared connection. Retries are cancellable via **`AbortController`** (resubscribe, `destroy`, reconnect drops `refCount === 0` entries). **`destroy`** clears grace and retry timers; reconnect no longer resurrects stuck cleanup channels.
> 
> Adds **`cockatiel`** to `@metamask/core-backend` and expands **`OHLCVService.test.ts`** for flush, retry, destroy races, and connection-state edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f44eea8358308e559e4f6f28b7bc7960d6bdcc3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core-backend/src/ws/ohlcv/OHLCVService.ts
```diff
@@ -329,16 +362,110 @@ export class OHLCVService {
...
+      this.#messenger.publish('OHLCVService:subscriptionError', {
+        channel,
+        error: String(error),
+        operation: 'unsubscribe',
+      });
+      return false;
+    }
+  }
+
+  #scheduleUnsubscribeRetry(channel: string): void {
```
Kriys94, line 436: I wonder if we can handle the retries to TanStack?

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
```diff
@@ -43,13 +43,21 @@ const SYSTEM_NOTIFICATIONS_CHANNEL = `system-notifications.v1.${SUBSCRIPTION_NAM
...
+
 // =============================================================================
 // Types — Channel Tracking
 // =============================================================================
 
 type ChannelEntry = {
   refCount: number;
   gracePeriodTimer?: ReturnType<typeof setTimeout>;
+  retryTimer?: ReturnType<typeof setTimeout>;
+  retryCount?: number;
```
Kriys94: Potentially TanStack would remove this kind of trackers and all underlying logic. (gracePeriodTimer, retryTimer, retryCount). I'm not sure about refCount

Prithpal-Sooriya: Agreed here, maybe we can move away from imperative retry by using either TanStack or cockatiel.

```ts
import { ExponentialBackoff, handleAll, retry } from 'cockatiel';
import { QueryClient } from '@tanstack/query-core';

const testFn1 = async () => {
  const result = await retry(handleAll, {
    maxAttempts: 3,
    backoff: new ExponentialBackoff({ initialDelay: 200, maxDelay: 5_000 }),
  }).execute(async () => await doSomethingThatCanFail());
}

const testFn2 = async () => {
  const result = await new QueryClient().fetchQuery({
    queryKey: ['ohlcv-disconnect', someIDIfWannaMakeUnique],
    retry: 3,
    queryFn: () => doSomethingThatCanFail(),
  });
}
```

sahar-fehri: could you guys check this refactor here [74a7e01](https://github.com/MetaMask/core/pull/9678/commits/74a7e01cf6cd0ab9054151e4455866da8a70d7ce)
i went with cockatiel

Prithpal-Sooriya (dismissed):

Kriys94 (dismissed):
Slight preference for re-using TanStack, but I'm also good with the implementation

sahar-fehri:
@metamaskbot publish-preview

sahar-fehri:
@metamaskbot publish-preview

packages/core-backend/src/ws/ohlcv/OHLCVService.test.ts
```diff
@@ -438,33 +438,295 @@ describe('OHLCVService', () => {
...
+            new Promise<void>((resolve) => {
+              releaseConnect = resolve;
+            }),
+        );
+        mocks.channelHasSubscription.mockReturnValue(true);
+
+        const subscribePromise = service.subscribe(SUB_OPTS);
+
+        jest.advanceTimersByTime(3000);
+        await flushPromises();
```
Prithpal-Sooriya, line 583: If we see flakiness here, then let's replace with `waitFor()`. Not blocking

packages/core-backend/src/ws/ohlcv/OHLCVService.test.ts
```diff
@@ -438,33 +438,295 @@ describe('OHLCVService', () => {
...
+
+        service.destroy();
+        releaseUnsub(new Error('ws gone'));
+        await completeAsyncOperations();
+
+        jest.advanceTimersByTime(1000);
+        await completeAsyncOperations();
```
Prithpal-Sooriya, line 648: Same here, not blocking but the test specs need a proper refactor

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
```diff
@@ -329,16 +365,136 @@ export class OHLCVService {
 
     entry.gracePeriodTimer = setTimeout(() => {
       entry.gracePeriodTimer = undefined;
-      this.#performUnsubscribe(channel).catch(() => {
-        // no-op
-      });
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.#performUnsubscribe(channel);
```
Prithpal-Sooriya: Can we add back the catch noop, protects from throwing?

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
```diff
@@ -329,16 +365,136 @@ export class OHLCVService {
...
+    const entry = this.#channels.get(channel);
+    if (!entry) {
+      // The channel was removed (e.g. destroy() or reconnect cleanup) while the
+      // unsubscribe was in flight. Do not resurrect it or start a new retry loop
+      // with an AbortController that teardown can no longer cancel — that could
+      // force a reconnection on the shared WebSocket after teardown.
+      return;
+    }
+
+    entry.retryAbort?.abort();
```
Prithpal-Sooriya, line 446: Not - Hmm I wonder if this additional abort logic can be captured by a "debounce"?

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
```diff
@@ -329,16 +365,136 @@ export class OHLCVService {
...
+    this.#runUnsubRetryLoop(channel, signal);
+  }
+
+  async #runUnsubRetryLoop(
+    channel: string,
+    signal: AbortSignal,
+  ): Promise<void> {
+    try {
+      await unsubRetryPolicy.execute(async () => {
+        const releaseLock = await this.#mutex.acquire();
```
Prithpal-Sooriya, line 461: Locking worries me.
1. Do we need locks for this logic?
2. Are we confident that we cover all possible unlocks?
3. Is this shared mutex used on other operations. Does that mean if lock is "stuck" other operations are blocked?

sahar-fehri: Fair questions

1- Do we need locks? Yes; #channels and refCount are mutated from several concurrent async entrypoints (UI subscribe/unsubscribe, the grace-timer callback, the retry loop, and reconnect resubscribe). Without serialization we'd race on refCount and on WS subscribe/unsubscribe ordering (e.g. a double-subscribe, or unsubscribing a channel mid-subscribe).

2- Are all unlocks covered? Yes; every acquire() is paired with try { … } finally { releaseLock() }: subscribe, unsubscribe, #performUnsubscribe, the #runUnsubRetryLoop execute callback, and #resubscribeActiveChannels. Release always runs, even on throw.

3- Is it a shared mutex that could block other ops? It's private per OHLCVService instance (readonly #mutex = new Mutex()), not shared with BackendWebSocketService or any other service  so worst case it only serializes OHLCV's own subscribe/unsubscribe, nothing else. Also worth noting the lock is released between retries: Cockatiel's backoff waits happen outside the execute callback, so the 1s/2s/4s delays don't hold the mutex.

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
sahar-fehri: nice catch

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
sahar-fehri: The timing side is actually already a debounce: the grace-period timer waits GRACE_PERIOD_MS before tearing down and a fast re-subscribe cancels it. But the bit flagged here (retryAbort) is doing something a debounce can't: cancelling an in-flight retry loop mid-backoff (the 1s → 2s → 4s window before we fall back to forceReconnection).

A debounce only reacts to repeats of the same event and only delays the start. Here we need to interrupt work that's already running, triggered by three different external events:

1- User returns to the chart (#subscribeInner): abort the teardown and reuse the still-alive WS subscription instead of unsubscribing the channel they're now viewing.

2- Service teardown (destroy() → #clearChannelTimers): abort so an orphaned retry can't fire forceReconnection on the shared socket after teardown (the resurrection bug this PR fixes).

3- Reconnect cleanup (#resubscribeActiveChannels): abort a retry that's targeting the pre-reconnect subscription so we don't trigger a redundant reconnect storm.

packages/core-backend/src/ws/ohlcv/OHLCVService.ts
sahar-fehri: [f44eea8](https://github.com/MetaMask/core/pull/9678/commits/f44eea8358308e559e4f6f28b7bc7960d6bdcc3e)

Prithpal-Sooriya (approved):

PR #9695, opened 2026-07-29
A couple of fixes for tsconfig linting

mcmire:
## Explanation

- Fix per-package tsconfig validation
  - Per-package lint-specific tsconfigs were only required to reference internal dependencies that had a `tsconfig.lint.json`, not the full set of internal dependencies. This means that as we add support for linting tsconfigs for more packages, their `tsconfig.lint.json` files may not be fully synced with `tsconfig.json`.
- Hook `lint:tsconfig:fix:all` into `lint:fix`

## Checklist


mcmire:
This won't work. We need to exclude dependencies that don't have `tsconfig.lint.json` or else `lint:tsc` won't work.

PR #9696, opened 2026-07-29
feat(snap-account-service): report migration error

ccharly:
## Explanation

Right now, we only `console.log` if the migration fails (which is not supposed to happen). We have seen some user reports (with the console log `Migration failed during unlock`), though, it's hard to assess what's the underlying error of this failure.

We now want to instrument each steps and report those error properly so we can (hopefully) have a better understanding of what's the real failure.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches mandatory unlock-time keyring migration and error reporting; behavior on failure is clearer but migration logic itself is unchanged aside from error wrapping.
> 
> **Overview**
> **Snap keyring v2 migration failures are now reported to Sentry** instead of only logging to the console, with messages sanitized so addresses and other sensitive data are not sent upstream.
> 
> A new `errors` module introduces **`SafeError`**, **`withSafeError`**, and **`reportError`**. Migration steps (`addNewKeyring`, `removeKeyring`, and the overall unlock-time migration) run inside `withSafeError`, which keeps `KeyringControllerError` and existing `SafeError`s as-is, turns superstruct `StructError`s into path/type-only summaries, and replaces other errors with generic step-labeled messages. On unlock failure, **`reportError`** logs and calls the messenger’s optional **`captureException`** with a wrapper error whose **`cause`** is the (sanitized) failure.
> 
> Tests cover sanitization behavior, Sentry wiring, and migration error paths; the changelog documents the addition. **`SnapTracker`** gains a small test that removal events for untracked Snaps are ignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1d0bbfa87da2e8f594e89ad16681215c854074c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/snap-account-service/CHANGELOG.md
+++ b/packages/snap-account-service/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Report migration error in case of failure ([#9696](https://github.com/MetaMask/core/pull/9696))
+  - The migration is now a mandatory step and should never fail, so we want to report this as critical errors if anything goes wrong during that step.
+
 ### Changed
 
 - Bump `@metamask/account-tree-controller` from `^7.5.4` to `^7.5.5` ([#9470](https://github.com/MetaMask/core/pull/9470))
--- a/packages/snap-account-service/CHANGELOG.md
+++ b/packages/snap-account-service/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Report migration error in case of failure ([#9696](https://github.com/MetaMask/core/pull/9696))
-  - The migration is now a mandatory step and should never fail, so we want to report this as critical errors if anything goes wrong during that step.
+  - The migration is a mandatory and it is not suppose to fail.
+  - Reporting should help identifying which step could have failed and help addressing the actual issue.
 
 ### Changed
 
--- a/packages/snap-account-service/CHANGELOG.md
+++ b/packages/snap-account-service/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Report migration error in case of failure ([#9696](https://github.com/MetaMask/core/pull/9696))
-  - The migration is a mandatory and it is not suppose to fail.
+  - The migration is a mandatory and it is not supposed to fail.
   - Reporting should help identifying which step could have failed and help addressing the actual issue.
 
 ### Changed
--- a/packages/snap-account-service/src/SnapAccountService.ts
+++ b/packages/snap-account-service/src/SnapAccountService.ts
@@ -65,7 +65,7 @@ import { SnapId } from '@metamask/snaps-sdk';
 import type { Json } from '@metamask/utils';
 import { assertStruct } from '@metamask/utils';
 
-import { reportError, safe } from './errors.js';
+import { reportError, withSafeError } from './errors.js';
 import { projectLogger as log } from './logger.js';
 import type {
   SnapAccountServiceEnsureReadyAction,
@@ -344,7 +344,9 @@ export class SnapAccountService {
    */
   #handleUnlock(): void {
     // eslint-disable-next-line no-void
-    void safe('Snap keyring v2 migration', () => this.ensureMigrated()).then(
+    void withSafeError('Snap keyring v2 migration', () =>
+      this.ensureMigrated(),
+    ).then(
       async () => {
         // If the migration is successful, we re-forward the current groups to each new keyrings!
         const groupId = this.#getSelectedAccountGroupId();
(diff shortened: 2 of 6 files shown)


packages/snap-account-service/src/SnapAccountService.ts
```diff
@@ -343,18 +344,20 @@ export class SnapAccountService {
...
           groupId,
           this.#getAccountGroup(groupId)?.accounts,
         );
-      })
-      .catch((error) => {
-        console.error('Migration failed after unlock:', error);
-      });
+      },
+      (error) => {
+        reportError(this.#messenger, 'Migration failed after unlock', error);
```
ccharly, line 361: No real sensitive data (from `error`) should escape from here. I ran multiple checks with claude on this.

This can be compared with living an error bubbles up from a `:withKeyring*` call basically, which means, errors from `keyring.{serialize,deserialize}` could be thrown here.

In this context, we mostly use Snap keyrings (legacy + Snap keyring v2), so we could expect errors coming from there.

The real errors we don't have much control over are `superstruct` errors. We now wrap sensitive fields with our new decorator (`superstruct.sensitive`), but it's not widely    used yet.

The other alternative is to wrap each steps of the migration with various `try { ... } catch { throw Error('This step failed with ${text}'); }`, but that would make any analysis still pretty difficult I guess...

packages/snap-account-service/src/errors.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * NOTE: Sentry defaults to a depth of 3 when extracting non-native attributes.
+ * As such, the context depth shouldn't be too deep.
+ *
+ * @param message - The error message to create a Sentry error from.
+ * @param innerError - The inner error to create a Sentry error from.
+ * @param context - The context to add to the Sentry error.
+ * @returns A Sentry error.
+ */
+export const createSentryError = <
```
hmalik88, line 154: Don't we have something similar to this in the multichain-account-service package? Maybe we can extract that somewhere so we're not duplicating?

ccharly: IMO we could put that in `controller-utils` at some point. Exporting it from the `multichain-account-service` feels wrong to me.

I'm ok with duplicating until we move it to a better place, WDYT?

packages/snap-account-service/src/SnapAccountService.ts
ccharly: Ok, finally introduce a new pattern where we only report errors we know are "safe" to be reported.

It's not perfect, but I prefer this solution for now. Once we know why the migration is failing, we'll fix and eventually, remove those `SafeError`s.

Though, I think the pattern is elegant enough that we could use it in place where we need it.

On a side note, we should over-abuse our new `superstruct.sensitive` too to avoid letting dynamic value escape through `superstruct` validation IMO. But for now, it's not widespread enough.

packages/snap-account-service/src/errors.ts
hmalik88: OK that's fine with me, but let's open a ticket to not forget 😄

ccharly:
@metamaskbot publish-preview

packages/snap-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Report migration error in case of failure ([#9696](https://github.com/MetaMask/core/pull/9696))
+  - The migration is now a mandatory step and should never fail, so we want to report this as critical errors if anything goes wrong during that step.
```
mathieuartu: Might be nitpicky, but why *should* it never fail? 
Is it just a phrasing to say we don't want any migration to fail, or is it because it should genuinely be impossible that it fails? 

This + PR description confuses me a bit as a newcomer on this subject 😛

ccharly: Ah sorry, I meant more like "it's not suppose to fail", but we do not consider it like a "true" critical error for now (since we'll keep trying to migrate until it works)

I'll rephrase that!

ccharly: I tried something different:
https://github.com/MetaMask/core/pull/9696/commits/623b83bdea20252f0d97fcd4d6e270eacc166a33

WDYT?

packages/snap-account-service/src/errors.ts
```diff
@@ -0,0 +1,189 @@
...
+ *   never the actual value.
+ * - Anything else: wrapped in a {@link SafeError} with a generic fallback
+ *   message; the original error is discarded to prevent accidental leakage.
+ *
+ * @param step - Human-readable label for this execution step, prepended to the
+ *   `SafeError` message when the error is not already safe.
+ * @param fn - The async callback to execute.
+ * @returns The result of `fn`.
+ */
+export async function safe<ReturnType>(
```
mathieuartu: Nit: maybe a name change so it's more evident what it does? Something like `withSafeError `?

packages/snap-account-service/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Report migration error in case of failure ([#9696](https://github.com/MetaMask/core/pull/9696))
+  - The migration is a mandatory and it is not suppose to fail.
```
mathieuartu:
```suggestion
  - The migration is a mandatory and it is not supposed to fail.
```

packages/snap-account-service/src/errors.ts
ccharly: I like `withSafeError`! I'll go with that, thanks

ccharly: https://github.com/MetaMask/core/pull/9696/commits/677c4d2e5117e9852bf148217669e6dab0b4761e

packages/snap-account-service/CHANGELOG.md
ccharly: Related, I also changed that to avoid using "cannot fail" which is a bit "too strong" here:
https://github.com/MetaMask/core/pull/9696/commits/a1d0bbfa87da2e8f594e89ad16681215c854074c

ccharly:
@metamaskbot publish-preview

mathieuartu (approved):
Awesome!

PR #9401, opened 2026-07-07
fix: ignore saved advanced gas fees for bridge transactions

cloudonshore:
## Explanation

Bridge transactions apply the user's **saved advanced gas fees**, whereas swaps deliberately ignore them.

In `updateGasFees`, saved gas fees are skipped only for `SWAP_TRANSACTION_TYPES` (`swap`, `swapAndSend`, `swapApproval`). `bridge` / `bridgeApproval` are not in that list, so a user's saved advanced gas fees are applied to bridge transactions.

For a user who has saved a low max base fee (observed in production: `maxBaseFee: "0.05"` gwei, `priorityFee: "0"` for mainnet), every bridge is submitted **underpriced** — below the current base fee — causing it to fail (`GAS_TOO_LOW` on the relay) or get stuck as pending. This is exactly the failure swaps were already protected from.

### Fix

Ignore saved gas fees for bridge transactions as well, via a dedicated list:

```ts
const SAVED_GAS_FEES_IGNORED_TRANSACTION_TYPES: TransactionType[] = [
  ...SWAP_TRANSACTION_TYPES,
  TransactionType.bridge,
  TransactionType.bridgeApproval,
];
```

I intentionally did **not** add the bridge types to `SWAP_TRANSACTION_TYPES` itself, because that constant also gates swap-specific behavior in `updateSwapsTransaction` (e.g. the `simulationFails` cancel-and-throw, and `transactionNewSwap*` events). Keeping a separate, purpose-named list decouples "ignore saved gas fees" from "is a swap".

## References

- Root-caused from production state logs where a bridge smart transaction was submitted with `maxFeePerGas: 0x2faf080` (0.05 gwei) / `maxPriorityFeePerGas: 0x0`, sourced from `advancedGasFee["0x1"] = { maxBaseFee: "0.05", priorityFee: "0" }`.

### `@metamask/transaction-controller`

- **FIXED**: `bridge` and `bridgeApproval` transactions now ignore user-saved (advanced) gas fees, matching swaps.

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc) for new or updated code as appropriate
- [x] I've highlighted breaking changes using the "BREAKING" category above as appropriate

---

> [!NOTE]
> **Medium Risk**
> Changes gas pricing for all internal transactions (bridges, swaps, etc.); behavior is intentional but affects submission paths where wrong fees caused production failures.
> 
> **Overview**
[...]

PR #8993, opened 2026-06-03
feat(transaction-controller): support saved gas fee levels

pedronfigueiredo:
## Explanation

This expands transaction-controller saved gas fee handling so clients can resolve saved preferences with full transaction metadata instead of only a chain ID. That lets consumers key saved gas settings by both chain and account while keeping the controller API responsible for applying saved fees.

The saved gas payload now supports gas fee estimate levels (`low`, `medium`, `high`) and legacy custom `gasPrice` values in addition to EIP-1559 custom fee values. Saved estimate levels reuse the current gas fee flow estimate for the saved level; custom values override only the specific saved fields.

Transactions that already include gas fee params continue to use those initial params, so dapp-proposed fees keep precedence over saved preferences. Swap transaction types continue to skip saved gas fees.

## References

Related PRs:
- https://github.com/MetaMask/metamask-extension/pull/43317
- https://github.com/MetaMask/metamask-mobile/pull/31649

## Checklist

- [x] I’ve updated the test suite for new or updated code as appropriate
- [x] I’ve updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I’ve communicated my changes to consumers by updating changelogs for packages I’ve changed
- [x] I’ve introduced breaking changes in this PR and have prepared client adoption work to resolve them

---

> [!NOTE]
> **Medium Risk**
> Breaking callback signature and changed gas-fee precedence logic can misprice dApp txs or break extension/mobile until consumers update; changes are confined to gas fee application paths with substantial test coverage.
> 
> **Overview**
> **Breaking change:** `getSavedGasFees` now receives full `TransactionMeta` instead of chain ID so clients can key saved preferences by chain and account. The `SavedGasFees` shape adds optional estimate `level`, legacy `gasPrice`, and makes EIP-1559 fields optional.
> 
> Gas fee application in `updateGasFees` now skips saved lookup when the tx already has initial gas params (dapp fees win), still skips internal txs, and can pick **low/medium/high** from the current gas fee flow when only a level is saved. Custom `maxBaseFee` / `priorityFee` / `gasPrice` override specific fields; `userFeeLevel` stays on the saved level only when pricing is truly level-derived (otherwise `custom`).
> 
> Deprecated **incoming transaction** constructor options and restored **no-op** `startIncomingTransactionPolling` / `stopIncomingTransactionPolling` for backward compatibility, with tests.
> 
[...]

PR #9704, opened 2026-07-30
fix: replace underpriced dapp-suggested and saved gas fees with suggested estimates

cloudonshore:
## Explanation

Underpriced gas fees — a `maxFeePerGas` below (or barely above) the current base fee — produce transactions that are very unlikely to be included before fee conditions change further. They typically end up stuck as pending until eventually marked as failed (e.g. `Transaction not found on network after timeout`), with no actionable feedback to the user. Telemetry shows this is a meaningful contributor to dropped/timed-out transactions.

Two fee sources can produce such values, and nothing in the client validates either against current network conditions. This PR adds a feature-flag-gated staleness gate for both, at transaction creation time — the last point where fees can still be changed before signing.

### 1. Dapp-suggested fees (`replaceUnderpricedDappGasFees` flag)

Currently:

- `updateGasFees` uses `initialParams.maxFeePerGas` as-is.
- Gas estimation deliberately strips fee fields before calling the RPC node, so the node's own fee-cap validation never runs.
- `GasFeePoller` only refreshes fees for preset levels, so `dappSuggested` fees stay frozen while the confirmation is open.

A dapp that computes fees from stale data can therefore produce a signed transaction that cannot mine. With the flag enabled:

- A new `shouldIgnoreDappGasFees` check fires when the dapp-suggested `maxFeePerGas` (or a legacy `gasPrice` used as `maxFeePerGas` on EIP-1559 networks) is below the current **low** estimate. The low estimate is the floor rather than the raw base fee, since fees between the base fee and the low estimate have near-zero effective priority fee and no headroom for base fee movement, so they share the same failure mode.
- When it fires, the transaction falls through to the existing suggested (medium) fee logic for both `maxFeePerGas` and `maxPriorityFeePerGas`.
- `userFeeLevel` is set to `medium` instead of `dappSuggested`, which also opts the transaction into `GasFeePoller` refresh while unapproved.
- The original values remain available via `dappSuggestedGasFees` on the transaction metadata, so UIs can still display the site suggestion and users can revert via the edit modal.

Note: for transactions with complete dapp-suggested fees, `getSuggestedGasFees` previously skipped fetching estimates entirely. When the flag is enabled, estimates are now fetched for these transactions so the comparison can run — one additional gas fee estimate call per dapp transaction on enabled chains.

### 2. Saved (advanced) gas fee preferences (`replaceUnderpricedSavedGasFees` flag)

A saved *custom* preference is a static `maxBaseFee` captured at some point in the past; once the base fee rises past it, every transaction priced from it strands. Nothing else guards this source: the dapp gate deliberately excludes `savedGasFees`, and `hasInitialGasFeeParams` only protects transactions carrying explicit fee params — saved fees apply precisely when they don't. This is increasingly relevant as clients expand saved-preference usage (per-account custom gas settings persisted from the confirmation flow, and #9682 extending saved fees to wallet-initiated transfers).

With the flag enabled:

- After suggested fees are fetched, if the saved preference has a custom `maxBaseFee` below the current low estimate, the saved preference is ignored for this transaction and the suggested (medium) values are used instead.
- **Level-based saved preferences (`low`/`medium`/`high`) are never touched** — they track current estimates by construction.
- `userFeeLevel` becomes `medium` instead of `custom`, with the same `GasFeePoller` refresh benefit.
- No fetch-path changes needed: transactions eligible for saved fees carry no fee params, so estimates are already fetched.

### Guards / fail-open behavior (both gates)

Fees are kept unchanged when: the respective feature flag is not enabled for the chain (both default to disabled; per-chain rollout via `perChainConfig` with `default` fallback, mirroring `timeoutAttempts`); the transaction is internal (dapp gate; swaps/bridges are priced by the aggregator/relay); the fee is at or above the low estimate; or estimates are unavailable / not fee-market type.

### The product decision (saved-fees gate)

Overriding a value the *user* saved is a stronger intervention than overriding a dapp's computed value. Three defensible behaviors when the saved-fees gate fires: (1) silently use suggested fees (this PR's behavior), (2) use suggested fees + surface an alert ("your saved gas setting is below current network fees"), (3) keep the saved value + warn only. The flag ships disabled, so this can be settled before enablement.

An argument for intervening (1 or 2) rather than deferring entirely to the saved value: **saved preferences already do not grant full autonomy.** The controller already ignores them for swaps and bridges (`SAVED_GAS_FEES_IGNORED_TRANSACTION_TYPES`, #9401/#9682) — explicitly because a saved fee could underprice aggregator-priced transactions — and already yields them to explicit fee params (`hasInitialGasFeeParams`, #8993). The established principle is that saved preferences apply *where they can work* and yield *where they would predictably break the transaction*. A custom `maxBaseFee` below the current low estimate is the clearest instance of the latter — this extends existing precedent rather than setting new one.

## References

- Prior art for saved-fee exclusions: #9401, #9682, #8993
- Known edge case (fee bump vs. dapp max-spend arithmetic) and proposed balance-clamp refinement: see PR comment below — it applies to both gates.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes pre-sign gas selection for dapp and saved-fee paths when flags are enabled; behavior is gated and fail-open, but wrong thresholds or rollout could surprise users or sites.
> 
> **Overview**
> Adds **opt-in, per-chain** gates in `updateGasFees` so transactions are less likely to be priced below current network conditions before approval.
> 
> When **`replaceUnderpricedDappGasFees`** is enabled, dapp-suggested EIP-1559 fees whose effective `maxFeePerGas` (including legacy `gasPrice` used as max fee) is **below the wallet’s low estimate** are replaced with **suggested medium** `maxFeePerGas` / `maxPriorityFeePerGas`. **`userFeeLevel`** becomes **`medium`** instead of **`dappSuggested`**, so fees can keep updating while unapproved; originals stay on **`dappSuggestedGasFees`**. Complete dapp fee params no longer skip estimate fetch when this flag is on (needed for the comparison; fail-open if estimates are missing).
> 
> When **`replaceUnderpricedSavedGasFees`** is enabled, saved **custom** `maxBaseFee` below the low estimate is **dropped** for that tx so medium suggestions apply; **level-based** saved prefs are unchanged. **`userFeeLevel`** becomes **`medium`** instead of **`custom`**.
> 
> New remote feature-flag helpers and types mirror existing per-chain `default` / `perChainConfig` patterns; both flags default to **disabled**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95e622d6bc48a4df2aa6fed0079724ae8b99cf80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cloudonshore:
## Known edge case: dapp max-spend calculations

If a dapp computes a maximum spendable amount from its own fee suggestion (e.g. `value = balance - gas × dappMaxFeePerGas` for a native max send/wrap/bridge), replacing an underpriced fee with the medium estimate can push the worst-case reservation (`value + gas × maxFeePerGas`) above the account balance, surfacing an insufficient-funds state for a transaction the dapp's UI presented as affordable.

Worth noting: for the gate to fire, the base fee has already risen past the dapp's data — so the dapp's *amount* is stale too, not just its fee. There is no fee assignment that both fits the balance and can currently be included: the affordable (dapp) fee can't mine, and any mineable fee breaks the max-spend arithmetic. The replacement converts a silent, un-actionable failure (stuck pending, nonce occupied) into an explicit one, but it can still contradict the dapp's arithmetic in a user-visible way.

**Proposed refinement (follow-up):** clamp the replacement to what the balance can bear, since the actual charge (`baseFee + tip`) is far below the reservation:

```
replacementMaxFee = min(medium, (balance − value) / gas)
```

and only proceed if the result is still ≥ the low estimate. If it is, the transaction fits the balance *and* can mine — the max-spend flow works unchanged. If it isn't, the amount is genuinely unaffordable at any viable fee, and the replacement is skipped (fail open, current behavior). Cost is one balance read, only when the gate fires. The clamp should also be skipped for sponsored-gas / gas-fee-token transactions, where `maxFeePerGas` never draws on the native balance (those are safe to bump freely).

Suggested alongside: break down the eventual metrics by outcome (`replaced` / `replaced_clamped` / `skipped_unaffordable`) so the real-world rate of this edge case is measurable.

packages/transaction-controller/src/utils/gas-fees.test.ts
```diff
@@ -753,6 +766,247 @@ describe('gas-fees', () => {
...
+    });
+
+    it('uses gasPrice fallback if flow returns unsupported estimate type', async () => {
+      mockGasFeeFlowMockResponse({
+        estimates: {
+          type: 'unsupportedType',
+        },
+        // TODO: Replace `any` with type
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any as GasFeeFlowResponse);
```
pedronfigueiredo: can we use unknown here or try to define this?

packages/transaction-controller/src/utils/gas-fees.ts
```diff
@@ -398,11 +550,21 @@ async function getSuggestedGasFees(
...
+    hasCompleteDappFees &&
+    !txMeta.isInternal &&
+    getReplaceUnderpricedDappGasFeesEnabled(txMeta.chainId, messenger);
+
   if (
     (!eip1559 && txMeta.txParams.gasPrice) ||
-    (eip1559 &&
-      txMeta.txParams.maxFeePerGas &&
-      txMeta.txParams.maxPriorityFeePerGas)
+    (hasCompleteDappFees && !isEstimateRequiredForDappFees)
```
pedronfigueiredo, line 567: It would be good to preserve complete dapp-supplied fees when fee estimation and the fallback RPC both fail.

Enabling replaceUnderpricedDappGasFees makes transactions with complete fees enter the estimation path at the bottom of the file:

```typescript
  const gasPriceHex = (await rpcRequest({
    messenger,
    networkClientId,
    method: 'eth_gasPrice',
  })) as Hex | undefined;
```

An estimation failure falls through to the unguarded eth_gasPrice request; if that request rejects, transaction creation now fails instead of retaining the original complete fees. Catch the fallback failure or return empty suggestions on this comparison-only path so the feature continues to fail open.

In summary:
 - Flag disabled: no regression; the old early-return behavior remains.
 - Flag enabled, estimate succeeds: intended new behavior.
 - Flag enabled, estimate fails but eth_gasPrice succeeds: likely still fails open and retains the dapp fees.
 - Flag enabled, both estimate and eth_gasPrice fail: new failure mode; transaction creation rejects where it previously retained the complete dapp fees.

pedronfigueiredo (changes requested):
looking good just two small nits

packages/transaction-controller/src/utils/gas-fees.test.ts
cloudonshore: Done in 7f376164f — replaced with `as unknown as GasFeeFlowResponse` and dropped the eslint suppression.

packages/transaction-controller/src/utils/gas-fees.ts
cloudonshore: Good catch, fixed in 7f376164f. The comparison-only path now returns empty suggestions after a flow failure instead of falling through to the unguarded `eth_gasPrice` request, so both-failing can no longer reject transaction creation and the dapp fees are retained.

pedronfigueiredo (dismissed):

pedronfigueiredo (approved):

PR #9686, opened 2026-07-28
chore: update `@tanstack/query-core` to v5 in remaining packages

Gudahtt:
## Explanation

A recent PR updated this package to v5 in one package (#9563). This PR updates to v5 in all remaining packages.

## References

N/A

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

PR #9712, opened 2026-07-30
fix(base-data-service): adapt to `@tanstack/query-core` v5 API

cryptodev-2s:
## What

Stacked on top of #9686. That PR bumps `@tanstack/query-core` this adapts the code to v5.

## Changes

`packages/base-data-service/src/BaseDataService.ts`
* `invalidateQueries` filter type: in v5 the generic is the query key (constrained to `readonly unknown[]`), not page data, so `InvalidateQueryFilters<Json>` no longer type checks. Dropped the argument to keep the loose v4 ergonomics.
* `queryFn` can now be the `skipToken` sentinel (a `unique symbol`, not callable). Typed the options so `queryFn` is always a concrete function, since data services never use `skipToken`.
* `fetchInfiniteQuery`: v5 requires `initialPageParam`, and `fetchMore` meta no longer carries an explicit `pageParam` (v5 derives it from `getNextPageParam` / `getPreviousPageParam`). Now passes `initialPageParam` and injects page param resolvers at `query.fetch` time so pagination still fetches the exact requested page. This keeps cursor pagination working even for consumers that do not define page param callbacks.

`packages/chomp-api-service/src/chomp-api-service.ts`
* `cacheTime` was renamed to `gcTime` in v5. Without this, the "evict on settle" behavior stopped working.

Tests / helpers
* `hashQueryKey` was renamed to `hashKey`.
* `dehydrate` now writes a `dehydratedAt` field into query state.
* Renamed `cacheTime` to `gcTime` and typed the page param in the example service.

## Why the injection matters

`MoneyAccountApiDataService` paginates via an explicit cursor without defining `getNextPageParam`. Under the raw v5 bump its pagination test crashed with `options.getNextPageParam is not a function`. The resolver injection restores that behavior.

---

> [!NOTE]
> **Medium Risk**
> Touches shared caching and infinite-pagination logic used by many API services; incorrect merge or invalidation behavior could cause stale data or extra network calls, though tests cover key paths.
> 
> **Overview**
> Upgrades **`@tanstack/query-core`** from v4 to v5 across `base-data-service` and dependent packages, with **breaking** API updates for subclasses: rename **`cacheTime` → `gcTime`**, use **`hashKey`** instead of **`hashQueryKey`**, and require **`initialPageParam`** plus **`getNextPageParam` / `getPreviousPageParam`** on infinite queries.
> 
> **`BaseDataService`** is reworked for v5 typing and behavior: **`invalidateQueries`** filters use the query-key generic; **`fetchQuery` / `fetchInfiniteQuery`** require a concrete **`queryFn`** (no skip token) and align with v5 option types. Pagination no longer relies on v4 **`fetchMore` meta**—the first page goes through query-core’s **`fetchInfiniteQuery`** (reusing fresh cache), while later pages are fetched directly and merged into the infinite cache via **`setQueryData`**, preserving cursor-style callers that only pass an explicit page param (e.g. **`MoneyAccountApiDataService`**).
> 
> Downstream services and tests are updated (**`gcTime`**, **`initialPageParam`**, nullable page-param callbacks, persisted **`dehydratedAt`** expectations), plus a test that repeat activity requests hit cache without a second fetch. Yarn consistency allowlisting for the old v4 range is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea8763e0c7de9d9cbf81c459c0af29293df2ed95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
 
-    const result = await query.fetch(undefined, {
-      meta: {
-        fetchMore: {
-          direction,
-          pageParam,
+    // query-core v5 no longer accepts an explicit page param via the `fetchMore`
+    // meta; it derives the next/previous param from these callbacks instead.
+    // Override them to return exactly the requested page so pagination works
+    // even when the consumer did not provide page-param callbacks.
```
mcmire: Nit: Saying why we're _not_ doing something could be confusing to people who don't have context. What if we simply say:

``` suggestion
    // Override the next/previous param callbacks to return exactly the requested page so pagination works
    // even when the consumer did not provide page-param callbacks.

File: packages/base-data-service/src/BaseDataService.ts [hunk 2]
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
-      const result = await this.#queryClient.fetchInfiniteQuery({
+      // query-core v5 requires an `initialPageParam`, which becomes the param of
+      // the first (and only) page this fetches. Prefer an explicit per-call
+      // `pageParam` (a cold jump to a specific page); otherwise use the
+      // consumer's `initialPageParam`. Branching on a strict `undefined` check
+      // (rather than `??`) preserves `null`, which is a valid `Json` page param
+      // and query-core's usual first-page sentinel.
+      let initialPageParam: TPageParam;
+      if (pageParam === undefined) {
+        initialPageParam = options.initialPageParam as TPageParam;
```
mcmire: Not requiring `initialPageParam` is fine — our version of `fetchInfiniteQuery` is intentionally different from TanStack Query's version — but is there a way to not use a typecast here? Basically we are allowing the user to specify a `initialPageParam` that doesn't match `TPageParam` which seems odd. It seems that we need a variant of `TPageParam` that allows `undefined`.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -251,10 +253,14 @@ export class BaseDataService<
...
-        'retry' | 'retryDelay'
+        'retry' | 'retryDelay' | 'queryFn'
       >,
-      'queryKey' | 'queryFn'
-    >,
+      'queryKey'
+    > & {
+      // Data services always provide a concrete query function; the `skipToken`
+      // sentinel added in query-core v5 is not supported here.
+      queryFn: QueryFunction<TQueryFnData, TQueryKey>;
```
mcmire: Nit: Would the following be more future-proof?

``` suggestion
      queryFn: NonNullable<
        Exclude<
          FetchQueryOptions<
            TQueryFnData,
            TError,
            TData,
            TQueryKey,
            TPageParam
          >['queryFn'],
          SkipToken
        >
      >;
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
+        >,
+        'retry' | 'retryDelay' | 'queryFn' | 'initialPageParam'
       >,
-      'queryKey' | 'queryFn'
-    >,
+      'queryKey'
+    > & {
+      // Data services always provide a concrete query function; the `skipToken`
+      // sentinel added in query-core v5 is not supported here.
+      queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam>;
```
mcmire: Nit: Would the following be more future-proof?

``` suggestion
      queryFn: NonNullable<
        Exclude<
          FetchInfiniteQueryOptions<
            TQueryFnData,
            TError,
            TData,
            TQueryKey,
            TPageParam
          >['queryFn'],
          SkipToken
        >
      >;
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
+      // These are required by query-core v5 for infinite queries but remain
+      // optional here: consumers may drive pagination purely by passing an
+      // explicit `pageParam` (see below).
```
mcmire: Nit: Maybe we can drop mention of v5 so this comment doesn't become stale?

``` suggestion
      // These are required by @tanstack/query-core for infinite queries but remain
      // optional here: consumers may drive pagination purely by passing an
      // explicit `pageParam` (see below).

File: packages/base-data-service/src/BaseDataService.ts [hunk 6]
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
+        TPageParam
+      >({
         ...options,
+        initialPageParam,
+        // Provide a no-op `getNextPageParam` when the consumer omits one.
+        // query-core v5 walks `getNextPageParam` when it refetches a multi-page
+        // infinite query, so a missing resolver would throw once more than one
+        // page has been cached. The no-op rebuilds the cache down to the first
+        // page; the consumer repopulates it by re-navigating with explicit
+        // page params.
```
mcmire: Hmm, this doesn't seem as understandable as it could be. What do you think about dropping the last sentence? Also I wonder if we can drop mention of v5 to prevent this comment from becoming stale:

``` suggestion
        // Provide a no-op `getNextPageParam` when the consumer omits one.
        // @tanstack/query-core walks `getNextPageParam` when it refetches a multi-page
        // infinite query, so a missing resolver would throw once more than one
        // page has been cached.

File: packages/base-data-service/src/BaseDataService.ts [hunk 7]
```diff
@@ -251,10 +253,14 @@ export class BaseDataService<
...
         FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
-        'retry' | 'retryDelay'
+        'retry' | 'retryDelay' | 'queryFn'
       >,
-      'queryKey' | 'queryFn'
-    >,
+      'queryKey'
+    > & {
+      // Data services always provide a concrete query function; the `skipToken`
+      // sentinel added in query-core v5 is not supported here.
```
mcmire: Nit: What do you think about dropping mention of v5 to prevent this comment from becoming stale?

``` suggestion
      // @tanstack/query-core's fetchQuery function accepts a "skip" token,
      // but data services always provide a concrete query function.

File: packages/base-data-service/src/BaseDataService.ts [hunk 8]
```diff
@@ -280,47 +286,107 @@ export class BaseDataService<
...
+          TPageParam
+        >,
+        'retry' | 'retryDelay' | 'queryFn' | 'initialPageParam'
       >,
-      'queryKey' | 'queryFn'
-    >,
+      'queryKey'
+    > & {
+      // Data services always provide a concrete query function; the `skipToken`
+      // sentinel added in query-core v5 is not supported here.
```
mcmire: Nit: What do you think about dropping mention of v5 to prevent this comment from becoming stale?

``` suggestion
      // @tanstack/query-core's fetchInfiniteQuery function accepts a "skip" token,
      // but data services always provide a concrete query function.

File: packages/base-data-service/tests/ExampleDataService.ts [hunk 9]
```diff
@@ -101,15 +101,21 @@ export class ExampleDataService extends BaseDataService<
...
+      gcTime: inMilliseconds(1, Duration.Day),
     });
   }
 
   async getActivity(
     address: string,
     page?: PageParam,
   ): Promise<GetActivityResponse> {
-    return this.fetchInfiniteQuery<GetActivityResponse>(
+    return this.fetchInfiniteQuery<
```
mcmire: It seems unfortunate that consumers would need to supply all of these type parameters now. I wonder if there is a way that they can be inferred. I'll have to think about this...

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -54,7 +56,7 @@ type CacheUpdatedType = DataServiceCacheUpdatedPayload['type'];
 export type DataServiceInvalidateQueriesAction<ServiceName extends string> = {
   type: `${ServiceName}:invalidateQueries`;
   handler: (
-    filters?: InvalidateQueryFilters<Json>,
+    filters?: InvalidateQueryFilters,
     options?: InvalidateOptions,
   ) => Promise<void>;
```
mcmire: What do you think about basing this off of `BaseDataService` so this never falls out of sync again?

``` suggestion
  handler: BaseDataService<
    ServiceName,
    Messenger<
      ServiceName,
      ActionConstraint,
      EventConstraint,
      // Use `any` to allow any parent to be set. `any` is harmless in a type constraint anyway,
      // it's the one totally safe place to use it.
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      any
    >
  >['invalidateQueries'];
```

If you want to go with this, I might suggest extracting a `BaseMessenger` type:

``` suggestion
  handler: BaseDataService<
    ServiceName,
    BaseMessenger<ServiceName>
  >['invalidateQueries'];
```

This type would look like this. You could keep it here or keep it in `@metamask/messenger`:

```typescript
/**
 * The supertype of all messengers, scoped to a namespace.
 *
 * @template Namespace - The namespace for the messenger's own actions and
 * events.
 */
export type BaseMessenger<Namespace extends string> = Messenger<
  Namespace,
  ActionConstraint,
  EventConstraint,
  // Use `any` to allow any parent to be set. `any` is harmless in a type constraint anyway,
  // it's the one totally safe place to use it.
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  any
>;
```

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -337,7 +403,7 @@ export class BaseDataService<
    * @returns Nothing.
    */
   async invalidateQueries(
-    filters?: InvalidateQueryFilters<Json>,
+    filters?: InvalidateQueryFilters,
```
mcmire: By default the type parameter to `InvalidateQueryFilters` is `unknown`. Do we want that? I think we were trying to keep this method JSON-compatible, so perhaps we should say:

``` suggestion
    filters?: InvalidateQueryFilters<Json[]>,
```

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -627,4 +837,151 @@ describe('BaseDataService', () => {
       expect(publishSpy).not.toHaveBeenCalled();
     });
   });
+
+  describe('fetchInfiniteQuery', () => {
```
mcmire: While structuring these tests using a `describe` block definitely matches our guidelines elsewhere, it does not follow the existing conventions of this file. We already have tests for `fetchInfiniteQuery`: see "handles paginated queries", "handles paginated queries starting at a specific page", etc. We also already have a method that makes use of `fetchInfiniteQuery` internally in `ExampleDataService`: see `getActivity`. Perhaps we can adapt or extend those tests and also `ExampleDataService`?

packages/base-data-service/src/BaseDataService.ts
mcmire, line 252: What do you think about updating this JSDoc?

``` suggestion
   * @param options - The options defining the query. Note that although this
   * method wraps `fetchQuery` from `@tanstack/query-core`, there are a few
   * restrictions:
   * - `queryKey` and `queryFn` are required
   * - `queryFn` must be a function, not a skip token
   * - `retry` and `retryDelay` are not available (retries can be customized
   *   using the constructor's `servicePolicyOptions`).
```

mcmire, line 280: What do you think about updating this JSDoc?

``` suggestion
  * @param options - The options defining the query. Note that although this
   * method wraps `fetchInfiniteQuery` from `@tanstack/query-core`, there are a
   * few differences:
   * - `queryKey` and `queryFn` are required
   * - `queryFn` must be a function, not a skip token
   * - `retry` and `retryDelay` are not available (retries can be customized
   *   using the constructor's `servicePolicyOptions`).
```

mcmire, line 282: Nit: Thoughts on tweaking this while we're at it?

``` suggestion
   * @returns A page's worth of data (i.e. what `queryFn` returns). Note that
   * this is different from `@tanstack/query-core`'s `fetchInfiniteQuery`
   * method, which returns all pages.
```

cryptodev-2s:
@metamaskbot publish-preview

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done, dropped the last sentence and the version mention.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done, using `InvalidateQueryFilters<Json[]>`.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Done.

cryptodev-2s: Done.

cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Applied. Dropped the `TPageParam` argument since `fetchQuery` doesn't have that generic.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Applied.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: Went with the `BaseMessenger` version. Extracted `BaseMessenger<Namespace>` and based the handler on `BaseDataService['invalidateQueries']`. No circular reference, and the full build passes. I also used `BaseMessenger` for the class's own messenger constraint so it stays in one place.

packages/base-data-service/src/BaseDataService.ts
cryptodev-2s: I tried this. Widening the page param type to allow `undefined` just moves the cast onto `queryFn` and `getNextPageParam`: their parameters are contravariant, so a consumer's callbacks typed for `TPageParam` stop fitting `TPageParam | undefined`. It produces two compile errors instead of one cast. So I kept the single cast and added a comment noting the value can legitimately be `undefined` at runtime even though query-core's type does not allow it.

packages/base-data-service/tests/ExampleDataService.ts
cryptodev-2s: Left as is for now. I don't have a clean way to infer these yet, and the `getActivityWithoutCallbacks` I added for the tests has the same verbosity. Happy to explore a helper if you have a direction in mind.

packages/base-data-service/src/BaseDataService.test.ts
cryptodev-2s: Reworked. Dropped the standalone harness and the nested describe, and moved the cases into flat tests built on `ExampleDataService`. Extended `ExampleDataService` with `getActivityWithoutCallbacks` (paginates by explicit cursor with no page param callbacks, and records the params its query function sees), so we cover pagination without callbacks, a stale refetch of a query with several pages, and preserving a `null` initial page param. Also made `PageParam`'s properties optional, which the query function already assumed.

cryptodev-2s:
@metamaskbot publish-preview

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -308,19 +347,34 @@ export class BaseDataService<
...
+      } as typeof query.options,
+      {
+        meta: {
+          fetchMore: {
+            direction,
+          },
         },
       },
-    });
+    );
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed. The base service now supplies a no-op `getNextPageParam` when the consumer omits one, so refetching a query with more than one cached page no longer calls a missing resolver. Added a regression test.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,22 +286,62 @@ export class BaseDataService<
...
+        TError,
+        TData,
+        TQueryKey,
+        TPageParam
+      >({
         ...options,
+        // query-core v5 requires an `initialPageParam`. When the caller drives
+        // pagination with an explicit `pageParam`, use it as the initial param
+        // so the first (and only) page fetched is the requested one.
+        initialPageParam: (options.initialPageParam ?? pageParam) as TPageParam,
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed. `initialPageParam` is no longer resolved with `??`, so an explicit `null` is preserved and reaches the query function. Added a regression test.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +312,117 @@ export class BaseDataService<
...
+        // than one page has been cached.
+        getNextPageParam: options.getNextPageParam ?? ((): null => null),
         queryFn: (context) =>
-          this.#policy.execute(() =>
-            options.queryFn({
-              ...context,
-              pageParam: context.pageParam ?? pageParam,
-            }),
-          ),
+          this.#policy.execute(() => options.queryFn(context)),
```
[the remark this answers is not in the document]
cryptodev-2s: Confirmed and fixed. The cold jump was writing the explicit `pageParam` into `initialPageParam`, so a stale refetch re-fetched the jumped page instead of recovering the first page. The jump param is now injected through `queryFn` while `initialPageParam` stays the consumer's value. Added a regression test.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +312,121 @@ export class BaseDataService<
...
+            // Use query-core's context param when it has one (the derived param
+            // during a refetch, or an explicit/`null` `initialPageParam`). Only
+            // when it has none (the very first page) substitute the caller's
+            // explicit `pageParam`, which is how a cold jump fetches a specific
+            // page. The strict `undefined` check (rather than `??`) forwards a
+            // `null` param unchanged.
+            options.queryFn(
+              context.pageParam === undefined
+                ? { ...context, pageParam }
+                : context,
```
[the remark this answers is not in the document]
cryptodev-2s: Confirmed and fixed. A cold jump with a non-`undefined` `initialPageParam` (for example `null`) was ignoring the jump. The fresh path now keys off the caller's explicit `pageParam`, and the cached path uses its own query function so nothing leaks into later navigation. Added a regression test. (This and the two duplicate comments are the same finding.)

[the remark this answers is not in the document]
cryptodev-2s: Confirmed and fixed. A cold jump with a non-`undefined` `initialPageParam` (for example `null`) was ignoring the jump. The fresh path now keys off the caller's explicit `pageParam`, and the cached path uses its own query function so nothing leaks into later navigation. Added a regression test. (This and the two duplicate comments are the same finding.)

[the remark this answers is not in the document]
cryptodev-2s: Confirmed and fixed. A cold jump with a non-`undefined` `initialPageParam` (for example `null`) was ignoring the jump. The fresh path now keys off the caller's explicit `pageParam`, and the cached path uses its own query function so nothing leaks into later navigation. Added a regression test. (This and the two duplicate comments are the same finding.)

packages/base-data-service/src/BaseDataService.ts
mcmire: Ah right, probably just a copy-paste error.

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -121,6 +121,102 @@ describe('BaseDataService', () => {
...
+
+  it('preserves a `null` initial page param', async () => {
+    const messenger = new Messenger({ namespace: serviceName });
+    const service = new ExampleDataService(messenger);
+
+    await service.getActivityWithoutCallbacks(TEST_ADDRESS, {
+      initialPageParam: null,
+    });
+
+    // `null` is a valid page param, so it must reach the query function rather
```
mcmire: Do we need a test for this? It seems that `null` is only a valid page param because we've said that it is in `ExampleDataService`, but otherwise the page param could technically be of any type (as long as it's JSON-compatible). Is this correct?

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -121,6 +121,102 @@ describe('BaseDataService', () => {
...
+
+    expect(page.data).toHaveLength(3);
+    // The explicit jump target must reach the query function, not the `null`
+    // initial page param.
+    expect(service.pageParamsSeen).toStrictEqual([
+      { after: TRANSACTIONS_PAGE_2_CURSOR },
+    ]);
+  });
+
+  it('does not refetch a fresh paginated query', async () => {
```
mcmire: Nit: The use of "fresh" here indicates that the query is new, but that does not seem to be the case. What about this wording:

``` suggestion
  it('does not refetch a previously fetched paginated query', async () => {
```

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -121,6 +121,102 @@ describe('BaseDataService', () => {
...
+    const messenger = new Messenger({ namespace: serviceName });
+    const service = new ExampleDataService(messenger);
+
+    // Only one page-1 response is mocked, so a second fetch would fail. A fresh
+    // cached query must be served without another request.
```
mcmire: There is no mention of mocking in this test — that happens in a previous step — so this comment is a bit distracting in my opinion. Thoughts on removing?

``` suggestion
```

packages/base-data-service/tests/ExampleDataService.ts
```diff
@@ -146,6 +155,66 @@ export class ExampleDataService extends BaseDataService<
...
+   *
+   * @param address - The account address.
+   * @param options - Optional page param and initial page param.
+   * @param options.page - The page to fetch.
+   * @param options.initialPageParam - The initial page param to configure.
+   * @returns A page of activity.
+   */
+  async getActivityWithoutCallbacks(
+    address: string,
+    {
```
mcmire: We list this file as an example that engineers can use to build their own data services. To make sure this is still a working example, we probably don't want to declare `page` as a named argument like this. The reason for this is due to `createUIQueryClient`, which is designed to create a TanStack query client object that intercepts queries and forwards them to a data service. When handling an infinite query (via e.g. `useInfiniteQuery`), `createUIQueryClient` will receive the `pageParam` and pass it to the data service method as the last argument: https://github.com/MetaMask/core/blob/27cb9e327c24091b154e21bb63361832a7dc8652/packages/react-data-query/src/createUIQueryClient.ts#L149

So I think we need to have this method take `page` as the last argument.

packages/base-data-service/tests/ExampleDataService.ts
```diff
@@ -146,6 +155,66 @@ export class ExampleDataService extends BaseDataService<
     );
   }
 
+  /**
+   * Fetch activity without providing page-param callbacks, driving pagination
```
mcmire: I'm curious why we have two different versions of `getActivity`. With the changes we've made to BaseDataService, are we introducing a new way to use `fetchInfiniteQuery` that wasn't possible before?

mcmire (commented):
Still working through this PR, but I had some questions/comments below in the meantime.

packages/base-data-service/tests/ExampleDataService.ts
cryptodev-2s: Good catch. `getActivityWithoutCallbacks` now takes `page` as the last positional argument like `getActivity`, so it works through `createUIQueryClient`. The `null` initial page param is set inside the method.

packages/base-data-service/src/BaseDataService.test.ts
cryptodev-2s: Done.

packages/base-data-service/src/BaseDataService.test.ts
cryptodev-2s: Done, removed.

packages/base-data-service/src/BaseDataService.test.ts
cryptodev-2s: Kept it. It guards the regression Bugbot flagged in [this thread](https://github.com/MetaMask/core/pull/9712#discussion_r3712890228): the base method was coercing a caller's page param to `undefined` via `??`. `null` is a valid `Json` page param and query-core's first page sentinel, so a data service that uses `null` (like this example now does) would silently fetch the wrong page. The test is about the base method preserving whatever the consumer passes, not about `null` being special to `ExampleDataService`.

packages/base-data-service/tests/ExampleDataService.ts
cryptodev-2s: No new capability. `getActivity` uses page param callbacks; `getActivityWithoutCallbacks` omits them, which is the shape `MoneyAccountApiDataService` already uses (paginating purely by explicit cursor). That callback free path is where several v5 edge cases showed up, so I wanted an example that exercises it: missing `getNextPageParam` on refetch ([thread](https://github.com/MetaMask/core/pull/9712#discussion_r3711435411)), a cold jump poisoning the first page ([thread](https://github.com/MetaMask/core/pull/9712#discussion_r3776124910)), and a cold jump ignoring a defined initial param ([thread](https://github.com/MetaMask/core/pull/9712#discussion_r3776317352)). `getActivity` needs its callbacks for its own tests, hence the second method.

cryptodev-2s: Renamed `getActivityWithoutCallbacks` to `getActivityByCursor` so the two are clearly deliberate: `getActivity` is the idiomatic infinite query with page param callbacks, and `getActivityByCursor` drives pagination purely by explicit cursor. They also cover different branches of `fetchInfiniteQuery` (backward-direction detection needs the callbacks; the no-op `getNextPageParam` fallback needs them absent), so a single method can't exercise both.

packages/base-data-service/CHANGELOG.md
```diff
@@ -38,6 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **BREAKING:** Remove `TPageData` type parameter from `invalidateQueries` method ([#9526](https://github.com/MetaMask/core/pull/9526))
   - This is technically a breaking change, but this was not used in any of our codebases
+- **BREAKING:** Bump `@tanstack/query-core` from `^4.43.0` to `^5.62.16` ([#9712](https://github.com/MetaMask/core/pull/9712))
+  - The option types accepted by `fetchQuery`, `fetchInfiniteQuery`, and `invalidateQueries` now follow the query-core v5 API. Subclasses may need to rename `cacheTime` to `gcTime`, and infinite queries no longer accept an explicit page param through the `fetchMore` meta.
```
FrederikBolding:
```suggestion
  - The option types accepted by `fetchQuery`, `fetchInfiniteQuery`, and `invalidateQueries` now follow the query-core v5 API. Subclasses need to rename `cacheTime` to `gcTime`.
```
`fetchMore` is not exposed publicly.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -280,47 +313,170 @@ export class BaseDataService<
...
       queryKey: options.queryKey,
     });
 
+    // A param-less call must return the first page. If the cached first page
+    // reports a previous page, it was produced by a cold jump (it is not
+    // actually the first page), so drop the query and refetch from
+    // `initialPageParam`. This detection needs `getPreviousPageParam`; a
+    // forward-only consumer cannot detect it, so for those a cold jump followed
+    // by a param-less read within `staleTime` still returns the jumped page.
+    if (query?.state.data && pageParam === undefined) {
```
FrederikBolding: The change to `fetchInfiniteQuery` seem extremely complicated. I haven't been able to wrap my head around whether they are all needed.

FWIW `pageParam` was added as a way to support the two categories of requests that data services need to support:
- Requests that come directly from `useInfiniteQuery` in the UI
- Requests that come from other services or controllers

It's convoluted, but I couldn't find a better way to do it. It seems more convoluted now that we have `getNextPageParam` 🤔 

Can we take a step back and try to simplify this?

packages/base-data-service/tests/ExampleDataService.ts
FrederikBolding: Agreed. Why is this suddenly not inferable?

packages/base-data-service/CHANGELOG.md
cryptodev-2s: Done.

packages/base-data-service/tests/ExampleDataService.ts
cryptodev-2s: Removing `getNextPageParam` shrinks the option list, but it doesn't fix inference on its own: `TPageParam` still only shows up in the `queryFn` argument and the optional `getPreviousPageParam`, so there's nothing required for TS to latch onto. I left the explicit generics for now and can take a follow up pass at getting the common case to infer.

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

FrederikBolding (approved):
Nice, lgtm!

ffmcgee725 (approved):

lwin-kyaw (approved):

Matt561 (approved):

jiexi (approved):
LGTM for kyc-controller

PR #9717, opened 2026-07-30
feat(assets-controller)!: fetch multicall address from config registry

mikesposito:
## Explanation

`MulticallClient` from `assets-controller` is being wired to `config-registry-controller` via messenger call. The registry is used as primary source for multicall3 addresses, and the existing bundle constant is used only as fallback.

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WPN-1746

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes affect on-chain balance batching for all EVM RPC paths and require a new messenger delegation; wrong or missing multicall addresses could break token balance fetches on some chains until registry data is correct.
> 
> **Overview**
> **BREAKING:** `AssetsControllerMessenger` must delegate `ConfigRegistryController:getNetworkConfigByCaip2ChainId`. Integrators need to wire that action when adopting this release.
> 
> `RpcDataSource` now passes a registry-backed resolver into `MulticallClient`: it calls the config registry for each chain’s `contracts.multicall3` (via CAIP-2), and only uses the built-in `MULTICALL3_ADDRESS_BY_CHAIN` map when the registry has no address.
> 
> `MulticallClient`’s constructor is refactored to take an options object (`getProvider`, `getMulticall3AddressForChain`, optional `config`) instead of positional args. Tests and mocks were updated accordingly.
> 
> On **config-registry-controller**, `RegistryNetworkConfig` gains optional `contracts.multicall3` (validated in the API schema), and exports drop `ConfigRegistryControllerMethodActions` in favor of the specific action type `ConfigRegistryControllerGetNetworkConfigByCaip2ChainIdAction`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a90c58c77e3eb38a64da492dbee4bebc30c2ed7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/data-sources/RpcDataSource.test.ts
```diff
@@ -210,6 +212,19 @@ async function withController<ReturnValue>(
...
+      !actionHandlerOverrides[
+        'ConfigRegistryController:getNetworkConfigByCaip2ChainId'
+      ]
+    ) {
+      rootMessenger.registerActionHandler(
+        'ConfigRegistryController:getNetworkConfigByCaip2ChainId',
+        () => {
+          console.log('Mock getNetworkConfigByCaip2ChainId called');
+          return undefined;
+        },
```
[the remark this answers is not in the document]
mikesposito: removed in [e0c099f](https://github.com/MetaMask/core/pull/9717/commits/e0c099f333fb054f90bd5e0fdc388fc21601b402)

packages/config-registry-controller/src/config-registry-api-service/types.ts
```diff
@@ -65,6 +69,7 @@ export const RegistryNetworkConfigSchema = type({
   rpcProviders: RpcProvidersSchema,
   blockExplorerUrls: BlockExplorerUrlsSchema,
   config: ChainConfigSchema,
+  contracts: optional(NetworkContractsSchema),
```
mikesposito, line 72: Added to the registry in https://github.com/MetaMask/mm-config-registry/pull/30

mikesposito:
@metamaskbot publish-preview

packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
```diff
@@ -591,7 +614,9 @@ export class MulticallClient {
       return [];
     }
 
-    const multicallAddress = MULTICALL3_ADDRESS_BY_CHAIN[chainId];
+    const multicallAddress =
+      this.#getMulticall3AddressForChain(chainId) ??
+      MULTICALL3_ADDRESS_BY_CHAIN[chainId];
```
maxime-oe, line 619: Could we rename this to `MULTICALL3_ADDRESS_BY_CHAIN_FALLBACK` ? Otherwise I can very confidently predict that we will keep updating this map :D

maxime-oe (dismissed):

packages/assets-controller/src/data-sources/evm-rpc-services/clients/MulticallClient.ts
mikesposito: I can consider renaming this, but keep in mind that we'd have to update the fallback anyway at some point

packages/assets-controller/src/data-sources/RpcDataSource.test.ts
```diff
@@ -4,6 +4,7 @@ import type { NetworkState } from '@metamask/network-controller';
 import { NetworkStatus, RpcEndpointType } from '@metamask/network-controller';
 import type { TransactionMeta } from '@metamask/transaction-controller';
 
+import { ConfigRegistryControllerGetNetworkConfigByCaip2ChainIdAction } from '../../../config-registry-controller/src/ConfigRegistryController-method-action-types.js';
```
salimtb: i think we should remove the `.js` here

mikesposito: oops, good catch. Will import with the package name

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -272,8 +273,17 @@ export class RpcDataSource extends AbstractDataSource<
...
+    this.#multicallClient = new MulticallClient({
+      getProvider: (hexChainId: string): RpcProvider => {
+        return this.#getMulticallProvider(hexChainId);
+      },
+      getMulticall3AddressForChain: (hexChainId: Hex): Hex | undefined =>
+        // The messenger is not being called from a constructor, so this is safe.
+        // eslint-disable-next-line no-restricted-syntax
+        this.#messenger.call(
+          'ConfigRegistryController:getNetworkConfigByCaip2ChainId',
+          `eip155:${hexToNumber(hexChainId)}`,
```
salimtb: nit: you can use metamask utils and the function `toCaipChainId`

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -272,8 +275,20 @@ export class RpcDataSource extends AbstractDataSource<
...
+        return this.#getMulticallProvider(hexChainId);
+      },
+      getMulticall3AddressForChain: (hexChainId: Hex): Hex | undefined =>
+        // The messenger is not being called from a constructor, so this is safe.
+        // eslint-disable-next-line no-restricted-syntax
+        this.#messenger.call(
+          'ConfigRegistryController:getNetworkConfigByCaip2ChainId',
+          toCaipChainId(
+            KnownCaipNamespace.Eip155,
+            String(hexToNumber(hexChainId)),
```
mikesposito:
```suggestion
            hexToNumber(hexChainId).toString(),
```

salimtb (dismissed):

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -272,8 +275,20 @@ export class RpcDataSource extends AbstractDataSource<
     });
 
     // Initialize MulticallClient with a provider getter
-    this.#multicallClient = new MulticallClient((hexChainId: string) => {
-      return this.#getMulticallProvider(hexChainId);
+    this.#multicallClient = new MulticallClient({
+      getProvider: (hexChainId: string): RpcProvider => {
```
ulissesferreira, line 279: Would be nice if this was Caip but out of scope...

packages/assets-controller/src/data-sources/RpcDataSource.ts
```diff
@@ -272,8 +275,20 @@ export class RpcDataSource extends AbstractDataSource<
...
+        return this.#getMulticallProvider(hexChainId);
+      },
+      getMulticall3AddressForChain: (hexChainId: Hex): Hex | undefined =>
```
ulissesferreira, line 282: Do we need the explicit type?

salimtb (approved):

packages/assets-controller/src/data-sources/RpcDataSource.ts
mikesposito: Yes:

Screenshot 2026-07-31 at 16 00 30

packages/assets-controller/src/data-sources/RpcDataSource.ts
mikesposito: I required an hex chain ID instead of a Caip because `MulticallClient` seem to only use EVM hex chains, so ti will be always called with hex only

mikesposito: Oh actually that function was pre-existing (it was just moved inside an object), but we can apply the same rationale for using hexes

ulissesferreira (approved):

PR #9747, opened 2026-07-31
Feat: Seed remote feature flag controller with default flags

Cal-L:
## Explanation

This is part of an effort to keep the RemoteFeatureFlagController as the source of truth for feature flags. As part of that effort, we've added a new optional constructor arg named `defaultFeatureFlags`, which will be provided by the platform apps. Under the hood, the controller will account for these flags when processing the effective flags that the consumers will use. The order of priority for the flags are - default flags > remote flags > override flags.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Additive optional API with centralized merge logic and broad test coverage; behavior change is limited to consumers that pass defaults or rely on override removal without a remote value.
> 
> **Overview**
> Adds an optional **`defaultFeatureFlags`** constructor option so platform apps can seed **`RemoteFeatureFlagController`** with client-side defaults that are **not persisted**. Effective flags exposed in **`remoteFeatureFlags`** are now merged through **`#getEffectiveFeatureFlags`** with precedence **defaults → processed remote → local overrides**.
> 
> Remote fetch, override set/remove/clear, and constructor hydration all use that helper so defaults remain for flags the server omits, remote values win on conflicts, and removing an override falls back to remote or default when no remote value exists. The wallet initializer forwards **`instanceOptions.remoteFeatureFlagController.defaultFeatureFlags`** into the controller.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 615b161359b0a375f0ef1f626289ab859b522fd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -191,6 +191,50 @@ function redactMetaMetricsIds(flags: FeatureFlags): FeatureFlags {
...
+      delete processedRemoteFeatureFlags[flagName];
+    }
+  }
+
+  const rawRemoteKeys = new Set(Object.keys(rawRemoteFeatureFlags ?? {}));
+  for (const flagName of Object.keys(defaultFeatureFlags)) {
+    if (!rawRemoteKeys.has(flagName)) {
+      delete processedRemoteFeatureFlags[flagName];
+    }
+  }
```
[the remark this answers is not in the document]
Cal-L: Updated code to omit versioned flags if it doesn't apply

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -55,6 +56,12 @@ const remoteFeatureFlagControllerMetadata = {
     includeInDebugSnapshot: true,
     usedInUi: false,
   },
+  processedRemoteFeatureFlags: {
```
Cal-L: This persisted state is the processed version of REMOTE feature flags, excluding defaults and overrides. Used for flag reconstruction on controller creation, preventing the need for a deconstruction.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -302,6 +322,25 @@ export class RemoteFeatureFlagController extends BaseController<
     );
   }
```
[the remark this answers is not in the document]
Cal-L: While this is possible, IMO - leaving it alone is a fair trade-off for the following reasons:
- removeFlagOverride, clearAllFlagOverrides, and setFlagOverrides are not actively used in user facing features on both mobile and extension
-  The problem with `#processedRemoteFeatureFlags` is that it was never persisted and re-calculated in the constructor by removing the override values. However, that's not a true representation of what the processed remote flags either since it removed flags that could've existed in remote.
- #processedRemoteFeatureFlags resolves to its accurate value after the first fetch with this PR. The correct value would exclude both defaults and overrides

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -121,6 +128,7 @@ export function getDefaultRemoteFeatureFlagControllerState(): RemoteFeatureFlagC
     remoteFeatureFlags: {},
     localOverrides: {},
     rawRemoteFeatureFlags: {},
+    processedRemoteFeatureFlags: undefined,
```
Cal-L: Processed remote feature flags are persisted so that it can be used to synchronously reconstruct feature flags on controller initialization

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -238,6 +249,7 @@ export class RemoteFeatureFlagController extends BaseController<
     getMetaMetricsId,
     clientVersion,
     prevClientVersion,
+    defaultFeatureFlags = {},
```
Cal-L, line 244: Ability for platforms to provide feature flag defaults

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -264,32 +277,39 @@ export class RemoteFeatureFlagController extends BaseController<
...
+      initialState.processedRemoteFeatureFlags !== undefined;
+
+    // When processed remote is already persisted, use it. When missing (pre-upgrade), use current effective flags.
+    const processedRemoteFeatureFlags =
+      initialState.processedRemoteFeatureFlags ??
+      initialState.remoteFeatureFlags;
+
+    // Force a refetch when processed remote has never been persisted so the
+    // new field is populated (from the API).
+    const shouldForceRefetch =
```
Cal-L: Force refetch if processed remote feature flags do not exist

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -264,32 +277,39 @@ export class RemoteFeatureFlagController extends BaseController<
...
+      !hasPersistedProcessedRemote || hasClientVersionChanged;
+
+    // Derive effective flags from defaults + processed + overrides.
+    const remoteFeatureFlags = {
```
Cal-L: Prioritize feature flags in this order

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
davidmurdoch: > preventing the need for a deconstruction.

hm. my full time job right now is pretty much just removing persisted state that can be derived from other state. is that what this is?

Gudahtt: Could you elaborate on what this is for? `remoteFeatureFlags` is _already_ the "processed" flags. What do you mean by "preventing the need for a deconstruction"?

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -228,6 +238,7 @@ export class RemoteFeatureFlagController extends BaseController<
    * @param options.getMetaMetricsId - Returns metaMetricsId.
    * @param options.clientVersion - The current client version for version-based feature flag filtering. Must be a valid 3-part SemVer version string.
    * @param options.prevClientVersion - The previous client version for feature flag cache invalidation.
+   * @param options.defaultFeatureFlags - Client-side default feature flags used as the lowest-precedence layer under processed remote flags and local overrides. Not persisted.
```
DDDDDanica, line 233: Question: what is the different between `rawRemoteFeatureFlags` and `defaultFeatureFlags`?

Cal-L: `rawRemoteFeatureFlags` are unprocessed flags (but has metametrics IDs redacted), which are returned from the server. They include threshold arrays, which aren't yet reconciled. `defaultFeatureFlags` are defaults flags that the platforms will pass into the controller. This was previously a missing gap since there was no default option.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Cal-L: @davidmurdoch @Gudahtt I think I made these changes more complex than needed. I ended up removing this field and simplified the PR to just intake a defaultFeatureFlags arg. The state in question is no longer relevant.

weitingsun (dismissed):

DDDDDanica (dismissed):

weitingsun (dismissed):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -264,6 +269,23 @@ export class RemoteFeatureFlagController extends BaseController<
...
+    for (const [flagName, overrideValue] of Object.entries(localOverrides)) {
+      if (processedRemoteFeatureFlags[flagName] === overrideValue) {
+        delete processedRemoteFeatureFlags[flagName];
+      }
+    }
+    for (const flagName of Object.keys(defaultFeatureFlags)) {
+      if (rawRemoteFeatureFlags[flagName] === undefined) {
+        delete processedRemoteFeatureFlags[flagName];
+      }
+    }
```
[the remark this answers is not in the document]
Cal-L: Removed this logic. @davidmurdoch will be working on removed unnecessary states

weitingsun (approved):

mcmire (approved):
LGTM.

davidmurdoch (approved):

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -265,14 +270,26 @@ export class RemoteFeatureFlagController extends BaseController<
 
     const localOverrides = initialState.localOverrides ?? {};
 
+    // Rebuild the processed remote layer from last session's effective flags by
+    // stripping local overrides.
+    const processedRemoteFeatureFlags = {
```
Gudahtt, line 275: Nit: This step is unnecessary, and doesn't achieve the stated goals

There is no reason to attempt to strip out the impact of previous overrides here. And even if we wanted to, this would not achieve that goal (because we don't know what the overrides were when this state was generated).

This code is assuming the overrides are unchanged. But if that was true, why remove them only to re-apply them immediately afterwards? The only reason I can foresee for re-applying overrides is to accommodate a change in overrides, which violates the assumption made here.

Gudahtt: If we _really_ want to accommodate making the initial state of `remoteFeatureFlags` correct _even when overrides change_, we need to recalculate the feature flags from the _raw_ values. Not from the persisted processed values.

If this is not a priority, we can simply delete this step instead.

Gudahtt: I see this was a pre-existing bug though, just moved by this PR.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -586,7 +608,10 @@ export class RemoteFeatureFlagController extends BaseController<
       return {
         ...this.state,
         localOverrides: {},
-        remoteFeatureFlags: { ...this.#processedRemoteFeatureFlags },
+        remoteFeatureFlags: this.#getEffectiveFeatureFlags(
+          this.#processedRemoteFeatureFlags,
+          {},
```
Gudahtt, line 613: Nit: This is another pre-existing bug: `this.#processedRemoteFeatureFlags` _may already contain overrides_, because the attempt to strip them out int he constructor is ineffective. We must construct the feature flags from the raw values to remove overrides.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -560,20 +588,14 @@ export class RemoteFeatureFlagController extends BaseController<
...
-    } else {
-      remoteFeatureFlags[flagName] = processedValue;
-    }
-
     this.update(() => {
       return {
         ...this.state,
         localOverrides: newLocalOverrides,
-        remoteFeatureFlags,
+        remoteFeatureFlags: this.#getEffectiveFeatureFlags(
```
Gudahtt, line 595: Nit: Another pre-existing bug, where we're attempting to remove overrides incorrectly

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -388,10 +417,9 @@ export class RemoteFeatureFlagController extends BaseController<
...
       return {
         ...this.state,
-        remoteFeatureFlags: {
-          ...redactedProcessedFlags,
-          ...this.state.localOverrides,
-        },
+        remoteFeatureFlags: this.#getEffectiveFeatureFlags(
+          redactedProcessedFlags,
+        ),
         rawRemoteFeatureFlags: redactMetaMetricsIds(remoteFeatureFlags),
```
Gudahtt, line 423: Ah, this explains the other bugs.

Redacting IDs here makes it impossible to use `rawRemoteFeatureFlags` for any useful work. This is a mistake, and is probably what motivated the flawed attempts at removing overrides. We must redact IDs later at the point of collection, not here. We need complete raw flags in order to effectively work with overrides.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Cal-L: +1 - Since @davidmurdoch mentioned that he's working on updating these states, I'll reach out to coordinate on this. I contemplated on doing a refactor but is out of scope for this PR.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Cal-L: +1 - Since @davidmurdoch mentioned that he's working on updating these states, I'll reach out to coordinate on this. I contemplated on doing a refactor but is out of scope for this PR.

PR #9765, opened 2026-08-03
fix: preserve Money Account EIP-7702 auths on same-chain pay batches

jpuri:
## Summary
- Add optional `authorizationList` to `TransactionBatchRequest` and merge it with the batch payer (`from`) upgrade authorization when building the type-4 batch transaction.
- Retain pre-signed EIP-7702 authorizations in `signAuthorizationList` instead of re-signing them with `txParams.from`.
- In Relay same-chain batch submit with account override, forward the quote's full signed `authorizationList` through `addTransactionBatch` so Money Account vault upgrades are not dropped (CONF-1741 / Monad mUSD deposits).

Related to task: https://consensyssoftware.atlassian.net/browse/CONF-1757

## Test plan
- [x] Unit: `@metamask/transaction-controller` — `batch.test.ts` / `eip7702.test.ts` (merge provided auths; retain pre-signed)
- [x] Unit: `@metamask/transaction-pay-controller` — `relay-submit.test.ts` (batch gets signed auth list only when same-chain + account override; single-tx path unchanged)
- [x] Manual (with mobile Gas Station merge): new user Money Account deposit, pay with Monad mUSD, same-chain batch path — Sentinel request includes 2 authorizations (EOA + Money Account) and vault leg succeeds
- [x] Regression: single-tx same-chain route and cross-chain routes unchanged

---

> [!NOTE]
> **Medium Risk**
> Touches EIP-7702 signing, batch assembly, and nonce attribution on the transaction publish path; incorrect authority recovery could skew nonces, but changes are scoped with unit tests and limited relay batch conditions.
> 
> **Overview**
> Fixes same-chain MetaMask Pay relay batches when an EOA pays for a Money Account deposit: **Money Account vault EIP-7702 authorizations are kept on the type-4 batch** instead of being dropped or re-signed as the payer.
> 
> **Transaction controller:** `TransactionBatchRequest` gains optional `authorizationList`. Batch building merges caller-provided auths with the batch payer’s upgrade auth when needed; security `delegationMock` only uses the payer’s upgrade entry, not foreign auths. `signAuthorizationList` **skips re-signing** entries that already have a full signature. Nonce tracking **recovers each authorization’s authority** (new `recoverAuthorizationAuthority` / `getAuthorizationAuthority`) so another account’s auth nonces do not count against the payer.
> 
> **Transaction pay controller:** On same-chain multi-tx relay submit with **account override**, the quote’s **fully signed** `authorizationList` is forwarded into `addTransactionBatch`; single-tx and non-override paths stay as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10ffd51d56faf5d413172dad59ffb868b6771455. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-controller/src/utils/batch.ts
```diff
@@ -288,6 +289,58 @@ async function getNestedTransactionMeta(
...
+
+    if (!upgradeContractAddress) {
+      throw rpcErrors.internal(ERROR_MESSAGE_NO_UPGRADE_CONTRACT);
+    }
+
+    authorizationList.push({ address: upgradeContractAddress });
+  }
+
+  if (providedAuthorizationList?.length) {
+    authorizationList.push(...providedAuthorizationList);
```
pedronfigueiredo, line 332: Appending the caller-provided list allows this transaction to contain authorization nonces belonging to accounts other than `from`. `getAndFormatTransactionsForNonceTracker` currently flattens every authorization nonce into records whose `from` is the outer transaction sender, but EIP-7702 recovers and increments each authorization authority independently. For example, an EOA transaction with nonce `5`, an EOA authorization at `6`, and a Money Account authorization at `20` is currently represented as three EOA nonces, potentially making the next EOA transaction jump to `21` instead of `7`. Please associate authorization nonces with their recovered authorities—or at minimum exclude foreign authorizations from the payer’s nonce history—and add mixed-authority nonce-tracker coverage.

packages/transaction-controller/src/utils/batch.ts
```diff
@@ -288,6 +289,58 @@ async function getNestedTransactionMeta(
  * @param request - The request object including the user request and necessary callbacks.
  * @returns The batch result object including the batch ID.
  */
+/**
+ * Build the authorization list for an EIP-7702 batch transaction.
```
pedronfigueiredo, line 287: The “Process a batch transaction…” JSDoc describes `addTransactionBatchWith7702`, but inserting the new helper and its JSDoc here has detached that block from the function it documents. This leaves two documentation blocks above `buildBatchAuthorizationList` and leaves `addTransactionBatchWith7702` undocumented. Please move the original block so that it immediately precedes `addTransactionBatchWith7702`.

pedronfigueiredo (changes requested):

packages/transaction-controller/src/utils/batch.ts
jpuri: PR is updated to address this

packages/transaction-controller/src/utils/batch.ts
jpuri: PR is updated to address this

pedronfigueiredo (dismissed):

pedronfigueiredo (approved):

PR #9768, opened 2026-08-03
fix(assets-controller): exempt user-initiated activity from spam filtering

salimtb:
Occurrence-floor / Blockaid spam filtering now applies only to passively discovered tokens (auto-detection, incoming-only websocket transfers such as airdrops). Assets acquired through user-initiated activity — the account sent funds in the same transaction, e.g. swap outputs — are listed on a new DataResponse.userInteractedAssets field by AccountActivityDataSource and bypass the filter in TokenDataSource, so they always show even when below the per-chain occurrence floor.

Filtered-out spam is now also stripped from stub assetsInfo on the response, and balance deletions match asset IDs case-insensitively.

## Checklist


Prithpal-Sooriya:
Hmm, I think the intention it correct here, but this still impacts:
- receiving assets (ensure no regression if you were to send from a diff wallet)
- metadata override
- no occurrence filtering

PR #9801, opened 2026-08-07
feat: add getPaymentMethodsForContext to ramps-controller

saustrie-consensys:
## Explanation

Money account "Pay with" previously read Buy-scoped payment methods even though Money quotes use a deposit asset and a different eligible-provider set. That mismatch could offer methods such as Revolut Pay that cannot produce a Money deposit quote (TRAM-3838 / metamask-mobile#34109).

### Shared context API and provenance

`getPaymentMethodsForContext` is a new shared Core API. It combines payment-method fetch and automatic-selection behavior that already existed in Mobile UB2 with Core's existing provider-resolution helpers, while allowing callers to choose whether the result updates Buy state.

`#resolveProviderIdsForPaymentMethods` is a new adapter over existing Core behavior. It reuses `#resolveAllProvidersFlag`, `#filterProviderIdsBySupport`, `#getSupportingProvidersForRegion`, and `#resolveProviderIdsForQuote` instead of creating a separate provider policy.

`mergePaymentMethodsById` was newly designed for multi-provider fan-out. It was not copied from Mobile.

`normalizeRampsAssetId` is not new behavior. Mobile has carried the same rule since April 2026 as `normalizeAssetIdForApi` (metamask-mobile#29037). This PR promotes it into Core's public API so the rule lives in one place and Mobile can drop its copy. `providerServesAsset` now uses it too, which is a behavior change: non-EVM CAIP-19 references are no longer lowercased, since Solana base58 and bitcoin bech32 references are case-sensitive and lowercasing both sides could report a provider as serving an asset it does not serve.

### Merge scope

`mergePaymentMethodsById` returns a single provider list unchanged by identity. It preserves the provider's array, row references, ordering, length, and duplicate IDs, so single-provider consumers such as standalone Buy keep verbatim provider output just as legacy `getPaymentMethods` did.

Deduping runs only when two or more provider lists are combined, and it keeps the first-seen entry per canonical ID. There is no field-level collision policy, because payment-method metadata is provider-invariant: the API serves it from a per-region catalog (`PaymentMethodsList.ts`) and the `provider` query narrows that catalog by ID-set membership (`RegionsV2Service.ts`, `providerPaymentIds.has(payment.id)`) without rewriting fields. Two providers returning the same payment ID therefore return identical `score`, `delay`, and `name`. The same data is already deduped first-seen-wins by the API itself for its multi-region path.

Provider fan-out does vary array order, because the API sorts by reliability per request. That affects ordering, not field values, and `mergePaymentMethodsById` preserves first-seen encounter order.

### Stateful write guards

`updateState: true` reproduces legacy Buy catalog and selection writes, but it is guarded because the controller mutates shared state from inside the query function. A consumer's query cache cannot undo that side effect after a stale request commits.

The write requires the stored region, selected asset, and selected provider to match the request context. Stored region is trimmed and lowercased before comparison, and a missing region remains fail-closed. EVM asset IDs are compared case-insensitively while non-EVM asset case is preserved.

A region-only stale write is reachable when an older region request remains in flight, the user changes region, and the new region repopulates the same asset and provider before the older response returns. Legacy behavior checked asset and provider but not region. These guards close that latent gap rather than adding a cache-only safeguard.

Two same-context stateful requests cannot be ordered by those guards, because the context key falls back to the globally selected provider when more than one provider is passed. Rather than order them with a request sequence, `updateState: true` now throws when provider resolution yields more than one provider, before any fetch is issued. The shared Buy catalog has one slot per region, token, and selected provider, so a fan-out write has nowhere unambiguous to land. Callers wanting a multi-provider catalog pass `updateState: false` and read the returned `methods`. UB2 already passes exactly one explicit selected provider, and MMPay deposit contexts are read-only, so no current caller is affected.

An earlier revision of this PR ordered these writes with a per-request sequence counter instead. It was removed: the race it guarded is unreachable for a single-provider write (the state context key and the fetch cache key move together, so concurrent duplicates collapse in `executeRequest`'s pending dedupe), and the counter was the source of both Bugbot findings on this PR, the second introduced by the fix for the first.

### Catalog writers

`paymentMethods.data` and `paymentMethods.selected` have two writers in Core: the pre-existing `getPaymentMethods` and the new `getPaymentMethodsForContext`. Neither one's write guard is aware of the other, so a client calling both for the same context has always been able to interleave them.

The consumer PR removes Mobile's last call to `getPaymentMethods` (it was in `app/components/UI/Ramp/queries/paymentMethods.ts` on `main`), so once both land, `getPaymentMethodsForContext` is Mobile's only writer of that state. `getPaymentMethods` stays exported and unchanged for other consumers.

### Scope boundary

This PR does not include `getSmartSelectedQuote` or `applyQuoteContext`. Payment-method context is the minimum shared boundary required by the implemented Money and UB2 consumers. Quote choice and quote-state application still have consumer-specific orchestration and remain deferred to the tracked UB2 dumbification follow-up (TRAM-3856).

## References

* Fixes: TRAM-3838
* Related: https://github.com/MetaMask/metamask-mobile/issues/34109
* Implemented and tested consumer: https://github.com/MetaMask/metamask-mobile/pull/35346 (supersedes #34469)
* Final consumer preview: `@metamask-previews/ramps-controller@20.0.0-preview-a2a428472`
* Final Core commit: `a2a428472d39421ffb5aaa04789bfdc45f573bd1`
* Preview publication: https://github.com/MetaMask/core/pull/9801#issuecomment-5413123492
* Related tech debt: TRAM-3856 (UB2 dumbification follow-up under TRAM-3460)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches payment-method catalog state, provider eligibility matching, and new concurrent-write guards; behavior change for non-EVM asset matching could affect which providers are considered to serve an asset.
> 
> **Overview**
> Adds **`getPaymentMethodsForContext`** (and the messenger action) so Money/headless flows can load payment methods for a **deposit asset and provider set** using the same resolution rules as **`getQuotes`**, without defaulting to the Buy catalog’s selected token/provider. By default the call is **request-only**; optional **`updateState: true`** writes `paymentMethods` only when region, selected token, and selected provider still match, and **throws** if resolution yields more than one provider.
> 
> Multi-provider fetches **fan out**, **merge** via new **`mergePaymentMethodsById`** / **`pickPaymentMethod`**, and tolerate partial failures. **`normalizeRampsAssetId`** is exported and used for comparisons; **`providerServesAsset`** / **`getProvidersServingAsset`** now **stop lowercasing non-EVM** CAIP-19 ids so Solana/bitcoin references stay case-sensitive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aefeed75a64c459459e5ee1c2d965a4be2924228. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
@metamaskbot publish-previews

packages/ramps-controller/src/RampsController.ts
```diff
@@ -274,6 +275,13 @@ function normalizeRampsErrorForRethrow(
   });
 }
 
+function normalizeAssetIdForContext(assetId: string): string {
```
amitabh94: Do we not have a util somewhere to normalize asset ?

packages/ramps-controller/src/RampsController.ts
```diff
@@ -807,6 +839,7 @@ const MESSENGER_EXPOSED_METHODS = [
   'setSelectedToken',
   'getProviders',
   'getPaymentMethods',
+  'getPaymentMethodsForContext',
```
amitabh94, line 840: what is this method supposed to be doing?

saustrie-consensys:
@metamaskbot publish-previews

amitabh94:
Is there some code in this PR that can be done on the backend or does it need to be on core?

packages/ramps-controller/src/RampsController.ts
saustrie-consensys: https://github.com/search?q=repo%3AMetaMask%2Fmetamask-mobile+normalizeAssetIdForApi&type=code

It used to be done in the frontend mobile repo. Moving it to the core repo

packages/ramps-controller/src/RampsController.ts
saustrie-consensys: getPaymentMethodsForContext gets the payment method based on what params are passed (region, asset, provider set), instead of reading whatever the Buy flow happens to have selected globally. Makes it more abstract so we can beyond UB2, MMPay, etc. @amitabh94

saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
> Is there some code in this PR that can be done on the backend or does it need to be on core?

mergePaymentMethodsById (and maybe the Promise.allSettled fan-out before it) could move to the backend. Either by modifying [getPaymentsV2](https://github.com/consensys-vertical-apps/va-mmcx-onramp-api/blob/e5a811c44d93dd77efe6b2cb94ceae7ca7df6abc/src/modules/regions-v2/RegionsV2Controller.ts#L500) to accept a repeated provider (drop [assertAtMostOneFilterValuePerQueryKey](https://github.com/consensys-vertical-apps/va-mmcx-onramp-api/blob/e5a811c44d93dd77efe6b2cb94ceae7ca7df6abc/src/modules/regions-v2/RegionsV2Controller.ts#L509) and widen one type in [getPaymentsWithRegions](https://github.com/consensys-vertical-apps/va-mmcx-onramp-api/blob/e5a811c44d93dd77efe6b2cb94ceae7ca7df6abc/src/modules/regions-v2/RegionsV2Service.ts#L1403)), or by using [getRegionV2](https://github.com/consensys-vertical-apps/va-mmcx-onramp-api/blob/e5a811c44d93dd77efe6b2cb94ceae7ca7df6abc/src/modules/regions-v2/RegionsV2Controller.ts#L866), which already takes a provider array and returns the same combined list. I looked into it out yesterday, might be worth doing it as a future follow up ticket, since this is becoming a lot of tech debt being handled for one Revolut ticket at this point. Handling more refactoring increases time by a lot since I'm carefully trying to minimize side effects.

The rest wouldn't be easily moveable. #resolveProviderIdsForPaymentMethods and #resolveAllProvidersFlag need moneyHeadlessAllProviders and its allowlist. The updateState write path needs userRegion.regionCode, tokens.selected, providers.selected, and paymentMethods.data / .selected, all of which live in Core state.

@amitabh94

saustrie-consensys:
@metamaskbot publish-previews

saustrie-consensys:
@metamaskbot publish-previews

packages/ramps-controller/src/providerAvailability.ts
```diff
@@ -1,15 +1,37 @@
...
+ * so lowercasing is restricted to EVM ids rather than applied to the whole
+ * string.
+ *
+ * The namespace itself is matched case-insensitively, so an `EIP155:` prefix
+ * is treated as EVM even though CAIP-19 defines namespaces as lowercase.
+ *
+ * @param assetId - CAIP-19 asset id, in either case form.
+ * @returns The asset id in the form the Ramps API uses.
+ */
+export function normalizeRampsAssetId(assetId: string): string {
```
amitabh94, line 28: Isnt Ramps api giving out caip id?

amitabh94 (dismissed):

wenfix (approved):

PR #9816, opened 2026-08-11
chore: Fix broken override feature on the remote feature flag controller

Cal-L:
## Explanation

This PR fixes issues related to incorrectly setting and removing flags based on overrides. Prior to this change, the persisted `remoteFeatureFlags` needed to cleaned prior to effective flag reconstruction, which meant overrides needed to be removed first. This led to broken behavior where flags were completely removed instead of falling back to remote feature flags. We're now leveraging the existing `rawRemoteFeatureFlags` to reconstruct effective flags. To make this happen, we also removed metametrics ID redacting. For context, there's there's a scenario where QA/PMs/Devs can provide their own metametrics IDs for specific device targeting. That being said, the IDs do not belong to users.

## References

Fixes https://consensyssoftware.atlassian.net/browse/MCWP-774

## Checklist

---

> [!NOTE]
> **High Risk**
> Breaking API (`init()` required), changed persistence semantics for raw flags, and core flag/override behavior affects every consumer of remote feature flags.
> 
> **Overview**
> Fixes **local override removal** by keeping a separate processed remote layer (`#processedRemoteFeatureFlags`) instead of inferring remote values by subtracting overrides from persisted `remoteFeatureFlags`. **Removing or clearing overrides** now restores the underlying remote (or default) value instead of dropping the flag.
> 
> Adds a **breaking** `RemoteFeatureFlagController.init()` that must run after construction. It re-processes persisted `rawRemoteFeatureFlags` (threshold selection, version gating, explicit `metaMetricsIds` targeting) and merges defaults + overrides into `remoteFeatureFlags`. The constructor no longer performs that merge; without `init`, effective flags may be incomplete until it runs.
> 
> **Stops redacting `metaMetricsIds`** from `rawRemoteFeatureFlags` (and removes `redactMetaMetricsIds`) so restart can re-derive targeting; changelog notes consumers may need to clear stale redacted `rawRemoteFeatureFlags` from persisted state.
> 
> Wallet integration tests expect one more `init()` participant (`Wallet.test` init count 6); wallet `defaultFeatureFlags` test now awaits `instance.init()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70345027de553b24c63a522a28733fa46f38b0a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add optional `defaultFeatureFlags` constructor option to `RemoteFeatureFlagController` for client-side defaults as the lowest-precedence layer under processed remote flags and local overrides ([#9747](https://github.com/MetaMask/core/pull/9747))
+- Add `RemoteFeatureFlagController.init` method, which re-evaluates version and threshold selection against `rawRemoteFeatureFlags` already in state and recomputes `remoteFeatureFlags` without making a network request ([#XXXX](https://github.com/MetaMask/core/pull/XXXX))
```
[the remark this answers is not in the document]
Cal-L: Fixed

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -371,59 +338,62 @@ export class RemoteFeatureFlagController extends BaseController<
...
+    this.update(() => {
+      return {
+        ...this.state,
+        remoteFeatureFlags: this.#getEffectiveFeatureFlags({
+          processedRemoteFeatureFlags: resolved.processedFlags,
+        }),
+        thresholdCache: resolved.thresholdCache,
+        featureFlagThresholdGroups: resolved.featureFlagThresholdGroups,
+      };
+    });
```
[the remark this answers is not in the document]
Cal-L: I'm thinking of leaving this alone for the following reasons:
- Only affects devs who are still using the metametrics IDs targeting in Launch Darkly
- Would require persisting another state for triggering a force refetch on this controller version
- After 24 hrs and calling `updateRemoteFeatureFlags` will resolve this

weitingsun (dismissed):

DDDDDanica (dismissed):

packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add optional `defaultFeatureFlags` constructor option to `RemoteFeatureFlagController` for client-side defaults as the lowest-precedence layer under processed remote flags and local overrides ([#9747](https://github.com/MetaMask/core/pull/9747))
+- Add `RemoteFeatureFlagController.init` method, which re-evaluates version and threshold selection against `rawRemoteFeatureFlags` already in state and recomputes `remoteFeatureFlags`. This also fixes issues around setting overrides properly. ([#9816](https://github.com/MetaMask/core/pull/9816))
```
Gudahtt: This should be documented as breaking, since the controller won't work properly unless `init` is called.

It's also not really a new feature, so it should go under "Changed" rather than "Added".

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -290,7 +251,7 @@ export class RemoteFeatureFlagController extends BaseController<
         remoteFeatureFlags: {
```
Gudahtt: Nit: This doesn't _strongly_ matter, but I would recommend removing this `remoteFeatureFlags` entry altogether and just leaving that property alone. There isn't really any benefit to re-applying defaults and overrides here, since it gets re-derived and overwritten in `init` anyway. It'll just waste a few CPU cycles.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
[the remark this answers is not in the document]
Gudahtt: Hmm. This could still be very disruptive if it _does_ end up impacting users though. It would temporarily enable or disable a feature flag unexpectedly until the next update.

It would be easiest to handle with a client migration though, by deleting `rawRemoteFeatureFlags`.

Maybe we can leave a note in the changelog about this?

packages/remote-feature-flag-controller/CHANGELOG.md
Cal-L: Updated to be part of Changed and added breaking note.

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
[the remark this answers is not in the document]
Cal-L: That's a sound solution. Included a note about this in the change logs

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
Cal-L: Removed. We're now just relying on `init` for reconciling `remoteFeatureFlags`

packages/remote-feature-flag-controller/src/remote-feature-flag-controller.ts
```diff
@@ -371,59 +331,61 @@ export class RemoteFeatureFlagController extends BaseController<
...
-    const metaMetricsId = this.#getMetaMetricsId();
-    const currentFlagNames = Object.keys(remoteFeatureFlags);
+    const resolved = hasRawRemoteFeatureFlags
+      ? await this.#processRemoteFeatureFlags(rawRemoteFeatureFlags)
+      : undefined;
 
-    // Build updated threshold cache
-    const updatedThresholdCache = { ...(this.state.thresholdCache ?? {}) };
+    this.#processedRemoteFeatureFlags =
+      resolved?.processedFlags ?? this.state.remoteFeatureFlags;
```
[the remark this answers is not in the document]
Gudahtt: This Cursor comment is correct, but this is still the best option available to us.

Yes, it's true that removing overrides won't work properly until the first fetch. But defaulting to the previous processed flags is the only way to prevent flags from flipping on or off for the end-user, which would be far worse than "remove override" not working properly. It is a dev-only feature after all.

packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Changed
 
-- Add optional `defaultFeatureFlags` constructor option to `RemoteFeatureFlagController` for client-side defaults as the lowest-precedence layer under processed remote flags and local overrides ([#9747](https://github.com/MetaMask/core/pull/9747))
+- **BREAKING:** Add `RemoteFeatureFlagController.init` method, which re-evaluates version and threshold selection against `rawRemoteFeatureFlags` already in state and recomputes `remoteFeatureFlags`. This also fixes issues around setting overrides properly. Add optional `defaultFeatureFlags` constructor option to `RemoteFeatureFlagController` for client-side defaults as the lowest-precedence layer under processed remote flags and local overrides. Clients to this version should delete `rawRemoteFeatureFlags` from persisted state in a migration. Previous versions stripped `metaMetricsIds` before persisting, so re-deriving from those payloads finds no explicit match and falls back to hash-based selection, moving explicitly targeted users off their variant until the next fetch. ([#9816](https://github.com/MetaMask/core/pull/9816))
```
Gudahtt: Huh, why did these two entries get combined 🤔 I don't see why we'd group together the `defautFeatureFlags` option and the `init` function.

I would probably word it something like this:
```
### Added

- Add optional `defaultFeatureFlags` constructor option to `RemoteFeatureFlagController` for client-side defaults as the lowest-precedence layer under processed remote flags and local overrides ([#9747](https://github.com/MetaMask/core/pull/9747))

### Changed

- **BREAKING:** Add `RemoteFeatureFlagController.init` method ([#9816](https://github.com/MetaMask/core/pull/9816))
  * This must be called during initialization to ensure `remoteFeatureFlags` is properly recomputed
- **BREAKING:** Stop redacting IDs from `rawRemoteFeatureFlags` [#9816](https://github.com/MetaMask/core/pull/9816))
  - Existing `rawRemoteFeatureFlags` properties should be deleted in a migration, so they do not get used for recomputing flags (which would not work properly with a redacted input).

### Fixed

- Restore remote flag value when overrides are removed/cleared ([#9816](https://github.com/MetaMask/core/pull/9816))
  - Previously the underlying remote value would be removed as well
```

Cal-L: Updated!

weitingsun (approved):

Gudahtt (approved):
LGTM!

PR #9817, opened 2026-08-11
fix: Add `resolveSourceAmount` callback

OGPoyraz:
## Explanation

For a Money Account Max deposit (e.g. "Send from Money Account → Perps"), `calculateSourceAmount` derives the Relay quote's source `amount` as `token.amountUsd / usdRate * 10^decimals`. Since mUSD's `usdRate` is just under `$1`, this inflates the mUSD source above the on‑chain redeemable, so the Relay simulation reverts (the error scales with amount, hence the intermittent failures). The exact redeemable (bare mUSD + vaulted vmUSD) is only known to the client.

This adds an optional, generic `resolveSourceAmount` constructor callback: `(request) => { sourceAmountRaw } | undefined`. It's consulted in `calculateSourceAmount` after the existing skip guards and before the fiat gross‑up; a returned raw value is used verbatim (human derived from it), and `undefined` falls through unchanged. Opt‑in per return value, so existing flows are unaffected and no new state fields are added. Also exports the related types.

## References

- Fixes CONF-1787 — [MA → Perps] max perps deposit with Money Account doesn't always work
- Consumer: MetaMask/metamask-mobile implements the callback to return the on‑chain redeemable for Money Account Max

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how quote source amounts are computed on an opt-in path in transaction-pay quoting; incorrect callback values could cause failed simulations or wrong pay amounts, though default flows are unchanged.
> 
> **Overview**
> Adds an optional **`resolveSourceAmount`** constructor hook on `TransactionPayController` so clients can supply an exact atomic source amount for Relay quotes instead of the default **fiat-derived** calculation (`amountUsd / usdRate`).
> 
> The callback is synchronous, receives `{ isMaxAmount, paymentOverride }`, and returns `{ sourceAmountRaw }` or **`undefined`** to keep existing behavior. `TransactionPayController` forwards it into **`updateSourceAmounts`** / **`calculateSourceAmount`**, which consult it **after** skip guards (same-token, zero amount) and **before** fiat math and max-balance logic; a returned raw value is used verbatim with human amount derived from payment token decimals.
> 
> Related types are exported from the package public API. **`calculateSourceAmount`** also moves the zero-amount check earlier so **`resolveSourceAmount` is not called** for skipped tokens. Tests cover forwarding from the controller, verbatim use, fallback, and guard behavior; changelog documents the new option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9833a9e3e58fa41596ae1e7b907888b5cc90d41f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/types.ts
```diff
@@ -215,6 +215,43 @@ export type GetAmountDataCallback = (
...
+
+  /** Selected payment (source) token. */
+  paymentToken: TransactionPaymentToken;
+
+  /** Optional payment source override for the transaction. */
+  paymentOverride?: PaymentOverride;
+
+  /** Required (target) token to be covered. */
+  token: TransactionPayRequiredToken;
+};
```
jpuri, line 225: If we just pass may be transaction object and transactionMeta here ?

OGPoyraz: Agreed it's relevant for callbacks like `getDelegationTransaction` but for specific to MA override, `isMax` and `paymentOverride` is in fact enough for now. (I will remove `paymentToken` and `token` thanks for the ping)

Shall we add that in future if we need `transaction`?

jpuri: ok, lets keep it as future task

OGPoyraz: Mobile callback: https://github.com/MetaMask/metamask-mobile/pull/34590/files#diff-8846dd9b37d5f76694facdfd1f952ecaa1f3c4235e176f3f07f3943d65577ab5

jpuri (dismissed):

jpuri (approved):

PR #9826, opened 2026-08-11
feat(account-tree-controller): add payload/snapshot support

ccharly:
## Explanation

First part of the new `:{import,export}State` actions for the account-tree.

This PR only exposes the new payload/snapshot types that will be used to import/export.

## References

- https://github.com/MetaMask/core/pull/9663

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches secret-bearing payload shapes and validation redaction for mnemonics/private keys; behavior is new and isolated to types/tests until controller actions land.
> 
> **Overview**
> Introduces a new `state/` layer for upcoming account-tree **import/export**: a versioned **`AccountTreePayload`** (v1, mnemonic + private-key wallets), **`IdMap`** for local controller IDs ↔ stable `wallet:…` payload IDs, and an immutable **`AccountTreeSnapshot`** for serialize/deserialize and selective export via `filterWallets` / `filterGroups` / `filterAllGroups`.
> 
> **`payload.ts`** defines portable wallet/group shapes (optional secrets), Superstruct validation with **fail-closed** version pinning, contiguous mnemonic group indices, and **`sensitive()`** on mnemonic/private-key fields so validation errors do not leak secrets. **`utils`** adds byte encoding helpers, `deepFreeze`, and safe validation error formatting.
> 
> Controller **`exportState` / `importState`** are not wired in this PR—only the types and snapshot machinery plus tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7aaee50b2ef079fc6b71211327f1385d09f9f714. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ccharly:
Using `no-changelog` for this one, since it's purely internal for the moment. I'll make sure to update the main PR with this PR reference:
- https://github.com/MetaMask/core/pull/9663

See: https://github.com/MetaMask/core/pull/9663/commits/9e0e993a75dfd4a7394e6582ebc152c4c800008d

packages/account-tree-controller/src/state/id-map.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Registers a local↔payload ID pair.
+   *
+   * @param localId - Local controller wallet or group ID.
+   * @param payloadId - Corresponding portable payload ID.
+   */
+  add(localId: LocalId, payloadId: PayloadId): void {
+    this.#localToPayload.set(localId, payloadId);
+    this.#payloadToLocal.set(payloadId, localId);
+  }
```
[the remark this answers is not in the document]
ccharly: We're never supposed to have any overwritten IDs. Added some `assert` to make this more explicit (about our invariants):

https://github.com/MetaMask/core/pull/9826/commits/8bca2e7fff94f2553be6657d02c278744d68b7f4

packages/account-tree-controller/src/state/snapshot.ts
```diff
@@ -0,0 +1,233 @@
...
+      };
+    } else {
+      filteredEntries[walletIndex] = {
+        ...wallet,
+        groups: filteredGroups as AccountWalletPrivateKeyGroupEntry[],
+      };
+    }
+
+    return new AccountTreeSnapshot(filteredEntries, this.#idMap);
+  }
```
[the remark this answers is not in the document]
ccharly: It feels wrong to address this during filtering. The payload will get validated by the receiver, so we let this constraint to the caller/producer of the payload. Added some more comments/jsdocs to explain this:

https://github.com/MetaMask/core/pull/9826/commits/e2009c40cdf999b4520316e7c4dc9531a4c37102

packages/account-tree-controller/src/state/payload.ts
```diff
@@ -0,0 +1,352 @@
...
+ * Superstruct schema for a fully versioned {@link AccountTreePayload}.
+ *
+ * Pins `version` to exactly {@link ACCOUNT_TREE_PAYLOAD_CURRENT_VERSION} so
+ * payloads from newer clients (v2+) are rejected rather than silently
+ * mis-handled as v1. Once a migration framework is wired up in
+ * {@link AccountTreeSnapshot.deserialize}, older versions will be up-migrated
+ * before this struct is checked, and newer versions will require a new struct.
+ *
+ * Secret fields (`value`, `privateKey`) use the Superstruct `sensitive()`
+ * wrapper so validation failures redact secrets from error output.
```
gantunesr, line 313: maybe I missed this, are there any tests about error message redaction when it comes to sensitive material?

packages/account-tree-controller/src/state/payload.ts
```diff
@@ -0,0 +1,352 @@
...
+  subId: string | number,
+): AccountGroupPayloadId {
+  return `${walletId}/${subId}`;
+}
+
+/** Options accepted by {@link AccountTreeController.exportState}. */
+export type ExportStateOptions = {
+  /** When `true`, secrets (mnemonic / private keys) are included. Requires the vault to be unlocked. */
+  includeSecrets?: boolean;
+};
```
gantunesr, line 215: do you think there is any explicit cleanup/teardown to do when secrets are exported?

gantunesr (commented):
Looks good overall

- Would be good to have an integration test for the flow export→filter→serialize→deserialize. This is probably covered in the next PR for the controller wiring

packages/account-tree-controller/src/state/payload.ts
ccharly: To like clearing them out of memory?

Like you'd like to have them being passed as options to control this?

packages/account-tree-controller/src/state/payload.ts
ccharly: So yes, all sensitive fields use `sensitive` (from our own fork of `superstruct`)

However, I thought I added explicit redaction tests for the payload here, but I don't see any 🤔

I'll add more! Also, while we assert the shape of the tree payload, all `superstruct` errors are wrapped and we use `formatValidationErrorMessages` that do not log any actual values, only the "field path". It's a 2nd layer of safety net 😄

packages/account-tree-controller/src/state/snapshot.ts
```diff
@@ -0,0 +1,242 @@
...
+   * The returned snapshot has no ID map — {@link toLocalId} / {@link toPayloadId}
+   * return `undefined`. Pass an {@link IdMap} to the constructor when you need
+   * the map.
+   *
+   * @param raw - Unknown value to parse.
+   * @returns A validated snapshot.
+   * @throws If `raw` is not a valid payload or its version is unsupported.
+   */
+  static async deserialize(raw: unknown): Promise<AccountTreeSnapshot> {
+    // TODO: Use migration framework here.
```
ccharly, line 238: Decided to omit the migration part for now. My plan was to use our new generic migration framework that we created for keyrings. Though, it's not needed here (we expect a v1 payload + the migration framework needs a slight rework around version handling + flattening the state/data being migrated).

Delaying this work for a bit to not block this PR.

packages/account-tree-controller/src/state/payload.ts
ccharly: Added even more tests: https://github.com/MetaMask/core/pull/9826/commits/585b0455ff5674c4b22ec006bce6e8a7dcbb5784

If you think of any others I can add, let me know! For now we have multiple layers:
1. `superstruct.sensitive` is redacting values (even if a sibling validation fails, you cannot extract the original value out of the validation tree)
2. `formatValidationErrorMessages` which only displays the field-path, no value (similar to our new `withSafeError` in the `SnapAccountService`

Those new tests are also here to future-proof any change in those areas as well. So if one of the 2 layers above change, the tests will fail too.

packages/account-tree-controller/src/state/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  it('uses type/refinement and never failure.message', () => {
+    // sensitive() redacts the value to *** in failure.message — if
+    // formatValidationErrorMessages were to use message instead of type, *** would
+    // appear in the output. Asserting it does not pins the mechanism.
+    const schema = object({ secret: sensitive(string()) });
+    const error = makeStructError({ secret: 12345 });
+    const result = formatValidationErrorMessages(error);
+    expect(result).toContain('expected: string');
+    expect(result).not.toContain('***');
+  });
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/9826/commits/d611e59a5ed8298d17724ffd60a7f0131b13d320

packages/account-tree-controller/src/state/payload.ts
ccharly: Added some more tests to reflect the 2 layers of redaction we have in place: https://github.com/MetaMask/core/pull/9826/commits/ee63cf4a28cefb9556d7d2fb050ede8cfdccfb31 (lint fix: https://github.com/MetaMask/core/pull/9826/commits/e5770675d2f275bc9701953c8944f54f35ae594a + cursor issue: https://github.com/MetaMask/core/pull/9826/commits/d611e59a5ed8298d17724ffd60a7f0131b13d320)

ccharly:
> Looks good overall
> 
> * Would be good to have an integration test for the flow export→filter→serialize→deserialize. This is probably covered in the next PR for the controller wiring

We indeed have some round-trips "integration" tests on the controller itself, to make sure the combo `:exportState` and `:importState` can be used together.

If needed, I'll add even more scenarios to also include filtering in this mix!

packages/account-tree-controller/src/state/payload.ts
ccharly: I investigated this a bit with claude.

The real problem is that we use raw `string` for the secrets, and it makes it hard to zero-ing memory completely with those (since we don't own the real memory buffer for those, only the view/value on it).

If we really wanted to do this, we would need to use real buffer types like `ArrayBuffer` or `Uint8Array`. But that means our payload will need special encoding functions (at least, to make it compatible with JSON-encoding for example).

Like we could have a `.stringify()` and `.parse(json)` for example.

But that would yet another steps before/after the snapshot's `serialize` and `deserialize`.

gantunesr (dismissed):

packages/account-tree-controller/src/state/payload.ts
ccharly: After discussing a bit about this internally, I actually went for a different way of encoding those in our payload, so we "limit" the window of when those sensitive values are used.

They are just encoded differently, so it's more or less security-by-obscurity, but it's slightly better than having them in clear I think (and that's aligned with how we encode them elsewhere in the codebase)

https://github.com/MetaMask/core/pull/9826/commits/ccc122ee9052b14f53bed083822f3b2ab88026f5

packages/account-tree-controller/src/state/payload.ts
```diff
@@ -0,0 +1,352 @@
...
+  },
+);
+
+const AccountWalletMnemonicPayloadStruct = object({
+  id: AccountWalletPayloadIdStruct,
+  type: literal(AccountWalletPayloadType.Mnemonic),
+  value: exactOptional(sensitive(BytesStruct)),
+  metadata: AccountWalletPayloadMetadataStruct,
+  groups: AccountWalletMnemonicGroupsStruct,
+});
```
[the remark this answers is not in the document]
ccharly: That's true even though we are the producer of this payload, so we "know" we won't be doing that.

We can fix that with `refine` and a stricter validation, but that's not vital honestly.

gantunesr (approved):

PR #9832, opened 2026-08-12
feat(perps): [perps-controller] Add Chase, TWAP, and Scale order types

abretonc7s:
## Explanation

`@metamask/perps-controller` supported six order types — `market`, `limit`, and the four trigger
placements — all of which resolve to a single order submitted through one code path. Chase, TWAP and
Scale are execution *strategies*: one request expands into a schedule of orders, and each needs a
different submission path, a different cancellation path, and rules of its own. None of that was
expressible.

This adds `twap`, `scale` and `chase` to `OrderType`, with the parameters, validation, placement and
cancellation each needs.

**Protocol research (the ticket's open question).** Answered from the SDK the package already depends
on, `@nktkas/hyperliquid@0.33.1`: TWAP **is** native (`twapOrder` / `twapCancel`, with their own
endpoints); Scale is **not** — it is a batch of ordinary limit orders; Chase is **not** — no such
action exists anywhere in the SDK, so it is emulated client-side. The SDK schema also pins the TWAP
window to a whole number of minutes in `[5, 1440]`, stricter than the ticket's "twapDuration > 0", so
validation enforces the real bound.

**What each placement does.**

- **TWAP** is submitted through the venue's TWAP action rather than the order book, and returns the
  venue's TWAP id. Cancelling it uses the TWAP cancel endpoint, never the order-book cancel.
- **Scale** fans out `scaleNumOrders` limit orders on an inclusive ladder between `scaleMinPrice` and
  `scaleMaxPrice`, submitted as one batch so the ladder rests together or fails together. Sizes are
  split in whole units of the asset's size grid, so the rungs sum to exactly the submitted size.
  Children come back as `OrderResult.childOrderIds`; the returned handle cancels all of them at once.
- **Chase** rests a post-only order at the near touch and returns a session handle immediately; a
  background tick re-prices it as the touch moves, stopping at the repricing cap, at the window
  deadline, when the order leaves the book, on cancel, or on `disconnect()`.

**Notable API effects.** `OrderType` is a wider union, which is breaking in the same way the trigger
types were in 11.0.0: consumer signatures that narrow it back to a smaller set must widen. `OrderParams`
gains eight optional strategy fields, `OrderResult` gains `childOrderIds`, and `CancelOrderParams`
gains an optional `orderType` that selects the cancellation path — omitting it, which every existing
caller does, keeps today's behaviour exactly. Invalid strategy parameters (inverted scale range,
out-of-range TWAP duration, a strategy field on a non-strategy order, a limit price on a strategy) are
rejected with a typed `PERPS_ERROR_CODES` value before any network call.

The public model stays provider-agnostic — no venue vocabulary reaches `OrderParams`,
`OrderResult` or `CancelOrderParams`. The one genuinely venue-specific constant is named for its
venue, `HYPERLIQUID_TWAP_LIMITS`, beside the existing `HYPERLIQUID_ORDER_LIMITS`.

One incidental fix: `TriggerOrderType` was `Exclude<OrderType, 'market' | 'limit'>`, so the three new
members would have been silently absorbed into the trigger union and started demanding a trigger
price. It is now spelled out. The resolved type is unchanged for existing consumers.

**Validation.** 108 new unit tests across three suites, driving the real provider against a stubbed
exchange client and asserting the exact actions submitted — that a TWAP reaches the TWAP action and
not the order action, that its cancel reaches the TWAP cancel endpoint and not the order cancel, the
exact scale ladder prices and the size split, the post-only chase placement at the touch and its
re-pricing loop, and that no exchange call is made for an invalid placement. The full package suite
passes with coverage thresholds met, the root build emits the new symbols, and the six pre-existing
order suites pass unmodified. A live read against HyperLiquid testnet confirms the controller still
instantiates and reads positions, orders and account state with the widened union in place.

**Follow-ups, deliberately out of scope.** Strategy-handle correlation is session-scoped (children
remain cancellable via `childOrderIds` after a restart); TWAP progress is not surfaced in controller
state, because the venue reports TWAPs through feeds the open-orders normalisation does not read; and
strategy placements are refused on sub-exchange (HIP-3) markets, whose pre-order margin transfer and
rollback are wired into the single-order submit path.

## References

- https://consensyssoftware.atlassian.net/browse/TAT-3723
- Client follow-up: mobile/extension order forms need to send the new fields and to pass `orderType`
  when cancelling a strategy handle. UI for these order types is explicitly out of scope here.

## Checklist

  - The `OrderType` widening is breaking in the same way as the trigger types in 11.0.0 and is marked
    **BREAKING** in the changelog, but no client draft PRs are prepared yet: the client work is a
    separate ticket, and no consumer in this repo narrows `OrderType`.

## **Screenshots/Recordings**

---

> [!NOTE]
> **High Risk**
> Widens the public `OrderType` union (breaking) and adds new signed trading paths, including a client-side chase loop with disconnect/cancel races and concurrency limits.
> 
> **Overview**
> Adds **strategy order types** `twap`, `scale`, and `chase` to `OrderType`, so one `placeOrder` request can expand into an execution schedule instead of a single resting order.
> 
> **TWAP** goes through HyperLiquid’s native TWAP action/cancel. **Scale** submits a batch limit ladder and tracks children under a group handle. **Chase** is emulated client-side: a post-only order rests one tick inside the spread and is cancel/replaced as the touch moves, bounded by interval, duration, and repricing caps.
> 
> `OrderResult.orderId` is a strategy *handle* for these types, with exchange ids in `childOrderIds`. `CancelOrderParams.orderType` selects the strategy cancel path; omitting it keeps ordinary single-order cancel. `editOrder` rejects strategy edits. Invalid strategy params fail with new typed `PERPS_ERROR_CODES` before signing.
> 
> Also **breaks** consumers that narrow `OrderType`, narrows single-order helpers/`closePosition` to `OrdinaryOrderType`, and spells out `TriggerOrderType` so the new types are not pulled into the trigger union.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a22d0e7a0d3c3db256ffdd393b0584be929d7a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -252,6 +282,26 @@ export const TP_SL_CONFIG = {
...
+ * instead of the per-order minimum — its suborders are its own business.
+ *
+ * Carries the venue prefix, like `HYPERLIQUID_ORDER_LIMITS`, because these are
+ * protocol constants rather than controller policy.
+ *
+ * From: https://hyperliquid.gitbook.io/hyperliquid-docs/trading/order-types
+ */
+export const HYPERLIQUID_TWAP_LIMITS = {
+  MinDurationMinutes: 5,
+  MaxDurationMinutes: 1440,
```
geositta, line 303: Could we resolve the duration discrepancy before publishing this as the venue limit? SDK 0.33.1 validates m <= 1440, but HyperLiquid’s current order-type documentation states that TWAPs can run for 5 minutes to 7 days (10,080 minutes). As written, the public API rejects venue-documented durations above one day. This requires either updating/patching the SDK and using 10,080, or documenting the one-day cap as an SDK limitation rather than HyperLiquid's protocol bound.

geositta (dismissed):
The implementation cleanly separates TWAP, scale, and chase execution paths while preserving existing order behavior. Validation, cancellation semantics, and focused test coverage support the expanded API. I left one nonblocking comment about clarifying the pinned SDK's TWAP duration limit.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -3655,224 +3943,1592 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+      const outcome = await this.#cancelChaseChild(session);
+
+      // The cancel is a second round trip, and this is the window where a
+      // caller's `cancelOrder` does the most damage: it would cancel the order
+      // this tick has already cancelled, report the cancel incomplete, and then
+      // this tick would rest a replacement the caller has no idea exists.
+      // Stopping here leaves nothing on the book.
+      if (!session.active) {
+        session.orderId = null;
+        return;
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4ad562e117. A refused tick cancel now preserves the live child ID, with a deterministic cancel-race regression.

packages/perps-controller/src/constants/perpsConfig.ts
abretonc7s: Clarified in 4ad562e117 that 1,440 minutes is the pinned SDK 0.33.1 cap; the venue documents 10,080 minutes.

michalconsensys (approved):

PR #9837, opened 2026-08-12
feat: retry failed quote loads and surface quote errors in TransactionPayController

dan437:
## Explanation

When a quote load in `TransactionPayController` failed unexpectedly, the error was swallowed (`.catch(noop)`), nothing was written to state, and the refresh loop skipped transactions with no quotes — so they were never retried. One failed fetch left the transaction permanently without quotes: clients showed no fees, and submission was blocked by the publish guard ("MetaMask Pay: Cannot submit without quote").

This change makes failed quote loads recoverable:

- `refreshQuotes` now also retries transactions that should have quotes but have none, on the default refresh interval. One transaction's failure no longer stops the others from refreshing.
- `QuoteRefresher` keeps the refresh loop alive while such a transaction exists.
- Unexpected failures are persisted to `quoteError` (reason `no-quotes`) instead of being swallowed — but only when quotes are needed and none are usable. Existing executable quotes stay usable until a refresh replaces them, and direct routes stay silent because they need no quotes. Stale no-op quotes from a previous direct route are dropped so the retry loop picks the transaction up.
- Every failed or skipped attempt stamps `quotesLastUpdated`, so retries wait a full interval instead of running every tick, and late writes cannot recreate pay state that cleanup already removed.

## References

- Related to https://github.com/MetaMask/metamask-mobile/pull/34471 (client-side submit gating for the same defect)
- CONF-1731

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core quote fetch, refresh, and error state for pay submission; behavior is well-tested but affects when users see errors and when submit is allowed.
> 
> **Overview**
> Fixes **MetaMask Pay** transactions that could stay permanently without quotes after a single failed or empty quote fetch, which blocked fees display and submit.
> 
> **Quote retry and refresh loop:** `refreshQuotes` now retries transactions that should have quotes but have none (failed load, empty result, fiat payment selected, etc.), respecting the normal refresh interval via `quotesLastUpdated`. `QuoteRefresher` keeps polling while `isQuoteRetryPending` is true, not only when executable quotes already exist. Per-transaction failures are isolated so one bad transaction does not stop others from refreshing.
> 
> **Failure persistence:** Unexpected quote pipeline errors call `persistQuoteLoadFailure`, which stamps `quotesLastUpdated`, sets `quoteError` (`no-quotes`) only when quotes are needed and none are usable, drops stale no-op quotes so retries can run, and skips writes if pay state was removed. Refresh paths that skip or throw still stamp attempt time so retries do not hammer every tick.
> 
> **Controller:** `updateQuotes` rejections are logged instead of swallowed with `noop`, without changing the fire-and-forget call pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b94dcdd9d7bdda4a6f47f0a535ccebd928d0090c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -36,6 +37,8 @@ import {
   subscribeTransactionChanges,
 } from './utils/transaction.js';
 
+const log = createModuleLogger(projectLogger, 'controller');
```
matthewwalsh0, line 40: The `projectLogger` is the `log` function itself for the main controller file.

packages/transaction-pay-controller/src/helpers/QuoteRefresher.ts
```diff
@@ -107,15 +107,20 @@ export class QuoteRefresher {
...
-      ),
+    // Transactions that need quotes but have none had a failed or empty
+    // quote load and need the loop to retry them.
+    const needsRefreshLoop = Object.values(state.transactionData).some(
+      (transaction) =>
+        Boolean(
+          transaction.quotes?.some(
+            (quote) => quote.strategy !== TransactionPayStrategy.None,
+          ),
+        ) || isQuoteRetryPending(transaction),
```
matthewwalsh0, line 118: As discussed, we need to confirm this is some Barbara wants.

My original assumption was the error cases (provider or simulation errors) wouldn't fix themselves so redundant to have the extra traffic and UX skeletons etc.

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -310,47 +419,108 @@ export async function refreshQuotes(
...
-    const strategyName = quotes[0].strategy;
-    const strategy = getStrategyByName(strategyName);
+    const strategyName = firstQuote?.strategy;
 
-    const refreshInterval =
-      (await strategy.getRefreshInterval?.({
-        chainId: quotes[0].request.sourceChainId,
-        messenger,
-      })) ?? DEFAULT_REFRESH_INTERVAL;
+    const refreshInterval = firstQuote
```
matthewwalsh0, line 443: Minor, could we assign default then override if quotes and strategy interval, rather than ternary and fallbacks?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -310,47 +419,108 @@ export async function refreshQuotes(
...
     const { isLoading, quotes, quotesLastUpdated } = transactionData;
 
-    if (isLoading || !quotes?.length) {
+    if (isLoading) {
+      continue;
+    }
+
+    const firstQuote = quotes?.[0];
+
+    if (!firstQuote && !isQuoteRetryPending(transactionData)) {
```
matthewwalsh0, line 428: Could we replace `isQuoteRetryPending` with just checking if `quoteError` is set as we always set that?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -310,47 +419,108 @@ export async function refreshQuotes(
...
+  messenger: TransactionPayControllerMessenger,
+  transactionId: string,
+  updateTransactionData: UpdateTransactionDataCallback,
+): void {
+  updateExistingTransactionData(
+    messenger,
+    transactionId,
+    updateTransactionData,
+    (data) => {
+      data.quotesLastUpdated = Date.now();
```
matthewwalsh0, line 491: Why do we need to update this if we haven't derived new quotes?

What does this enable in the UI?

matthewwalsh0 (changes requested):

dan437:
Closing this for now. Mobile #34471 fixes the user-facing issue, while this retry logic also retries legitimate empty responses.

PR #9550, opened 2026-07-20
feat(perps-controller): add pro-mode layout preferences and mode state

michalconsensys:
## Explanation

Before Perps Pro mode v1 UI work can start, `@metamask/perps-controller` needs two core-state additions. This PR is core-package preparation only — there are no mobile UI changes.

- **Pro-mode layout preferences.** Adds a new top-level `proLayoutPreferences` field to `PerpsControllerState` (`orderBookExpanded`, `chartExpanded`, `orderBookPosition`, `orderFormPosition`). It follows the existing `marketFilterPreferences` pattern — a flat, network-independent, persisted object — rather than the per-market `tradeConfigurations.orderBookGrouping` pattern, because these preferences must survive across markets instead of resetting per symbol. All four fields get sensible defaults now (`chartExpanded`/positions are reserved for future container-position UI), so no state-shape migration is needed later. Access is via `getProLayoutPreferences()` and a patch-style `setProLayoutPreferences(patch)` setter (so future fields don't require new setter signatures), plus a `selectProLayoutPreferences` selector.
- **Lite/Pro mode field.** Adds a `PerpsMode` enum (`Lite`/`Pro`) and a `mode: PerpsMode` field defaulting to `PerpsMode.Lite`, with a `setPerpsMode(mode)` method (mirroring the existing `toggleTestnet` write pattern) and a `selectPerpsMode` selector. This supersedes the "State:" bullet previously sketched in TAT-3551, which will consume this field/enum/setter instead of redefining them.

Both fields are persisted (`persist: true`) consistent with the controller's existing global-preference metadata (`marketFilterPreferences`/`maxSlippageBps`). The new methods are exposed as messenger actions via the existing method-action codegen (`PerpsController-method-action-types.ts` regenerated), and the enum, type, defaults, selectors, and action types are exported from the package entry point.

To keep the API ergonomic and robust for consumers:

- The canonical `PerpsMode` enum, `ProLayoutPreferences` type, and the `DEFAULT_PERPS_MODE` / `DEFAULT_PRO_LAYOUT_PREFERENCES` constants live in the dependency-free `constants/perpsConfig` module (alongside `SortOptionId`/`MARKET_SORTING_CONFIG`) and are re-exported from the controller/package entry points. This preserves the one-way `PerpsController → perpsConfig` dependency (no cycle) and keeps `selectors.ts` free of heavy transitive imports.
[...]

PR #9838, opened 2026-08-12
feat(perps): persist Pro positions sort and side filter prefs

michalconsensys:
## Explanation

In Pro mode, mobile needs Positions/Orders panel sort and side-filter preferences to persist across markets and app restarts. Today those values live in local React state and reset on remount (e.g. market navigation via stack push).

This extends the existing network-independent flat `proLayoutPreferences` bag (already used for `chartExpanded`) with:

- `positionsSideFilter` — default `'all'`
- `positionsSortField` — default `'positionValue'`
- `positionsSortDirection` — default `'desc'`

Sort is stored as two top-level scalars (same shape as `marketFilterPreferences`), so getter/setter/selector stay on shallow spread with no custom merge helper. Callers continue to use `getProLayoutPreferences()` / `setProLayoutPreferences(patch)`.

**TypeScript breaking:** consumers that construct a full `ProLayoutPreferences` object must include the new required fields. Runtime persisted state is forward-compatible via merge-over-defaults.

Mobile consumer PR: https://github.com/MetaMask/metamask-mobile/pull/34667 ([TAT-3644](https://consensyssoftware.atlassian.net/browse/TAT-3644)).

## References

* Fixes https://consensyssoftware.atlassian.net/browse/TAT-3644
* Related: https://github.com/MetaMask/metamask-mobile/pull/34667

## Checklist

[TAT-3644]: https://consensyssoftware.atlassian.net/browse/TAT-3644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> UI preference persistence only; runtime remains forward-compatible via merge-over-defaults, with a TypeScript-only break for full-object constructors.
> 
> **Overview**
> Extends the flat `ProLayoutPreferences` bag so Pro Positions/Orders **side filter** and **sort** survive market navigation and app restarts via the existing `getProLayoutPreferences()` / `setProLayoutPreferences(patch)` API.
> 
> Adds three required fields with defaults: `positionsSideFilter` (`'all'`), `positionsSortField` (`'positionValue'`), and `positionsSortDirection` (`'desc'`), plus exported types `ProPositionsSideFilter`, `ProPositionsSortField`, and `ProPositionsSortDirection`.
> 
> **Breaking for TypeScript:** callers that construct a full `ProLayoutPreferences` object must include the new fields. Older persisted state stays valid at runtime because the getter and `selectProLayoutPreferences` still merge over `DEFAULT_PRO_LAYOUT_PREFERENCES`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ec9a8f45d5ae897ad7c569071d9a262cb48ee76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -486,19 +486,45 @@ export enum PerpsMode {
...
+ * the Positions/Orders panel sort and side filter so they survive market
+ * navigation and app restarts.
  */
 export type ProLayoutPreferences = {
   orderBookExpanded: boolean;
   chartExpanded: boolean;
   orderBookPosition: 'left' | 'right';
   orderFormPosition: 'left' | 'right';
+  positionsSideFilter: ProPositionsSideFilter;
+  positionsSortConfig: ProPositionsSortConfig;
```
aganglada: Nesting `positionsSortConfig` is the only reason we need a custom merge helper — the original `proLayoutPreferences` bag from #9550 was intentionally flat.

Closest in-package precedent is `marketFilterPreferences`, which stores sort as two top-level scalars (`optionId` + `direction`) with a simple `pref ?? defaults` getter and an atomic setter. Consider flattening to `positionsSortField` + `positionsSortDirection` here so getter/setter/selector can stay on shallow spread and you can drop `mergeProLayoutPreferences` entirely.

If Mobile's UI state is nested, mapping at the boundary is a small cost for a much simpler persisted shape.

packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -513,8 +539,45 @@ export const DEFAULT_PRO_LAYOUT_PREFERENCES: ProLayoutPreferences = {
...
+/**
+ * Merge a partial/persisted pro-layout preference blob over defaults.
+ *
+ * Nested `positionsSortConfig` is deep-merged so a persisted object that
+ * predates one of its fields still yields a fully-populated config.
+ *
+ * @param prefs - Partial preferences from persisted state or a setter patch.
+ * @returns A fully-populated `ProLayoutPreferences` object.
+ */
+export const mergeProLayoutPreferences = (
```
aganglada: If you keep the nested object, this helper only deep-merges `positionsSortConfig` over **defaults**, not over successive layers (persisted state → patch). That forces the setter to re-implement nested merge before calling in.

Consider a variadic signature that folds layers in one place:

```ts
export const mergeProLayoutPreferences = (
  ...layers: Array<ProLayoutPreferencesPatch | null | undefined>
): ProLayoutPreferences => { /* merge top-level + accumulate positionsSortConfig across layers */ };
```

Then:
- getter: `mergeProLayoutPreferences(this.state.proLayoutPreferences)`
- setter: `mergeProLayoutPreferences(state.proLayoutPreferences, patch)`

No need for `deepmerge` (devDependency only here) — a ~15-line loop is enough for one nested key.

packages/perps-controller/src/PerpsController.ts
```diff
@@ -5141,26 +5150,33 @@ export class PerpsController extends BaseController<
...
    */
-  setProLayoutPreferences(patch: Partial<ProLayoutPreferences>): void {
+  setProLayoutPreferences(patch: ProLayoutPreferencesPatch): void {
     this.update((state) => {
-      state.proLayoutPreferences = {
+      state.proLayoutPreferences = mergeProLayoutPreferences({
         ...state.proLayoutPreferences,
         ...patch,
-      };
+        ...(patch.positionsSortConfig
```
aganglada: This manual `positionsSortConfig` spread exists because `{ ...state, ...patch }` clobbers the nested object when the patch only supplies `field` or `direction`. The logic belongs inside `mergeProLayoutPreferences` (see comment on the helper) so the setter can be:

```ts
state.proLayoutPreferences = mergeProLayoutPreferences(
  state.proLayoutPreferences,
  patch,
);
```

Alternative if you want to avoid deep merge altogether: keep nested shape but add an atomic `setPositionsSortConfig(config)` (like `saveMarketFilterPreferences(optionId, direction)`) and restrict `setProLayoutPreferences` to shallow top-level patches.

packages/perps-controller/src/index.ts
```diff
@@ -35,6 +35,7 @@ export {
   PerpsMode,
   DEFAULT_PERPS_MODE,
   DEFAULT_PRO_LAYOUT_PREFERENCES,
+  mergeProLayoutPreferences,
```
aganglada: Do consumers need `mergeProLayoutPreferences` on the public surface? Getter/setter/selector already wrap it. Exporting adds API surface for an internal persistence detail — I'd keep it module-private unless Mobile has a concrete out-of-controller hydration path that needs it.

aganglada (commented):
Focused on the merge helper and whether existing perps patterns apply. No approval/block — inline notes on specific lines.

aganglada (approved):

PR #9839, opened 2026-08-12
feat(logging-controller): Add expiration to logs

FrederikBolding:
## Explanation

Add `expiryTime` constructor argument to `LoggingController`, it defaults to 7 days.

When adding a new log entry, we iterate through the existing logs and discard any that were logged before the expiry cutoff.

Additionally simplifies the controller by using UUID v4, not checking for conflicts and removing an unused dependency on `controller-utils`.

## References

N/A

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes persisted log retention by auto-deleting old entries on `add`, which can remove historical support/debug data. Scope is limited to `LoggingController` and the new constructor option is backward-compatible.
> 
> **Overview**
> Adds an optional `expiryTime` constructor argument to `LoggingController` (default **7 days**). On each `add`, expired entries are removed before the new log is stored, so persisted logs no longer grow indefinitely.
> 
> Also simplifies ID generation by switching from UUID v1 with collision retries to UUID v4, and drops the unused `@metamask/controller-utils` dependency in favor of `@metamask/utils` for duration helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e0a54e931657607c8b468e3add54f247b3bddf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/logging-controller/src/LoggingController.ts
+++ b/packages/logging-controller/src/LoggingController.ts
@@ -80,8 +80,7 @@ export class LoggingController extends BaseController<
   LoggingControllerState,
   LoggingControllerMessenger
 > {
-
-  #expiryTime: number;
+  readonly #expiryTime: number;
 
   /**
    * Creates a LoggingController instance.
--- a/packages/logging-controller/src/LoggingController.ts
+++ b/packages/logging-controller/src/LoggingController.ts
@@ -5,7 +5,8 @@ import type {
 } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
 import type { Messenger } from '@metamask/messenger';
-import { v1 as random } from 'uuid';
+import { Duration, inMilliseconds } from '@metamask/utils';
+import { v4 as uuid } from 'uuid';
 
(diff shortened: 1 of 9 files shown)


packages/logging-controller/src/LoggingController.ts
GuillaumeRx, line 1: What about also doing it when we initialize the controller ?

FrederikBolding: Yeah, we could

OGPoyraz (approved):

packages/logging-controller/src/LoggingController.ts
FrederikBolding: Though it probably doesn't matter much as I think we wouldn't persist until an actual change happens 🤔 WDYT?

GuillaumeRx: Yes that makes sense !

GuillaumeRx (approved):

PR #9840, opened 2026-08-12
fix: defer re-entrant publishes of the same event

GuillaumeRx:
## Explanation

`Messenger` delivered a re-entrant publish of the same event **inline**. When a subscriber published the event it was currently handling — directly, or indirectly through a delegated messenger (for example a `stateChange` subscriber that calls an action which triggers `update()` on the same controller) — the nested publish ran to completion, and then the in-progress publish **resumed its subscriber loop and re-delivered its now-stale payload** to the subscribers it had not yet reached. A downstream subscriber such as a UI store bridging controller state could therefore end up observing an older payload after a newer one.

This surfaced in the extension as a network filter that stayed on a just-added network even though the enabled-network map had already been restored: the restore's `stateChange` re-entered `publish`, and the stale "switched" payload was the last thing the UI received.

This change makes `#publish` **defer** a publish of an event that is already being published on the same messenger, then **drain** the deferred publishes in order once the in-progress publish finishes. Because the event stays marked as "publishing" while draining, further re-entrant publishes of that event are deferred and drained too. Re-entrant publishes of a *different* event still run inline. The fix covers delegated messengers as well, since delegated publishes funnel through the same `#publish`.

Note this is a small behavioral change: a re-entrant publish of the same event is now delivered after the current publish completes (still synchronously, within the same tick) rather than nested inside it.

## References

- Related to MetaMask/metamask-extension#44142, where this was worked around at the call site (restoring the map outside the `stateChange` publish via `await`).

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core event-delivery ordering for re-entrant same-event publishes across all Messenger consumers; behavior is intentional but could affect code that relied on nested delivery order.
> 
> **Overview**
> Fixes **`Messenger`** so a subscriber that publishes the **same** event while handling it (including via delegation) no longer causes later subscribers to see an **outdated** payload.
> 
> `#publish` now tracks in-flight publishes per event type and **queues** same-event re-entrancy until the current delivery finishes, then drains queued publishes in order. Subscriber iteration moves to **`#deliverToSubscribers`**; re-entrant publishes of **different** events still run inline. Tests and the changelog document the behavior change (deferred delivery is still synchronous in the same tick).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ab366a8be2dd82af255661d247e3618bcbab913. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/messenger/src/Messenger.ts
+++ b/packages/messenger/src/Messenger.ts
@@ -263,6 +263,13 @@ export class Messenger<
   readonly #events = new Map<Event['type'], EventSubscriptionMap<Event>>();
 
   /**
+   * In-progress publishes, keyed by event type. A key is present for the
+   * duration of a publish (so presence means "publishing"); its array collects
+   * re-entrant publishes of that event, drained when the publish finishes.
+   */
+  readonly #deferredPublishes = new Map<Event['type'], (() => void)[]>();
+
+  /**
    * The set of messengers we've delegated events to and their event handlers, by event type.
    */
   readonly #subscriptionDelegationTargets = new Map<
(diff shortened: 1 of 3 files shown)


packages/messenger/src/Messenger.ts
```diff
@@ -648,6 +655,37 @@ export class Messenger<
...
+
+    const deferred: (() => void)[] = [];
+    this.#deferredPublishes.set(eventType, deferred);
+    try {
+      this.#deliverToSubscribers(eventType, ...payload);
+
+      // Drain deferred publishes in order. The array grows as further
+      // re-entrant publishes push onto it; the iterator reads those too.
+      for (const run of deferred) {
+        run();
```
Gudahtt, line 679: Nit: Should we wrap this in another `try...catch`, and report the result with `this?.captureException`?

As currently written, one deferred event handler could prevent a second deferred event from executing. This differs from today, where each event is independent (a _subscriber_ can interrupt later subscribers, but separate publishes cannot impact each other).

In general we expect subscribers to never crash, so I'm not sure this case is especially important to handle. But we don't enforce this expectation.

Gudahtt: Oh wait, sorry we're already doing this in `#deliverToSubscribers`, that's good. Nevermind

Gudahtt (approved):
LGTM!

PR #9867, opened 2026-08-13
fix: never serve Accounts API balances from cache

salimtb:
AccountsApiDataSource now invalidates cached balance queries before every fetch and fetches v5/v6 balances with staleTime: 0, gcTime: 0, so balances always hit the network in every flow (polling, forced refresh, websocket-triggered) and are never retained for reuse. Previously non-forced polls could reuse balances cached for up to 60s and forceUpdate: true only shrank the cache window to 100ms, so pull-to-refresh and post-transaction refreshes could show stale balances.

Extension PR: https://github.com/MetaMask/metamask-extension/pull/45494
Mobile PR: https://github.com/MetaMask/metamask-mobile/pull/34748

## Checklist

### DIFF
--- a/packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
+++ b/packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
@@ -427,9 +427,13 @@ export class AccountsApiDataSource extends AbstractDataSource<
         return response;
       }
 
-      const fetchOptions = request.forceUpdate
-        ? { staleTime: 100, gcTime: 100 }
-        : undefined;
+      // Balances must never be served from the TanStack cache in any flow
+      // (polling, forced refresh, websocket-triggered): invalidate cached
+      // balance queries first (fetchQuery refetches invalidated queries
+      // regardless of freshness), then fetch with staleTime/gcTime 0 so the
+      // response is neither read back nor retained.
+      const fetchOptions = { staleTime: 0, gcTime: 0 };
+      await this.#apiClient.accounts.invalidateBalances();
 
       // Feature-flagged: v6 endpoint with a fallback to legacy v5. The flag is
       // read here (not cached) so a runtime toggle can revert v6 -> v5.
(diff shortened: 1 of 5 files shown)


packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -427,9 +427,13 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
 
-      const fetchOptions = request.forceUpdate
-        ? { staleTime: 100, gcTime: 100 }
-        : undefined;
+      // Balances must never be served from the TanStack cache in any flow
+      // (polling, forced refresh, websocket-triggered): invalidate cached
+      // balance queries first (fetchQuery refetches invalidated queries
+      // regardless of freshness), then fetch with staleTime/gcTime 0 so the
+      // response is neither read back nor retained.
+      const fetchOptions = { staleTime: 0, gcTime: 0 };
```
salimtb, line 435: passing this to 0 now , we can also consider removing it

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -427,9 +427,13 @@ export class AccountsApiDataSource extends AbstractDataSource<
...
+      await this.#apiClient.accounts.invalidateBalances();
```
salimtb, line 436: staleTime/gcTime only apply to this one call,  old entries in the shared cache keep their 60s freshness. Invalidating first marks them stale for all consumers and forces an unconditional refetch, guaranteeing balances never come from cache.

salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
Prithpal-Sooriya: Reverts this PR:
https://github.com/MetaMask/core/pull/9591

Worth double checking - I think instead of caching, we should attempt debouncing. But open to discuss

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
Prithpal-Sooriya: Yes, 100% agreed here

salimtb:
closing this one to do a revert instead: https://github.com/MetaMask/core/pull/9870

PR #9875, opened 2026-08-13
feat(phishing-controller): export extractSignatureAddresses for EIP-712 typed-data scanning

wzrdk3lly:
## Explanation
Ticket: [PSAFE-441](https://consensyssoftware.atlassian.net/browse/PSAFE-441)

MetaMask validates \`eth_signTypedData_v3\` and \`eth_signTypedData_v4\` requests with PPOM, which inspects the address fields of a signature. PPOM's threat data is refreshed on a delay, so an address that has been flagged recently can still be reported as benign, and it does not cover every chain or protocol.

Separately, the real-time per-address scan (\`/address/evm/scan\`) is only applied to a few fields today: the token \`verifyingContract\`, permit \`spender\`, and delegation \`delegate\`. Addresses carried in any other field — recipients, makers, tokens, custom protocol fields — receive no real-time scan at all.

This change adds a schema-driven extractor that returns up to 10 distinct \`address\`-typed values found in a typed-data message (including nested structs and arrays, matched by declared type rather than field name), and reports when the message could not be fully walked because the address cap, depth limit, or work budget was reached. When PPOM has not already flagged the request, the extracted addresses are passed to the real-time address scan. Results reuse the existing address cache, so fields already scanned elsewhere are not requested again.

### What this exports

- \`extractSignatureAddresses(typedData, options)\` — walks \`types\`/\`primaryType\`/\`message\`, returns \`{ addresses, fields, overflow }\`
- \`ExtractedSignatureAddresses\` — return type
- \`ExtractSignatureAddressesOptions\` — options: \`exclude\` (addresses to skip, e.g. the signer), \`excludeFields\` (top-level field names to skip, e.g. \`spender\` when already covered by an existing scan)

### Safeguards

- Cap of 10 distinct addresses per message
- Max traversal depth of 12
- Max 5000 nodes visited
- \`overflow: true\` is returned whenever the walk is incomplete so callers can surface a caution

## References

- MetaMask Extension (client wiring): MetaMask/metamask-extension#45521
- MetaMask Mobile (client wiring): MetaMask/metamask-mobile#34786
- Prior implementation with local extractor copy (Extension): MetaMask/metamask-extension#45058
- Prior implementation with local extractor copy (Mobile): MetaMask/metamask-mobile#34428

## Changelog

See \`packages/phishing-controller/CHANGELOG.md\`.

## Checklist

- [x] Tests written and passing
- [x] Changelog updated
- [x] Exports added to \`index.ts\`
- [ ] Version bump (pending release cut)

[PSAFE-441]: https://consensyssoftware.atlassian.net/browse/PSAFE-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> New security-adjacent parsing logic must stay aligned with EIP-712 signing and cap/overflow behavior; incomplete walks could miss addresses if clients ignore `overflow`.
> 
> **Overview**
> Adds a shared **`extractSignatureAddresses`** helper to `@metamask/phishing-controller` so clients can feed **real-time address scans** from `eth_signTypedData` payloads without hard-coding permit/spender-style field names.
> 
> The walker follows EIP-712 **`types` / `primaryType` / `message`**, collecting values for fields declared as **`address`** (including nested structs and arrays), normalizing encodings to match signer behavior, and returning **`addresses`**, per-address **`fields`**, and **`overflow`** when caps or traversal limits stop a full walk. Callers can **`exclude`** addresses (e.g. signer), skip top-level **`excludeFields`**, and tune **`maxAddresses`** (default 10, ceiling 50). **`DEFAULT_MAX_SIGNATURE_ADDRESSES`** and **`MAX_SIGNATURE_ADDRESSES_CEILING`** are exported alongside the types.
> 
> Also documents the feature in the changelog, adds **`@metamask/eth-sig-util`** as a **devDependency** for normalization tests, and bumps **`@metamask/transaction-controller`** to `^69.8.1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 541dff0d9dd7d1bc164a9b2cc5b7a9100d1fc125. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/phishing-controller/src/signature-address-extraction.ts
```diff
@@ -0,0 +1,300 @@
...
+    value >= 0
+  ) {
+    numeric = BigInt(value);
+  } else {
+    return undefined;
+  }
+
+  // Reduce into the 20-byte address space, matching how the signer encodes an
+  // `address` field, so non-canonical encodings resolve to the signed address.
+  numeric %= ADDRESS_MODULUS;
```
adonesky1: The %= ADDRESS_MODULUS is backwards relative to the signer. [encodeField](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L297-L305) returns reallyStrangeAddressToBytes(value).subarray(0, 20), and since padStart never truncates, an oversized value keeps its leading 20 bytes. Modulo keeps precisely the low bytes the signer discards.

This is exploitable: verified against `TypedDataUtils.encodeData, decimal(ADDR_A * 256 + 0x42)` signs as exactly `ADDR_A` but normalizes here to `0x1111…1142`. An attacker appends one byte, the real spender never reaches the scanner, and overflow stays false.

Suggested fix: take the leading 20 bytes rather than reducing, plus a test asserting agreement with `TypedDataUtils.encodeData` so the two can't drift again.

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `extractSignatureAddresses` utility, plus `ExtractedSignatureAddresses` and `ExtractSignatureAddressesOptions` types, to collect the `address`-typed values from an EIP-712 typed-data message for real-time address scanning ([#9999](https://github.com/MetaMask/core/pull/9999))
```
adonesky1: Two fixes here: the link should be #9875 rather than #9999, and the "reduced mod 2^160 as the signer does" wording on the next line describes behaviour the signer does not have, so it should change with the normalisation fix.

packages/phishing-controller/src/signature-address-extraction.test.ts
```diff
@@ -0,0 +1,542 @@
...
+      'X',
+      { X: [{ name: 'a', type: 'address' }] },
+      { a: mixed },
+    );
+    expect(addressesOf(data)).toStrictEqual([mixed.toLowerCase()]);
+  });
+
+  it('reduces an oversized decimal-encoded address to the signed address', () => {
+    // The signer reduces an `address` mod 2^160, so `value + 2^160` signs as
+    // `value`. The extractor must resolve it to the same address.
```
adonesky1: The premise in this comment is inverted: the signer does not reduce an `address` mod 2^160.

For a string that isn't strict hex, [`encodeField`'s `address` branch](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L297-L305) falls through to the last path:

```ts
return ['address', reallyStrangeAddressToBytes(value).subarray(0, 20)];
```

packages/phishing-controller/src/signature-address-extraction.ts
```diff
@@ -0,0 +1,300 @@
...
+  for (const address of options.exclude ?? []) {
+    const normalized = normalizeAddress(address);
+    if (normalized) {
+      excluded.add(normalized);
+    }
+  }
+
+  const excludedFields = new Set(
+    (options.excludeFields ?? []).map((field) => field.toLowerCase()),
+  );
```
adonesky1: EIP-712 field names are case-sensitive. [encodeData](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L389-L415) reads data[field.name] verbatim and this file does too at [L243](https://github.com/MetaMask/core/blob/82d0fb4812922baceb6b76e8b18a4ffc086094db/packages/phishing-controller/src/signature-address-extraction.ts#L243), so lowercasing only in the [exclusion check](https://github.com/MetaMask/core/blob/82d0fb4812922baceb6b76e8b18a4ffc086094db/packages/phishing-controller/src/signature-address-extraction.ts#L236) means excludeFields: ['spender'] silently drops Spender too.

A Permit declaring both commits both addresses, and the handler this option defers to reads [message?.spender](https://github.com/MetaMask/metamask-extension/blob/c7a3997b612c5de5f8e5a2ff473d508e3885a5d9/app/scripts/lib/trust-signals/trust-signals-middleware.ts#L367) exactly while gating only on primaryType. So Spender is scanned by nobody.

Suggested fix: compare exactly and document that entries must match the declared name verbatim. Over-excluding fails silently; under-excluding only costs a duplicate scan.

packages/phishing-controller/src/signature-address-extraction.ts
```diff
@@ -0,0 +1,300 @@
...
+    value: unknown,
+    depth: number,
+  ): void {
+    visited += 1;
+    if (truncated(depth)) {
+      return;
+    }
+
+    // Handle one array dimension at a time, e.g. `address[]` or `Type[][]`.
+    const arrayMatch = type.match(/^(.*)\[\d*\]$/u);
```
adonesky1: `visitField` checks types in a different order than the signer and accepts fewer array shapes, so some address values get signed without being scanned.

[encodeField](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L267-L385) resolves custom types [first](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L277) and arrays [last](https://github.com/MetaMask/eth-sig-util/blob/v9.0.0/src/sign-typed-data.ts#L354-L361) via endsWith(']'); [visitField](https://github.com/MetaMask/core/blob/82d0fb4812922baceb6b76e8b18a4ffc086094db/packages/phishing-controller/src/signature-address-extraction.ts#L269-L290) inverts that and requires /^(.*)\[\d*\]$/u.

Two cases I verified via TypedDataUtils.eip712Hash. With type: 'address[abc]' the signer slices to address and commits, while the regex rejects it and no later branch matches, so we collect nothing. With a type declared as address[] the signer treats it as a struct and commits the inner address, while we array-match first and bail on the Array.isArray(value) guard. Neither sets overflow.

Suggested fix: mirror the signer's precedence (custom type, primitives, then endsWith(']') with lastIndexOf('[') slicing), and set overflow = true when an address-bearing type can't be walked, so silence stops looking like "no addresses"

packages/phishing-controller/src/signature-address-extraction.ts
```diff
@@ -0,0 +1,355 @@
...
+ */
+function normalizeAddress(value: unknown): string | undefined {
+  if (typeof value === 'string') {
+    const trimmed = value.trim();
+    if (HEX_STRING_REGEX.test(trimmed)) {
+      let digits = trimmed.slice(2);
+      if (digits.length % 2 === 1) {
+        digits = `0${digits}`;
+      }
+      return `0x${digits.slice(0, 40).padStart(40, '0').toLowerCase()}`;
```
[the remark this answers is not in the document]
wzrdk3lly: Checked this against `@metamask/eth-sig-util` `encodeField` / `TypedDataUtils.encodeData` (the same stack as the Donesky leading-20-byte fix).

The split is not even- vs odd-length hex. In the `@metamask/utils` this signer uses, **odd-length `0x`-hex is still `isStrictHexString`**, so it never hits `reallyStrangeAddressToBytes`. Both sides `hexToBytes` (pad the odd nibble) and take 20 bytes.

- `0x1` and 39-nibble `0x111…` encode and scan as the same address.
- Oversized `0x`-hex (`ADDR_A` + extra byte) **does not sign** (`Invalid address value … 21 bytes`). We still collect the leading 20; that is an extra scan of an unsignable value, not a signed spender missing the scanner.

Pinned in `9fb5b04` (`agrees with eth-sig-util on odd-length 0x-hex`, `does not treat oversized 0x-hex as a signable address`). No `normalizeAddress` change.

packages/phishing-controller/src/signature-address-extraction.ts
```diff
@@ -0,0 +1,355 @@
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+const HEX_STRING_REGEX = /^0x[0-9a-fA-F]+$/u;
+const DECIMAL_STRING_REGEX = /^[0-9]+$/u;
+
+// Cap the number of addresses returned for a single signature. A legitimate
+// signature references far fewer; exceeding this is treated as unusual and
+// surfaced to the caller (via `overflow`) rather than scanned in full.
+const MAX_SIGNATURE_ADDRESSES = 10;
```
adonesky1: Is 10 the right cap, and should callers be able to override it?

Using the Permit2 `PermitBatch` fixture from the extension's own `test/integration/confirmations/signatures/signature-helpers.ts`, `overflow` goes true at 11 tokens under the extension's `excludeFields: ['spender']` config, and at 10 under mobile's, which passes no exclusion so `spender` counts against the cap. Both clients render `overflow` as a user-visible "scan incomplete" warning, so a batch that size gets a caution with nothing actually wrong.

Could we add a `maxAddresses` option (or raise the default), and export the constant so callers can word that message accurately? I don't have data on how common batches that large are, so this may be rare in practice, but it seems easier to settle before two clients ship against the current behaviour.

adonesky1 (dismissed):
Just one more non-blocking comment. Otherwise LGTM

packages/phishing-controller/src/signature-address-extraction.ts
wzrdk3lly: Agreed, I’ll add an optional maxAddresses (default stays 10), export the constant so copy/telemetry can stay accurate, and put a hard ceiling so callers cannot set an unbounded value. Clients can keep omitting the option and get 10.

adonesky1 (dismissed):

adonesky1 (dismissed):

.github/actions/check-merge-queue-changelogs/action.yml
```diff
@@ -40,23 +40,20 @@ runs:
           const number = parseInt(match[1], 10);
           core.setOutput('pr-number', number);
 
-    - name: Get pull request branch
-      id: pr-branch
+    - name: Get pull request head ref
```
mcmire: Are these changes necessary to merge this PR? They seem unrelated.

.github/workflows/changelog-check.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         uses: MetaMask/github-tools/.github/actions/check-changelog@v1
         with:
           base-branch: ${{ github.event.pull_request.base.ref }}
-          head-ref: ${{ github.head_ref }}
+          head-ref: refs/pull/${{ github.event.pull_request.number }}/head
```
mcmire: Same for this one.

.github/actions/check-merge-queue-changelogs/action.yml
wzrdk3lly: These are not necessary. Dropping these now.

.github/workflows/changelog-check.yml
wzrdk3lly: Dropping this as well.

PR #9877, opened 2026-08-14
fix(accounts-controller): guard undefined scopes in multichain listing

rajanpanth:
## Summary

Adds a guard for undefined `scopes` when listing multichain accounts in the accounts controller flow.

## Why

Undefined scope payloads can surface in edge cases and currently trigger avoidable runtime failures.

---

> [!NOTE]
> **Low Risk**
> Narrow defensive change to multichain account filtering; only excludes malformed legacy accounts from chain-scoped lists.
> 
> **Overview**
> Fixes a **runtime `TypeError`** when `listMultichainAccounts` is called with a CAIP-2 chain ID while persisted state includes internal accounts whose `scopes` is missing (`undefined`), which can happen for legacy or partially migrated accounts.
> 
> Chain filtering now requires `Array.isArray(account.scopes)` before calling `isScopeEqualToAny`; accounts without valid scopes are **omitted** from filtered results instead of crashing. Unfiltered listing (no `chainId`) is unchanged.
> 
> A regression test and changelog entry document the fix (metamask-extension [#41962](https://github.com/MetaMask/metamask-extension/issues/41962)).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6f12e2b8fbea7c34d0c096ba2ef6d561d643b42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/accounts-controller/CHANGELOG.md
+++ b/packages/accounts-controller/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/utils` from `^11.11.0` to `^11.12.0` ([#10076](https://github.com/MetaMask/core/pull/10076))
 
+### Fixed
+
+- Prevent `listMultichainAccounts` from throwing a `TypeError` when an internal account has an undefined `scopes` field ([#41962](https://github.com/MetaMask/metamask-extension/issues/41962))
+
 ## [39.1.1]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


rajanpanth:
Rebased onto `main` (107 commits) and this is mergeable again.

The conflict was only in `packages/accounts-controller/CHANGELOG.md`: `39.1.1` was cut in the meantime, so my entry now sits under `[Unreleased]` with the released section untouched below it.

Since that is a lot of drift, I re-verified the fix rather than assuming it survived:

- `yarn workspace @metamask/accounts-controller run test` passes, coverage 100% statements / 100% lines.
- Reverting just the guard fails exactly one test, `listMultichainAccounts > does not throw when an account has an undefined scopes field and excludes it from chain-filtered results`, so the test still pins the behaviour rather than passing incidentally.

The change itself is unchanged in substance, an `Array.isArray` guard before `isScopeEqualToAny` so a legacy or partially-migrated account persisted without `scopes` is skipped instead of throwing a `TypeError`.

PR #9887, opened 2026-08-14
docs(releasing): clarify major-bump dependents in release PRs

rajanpanth:
## Summary
- clarify in `docs/processes/releasing.md` that when a package gets a major bump, impacted dependents should be included in the same release branch/PR
- add this note in both interactive and manual release workflows where dependency issues are reviewed

## Why
`create-release-branch` already flags dependent packages for breaking changes, but the contributor docs did not explicitly call out that these dependents should be released together.

Closes #8066.

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to the release process guide; no runtime, tooling, or release automation behavior changes.
> 
> **Overview**
> Updates **`docs/processes/releasing.md`** so contributors know that a **major version bump** (breaking change) should trigger including **dependent packages** in the same release branch/PR—not only peer-dependent edge cases.
> 
> The **interactive** workflow now says the UI will **strongly advise** including those dependents when resolving dependency issues. The **manual** workflow uses parallel wording (prompt to include dependents) and tightens how **missing dependencies** and **version incompatibilities** are described (errors vs. flags).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5db897c8c2047de7cbfaf9e34ac74c10a393e7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/processes/releasing.md
+++ b/docs/processes/releasing.md
@@ -39,8 +39,7 @@ This option provides a visual interface to streamline the release process:
 
    Common types of dependency issues you might encounter:
    - **Missing dependencies**: If you're releasing Package A that depends on Package B, the UI will prompt you to include Package B
-   - **Breaking change impacts**: If you're releasing Package B with breaking changes, the UI will identify packages that have peer dependencies on Package B that need to be updated
-   - **Major-version dependents**: If Package B is getting a major bump, include impacted dependents in the same release branch and update their version ranges/changelogs in that release PR
+   - **Breaking change impacts**: If you're releasing a package with breaking changes (and therefore bumping the major part of its version), the UI will identify dependents of that package and strongly advise you to also include them in the release
    - **Version incompatibilities**: The UI will flag if your selected version bumps don't follow semantic versioning rules relative to dependent packages
 
    Unlike the manual workflow where you need to repeatedly edit a YAML file, in the interactive mode you can quickly resolve these issues by checking boxes and selecting version bumps directly in the UI.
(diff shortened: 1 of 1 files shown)


docs/processes/releasing.md
```diff
@@ -40,6 +40,7 @@ This option provides a visual interface to streamline the release process:
    Common types of dependency issues you might encounter:
    - **Missing dependencies**: If you're releasing Package A that depends on Package B, the UI will prompt you to include Package B
    - **Breaking change impacts**: If you're releasing Package B with breaking changes, the UI will identify packages that have peer dependencies on Package B that need to be updated
+   - **Major-version dependents**: If Package B is getting a major bump, include impacted dependents in the same release branch and update their version ranges/changelogs in that release PR
```
mcmire: This is kind of covered by the previous point: a major version would only be issued if there are breaking changes. That said, it's mildly out of date.

Perhaps this would satisfy your concern?

``` suggestion
   - **Breaking change impacts**: If you're releasing a package with breaking changes (and therefore bumping the major part of its version), the UI will identify dependents of that package and strongly advise you to also include them in the release

File: docs/processes/releasing.md [hunk 2]
```diff
@@ -109,6 +110,7 @@ If you prefer more direct control over the release process:
...
    - **Version incompatibilities**: The UI will flag if your selected version bumps don't follow semantic versioning rules relative to dependent packages
```
mcmire: This is actually wrong — there is no UI in the manual flow. Maybe we should fix these while we're at it?

``` suggestion
   - **Missing dependencies**: If you're releasing Package A that depends on Package B, you will be prompted to include Package B
   - **Breaking change impacts**: If you're releasing a package with breaking changes (and therefore bumping the major part of its version), you will be asked to include dependents of that package in the release
   - **Version incompatibilities**: You will receive an error if your selected version bumps don't follow semantic versioning rules relative to dependent packages

mcmire (commented):
Hey, thanks for the PR. I just had some suggestions.

docs/processes/releasing.md
rajanpanth: Updated the interactive-flow guidance to use the breaking-change framing and make the dependent-package recommendation explicit.

docs/processes/releasing.md
rajanpanth: Updated all three manual-flow bullets to remove the UI wording and describe the command-line prompts and errors.

rajanpanth:
Thanks for the suggestions — I pushed the requested wording updates in both sections and replied inline on each thread. Could you take another look when you have a moment?

rajanpanth:
Thanks for the detailed suggestions. I incorporated the requested wording corrections in the latest commit, including the manual-flow wording. All visible checks are passing. Could you please take another look when convenient?

PR #9889, opened 2026-08-15
feat(perps-controller): add Lighter perps venue (initial implementation, flag-gated)

abretonc7s:
## Explanation

**TAT-3766: add Lighter as a second perps venue.** Started as a spike, now a full `LighterProvider` in `@metamask/perps-controller`, hardened through 23 rounds of adversarial cross-review and validated end-to-end against live Lighter testnet.

### What's in it

- **Full trading surface** behind the `PerpsProvider` interface: place/cancel/close orders, venue-linked OCO TP/SL (grouped trigger orders), isolated margin add/remove, leverage updates, signed L2 withdraw, historical orders/fills/funding/ledger, and live streams (prices, account, positions, orders, fills, order book, candles) over a shared WebSocket with REST-polling fallback.
- **Signer model**: deterministic venue-key derivation via `personal_sign` (raw EVM key never leaves the keyring — hardware-wallet compatible). The Go/WASM signer is injected through a `LighterSignerBridge` seam so mobile's WebView bridge and the headless Node adapter are interchangeable.
- **Financial-safety architecture** (the bulk of the review rounds): a durable nonce-dispatch ledger with exact tx-hash reconciliation — an ambiguous dispatch is never blindly retried; proven-committed outcomes are quarantined until explicitly acknowledged per-outcome (`getRecoveredDispatches`/`acknowledgeRecoveredDispatch`); a TP/SL settlement journal that survives crashes and restores nothing it cannot prove, parking unprovable protection as durable manual recovery (`getPendingManualRecoveries`); process-wide serialization of venue writes, the ledger, and the bridge's singleton WASM client; session fences on every account switch.
- **Controller wiring behind a flag**: `PerpsProviderType` gains `'lighter'`; enablement via `providerCredentials.lighter.enabled` or the `perpsLighterProviderEnabled` remote flag. New controller actions expose the safety states to clients.

### How it was validated

- **Unit**: 213 provider tests (3,031 package-wide), including deterministic fault-injection proofs for response loss, masked commits, restarts, account switches mid-flight, concurrent providers, and storage failures. Every review-round fix was proven RED against the prior commit before landing.
- **Live testnet e2e** (real venue, real WASM signer built from source): sign-only 9/9 (pinned WASM identity contract), key registration, order-lifecycle 7/7, TP/SL 12/12 (including a real venue-linked OCO pair), close-position 5/5 with both fills observed on the trades stream, margin/leverage 7/7 on a fresh faucet account.
- **Mainnet read path**: 227 active perp markets, full-catalog WS snapshot, candles.
- **On-device (iOS simulator)**: live prices/candles/positions over WS-only transport; WASM signer mount→ready ~2.1 s, round-trip 32–78 ms.

### Open items

- `editOrder` deliberately returns an error: Lighter accepts `ModifyOrder` submissions without executing the reprice — raised with the venue; cancel + re-place instead.
- Mobile client integration (WebView signer bridge + recovery-state UI): MetaMask/metamask-mobile#34865 (draft, consumes this package).

## References

- https://consensyssoftware.atlassian.net/browse/TAT-3766
- Signer reference: https://github.com/elliottech/lighter-go/tree/web-wasm
- Risk context: https://l2beat.com/scaling/projects/lighter

## Checklist

- [x] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them — not applicable: no breaking changes

---

> [!NOTE]
> **High Risk**
> Introduces a new trading venue, client signer bridge, and durable dispatch/recovery flows for ambiguous L2 submissions—areas that affect funds, leverage, and order settlement even when feature-gated.
> 
> **Overview**
> Adds **Lighter** as an opt-in perps venue (`PerpsProviderType` now includes `'lighter'`, a breaking change for exhaustive switches). Registration mirrors MYX: dynamic import of `LighterProvider`, gated by `providerCredentials.lighter.enabled` or remote `perpsLighterProviderEnabled` **only when** the client supplies `LighterSignerBridge` (venue keys stay client-side; no deposit ads until bridge deposit flow exists). `PROVIDER_CONFIG.LIGHTER_TESTNET_ONLY` and cache keys follow the same pattern as MYX.
> 
> Exposes **durable settlement** on the controller and aggregator: `getPendingManualRecoveries`, `getRecoveredDispatches`, and `acknowledgeRecoveredDispatch`, with exported `PerpsPendingManualRecovery` / `PerpsRecoveredDispatch` types. Aggregated mode fans these out across providers that implement the contract.
> 
> **Multi-provider routing** for margin math: optional `providerId` on liquidation/maintenance params and on `getMaxLeverage(asset, providerId?)`, with `AggregatedPerpsProvider` delegating to the owning venue instead of always using the default.
> 
> Ships Lighter **constants/types exports** (`lighterConfig`, signer bridge types) and changelog notes for the initial (disabled-by-default) implementation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8a378e3094347614c82d531e975301d654762fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/constants/perpsConfig.ts
```diff
@@ -608,6 +608,12 @@ export const PROVIDER_CONFIG = {
   DefaultProvider: 'hyperliquid' as const,
   /** Force MYX to testnet only (mainnet credentials not yet available) */
   MYX_TESTNET_ONLY: false,
+  /**
+   * Force Lighter to testnet only. Off: Lighter follows the global network
+   * toggle so mainnet reads (full market catalog, prices, candles) work;
+   * writes stay testnet-gated inside LighterProvider (POC).
+   */
+  LIGHTER_TESTNET_ONLY: false,
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 50624d3731 — every nonce-consuming write (including key registration) is now refused on mainnet at the venue write lock; mainnet stays read-only until mainnet trading is validated end-to-end.

packages/perps-controller/src/PerpsController.ts
```diff
@@ -3347,6 +3484,54 @@ export class PerpsController extends BaseController<
...
+   */
+  async acknowledgeRecoveredDispatch(recoveryId: string): Promise<void> {
+    const provider = await this.#getActiveProviderWhenReady();
+    if (!provider.acknowledgeRecoveredDispatch) {
+      throw new Error(
+        'The active perps provider has no recovered dispatches to acknowledge',
+      );
+    }
+    return provider.acknowledgeRecoveredDispatch(recoveryId);
+  }
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 50624d3731 — AggregatedPerpsProvider now forwards getPendingManualRecoveries/getRecoveredDispatches (aggregated, errors propagate) and routes acknowledgeRecoveredDispatch to the owning provider.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,7814 @@
...
+    // INITIAL ROLLOUT GATE: every nonce-consuming venue write (including
+    // signer-key registration) is limited to testnet. Mainnet stays
+    // read-only until mainnet writes have been validated end-to-end —
+    // the enablement flags alone must not be able to unlock unvalidated
+    // mainnet trading.
+    if (!this.#isTestnet) {
+      throw new Error(
+        'Lighter mainnet trading is not enabled yet; venue writes are limited to testnet',
+      );
+    }
```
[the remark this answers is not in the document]
abretonc7s: Fixed in a045856429 — signer setup may enter the lock on mainnet (bridge-local client creation + read-only nonce fetch) so auth-token minting and authenticated reads work with a registered key; the gate moved to a dispatch backstop inside submit, so key registration and every other write are still refused before anything reaches the venue.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,7827 @@
...
+        // happened while SIGNING must abort before submission.
+        this.#assertSession(generationAtIntent);
+        // Mainnet dispatch backstop: covers the signer-setup path that
+        // is allowed to ENTER the lock on mainnet — nothing may be
+        // submitted there.
+        if (!this.#isTestnet) {
+          throw new Error(
+            'Lighter mainnet trading is not enabled yet; venue writes are limited to testnet',
+          );
+        }
```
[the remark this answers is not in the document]
abretonc7s: Fixed in ce81a72d7d — registration is refused before the L1 personal_sign and ChangePubKey signing on mainnet; no wallet/hardware prompt for a dispatch the gate is guaranteed to refuse.

packages/perps-controller/tests/e2e/lighter/build-wasm.sh
```diff
@@ -0,0 +1,107 @@
...
+if git -C "$REPO_DIR" cat-file -e "HEAD:web-wasm/main.wasm" 2>/dev/null; then
+  git -C "$REPO_DIR" show "HEAD:web-wasm/main.wasm" > "$OUT_DIR/upstream-main.wasm"
+  COMMITTED_SHA="$(shasum -a 256 "$OUT_DIR/upstream-main.wasm" | awk '{print $1}')"
+fi
+# Re-run the build so the artifact we ship is unambiguously source-built,
+# and prove SELF-reproducibility: a forced full recompile must produce
+# the identical hash.
+(cd "$REPO_DIR/web-wasm" && GOOS=js GOARCH=wasm GOTOOLCHAIN=go1.26.0 go build -trimpath -ldflags="-s -w" -o main.wasm)
+FIRST_SHA="$(shasum -a 256 "$REPO_DIR/web-wasm/main.wasm" | awk '{print $1}')"
+(cd "$REPO_DIR/web-wasm" && GOOS=js GOARCH=wasm GOTOOLCHAIN=go1.26.0 go build -a -trimpath -ldflags="-s -w" -o main.wasm)
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4c097f83c5. All Go build, GOROOT, and version lookups now use the same pinned go1.26.0 toolchain.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,7890 @@
...
+        if (!verifiedFullClose) {
+          return {
+            isValid: false,
+            error: `Order size ${requestedSize} is below the Lighter minimum of ${minSize} ${params.symbol}`,
+          };
+        }
+      }
+      // Wire-format parity: placement integerizes size and the
+      // slippage-adjusted EXECUTION
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4c097f83c5. A testnet-bound Lighter provider now returns no bridge routes even when a caller requests mainnet, while mainnet providers retain their route.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8003 @@
...
+  }
+  return {
+    time,
+    open: String(candle.o),
+    high: String(candle.h),
+    low: String(candle.l),
+    close: String(candle.c),
+    volume: String(candle.v),
+  };
+};
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4c097f83c5. Candle fields now use strict finite parsing and serialize the validated numeric values, rejecting null, blank, boolean, and malformed inputs.

packages/perps-controller/recipes/lighter-e2e.recipe.json
```diff
@@ -0,0 +1,553 @@
...
+        "next": "assert-sign-only-checks",
+        "intent": "Verify the sign-only phase exited green"
+      },
+      "assert-sign-only-checks": {
+        "action": "assert_output",
+        "source": "sign-only",
+        "stream": "stdout",
+        "contains": "PHASE_PASS: sign-only (7/7 checks)",
+        "next": "assert-sign-only-json",
+        "intent": "Verify all 7 sign-only checks passed"
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4c097f83c5. The sign-only recipe assertion and intent now require the current 9/9 completion marker.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8009 @@
...
+      const leverageImfHundredths = await this.#resolveLeverageIntent(params);
+      // The app manages ISOLATED positions only (there is no cross-margin
+      // management UI, and small cross positions report no liquidation
+      // price), so a flat market opens isolated. The venue refuses
+      // changing the mode of a market with an open position, so an
+      // existing position keeps whatever mode it already has.
+      const leverageMarginMode =
+        leverageImfHundredths === null
+          ? LIGHTER_MARGIN_MODE_ISOLATED
+          : await this.#resolveMarginModeForSymbol(params.symbol);
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4c097f83c5. Position adaptation now maps venue marginMode 1 to isolated and preserves cross for 0 or older omitted fields.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8037 @@
...
+              );
+            } else if (lookedUp.status === 3) {
+              quarantine(entry, 'succeeded', 'tx-status:3');
+            } else {
+              quarantine(
+                entry,
+                'unknown',
+                `tx-status:${String(lookedUp.status ?? -1)}`,
+              );
+            }
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 07f530f140.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+      return {
+        isValid: false,
+        error:
+          'Lighter does not support TP/SL attached at placement; place the order, then call updatePositionTPSL',
+      };
+    }
+    if (params.timeInForce === 'ALO') {
+      return {
+        isValid: false,
+
```
geositta, line 7943: **[P1] Enforce order-book nonce continuity before applying this delta.** Lighter documents that each update's `begin_nonce` must equal the previous update's `nonce`; this implementation omits both fields from `LighterWsOrderBookMessage` and applies every delta to the local maps. If one frame is dropped while the socket remains connected, deleted and changed levels remain indefinitely, so the depth chart, best bid/ask, spread, and mid-price become silently wrong. Please retain the last nonce per market and, on a mismatch, discard the local book and resubscribe/reconnect until a new `subscribed/order_book` snapshot arrives. Add a test that skips one nonce and proves no mixed snapshot is emitted. Official contract: https://apidocs.lighter.xyz/docs/websocket-reference#order-book. This defect is introduced by this PR's new order-book stream handler.

packages/perps-controller/src/services/LighterClientService.ts
```diff
@@ -0,0 +1,505 @@
...
+
+    try {
+      const response = await fetch(url, {
+        method: init.method,
+        headers: init.headers,
+        body: init.body,
+        signal: controller.signal,
+      });
+
+      const payload = convertKeysToCamelCase(await response.json()) as Result;
```
geositta: **[P1] Decode this untrusted JSON instead of asserting the caller-selected generic type.** `convertKeysToCamelCase(...) as Result` validates only `code`; it does not prove endpoint fields such as `nonce`, `accounts`, `orderBooks`, `orders`, or `txHash`. A schema change or malformed `code: 200` response therefore reaches signing/trading logic as trusted data—for example, `getNextNonce()` can return an absent/non-numeric nonce despite its declared type. Following Total TypeScript's boundary guidance, keep the payload `unknown` and use endpoint-specific Superstruct decoders (including finite/integer/range checks) that fail closed before returning. Add malformed-success-response tests for nonce and transaction responses. This unsafe boundary is newly added by this PR.

packages/perps-controller/tests/e2e/lighter.e2e.ts
```diff
@@ -0,0 +1,2110 @@
...
+
+const PHASE = args.get('phase') ?? 'sign-only';
+const OUT_DIR = resolve(
+  args.get('out') ?? join(REPO_ROOT, 'temp', 'lighter-e2e'),
+);
+const MARKET = args.get('market') ?? 'SOL';
+const WASM_DIR = resolve(
+  args.get('wasm-dir') ?? join(REPO_ROOT, 'temp', 'lighter-wasm'),
+);
+const ETH_KEY = (args.get('eth-key') ?? DEFAULT_DUMMY_ETH_KEY) as `0x${string}`;
```
geositta: **[P1] Do not accept a private key through `--eth-key`.** Command-line arguments are visible in shell history and to local process inspection, so the documented path exposes any non-dummy funded testnet key supplied to this driver. Read a dedicated environment variable or protected file descriptor instead, keep the public dummy key as an explicit opt-in, and fail with a clear message when write phases require credentials. This credential-ingress path is introduced by this PR.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+};
+
+// ============================================================================
+// LighterProvider
+// ============================================================================
+
+/**
+ * Lighter provider implementation (POC).
+ */
+export class LighterProvider implements PerpsProvider {
```
geositta, line 966: **[P2, non-blocking architecture concern] Separate the independent state machines represented by this class.** This new 8,149-line provider has roughly 100 private members and owns HTTP decoding, WebSocket assembly, account-session binding, bridge ownership, nonce allocation, durable dispatch recovery, TP/SL settlement, validation, calculations, and public provider orchestration. Those concerns share mutable state, so a change in one protocol can invalidate financial invariants in another and reviewers cannot reason locally about correctness. Please track extraction of typed collaborators such as `LighterSession`, `LighterDispatchLedger`, `LighterTpslSettlement`, and `LighterStreamClient`; keep `LighterProvider` as composition/orchestration. I am marking this non-blocking for this PR because the extraction itself carries financial regression risk, but the current shape should not become the template for additional venue behavior.

geositta (changes requested):
Thank you for the extensive failure-injection coverage and the work to make ambiguous financial dispatches fail closed. I am requesting changes because two remaining boundary defects can present incorrect venue state or allow malformed venue data into signing and trading paths.

Required before merge:
- Enforce Lighter order-book `begin_nonce`/`nonce` continuity and obtain a fresh snapshot after a gap. Without this, a dropped delta leaves pro traders looking at a silently corrupted depth book, spread, and mid-price.
- Decode REST responses from `unknown` with endpoint-specific runtime validators before assigning TypeScript response types. The current generic assertion makes untrusted JSON appear valid and permits missing or malformed nonce, account, market, and transaction fields to reach financial logic.
- Stop accepting private keys through command-line arguments in the e2e driver; process arguments and shell history expose them.

I also left a non-blocking design comment on the 8,149-line provider. Its signer, persistence, recovery, REST, WebSocket, validation, and trading state machines should be separated behind explicit typed interfaces so future venue changes can be reviewed and tested in isolation.

`yarn workspace @metamask/perps-controller run test` passes locally. Current CI checks are green. This PR does not change React or React Native components, so there are no component-level React findings. All findings are in files introduced by this branch relative to `origin/main`.

geositta (commented):
Follow-up review identified additional correctness blockers after the initial submission. The highest-risk items are the Lighter mainnet deposit transaction construction, transaction-status decoding, and automatic venue-key replacement after a failed key lookup. These need resolution before the earlier change request can be cleared. The comments below are limited to behavior introduced by this branch and verified against the current Lighter documentation.

packages/perps-controller/src/constants/lighterConfig.ts
```diff
@@ -0,0 +1,369 @@
...
+ * `GET /api/v1/layer1BasicInfo` (contract addresses) and the venue docs
+ * (minimums). Mainnet settles against Ethereum L1; testnet runs on a
+ * venue-hosted devnet L1 (chain id 123456), so its route is informational.
+ */
+export const LIGHTER_BRIDGE_CONFIG = {
+  mainnet: {
+    /** CAIP-2 chain the bridge contract lives on (Ethereum mainnet). */
+    chainId: 'eip155:1',
+    /** ZkLighter L1 contract (deposits via `deposit`, selector 0x8a857083). */
+    bridgeContract: '0x3B4D794a66304F130a4Db8F2551B0070dfCf5ca7',
```
geositta, line 359: **[P0] Do not advertise this deposit route until the transaction builder calls Lighter's bridge contract.** `DepositService.prepareTransaction` currently builds `ERC20.transfer(bridgeContract, 0)` and sends it to the USDC token contract. Lighter requires an approval followed by `deposit(_to, _assetIndex, _routeType, _amount)` on this proxy (`0x8a857083`). A mainnet user can therefore transfer USDC to the bridge address without executing a deposit that credits their Lighter account. Please move deposit transaction construction behind a provider-owned contract or suppress the route until the approval + bridge-deposit flow is implemented and integration-tested. Official flow: https://apidocs.lighter.xyz/docs/deposits-transfers-and-withdrawals. This unsafe route becomes reachable through this PR's new config.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+            lookedUp.apiKeyIndex === this.#apiKeyIndex &&
+            lookedUp.nonce === entry.nonce &&
+            typeof lookedUp.status === 'number';
+          if (matchesIdentity) {
+            // A matching hash is not enough to prove the transaction's
+            // outcome. Non-terminal statuses must retain the unresolved
+            // entry so no acknowledgment can make the intent retryable
+            // while the original transaction is still in flight.
+            if (
+              lookedUp.status !== 3 &&
```
geositta: **[P1] Correct the transaction-status mapping before using it for recovery.** Lighter documents `0 = Failed`, `1 = Pending`, `2 = Executed`, and `3 = Pending - Final State`; this code treats 3 as succeeded, 4/5 as failed, and leaves 0/2 unresolved. After response loss, an executed transaction can block writes indefinitely, while a pending-final transaction can be acknowledged as completed. Please centralize the documented status union and use it here and in TP/SL recovery, with a test for every status. Source: https://apidocs.lighter.xyz/docs/data-structures-constants-and-errors#transaction-status-mapping. The incorrect reconciler is entirely introduced by this PR.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+      const response = await this.#clientService.getApiKeys(
+        accountIndex,
+        this.#apiKeyIndex,
+      );
+      return response.apiKeys.some(
+        (key) =>
+          key.apiKeyIndex === this.#apiKeyIndex &&
+          key.publicKey === this.#venuePublicKey,
+      );
+    } catch {
```
geositta: **[P1] Fail closed when the venue-key lookup fails.** This catch converts transport, authentication, rate-limit, and malformed-response errors into an authoritative `false`; `#setupSigner` then calls `#registerVenueKey` and submits `ChangePubKey`. A transient read failure can therefore replace a valid key, invalidate another client/device, and consume a nonce. Only a validated successful response that proves the configured slot does not contain this key should return `false`; all lookup errors must propagate. This rotation path is introduced by the new provider.

packages/perps-controller/src/utils/lighterAdapter.ts
```diff
@@ -0,0 +1,618 @@
...
+    case 'canceled-post-only':
+    case 'canceled-reduce-only':
+    case 'canceled-position-not-allowed':
+    case 'canceled-margin-not-allowed':
+    case 'canceled-too-much-slippage':
+    case 'canceled-not-enough-liquidity':
+    case 'canceled-self-trade':
+    case 'canceled-expired':
+      return 'canceled';
+    default:
```
geositta, line 606: **[P1] Do not map unhandled terminal states to `open`.** Lighter currently documents terminal values including `canceled-oco`, `canceled-child`, `canceled-liquidation`, and `canceled-invalid-balance`; none is handled above, so each becomes an actionable open order in the Pro UI and remains in the live-order cache. Please decode the documented status union exhaustively and fail with a data-integrity error for an unknown value instead of asserting that it is live. Source: https://apidocs.lighter.xyz/docs/websocket-reference#order-json. This adapter and fallback are new in this PR.

packages/perps-controller/src/services/LighterWalletService.ts
```diff
@@ -0,0 +1,165 @@
...
+   */
+  async deriveKeySeed(apiKeyIndex: number): Promise<string> {
+    const address = this.getUserAddress();
+    const message = buildLighterKeyDerivationMessage({
+      address,
+      chainId: getLighterChainId(this.network),
+      apiKeyIndex,
+    });
+    const signature = await this.signPersonalMessage(message);
+    const seedBytes = await sha256(hexToBytes(signature));
```
geositta: **[P1] Do not make account recovery depend on byte-identical `personal_sign` output.** EIP-191 defines the signed message format, but it does not require deterministic ECDSA nonces or identical signature encoding across software and hardware signers. Two valid signatures can therefore hash to different seeds, causing the next device to derive another Lighter key and enter the automatic replacement path. Please use a stable persisted/recoverable key mechanism, or constrain and verify the signer contract before registration and document the resulting wallet limitations. This derivation scheme is introduced by this PR.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+      });
+      return { success: false, error: wrappedError.message };
+    }
+  }
+
+  // ============================================================================
+  // History Operations (POC: stubbed)
+  // ============================================================================
+
+  async getOrderFills(
```
geositta, line 6516: **[P1] Honor the account and history query contract instead of silently reading the selected wallet's first page.** This method ignores `accountId`, `startTime`, `endTime`, `symbol`, and aggregation/pagination fields and forwards only `limit`. Account-pinned cache requests can therefore be populated with another selected account after a switch, while date or symbol pagination repeats unrelated recent fills. Please implement these parameters or explicitly reject unsupported combinations, and add account-switch, symbol, and range tests. The mismatch is introduced by this new provider method.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8149 @@
...
+    // binding: they compare this generation after their awaits and retry
+    // instead of caching results for the wrong account.
+    this.#sessionGeneration += 1;
+    this.#accountIndex = null;
+    this.#signerReadyPromise = null;
+    this.#authToken = null;
+    // #tpslUnsettled is NOT cleared: entries are keyed by
+    // address+accountIndex+symbol, so B never consumes A's pending ids and
+    // switching back to A retains its reconciliation obligation.
+    this.#teardownStream();
```
geositta, line 1433: **[P1] Clear subscriber-visible account state synchronously when the wallet binding changes.** Teardown clears internal maps but emits no empty account, position, or order snapshot. Consumers keep showing account A's balances and positions under account B until every new authenticated channel succeeds; on an auth/capability failure, stale financial state can remain indefinitely. Please emit the reset before rebuilding channels and test a switch whose new subscription fails. This stale-state path is introduced by the PR's new session binding.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in d860bdfb40. The order-book stream now enforces begin_nonce continuity, discards a gapped local book, and resubscribes for a fresh snapshot; the skipped-nonce test verifies no mixed book is emitted.

packages/perps-controller/src/services/LighterClientService.ts
abretonc7s: Fixed in d860bdfb40. LighterClientService now decodes each successful endpoint response with Superstruct before returning typed data, including strict nonce, account, market, order, transaction, and sendTx validation with malformed-success tests.

packages/perps-controller/tests/e2e/lighter.e2e.ts
abretonc7s: Fixed in d860bdfb40. The e2e driver rejects --eth-key, reads LIGHTER_E2E_ETH_KEY instead, and requires explicit --use-dummy-eth-key opt-in for the public dummy account.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Tracked as a follow-up. This PR keeps the provider intact because extracting its financial state machines now would add regression risk; new behavior should move behind typed collaborators rather than expanding the class further.

packages/perps-controller/src/constants/lighterConfig.ts
abretonc7s: Fixed in d860bdfb40. Lighter now advertises no deposit route until a provider-owned approval plus bridge-deposit transaction builder is implemented and integration-tested; withdrawal routing remains unchanged.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in d860bdfb40. Transaction status handling now centralizes the documented 0 failed, 1 pending, 2 executed, and 3 pending-final mapping and uses it in generic and TP/SL recovery, with coverage for every value.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in d860bdfb40. API-key lookup failures now propagate, an empty validated slot alone permits registration, and a different existing key fails explicitly instead of triggering automatic replacement.

packages/perps-controller/src/utils/lighterAdapter.ts
abretonc7s: Fixed in d860bdfb40. All documented canceled terminal statuses map to canceled, while unknown statuses throw a Lighter data-integrity error instead of becoming open orders.

packages/perps-controller/src/services/LighterWalletService.ts
abretonc7s: Fixed in d860bdfb40. Core no longer derives a venue key from personal_sign. The client-injected signer bridge now owns key creation and persistence; Core passes only named public session metadata and refuses mismatched registered keys.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in d860bdfb40. Fill reads validate the selected account, reject unsupported aggregation, honor limit and time filters, and apply the symbol filter on getOrFetchFills, with account, range, and symbol coverage.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in d860bdfb40. Wallet rebinding now synchronously emits empty account, position, and order state before socket teardown/rebuild, including the failing-new-subscription case.

packages/perps-controller/src/constants/lighterConfig.ts
```diff
@@ -0,0 +1,369 @@
...
+    LighterOrderBookMeta,
+    'minBaseAmount' | 'minQuoteAmount' | 'supportedSizeDecimals'
+  >,
+  price: number,
+): number {
+  const minBase = parseFloat(market.minBaseAmount);
+  const minQuote = parseFloat(market.minQuoteAmount);
+  const step = 10 ** -market.supportedSizeDecimals;
+  const byQuote = price > 0 ? minQuote / price : minBase;
+  const raw = Math.max(minBase, byQuote);
```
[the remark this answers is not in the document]
abretonc7s: Fixed in a95b9b5584. Minimum-size calculation now strictly parses and validates venue minima, price, and size decimals, throwing a data-integrity error for malformed or non-finite input; focused tests cover prefix-numeric, missing, NaN, and zero cases.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8241 @@
...
+        // is exact user intent and is never adjusted here.
+        requestedSize = snapToLighterSizeGrid(
+          usdAmount / referencePrice,
+          market.supportedSizeDecimals,
+        );
+      }
+      if (!(requestedSize > 0)) {
+        return { success: false, error: 'Order size must be positive' };
+      }
+      const minSize = computeLighterMinOrderSize(market, referencePrice);
```
geositta: **[P1] Apply these venue minimums only to maker orders.** Lighter explicitly documents `min_base_amount` and `min_quote_amount` as maker-only constraints, but this branch applies `computeLighterMinOrderSize` to market/IOC orders as well. A legitimate small taker order or partial reduce-only close is therefore rejected locally even though the venue permits it. Please condition this check on maker order semantics and add tests proving below-maker-minimum market opens and partial closes reach signing without changing the existing limit-order guard. Source: https://apidocs.lighter.xyz/docs/trading. This behavior is introduced by the Lighter provider in this PR.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8241 @@
...
+    ) {
+      throw new Error('Lighter fill history startTime must not exceed endTime');
+    }
+    try {
+      this.#ensureSessionBinding();
+      const generation = this.#sessionGeneration;
+      const accountIndex = await this.#ensureAccountIndex();
+      const token = await this.#getAuthToken();
+      await this.#ensureMarkets();
+      const response = await this.#clientService.getTrades(
```
geositta: **[P1] Fetch the requested history range instead of filtering one newest page.** This call retrieves only `limit` recent trades and then applies `startTime`, `endTime`, and symbol filtering in memory. If the newest page is outside the requested range, older matching fills are never queried and the API returns a false empty/incomplete result. Lighter's trades endpoint supports `cursor`/`from`; please page until the requested boundary is covered, or use the export endpoint for bounded time ranges, and preserve the requested limit after filtering. Add a test where the first page contains no match and the second page does. This is the incomplete portion of the previous fill-contract finding.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8241 @@
...
+      ),
+    );
+  };
+
+  async getOpenOrders(_params?: GetOrdersParams): Promise<Order[]> {
+    // Public reads re-kick pending journal recovery (deduped, detached).
+    this.#kickTpslRecovery();
+    try {
+      return await this.#readOpenOrdersStrict();
+    } catch (caughtError) {
```
geositta, line 4884: **[P1] Preserve fail-closed decoding through the provider boundary.** The adapter now correctly throws for unknown order states, but this catch converts that integrity failure—and REST schema failures from `LighterClientService`—to `[]`, making malformed venue data indistinguishable from no open orders. The WebSocket snapshot path has the related transactional issue at line 7617: it clears `#wsOrders` before all orders adapt successfully, so one invalid entry can leave the cache empty/partial. Please classify and rethrow data-integrity failures, and adapt a complete snapshot into a temporary map before replacing visible state. Add REST and WS tests showing the previous valid state is not replaced by malformed data.

packages/perps-controller/src/types/lighter-types.ts
```diff
@@ -0,0 +1,746 @@
...
+  createClient(
+    params: LighterCreateClientParams,
+  ): Promise<LighterCreateClientResult>;
+
+  /**
+   * Execute one WASM signer call and resolve with its result object.
+   *
+   * @param call - The function name and positional params to invoke.
+   */
+  execute<Result>(call: LighterWasmCall): Promise<Result>;
```
geositta: **[P1] Make the signer operation determine its arguments and result type.** `execute<Result>(LighterWasmCall)` lets the caller independently choose an arbitrary result type while `function` is any string and `params` is `unknown[]`; misspelled functions, shifted numeric arguments, and incompatible responses all compile. A React Native WebView implementation cannot truthfully implement this contract without an unsafe assertion. Please define a function-keyed operation map/discriminated union with exact parameter tuples and result types, or return `unknown` and runtime-decode each signer response before financial code consumes it. This public unsound API is introduced by this PR.

packages/perps-controller/src/services/LighterClientService.ts
```diff
@@ -0,0 +1,814 @@
...
+  timestamp: NonNegativeIntegerStruct,
+  triggerPrice: optional(string()),
+  orderId: optional(string()),
+  parentOrderIndex: optional(NonNegativeIntegerStruct),
+  parentOrderId: optional(string()),
+  toCancelOrderId0: optional(string()),
+  toTriggerOrderId0: optional(string()),
+  toTriggerOrderId1: optional(string()),
+});
+const TradeStruct = type({
```
geositta, line 206: **[P1] Validate every financial field consumed by the fill adapter.** The endpoint-level decoder is a strong improvement, but `TradeStruct` validates price and size only as arbitrary strings and omits fields the adapter trusts (`askAccountPnl`, `bidAccountPnl`, maker/taker fees, position-before values, and sign-change flags). Superstruct therefore accepts malformed values such as `size: '1oops'` or an unchecked non-numeric fee; downstream `parseFloat` can truncate them or produce `NaN`, and missing PnL silently becomes zero. Please add strict decimal/domain structs and include every consumed optional field in this schema. Apply the same strict decimal type to market minima and account/margin values so typed payloads represent validated financial data, not merely strings. This is the remaining portion of the prior REST-boundary finding.

geositta (changes requested):
Thank you for addressing the prior review. I re-verified the updated branch and confirmed that the deposit route, transaction-status mapping, key-registration behavior, signer-key ownership, account-switch reset, order-book continuity, documented order statuses, and e2e credential handling are now corrected.

I am keeping the review at **Request changes** for five remaining items:

- Lighter's maker-only minimums are still applied to market/IOC orders, which rejects valid small taker orders and partial closes.
- Fill-history time and symbol filters operate on only the newest page, so valid older results can be reported as absent.
- Newly strict order decoding is not applied transactionally: REST errors become empty state, and a malformed WebSocket snapshot can clear valid cached orders before failing.
- The public WASM bridge still lets callers independently assert any result type for any function and argument list.
- The new REST schemas verify that financial decimals are strings but do not verify their numeric grammar or domain before adapters consume them.

These are concrete correctness and contract-boundary issues rather than requests to broaden the feature. The package test suite passes locally and the latest CI is green; additional tests are needed for these failure cases. No React or React Native component code changed in this revision, but the bridge typing directly affects the safety of React Native WebView implementations.

packages/perps-controller/src/utils/lighterAdapter.ts
```diff
@@ -0,0 +1,624 @@
...
+    case 'canceled-expired':
+    case 'canceled-oco':
+    case 'canceled-child':
+    case 'canceled-liquidation':
+    case 'canceled-invalid-balance':
+      return 'canceled';
+    default:
+      throw new Error(
+        `${LIGHTER_DATA_INTEGRITY_PREFIX} unknown order status '${status}'`,
+      );
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 521b8c65f2. REST integrity errors now rethrow, and WebSocket order snapshots replace cached state only after every row adapts successfully.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in 521b8c65f2. Maker minimums now apply only to resting limit orders; market, IOC, and partial market closes keep their requested size.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in 521b8c65f2. Fill reads now follow venue cursors until the requested range is covered, apply the market filter, and preserve the caller limit after filtering.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in 521b8c65f2. Provider reads preserve data-integrity failures, and malformed WebSocket snapshots cannot clear or partially replace the previous valid order state.

packages/perps-controller/src/types/lighter-types.ts
abretonc7s: Fixed in 521b8c65f2. The signer bridge now uses named operation tuples and function-keyed result types; client creation returns public registration data only, while the client owns the venue key.

packages/perps-controller/src/services/LighterClientService.ts
abretonc7s: Fixed in 521b8c65f2. Endpoint decoders now validate strict finite decimal grammar for every financial field consumed by the market, account, order, fill, funding, and transfer adapters.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8403 @@
...
+    let address: string;
+    try {
+      address = this.#walletService.getUserAddress().toLowerCase();
+    } catch {
+      if (this.#boundAddress !== null) {
+        // All accounts deselected while a session existed: invalidate so
+        // nothing in flight can still act for the old account.
+        this.#emitAccountBindingReset();
+        this.#invalidateSessionState();
+        this.#removeSignerResetListener();
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 42d8e2873e. Reselection now re-arms one signer reset listener, and tests cover reset-driven signer recreation, deduplication, and disconnect cleanup.

packages/perps-controller/src/services/LighterClientService.ts
```diff
@@ -0,0 +1,847 @@
...
+  timestamp: NonNegativeIntegerStruct,
+  triggerPrice: optional(StrictDecimalStringStruct),
+  orderId: optional(string()),
+  parentOrderIndex: optional(NonNegativeIntegerStruct),
+  parentOrderId: optional(string()),
+  toCancelOrderId0: optional(string()),
+  toTriggerOrderId0: optional(string()),
+  toTriggerOrderId1: optional(string()),
+});
+const TradeStruct = type({
```
geositta, line 206: **[P1] Encode financial domains and required trade context in this schema.** `StrictDecimalStringStruct` accepts negative sizes, prices, balances, and market minima because it proves only finite decimal syntax. This trade schema also makes documented PnL and position-before/sign-change fields optional; downstream adaptation turns absent PnL into `'0'` and missing context into neutral direction. That produces valid-looking but false execution history. Please use field-specific structs (positive price/size/minimum, non-negative balances/magnitudes, signed PnL/funding) and require the trade fields guaranteed by Lighter's documented response. If live payloads legitimately omit a field, represent that as unavailable rather than manufacturing zero. Source: https://apidocs.lighter.xyz/reference/trades. This is the unresolved portion of the previous REST-boundary finding.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8497 @@
...
+      const generation = this.#sessionGeneration;
+      const accountIndex = await this.#ensureAccountIndex();
+      const token = await this.#getAuthToken();
+      await this.#ensureMarkets();
+      const response = await this.#clientService.getPositionFundings(
+        accountIndex,
+        token,
+      );
+      this.#assertSession(generation);
+
```
geositta: **[P1] Do not substitute a global maintenance fraction when venue metadata is unavailable.** This fallback feeds `calculateLiquidationPrice` a fabricated 1% MMR, while `getMaxLeverage` similarly returns 50x on missing or failed per-market metadata. Lighter publishes market-specific maintenance and minimum-initial fractions; for a 2.4% market, the fallback materially understates liquidation risk and exposes unsupported leverage in Pro UI. The trading validator already fails closed via `#requireMarketMaxLeverage`; please apply the same contract to display/calculation APIs and report unavailable instead of returning invented parameters. Source: https://apidocs.lighter.xyz/reference/orderbookdetails.

geositta, line 6656: **[P1] Preserve data-integrity failures in every financial-history API.** `LighterClientService` now labels malformed endpoint responses with `LIGHTER_DATA_INTEGRITY_PREFIX`, but this catch rethrows only capability errors and converts integrity failures to `[]`. The same pattern remains in `getUserNonFundingLedgerUpdates` and `getUserHistory`. A schema change or malformed amount then appears as a successful history with no funding, deposits, withdrawals, or transfers. Please rethrow integrity errors consistently (ideally through one shared predicate/helper) and add one malformed-response test per public history path.

geositta, line 7579: **[P1] Do not replace valid account state with empties after an integrity failure.** If account discovery or authenticated channel setup fails because a Lighter response is malformed, this path emits `EMPTY_ACCOUNT_STATE`, `[]` positions, and `[]` orders to active subscribers. That presents data corruption as a zero-balance account and can hide exposure. Please distinguish an explicit account switch/deselection (where the synchronous reset is correct) from transport/integrity failure (where the error/connection state must surface without asserting authoritative empties). The nested authenticated-order setup path needs the same treatment.

packages/perps-controller/src/providers/LighterProvider.ts
```diff
@@ -0,0 +1,8497 @@
...
+      const firstPage = await this.#clientService.getInactiveOrders(
+        accountIndex,
+        authToken,
+        100,
+        undefined,
+        marketId,
+      );
+      this.#assertSession(generation);
+      ingest(firstPage.orders);
+      if (missing() && !deepTraversalDone) {
```
geositta, line 3049: **[P2] Make the inactive-order traversal exhaustive or explicitly terminal when truncated.** This reader scans at most 1,000 rows, sets `deepTraversalDone`, and discards the remaining cursor even when target client IDs are still missing. Every later recovery starts from the first page and reaches the same cutoff, so a high-activity account can leave TP/SL settlement permanently unresolved with no actionable truncation signal. Please retain/resume the cursor until exhaustion, or throw a specific incomplete-history error that parks/surfaces manual recovery rather than treating the partial cache as authoritative. Add a target-beyond-page-10 test.

geositta (changes requested):
Thank you for the follow-up work. I verified that the maker-only minimum handling, cursor-based fill history, transactional REST/WebSocket order updates, and function-keyed signer bridge are now correctly addressed. The disconnect and cache fences added in the subsequent commits are also well covered.

I cannot approve this revision yet. The remaining issues share one correctness requirement: when authoritative Lighter financial data is missing, malformed, or incomplete, the provider must surface that state rather than substitute plausible values or empty histories. Specifically:

- Financial decimal schemas validate syntax but not field domains, and documented trade context can still be omitted and converted into zero/neutral history.
- Liquidation and leverage APIs fabricate 1%/50x risk parameters when market metadata cannot be established.
- Funding, ledger, and user-history reads convert integrity failures into empty histories.
- Account-channel setup converts integrity failures into empty account, position, and order state.
- TP/SL recovery repeatedly stops at the same 1,000 inactive orders without reporting that authoritative history was truncated.

These paths can produce materially incorrect Pro trading state while presenting a successful result, so they need to fail closed before merge. The exact-head package tests and all current CI checks pass; additional tests should exercise malformed/missing financial fields, metadata failure, integrity failures in every history/stream path, and recovery targets beyond the traversal bound. All comments are on code introduced by this PR relative to `origin/main`.

geositta (dismissed):

packages/perps-controller/src/services/LighterClientService.ts
abretonc7s: Fixed in 5e303106a1. REST financial fields now use positive, non-negative, or signed domain validators, and documented trade PnL, role, and position context are required instead of defaulted.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in 5e303106a1. Maintenance margin, position max leverage, liquidation previews, and getMaxLeverage now fail when authoritative per-market margin metadata is unavailable.

abretonc7s: Fixed in 5e303106a1. Funding, ledger, and user-history APIs now rethrow venue integrity failures, with malformed funding, deposit, withdrawal, and transfer coverage.

abretonc7s: Fixed in 5e303106a1. Account discovery and authenticated-channel failures preserve subscriber state; only explicit account switches or deselection emit empty resets.

packages/perps-controller/src/providers/LighterProvider.ts
abretonc7s: Fixed in 5e303106a1. Inactive-order recovery now walks beyond page ten until targets or exhaustion, and fails explicitly on a cursor cycle or the 10,000-row safety bound.

geositta (approved):

PR #9892, opened 2026-08-17
feat: start tracking on AccountTreeController:initialized

salimtb:
Subscribe to AccountTreeController:initialized so asset tracking starts only after the tree is fully built, instead of on intermediate stateChange events during init. Hosts must now also delegate the new event through the AssetsController messenger.

Extension PR: https://github.com/MetaMask/metamask-extension/pull/45579
Mobile PR: https://github.com/MetaMask/metamask-mobile/pull/34866

## Checklist

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 22007ec7074e55acb45569660bf6997d99aea165. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -3,6 +3,7 @@ import type {
   AccountTreeControllerInitializedEvent,
   AccountTreeControllerSelectedAccountGroupChangeEvent,
   AccountTreeControllerStateChangeEvent,
+  AccountTreeControllerUninitializedEvent,
 } from '@metamask/account-tree-controller';
 import type { AccountsControllerGetSelectedAccountAction } from '@metamask/accounts-controller';
 import { BaseController } from '@metamask/base-controller';
@@ -347,6 +348,7 @@ type AllowedEvents =
   | AccountTreeControllerSelectedAccountGroupChangeEvent
   | AccountTreeControllerStateChangeEvent
   | AccountTreeControllerInitializedEvent
+  | AccountTreeControllerUninitializedEvent
   | ClientControllerStateChangeEvent
   | KeyringControllerLockEvent
   | KeyringControllerUnlockEvent
@@ -747,6 +749,12 @@ export class AssetsController extends BaseController<
   /** Whether the keyring is unlocked. Combined with #uiOpen for #updateActive. */
   #keyringUnlocked = false;
 
+  /**
+   * Whether `AccountTreeController` has finished `init()`. Unlock / UI-open
+   * alone must not start fetches — the tree can still be mid-build.
+   */
+  #accountTreeInitialized = false;
+
   readonly #controllerMutex = new Mutex();
 
   /** Serializes account-switch fetch + subscribe to prevent overlapping races. */
@@ -1105,8 +1113,10 @@ export class AssetsController extends BaseController<
     // Subscribe to account group changes (when user switches between account groups like Account 1 -> Account 2)
     this.messenger.subscribe(
       'AccountTreeController:selectedAccountGroupChange',
-      (groupId) => {
-        this.#handleAccountGroupChanged(groupId).catch(console.error);
+      (groupId, previousGroupId) => {
+        this.#handleAccountGroupChanged(groupId, previousGroupId).catch(
+          console.error,
+        );
       },
     );
 
@@ -1199,7 +1209,12 @@ export class AssetsController extends BaseController<
     // happen before `AccountTreeController.init()`, and `:stateChange` fires
     // for intermediate mutations during that build.
     this.messenger.subscribe('AccountTreeController:initialized', () => {
-      this.#handleAccountTreeInitialized();
+      this.#accountTreeInitialized = true;
+      this.#updateActive();
+    });
+    this.messenger.subscribe('AccountTreeController:uninitialized', () => {
+      this.#accountTreeInitialized = false;
+      this.#updateActive();
     });
   }
 
@@ -1260,29 +1275,18 @@ export class AssetsController extends BaseController<
   }
 
   /**
-   * Start or stop asset tracking based on client (UI) open state and keyring
-   * unlock state. Only runs when both UI is open and keyring is unlocked.
+   * Start or stop asset tracking based on client (UI) open state, keyring
+   * unlock state, and account-tree readiness. Only runs when the UI is open,
+   * the keyring is unlocked, and the account tree has finished `init()`.
    */
   #updateActive(): void {
-    const shouldRun = this.#uiOpen && this.#keyringUnlocked;
+    const shouldRun =
+      this.#uiOpen && this.#keyringUnlocked && this.#accountTreeInitialized;
     if (shouldRun) {
       this.#start();
     } else {
       this.#stop();
     }
-  }
-
-  /**
-   * Handle `AccountTreeController:initialized`. Starts asset tracking once
-   * the tree is fully built. No-op unless the UI is open and the keyring is
-   * unlocked; `#start()` is idempotent if subscriptions are already active.
-   */
-  #handleAccountTreeInitialized(): void {
-    const shouldRun = this.#uiOpen && this.#keyringUnlocked;
-    if (!shouldRun) {
-      return;
-    }
-    this.#start();
   }
 
   /**
@@ -1293,8 +1297,12 @@ export class AssetsController extends BaseController<
    * `:initialized` instead, so we do not fetch on every tree mutation.
    */
   #handleAccountTreeStateChange(): void {
-    const shouldRun = this.#uiOpen && this.#keyringUnlocked;
-    if (!shouldRun || this.#activeSubscriptions.size === 0) {
+    const shouldRun =
+      this.#uiOpen &&
+      this.#keyringUnlocked &&
+      this.#accountTreeInitialized &&
+      this.#activeSubscriptions.size > 0;
+    if (!shouldRun) {
       return;
     }
     const accounts = this.#getSelectedAccounts();
@@ -1342,7 +1350,7 @@ export class AssetsController extends BaseController<
         chainIds: [...this.#enabledChains],
         forceUpdate: true,
       });
-      this.#subscribeAssets();
+      this.#subscribeAssets({ skipInitialFetch: true });
       if (newAccounts.length > 0) {
         await this.getAssets(newAccounts, {
           chainIds: [...this.#enabledChains],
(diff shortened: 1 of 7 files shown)


salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

salimtb:
@metamaskbot publish-preview

packages/assets-controller/src/AssetsController.ts
```diff
@@ -345,6 +347,8 @@ type AllowedEvents =
   // AssetsController
   | AccountTreeControllerSelectedAccountGroupChangeEvent
   | AccountTreeControllerStateChangeEvent
+  | AccountTreeControllerInitializedEvent
+  | AccountTreeControllerUninitializedEvent
```
salimtb, line 351: Hosts that filter messenger events must now also allow `:initialized` / `:uninitialized`. These are the tree-ready / torn-down signals. Startup used to ride on `:stateChange` during `AccountTreeController.init()`, which fired many times and spammed the Accounts API.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -745,6 +749,12 @@ export class AssetsController extends BaseController<
   /** Whether the keyring is unlocked. Combined with #uiOpen for #updateActive. */
   #keyringUnlocked = false;
 
+  /**
+   * Whether `AccountTreeController` has finished `init()`. Unlock / UI-open
+   * alone must not start fetches — the tree can still be mid-build.
+   */
+  #accountTreeInitialized = false;
```
salimtb, line 756: New flag. Unlock and UI-open can happen before `AccountTreeController.init()` finishes; without it `#updateActive()` could start fetches on a half-built tree. Tracking now runs only when UI is open **and** unlocked **and** the tree is initialized.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1103,18 +1113,17 @@ export class AssetsController extends BaseController<
     // Subscribe to account group changes (when user switches between account groups like Account 1 -> Account 2)
     this.messenger.subscribe(
       'AccountTreeController:selectedAccountGroupChange',
-      (groupId) => {
-        this.#handleAccountGroupChanged(groupId).catch(console.error);
+      (groupId, previousGroupId) => {
```
salimtb, line 1116: Thread `previousGroupId` through. ATC re-emits `selectedAccountGroupChange` on every `init()` even when the group didn't change, so we need the previous id to tell real user switches from init noise.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1196,6 +1205,17 @@ export class AssetsController extends BaseController<
         this.#onTransactionConfirmed(transactionMeta);
       },
     );
+    // Start tracking only after the account tree is fully built. Unlock can
+    // happen before `AccountTreeController.init()`, and `:stateChange` fires
+    // for intermediate mutations during that build.
+    this.messenger.subscribe('AccountTreeController:initialized', () => {
```
salimtb, line 1211: `:initialized` flips the flag and calls `#updateActive()` to start tracking once; `:uninitialized` clears it and stops tracking (wallet reset / `clearState`) so we don't keep polling stale accounts.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1255,11 +1275,13 @@ export class AssetsController extends BaseController<
...
-   * Start or stop asset tracking based on client (UI) open state and keyring
-   * unlock state. Only runs when both UI is open and keyring is unlocked.
+   * Start or stop asset tracking based on client (UI) open state, keyring
+   * unlock state, and account-tree readiness. Only runs when the UI is open,
+   * the keyring is unlocked, and the account tree has finished `init()`.
    */
   #updateActive(): void {
-    const shouldRun = this.#uiOpen && this.#keyringUnlocked;
+    const shouldRun =
+      this.#uiOpen && this.#keyringUnlocked && this.#accountTreeInitialized;
```
salimtb, line 1284: Single start/stop gate now also requires `#accountTreeInitialized`. Unlock alone is no longer enough — this prevents the unlock-before-init Accounts API burst.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1268,56 +1290,54 @@ export class AssetsController extends BaseController<
...
+   * `:stateChange` events during `init()` are ignored — startup is driven by
+   * `:initialized` instead, so we do not fetch on every tree mutation.
    */
   #handleAccountTreeStateChange(): void {
-    const shouldRun = this.#uiOpen && this.#keyringUnlocked;
+    const shouldRun =
+      this.#uiOpen &&
+      this.#keyringUnlocked &&
+      this.#accountTreeInitialized &&
+      this.#activeSubscriptions.size > 0;
```
salimtb, line 1304: `:stateChange` still fires for intermediate `init()` mutations, so we ignore them unless already tracking. After that we only refresh when the selected account set actually changes (e.g. a snap account added), not on every tree mutation.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1330,7 +1350,7 @@ export class AssetsController extends BaseController<
         chainIds: [...this.#enabledChains],
         forceUpdate: true,
       });
-      this.#subscribeAssets();
+      this.#subscribeAssets({ skipInitialFetch: true });
```
salimtb, line 1353: `#runAccountTreeRefresh` just did `getAssets(..., { forceUpdate: true })`, which already hit the Accounts API. `skipInitialFetch: true` drops AccountsApi's duplicate subscribe-time poll; the interval still polls on the normal cadence afterward.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1363,13 +1383,14 @@ export class AssetsController extends BaseController<
       // and default tracked assets that were never returned by balance APIs.
       this.#ensureNativeBalancesDefaultZero();
       this.#ensureDefaultTrackedAssetsSeeded();
-      this.#subscribeAssets();
+      // Balances were just force-fetched — skip AccountsApi's subscribe-time poll.
+      this.#subscribeAssets({ skipInitialFetch: true });
```
salimtb, line 1387: Same as the tree-refresh path: `#runStartupRefresh` already force-fetched, so skip AccountsApi's one-shot subscribe poll (interval still runs).

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3568,17 +3608,33 @@ export class AssetsController extends BaseController<
...
     // The selected account group can be empty during onboarding or wallet reset.
     if (!groupId) {
       return;
     }
 
+    // First start is owned by `AccountTreeController:initialized` / `#start`.
+    // ATC also re-publishes `selectedAccountGroupChange` on every `init()`
+    // (including when the group did not change) so late subscribers can catch
+    // up — ignore those until we are already tracking.
+    if (this.#activeSubscriptions.size === 0) {
```
salimtb, line 3624: Early returns. First start is owned by `:initialized` → `#start`, not by the init-time group event. Skipping empty-subscription and same-group (`groupId === previousGroupId`) events avoids a duplicate force-fetch around startup.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3595,7 +3651,8 @@ export class AssetsController extends BaseController<
       this.#ensureNativeBalancesDefaultZero();
       this.#ensureDefaultTrackedAssetsSeeded();
       // Subscribe after seed so the price poll sees natives / defaults.
-      this.#subscribeAssets();
+      // Balances were just force-fetched — skip AccountsApi's subscribe-time poll.
+      this.#subscribeAssets({ skipInitialFetch: true });
```
salimtb, line 3655: Group switch already force-fetched via `getAssets`, so skip AccountsApi's subscribe-time poll here too.

packages/assets-controller/src/data-sources/AbstractDataSource.ts
```diff
@@ -26,6 +26,12 @@ export type SubscriptionRequest = {
    * Provided by the controller when subscribing.
    */
   getAssetsState?: () => AssetsControllerStateInternal;
+  /**
+   * When true, skip the one-shot fetch that normally runs when a subscription
+   * is created. Used after the controller has already force-fetched balances
+   * for the same accounts/chains so we do not immediately hit the API again.
+   */
+  skipInitialFetch?: boolean;
```
salimtb, line 34: Optional subscribe flag so AssetsController can tell AccountsApi that balances for this scope are already fresh. Non-breaking — sources that don't read it behave as before.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -721,6 +721,11 @@ export class AccountsApiDataSource extends AbstractDataSource<
     return async (context, next) => {
       const { request } = context;
 
+      // Price/metadata-only requests must not hit the Accounts API.
+      if (!request.dataTypes.includes('balance')) {
```
salimtb, line 725: Skip the Accounts API balance middleware for price-only refreshes (currency change, missing-price backfill, etc.). Balances aren't needed there, so pass through to the next middleware without the HTTP call.

packages/assets-controller/src/data-sources/AccountsApiDataSource.ts
```diff
@@ -879,8 +885,12 @@ export class AccountsApiDataSource extends AbstractDataSource<
       onAssetsUpdate: subscriptionRequest.onAssetsUpdate,
     });
 
-    // Initial fetch
-    await pollFn();
+    // Interval above still polls on the normal cadence. This only skips the
+    // one-shot fetch at subscribe time when the controller already ran a
+    // force getAssets for the same scope (startup / group refresh).
+    if (!skipInitialFetch) {
```
salimtb, line 891: `setInterval` is still armed , polling is not disabled. We only skip the immediate one-shot fetch when the controller already force-fetched the same accounts/chains; the next tick still runs on `pollInterval`.

packages/assets-controller/src/__fixtures__/MockAssetControllerMessenger.ts
```diff
@@ -67,6 +67,8 @@ export function createMockAssetControllerMessenger(): {
       // AssetsController
       'AccountTreeController:selectedAccountGroupChange',
       'AccountTreeController:stateChange',
+      'AccountTreeController:initialized',
```
salimtb, line 70: Mirror the host messenger allowlist so the fixture can publish the new `:initialized` / `:uninitialized` lifecycle events.

juanmigdr (approved):

PR #9041, opened 2026-06-08
feat: add @metamask/network-connection-banner-controller

cryptodev-2s:
## Explanation

The RPC connection banner ("Still connecting" / "Unable to connect") is currently rendered by both `metamask-extension` and `metamask-mobile` based on duplicated logic in each repo:

- The 5s / 30s timer state machine.
- The show/hide rule (custom-RPC override, ≥2 distinct registrable domains failing, or all-enabled-failed escape hatch).
- eTLD+1 grouping via `psl` so a single provider's wide outage doesn't pop the banner.
- Infura URL detection.
- The `{ chainId, networkName, rpcUrl, isInfuraEndpoint, infuraNetworkClientId }` payload shape.

This PR introduces `@metamask/network-connection-banner-controller` so there's a single source of truth. The controller:

- Subscribes to `NetworkController:stateChanged`, `NetworkEnablementController:stateChanged`, and `ConnectivityController:stateChanged` via the messenger.
- Runs the 5s/30s escalation timers internally.
- Exposes a flat state `{ status: 'available' | 'degraded' | 'unavailable', network: FailedNetwork | null }` that clients subscribe to.
- Exposes two messenger actions: `dismissBanner()` and `switchToDefaultInfuraRpc({ chainId })` — the latter looks up the chain's first Infura endpoint and calls `NetworkController:updateNetwork` with `replacementSelectedRpcEndpointIndex`.
- Stays analytics-free; clients diff state-change events into their own analytics translation.

`psl` 1.15 ships its `types` field outside its `exports` map so TypeScript's `node16` module resolution can't find it; a small ambient `src/psl.d.ts` shim mirrors the workaround already in use in extension and mobile. `ip-regex` is pinned to `^4.3.0` (last CJS release) because the monorepo's Jest config doesn't transform ESM-only deps.

## References

- Jira: [WPC-1014](https://consensyssoftware.atlassian.net/browse/WPC-1014)
- Extension Testing PR https://github.com/MetaMask/metamask-extension/pull/43361

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by updating changelogs for packages I've changed
- [ ] I've introduced breaking changes in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

[...]

PR #9893, opened 2026-08-17
fix(network-connection-banner-controller): subscribe to ClientController:stateChange

cryptodev-2s:
## Explanation

`NetworkConnectionBannerController` subscribes to `ClientController:stateChanged` using an event type it defines locally. `@metamask/client-controller` only declares and exports `ClientController:stateChange`.

It works at runtime because `BaseController` publishes both names. This is a contract cleanup, not a behavior fix.

Now the controller uses the exported `ClientControllerStateChangeEvent`, dropping the duplicated local type and matching the three other upstream subscriptions in the file.

Breaking: clients must delegate `ClientController:stateChange` instead of `ClientController:stateChanged`, otherwise the controller stops receiving UI open state and the banner never shows. Both extension and mobile need the one line swap.

## References

- Introduced in #9041

## Checklist

- [x] I've introduced breaking changes in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit ec8b9d7d301e82bafb66b23a463bf63497ecf360. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/network-connection-banner-controller/src/NetworkConnectionBannerController.ts
+++ b/packages/network-connection-banner-controller/src/NetworkConnectionBannerController.ts
@@ -5,7 +5,7 @@ import type {
 } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
 import { clientControllerSelectors } from '@metamask/client-controller';
-import type { ClientControllerState } from '@metamask/client-controller';
+import type { ClientControllerStateChangeEvent } from '@metamask/client-controller';
 import {
   CONNECTIVITY_STATUSES,
   connectivityControllerSelectors,
@@ -265,16 +265,6 @@ export type NetworkConnectionBannerControllerEvents =
   NetworkConnectionBannerControllerStateChangedEvent;
 
 /**
- * Published when the state of `ClientController` changes. Defined here
- * because the `client-controller` package still exports the legacy
- * `:stateChange` event type.
- */
-type ClientControllerStateChangedEvent = ControllerStateChangedEvent<
-  'ClientController',
-  ClientControllerState
->;
-
-/**
  * Events from other messengers that
  * {@link NetworkConnectionBannerControllerMessenger} subscribes to.
  */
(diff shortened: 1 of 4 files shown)


cryptodev-2s:
@metamaskbot publish-preview

packages/network-connection-banner-controller/src/NetworkConnectionBannerController.ts
```diff
@@ -474,7 +464,8 @@ export class NetworkConnectionBannerController extends BaseController<
     // Lifecycle: evaluate RPC health (and run the banner escalation timers)
     // only while the client UI is open on an unlocked wallet.
     this.messenger.subscribe(
-      'ClientController:stateChanged',
+      // eslint-disable-next-line no-restricted-syntax -- awaiting upstream :stateChanged migration
+      'ClientController:stateChange',
```
FrederikBolding, line 468: Why is this necessary? Does `ClientController` not publish both?

cryptodev-2s: Because it's marked as deprecated.

cryptodev-2s: If remove it we get 
```
Subscribing to ':stateChange' events is deprecated. Use ':stateChanged' instead.eslint[no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax)

FrederikBolding: I wasn't referring to the eslint comment. I am curious why we are making this change in the first place and why we can't use `:stateChanged` in this controller? Is it not published by `ClientController`?

cryptodev-2s: No it's not defined by `ClientController` because it still uses `:stateChange` https://github.com/MetaMask/core/blob/main/packages/client-controller/src/ClientController.ts#L82

FrederikBolding (approved):

PR #9907, opened 2026-08-18
fix(perps-controller): map Terminal v1 category onto client marketType

geositta:
## What is the current state of things and why does it need to change?

Mobile's Pre-IPO chip filters on `marketType === "pre-ipo"`. Production Terminal v1 already sends Unitree, CXMT, and SKHY as `category: "pre_ipo"`, but v1 `#extractMetadata` only copied `marketType` / `categories` and dropped `category`. Those markets left `marketType` unset and did not appear under Pre-IPO.

Cerebras (`xyz:CBRS`) and SpaceX (`xyz:SPCX`) are public. Terminal already classifies them as `stocks`, but `HIP3_ASSET_MARKET_TYPES` still tags them `pre-ipo` from an older client override, so they stay on the Pre-IPO chip when Terminal does not set `marketType`.

## What is the solution your changes offer and how does it work?

`#categoryToMarketType` is the single Terminal → client alias table (`pre_ipo` → `pre-ipo`, `stocks` → `stock`, plus passthrough of values already in `MarketCategory`). v2 `#marketTypeFor` and v1 `#extractMetadata` both use it. Explicit v1 `marketType` still wins.

When Terminal is on, Unitree / CXMT / SKHY get `pre-ipo` from `category`, and CBRS / SPCX get `stock`. `HIP3_ASSET_MARKET_TYPES` moves CBRS and SPCX to `stock` so the HyperLiquid-direct fallback matches. `xyz:IPOP` (Quantinuum) stays Pre-IPO.

## Are there any changes whose purpose might not obvious to those unfamiliar with the domain?

This is not a typo in the mobile filter and not an old-app shim. Terminal and the client use different field names and spellings today (`category: "pre_ipo"` vs `marketType: "pre-ipo"`). The controller is the adapter; changing the chip to compare `"pre_ipo"` would still miss Unitree (`marketType` is unset) and would drop markets that already use `"pre-ipo"`.

We do not map only `pre_ipo` and leave `stocks` alone. That was protecting the stale CBRS/SPCX Pre-IPO override.

## If your primary goal was to update one package but you found you had to update another one along the way, why did you do so?

Only `@metamask/perps-controller` changed. No other package in this repo.

## If you had to upgrade a dependency, why did you do so?

No dependency upgrades.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Localized market classification and filter-bucket behavior in perps-controller; no auth, trading, or payment paths.
> 
> **Overview**
> Fixes Pre-IPO and stock category filters when Mobile uses Terminal v1 metadata: markets like Unitree were sent as `category: "pre_ipo"` but never got `marketType`, so they did not match `pre-ipo`.
> 
> **Terminal → client mapping:** Adds shared `#categoryToMarketType` (`pre_ipo` → `pre-ipo`, `stocks` → `stock`, plus valid `MarketCategory` passthrough). v1 `#extractMetadata` now reads `category` on the schema and uses this when `marketType` is missing; explicit `marketType` still wins. v2 `#marketTypeFor` delegates to the same helper.
> 
> **Enrichment parity:** When v1 Terminal metadata applies a `marketType`, `MarketDataService` clears `isNewMarket` so categorized HIP-3 markets are not also in the `new` filter (aligned with the v2 snapshot path).
> 
> **Static fallback:** Moves `xyz:CBRS` and `xyz:SPCX` from `pre-ipo` to `stock` in `HIP3_ASSET_MARKET_TYPES` when Terminal/HyperLiquid fallback is used; `xyz:IPOP` stays Pre-IPO.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d1b7ca56c2ffd6fa6f9505ba25d20a973121a6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/perps-controller/CHANGELOG.md
+++ b/packages/perps-controller/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Report the take profit (or stop loss) price on a `Position` when its only trigger for that direction is a partial, quantity-scoped one ([#9912](https://github.com/MetaMask/core/pull/9912))
   - `takeProfitPrice`/`stopLossPrice` were only ever scanned from position-bound triggers, so a position whose sole take profit closed it partially reported `takeProfitCount: 1` with no price, and clients rendering the scalar showed none. Applies to the REST `getPositions`, `getUserDataSnapshot`, and WebSocket position paths alike.
   - Two or more triggers in a direction still report the scanned price, because no single price describes them and clients render the count instead.
+- Map Terminal v1 `category` aliases (`pre_ipo` → `pre-ipo`, `stocks` → `stock`) when `marketType` is absent, so Pre-IPO markets such as Unitree appear under that filter without a static symbol list
+- Reclassify `xyz:CBRS` and `xyz:SPCX` from `pre-ipo` to `stock` in `HIP3_ASSET_MARKET_TYPES` now that they are public (Terminal already sends `stocks`). Leave `xyz:IPOP` as Pre-IPO
 
 ## [12.1.0]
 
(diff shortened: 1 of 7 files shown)


michalconsensys (dismissed):

abretonc7s (changes requested):
We should be using v2 now and deprecate the v1 endpoint.

abretonc7s:
Closing as obsolete. This PR restores deprecated Terminal v1 behavior. The correct local provider-fallback classification shipped through #9988 in @metamask/perps-controller 13.1.0, which Mobile main already consumes.

PR #9913, opened 2026-08-19
feat(platform-api-docs): add `root-messenger` strategy

cryptodev-2s:
## Explanation

`@metamask/platform-api-docs` documents the platform API — every messenger action and event a project exposes. Until now it had one way of finding them: parse every TypeScript source and declaration file it can reach (the scan directories, `packages/*/src`, and `node_modules/@metamask/*/dist/**/*.d.cts`) and walk every type alias named `*Messenger`.

That is the right approach for this monorepo, which has no single messenger aggregating every capability. It is a poor fit for a client, which already declares the complete set on its root messenger. Re-deriving that from the whole dependency tree means parsing ~11,600 files in `metamask-mobile` and ~4,500 in `metamask-extension`, to rediscover something the client has written down in one place.

This PR adds a second strategy that reads what the client already declares.

### `--strategy`

- **`scan`** (default) — unchanged behaviour, and the only option for a project with no single aggregating messenger.
- **`root-messenger`** — resolves the two types named by `--root-actions` and  `--root-events` (each written `<file>#<TypeName>`) and lets the TypeScript type  checker enumerate them. Only the named files are opened; the checker pulls in the rest.

Flags belonging to the strategy that wasn't selected are rejected rather than ignored, via a yargs `.check`, so a mistaken invocation fails loudly instead of quietly producing docs built the wrong way. The `<file>#<TypeName>` references are parsed in a yargs `.coerce`, so a malformed one is reported like any other
bad argument before work begins.

### Why the type checker rather than the AST

This is the non-obvious part. The two clients declare their root unions differently:

- `metamask-mobile` writes `GlobalActions` by hand as a union of type references.  A syntactic walk would work.
- `metamask-extension` derives `RootMessengerActions` from a registry of  messenger factories via
  `MessengerActions<ReturnType<(typeof MESSENGER_FACTORIES)[…]['getMessenger']>>`.
  **There is no syntactic union to walk** — only the type checker can say what it contains.

Going through the checker handles both shapes with one code path. Once it reports *which* capability types are in the union, each declaration is handed to the **existing** extractor in `extraction.ts`, so JSDoc, handler/payload signatures, source links, and deprecation flags come out identical to `scan`. The new module is a discovery front-end, not a second extractor.

Two details worth knowing:

- A capability declared as a **type alias** carries its name and JSDoc on the  *alias* symbol; the plain symbol points at the anonymous object type. An **interface** has no alias symbol, being its own declaration, so both are
  consulted. Missing the interface case silently dropped 61 actions and 8 events  on mobile before it was fixed.
- For a lone generic instantiation (`type Actions = Foo<Bar>`) the checker attributes the alias to the *root union itself*, which would hand the extractor the wrong declaration. That case is guarded.

### Failure behaviour

Generation now fails loudly instead of producing an empty site. `writeOutput` deletes `docs/` before writing, so a root union that resolves to nothing — a renamed type, or imports that don't resolve — would previously have replaced a
published docs directory with an empty one and exited `0`. It now throws, naming both references.

Capability types that can't be documented are reported with their names rather than counted, in three buckets: declared inline (no name or JSDoc), unresolved (`any`/`unknown`, usually a failed import), and unextractable (a shape the extractor rejects). A count alone isn't actionable at this scale.

### Also fixed: MDX escaping

`escapeJsDocTextForMdx` escaped `{` and `}` but not `<`, which MDX reads as the start of a JSX tag. A `@returns` comment such as `Promise<PointsBoostDto[]>` therefore **failed the site build** rather than rendering:

```
Unexpected character `[` (U+005B) in name, expected a name character…
```

This is pre-existing and independent of the new strategy — `scan` produces the byte-identical line — but it blocked `--build` and `--serve` for both clients, so it is fixed here. Affects description, `@param`, and `@returns` text; handler and payload signatures were already safe inside fenced code blocks.

---

## Benchmarks

Measured on a warm checkout, doc generation only.

| Project | `scan` | `root-messenger` | Speedup |
| --- | --- | --- | --- |
| `metamask-extension` | 47.9s | **4.7s** | ~10× |
| `metamask-mobile` | 96.9s | **5.7s** | ~17× |

`scan` parses ~5,600 `app/**/*.ts` plus ~6,026 `.d.cts` in mobile, and ~4,472 `.d.cts` in the extension. `root-messenger` opens the entry file and lets the checker pull in only what the union references.

### `metamask-mobile`

| | `scan` | `root-messenger` |
| --- | --- | --- |
| Namespaces | 112 | 103 |
| Actions | 1184 | 1157 |
| Events | 181 | 171 |
| Unique capabilities | 1365 | 1328 (97.3%) |

The 37 not documented are mostly controllers genuinely **not on the root messenger** — `PasskeyController` alone accounts for 17, plus `RatesController` (4), the sample controllers, and the decrypt/encrypt message managers. Nothing is found by `root-messenger` that `scan` misses.

### `metamask-extension`

| | `scan` | `root-messenger` |
| --- | --- | --- |
| Namespaces | 119 | 115 |
| Actions | 1209 | 1085 |
| Events | 183 | 176 |
| Unique capabilities | 1392 | 1261 (90.6%) |

99 of the 131-capability gap is `PerpsController`, and the cause is worth flagging to the extension team rather than treating as a tool limitation:

```ts
export type PerpsControllerMessenger = Messenger<
  'PerpsController',
  AllowedActions,   // actions Perps may CALL
  AllowedEvents
>;
```

`RootMessengerActions` is `MessengerActions<ChildMessengers>` — the union of what each child messenger is **allowed to call**, not what each controller **provides**. A controller whose actions are only invoked from the UI, never from another controller's messenger, never appears. `PerpsController` is registered in
`MESSENGER_FACTORIES` yet contributes **zero** constituents.

`root-messenger` documents exactly what the named types contain. Full coverage in. the extension needs an aggregate of *provided* actions, which is an extension-side change.

Conversely, `root-messenger` finds 6 capabilities `scan` misses(`MultichainRoutingService` ×4, `PPOMController` ×2).

Reported-but-skipped, current run: mobile 39 unextractable; extension 2 inline + 33 unextractable.

---

## Usage in clients

Once published, add the dependency and two scripts. For `metamask-mobile`:

```json
{
  "scripts": {
    "docs:platform-api:build": "platform-api-docs --build --project-label Mobile --strategy root-messenger --root-actions 'app/core/Engine/types.ts#GlobalActions' --root-events 'app/core/Engine/types.ts#GlobalEvents'",
    "docs:platform-api:serve": "platform-api-docs --serve --project-label Mobile --site-base-url / --strategy root-messenger --root-actions 'app/core/Engine/types.ts#GlobalActions' --root-events 'app/core/Engine/types.ts#GlobalEvents'"
  }
}
```

For `metamask-extension`, the label and references change:

```
--project-label Extension
--root-actions 'app/scripts/lib/messenger.ts#RootMessengerActions'
--root-events  'app/scripts/lib/messenger.ts#RootMessengerEvents'
```

Notes for consumers:

- Keep the `#` **inside quotes** — unquoted, most shells treat it as a comment and silently truncate the argument.
- `--root-actions` / `--root-events` are relative to the project path, not the shell's working directory.
- Output defaults to `<project-path>/.platform-api-docs`; gitignore it.
- `metamask-extension` additionally needs its `postcss-loader/jiti` resolution narrowed to `postcss-loader@^8.2.1/jiti`. The unversioned form also stubs the `postcss-loader@^7.3.4` that `@docusaurus/bundler` depends on, replacing `jiti` with an empty package and breaking the site build. Narrowing preserves the stub's original intent for the extension's own `postcss-loader@8.2.1`.

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WPC-1202
* Consumer PR (mobile): https://github.com/MetaMask/metamask-mobile/pull/26526
* Consumer PR (extension, includes the `jiti` resolution fix): https://github.com/MetaMask/metamask-extension/pull/40352

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the docs CLI and generation pipeline; default `scan` behavior is preserved with an explicit `strategy` field in tests.
> 
> **Overview**
> Adds a **`root-messenger`** discovery path alongside the existing **`scan`** default: the CLI accepts `--strategy`, `--root-actions`, and `--root-events` (`<file>#<TypeName>`), validates that strategy-specific flags are not mixed, and routes generation through type-checker resolution of the project’s root action/event unions instead of scanning the whole tree.
> 
> New **`root-messenger-discovery`** walks those unions (including checker-derived unions like `MessengerActions<…>`), reuses the shared extractor in **`extraction.ts`** (with **`classifyMessengerCapabilityTypeDeclaration`** exported for reuse), warns on skipped inline or unextractable capabilities, and **throws** when unions resolve to `any`/`unknown` or would produce zero docs so an empty site cannot overwrite published output.
> 
> **`escapeJsDocTextForMdx`** now escapes `<` as well as braces so generic types in JSDoc do not break MDX builds. README, changelog, and broad CLI/generate/discovery tests cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1da2bbd6e90b2a35cd3a6be63a419eb258346870. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryptodev-2s:
@metamaskbot publish-preview

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,301 @@
...
+  for (const constituent of constituents) {
+    // `any`/`unknown` means the checker couldn't resolve the type at all,
+    // usually a failed import. Worth separating from a genuinely anonymous
+    // capability, because the fix is entirely different.
+    if (constituent.isAny() || constituent.isUnknown()) {
+      skipped.unresolved.push(
+        `${rootDeclaration.getName()} (resolved to \`${constituent.getText()}\`)`,
+      );
+      continue;
+    }
```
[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/9913/commits/5fe1f30c7ef629f62c126c92c4ee71d50d25ba69

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,301 @@
...
+  const declarations =
+    aliasDeclarations.length > 0
+      ? aliasDeclarations
+      : (constituent.getSymbol()?.getDeclarations() ?? []);
+
+  return declarations.find(
+    (node): node is TypeAliasDeclaration | InterfaceDeclaration =>
+      NodeGuards.isTypeAliasDeclaration(node) ||
+      NodeGuards.isInterfaceDeclaration(node),
+  );
```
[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/9913/commits/5fe1f30c7ef629f62c126c92c4ee71d50d25ba69

cryptodev-2s:
@metamaskbot publish-preview

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,344 @@
...
+  if (found) {
+    return found;
+  }
+
+  // Nothing named behind the type itself. When the root aliases a single
+  // generic instantiation, the declaration we want is the one its type node
+  // references — `Foo` in `type Actions = Foo<Bar>`.
+  return findDeclarationFromRootTypeNode(rootDeclaration);
```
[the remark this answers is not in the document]
cryptodev-2s: fixed here https://github.com/MetaMask/core/pull/9913/commits/11c0860f91e6453d711f897623cf5aa556cffa47

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

packages/platform-api-docs/src/generate.ts
```diff
@@ -128,6 +130,33 @@ async function resolveRepoBaseUrl(
   return `${repoUrl}/blob/${ref}/`;
 }
 
+/**
+ * Options for the `scan` strategy, which reads every `*Messenger` type alias
+ * in every file it can find. Used when no single messenger aggregates every
+ * capability.
+ */
+type ScanStrategyOptions = {
```
mcmire: What do you think about putting these types at the top of the file, before functions, following conventions for other files?

packages/platform-api-docs/src/generate.ts
```diff
@@ -128,6 +130,33 @@ async function resolveRepoBaseUrl(
...
+  strategy?: 'scan';
+  /** Directories (relative to projectPath) to scan for .ts source files. */
+  scanDirs: string[];
+  rootActions?: never;
+  rootEvents?: never;
```
mcmire: Do we need to mark these as `never`? Can we omit them entirely?

``` suggestion
```

packages/platform-api-docs/src/generate.ts
```diff
@@ -128,6 +130,33 @@ async function resolveRepoBaseUrl(
...
+};
+
+/**
+ * Options for the `root-messenger` strategy, which resolves the unions a
+ * project declares for its root messenger instead of scanning. Used when one
+ * messenger carries every action and event.
+ */
+type RootMessengerStrategyOptions = {
+  strategy: 'root-messenger';
+  scanDirs?: never;
```
mcmire: Same as above.

``` suggestion
```

packages/platform-api-docs/src/cli.ts
```diff
@@ -119,6 +121,53 @@ async function resolveCommitSha(projectPath: string): Promise<string | null> {
   }
 }
 
+/** The subset of parsed arguments {@link checkStrategyArgs} validates. */
+type StrategyArgs = {
```
mcmire: What do you think about putting these types at the top of the file, before functions, following conventions for other files?

Alternatively what do you think about inlining this into `checkStrategyArgs`?

packages/platform-api-docs/src/cli.ts
```diff
@@ -179,6 +245,9 @@ async function main(): Promise<void> {
       description:
         'Path prefix the built site will be served under, e.g. /core/platform-api/',
     })
+    .coerce('root-actions', parseRootTypeReference)
```
mcmire: Nit: I think `.option()` above takes a `coerce` option, so is it possible to move this up? e.g.

``` typescript
    .option('root-actions', {
      type: 'string',
      description:
        'Type aliasing the union of every action on the root messenger, written as "<file>#<TypeName>" (required with --strategy root-messenger)',
      coerce: parseRootTypeReference,
    })
    .option('root-events', {
      type: 'string',
      description:
        'Type aliasing the union of every event on the root messenger, written as "<file>#<TypeName>" (required with --strategy root-messenger)',
      coerce: parseRootTypeReference,
    })

File: packages/platform-api-docs/src/cli.ts [hunk 9]
```diff
@@ -203,9 +272,15 @@ async function main(): Promise<void> {
...
     projectPath: resolvedProjectPath,
     outputDir: resolvedOutputDir,
-    scanDirs,
     projectLabel,
     commitSha,
+    ...(argv.strategy === 'root-messenger'
+      ? {
+          strategy: 'root-messenger' as const,
+          rootActions: argv['root-actions'] as RootTypeReference,
+          rootEvents: argv['root-events'] as RootTypeReference,
```
mcmire: If you specify an option above which contains dashes, Yargs *should* give you `camelCase` aliases in type.

To go further, I see that we have to typecast these options because Yargs doesn't know that we've validated them. I wonder if using `.check` is limiting us, and instead we ought to implement our own logic after we parse args? Typically the way I've handled this in other packages is to extract the `yargs(...)` call + any validation needed to a function called `parseArguments`, which doesn't return the same type that Yargs returns but rather something that is more useful. Using type guards you could then get TypeScript to understand that if `strategy` is "root-messenger" then `rootActions` and `rootEvents` are `RootTypeReference`.

Then you could say this:

``` suggestion
          rootActions: argv.rootActions,
          rootEvents: argv.rootEvents,
```

packages/platform-api-docs/src/root-messenger-discovery.ts
mcmire: Since we now have a `root-messenger-discovery.ts` file should we rename the existing `discovery.ts` file to `scan-discovery.ts`?

packages/platform-api-docs/src/generate.ts
mcmire: Do we need to mark `strategy` as optional and the rest as `never`?

``` suggestion
  strategy: 'scan';
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,359 @@
...
+
+/**
+ * Create a ts-morph Project for resolving root messenger types.
+ *
+ * No file list is loaded: this strategy opens only the entry files and lets
+ * the checker pull in the rest.
+ *
+ * @returns A new ts-morph Project.
+ */
+function createRootMessengerProject(): TsMorphProject {
```
mcmire, line 91: In the previous implementation, we had code for finding files in `discovery.ts` and we had code for scanning in `generate.ts`, and code for creating a TypeScript project in `extraction.ts`.

But now some of that logic is combined into `root-messenger-discovery.ts`, including this line. So it seems like we have code that's spread out.

It would make this package more readable if we could split out the concerns better. I like the direction you've gone with `root-messenger-discovery.ts`, and it would be nice if a similar pattern were followed for the existing code. In other words, ideally, `generate.ts` shouldn't know about the exact strategy being used; its goal should be to generate data that can be used to create pages in the documentation site. I think each strategy should be more self-contained; there should be a file or a directory (whatever is more appropriate) that holds all the code for that strategy. 

What do you think?

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,359 @@
...
+ * Resolve the type alias a reference names.
+ *
+ * @param project - The ts-morph project to load the file into.
+ * @param projectPath - Absolute path to the project root.
+ * @param reference - The reference to resolve.
+ * @param flagName - The CLI flag the reference came from, used in errors.
+ * @returns The type alias declaration.
+ * @throws If the file can't be read or declares no such type alias.
+ */
+function resolveRootDeclaration(
```
mcmire: "Root" seems ambiguous here. What do you think about being more specific?

``` suggestion
function findRootActionsOrEventsTypeDeclaration(
```

or

``` suggestion
function findRootCapabilitiesUnionTypeDeclaration(
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,359 @@
...
+      skipLibCheck: true,
+      target: ts.ScriptTarget.ESNext,
+      module: ts.ModuleKind.ESNext,
+      moduleResolution: ts.ModuleResolutionKind.NodeJs,
+    },
+  });
+}
+
+/**
+ * Resolve the type alias a reference names.
```
mcmire: Seems like we could be a little less abstract in the description of this function. Maybe:

``` suggestion
 * Locate the type alias for a root actions or events union in a source file
 * within the TypeScript project.
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,359 @@
...
+  reference: RootTypeReference,
+  flagName: string,
+): TypeAliasDeclaration {
+  const absolutePath = path.resolve(projectPath, reference.filePath);
+
+  let sourceFile;
+  try {
+    sourceFile =
+      project.getSourceFile(absolutePath) ??
+      project.addSourceFileAtPath(absolutePath);
```
mcmire, line 138: It seems like we ought to have already added the source files up front, but I guess this doesn't hurt.

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,359 @@
...
+ * @param rootDeclaration - The root union's declaration.
+ * @param kind - Whether these are actions or events.
+ * @param projectPath - Absolute path to the project root.
+ * @param skipped - Labels collected as undocumentable constituents are found.
+ * @param reference - The reference that named this type, used in errors.
+ * @param flagName - The CLI flag the reference came from, used in errors.
+ * @returns The extracted capabilities.
+ * @throws If the union resolved to `any` or `unknown`.
+ */
+function extractFromRootType(
```
mcmire: Similar as above. What root type are we talking about? I wonder if being more verbose here would help readability.

``` suggestion
function extractFromRootMessengerCapabilitiesUnion(

File: packages/platform-api-docs/src/root-messenger-discovery.ts [hunk 15]
```diff
@@ -0,0 +1,359 @@
...
+  rootDeclaration: TypeAliasDeclaration,
+  kind: 'action' | 'event',
+  projectPath: string,
+  skipped: SkippedCapabilities,
```
mcmire: Nit: In my experience, AI has a habit of abbreviating variables. This sounds like it's a boolean, thoughts on renaming to `skippedCapabilities` to match the type?

``` suggestion
  skippedCapabilities: SkippedCapabilities,

File: packages/platform-api-docs/src/root-messenger-discovery.ts [hunk 16]
```diff
@@ -0,0 +1,359 @@
...
+ * @param kind - Whether these are actions or events.
+ * @param projectPath - Absolute path to the project root.
+ * @param skipped - Labels collected as undocumentable constituents are found.
+ * @param reference - The reference that named this type, used in errors.
+ * @param flagName - The CLI flag the reference came from, used in errors.
+ * @returns The extracted capabilities.
+ * @throws If the union resolved to `any` or `unknown`.
+ */
+function extractFromRootType(
+  rootDeclaration: TypeAliasDeclaration,
```
mcmire: Thoughts on renaming this?

``` suggestion
  rootMessengerCapabilitiesUnion: TypeAliasDeclaration,

File: packages/platform-api-docs/src/root-messenger-discovery.ts [hunk 17]
```diff
@@ -0,0 +1,359 @@
...
+  enclosingNode: TypeAliasDeclaration,
+): string {
+  const text = type.getText(enclosingNode).replace(/\s+/gu, ' ');
+  return text.length > 80 ? `${text.slice(0, 77)}...` : text;
+}
+
+/**
+ * Extract every documentable capability from one root union.
+ *
+ * @param rootDeclaration - The root union's declaration.
```
mcmire: Similar as above.

``` suggestion
 * Extract every documentable capability from one root messenger actions or events union type.
 *
 * @param rootMessengerCapabilitiesUnion - The root messenger actions or events union type.

File: packages/platform-api-docs/src/root-messenger-discovery.ts [hunk 18]
```diff
@@ -0,0 +1,359 @@
...
+ */
+function extractFromRootType(
+  rootDeclaration: TypeAliasDeclaration,
+  kind: 'action' | 'event',
+  projectPath: string,
+  skipped: SkippedCapabilities,
+  reference: RootTypeReference,
+  flagName: string,
+): MessengerCapabilityPacket[] {
+  const rootType = rootDeclaration.getTypeNodeOrThrow().getType();
```
mcmire: Similar as above, what about `rootMessengerCapabilitiesUnion`?

``` suggestion
  const rootMessengerCapabilitiesUnion = rootDeclaration.getTypeNodeOrThrow().getType();

File: packages/platform-api-docs/src/root-messenger-discovery.ts [hunk 19]
```diff
@@ -0,0 +1,359 @@
...
+  const packets: MessengerCapabilityPacket[] = [];
+
+  for (const constituent of constituents) {
+    const declaration = findCapabilityDeclaration(
+      constituent,
+      rootDeclaration,
+      constituents.length === 1,
+    );
+    if (!declaration) {
+      skipped.unnamed.push(summarizeType(constituent, rootDeclaration));
```
mcmire: It looks like this function takes an argument (`skipped`) and then mutates it. I guess we do this as a way of making `discoverFromRootMessenger` more compact, but looking at this function alone it would be less surprising in my opinion if this function returned capability packets + skipped capabilities, and if `discoverFromRootMessenger` merged a "child" `skipped` into the composite `skipped`.

mcmire (commented):
I went through the changes here and made a bunch of comments. I've collected suggestions for most of them in this PR: https://github.com/MetaMask/core/pull/9981. Most of the changes are to `root-messenger-discovery.ts`, to make it easier to understand. I tried to break things up into commits so if you want to cherry-pick anything you can.

The only thing I didn't do was to consolidate the existing "scan strategy" code into one file just like we've done with the "root messenger strategy" code. I figured that was a larger refactor.

I also haven't manually tested this flow yet. I will do that shortly.

packages/platform-api-docs/src/root-messenger-discovery.ts
mcmire: I realize that this is a larger refactor so we can always do this later, but I wanted to point it out as a way of making this package more self-consistent and easier to browse if you're new to how it works.

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Done here https://github.com/MetaMask/core/pull/9913/commits/a846a890c212e359b4288b6a24fade0398f45952

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Done here https://github.com/MetaMask/core/pull/9913/commits/a846a890c212e359b4288b6a24fade0398f45952

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: I can look up this part of the follow up refactors

packages/platform-api-docs/src/generate.ts
```diff
@@ -15,8 +15,78 @@ import {
...
+  /** The selected strategy. */
+  strategy: 'scan';
+  /** Directories, relative to the project root, to scan. */
+  scanDirs: string[];
+};
+
+/**
+ * Options for the `root-messenger` strategy, which resolves the unions a
+ * project declares for its root messenger instead of scanning. Used when one
+ * messenger carries every action and event.
```
mcmire: "Union" isn't totally accurate here from an implementation perspective:

``` suggestion
 * Options for the `root-messenger` strategy, which walks the types a
 * project declares for its root messenger capabilities instead of scanning the
 * entire repo. Used when one messenger carries every action and event.
```

packages/platform-api-docs/src/generate.ts
```diff
@@ -535,7 +562,102 @@ export async function generate(
 
   logScanPlan(sources);
 
-  const allItems = await scanSources(projectPath, sources);
+  return await scanSources(projectPath, sources);
+}
+
+/**
+ * Collect capabilities by resolving the project's root messenger unions.
```
mcmire:
``` suggestion
 * Using the project's root messenger capability collection types as
 * entrypoints, collect the constituent individual capability types and package
 * them so that they can be displayed within the documentation site.
```

cryptodev-2s:
@metamaskbot publish-preview

packages/platform-api-docs/src/generate.ts
```diff
@@ -535,7 +562,102 @@ export async function generate(
...
+  );
+  warnSkipped(
+    'whose shape could not be read',
+    skippedCapabilities.unextractableCapabilities,
+  );
+
+  // Both unions resolving to nothing is always a misconfiguration — a wrong
+  // type name, or imports that didn't resolve. Failing here matters because
+  // generation would otherwise replace an existing docs directory with an
+  // empty one and exit successfully.
```
mcmire: The actions/events types may not truly be unions:

``` suggestion
  // If both capability collection types resolve to nothing, it's always a
  // misconfiguration (a wrong type name, or imports that didn't resolve).
  // Failing here matters because generation would otherwise replace an existing
  // docs directory with an empty one and exit successfully.
```

packages/platform-api-docs/src/generate.ts
```diff
@@ -535,7 +562,102 @@ export async function generate(
...
+        `Check that these types name the unions carrying every capability, ` +
+        `and that their imports resolve.`,
+    );
+  }
+
+  return capabilityPackets;
+}
+
+/** How many skipped capability types to name before summarizing the rest. */
+const MAX_SKIPPED_SHOWN = 10;
```
mcmire: What are your thoughts on moving this constant to the top of the file to follow conventions?

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+} from './extraction.js';
+import type { MessengerCapabilityPacket } from './types.js';
+
+// ---------------------------------------------------------------------------
+// The `root-messenger` strategy: resolve the unions a project declares for its
+// root messenger and let the type checker enumerate them. Going through the
+// checker rather than the AST means a union works whether it is written by
+// hand or computed (e.g. derived from a registry via `ReturnType<...>`). Each
+// capability it reports is handed to the shared extractor in `extraction.ts`,
+// so the output matches what the `scan` strategy produces.
```
mcmire: This isn't as accurate as it could be:

``` suggestion
// The `root-messenger` strategy: resolve the types a project declares for its
// collection of root messenger actions and events and let TypeScript walk them.
// Each capability type found is handed to the shared extractor in
// `extraction.ts`, so the output matches what the `scan` strategy produces.
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+};
+
+/**
+ * Labels for capability types that were found but couldn't be documented,
+ * grouped by why. Labels rather than counts, so warnings can name what to fix.
+ */
+type SkippedCapabilities = {
+  /**
+   * Capabilities declared inline in the union, so there is no name or JSDoc to
+   * document.
```
mcmire: This isn't as clear as it could be:

``` suggestion
   * Capabilities declared inline in the capability collection type, so there is
   * no name or JSDoc to document.
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+  skippedCapabilities: SkippedCapabilities;
+} {
+  const capabilityCollectionType = capabilityCollectionTypeDeclaration
+    .getTypeNodeOrThrow()
+    .getType();
+
+  // TypeScript absorbs `any | T` into `any` and `unknown | T` into `unknown`,
+  // so a single member the checker can't resolve — typically a failed import —
+  // erases every other capability in the union. Fail instead of emitting a
+  // catalog that looks complete but silently isn't.
```
mcmire: This is fine, but "union" is a bit inaccurate, and it's not as understandable as it could be:

``` suggestion
  // If one or more of the constituent capabilities in the collection is `any`
  // or `unknown` — e.g. its import failed — then the type of the whole
  // collection will also be `any` or `unknown`. Fail instead of emitting a
  // catalog that looks complete but silently isn't.
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+  if (
+    capabilityCollectionType.isAny() ||
+    capabilityCollectionType.isUnknown()
+  ) {
+    throw new Error(
+      `${capabilityCollectionTypeReference.filePath}#${capabilityCollectionTypeReference.typeName}, named by ${commandLineOptionName}, ` +
+        `resolved to \`${capabilityCollectionType.getText()}\` rather than a union of ` +
+        `capabilities. This usually means an import in that file could not be ` +
+        `resolved; because TypeScript absorbs the rest of a union into ` +
+        `\`any\`, every other capability in it would be missing.`,
```
mcmire: This is a bit more detailed than it should be, and it's lacking next steps:

``` suggestion
      `${capabilityCollectionTypeReference.filePath}#${capabilityCollectionTypeReference.typeName}, named by ${commandLineOptionName}, ` +
        `resolved to \`${capabilityCollectionType.getText()}\`. ` +
        `It's likely that an individual action or event type is also \`${capabilityCollectionType.getText()}\`, ` +
        `which may be due to a failed import. ` +
        `You will need to fix this first before generating docs for this project.`
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+
+  return { capabilityPackets, skippedCapabilities };
+}
+
+/**
+ * Resolves `<file>#<TypeName>` references to messenger actions and events union
+ * types within the given project (e.g. `RootMessengerActions` or
+ * `RootMessengerEvents`), walks the union to collect all of the containing
+ * capability types (e.g. `NetworkControllerAddNetworkAction`), then packages
+ * them so that they can be displayed within the documentation site.
```
mcmire: "Union" isn't completely accurate:

``` suggestion
 * Resolves `<file>#<TypeName>` references to messenger actions and events
 * collection types within the given project (e.g. `RootMessengerActions` or
 * `RootMessengerEvents`), walks the collection to gather all of the containing
 * capability types (e.g. `NetworkControllerAddNetworkAction`), then packages
 * them so that they can be displayed within the documentation site.
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -0,0 +1,527 @@
...
+ *
+ * @param args - The arguments to this function.
+ * @param args.projectPath -Absolute path to the project to scan.
+ * @param args.rootActionsTypeReference - A reference to a messenger
+ * actions union type within the project, in `<file>#<TypeName>` format.
+ * @param args.rootEventsTypeReference - A reference to a messenger events
+ * union type within the project, in `<file>#<TypeName>` format.
```
mcmire:
``` suggestion
 * @param args.rootActionsTypeReference - A reference to a messenger
 * actions collection type within the project, in `<file>#<TypeName>` format.
 * @param args.rootEventsTypeReference - A reference to a messenger events
 * collection type within the project, in `<file>#<TypeName>` format.
```

packages/platform-api-docs/README.md
```diff
@@ -35,12 +35,61 @@ Options:
...
+  --root-events <ref>      Type aliasing the union of every event, as "<file>#<TypeName>"
+                           (required with --strategy root-messenger)
   --output <dir>           Output directory (default: <project-path>/.platform-api-docs)
   --project-label <label>  Short label identifying the project (e.g. "Core", "Extension")
   --help                   Show this help message
 ```
 
+## Strategies
+
+Which strategy to use depends on whether the project has a single messenger
```
mcmire: This text is wrapped but the text at the top of the file is not wrapped. Is it possible to make this consistent? (I believe that we generally do *not* wrap text in Markdown files.)

mcmire (commented):
Some more suggestions. I forgot that in my PR I made an effort to represent `--root-actions` and `--root-events` as references to capability *collection* types rather than unions (because technically neither may be a true union). So most of these changes are to make sure the JSDoc captures that point of view.

I have some more changes to make to the README but I left a comment there about reformatting first.

packages/platform-api-docs/README.md
cryptodev-2s: Good catch my bad I usually pay attention to this as it happens so often, but I got it missed this time

mcmire: Yeah, maybe you can add a rule for your agent to ask it to check the existing line wrapping convention for the file and honor that when making changes?

cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Applied here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Moved here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/9913/commits/1da2bbd6e90b2a35cd3a6be63a419eb258346870

mcmire:
@cryptodev-2s I found https://github.com/MetaMask/metamask-extension/pull/40352 and https://github.com/MetaMask/metamask-mobile/pull/26526 and tried them out.

On Extension this is what I'm seeing:

```
Warning: skipped 2 capability types declared inline, with no name to document: { type: "PreferencesController:getState"; handler: () => { isMultiAccountBala..., { type: "RewardsController:stateChange"; payload: [import("../controllers/rew...
Warning: skipped 33 capability types whose shape could not be read: AuthenticatedUserStorageInvalidateQueriesAction (node_modules/@metamask/authenticated-user-storage/dist/authenticated-user-storage.d.cts:23), MoneyAccountApiDataServiceInvalidateQueriesAction (node_modules/@metamask/money-account-api-data-service/dist/money-account-api-data-service.d.cts:30), MoneyAccountBalanceServiceInvalidateQueriesAction (node_modules/@metamask/money-account-balance-service/dist/money-account-balance-service.d.cts:33), SentinelApiServiceInvalidateQueriesAction (node_modules/@metamask/sentinel-api-service/dist/types.d.cts:11), ControllerGetStateAction (node_modules/@metamask/base-controller/dist/BaseController.d.cts:112), ClaimsServiceInvalidateQueriesAction (node_modules/@metamask/claims-controller/dist/ClaimsService.d.cts:15), ShieldApiServiceInvalidateQueriesAction (node_modules/@metamask/shield-controller/dist/shield-api-service.d.cts:57), SubscriptionServiceInvalidateQueriesAction (node_modules/@metamask/subscription-controller/dist/SubscriptionService.d.cts:12), AuthenticatedUserStorageCacheUpdatedEvent (node_modules/@metamask/authenticated-user-storage/dist/authenticated-user-storage.d.cts:45), AuthenticatedUserStorageGranularCacheUpdatedEvent (node_modules/@metamask/authenticated-user-storage/dist/authenticated-user-storage.d.cts:50), and 23 more
```

On Mobile this is what I'm seeing:

```
Warning: skipped 39 capability types whose shape could not be read: MoneyAccountBalanceServiceInvalidateQueriesAction (node_modules/@metamask/money-account-balance-service/dist/money-account-balance-service.d.cts:33), MoneyAccountApiDataServiceInvalidateQueriesAction (node_modules/@metamask/money-account-api-data-service/dist/money-account-api-data-service.d.cts:30), DataServiceInvalidateQueriesAction (node_modules/@metamask/base-data-service/dist/BaseDataService.d.cts:16), SubscriptionServiceInvalidateQueriesAction (node_modules/@metamask/subscription-controller/dist/SubscriptionService.d.cts:12), ShieldApiServiceInvalidateQueriesAction (node_modules/@metamask/shield-controller/dist/shield-api-service.d.cts:57), ClaimsServiceInvalidateQueriesAction (node_modules/@metamask/claims-controller/dist/ClaimsService.d.cts:15), DataServiceInvalidateQueriesAction (node_modules/@metamask/base-data-service/dist/BaseDataService.d.cts:16), AuthenticatedUserStorageInvalidateQueriesAction (node_modules/@metamask/authenticated-user-storage/dist/authenticated-user-storage.d.cts:23), ChompApiServiceInvalidateQueriesAction (node_modules/@metamask/chomp-api-service/dist/chomp-api-service.d.cts:17), SentinelApiServiceInvalidateQueriesAction (node_modules/@metamask/sentinel-api-service/dist/types.d.cts:11), and 29 more
```

I wonder if it may be worth it eventually to throw errors instead of printing warnings if we encounter issues. That will motivate us to correct them. Or maybe there's a lint rule or some other check we can add to prevent these problems from creeping into the codebase altogether that would be nice too...

Still, this worked great and I can see that there are clear differences between Extension and Mobile (for instance, Extension includes `AccountOrderController` but Mobile does not, and Mobile includes `CardController` but Extension does not). And the `platform-api-docs` command ran super fast, which is great. The one for `core` still runs slow, but it sounds like you are working on that in your next PR.

At this point ~I think all of the remaining comments I have are super minor~ everything looks good. Nice work 👏🏻

mcmire (approved):
LGTM!

PR #6388, opened 2025-08-26
Fix: phishing controller c2 optimization

mindofmar:
## Explanation

`isMaliciousC2Domain` previously checked each request against the C2 blocklist using `Array.includes`, which performs a linear O(n) scan of hashed domain strings on every call. For a list in the thousands-of-entries range, each call scans the full list up to 6 times — once for the exact hostname hash, up to 5 times for parent domain hashes.

This PR switches the internal representation to `Set<string>`, reducing each of those scans to an O(1) hash lookup.

**Benchmark results (internal profiling):**
- Per-request time: 2–8ms → 0.05–0.5ms (~10–100x improvement depending on page complexity; ~50x averaged across pages)
- Potential CPU savings: up to ~300ms per page load
- Background tabs also benefit — confirmed they generate significant background network requests that all route through this check

**Implementation notes:**
- The conversion from `string[]` → `Set<string>` happens once at construction time via a new unexported `InternalPhishingDetectorConfiguration` type. The public `PhishingDetectorConfiguration` type keeps `c2DomainBlocklist?: string[]` unchanged — no API break for downstream consumers.
- `isMaliciousC2Domain` now uses `.size` / `.has` in place of `.length` / `.includes`
- `getDefaultPhishingDetectorConfig` accepts and threads `c2DomainBlocklist` through as `string[]`; `processConfigs` no longer performs a redundant intermediate `Set` construction

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5611

## References

- Fixes https://github.com/MetaMask/MetaMask-planning/issues/5611

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary
- [x] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving internal data-structure change on a hot path; public types and matching logic are unchanged.
> 
> **Overview**
> **C2 domain blocklist checks** in `PhishingDetector.isMaliciousC2Domain` now use a `Set` internally instead of scanning a `string[]` with `includes`, so each hostname and parent-domain hash lookup is O(1) rather than O(n) (up to several lookups per request).
> 
[...]

PR #9914, opened 2026-08-20
refactor: migrate PhishingController data fetching to PhishingDataService

0xOhm:
## Description

Extracts all remote data fetching from `PhishingController` into a new `PhishingDataService` built on the `BaseDataService` pattern (PSAFE-593). `CacheManager` is removed in favor of the service's built-in caching, persistence, and request coalescing.

### Changes

- **New `PhishingDataService`** owning all four remote APIs (stalelist/hotlist, request-blocklist/C2, dapp scanning, token/address scanning) with generated method action types
- **`PhishingController`** slimmed to detection/state logic; queries data through the service via the messenger
- **`CacheManager` removed** — caching, persistence (`StorageService`), and request coalescing now come from `BaseDataService`
- Response validation intentionally kept loose (e.g. token scan `result_type` as `string`, not enum) to match pre-refactor behavior — production already returns values outside the documented enum (`Verified`)

### Verification

- 501 unit tests / 7 suites, 100% stmt/fn/line coverage, 99.29% branch
- Full monorepo build clean against the new API
- Live production API contract: 9/9 endpoints accepted by validation
- Outage simulation: detection fails open to stale lists (never off), bounded retries
- Persistence lifecycle: rehydration after simulated MV3 SW restart with zero network; `maxAge`-expired verdicts discarded
- Coalescing parity: 120-token chunking, miss-only fetch, zero-fetch cached batches

### Client integration PRs

- Extension and mobile integration PRs to follow (wiring + state migrations); they depend on this merging and publishing first.

---

> [!NOTE]
> **High Risk**
> Large breaking refactor of security-sensitive phishing and scan networking, new required wiring and state migrations, plus changed caching/persistence semantics for scan verdicts.
> 
> **Overview**
> Introduces **`PhishingDataService`** on top of **`BaseDataService`** so stalelist/hotlist, C2 blocklist, URL/token/address scans, and approvals go through messenger actions with TanStack query caching, optional **`StorageService`** persistence, and batched bulk scans. **`PhishingController`** keeps list/detection logic only and calls the service; the in-controller **`CacheManager`** and persisted **`urlScanCache` / `tokenScanCache` / `addressScanCache`** state (plus related constructor options) are removed.
> 
> **`BaseDataService`** gains bounded wait for cache rehydration after **`init`** (`hydrationTimeout`, **`DEFAULT_HYDRATION_TIMEOUT`**), optional **`shouldHydrateQuery`**, persisted-cache shape validation with discard-on-failure, and a protected **`policy`** getter for subclasses.
> 
> Behavior fixes bundled with the move: partial **`bulkScanUrls`** / bulk token results instead of failing the whole batch, skip unrecognized hotlist diff list types, negative caching for tokens with no API row, stricter rejection of malformed responses (not cached), and transaction-driven token scans coalesced per chain.
> 
> **Breaking for integrators:** register **`PhishingDataService`**, delegate its actions, call **`init`** when persistence is enabled, and migrate persisted controller state away from the removed scan-cache fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0b3196ba1f412eb10add4720767ab82c80b62c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


0xOhm:
@metamaskbot publish-previews

0xOhm:
@metamaskbot publish-previews

adonesky1:
@metamaskbot publish-previews

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+      },
+    });
+
+    const entries = urls.map((url) => {
+      const [hostname] = getHostnameFromWebUrl(url);
+      return this.fetchQuery({
+        queryKey: [`${this.name}:scanUrl`, hostname],
+        queryFn: async () => loader.load(url),
+        staleTime: SCAN_RESULT_STALE_TIME,
+      }).then((result) => [url, hostname, result] as const);
```
adonesky1: Could we key this with `getPhishingDetectionScanUrlParam(url)`? TanStack deduplicates in-flight queries that share this key, so two ipfs.io paths in one bulk call currently send only the first URL to the loader and reuse its verdict for both; the old implementation sent both URLs.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+  const executeChunk = async (chunk: PendingItem[]): Promise<void> => {
+    try {
+      const results = await executeBatch(chunk.map((item) => item.key));
+      for (const item of chunk) {
+        item.resolve(results[item.key] ?? null);
+      }
+    } catch (error) {
+      for (const item of chunk) {
+        item.reject(error);
+      }
```
adonesky1, line 336: We should retry `executeBatch` once as a batch, or disable policy retries for these per-item loader queries. A failed 50-item POST rejects 50 fetchQuery calls, and each starts its own default three-retry Cockatiel loop, so one failed batch can produce up to 150 independently scheduled follow-up POSTs.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+      return this.fetchQuery({
+        queryKey: [`${this.name}:scanUrl`, hostname],
+        queryFn: async () => loader.load(url),
+        staleTime: SCAN_RESULT_STALE_TIME,
+      }).then((result) => [url, hostname, result] as const);
+    });
+    loader.flush();
+
+    const results: Record<string, PhishingDetectionScanResult> = {};
+    for (const [url, hostname, result] of await Promise.all(entries)) {
```
adonesky1: We should preserve successful cache hits when a cache miss fails. `Promise.all` rejects the entire service call, so one failed miss causes the controller to return an empty result set, including dropping cached BLOCK verdicts that the previous implementation returned.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+        );
+        return this.#toJson(response);
+      },
+      staleTime: SCAN_RESULT_STALE_TIME,
+    });
+
+    return this.#validate(
+      jsonResponse,
+      ScanUrlResponseStruct,
+      'URL scan',
```
adonesky1: We should pass `responseStruct` into `fetchQuery` for direct queries, especially `scanUrl` and `scanAddress`, instead of validating afterward. TanStack calls `setData` before control reaches `#validate`, so a malformed 200 becomes a successful one-minute cache entry and may be persisted even though the caller receives an error.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+  };
+  let pending: PendingItem[] = [];
+  let flushScheduled = false;
+
+  const executeChunk = async (chunk: PendingItem[]): Promise<void> => {
+    try {
+      const results = await executeBatch(chunk.map((item) => item.key));
+      for (const item of chunk) {
+        item.resolve(results[item.key] ?? null);
+      }
```
adonesky1: I think we should avoid caching `null` for URLs present in `response.errors`. The first call returns the API error, but a repeat call within staleTime hits fresh null, skips the batch request, and returns neither a result nor the original error.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+      queryClientConfig,
+      // Circuit breaking is disabled by default: this service talks to four
+      // independent API hosts through a single shared policy, so a broken
+      // circuit caused by one host's outage would also pause phishing-list
+      // updates from the others. Protection against hammering a failing host
+      // comes from the controller's refresh-interval bookkeeping and the scan
+      // result stale times, matching the previous in-controller behavior.
+      policyOptions: {
+        maxConsecutiveFailures: Number.MAX_SAFE_INTEGER,
+        ...policyOptions,
```
adonesky1, line 448: Lets make the retry policy explicit rather than inheriting handleAll with maxRetries: 3. The previous controller made one request and comparable migrations defaulted to maxRetries: 0; if retries are intentional, could we use the Chomp/Subscription filter so deterministic 400/404 responses are not retried?

packages/phishing-controller/src/PhishingDataService.test.ts
```diff
@@ -0,0 +1,712 @@
...
+        rootMessenger.call(
+          'PhishingDataService:getApprovals',
+          'ethereum',
+          '0x1234567890123456789012345678901234567890',
+        ),
+      ).rejects.toThrow('Malformed response received from approvals endpoint');
+    });
+  });
+
+  describe('persistence', () => {
```
adonesky1, line 764: Can we add Phishing-specific rehydration and maxAge expiry tests before making persistence the default. This test only verifies StorageService:setItem; it does not register getItem/removeItem or call init(), so the claimed zero-network MV3 restart and expired-cache behavior are not currently covered.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+  /**
+   * Fetches the full phishing detection stalelist.
+   *
+   * @returns The stalelist response.
+   */
+  async getStalelist(): Promise<DataResultWrapper<PhishingStalelist>> {
+    const jsonResponse = await this.fetchQuery({
+      queryKey: [`${this.name}:getStalelist`],
+      queryFn: async () => this.#getJson(METAMASK_STALELIST_URL),
+      staleTime: 0,
```
adonesky1, line 475: Could we set `gcTime: 0` on the three list queries, the way `getApprovals` already does? They use `staleTime: 0` so they always refetch, but the response still lands in the query cache, and `#persistCache` dehydrates every success query with no key filter on a debounce that fires on any cache event, including every URL scan.

The live `/v1/stalelist` is 2.3 MB across 98,829 blocklist entries, so an active browsing session rewrites a multi-megabyte blob to client storage roughly every 10s, duplicating lists we already persist in `phishingLists`. I confirmed with a 20k-entry list that the stalelist body does end up inside the `StorageService:setItem` payload.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,789 @@
...
+          {
+            method: 'GET',
+            headers: {
+              Accept: 'application/json',
+            },
+          },
+        );
+        return this.#toJson(response);
+      },
+      staleTime: SCAN_RESULT_STALE_TIME,
```
adonesky1, line 562: Could we set an explicit `gcTime` on the scan queries? TanStack defaults `gcTime` to `Infinity` when `isServer` is true, and `isServer` is `typeof window === 'undefined'`, which holds in the MV3 service worker. Entries go stale after `SCAN_RESULT_STALE_TIME` but are never evicted, so extension background memory grows with every distinct hostname, token, and address scanned; the `CacheManager` this replaces had explicit `maxSize` bounds. Mobile is unaffected because React Native defines `window`.

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -67,209 +70,53 @@ import {
...
- * provider to downstream client attribution name.
- */
-export const phishingListKeyNameMap = {
-  [ListKeys.EthPhishingDetectConfig]: ListNames.MetaMask,
-};
+// Request timeouts, in milliseconds.
+const URL_SCAN_TIMEOUT = 8000;
+const BULK_URL_SCAN_TIMEOUT = 15000;
+const TOKEN_SCAN_TIMEOUT = 8000;
+const ADDRESS_SCAN_TIMEOUT = 5000;
```
adonesky1, line 118: These timeouts were sized when each call was a single `fetch`; they now race a policy that can make four attempts with backoff between them, and nothing cancels the query on timeout since no `AbortSignal` is threaded into `fetch`. A slow host leaves the query retrying after the caller has given up, and because `fetchQuery` dedupes in-flight queries, the next call for the same address attaches to that hung request and times out too.

If we make the retry policy explicit as suggested above, could we also either size these to cover the retry budget or scope the timeout inside `queryFn` so it bounds each attempt?

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,11 +7,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Requests are wrapped in a shared retry policy, configurable via the `policyOptions` constructor option; circuit breaking is disabled by default because the service spans four independent API hosts and a broken circuit caused by one host would pause phishing-list updates from the others
+  - Scan results are cached per URL hostname, token, and address for `SCAN_RESULT_STALE_TIME` (1 minute, matching the previous cache TTLs); bulk scans only request items without a fresh cached result and coalesce them into batched API calls (up to 50 URLs / 100 tokens per request), and single and bulk URL scans share cache entries; approvals are never cached
+  - The query cache is persisted between sessions by default (`persistenceConfig`, max age 5 minutes), which requires the `StorageService:setItem`, `StorageService:getItem`, and `StorageService:removeItem` messenger actions and an `init` call during client initialization (automatic with `@metamask/wallet`); pass `persistenceConfig: null` to disable
```
adonesky1: Two corrections here. `PhishingDataService` is not registered in `packages/wallet/src/initialization/instances`, so `init()` is not automatic with `@metamask/wallet` yet; mobile works only because `phishing-data-service-init.ts` calls it explicitly. Could we either register the service in `packages/wallet` or reword this to say clients must call `init()` themselves?

Separately, the bullet above claims single and bulk URL scans share cache entries, but that holds only for non-gateway hosts: `scanUrl` keys on `scanUrlParam` (hostname plus path) while `bulkScanUrls` keys on bare hostname.

adonesky1 (commented):
I merged main into this branch to clear the conflicts (CHANGELOG, package.json, yarn.lock, all union resolutions). Worth knowing that #10007 landed on main while this was open and removed `fetchQuery` from every `KycService` POST for three reasons that all apply here: deduped in-flight writes, responses published on the messenger as `cacheUpdated` payloads, and the service policy retrying non-idempotent POSTs. Those are the same root cause behind the bulk dedupe and retry fan-out comments I left earlier.

Could we reconcile with that precedent, keeping `fetchQuery` for the GET list endpoints and `scanUrl` but calling `#postJson` directly for `scanAddress` and `getApprovals`? The bulk endpoints genuinely need per-item caching so they are a real exception, but it would help to say so explicitly in a comment.

Also note #6388 (C2 blocklist `Array` to `Set`) is now in the base. It only touched `PhishingDetector.ts`, which this PR does not modify, so the merge is clean there.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Measured this: 10 URLs against a failing bulk endpoint produced 22 POSTs, one batch of 10 followed by 21 single-URL retries. The retries de-batch rather than re-sending the batch, so a failing host gets many small requests instead of one retried batch.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Confirmed: two consecutive `scanUrl` calls against a malformed 200 both throw, but only one network request is made. The malformed body sits in the cache as a success for the full minute, and `#persistCache` dehydrates `status === 'success'` queries, so it reaches disk too.

adonesky1:
@metamaskbot publish-previews

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,911 @@
...
+ * actions and events.
+ */
+export const serviceName = 'PhishingDataService';
+
+export const PHISHING_CONFIG_BASE_URL =
+  'https://phishing-detection.api.cx.metamask.io';
+export const METAMASK_STALELIST_FILE = '/v1/stalelist';
+export const METAMASK_HOTLIST_DIFF_FILE = '/v2/diffsSince';
+
+export const CLIENT_SIDE_DETECION_BASE_URL =
```
mcmire, line 78: Typo

``` suggestion
export const CLIENT_SIDE_DETECTION_BASE_URL =

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 13]
```diff
@@ -0,0 +1,911 @@
...
+    messenger: PhishingDataServiceMessenger;
+    queryClientConfig?: QueryClientConfig;
+    policyOptions?: CreateServicePolicyOptions;
+    persistenceConfig?: PersistenceConfiguration | null;
+  }) {
+    super({
+      name: serviceName,
+      messenger,
+      queryClientConfig,
+      // Circuit breaking is disabled by default: this service talks to four
```
mcmire: Instead of having this class represent four distinct APIs, have we considered creating four separate service classes? Ideally each data service class should act more or less like a traditional API client.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,911 @@
...
+
+  /**
+   * Fetches the full phishing detection stalelist.
+   *
+   * @returns The stalelist response.
+   */
+  async getStalelist(): Promise<DataResultWrapper<PhishingStalelist>> {
+    const jsonResponse = await this.fetchQuery({
+      queryKey: [`${this.name}:getStalelist`],
+      // Validated inside the query function so that a malformed response is
```
mcmire: This is the second time I've seen a team opt for this strategy. I suppose maybe we should bake this into `fetchQuery`.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,911 @@
...
+        this.#validate(
+          await this.#getJson(METAMASK_STALELIST_URL),
+          StalelistResponseStruct,
+          'stalelist',
+        ) as Json,
+      staleTime: 0,
+      gcTime: LIST_GC_TIME,
+    });
+
+    return jsonResponse as DataResultWrapper<PhishingStalelist>;
```
mcmire: Why the typecast? Ideally `this.#validate` should return the correct type (as it's the one ensuring that the data matches a known type).

(edit: I'll try to come back to this and make a better suggestion here)

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Client state migrations should remove these properties from persisted `PhishingController` state
+- **BREAKING:** Remove the `urlScanCacheTTL`, `urlScanCacheMaxSize`, `tokenScanCacheTTL`, `tokenScanCacheMaxSize`, `addressScanCacheTTL`, and `addressScanCacheMaxSize` options from `PhishingControllerOptions`; scan result freshness is now controlled by `SCAN_RESULT_STALE_TIME` in `PhishingDataService` ([#9914](https://github.com/MetaMask/core/pull/9914))
+- Tokens for which the bulk scanning API returns no result are now negatively cached for `SCAN_RESULT_STALE_TIME` instead of being re-requested on every call ([#9914](https://github.com/MetaMask/core/pull/9914))
+- `scanUrl` now reports the underlying error message in `fetchError` for network errors instead of `'timeout of 8000ms exceeded'` ([#9914](https://github.com/MetaMask/core/pull/9914))
+- Requests are no longer retried by default; the previous in-controller implementation made a single request per call, and the controller's timeouts are sized for one attempt. Pass `policyOptions.maxRetries` to opt back in ([#9914](https://github.com/MetaMask/core/pull/9914))
```
mcmire: Pass `policyOptions.maxRetries` to what? It would be helpful to more clear about which class is being changed here.

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING:** Remove the `urlScanCache`, `tokenScanCache`, and `addressScanCache` properties from `PhishingControllerState`; scan results are now cached (and persisted) by `PhishingDataService`'s query cache ([#9914](https://github.com/MetaMask/core/pull/9914))
+  - Client state migrations should remove these properties from persisted `PhishingController` state
+- **BREAKING:** Remove the `urlScanCacheTTL`, `urlScanCacheMaxSize`, `tokenScanCacheTTL`, `tokenScanCacheMaxSize`, `addressScanCacheTTL`, and `addressScanCacheMaxSize` options from `PhishingControllerOptions`; scan result freshness is now controlled by `SCAN_RESULT_STALE_TIME` in `PhishingDataService` ([#9914](https://github.com/MetaMask/core/pull/9914))
+- Tokens for which the bulk scanning API returns no result are now negatively cached for `SCAN_RESULT_STALE_TIME` instead of being re-requested on every call ([#9914](https://github.com/MetaMask/core/pull/9914))
+- `scanUrl` now reports the underlying error message in `fetchError` for network errors instead of `'timeout of 8000ms exceeded'` ([#9914](https://github.com/MetaMask/core/pull/9914))
```
mcmire: Should we clarify that it's `scanUrl` in `PhishingController` that's being changed?

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Requests are no longer retried by default; the previous in-controller implementation made a single request per call, and the controller's timeouts are sized for one attempt. Pass `policyOptions.maxRetries` to opt back in ([#9914](https://github.com/MetaMask/core/pull/9914))
+- Malformed API responses (e.g. a stalelist without a numeric `lastUpdated`, or scan results without a `recommendedAction`/`result_type`) are now rejected and treated as request failures instead of being passed through, and are not cached ([#9914](https://github.com/MetaMask/core/pull/9914))
+- `bulkScanUrls` now returns the results it was able to resolve even if some lookups fail, reporting the failures per URL in `errors`; it only rejects when no result could be resolved at all. Previously a single failed lookup discarded every result in the batch, including cached `BLOCK` verdicts for unrelated URLs ([#9914](https://github.com/MetaMask/core/pull/9914))
```
mcmire: Using multiple sentences for a changelog entry is a sign that it's too long, can more details be given in a sub-bullet?

packages/phishing-controller/CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `bulkScanUrls` no longer caches a "no result" verdict for URLs the API reported an error for, so those URLs are retried on the next call instead of being silently skipped for a minute ([#9914](https://github.com/MetaMask/core/pull/9914))
```
mcmire: Is a bug being fixed? If so should this be in the "Fixed" category?

packages/phishing-controller/package.json
```diff
@@ -52,10 +52,15 @@
   "dependencies": {
     "@metamask/address-book-controller": "^8.0.0",
     "@metamask/base-controller": "^10.0.0",
+    "@metamask/base-data-service": "^2.0.0",
```
mcmire: Nit: Should we document that `@metamask/base-data-service` was added as a dependency in the changelog?

packages/phishing-controller/package.json
```diff
@@ -52,10 +52,15 @@
...
     "@metamask/controller-utils": "^13.0.0",
     "@metamask/messenger": "^3.0.0",
+    "@metamask/storage-service": "^2.0.0",
```
mcmire: Nit: Should we document that `@metamask/storage-service` was added as a dependency in the changelog?

packages/phishing-controller/CHANGELOG.md
mcmire: Nit: Usually we note in the changelog when dependencies are added or updated. Can we do this for `@metamask/base-data-service` and `@metamask/storage-service`?

``` suggestion
- `bulkScanUrls` no longer caches a "no result" verdict for URLs the API reported an error for, so those URLs are retried on the next call instead of being silently skipped for a minute ([#9914](https://github.com/MetaMask/core/pull/9914))
- Add `@metamask/base-data-service` (`^2.0.0`) as a dependency ([#9914](https://github.com/MetaMask/core/pull/9914))
- Add `@metamask/base-data-service` (`^2.0.0`) as a dependency ([#9914](https://github.com/MetaMask/core/pull/9914))
```

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1159 @@
...
+   * @returns The approvals response.
+   */
+  async getApprovals(
+    chain: string,
+    address: string,
+  ): Promise<ApprovalsResponse> {
+    // Deliberately not routed through `fetchQuery`. Approvals reflect live,
+    // account-specific state that is never cached, so the query cache would
+    // provide no benefit while publishing the response on the messenger as a
+    // `cacheUpdated` payload. This matches the handling of non-cached POSTs
```
mcmire: `fetchQuery` provides some other benefits, for instance, you can pass a `responseStruct` option to automatically validate the response. Could you pass `staleTime: 0` to bypass the caching? I think this is usually the preferred approach in other packages.

mcmire (commented):
I have just started looking at this PR. Here are some comments for now. I will do another pass later.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1278 @@
...
+            ...queryClientConfig.defaultOptions?.queries,
+            // Hydration reconstructs queries using these defaults. Without an
+            // explicit value, service workers receive TanStack's server
+            // default of `Infinity`, which cannot later be reduced by a
+            // per-query option.
+            gcTime: SCAN_RESULT_GC_TIME,
+          },
+        },
+      },
+      // Circuit breaking is disabled by default: this service talks to four
```
mcmire, line 624: Have we considered creating four data services, one for each API? Would that it easier to finetune the behavior? Or is our goal here to mimic what the controller was doing before?

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1278 @@
...
+   * @param timeout - The optional request timeout, in milliseconds.
+   * @returns The parsed JSON response.
+   */
+  async #fetchJson(
+    url: string,
+    init: RequestInit,
+    timeout?: number,
+  ): Promise<Json> {
+    const controller = new AbortController();
+    const sourceSignal = init.signal;
```
mcmire: Why are we handling two AbortController signals here?

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -390,6 +435,19 @@ export class BaseDataService<
...
   }
 
+  /**
+   * Executes an operation using this service's retry and circuit-breaker
+   * policy without adding the result to the query cache.
+   *
+   * @param operation - The asynchronous operation to execute.
+   * @returns The operation result.
+   */
+  protected async executeWithPolicy<Result>(
```
mcmire: I wonder if it makes more sense to expose the `policy` property rather than wrapping `policy.execute()`? I recall data services from other teams where it seems like it would have been nice to have access to the policy object.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1278 @@
...
+   * @param init - The fetch request options.
+   * @param timeout - The optional request timeout, in milliseconds.
+   * @returns The parsed JSON response.
+   */
+  async #fetchJson(
+    url: string,
+    init: RequestInit,
+    timeout?: number,
+  ): Promise<Json> {
+    const controller = new AbortController();
```
mcmire, line 1204: We already have an AbortController kept under `this.#abortController`. Why are we creating another one?

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1278 @@
...
+        () => undefined,
+      );
+    }
+  };
+
+  return {
+    async load(key: string): Promise<Json | null> {
+      return new Promise((resolve, reject) => {
+        pending.push({ key, resolve, reject });
+        // Items registered outside an explicit flush (e.g. by the retry
```
mcmire: I don't understand why this is necessary, can you shine some light here?

mcmire (commented):
Still working through this. Left some more comments.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Agreed on the fix, but `CLIENT_SIDE_DETECION_BASE_URL` predates this PR ([PhishingController.ts#L84 at the base commit](https://github.com/MetaMask/core/blob/aa715321f6c3ba8109c95517294512077b0ed4e4/packages/phishing-controller/src/PhishingController.ts#L84)) and is public via `export *` from the package index ([index.ts#L1](https://github.com/MetaMask/core/blob/aa715321f6c3ba8109c95517294512077b0ed4e4/packages/phishing-controller/src/index.ts#L1)); this PR only moved it and kept a compat re-export ([PhishingController.ts#L91](https://github.com/MetaMask/core/blob/337bbda8793d1f8cf474bfa761a9c6f0b027b087/packages/phishing-controller/src/PhishingController.ts#L91)). A plain rename would be breaking, so I'd add `CLIENT_SIDE_DETECTION_BASE_URL` and keep the old name as a deprecated alias. Fine with a hard rename instead if you'd rather take it in this major?

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: The casts come from `fetchQuery`'s `TQueryFnData extends Json` bound: superstruct `optional()` infers `T | undefined`, which isn't `Json`, and `responseStruct` needs a `Struct<Json>` for the same reason, so it can't take these structs without casting the raw response first. For the two structs without optionals (stalelist and address scan) the validated type satisfies the bound on its own, so I dropped the casts there in d04bcaa2da. Removing the rest means relaxing that bound in `base-data-service`; I'd do that as a follow-up unless you want it here.

packages/phishing-controller/CHANGELOG.md
adonesky1: Folded into the `PhishingDataService` entry under Added and named the constructor option explicitly in 337bbda879; this line was duplicating that sub-bullet.

packages/phishing-controller/CHANGELOG.md
adonesky1: Changed to `PhishingController.scanUrl` in 337bbda879.

packages/phishing-controller/CHANGELOG.md
adonesky1: Moved to Fixed with the previous behavior as a sub-bullet in 337bbda879, and gave the per-entry validation entry the same split.

packages/phishing-controller/CHANGELOG.md
adonesky1: Whoops, some of these entries are stale from past iterations of this work: this one and the entries for the approvals policy, `scanUrl` hostname, `fetchError` results, accepted verdicts, `bulkScanTokens` partial results, and destroy aborting requests all fixed code that never shipped. Consolidated and cleaned up in 337bbda879; the two that fix released behavior (`bulkScanUrls` partial results and unknown hotlist list types) now sit under Fixed.

adonesky1: Added all five in 337bbda879: `@metamask/base-data-service` `^2.0.0`, `@metamask/storage-service` `^2.0.0`, `@metamask/superstruct` `^3.4.1`, `@metamask/utils` `^11.12.0`, and `@tanstack/query-core` `^5.62.16`. Your suggestion listed `base-data-service` twice; I assumed the second was `storage-service`.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Agreed, done in d04bcaa2da with `gcTime: 0` alongside `staleTime: 0` (staleTime alone leaves the entry cached until GC). Reading `KycService` again, the bypass there is for writes, and `shield-api-service` routes POST-shaped reads through `fetchQuery` the same way, so this matches precedent better than the direct call did. `responseStruct` doesn't fit here because approvals are validated per entry, so validation stays inside the query function.

packages/phishing-controller/src/PhishingDataService.ts
mcmire: Ah I didn't realize. I'm fine with renaming + adding a deprecated alias to avoid a new major version. But I'm also fine to leave that for another PR.

packages/phishing-controller/src/PhishingDataService.ts
mcmire: Gotcha. Yeah, we'll have to figure out a way to relax the types in `BaseDataService`. That's clearly out of scope here so these typecasts are fine.

cc @FrederikBolding

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Leaving the rename and deprecated alias for a follow-up PR so this one stays scoped to the data service migration.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Parity with the controller is the goal here, deliberately: one service mirrors what the controller did, and the shared policy is the reason retries and circuit breaking default to off. Splitting into per-host services would let each enable a breaker safely, but it multiplies client wiring (four messengers, action-type files, allowed-action sets, and init/storage keys in both clients) for a capability the released controller never had. I'd keep one service in this PR and note the trade-off in the class docblock; the controller only reaches the service via actions, so a split later is contained. Objections?

packages/base-data-service/src/BaseDataService.ts
adonesky1: Done in 2c421e3a0e: `executeWithPolicy` only ever existed on this branch, so I swapped it for a protected `policy` getter rather than keeping a wrapper. Subclasses can still run uncached requests under the shared policy with `this.policy.execute(...)`, and they now also get `onBreak`, `onDegraded`, and `onRetry`, which the hand-rolled services forward today but `BaseDataService` subclasses couldn't reach.

packages/phishing-controller/src/PhishingDataService.ts
mcmire: Okay, that's what I thought. Typically we've tried to keep the responsibility of data services small and focused on _one_ API. In the future, we may want to consolidate data services so that all interactions for an API always go through _one_ data service instead of potentially being split up among packages. But it seems like there is a good reason for taking the approach we are taking, so I'm not opposed. We can think about this more later.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: `fetch` only takes one cancel signal, but a request can be cancelled for up to three reasons: TanStack cancels that query (single-item requests only; batched requests aren't tied to one query), the service is destroyed (that's `this.#abortController`), or the request's own timeout fires. The per-request controller merges those into the one signal `fetch` accepts, and keeps the timeout separate so the error can say it was a timeout. Comment added in 8d5a1c9ed6 and cff89ad93c.

To be clear this isn't shape I think we want long term. `AbortSignal.any` would be the one-liner but mobile's RN polyfill doesn't have it (its `shim.js` has to back-fill even `AbortSignal.timeout`), so it's the portable option for now. The better home is `BaseDataService`: hand each query function a signal covering both the query and the service's teardown, with an optional timeout built on cockatiel's `timeout` policy, which already models this and throws a typed `TaskCancelledError`. That would delete this wiring here and give every subclass the same guarantees. Happy to write that up as a follow-up if you agree with the direction.

packages/phishing-controller/src/PhishingDataService.ts
adonesky1: Fair question, and the comment there was out of date: it pointed at retried item queries, which no longer exist. What the microtask flush actually covers now is the `init()` case, where `fetchQuery` waits for cache rehydration before running each lookup, so the lookups arrive after the explicit `flush()` has already run on an empty list. Looking at that, the explicit flush turned out to be redundant: all the lookups of one bulk call are issued in the same turn, so scheduling one flush as a microtask catches them all. Removed the explicit `flush` and its API in c6bd755740. Checking this also turned up a gap in the rehydration wait: it used a timer per query, which split a bulk call into single requests if rehydration timed out and made every query pay the full timeout while storage hung. 5f89fc129e shares one deadline per service instead, with tests for both cases.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -512,6 +607,14 @@ export class BaseDataService<
...
 
+    if (!is(untypedCache, PersistedCacheStruct)) {
+      await this.#externalMessenger.call(
+        'StorageService:removeItem',
+        this.name,
+        STORAGE_SERVICE_KEY,
+      );
+      return;
+    }
     const cache = untypedCache as unknown as PersistedCache;
```
mcmire: Since we are now using a struct to validate the cache object, can we use `validate` to get a typed cache, this way we can avoid the type assertion below? 

``` suggestion
    const [cache, errors] = validate(untypedCache, PersistedCacheStruct);

    if (errors.length > 0) {
      await this.#externalMessenger.call(
        'StorageService:removeItem',
        this.name,
        STORAGE_SERVICE_KEY,
      );
      return;
    }

File: packages/base-data-service/src/BaseDataService.ts [hunk 29]
```diff
@@ -131,13 +145,35 @@ export type PersistenceConfiguration = {
...
+   */
+  shouldHydrateQuery?: (query: DehydratedState['queries'][number]) => boolean;
 };
 
 type PersistedCache = {
   state: DehydratedState;
   timestamp: number;
 };
 
+const PersistedCacheStruct = objectType({
```
mcmire, line 169: What do you think about grouping this `const`s with the others and moving it below `DEFAULT_HYDRATION_TIMEOUT`?

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -131,13 +145,35 @@ export type PersistenceConfiguration = {
...
+   * to validate persisted data before it can be served from the cache.
+   * Defaults to restoring every persisted query.
+   */
+  shouldHydrateQuery?: (query: DehydratedState['queries'][number]) => boolean;
 };
 
 type PersistedCache = {
   state: DehydratedState;
   timestamp: number;
 };
```
mcmire: Now that we have a struct, can we derive this type from the `const`? (`Infer` comes from `@metamask/superstruct`.)

``` suggestion
type PersistedCache = Infer<typeof PersistedCacheStruct>;

File: packages/base-data-service/src/BaseDataService.ts [hunk 31]
```diff
@@ -408,12 +468,47 @@ export class BaseDataService<
...
+   * waits again while storage remains unresponsive.
+   *
+   * Callers must only await this when `init` has been called: an asynchronous
+   * hop before a query starts changes its timing relative to callers' own
+   * timers, so services without persistence keep starting queries
+   * synchronously.
+   *
+   * @param initialization - The pending rehydration.
+   */
+  async #waitForInitialization(initialization: Promise<void>): Promise<void> {
```
mcmire, line 495: What are your thoughts on baking this into `init` itself? That way we only need one property to hold the initialization promise (`this.#initializationPromise`) and not two.

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -512,6 +607,14 @@ export class BaseDataService<
       return;
     }
 
+    if (!is(untypedCache, PersistedCacheStruct)) {
```
mcmire, line 610: Considering that `PersistedCacheStruct` and `PersistedCache` are similarly named, it might be tempting to make use of the struct to more idiomatically assign a type to `cache` rather than using a type assertion, e.g.:

``` typescript
    const [error, cache] = validate(untypedCache, PersistedCacheStruct);

    if (error) {
      await this.#externalMessenger.call(
        'StorageService:removeItem',
        this.name,
        STORAGE_SERVICE_KEY,
      );
      return;
    }
```

However, this would not work, because `PersistedCacheStruct` is more loosely defined than `PersistedCache`. 

Maybe using a struct here is too heavy-handed and we can get away with simple checks?

``` suggestion
    if (
      !(
        isPlainObject(untypedCache) &&
        hasProperty(untypedCache, 'timestamp') &&
        typeof untypedCache.timestamp === 'number' &&
        hasProperty(untypedCache, 'state') &&
        isPlainObject(untypedCache.state) &&
        hasProperty(untypedCache.state, 'queries') &&
        Array.isArray(untypedCache.state.queries) &&
        hasProperty(untypedCache.state, 'mutations') &&
        Array.isArray(untypedCache.state.mutations)
      )
    ) {
```

We can also move this to an `isPersistedCache` function if it's too much to see here.

What do you think?

packages/base-data-service/src/BaseDataService.ts
mcmire: Alternatively, now that we have this struct, it's a bit confusing to have this + `PersistedCache`. Usually we derive types from structs, but in this case, if we were to derive a type from `PersistedCacheStruct` and call it `PersistedCache`, it would be too broad.

Maybe we don't need this struct after all? (also see below)

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -390,6 +438,18 @@ export class BaseDataService<
...
 
+  /**
+   * The retry and circuit-breaker policy that wraps every query made through
+   * this service. Subclasses can run uncached requests under the same policy
+   * with `this.policy.execute(...)`, or observe its `onBreak`, `onDegraded`,
+   * and `onRetry` events.
+   *
+   * @returns The service policy.
+   */
+  protected get policy(): ServicePolicy {
```
mcmire, line 449: Is it worth adding a comment explaining why this is a getter?

``` suggestion
  // A getter so that `policy` cannot be overwritten in a JavaScript project.
  protected get policy(): ServicePolicy {

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 34]
```diff
@@ -0,0 +1,1281 @@
...
+   * @param response - The response to validate.
+   * @param struct - The struct to validate against.
+   * @param endpointName - The name of the endpoint, used in error messages.
+   * @returns The validated response.
+   */
+  #validate<Type, Schema>(
+    response: unknown,
+    struct: Struct<Type, Schema>,
+    endpointName: string,
+  ): Infer<Struct<Type, Schema>> {
```
mcmire, line 1273: Would this also work?

``` suggestion
  ): Type {

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 35]
```diff
@@ -0,0 +1,1281 @@
...
+            signal,
+          }),
+          HotlistDiffsResponseStruct,
+          'hotlist diffs',
+        ) as Json,
+      staleTime: 0,
+      gcTime: LIST_GC_TIME,
+    });
+
+    return jsonResponse as DataResultWrapper<Hotlist>;
```
mcmire, line 723: It seems that you are validating that the JSON response matches a certain shape — at the same time getting TypeScript to assign a type to your response data — and then you're saying "no, actually, the data is this".

From the comment that sits above `targetList` in `HotlistDiffsResponseStruct`, it appears that `HotlistDiffsResponseStruct` allows `targetList` to be  _any_ string, and yet `Hotlist` wants it to be a _specific_ string. Should this be loosened in `Hotlist`? Otherwise, you're allowing unverified data and thus a bug to slip into the processing logic.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+        this.#validate(
+          await this.#getJson(url, { signal }),
+          C2DomainBlocklistResponseStruct,
+          'C2 domain blocklist',
+        ) as Json,
+      staleTime: 0,
+      gcTime: LIST_GC_TIME,
+    });
+
+    return jsonResponse as C2DomainBlocklistResponse;
```
mcmire, line 753: Is it possible to drop these type assertions?

``` suggestion
        ),
      staleTime: 0,
      gcTime: LIST_GC_TIME,
    });

    return jsonResponse;

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 37]
```diff
@@ -0,0 +1,1281 @@
...
+        return {
+          ...scanResult,
+          hostname: scanResult.hostname ?? hostname,
+        } as Json;
+      },
+      staleTime: SCAN_RESULT_STALE_TIME,
+      gcTime: SCAN_RESULT_GC_TIME,
+    });
+
+    return jsonResponse as PhishingDetectionScanResult;
```
mcmire, line 796: Is it possible to drop these type assertions?

``` suggestion
        };
      },
      staleTime: SCAN_RESULT_STALE_TIME,
      gcTime: SCAN_RESULT_GC_TIME,
    });

    return jsonResponse;

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 38]
```diff
@@ -0,0 +1,1281 @@
...
+          } else if (result.fetchError) {
+            itemErrors[key] = new BatchItemError(result.fetchError);
+          } else {
+            // Entries are shared with `scanUrl`, whose results always carry a
+            // hostname, so fill it in before the entry reaches the cache.
+            const [hostname] = getHostnameFromWebUrl(key);
+            results[key] = {
+              ...result,
+              hostname: result.hostname ?? hostname,
+            } as Json;
```
mcmire, line 860: Is it possible to drop this type assertion?

``` suggestion
            };

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 39]
```diff
@@ -0,0 +1,1281 @@
...
+      if (outcome.status === 'rejected') {
+        const reason = unwrapBatchError(outcome.reason);
+        addError(url, getErrorMessage(reason));
+        if (!(outcome.reason instanceof BatchItemError)) {
+          requestFailure ??= { reason };
+        }
+        continue;
+      }
+
+      results[url] = outcome.value as PhishingDetectionScanResult;
```
mcmire, line 916: Why the type assertion? `outcome.value` is `Json`, but `PhishingDetectionScanResult` is not compatible. Is there a better type you can assign to `entries`?

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+      return this.#fetchTokenScanQuery(loader, chain, normalizedToken).then(
+        (result) => [normalizedToken, result] as const,
+      );
+    });
+
+    const results: TokenScanApiResponse['results'] = {};
+    let firstError: Error | undefined;
+    for (const outcome of await Promise.allSettled(entries)) {
+      if (outcome.status === 'rejected') {
+        firstError ??= unwrapBatchError(outcome.reason) as Error;
```
mcmire, line 985: Why the typecast? `unwrapBatchError` returns `unknown`. Is there a better type you can assign to `firstError`? Can you wrap `firstError` in an error? We have a `ensureError` function in `@metamask/utils`.

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+        const response = this.#validate(
+          jsonResponse,
+          BulkScanTokensResponseStruct,
+          'bulk token scan',
+        );
+        const results: Record<string, Json> = {};
+        const errors: Record<string, Error> = {};
+        for (const [key, result] of Object.entries(response.results ?? {})) {
+          if (is(result, TokenScanResultStruct)) {
+            results[key] = result as Json;
```
mcmire, line 1029: Is this type assertion necessary?

``` suggestion
            results[key] = result;

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 42]
```diff
@@ -0,0 +1,1281 @@
...
+    chain: string,
+    token: string,
+  ): Promise<TokenScanResultResponse | null> {
+    const result = await this.fetchQuery({
+      queryKey: [`${this.name}:scanToken`, chain, token],
+      queryFn: async () => loader.load(token),
+      staleTime: SCAN_RESULT_STALE_TIME,
+      gcTime: SCAN_RESULT_GC_TIME,
+    });
+    return result as TokenScanResultResponse | null;
```
mcmire, line 1060: Instead of typecasting this, is there a way we can tell the loader what data we expect, by passing a type parameter into `BatchLoader`?

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+          ),
+        } as Json;
+      },
+      // Live account state: always refetch and evict as soon as the call
+      // settles, as other data services do for uncached reads.
+      staleTime: 0,
+      gcTime: 0,
+    });
+
+    return jsonResponse as ApprovalsResponse;
```
mcmire, line 1136: Similar as above:

``` suggestion
        };
      },
      // Live account state: always refetch and evict as soon as the call
      // settles, as other data services do for uncached reads.
      staleTime: 0,
      gcTime: 0,
    });

    return jsonResponse;

File: packages/phishing-controller/src/PhishingDataService.ts (hunk 26 above)
Comment (mcmire) [reply packages/phishing-controller/src/PhishingDataService.ts:1204]: Oh okay. So it sounds like one AbortController is service-level (and gets triggered when the service is destroyed) and the other is request-level (and may get triggered by TanStack Query or manually)? I'm asking because I'm wondering if this is a pattern that we need to port to BaseDataService.

File: packages/phishing-controller/src/PhishingDataService.ts [hunk 44]
```diff
@@ -0,0 +1,1281 @@
...
+   * responses. The error message intentionally matches the
+   * `<status> <statusText>` format historically produced by
+   * `PhishingController` so that consumers relying on it keep working.
+   *
+   * @param response - The response to parse.
+   * @returns The parsed JSON response.
+   */
+  async #toJson(response: Response): Promise<Json> {
+    if (!response.ok) {
+      throw new HttpError(
```
mcmire, line 1253: Nit: The HTTP status does not seem to correlate to converting the response to JSON. Maybe it's worth it to bake this into `#fetchJson` instead?

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+      return new Promise((resolve, reject) => {
+        pending.push({ key, resolve, reject });
+        // The first lookup in a turn schedules the flush as a microtask, which
+        // runs once the current synchronous work has finished, so every other
+        // lookup made in the same turn (for example all the item queries of
+        // one bulk call, whether they start synchronously or after awaiting
+        // cache rehydration) joins the same batch.
+        if (!flushScheduled) {
+          flushScheduled = true;
+          queueMicrotask(flush);
```
mcmire, line 544: This whole function looks similar to `reduceInBatchesSerially` from `assets-controller`: https://github.com/MetaMask/core/blob/1e24fcef9402a33502cf385c20c3539cc2f75226/packages/assets-controller/src/data-sources/evm-rpc-services/utils/batch.ts#L34

However, this part is a bit different. Do we need to schedule the flush as a microtask because requests go through a query client? `assets-controller` also uses a query client under the hood, but it doesn't need to do this. So I'm wondering what the differences are between the approach we're taking here and the approach that is being taken in `assets-controller` to batch requests and whether the complexity is worth it.

You can see where it's being used here for instance:

https://github.com/MetaMask/core/blob/1e24fcef9402a33502cf385c20c3539cc2f75226/packages/assets-controller/src/data-sources/PriceDataSource.ts#L357

(Asking because I'm wondering if this is a pattern we need to add to `base-data-service`, so I want to know the problems we're solving and what the best approach is.)

packages/phishing-controller/src/PhishingDataService.ts
```diff
@@ -0,0 +1,1281 @@
...
+            (retryFilterPolicy?.options.errorFilter(error) ?? true),
+        ),
+      },
+      persistenceConfig: persistenceConfig
+        ? {
+            ...persistenceConfig,
+            // Persisted scan results are validated before they can be served
+            // from the cache; a caller-provided filter is applied on top.
+            shouldHydrateQuery: (query): boolean =>
+              isValidPersistedScanQuery(query) &&
```
mcmire, line 656: What is this line connected to? Is it the fact we are batching requests or is it the fact that we are validating requests inside of the `queryFn`?

packages/phishing-controller/src/PhishingDataService.ts
FrederikBolding: `exactOptional` may be what we are looking for here to be JSON compatible?

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -167,6 +203,10 @@ export class BaseDataService<
 
   readonly #persistenceConfig?: PersistenceConfiguration;
 
+  #initializationPromise?: Promise<void>;
+
+  #boundedInitialization?: Promise<void>;
```
FrederikBolding, line 208: None of the changes to `BaseDataService` should be done in this PR, they should be opened as a separate PR IMO.

packages/base-data-service/src/BaseDataService.ts
FrederikBolding, line 610: What is this trying to guard against?

packages/base-data-service/src/BaseDataService.ts
```diff
@@ -284,6 +324,10 @@ export class BaseDataService<
     queryFn: QueryFunction<TQueryFnData, TQueryKey>;
     responseStruct?: TDataStruct;
   }): Promise<TData> {
+    if (this.#initializationPromise) {
+      await this.#waitForInitialization(this.#initializationPromise);
```
FrederikBolding, line 328: What's the use-case here? The thinking originally was that persisted state would be loaded during init anyways before queries start to fire.

packages/phishing-controller/src/PhishingDataService.ts
mcmire: I revisited this and came to a different conclusion as you @adonesky1. I don't think it's due to the types in `fetchQuery`. I left some more comments here: https://github.com/MetaMask/core/pull/9914#discussion_r4009732191

PR #9915, opened 2026-08-20
fix(base-data-service): replace cached page in place on re-fetch

### LINKED ISSUE #9900: base-data-service: fetchInfiniteQuery duplicates a page when a stale cached page is requested again
`BaseDataService.fetchInfiniteQuery` (query-core v5) appends a later page straight into the infinite cache without checking whether it is already there. Re-requesting a page that is already cached fetches it again and merges it, so the service's cache ends up with duplicated and out of order pages.

The realistic trigger is a refetch. When a UI's stale infinite query rebuilds, query-core re-requests each page through the service, so page 2 is requested again. Verified with `createUIQueryClient` and an `InfiniteQueryObserver` against the live behavior: after scrolling to page 2 (service cache params `[1, 2]`), a refetch leaves the service cache at `[2, 1, 2]`, and each further refetch grows it (`[2, 1, 2, 3]`, and so on).

The UI observers are not directly affected: query-core rebuilds each observer's own result cleanly, so users do not see duplicate items. The impact is on the service's shared cache, which accumulates duplicate and misordered pages over time (unbounded growth across refetches), and a newly hydrating observer can briefly receive the corrupted pages before its own fetch rebuilds them.

This is existing behavior, not a v5 regression. Verified against `@tanstack/query-core` 4.43.0: v4 also duplicated on every repeat request.

Suggested fix: before merging, if the requested page param is already present, replace that page in place instead of appending or prepending.

Scoped out of the query-core v5 bump (#9712), where Bugbot originally flagged it.

szupzj18:
## Explanation

Fixes #9900

`BaseDataService.fetchInfiniteQuery` forwarded re-fetches of an already-cached page param to `query.fetch` with a `fetchMore` direction. query-core's infinite-query behavior merges a fetched page by appending (forward) or prepending (backward) it, so re-requesting a page that was already in the cache left a **duplicate, out-of-order copy** in the cache. As the issue reports, the service cache grew on every refetch (e.g. `[1, 2]` → `[2, 1, 2]` → `[2, 1, 2, 3]`), and a newly hydrating observer could briefly receive the corrupted pages.

The fix detects that the requested page param is already present before the fetch, and after the fetch completes, collapses the edge-merged duplicate and places the fresh page at the original index via `setQueryData`. The fresh page is returned to the caller.

## Changes

- `fetchInfiniteQuery`: record the existing page index before fetching; when present, normalize the merged result so the cache holds each page exactly once, in order, with fresh data.
- Tests: two new cases covering the backward and forward `fetchMore` directions, asserting the cache holds each page once after a re-fetch.

## Checklist

- [x] Tests added/updated for the change (RED before the fix: cache held 2 copies; GREEN after)
- [x] `yarn workspace @metamask/base-data-service run test` passes (100% statements/lines/functions, branches above threshold)
- [x] `yarn lint:tsc` and ESLint on the changed files pass
- [x] Changelog entry added and validated

---

> [!NOTE]
> **Medium Risk**
> Touches infinite-query cache merge in BaseDataService, which can affect paginated data for all subclasses. Logic is small and covered by tests, but incorrect collapse would still corrupt page order.
> 
> **Overview**
> Stops `fetchInfiniteQuery` from duplicating pages when the same page param is requested again. query-core was appending or prepending a fresh copy, so the cache grew and went out of order.
> 
> If the page param is already in the cache, the extra merged page is collapsed and the fresh data is written back at the original index via `setQueryData`. If query-core skips the fetch (no next/previous page), the existing page is returned unchanged.
> 
> Tests cover backward and forward refetch, plus a no-adjacent-pages case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f7c5ad09f78c7ea6974e166465c307498611bba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -46,6 +46,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `@metamask/messenger` from `^1.2.0` to `^2.0.0` ([#9392](https://github.com/MetaMask/core/pull/9392))
 - Add dependency `cockatiel` (`^3.1.2`) ([#9418](https://github.com/MetaMask/core/pull/9418))
 
+### Fixed
+
+- Prevent `fetchInfiniteQuery` from duplicating a page when re-fetching a page param that is already present in the cache ([#9900](https://github.com/MetaMask/core/issues/9900))
+  - The fresh page now replaces the cached one in place instead of being appended or prepended, which previously left duplicate, out-of-order pages in the cache.
+
 ## [0.1.3]
 
 ### Changed
--- a/packages/base-data-service/CHANGELOG.md
+++ b/packages/base-data-service/CHANGELOG.md
@@ -48,7 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Prevent `fetchInfiniteQuery` from duplicating a page when re-fetching a page param that is already present in the cache ([#9900](https://github.com/MetaMask/core/issues/9900))
+- Prevent `fetchInfiniteQuery` from duplicating a page when re-fetching a page param that is already present in the cache ([#9915](https://github.com/MetaMask/core/pull/9915), [#9900](https://github.com/MetaMask/core/issues/9900))
   - The fresh page now replaces the cached one in place instead of being appended or prepended, which previously left duplicate, out-of-order pages in the cache.
 
 ## [0.1.3]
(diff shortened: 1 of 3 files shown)


szupzj18:
Good catch. Fixed in 6f7c5ad: the duplicate-page collapse now checks `result.pages.length > pages.length` before popping/shifting. When `getNextPageParam`/`getPreviousPageParam` returns null, `fetchInfiniteQuery` skips the fetch entirely and returns the existing data unchanged — in that case there is no duplicate to collapse, and the code now returns the cached page at the existing index instead of corrupting the cache.

Added a regression test: refetching a page with `hasNextPage: false, hasPreviousPage: false` (both param functions return null) must leave the cache with the single page intact.

All 69 tests pass; ESLint clean.

PR #9922, opened 2026-08-20
feat(perps-controller): persist pro-mode order, chart, and draft preferences

michalconsensys:
## Explanation

Perps Pro-mode trade preferences were not fully owned by `@metamask/perps-controller`. Clients had to keep order type, order-book "listed by" display, visible candle count, and the draft `reduceOnly` flag locally, while pending trade drafts lived for five minutes and were not cleared after a successful order.

This change moves those preferences into the controller and tightens draft restore:

- **`selectedOrderType`** is stored once for all markets (default `'market'`). Saving a pending trade configuration that carries an `orderType` also updates this value, so switching order type on one market carries over to the next.
- **`orderBookPreferences`** stores the "listed by" display as `{ currency: 'usd' | 'base', metric: 'total' | 'size' }` (defaults to USD totals). Order-book grouping remains per-market (`orderBookGrouping`).
- **`visibleCandleCount`** is stored globally for both Lite and Pro charts, defaults to `30`, and is clamped to `10–250` (non-finite input is ignored).
- **Pending trade drafts** now last **30 seconds** instead of five minutes, include `reduceOnly`, and are cleared after a **successful** `placeOrder`. Leverage (persisted per market) and the selected order type are retained across the reset.

Getters and selectors merge defaults, so older persisted state without the new fields still reads correctly, and the controller constructor merges defaults into any provided state — existing users need no migration. New messenger actions and selectors are exposed for each preference. Because three new fields were added to `PerpsControllerState`, consumers that construct a full state object must now include them (see References).

Last-market landing ([TAT-3702]) and market-list category persistence ([TAT-3736]) are intentionally out of scope and tracked separately.

## References

- Fixes: [TAT-3706]
- Related: [TAT-3702], [TAT-3736]
- Consumer impact (breaking): the new required `PerpsControllerState` fields (`selectedOrderType`, `orderBookPreferences`, `visibleCandleCount`) will require `metamask-mobile` to update the two spots that build a fully-typed state object — `app/components/UI/Perps/__mocks__/serviceMocks.ts` (`createMockPerpsControllerState`) and `app/core/Engine/controllers/perps-controller/index.test.ts` (`initialPerpsControllerState`). Runtime, imports, and persisted-state rehydration remain backward-compatible.

## Changelog

`@metamask/perps-controller`:

- **BREAKING:** Added persisted `selectedOrderType`, `orderBookPreferences`, and `visibleCandleCount` fields to `PerpsControllerState`, with controller methods and selectors for updating and reading each preference.
- Changed pending trade configuration restore to last 30 seconds instead of five minutes, include the `reduceOnly` setting, and clear the draft after a successful order while retaining leverage and the selected order type.

## Manual testing steps

```gherkin
Feature: Persist Pro-mode order, chart, and draft preferences

  Scenario: Selected order type is shared across markets
    Given I set the selected order type to "limit" on BTC
    When I switch to ETH
    Then getSelectedOrderType returns "limit"

  Scenario: Order-book display preferences default to USD totals
    Given a controller with no persisted orderBookPreferences
    When I read getOrderBookPreferences
    Then it returns { currency: "usd", metric: "total" }

  Scenario: Visible candle count is clamped and persisted
    Given the default visibleCandleCount is 30
    When I set the visible candle count to 9
    Then getVisibleCandleCount returns 10
    When I set the visible candle count to 251
    Then getVisibleCandleCount returns 250

  Scenario: Pending draft expires after 30 seconds
    Given I save a pending trade configuration for BTC
    When 30 seconds and 1 millisecond elapse
    Then getPendingTradeConfiguration for BTC returns undefined

  Scenario: Draft is cleared after a successful order but leverage and order type are kept
    Given I save a pending trade configuration for BTC with leverage and order type "limit"
    When placeOrder succeeds for BTC
    Then getPendingTradeConfiguration for BTC returns undefined
    And getTradeConfiguration for BTC still returns the saved leverage
    And getSelectedOrderType returns "limit"
```

## Screenshots/Recordings

N/A — controller/state changes only, no UI in this package.

## Checklist

[TAT-3706]: https://consensyssoftware.atlassian.net/browse/TAT-3706
[TAT-3702]: https://consensyssoftware.atlassian.net/browse/TAT-3702
[TAT-3736]: https://consensyssoftware.atlassian.net/browse/TAT-3736

---

> [!NOTE]
> **Medium Risk**
> Adds required persisted fields to PerpsControllerState (breaking for full-state constructors) and changes draft restore/clear behavior around order placement. Runtime rehydration still falls back to defaults.
> 
> **Overview**
> **BREAKING:** Persists market-agnostic Pro preferences on `PerpsControllerState`: `selectedOrderType` (default `market`), `orderBookPreferences` (USD totals), and `visibleCandleCount` (default 30, clamped 10–250). Getters, setters, messenger actions, and selectors are added; missing persisted fields still merge to defaults.
> 
> Pending trade drafts now expire after **30 seconds** (was 5 minutes), include `reduceOnly`, and are cleared after a successful `placeOrder`. Leverage and selected order type are kept. Saving a pending config with `orderType` also updates the global selected type.
> 
> `selectPendingTradeConfiguration` still memoizes the raw draft for stable references, but evaluates TTL on every call so expiry is not frozen by Reselect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 159f11e7e6da0a694fcbb687566c2616064fd288. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


abretonc7s:
## Automated Review — PR #9922

> **BETA** — Automated review from the farmslot pipeline.

| | |
|---|---|
| **Recommendation** | APPROVE |
| **Reviewed commit** | `159f11e7e6da0a694fcbb687566c2616064fd288` |
| **Tier** | standard |
| **Recipe** | N/A |

### Summary


No blocking defect found. The change is coherent, well tested, and the breaking
surface is labelled correctly. Four non-blocking findings and one downstream
follow-up are listed below.

<details>
<summary>Full review details</summary>

## What the PR does

Moves four Pro-mode preferences into `PerpsController`:

- `selectedOrderType` (global, default `market`), also written as a side effect of
  `savePendingTradeConfiguration` when the draft carries an `orderType`.
- `orderBookPreferences` (`{ currency, metric }`, default USD totals).
- `visibleCandleCount` (default 30, clamped 10–250 on write).
- Pending draft TTL 5 min → 30 s, `reduceOnly` added to the draft, and the draft is
  cleared after a successful `placeOrder`.

Correct and worth calling out:

- `PerpsController.ts:1193` shallow-merges defaults into provided state, and every
  getter/selector merges defaults, so older persisted state rehydrates without a
  migration.
- `selectors.ts:170` splits the memoized extractor from the TTL check, so expiry is
  evaluated on every call while the returned object keeps a stable reference. That
  fixes the real bug in the previous `createSelector` shape, where `Date.now()` was
  frozen inside the memo. Both properties are covered by tests
  (`tests/src/selectors.test.ts:487`, `:523`).
- `PerpsController.ts:2657` clears the draft only when `result.success` is true;
  the failure path is tested (`tests/src/PerpsController.configuration.test.ts:1260`).
- `messenger-action-types:check` passes, so the generated action-type file matches
  `MESSENGER_EXPOSED_METHODS`.

## Non-blocking findings

1. **Read paths do not clamp `visibleCandleCount`** —
   `packages/perps-controller/src/PerpsController.ts:5890` and
   `packages/perps-controller/src/selectors.ts:323`.
   The setter clamps to 10–250 and rounds, but both readers only guard against
   non-finite values. Any state that reaches the store by another route (persisted
   state written by an older or third-party client, a direct `update`, a test
   fixture) is returned verbatim — e.g. `visibleCandleCount: 5000` reads back as
   5000 and a chart would try to render it. Clamping in the getter/selector, or
   sharing one `normalizeVisibleCandleCount` helper between setter and readers,
   would make the invariant hold on the read side too.

2. **`setOrderBookPreferences` can write `undefined` over a default** —
   `packages/perps-controller/src/PerpsController.ts:5855`.
   `exactOptionalPropertyTypes` is off in `tsconfig.base.json`, so
   `setOrderBookPreferences({ currency: undefined })` type-checks. The spread stores
   an explicit `currency: undefined`, and because `getOrderBookPreferences` also uses
   a spread (`{ ...DEFAULT, ...state.orderBookPreferences }`), the explicit
   `undefined` wins over the default rather than being repaired. The existing test
   (`tests/src/PerpsController.configuration.test.ts:868`) covers a *missing* key,
   not an `undefined`-valued one. Filtering undefined values in the patch would close
   it.

3. **`selectOrderBookPreferences` returns a fresh object on every call** —
   `packages/perps-controller/src/selectors.ts:300`.
   Mobile's `usePerpsSelector` is plain `react-redux` `useSelector` with reference
   equality, so any component reading this selector re-renders on every dispatched
   action. This mirrors the pre-existing `selectProLayoutPreferences`, so it is not a
   regression, but it sits oddly next to the careful reference-stability work done
   for `selectPendingTradeConfiguration` in this same PR. A memoized selector (or a
   module-level frozen default returned when the slice is absent) would be consistent.

4. **Small test gaps.** No case asserts that `savePendingTradeConfiguration` leaves
   `selectedOrderType` untouched when the draft has no `orderType` — the implicit
   contract at `PerpsController.ts:5654`. No case covers `placeOrder` rejecting
   (throw rather than `success: false`), where the draft is also retained.

## Downstream compatibility (Mobile / Extension)

Clones inspected: `../metamask-mobile-3` @ `b7e4a5198c`, `../metamask-extension-3` @
`eaa84b92d4` (the paths in the checklist, `../metamask-mobile` / `../metamask-extension`,
do not exist on this machine). Full scan in `artifacts/downstream-usage.log`.

- **Exports: additive only.** No export was removed or renamed. `src/constants/index.ts`
  re-exports with `export *`, so `VISIBLE_CANDLE_COUNT_CONFIG`,
  `DEFAULT_ORDER_BOOK_PREFERENCES`, and `DEFAULT_SELECTED_ORDER_TYPE` are reachable
  from `@metamask/perps-controller/constants` without a manifest change.
- **Breaking surface is exactly what the PR claims.** Three new required
  `PerpsControllerState` fields break only code that builds a fully-typed state
  literal: `app/components/UI/Perps/__mocks__/serviceMocks.ts:114` and
  `app/core/Engine/controllers/perps-controller/index.test.ts:109` in Mobile. The
  five literals in `app/components/UI/Perps/hooks/usePerpsSelector.test.ts` use
  `as PerpsControllerState` casts and are unaffected. Extension uses only
  `Partial<PerpsControllerState>` and its own `ui/selectors/perps-controller.ts`, so
  it is unaffected.
- **`selectPendingTradeConfiguration` is no longer a reselect selector.** It is now a
  plain function, so `.resultFunc` / `.recomputations()` and friends are gone from the
  type. Neither clone uses them (Mobile calls it directly at
  `usePerpsOrderForm.ts:118` and `useInitPerpsPaymentToken.ts:17`; Extension does not
  import it), so nothing breaks today — but it is a public-shape change that is not in
  the changelog and would be worth one line there.
- **Six new messenger methods are additive.** Extension's
  `app/scripts/messenger-client-init/perps-controller-init.ts` maps methods
  explicitly and has no parity assertion against `MESSENGER_EXPOSED_METHODS`, so it
  keeps compiling.
- **TTL 5 min → 30 s is user-visible in Mobile.** `useInitPerpsPaymentToken` restores
  the draft's `selectedPaymentToken` after the user returns from token selection, and
  `usePerpsSavePendingConfig` saves on blur/unmount. Any detour longer than 30 s
  (token picker, a deposit, an app backgrounding) now drops the draft where it used to
  survive. The PR frames this as intentional; confirm with TAT-3706 that 30 s covers
  the pay-with-token round trip. Mobile's own comments still say "5 minutes"
  (`usePerpsSavePendingConfig.ts:9`, `usePerpsOrderForm.ts:116`) — doc drift to fix on
  the client side.
- **The new clear-on-success may be a no-op in Mobile's Lite order view.**
  `PerpsOrderView.tsx:1578` already clears the draft after a successful submit but does
  not reset `orderForm`; `usePerpsSavePendingConfig` then re-saves a draft on
  blur/unmount with the un-reset values. The Pro form resets the form first
  (`usePerpsProOrderForm.ts:1444`) and is fine. Not a Core defect — the controller-side
  clear is correct — but the stated benefit won't land in the Lite view without a
  client change.

## Validation

| Command | Result |
| --- | --- |
| `yarn workspace @metamask/perps-controller run jest --no-coverage tests/src/PerpsController.configuration.test.ts tests/src/selectors.test.ts` | 2 suites, 124 tests passed |
| `yarn workspace @metamask/perps-controller run jest --no-coverage tests/src/PerpsController.trading.test.ts tests/src/PerpsController.state.test.ts` | 2 suites, 66 tests passed |
| `yarn workspace @metamask/perps-controller run messenger-action-types:check` | up to date |
| `yarn workspace @metamask/perps-controller run changelog:validate` | pass (no output) |
| `yarn eslint` on the six changed source/test files | clean |

Not run, and why: no `packages/*/dist` exists in this checkout, so a build would
require the ~100 s root `yarn build` across 98 packages. The execution contract for
this run is `static-code` depth, and the diff changes no package export map, build
wiring, or generated output — only additive type/const exports from modules that are
already exported. Type coverage is therefore asserted via Jest (ts-jest/Babel path)
and ESLint rather than an emitted build.

Changelog: both entries are present, PR-linked, and the breaking one is marked
`**BREAKING:**` and ordered first in `### Added`. Release risk is a major bump for
`@metamask/perps-controller`, which the entry already signals.

</details>

abretonc7s (approved):
Automated review — see comment above for full details.

PR #9927, opened 2026-08-20
feat: send the MetaMask auth token to built-in Infura endpoints

basgys:
## Explanation

`NetworkController` presents the MetaMask auth token as `Authorization: Bearer` on requests to its built-in Infura endpoints, reading it through `AuthenticationController:getBearerToken` on its own messenger. Infura accepts the token alongside the v3 project ID in the URL.

**Built-in Infura endpoints only.** The token goes on the primary endpoint of `infura`-type clients, the ones built from `infuraProjectId`. Custom endpoints never present it, even on an `infura.io` host: that is someone else's key. Failover endpoints never present it: other providers. The gate is client type plus position in the chain, so `rpcFailoverMode: 'forced'` sends the token nowhere, since that mode leaves no Infura primary.

**Analytics opt-in only.** The token identifies the user, so `NetworkController` reads `AnalyticsController:getState` and presents it only while `optedIn` is `true`. The rule sits in the controller, not in a constructor option, so no client can redefine or forget it.

**Read per request.** A refreshed token is used on the next request. If the read throws, the request goes out without the header. That covers the locked state, where `getBearerToken` throws until unlock, and clients that register no `AuthenticationController`, such as `wallet-cli`.

## Breaking change

`NetworkControllerMessenger` now requires `AuthenticationController:getBearerToken` to be delegated from the root messenger. `@metamask/wallet` is the only consumer that constructs `NetworkController`, and it delegates the action here.

## Gate on client type, not URL

The project ID ships inside the built extension, so anyone can read it, and `wallet_addEthereumChain` lets a dApp propose an endpoint URL. Matching `/v3/<infuraProjectId>` would hand the user's identity token to any URL containing that string. Doing it safely means pinning the host and checking the subdomain against a hand-maintained allowlist, as `app/scripts/migrations/157.ts` does for failover URLs. Client type asserts the same thing without the allowlist.

Chains that should present the token belong in `InfuraNetworkType`. Linea, Base, Arbitrum, BSC, Optimism, Polygon and Monad already resolve as `infura`-type endpoints and receive it. The extension's `FEATURED_RPCS` still hardcodes `RpcEndpointType.Custom`, leaving Avalanche, zkSync Era and Sei without it: a fix to those network definitions, not to this gate.

## References

None.

## Checklist

  - Breaking for `@metamask/network-controller`, resolved here for `@metamask/wallet`, its only consumer that constructs the controller. No client constructs it directly, so no client draft PR is needed.

---

> [!NOTE]
> **High Risk**
> Sends user-identifying bearer tokens on RPC traffic and adds a breaking messenger dependency on authentication; incorrect scoping could leak tokens to non-MetaMask Infura URLs.
> 
> **Overview**
> **NetworkController** now attaches the MetaMask auth token as `Authorization: Bearer` on HTTP requests to **built-in Infura** RPC endpoints (primary endpoint of `infura`-type clients only—not custom URLs or failover hosts).
> 
> The token is fetched per request via **`AuthenticationController:getBearerToken`** (typed through **`@metamask/profile-sync-controller`**) and only when **`AnalyticsController`** reports **`optedIn`**. If token retrieval fails or returns nothing (e.g. wallet locked), RPC proceeds without the header.
> 
> **Breaking:** **`NetworkControllerMessenger`** must delegate **`AuthenticationController:getBearerToken`**; **`@metamask/wallet`** wiring and tests are updated accordingly. New integration tests cover header behavior, opt-in gating, and endpoint-type restrictions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5438094149f768fdfc826f714f8c7f1e1622739. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet/src/initialization/instances/network-controller/types.ts
```diff
@@ -19,4 +19,13 @@ export type NetworkControllerInstanceOptions = {
...
   analyticsOptions?: NetworkControllerAnalyticsOptions;
+  /**
+   * Returns an authentication token to present as a bearer credential on
+   * requests to Infura endpoints, or undefined if none is available.
+   *
+   * Optional; when omitted, no credential is presented. Requests to other
+   * endpoints are unaffected either way, as are requests that already carry an
+   * `Authorization` header.
+   */
+  getInfuraAuthToken?: () => Promise<string | undefined>;
```
Gudahtt: Is this token the same as the one returned by `AuthenticationController:getBearerToken`? If so, can we use that action instead of a constructor param?

basgys: Yes, it is the same. I reworked the PR to have this logic:

```ts
getInfuraAuthToken: async () =>
        messenger.call('AnalyticsController:getState').optedIn
          ? await messenger.call('AuthenticationController:getBearerToken')
          : undefined,
```

packages/wallet/src/initialization/instances/network-controller/network-controller.ts
```diff
@@ -18,9 +31,15 @@ export const networkController: InitializationConfiguration<
       infuraProjectId: options.infuraProjectId,
       failoverUrls: options.failoverUrls,
       analyticsOptions: options.analyticsOptions,
+      // The token identifies the user, which is only allowed for analytics
+      // opt-ins.
+      getInfuraAuthToken: async () =>
```
Gudahtt: Thoughts on moving this _inside_ of the NetworkController, rather than passing it in as a constructor parameter?

For context, our intention with the messenger is to use it for all communication between wallet components, reserving constructor parameters just for platform dependencies that haven't been migrated to a service yet. Eventually we hope to eliminate all (or nearly all) constructor parameters in favor of messenger actions/events.

Also, the messenger is meant to represent the full set of wallet capabilities that a particular component has. This strategy of _extending_ the capabilities of the messenger just for initialization creates a misleading perception of the capabilities of the NetworkController (it will _appear_ to not have the `AuthenticationController:getBearerToken` capability, but in reality it will), which is definitely not something we want.

basgys: @Gudahtt This means the analytics opt-in gate needs to move into the `NetworkController`, or do you have a different strategy in mind? 

Let me know what you would like to see implemented conceptually, and I will follow your recommendations. I don't know the codebase well enough to have a long-term vision.

Gudahtt: In what you're proposing, the opt-in gate is effectively already a part of the NetworkController. This is the only place it's constructed, and you've made it a required constructor parameter.

I don't think there's any benefit to disguising it as a constructor parameter when it's not optional, nor can I see a benefit of pretending the messenger instance is narrower than it is.

basgys: I reworked the PR. Let me know what you think.

packages/network-controller/CHANGELOG.md
```diff
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - The token is retrieved through `AuthenticationController:getBearerToken`. Because it identifies the user, it is only presented when the user has opted in to analytics (`AnalyticsController` state `optedIn`).
+  - Only endpoints of type `infura` (those reached through `infuraProjectId`) present the token. Custom RPC endpoints and failover endpoints never do, even when hosted by Infura, so a user-supplied Infura key is never paired with it.
+  - The token is read once per request, so a refreshed token is picked up on the next request. If the call throws, which it does while the wallet is locked, the request is made without the credential.
+- Add `@metamask/profile-sync-controller` as a dependency at `^29.0.0` ([#9927](https://github.com/MetaMask/core/pull/9927))
+
 ### Changed
 
+- **BREAKING:** `NetworkControllerMessenger` now requires the `AuthenticationController:getBearerToken` action to be delegated from the root messenger ([#9927](https://github.com/MetaMask/core/pull/9927))
+  - Clients that do not delegate it present no token: the call throws, and the request is made without the credential.
```
Gudahtt, line 21: Nit: While true, this is not necessary to mention. We can assume all required actions are delegated; it'll create a type error otherwise.

packages/network-controller/src/create-network-client.ts
```diff
@@ -353,6 +405,14 @@ function createRpcServiceChain({
         ...(overriddenOptions.policyOptions ?? {}),
       },
       ...overriddenOptions,
+      ...(isBuiltInInfuraEndpoint && getInfuraAuthToken
+        ? {
+            fetch: withInfuraAuthToken(
```
Gudahtt, line 410: We have a `fetchOptions` parameter that can be used to set headers, we don't need to create a wrapper around `fetch`

Gudahtt: Oh, but for `fetchOptions` we'd need to have synchronous access to the token. Hmm.

packages/network-controller/CHANGELOG.md
basgys: Removed. Confirmed the type system catches it: `Messenger.ts:136` folds missing delegations into `__MISSING_DELEGATIONS__`, so omitting the action fails to compile rather than degrading at runtime.

Kept the runtime note in the `getInfuraAuthToken` JSDoc, but corrected it to the case that is not type-checked: a client that delegates the action but registers no `AuthenticationController`, as `wallet-cli` does. That throws and the request goes out without the credential.

basgys: I removed it.

packages/network-controller/src/create-network-client.ts
basgys: What would you suggest?

Gudahtt: You could get the token in the `init` function of `NetworkController`, and pass it into the network clients from there

Gudahtt: Actually that might not work, since we may not be signed in at that point. Maybe the way you've implemented it is the best we can do with the current AuthController API

packages/network-controller/tests/NetworkController.test.ts
```diff
@@ -997,6 +997,62 @@ describe('NetworkController', () => {
     });
   });
 
+  describe('Infura auth token', () => {
```
mcmire, line 1000: If we move `getInfuraAuthToken` to `create-network-client.ts` then there isn't much that `NetworkController` is doing, so maybe we don't need these tests at all?

packages/network-controller/src/NetworkController.ts
```diff
@@ -2865,6 +2885,7 @@ export class NetworkController extends BaseController<
           },
           getRpcServiceOptions: this.#getRpcServiceOptions,
           getBlockTrackerOptions: this.#getBlockTrackerOptions,
+          getInfuraAuthToken: this.#getInfuraAuthToken,
```
mcmire, line 2888: Since `#getInfuraAuthToken` simply accesses the messenger and we already pass the messenger to `createAutoManagedNetworkClient`, do we need `#getInfuraAuthToken` here at all? Maybe we can move `getInfuraAuthToken` to `create-network-client.ts`.

packages/network-controller/src/create-network-client-tests/infura-auth-token.test.ts
```diff
@@ -0,0 +1,322 @@
...
+ * header of every request before forwarding it to the global `fetch`.
+ *
+ * @returns The recording `fetch` along with the list it records into.
+ */
+function buildRecordingFetch(): {
+  requests: RecordedRequest[];
+  recordingFetch: typeof fetch;
+} {
+  const requests: RecordedRequest[] = [];
+  const recordingFetch: typeof fetch = async (input, init) => {
```
mcmire, line 42: I realize it's not obvious here, but we are using Nock under the hood to mock these requests. There is a way we can make use of that fact instead of handrolling a way to capture requests. However, that would take more code, and I know you are trying to get this merged. So this is fine in the meantime.

packages/network-controller/src/create-network-client.ts
```diff
@@ -221,6 +226,44 @@ export function createNetworkClient({
...
+
+    if (!token) {
+      return await fetchImplementation(input, init);
+    }
+
+    return await fetchImplementation(input, {
+      ...init,
+      // These headers can be spread safely because `RpcService` builds them by
+      // deep-merging plain objects, so they never arrive as a `Headers`
+      // instance or as a list of pairs.
```
mcmire, line 261: Nit: Maybe we don't need this comment? I don't see why it wouldn't be safe, given that we don't need a type assertion.

``` suggestion
```

PR #9928, opened 2026-08-20
feat: add `config-registry-{controller, api-service}` to `wallet`

mikesposito:
## Explanation

Adding `@metamask/config-registry-controller` to the `wallet` package.

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WPN-1894

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking constructor API plus new remote config fetch on wallet startup. Duplicate messenger registration can fail existing consumers that already wire these controllers.
> 
> **Overview**
> **BREAKING:** Default wallet init now constructs `ConfigRegistryApiService` and `ConfigRegistryController`, so remote config registry is part of the core wallet graph.
> 
> `instanceOptions.configRegistryApiService` is **required** (`env`, optional `fetch` / `policyOptions`). `configRegistryController` is optional (`pollingInterval`, `fallbackConfig`). The controller messenger delegates keyring, remote feature flags, and `ConfigRegistryApiService:fetchConfig`.
> 
> Consumers that already register these on a custom root messenger must drop their own wiring or registrations will collide. `wallet-cli` passes `ConfigRegistryApiEnv.PRD`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4777e5348edf58cbe73ea6b340ed16484fb59dc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/wallet-cli/CHANGELOG.md
+++ b/packages/wallet-cli/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Wire the `configRegistryApiService` slot in the daemon wallet's instance options with `env: ConfigRegistryApiEnv.PRD`, required now that `@metamask/wallet` makes this slot mandatory ([#9928](https://github.com/MetaMask/core/pull/9928))
 - Add the `mm wallet send` command and a dedicated daemon `sendTransaction` RPC handler for sending a transaction through the daemon-hosted `TransactionController` ([#9636](https://github.com/MetaMask/core/pull/9636))
   - The command converts the ether `--value` to wei, resolves the network client (`--network-client-id` or `--chain-id`) and sender (defaulting to the selected account), previews the resolved plan, and broadcasts after confirmation, printing the resulting transaction hash. `--yes` skips the prompt; `--dry-run` resolves and validates without broadcasting.
   - The `sendTransaction` handler awaits the broadcast server-side and returns a serializable `{ transactionHash, transactionId, status }`, because `addTransaction`'s `result` promise cannot travel back over the generic `call` dispatch. Transactions are submitted as internal (auto-approved by the daemon).
--- a/packages/wallet-cli/CHANGELOG.md
+++ b/packages/wallet-cli/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Wire the `configRegistryApiService` slot in the daemon wallet's instance options with `env: ConfigRegistryApiEnv.PRD`, required now that `@metamask/wallet` makes this slot mandatory ([#9928](https://github.com/MetaMask/core/pull/9928))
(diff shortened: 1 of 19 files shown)


packages/wallet/src/initialization/instances/config-registry-api-service/config-registry-api-service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const configRegistryApiService: InitializationConfiguration<
+  ConfigRegistryApiService,
+  ConfigRegistryApiServiceMessenger
+> = {
+  name: 'ConfigRegistryApiService',
+  init: ({ messenger, options }) =>
+    new ConfigRegistryApiService({
+      messenger,
+      env: options.env ?? ConfigRegistryApiEnv.PRD,
```
mcmire: Is it intentional to always hit prod, even locally? Why not require clients to pass this in instead of having the Wallet Library default this?

mikesposito: The idea was to preserve the optionality, but we can make it a mandatory option

mikesposito: Updated in [4f4ccda](https://github.com/MetaMask/core/pull/9928/commits/4f4ccda7231af3d89f15e6a614e7673918572fbe)

packages/wallet-cli/src/daemon/wallet-factory.ts
```diff
@@ -94,6 +95,9 @@ function buildInstanceOptions(
   infuraProjectId: string,
 ): WalletOptions['instanceOptions'] {
   return {
+    configRegistryApiService: {
+      env: ConfigRegistryApiEnv.PRD,
```
mcmire, line 99: I don't know if I like this either, but at least it's consistent with the defaults `wallet-cli` uses for other controllers/services (i.e. it uses prod for other APIs).

packages/wallet-cli/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Wire the `configRegistryApiService` slot in the daemon wallet's instance options with `env: ConfigRegistryApiEnv.PRD`, required now that `@metamask/wallet` makes this slot mandatory ([#9928](https://github.com/MetaMask/core/pull/9928))
```
mcmire: Does this change anything for end users? If `wallet-cli` is not using `ConfigRegistryApiController` or `ConfigRegistryApiService` for anything then I wonder if we can leave this out.

``` suggestion
```

I guess we may have to use the `no-changelog` label if the changelog checker is complaining about this.

mcmire (commented):
One more thing.

packages/wallet-cli/CHANGELOG.md
mikesposito: Good point, I'm applying this change and the label

mcmire (approved):
Looks good!

PR #9940, opened 2026-08-24
fix(assets-controller): skip redundant Accounts API refetch when WebSocket already confirmed the transaction

juanmigdr:
## Explanation

`AssetsController` runs a forced Accounts API balance refetch every time a transaction is confirmed, even when `AccountActivityService`'s WebSocket already delivered the same balance update for that chain. That's an unnecessary HTTP call whenever the WebSocket connection is healthy.

This change makes `AssetsController` check whether the transaction's chain is WS-active, and if so, wait briefly for a matching `AccountActivityService:transactionUpdated` event (correlated by transaction hash) before forcing the API refetch. If the WebSocket confirms in time, the API call is skipped; if it doesn't (WS down, message dropped, etc.), the API call still happens as a fallback, so behavior is unchanged when the WebSocket isn't available.

One thing worth calling out for reviewers: `transactionConfirmed` and the WebSocket's `transactionUpdated` push are two independent signals that can arrive in either order — the WebSocket message frequently arrives *before* `transactionConfirmed` fires. The implementation accounts for this by tracking whichever one arrives first per transaction, so it works correctly regardless of ordering.

**Breaking change:** `AssetsController` now subscribes to `AccountActivityService:transactionUpdated`. Any host that restricts which events flow through the `AssetsController` messenger (e.g. via a scoped delegation list) must add this event, or `AssetsController` will never see it and will always fall back to the API call.

## References

- Mobile client adoption: https://github.com/MetaMask/metamask-mobile/pull/35202
- Extension client adoption: https://github.com/MetaMask/metamask-extension/pull/45718
- Ticket: https://consensyssoftware.atlassian.net/browse/ASSETS-3910 & https://consensyssoftware.atlassian.net/browse/ASSETS-3916

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes post-confirmation balance refresh timing and adds a breaking messenger subscription; stale balances are possible only if WS falsely signals confirmation and the API fallback is skipped, but timeout fallback preserves prior behavior when WS is unreliable.
> 
> **Overview**
> After a transaction confirms, **AssetsController** no longer always forces an Accounts API balance refetch. When the chain is WebSocket-active and the confirmation includes a **hash**, it waits up to **1.5s** for a matching **`AccountActivityService:transactionUpdated`** event; if that arrives first (including when WS fires before **`transactionConfirmed`**), the HTTP refetch is skipped.
> 
> If the chain is not WS-active, the hash is missing, or the WS signal never matches in time, behavior is unchanged: **`getAssets`** still runs with **`forceUpdate: true`**. Coordination uses a per-transaction map with cleanup on **`destroy`**.
> 
> **Breaking:** the controller now subscribes to **`AccountActivityService:transactionUpdated`**; hosts that scope messenger events must delegate it or the optimization never applies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a516b09c374d948138d95759051228c88a352b99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/CHANGELOG.md
+++ b/packages/assets-controller/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Skip the forced Accounts API balance refetch on `TransactionController:transactionConfirmed` when `AccountActivityService` (WebSocket) already reports the transaction's chain as active: `AssetsController` now waits up to 1.5s for a matching `AccountActivityService:transactionUpdated` event (correlated by transaction hash) before falling back to the API call, avoiding a redundant HTTP request when the WebSocket already delivered the same balance update
+- **BREAKING:** Subscribe to `AccountActivityService:transactionUpdated` (typed as `AccountActivityServiceTransactionUpdatedEvent`) to skip the forced Accounts API balance refetch on `TransactionController:transactionConfirmed` when `AccountActivityService` (WebSocket) already reports the transaction's chain as active and confirms the same transaction (correlated by transaction hash) within 1.5s of the confirmation, avoiding a redundant HTTP request when the WebSocket already delivered the same balance update ([#9940](https://github.com/MetaMask/core/pull/9940))
+  - Hosts that restrict which events flow through the `AssetsController` messenger must now also delegate `AccountActivityService:transactionUpdated`
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


juanmigdr:
Superseeded by https://github.com/MetaMask/core/pull/10030#pullrequestreview-5069046795

PR #9943, opened 2026-08-24
fix: only scan send payees for address poisoning

adonesky1:
## Explanation

Restrict address poisoning known recipients to user-chosen send payees so confirmed approves, swaps, and contract calls stop putting token and protocol addresses into the comparison set.

### Summary

Export `getSendRecipients` and hydrate the phishing known-recipient set from it instead of `getEffectiveRecipient`, which falls back to `txParams.to` for every non-transfer type.

### Problem

Address poisoning is a payee mixup: the user previously sent to Alice, an attacker grinds a lookalike EOA, and the user pastes from history. We were also storing token, Permit2, router, and other contract `to` values from confirmed approves and contract interactions. Vanity token factories (shared prefix plus suffix) then 4+4-matched each other, which is what fired on the o1.exchange Permit2 approve: on Base, dogue (`0xb2000000000000000000003779298b7eB3D8d501`) matched Base Uncle (`0xb200000000000000000000c02ce1aA07c9E3d501`) at 22 prefix characters plus the shared suffix `d501`.

`getEffectiveRecipient` is the right helper for first-time interaction (the contract `to` is who we are calling). It is the wrong helper for poisoning.

### Solution

Add `getSendRecipients` next to `getEffectiveRecipient` and use it only for poisoning hydration.

Included:

- `simpleSend` `to`, preferring `txParamsOriginal` when present
- Decoded `to` / `_to` for the three transfer methods: `transfer`, `transferFrom`, `safeTransferFrom` (no fallback to the token contract)
- `swapAndSendRecipient`
- Nested batch calls that themselves are sends or token transfers
- Untyped transactions with no calldata, treated as legacy native sends

Not included: approves and Permit2, `swap` and `swapApproval`, bridges, staking, and generic contract calls. Note `swapAndSend` is deliberately not in that list, since its recipient is entered by the user.

Also not included, and worth being explicit about: ERC-1155 `safeBatchTransferFrom`. There is no `TransactionType` for it, so it classifies as `contractInteraction` and its payees are never scanned. That is pre-existing behaviour inherited from `getEffectiveRecipient` rather than something this PR changes, but it does mean "ERC-1155 transfers are covered" is only true of the single-transfer method.

`getEffectiveRecipient` is unchanged so first-time interaction keeps using the contract `to`.

Two deliberate details a reviewer might question:

- `txParamsOriginal` is swapped in as a whole object rather than field by field, unlike `useTransferRecipient` in the clients. It is a `cloneDeep` snapshot, so its `to` and `data` always describe the same call. Mixing an original `to` with a wrapped `data` would misread an untyped transaction as a contract call and drop a real payee.
- `swapAndSendRecipient` is added unconditionally rather than gated on `TransactionType.swapAndSend`. It is only ever populated from the payee the user entered in the swap-and-send flow, and gating on the type as well would be dead code.

Follow-up client PR: https://github.com/MetaMask/metamask-extension/pull/45724

### Risk

- Existing known sets that currently contain token/router addresses will drop those entries on next hydration. `#knownRecipients` is in-memory and rebuilt in the constructor, so this self-heals on the next controller construction with no migration.
- A later send to a lookalike EOA will no longer match against a token the user only approved, which is correct.
- Clients still on the old phishing-controller keep the old known set until they bump. Extension pins `^17.4.0` and mobile pins `^17.3.1`, so both need a bump PR after this releases. Until then the extension fix is candidate-side only.

### Verification

Unit coverage in this PR:

- `getSendRecipients` returns `[]` for `tokenMethodApprove` and `contractInteraction`, and returns the decoded payee (not the token contract) for the three transfer types
- `PhishingController` does not add token contracts from confirmed approve transactions, and hydrates `swapAndSendRecipient` rather than the swap contract

#### Manual verification of the known-set change

The false positive in the report is fixed on the extension side alone, because an approve confirmation stops producing a candidate at all. That means the extension PR's manual steps cannot observe this PR's change: with no candidate, the known set is never consulted. To see the known-set cleanup you have to probe it from the **send** side.

### Setup

Needs a preview build of this package, since the extension pins `@metamask/phishing-controller@^17.4.0`.

In `metamask-extension`, on [#45724](https://github.com/MetaMask/metamask-extension/pull/45724), add a top-level `previewBuilds` block to `package.json` and run `yarn install`:

```json
"previewBuilds": {
  "@metamask/phishing-controller": {
    "type": "non-breaking",
    "previewVersion": "17.4.0-preview-878689a26"
  },
  "@metamask/transaction-controller": {
    "type": "non-breaking",
    "previewVersion": "69.6.1-preview-878689a26"
  }
}
```

Both are needed: the preview `phishing-controller` depends on `transaction-controller@^69.6.1`, which the extension's `^69.5.1` pin does not satisfy, so without the second entry you get a second copy on disk. Confirm the install with `yarn why @metamask-previews/transaction-controller`, not `yarn why @metamask/transaction-controller`; once remapped, the copies live under the preview scope. Do not commit the `package.json` or `yarn.lock` changes, and do not delete the vendored `getSendRecipients` in that PR, which still has to build against the published `^69.5.1`.

Then `yarn dist` and load the build. Everything below runs in the page console at https://metamask.github.io/test-dapp/ with the account connected.

### Helpers, paste once

Assigns to `window` so it survives separate pastes. Pasting a bare `const from = ...` in one block and using it in the next is what produces `from is not defined`.

```js
window.mm = {
  // approve(0x000000000022D473030F116dDEE9F6B43aC78BA3, 0)
  APPROVE_PERMIT2_ZERO:
    '0x095ea7b3000000000000000000000000000000000022d473030f116ddee9f6b43ac78ba30000000000000000000000000000000000000000000000000000000000000000',
  async tx(params) {
    const chainId = await ethereum.request({ method: 'eth_chainId' });
    if (chainId !== '0x2105') {
      throw new Error(`Switch MetaMask to Base. Current chainId: ${chainId}`);
    }
    const [from] = await ethereum.request({ method: 'eth_requestAccounts' });
    const hash = await ethereum.request({
      method: 'eth_sendTransaction',
      params: [{ from, ...params }],
    });
    console.log('submitted', hash);
    return hash;
  },
};
console.log('ready');
```

### 1. Seed the known set

Confirms an `approve(Permit2, 0)` on Base Uncle. On the published `phishing-controller` this puts the token contract into the known set; with this PR it does not. Approving `0` needs no token balance.

```js
mm.tx({
  to: '0xb200000000000000000000c02ce1aA07c9E3d501',
  data: mm.APPROVE_PERMIT2_ZERO,
});
```

Confirm it and wait for confirmed status in Activity. `#getRecipientAddressesFromTransaction` returns `[]` for anything not yet confirmed, so running step 2 while this is pending gives a false pass.

Expect a trust-signal alert on this confirmation regardless of this PR: the address scan returns `Warning` for Base Uncle. Separate detector, unrelated.

### 2. Probe the known set from the send side

`0xb20011…d501` scores prefix 4 / suffix 4 against the Base Uncle contract, exactly the detector's threshold. A dapp-initiated native send with no calldata types as `simpleSend`, so it does produce a candidate and the known set is consulted.

```js
mm.tx({
  to: '0xb20011111111111111111111111111111111d501',
  value: '0x0',
});
```

**Expected: no address poisoning warning.** Cancel rather than confirming, so the probe address stays out of the known set.

### BEFORE:

image (6)

### AFTER: using the same address [from the original report](https://consensys.slack.com/archives/C07BE24B8AC/p1787540982398789)

Video depicting using the same contract address with no address poisoning detection when compared against a similar looking contract:
https://github.com/user-attachments/assets/b1688fa2-2d2d-4868-9f70-eef3d44c89b7

To see the failing side, remove the `previewBuilds` block, `yarn install`, rebuild, and repeat steps 1 and 2. The published `^17.4.0` warns here, comparing the send address against the token contract the approve stored.

### 3. Control: the known set still holds real payees

Confirm this one, wait for confirmed status:

```js
mm.tx({
  to: '0x1234000000000000000000000000000000009abc',
  value: '0x0',
});
```

Then submit the lookalike and **cancel** it:

```js
mm.tx({
  to: '0x1234ffffffffffffffffffffffffffffffff9abc',
  value: '0x0',
});
```

**Expected: the address poisoning warning appears.** This must hold with the preview installed. If step 2 and step 3 both come back silent, the detector is broken rather than the fix working.

### Real poisoning for entered native send poisoning still working as expected
https://github.com/user-attachments/assets/73ed360f-898c-4557-b497-8088e7dfc2c1

Step 2 is the only check that distinguishes this PR from the extension PR. Note the known set is rebuilt in the `PhishingController` constructor from confirmed transaction state, so there is nothing to clear after swapping builds: restarting the extension re-derives it.

## References

- Ticket: https://consensyssoftware.atlassian.net/browse/PSAFE-633
- Slack: https://consensys.slack.com/archives/C07BE24B8AC/p1787540982398789
- Extension confirmation gate: https://github.com/MetaMask/metamask-extension/pull/45724

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how address-poisoning known recipients are derived from transaction history, which can reduce false positives but also narrows the comparison set; real send-based poisoning detection should remain covered by tests and address book entries.
> 
> **Overview**
> Adds and exports **`getSendRecipients`** in `@metamask/transaction-controller` so callers can resolve **user-chosen payees** (native sends, decoded ERC-20/721/1155 transfers including batch, `swapAndSendRecipient`, nested batch sends) while **excluding** approves, swaps, and generic contract `to` addresses. **`getEffectiveRecipient`** is unchanged for “who we’re calling” flows; docs now steer poisoning checks toward the new helper.
> 
> **`PhishingController`** builds its in-memory known-recipient set from **`getSendRecipients`** instead of **`getEffectiveRecipient`**, so confirmed token approves and similar txs no longer seed lookalike comparisons against token/router contracts. Invalid hex recipients are still dropped via existing normalization.
> 
> Tests cover **`getSendRecipients`** edge cases (cancellations, speed-ups, `txParamsOriginal`, batches) and phishing hydration (approve txs, swap-and-send payee vs swap contract).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5a186b983613384ddc49853c4da9e7d71fdf840. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


adonesky1:
@metamaskbot publish-previews

packages/transaction-controller/src/utils/recipient.test.ts
```diff
@@ -106,3 +106,211 @@ describe('getEffectiveRecipient', () => {
     expect(getEffectiveRecipient(transactionMeta)).toBe(TOKEN_CONTRACT);
   });
 });
+
+describe('getSendRecipients', () => {
```
imblue-dabadee, line 141: Perhaps we add test cases for ERC721 and ERC1155 as well?

packages/phishing-controller/src/PhishingController.ts
```diff
@@ -962,18 +962,11 @@ export class PhishingController extends BaseController<
...
     }
 
-    const transactionRecipient = this.#normalizeAddress(
-      getEffectiveRecipient(transaction),
-    );
-    const swapAndSendRecipient = this.#normalizeAddress(
-      transaction.swapAndSendRecipient,
-    );
-
     return Array.from(
```
imblue-dabadee, line 965: Nit: `getSendRecipients` dedupes so don't need this anymore and can just return the new function.

imblue-dabadee (dismissed):
2 comments but LGTM!

packages/transaction-controller/src/utils/recipient.ts
```diff
@@ -33,3 +43,130 @@ export function getEffectiveRecipient(
...
+  type,
+}: SendRecipientSource): string | undefined {
+  if (isNativeSendType(type, data)) {
+    return to;
+  }
+
+  if (type && TOKEN_TRANSFER_TYPES.includes(type) && hasCalldata(data)) {
+    return decodeTokenTransferRecipient(data);
+  }
+
```
imblue-dabadee, line 158: To get ERC1155 transfers, something cheap like this should work:
```suggestion
  if (
    hasCalldata(data) &&
    data.toLowerCase().startsWith(ERC1155Interface.getSighash('safeBatchTransferFrom'))
  ) {
    return decodeTokenTransferRecipient(data);
  }
```

adonesky1: Added `safeBatchTransferFrom` handling with top-level and nested coverage in `3f7b513ad2`, and mirrored it in Extension in `bf61b3c14b6`.

packages/transaction-controller/src/utils/recipient.test.ts
adonesky1: Added explicit ERC-721 and ERC-1155 `safeTransferFrom` coverage in `9fdd24dfc7`.

packages/transaction-controller/src/utils/recipient.ts
```diff
@@ -33,3 +45,146 @@ export function getEffectiveRecipient(
...
+  if (
+    hasCalldata(data) &&
+    data.slice(0, 10).toLowerCase() ===
+      ERC1155_SAFE_BATCH_TRANSFER_FROM_SELECTOR
+  ) {
+    return decodeTokenTransferRecipient(data);
+  }
```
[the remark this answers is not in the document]
imblue-dabadee: Seems legit given the comment here on line 108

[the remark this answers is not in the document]
imblue-dabadee:
```suggestion
    const effectiveType = getEffectiveType(
      transactionMeta.type,
      transactionMeta.originalType,
    );
     if (effectiveType === TransactionType.cancel) {
      return [];
    }
  }
```

Something like this (but at the beginning of the func) should rectify.

imblue-dabadee (dismissed):

packages/transaction-controller/src/utils/recipient.ts
```diff
@@ -33,3 +45,155 @@ export function getEffectiveRecipient(
...
+
+  addRecipient(
+    getSendRecipientFromSource({
+      data: params?.data,
+      to: params?.to,
+      type: getEffectiveType(
+        transactionMeta.type,
+        transactionMeta.originalType,
+      ),
+    }),
```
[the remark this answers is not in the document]
adonesky1: Fixed in [994d977712](https://github.com/MetaMask/core/commit/994d9777120628fbd7a0bc3a46bd8cd9c3d716fc). `getSendRecipients` now infers recognized transfer methods only when it uses `txParamsOriginal`, with regression coverage for ERC-20, ERC-721, and ERC-1155 while approvals remain excluded.

imblue-dabadee (dismissed):

jpuri (approved):

mcmire (approved):
LGTM.

mindofmar (approved):

PR #9947, opened 2026-08-24
feat(bridge-controller): add SwapBridge failure telemetry schema and classifiers

Battambang:
## Explanation

**Problem.** SwapBridge `Failed` today is one bucket. Mixpanel cannot tell quote-fetch vs submit vs source vs dest vs poll, or whether a source/dest hash existed. That hides Tron/non-EVM diagnosis behind a blended "on-chain" failure rate. Free-text `error_message` is not a reliable classifier.

This PR adds the **schema and classifiers only**. It does **not** change Mixpanel payloads - runtime analytics behavior is unchanged until the follow-up wires these into emit paths.

Classification is always derived from the failing code path and hash presence, never from `error_message` text.

### Content

Shared types live in bridge-controller: optional *Data for Mixpanel event context, and required *Properties as the classifier return shape. bridge-status-controller imports those types rather than redefining them.

| Package | Added |
|---|---|
| `@metamask/bridge-controller` | `FailurePhase`, `SwapBridgeErrorCode` enums · `HashPresenceData`, `FailureTelemetryData` · `HashPresenceProperties`, `FailureTelemetryProperties` . `getQuoteFetchErrorCode` |
| `@metamask/bridge-status-controller` | `getSubmitErrorCode` · `getHashPresenceProperties` · `getStatusFailurePhase` · `getSubmitFailureTelemetry` · `getStatusFailureTelemetry`|

### Event context types (all optional, additive)

| Event | New optional fields |
|---|---|
| Quotes Error | `failure_phase`, `error_code` |
| Failed | `failure_phase`, `error_code`, `source_hash_present`, `destination_hash_present` |
| Submitted | `source_hash_present`, `destination_hash_present` |
| Completed | `source_hash_present`, `destination_hash_present` |

### Property values

| Property | Values |
|---|---|
| `failure_phase` | `quote` · `broadcast` · `source_execution` · `destination_execution` · `poll` · `unknown` |
| `error_code` | `quote_fetch_failed` · `missing_error_object` · `non_error_rejection` · `status_failed_without_reason` · `unknown` |
| `source_hash_present` / `destination_hash_present` | boolean |

### Classification rules

- `getQuoteFetchErrorCode` — `missing_error_object` when the thrown value is nullish, `quote_fetch_failed` for an `Error`, `non_error_rejection` otherwise.
- `getSubmitFailureTelemetry` always reports `broadcast` with both hash flags `false`, because Core currently emits Failed before `executeSubmitStrategy`.
[...]

PR #9949, opened 2026-08-25
feat(bridge): emit SwapBridge failure_phase, error_code, and hash presence

Battambang:
## Explanation

Stacked on #9947 (schema + classifiers only). This PR **emits** the new Mixpanel fields so Failed is no longer a blended "on-chain" rate. Every value comes from the code path and hash presence, never from `error_message`.

### What each event emits

| Event | Emitted by | What we emit |
|---|---|---|
| Quotes Error | `bridge-controller` | `failure_phase=quote`, `error_code` from `getQuoteFetchErrorCode(error)` |
| Failed (submit catch, before `executeSubmitStrategy`) | `bridge-status-controller` | `getSubmitFailureTelemetry(error)` → `broadcast`, both hash flags `false` |
| Failed (EVM tx meta) | `bridge-status-controller` | `source_execution` when the tx has a hash, else `broadcast`; `error_code` `unknown` when the tx carries an error, else `missing_error_object` |
| Failed (status poll) | `bridge-status-controller` | `destination_execution` > `source_execution` > `poll`, from history src/dest hashes |
| Submitted | `bridge-status-controller` | both hash flags `false` (emit is still before the submit strategy) |
| Completed | `bridge-status-controller` | hash flags from bridge history src/dest hashes |

### Also added

`promoteFailurePhase` (new in `bridge-status-controller/src/utils/metrics.ts`) reconciles the phase reported by the emitting path with the combined hash presence — history flags OR-ed with any caller-supplied flags. It upgrades to `destination_execution` on a dest hash and to `source_execution` on a source hash, but only from `broadcast` / `poll` / `unknown`, so a genuine no-hash broadcast failure is never reclassified as `poll`.

On Failed, a caller-supplied `failure_phase` wins over the status-derived one, and its `error_code` is kept (defaulting to `unknown`); when the caller supplies no phase, both come from `getStatusFailureTelemetry`.

### Ownership

`bridge-status-controller` owns submit, poll, and EVM emit. `bridge-controller` owns Quotes Error and the Failed property defaults — when a client omits them, Failed falls back to `failure_phase=unknown`, `error_code=unknown`, and both hash flags `false`, so the properties are always present in the payload.

After #9947 merges, retarget this PR to `main`.

## References

* [WPN-1877](https://consensys.atlassian.net/browse/WPN-1877)
* Epic: [WPN-1875](https://consensys.atlassian.net/browse/WPN-1875)
* Schema: [WPN-1876](https://consensys.atlassian.net/browse/WPN-1876)
* Depends on: #9947

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes are limited to analytics payload shaping and classifier wiring; swap/bridge execution behavior is unchanged.
> 
> **Overview**
> This PR **wires up emission** of the SwapBridge failure telemetry schema from #9947 so Mixpanel events carry structured `failure_phase`, `error_code`, and hash-presence fields derived from code paths (not `error_message`).
> 
> **`bridge-controller`** adds `failure_phase=quote` and a classified `error_code` on **Quotes Error** (via `getQuoteFetchErrorCode` in the quote-fetch catch and in the event property builder). **Failed** event assembly is adjusted so client-supplied failure fields merge correctly with quote/request metadata.
> 
> **`bridge-status-controller`** attaches telemetry across the trade lifecycle: **Submitted** always emits both hash flags as `false`; **Completed** adds hash presence from bridge history; **Failed** combines broadcast catch helpers (`getBroadcastFailureProperties`), history-based classification (`getFailurePropertiesFromHistory`), EVM tx-meta fields, and **`promoteFailurePhase`** so phases align with combined hash presence without turning no-hash broadcast failures into `poll`. Failed events without a history item still get explicit defaults for all four fields.
> 
> Public helper renames: `getSubmitFailureTelemetry` → `getBroadcastFailureProperties`, `getStatusFailureTelemetry` → `getFailurePropertiesFromHistory`. The **Failed** event type now requires **`FailureTelemetryProperties`** instead of optional telemetry data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c73f0b77d193dbeee1b3b0fe2ae17921a398fe5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[WPN-1877]: https://consensyssoftware.atlassian.net/browse/WPN-1877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Battambang:
@metamaskbot publish-preview

Battambang:
@metamaskbot publish-preview

Battambang:
@metamaskbot publish-preview

Battambang:
@metamaskbot publish-preview

packages/bridge-status-controller/src/utils/metrics.ts
```diff
@@ -57,6 +57,117 @@ export type FailureTelemetryProperties = HashPresenceProperties & {
...
+  return phase;
+};
+
+/**
+ * Telemetry for Failed events emitted from the submit catch (no tx hash yet).
+ *
+ * @param error - The thrown value from submit.
+ * @returns Phase, error code, and hash-presence flags.
+ */
+export const getSubmitFailureTelemetry = (
```
micaelae: Can you rename this to `getBroadcastFailureProperties` to make it more descriptive?

packages/bridge-status-controller/src/utils/metrics.ts
```diff
@@ -57,6 +57,117 @@ export type FailureTelemetryProperties = HashPresenceProperties & {
...
+  return {
+    failure_phase: FailurePhase.Broadcast,
+    error_code: getSubmitErrorCode(error),
+    source_hash_present: false,
+    destination_hash_present: false,
+  };
+};
+
+/**
+ * Telemetry for Failed events derived from a status poll.
```
micaelae: Based on the usage, this is not limited to events emitted during polling. It's also called when TransactionController reports a failed swap tx, or anytime history is used to populate telemetry data. Something like `getFailureTelemetryFromHistory` would be a more intuitive name

packages/bridge-status-controller/src/utils/metrics.ts
```diff
@@ -379,90 +490,12 @@ export const getEVMTxPropertiesFromTransactionMeta = (
...
-  sourceHash?: string | null,
-  destinationHash?: string | null,
-): FailureTelemetryProperties => {
-  const hashPresence = getHashPresenceProperties(sourceHash, destinationHash);
-  return {
-    ...hashPresence,
-    failure_phase: getStatusFailurePhase(hashPresence),
-    error_code: SwapBridgeErrorCode.StatusFailedWithoutReason,
+    ...getHashPresenceProperties(transactionMeta.hash, undefined),
+    failure_phase: transactionMeta.hash
```
micaelae, line 502: Why not use getStatusFailurePhase and getSubmitErrorCode for consistency?

packages/bridge-status-controller/src/utils/metrics.ts
```diff
@@ -57,6 +57,117 @@ export type FailureTelemetryProperties = HashPresenceProperties & {
...
+  }
+  if (hashPresence.source_hash_present) {
+    return FailurePhase.SourceExecution;
+  }
+  return FailurePhase.Poll;
+};
+
+/**
+ * Align `failure_phase` with combined hash presence without turning a
+ * no-hash `broadcast` failure into `poll`.
```
micaelae, line 112: Which tx type and submission phase causes this to happen? Can the emitting code be updated to indicate the correct phase instead of overriding it using this util?

packages/bridge-status-controller/src/index.ts
```diff
@@ -59,6 +59,7 @@ export {
   getStatusFailureTelemetry,
   getSubmitErrorCode,
   getSubmitFailureTelemetry,
+  promoteFailurePhase,
 } from './utils/metrics.js';
 export type {
   FailureTelemetryProperties,
```
micaelae: Minor: These don't need to be exported

packages/bridge-status-controller/src/utils/metrics.ts
Battambang: Fixed in commit [https://github.com/MetaMask/core/pull/9949/changes/740b67846a162fe21b0fc172772bf108fed9a353](https://github.com/MetaMask/core/pull/9949/changes/740b67846a162fe21b0fc172772bf108fed9a353)

packages/bridge-status-controller/src/utils/metrics.ts
Battambang: Fixed with commit [https://github.com/MetaMask/core/pull/9949/changes/6561232fcaeb534cb48e90ff65e224578e00d885](https://github.com/MetaMask/core/pull/9949/changes/6561232fcaeb534cb48e90ff65e224578e00d885).
This runs for any Failed emit that has a history item (poll and TransactionController failures), not only polling.
=>Renaming to `getFailurePropertiesFromHistory` to better aligned with `getBroadcastFailureProperties`.

packages/bridge-status-controller/src/index.ts
Battambang: Fixed in commit [https://github.com/MetaMask/core/pull/9949/changes/290916c7125c9af9a2b3f20aefd65dc1f2b4ae17](https://github.com/MetaMask/core/pull/9949/changes/290916c7125c9af9a2b3f20aefd65dc1f2b4ae17).

packages/bridge-status-controller/src/utils/metrics.ts
Battambang: Those 2 helpers are not equivalent to the inline EVM Failed mapping mentioned here.

`getStatusFailurePhase` helper maps no hashes to `poll` phase. This path is a TransactionController failure with no dest hash: no hash means the tx never broadcast (broadcast phase), while hash means it landed (source_execution phase). Using `getStatusFailurePhase` would recode pre-hash EVM failures as poll. Also that hash/no hash split is why `promoteFailurePhase` exists.

`getSubmitErrorCode` is for thrown catch values (`instanceof Error`). `transactionMeta.error` is a plain `TransactionError` object, so that helper would emit `non_error_rejection` error_code instead of `unknown` error_code. That would weaken the segmentation of error_code that we are trying to properly isolate.

packages/bridge-status-controller/src/utils/metrics.ts
Battambang: Good remark there. This "hook" promote logic is to fix the Bugbot finding https://github.com/MetaMask/core/pull/9949#discussion_r3851879628.
Initial commit fixing this: https://github.com/MetaMask/core/pull/9949/changes/0e431b80c902d6ef8fc8abce9f2c318def5fe177

To be short, promoteFailurePhase is not tied to a special swap vs bridge type. It is fixing a  mismatch that only happens on one failed path: only applies to EVM TransactionController Failed (#onTransactionFailed for swap / bridge / batch) when a history item already exists.
Keeping it at the Failed merge sounds a rationale option because that is the only place that ORs history hashes with caller flags while emitters only see their local view (txMeta.hash vs poll hashes). It's also avoid duplicating that combine rule in every Failed caller.

packages/bridge-controller/src/bridge-controller.ts
```diff
@@ -1385,33 +1389,45 @@ export class BridgeController extends StaticIntervalPollingController<BridgePoll
...
-          ...baseProperties,
+          ...propertiesFromClient,
+          failure_phase:
+            failedClientProperties.failure_phase ?? FailurePhase.Unknown,
+          error_code:
+            failedClientProperties.error_code ?? SwapBridgeErrorCode.Unknown,
+          source_hash_present:
+            failedClientProperties.source_hash_present ?? false,
+          destination_hash_present:
+            failedClientProperties.destination_hash_present ?? false,
```
micaelae: Please move these fallbacks to bridge-status-controller where the events get emitted

Battambang: The status-controller already classifies Failed before calling `trackUnifiedSwapBridgeEvent`. These necessary fallbacks are not classification, they fill **Mixpanel** when a Failed caller omits the new optional fields.
Note that every Failed Mixpanel payload still goes through `BridgeController.#getEventProperties`, including direct `BridgeController:trackUnifiedSwapBridgeEvent(Failed, …)` that never hits status-controller. Moving the fallbacks would leave cases where those events without failure_phase / error_code / hash flags. This will defeat the purpose of trying to lower the the "blank"/"undefined"/"not set" kind of events values returned in Mixpanel, we do always want these values to be filled (even call without from status-controller).

Battambang:
@metamaskbot publish-preview

packages/bridge-controller/src/bridge-controller.ts
micaelae: Gotcha. The BridgeStatusController is the only emitter of Failed events and all status resolution logic should ideally live there

If we want to make sure these properties are always set, you can update the HashPresenceData and FailureTelemetryData types to require all the fields. `trackUnifiedSwapBridgeEvent` checks the payloads against these types so incomplete inputs will be caught at build time

packages/bridge-status-controller/src/utils/metrics.ts
```diff
@@ -49,6 +49,120 @@ import {
...
+export const getStatusFailurePhase = (
+  hashPresence: HashPresenceProperties,
+): FailurePhase => {
+  if (hashPresence.destination_hash_present) {
+    return FailurePhase.DestinationExecution;
+  }
+  if (hashPresence.source_hash_present) {
+    return FailurePhase.SourceExecution;
+  }
+  return FailurePhase.Poll;
```
micaelae, line 107: If the approval fails before the source tx is submitted, this evaluates to Poll. Can we either add an Approval phase or categorize this under SourceExecution?

packages/bridge-controller/src/bridge-controller.ts
Battambang: Refactored to move the fallbacks to the bridge-status-controller. Commit: [https://github.com/MetaMask/core/pull/9949/changes/efbf04aa1fca545797adad4916a5e95f1efbc794](https://github.com/MetaMask/core/pull/9949/changes/efbf04aa1fca545797adad4916a5e95f1efbc794)

packages/bridge-status-controller/src/utils/metrics.ts
Battambang: Added the `Approval` phase and its condition. commit: [https://github.com/MetaMask/core/pull/9949/changes/8fbba640fac276885a1b905fac755bfb61a12eda](https://github.com/MetaMask/core/pull/9949/changes/8fbba640fac276885a1b905fac755bfb61a12eda)

packages/bridge-status-controller/src/bridge-status-controller.ts
```diff
@@ -1830,6 +1858,34 @@ export class BridgeStatusController extends StaticIntervalPollingController<Brid
...
 
+    const historyHashPresence = getHashPresenceProperties(
+      historyItem.status.srcChain.txHash,
+      historyItem.status.destChain?.txHash,
+    );
+
+    const failurePropertiesFromHistory = getFailurePropertiesFromHistory(
+      historyItem.status.srcChain.txHash,
+      historyItem.status.destChain?.txHash,
+      Boolean(historyItem.approvalTxId) || historyItem.hasApprovalTx,
```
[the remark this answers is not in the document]
Battambang: Bugbot is accurate on all three points. The core issue (point 1) is a genuine design flaw in the implementation: using a persistent "approval was involved" flag as a proxy for "approval failed", is not the same thing. The net effect is that the Approval phase can mislabel non-EVM source-submit failures as approval, which defeats the diagnosability goal of the PR.

**Recommendation**
Revert the Approval phase. Bugbot's finding:

- For EVM, approval failures are already correctly classified (Broadcast/SourceExecution) via the EVM path — getStatusFailurePhase is never consulted.

- For non-EVM, Poll is the honest label for "status said failed, no hash context." Relabeling it approval based on hasApprovalTx introduces a worse mislabel, because that flag persists after the approval confirms.

- A precise approval label would require a positive "approval failed" signal that the history/status API does not currently expose.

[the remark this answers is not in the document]
Battambang: @micaelae From adding the `approval` phase. Not sure we can have this phase in the scope of this PR.

Battambang:
Related to Bugbot comment with the Approval phase:

Using a persistent "approval was involved" flag as a proxy for "approval failed" is not the same thing. The net effect is that the Approval phase can mislabel non-EVM source-submit failures as approval, which defeats the diagnosability goal of the PR.

> _Recommendation_

> Revert the Approval phase. Bugbot's finding:
> 
> For EVM, approval failures are already correctly classified (Broadcast/SourceExecution) via the EVM path — getStatusFailurePhase is never consulted.
> 
> For non-EVM, Poll is the honest label for "status said failed, no hash context." Relabeling it approval based on hasApprovalTx introduces a worse mislabel, because that flag persists after the approval confirms.
> 
> A precise approval label would require a positive "approval failed" signal that the history/status API does not currently expose.


@micaelae  From adding the approval phase. Not sure we can have this phase in the scope of this PR.

Battambang:
@metamaskbot publish-preview

packages/bridge-status-controller/src/bridge-status-controller.test.ts
```diff
@@ -1778,6 +1778,19 @@ describe('BridgeStatusController', () => {
...
         expect(fetchBridgeTxStatusSpy).toHaveBeenCalledTimes(1);
         expect(messengerCallSpy.mock.calls).toMatchSnapshot();
+        const failedCall = messengerCallSpy.mock.calls.find(
+          ([action, eventName]) =>
+            action === 'BridgeController:trackUnifiedSwapBridgeEvent' &&
+            eventName === UnifiedSwapBridgeEventName.Failed,
+        );
+        expect(failedCall?.[2]).toStrictEqual(
+          expect.objectContaining({
+            failure_phase: 'source_execution',
```
micaelae, line 1788: Bridge status polling doesn't start until the src hash is available. Shouldn't this result in a destination_execution failure phase?

```suggestion
            failure_phase: 'destination_execution',
```

packages/bridge-status-controller/src/__snapshots__/bridge-status-controller.test.ts.snap
```diff
@@ -6691,6 +6798,7 @@ exports[`BridgeStatusController subscription handlers TransactionController:tran
     "quoted_vs_used_gas_ratio": 0,
     "security_warnings": [],
     "slippage_limit": 0,
+    "source_hash_present": true,
```
micaelae, line 6801: This test is for a failed approval, for which I expect the suggested failure_phase and source_hash_present properties. Can this case be covered in this PR?

```suggestion
    "failure_phase": "approval",
   "feature_id": "unified_swap_bridge",
   "gas_included": false,
   "gas_included_7702": false,
@@ -6691,6 +6798,7 @@ exports[`BridgeStatusController subscription handlers TransactionController:tran
   "quoted_vs_used_gas_ratio": 0,
   "security_warnings": [],
   "slippage_limit": 0,
    "source_hash_present": false,
```

micaelae: Seeing the same property mismatches for these test cases
- `should track failed event for bridge transaction if approval is dropped`
- `Tron swap with approval should handle approval transaction errors`
- `EVM bridge should throw an error if approval tx fails`

packages/bridge-status-controller/src/bridge-status-controller.test.ts
Battambang: `source_execution ` is correct here. The test polls once and gets MockStatusResponse.getFailed() which sets `srcChain.txHash` (source hash present) but `destChain` with no `txHash` (destination hash absent). Logic for this is in `getStatusFailurePhase` from src/utils/metrics.ts: `destination_execution` requires the destination hash to be present which it is not in this case.

Phase is categorized poll regarding the hash presence because there is no thrown error for status failure. It has to be considered as a fallback where nothing is known, the status endpoint reported FAILED with no detail and no hash at source and destination.

packages/bridge-status-controller/src/__snapshots__/bridge-status-controller.test.ts.snap
Battambang: The "approval" phase is trying to introduce some edge cases logic that is not straightforward to handle in the scope of this PR. The risk here is having incoherent telemetry as reported previously by Bugbot.
It is better to handle this case in another ticket as a better enhancement with phase.
Reverting Approval addition.

packages/bridge-status-controller/src/index.ts
```diff
@@ -56,7 +56,7 @@ export {
 export {
   getHashPresenceProperties,
   getStatusFailurePhase,
-  getStatusFailureTelemetry,
+  getFailurePropertiesFromHistory,
   getSubmitErrorCode,
-  getSubmitFailureTelemetry,
+  getBroadcastFailureProperties,
```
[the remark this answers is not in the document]
Battambang: The realistic set of external consumers importing these methods is zero since it is only related to this current wire work.

micaelae (approved):

micaelae:
LGTM

PR #9951, opened 2026-08-25
Feat/aaguid extract

lwin-kyaw:
## Explanation

`PasskeyController` already parses the authenticator AAGUID during enrollment and persists it at `passkeyRecord.credential.aaguid`, but only as a side effect of `verifyRegistrationResponse`. Clients that want the AAGUID *before* enrollment completes — to record which authenticator a user is enrolling, or to show where their passkey lives (iCloud Keychain, Google Password Manager, a hardware key) — had no way to get it. The WebAuthn parsing helpers are internal to the package, and `verifyRegistrationResponse` can't stand in for them: it requires an expected challenge, origin, and RP ID, and performs full attestation verification.

This PR makes that value reachable in two steps.

**Export `getAAGUIDFromRegistrationResponse`.** Given the result of `navigator.credentials.create()`, it decodes the attestation object, parses the attested credential data in `authData`, and returns the AAGUID as a dashed UUID string. It returns `undefined` when the authenticator data carries no attested credential data, and throws when the attestation object or authenticator data is malformed. It lives in `verify-registration-response.ts` alongside the verifier and shares the same AAGUID formatting helper, so the string it returns is identical to the one persisted in state.

The response is parsed as-is, without verification. Until `protectVaultKeyWithPasskey` verifies the attestation, callers should treat the value as untrusted and use it for telemetry or display only.

**Request `attestation: 'direct'` during registration.** This is the change that makes the AAGUID meaningful. With `attestation: 'none'`, browsers anonymize the response and replace the AAGUID with all zeros, so the new util would return `00000000-0000-0000-0000-000000000000` for every enrollment.

Worth a reviewer's attention: `verifyRegistrationResponse` only accepts the `none` and self-attested `packed` attestation formats, and throws `Unsupported attestation format` otherwise. Requesting `direct` invites authenticators to respond with formats we don't implement (`apple`, `tpm`, `fido-u2f`, or `packed` with an `x5c` certificate chain), which would fail enrollment at the verification step rather than at registration. Platform authenticators on the targeted clients should be exercised before this ships.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Additive API and internal refactor with expanded tests; enrollment verification and controller attestation settings are unchanged in this diff.
> 
> **Overview**
> Adds **`getAAGUIDFromRegistrationResponse`**, a public helper that decodes a `navigator.credentials.create()` result and returns the authenticator **AAGUID** as a dashed UUID (or `undefined` when attested credential data is absent), **without** running full registration verification. It is exported from the package entry point and documented in the changelog for pre-enrollment telemetry or UI (e.g. which passkey provider was used).
> 
> **`verifyRegistrationResponse`** now formats the persisted AAGUID through a shared **`formatAAGUID`** helper so verified and pre-verify strings stay consistent.
> 
> Tests cover the new util (including malformed attestation data), **`tokenBinding`** statuses `present` / `supported` / `not-supported` on registration and authentication verification, and a **`buildSignedAssertion`** helper so authentication tests can mutate `clientDataJSON` without breaking signatures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 429fcf58478e325163fb1df289cd180967caf8c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/passkey-controller/CHANGELOG.md
+++ b/packages/passkey-controller/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `getAAGUIDFromRegistrationResponse` to read the authenticator AAGUID from a `navigator.credentials.create()` result, returned as a dashed UUID string, or `undefined` when the authenticator data carries no attested credential data. Throws if the attestation object or its authenticator data is malformed
+- `getAAGUIDFromRegistrationResponse` to read the authenticator AAGUID from a `navigator.credentials.create()` result. ([#9951](https://github.com/MetaMask/core/pull/9951))
+  - The function returned as a dashed UUID string, or `undefined` when the authenticator data carries no attested credential data. Throws if the attestation object or its authenticator data is malformed. 
   - The response is parsed as-is without verification, so treat the value as untrusted until `protectVaultKeyWithPasskey` verifies the attestation; the verified AAGUID is persisted on `passkeyRecord.credential.aaguid`.
 
 ### Changed
--- a/packages/passkey-controller/CHANGELOG.md
+++ b/packages/passkey-controller/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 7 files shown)


lwin-kyaw:
@metamaskbot publish-preview

chaitanyapotti (approved):

tanguyenvn (approved):

PR #9954, opened 2026-08-25
fix(transaction-pay-controller): use exact input for Relay deposits

pedronfigueiredo:
## Explanation

Relay quote behavior is determined after transaction processing rather than by broad transaction-type allowlists.

This change:

- uses `EXACT_INPUT` with `sourceTokenAmount` when a Relay quote has no embedded `txs`, so the entered source amount remains the total amount paid;
- keeps plain `perpsDeposit` flows input-based, while preserving `EXACT_OUTPUT` with `targetAmountMinimum` for `perpsDepositAndOrder` because the follow-on order requires guaranteed target margin;
- preserves `EXACT_OUTPUT` with the target amount for quotes that embed transactions, including Money Account calls;
- assigns the final Relay `tradeType` once after transaction processing and lets transaction helpers provide specialized amounts alongside their other request fields;
- adds the optional, client-facing `TransactionPayQuote.isInputBased` and `TransactionPayTotals.isInputBased` flags;
- has Relay, Across, and Server strategies expose normalized input-based semantics without requiring totals code to inspect provider-specific quote internals;
- treats omitted quote flags, including Fiat quotes, as output-based;
- marks totals as input-based only when every selected quote is input-based, preventing source-side fees from being counted twice; and
- clamps negative Relay dust to zero when exact-input output is below the prior target minimum.

## Consumer integration

The coordinated client PRs remain drafts until the Core package containing this change is published:

- [MetaMask Mobile #35249](https://github.com/MetaMask/metamask-mobile/pull/35249) consumes `TransactionPayTotals.isInputBased`, keeps input-based source amounts stable, and displays the authoritative target amount.
- [MetaMask Extension #45753](https://github.com/MetaMask/metamask-extension/pull/45753) consumes the same aggregate flag and preserves the legacy presentation for output-based totals.

Both clients treat only an explicit `isInputBased: true` as input-based. Mixed aggregates, embedded-transaction quotes, and `perpsDepositAndOrder` quotes retain the existing **Total** presentation. Input-based withdrawals use the controller's authoritative target amount; legacy or output-based withdrawals retain the existing fee-subtraction fallback.

Final client integration still requires installing the published Core package, removing temporary compatibility types or local links, regenerating dependency artifacts, and running each client's integration validation.

## Validation

- `yarn workspace @metamask/transaction-pay-controller run test`
  - 100% statement, line, and function coverage; 99.49% branch coverage
- `yarn workspace @metamask/transaction-pay-controller run jest --no-coverage src/strategy/relay/relay-quotes.test.ts`
  - 1 suite and 144 tests passed
- `yarn workspace @metamask/transaction-pay-controller run jest --no-coverage src/strategy/fiat/fiat-direct-musd.test.ts src/utils/totals.test.ts`
  - 2 suites and 27 tests passed
- `yarn workspace @metamask/transaction-pay-controller run build`
- `yarn workspace @metamask/transaction-pay-controller run changelog:validate`
- ESLint and `oxfmt --check` for all 14 changed TypeScript files
- `git diff --check origin/main...HEAD`

## References

- [CONF-1782](https://consensyssoftware.atlassian.net/browse/CONF-1782)
- Mobile consumer PR: https://github.com/MetaMask/metamask-mobile/pull/35249
- Extension consumer PR: https://github.com/MetaMask/metamask-extension/pull/45753

## Checklist

[CONF-1782]: https://consenssoftware.atlassian.net/browse/CONF-1782

---

> [!NOTE]
> **Medium Risk**
> Changes default Relay trade type and pay-total math for many deposit paths; incorrect classification of embedded or perps flows could mis-quote amounts or inflate totals.
> 
> **Overview**
> Relay quoting now picks **`EXACT_INPUT`** with **`sourceTokenAmount`** after transaction processing when the request has **no embedded `txs`**, replacing the prior default **`EXPECTED_OUTPUT`** / target-minimum behavior for plain deposits, max-amount, and token-transfer flows. **`EXACT_OUTPUT`** with **`targetAmountMinimum`** is kept when quotes embed transactions (delegation, Money Account post-quote) or when the parent type is **`perpsDepositAndOrder`**, so margin for the follow-on order stays guaranteed.
> 
> The controller exposes optional **`TransactionPayQuote.isInputBased`** and **`TransactionPayTotals.isInputBased`** (Relay, Across, Server set them from trade type; omitted flags count as output-based). **`calculateTotals`** no longer takes **`isMaxAmount`**; it marks totals input-based only when **every** selected quote is input-based, uses the quoted target for the payment slice in that case to avoid double-counting fees, and Relay dust is clamped to zero when exact-input output sits below the prior target minimum.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 519b8b5f1dc73ac33cd4a1d830d69b747e4965ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
@@ -205,6 +205,7 @@ function combineDirectMusdFiatQuote({
       },
       targetNetwork: { fiat: '0', usd: '0' },
     },
+    isInputBased: false,
     original: {
       rampsQuote: fiatQuote,
       relayQuote: undefined,
--- a/packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
+++ b/packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
@@ -205,7 +205,6 @@ function combineDirectMusdFiatQuote({
       },
       targetNetwork: { fiat: '0', usd: '0' },
     },
-    isInputBased: false,
     original: {
       rampsQuote: fiatQuote,
       relayQuote: undefined,
--- a/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
+++ b/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
@@ -82,6 +82,7 @@ const POST_QUOTE_GAS_BUFFER = 1.1;
 const PAYMENT_OVERRIDE_GAS = 75_000;
 const ZERO_AMOUNT = { fiat: '0', human: '0', raw: '0', usd: '0' };
 
+type RelayQuoteRequestDraft = Omit<RelayQuoteRequest, 'tradeType'>;
 type RelayStepData = RelayTransactionStep['items'][0]['data'];
 
 type RelayGasResult = {
@@ -323,7 +324,7 @@ async function getSingleQuote(
       ? { ...request, recipient: nonAtomicRecipient }
       : request;
 
-    const body: RelayQuoteRequest = {
+    const body: RelayQuoteRequestDraft = {
       amount: useExactInput ? sourceTokenAmount : targetAmountMinimum,
       destinationChainId: Number(targetChainId),
       destinationCurrency: targetTokenAddress,
@@ -337,7 +338,6 @@ async function getSingleQuote(
         : {}),
       recipient: effectiveRequest.recipient ?? from,
       slippageTolerance,
-      tradeType: useExactInput ? 'EXACT_INPUT' : 'EXPECTED_OUTPUT',
       user: from,
     };
 
@@ -378,13 +378,17 @@ async function getSingleQuote(
     }
 
     if (!body.txs?.length) {
-      body.tradeType = 'EXACT_INPUT';
       body.amount = sourceTokenAmount;
     }
 
-    log('Request body', body);
+    const finalBody: RelayQuoteRequest = {
+      ...body,
+      tradeType: body.txs?.length ? 'EXACT_OUTPUT' : 'EXACT_INPUT',
+    };
 
-    const quote = await fetchRelayQuote(messenger, body, signal);
+    log('Request body', finalBody);
+
+    const quote = await fetchRelayQuote(messenger, finalBody, signal);
 
     log('Fetched relay quote', quote);
 
@@ -469,7 +473,7 @@ async function resolveNonAtomicRecipient(
 async function processTransactions(
   transaction: TransactionMeta,
   request: QuoteRequest,
-  requestBody: RelayQuoteRequest,
+  requestBody: RelayQuoteRequestDraft,
   messenger: TransactionPayControllerMessenger,
 ): Promise<boolean> {
   // Skip when skipProcessTransactions (defaulting to isPostQuote) is set — the
@@ -523,7 +527,6 @@ async function processTransactions(
   requestBody.authorizationList = normalizeAuthorizationList(
     delegation.authorizationList,
   );
-  requestBody.tradeType = 'EXACT_OUTPUT';
 
   const tokenTransferData = nestedTransactions?.find((nestedTx) =>
     nestedTx.data?.startsWith(TOKEN_TRANSFER_FOUR_BYTE),
@@ -560,7 +563,7 @@ async function processTransactions(
 async function processMoneyAccountPostQuote(
   transaction: TransactionMeta,
   request: QuoteRequest,
-  requestBody: RelayQuoteRequest,
+  requestBody: RelayQuoteRequestDraft,
   messenger: TransactionPayControllerMessenger,
 ): Promise<void> {
   const { transactionData: transactionDataList } = messenger.call(
@@ -589,7 +592,6 @@ async function processMoneyAccountPostQuote(
   const rawAmount = transactionData?.tokens?.[0]?.amountRaw ?? '0';
 
   requestBody.authorizationList = normalizeAuthorizationList(authorizationList);
-  requestBody.tradeType = 'EXACT_OUTPUT';
   requestBody.amount = rawAmount;
   requestBody.txs = [
     {
--- a/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
+++ b/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
@@ -306,8 +306,7 @@ async function getSingleQuote(
     // Explicit source-driven flows use EXACT_INPUT immediately. Other flows
     // start as EXPECTED_OUTPUT until transaction processing determines whether
     // the request embeds transactions.
-    const useExactInput =
-      (isMaxAmount ?? false) || (request.isPostQuote ?? false);
+    const useExactInput = Boolean(isMaxAmount || request.isPostQuote);
 
     const useExecute =
       supports7702 &&
--- a/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
+++ b/packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
@@ -83,7 +83,8 @@ const POST_QUOTE_GAS_BUFFER = 1.1;
 const PAYMENT_OVERRIDE_GAS = 75_000;
 const ZERO_AMOUNT = { fiat: '0', human: '0', raw: '0', usd: '0' };
 
-type RelayQuoteRequestDraft = Omit<RelayQuoteRequest, 'amount' | 'tradeType'>;
+type RelayQuoteRequestDraft = Omit<RelayQuoteRequest, 'amount' | 'tradeType'> &
+  Partial<Pick<RelayQuoteRequest, 'amount'>>;
 type RelayStepData = RelayTransactionStep['items'][0]['data'];
 
 type RelayGasResult = {
@@ -352,15 +353,13 @@ async function getSingleQuote(
       messenger,
     );
 
-    let transactionAmount: string | undefined;
-
     if (
       !processedTransactions &&
       isAtomic &&
       effectiveRequest.isPostQuote &&
       effectiveRequest.paymentOverride === PaymentOverride.MoneyAccount
     ) {
-      transactionAmount = await processMoneyAccountPostQuote(
+      await processMoneyAccountPostQuote(
         transaction,
         effectiveRequest,
         body,
@@ -380,7 +379,7 @@ async function getSingleQuote(
     const finalBody: RelayQuoteRequest = {
       ...body,
       amount:
-        transactionAmount ??
+        body.amount ??
         (requiresExactOutput ? targetAmountMinimum : sourceTokenAmount),
       tradeType: requiresExactOutput ? 'EXACT_OUTPUT' : 'EXACT_INPUT',
     };
@@ -564,7 +563,7 @@ async function processMoneyAccountPostQuote(
   request: QuoteRequest,
   requestBody: RelayQuoteRequestDraft,
   messenger: TransactionPayControllerMessenger,
-): Promise<string | undefined> {
+): Promise<void> {
   const { transactionData: transactionDataList } = messenger.call(
     'TransactionPayController:getState',
   );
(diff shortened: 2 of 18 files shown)


packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -303,18 +303,11 @@ async function getSingleQuote(
...
-    // HyperCore perps deposits fund an order that requires the full target as
-    // margin, so the delivered amount must be guaranteed rather than expected.
-    // EXPECTED_OUTPUT only guarantees `target * (1 - slippage)`, which leaves
-    // the follow-on order short and it fails on insufficient margin.
-    const useExactOutput = !useExactInput && isHypercoreDeposit(request);
+    // Explicit source-driven flows use EXACT_INPUT immediately. Other flows
+    // start as EXPECTED_OUTPUT until transaction processing determines whether
+    // the request embeds transactions.
+    const useExactInput =
+      (isMaxAmount ?? false) || (request.isPostQuote ?? false);
```
matthewwalsh0: Minor, could also just `Boolean(isMaxAmount || request.isPostQuote)` if more readable?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -345,7 +338,7 @@ async function getSingleQuote(
         : {}),
       recipient: effectiveRequest.recipient ?? from,
       slippageTolerance,
-      tradeType: getTradeType(useExactInput, useExactOutput),
+      tradeType: useExactInput ? 'EXACT_INPUT' : 'EXPECTED_OUTPUT',
```
matthewwalsh0: Minor, was the `getTradeType` function useful if we pass it the request and `isMaxAmount` it owns the full decision.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -385,6 +378,11 @@ async function getSingleQuote(
       body.refundTo = effectiveRequest.refundTo;
     }
 
+    if (!body.txs?.length) {
+      body.tradeType = 'EXACT_INPUT';
```
matthewwalsh0: Minor, is it simpler to assign `tradeType` once after we've set the transactions?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0: This `EXPECTED_OUTPUT` branch is never hit as we override it regardless if transactions or not?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -385,6 +378,11 @@ async function getSingleQuote(
...
+      body.amount = sourceTokenAmount;
```
matthewwalsh0: Also here, is it more readable to just assign `amount` once after transactions are processed?

packages/transaction-pay-controller/src/types.ts
```diff
@@ -838,6 +838,9 @@ export type TransactionPayTotals = {
   /** Total fees for the target transaction and all quotes. */
   fees: TransactionPayFees;
 
+  /** Whether all selected quotes are driven by the source input amount. */
```
matthewwalsh0: Is this a bit ambiguous?

Maybe `Whether the quote(s) subtract fees from the destination amount meaning the input amount is static`

packages/transaction-pay-controller/src/utils/totals.ts
```diff
@@ -146,38 +152,67 @@ export function calculateTotals({
...
+ * @param quote - Selected quote.
+ * @param isMaxAmount - Whether the target transaction uses the maximum input.
+ * @returns Whether the quote uses input-based semantics.
+ */
+function isQuoteInputBased(
+  quote: TransactionPayQuote<unknown>,
+  isMaxAmount: boolean,
+): boolean {
+  if (quote.strategy === TransactionPayStrategy.Relay) {
+    const relayQuote = quote.original as RelayQuote;
```
matthewwalsh0: This breaks the abstraction provided by the `TransactionPayStrategy` and couples this code to the internals of each one.

The purpose of the `TransactionPayQuote` type is to abstract what a quote can return, so it would also need a `isInputBased` property?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Agreed, `Boolean(...)` is clearer here. I’ll simplify it.

Addressed in `5f0d8b805e` ([commit](https://github.com/MetaMask/core/commit/5f0d8b805e8da30cdf330d69a6ebc35f162c84ac)).

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Good point. I’ll derive `tradeType` once from whether transactions were added after processing, instead of setting and overriding it in multiple places.

Addressed in `b32aea78f3` ([commit](https://github.com/MetaMask/core/commit/b32aea78f3fe8fa30350747c9126374858cce1b4)).

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Agreed. I’ll defer `amount` too and set it once after transaction processing, preserving the Money Account transaction amount when present.

Addressed in `d0eb8912b4` ([commit](https://github.com/MetaMask/core/commit/d0eb8912b4633a06a7066460720a7ff1d701c943)).

packages/transaction-pay-controller/src/types.ts
pedronfigueiredo: Agreed—the fee behavior is more concrete. I’ll update the documentation to clarify that fees are subtracted from the destination amount while the input amount stays static.

Addressed in `c848146ce2` ([commit](https://github.com/MetaMask/core/commit/c848146ce29555d331d4a5578800552a258033fa)).

packages/transaction-pay-controller/src/utils/totals.ts
pedronfigueiredo: Agreed. I’ll expose `isInputBased` on `TransactionPayQuote`, have each strategy populate it, and make totals consume only the normalized property.

Addressed in `ccb7854521` ([commit](https://github.com/MetaMask/core/commit/ccb7854521d661df29f2357cd29a20b330532587)).

packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
```diff
@@ -205,6 +205,7 @@ function combineDirectMusdFiatQuote({
       },
       targetNetwork: { fiat: '0', usd: '0' },
     },
+    isInputBased: false,
```
matthewwalsh0: Minor, since it's optional we can just assume `undefined` is `false`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -366,13 +351,15 @@ async function getSingleQuote(
...
+    let transactionAmount: string | undefined;
+
     if (
       !processedTransactions &&
       isAtomic &&
       effectiveRequest.isPostQuote &&
       effectiveRequest.paymentOverride === PaymentOverride.MoneyAccount
     ) {
-      await processMoneyAccountPostQuote(
+      transactionAmount = await processMoneyAccountPostQuote(
```
matthewwalsh0: I think it's still okay for the utils to mutate the request directly for cleaner encapsulation.

But was just meaning in this function we were assigning twice which seemed unnecessary 👍

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -610,42 +603,8 @@ async function processMoneyAccountPostQuote(
...
-/**
- * Whether the quote deposits into HyperCore USDC.
- *
- * `normalizeRequest` remaps Arbitrum-USDC perps deposits to HyperCore before
- * the quote is built, so the check is against the normalized target.
- *
- * @param request - Normalized quote request.
- * @returns True when the target is HyperCore USDC.
- */
-function isHypercoreDeposit(request: QuoteRequest): boolean {
```
matthewwalsh0, line 624: Do we not need to keep this or something similar to ensure it's `EXACT_OUTPUT` for Hyperliquid deposits?

I'm assuming `EXACT_INPUT` is okay for `perpsDeposit` but **not** for `perpsDepositAndOrder` since we need a guaranteed return?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -585,15 +580,13 @@ async function processMoneyAccountPostQuote(
...
-    return;
+    return undefined;
   }
 
   const fundingRecipient = recipient ?? request.from;
   const rawAmount = transactionData?.tokens?.[0]?.amountRaw ?? '0';
 
   requestBody.authorizationList = normalizeAuthorizationList(authorizationList);
-  requestBody.tradeType = 'EXACT_OUTPUT';
-  requestBody.amount = rawAmount;
```
matthewwalsh0, line 596: Another reason to keep the assignment here is we still assign `authorizationList` and `txs` so diverges the pattern.

matthewwalsh0 (dismissed):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Good catch. `EXPECTED_OUTPUT` was redundant here. I removed that branch and now assign `tradeType` once after transaction processing: `EXACT_OUTPUT` when transactions are present and `EXACT_INPUT` otherwise.

Addressed in `b32aea78f3` ([commit](https://github.com/MetaMask/core/commit/b32aea78f3fe8fa30350747c9126374858cce1b4)).

packages/transaction-pay-controller/src/strategy/fiat/fiat-direct-musd.ts
pedronfigueiredo: Agreed. Since aggregation treats only strict `true` as input-based, an omitted value already means false. I’ll remove the explicit flag from the direct mUSD quote and update its test.

Addressed in `5e11e1b6bf` ([commit](https://github.com/MetaMask/core/commit/5e11e1b6bf88b05f565bb7a7efc8a154e135ccf8)).

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Thanks, understood. The helpers still mutate `authorizationList` and `txs` for encapsulation, while the Money Account amount is returned and assigned once alongside `tradeType` in the final request body. I’ll leave that structure as-is.

Addressed in `d0eb8912b4` ([commit](https://github.com/MetaMask/core/commit/d0eb8912b4633a06a7066460720a7ff1d701c943)).

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: Good catch. `perpsDepositAndOrder` still needs a guaranteed target amount for the follow-on order. I’ll restore that exception so it uses `targetAmountMinimum` with `EXACT_OUTPUT`, while plain `perpsDeposit` remains `EXACT_INPUT`.

Addressed in `0415517cbb` ([commit](https://github.com/MetaMask/core/commit/0415517cbb7462d53a843ca8582a67269af12200)).

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
pedronfigueiredo: That makes sense. I’ll keep the Money Account `amount` assignment with `authorizationList` and `txs` in the helper, and let final request assembly only supply the default amount when the helper did not set one. `tradeType` will still be assigned once after processing.

Addressed in `466b4fc7ef` ([commit](https://github.com/MetaMask/core/commit/466b4fc7eff848b172d5aa6899b66f3a89429a6f)).

OGPoyraz (approved):

PR #9958, opened 2026-08-25
fix: add a clearState function and fix ongoing flow wrongly reset

Akaryatrh:
## Explanation

`KycController.reset()` deliberately preserves the session email, the persisted terms
acceptance, the per-product KYC-required cache and the user-keyed status, because it exists
to restart a flow rather than to erase the customer. That left no way to erase them, so
after a full wallet reset the previous user's email was still sitting in controller state.

This PR adds `KycController.clearState()` (exposed as the `KycController:clearState`
messenger action) which restores `getDefaultKycControllerState()`, drops the auth-frame
client token, stops both polling loops (session status and user-keyed status), and bumps the
flow generation so async work started earlier discards its results.

While covering that last part, it turned out the generation guard did not actually stop an
in-flight flow. `#generation` only protects writes routed through `#updateIfCurrent`, and
`loadDisclaimers()` captures the generation at its own entry — i.e. the *post*-reset value —
so its guard passed. A `clearState()` landing mid-`initialize()` therefore left the
controller at `phase: 'terms'` with `geoCountry` repopulated instead of at its defaults.
`initialize()` now bails out after the geolocation await when it has been superseded, which
fixes the same hole in `reset()`.

Two smaller changes worth flagging for reviewers:

- The teardown shared by `reset()` and `clearState()` moved into a private
  `#supersedeInFlightWork()` helper.
- Two Iron-vendor tests were named for a reset landing *during* customer creation but
  simulated it by resetting before geolocation resolved. The new guard short-circuits that
  path earlier, so they now reset inside the `createVendorCustomer` mock, which is what their
  names describe and keeps the guards inside that block covered.

Two related gaps are intentionally left out of scope: `clearState()` does not touch the UKYC
`local_user_secret` in Encrypted User Storage, and it does not republish `statusChanged`, so
a consumer caching the last user status for the Money toast keeps a stale value (the event
payload type cannot express `null`).

## References

- Stacked on `feat/kyc-controller-generic-ukyc`; please review/merge that first.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes KYC controller lifecycle and wallet-reset behavior; the race fix affects when flows resume, but scope is limited to kyc-controller state and async guards.
> 
> **Overview**
> Adds **`KycController.clearState()`** (and **`KycController:clearState`**) for a full wallet reset: it restores **`getDefaultKycControllerState()`**, wiping email, persisted terms, the per-product KYC-required cache, and user-keyed status—data **`reset()`** intentionally keeps when restarting a flow.
> 
> **`reset()`** and **`clearState()`** now share **`#cancelPendingSession()`**, which clears the auth-frame client token, stops session and user-status polling, and bumps flow **generation** so in-flight async work is discarded.
> 
> **`initialize()`** returns immediately after the geolocation await when **generation** has changed, so a late geo response can no longer advance a superseded flow into **`terms`** or trigger disclaimer loading after **`reset()`** or **`clearState()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d23deb68f65fbeef4aca8030cea68ec76b47ef0e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/kyc-controller/src/KycController.ts
```diff
@@ -1420,6 +2029,34 @@ export class KycController extends BaseController<
...
+    });
+  }
+
+  /**
+   * Tears down everything that lives outside state: drops the auth-frame
+   * client token, stops both polling loops, and bumps the flow generation so
+   * async steps started earlier discard their results instead of writing them
+   * onto the controller. Shared by {@link reset} and {@link clearState}.
+   */
+  #supersedeInFlightWork(): void {
```
jiexi: maybe a better term would be something like `cancelInflightSession`, `cancelPendingSession`?

jiexi: https://github.com/MetaMask/core/pull/9958/commits/bfaf930fd210877e3d269ab3957e4cce6ba265b8

jiexi (approved):

Akaryatrh:
@metamaskbot publish-previews

PR #9961, opened 2026-08-25
feat(perps-controller): add Chase order lifecycle

abretonc7s:
## Description

Completes the controller lifecycle Mobile needs for Chase orders. It adds retained order snapshots, background suspension, provider routing, max-distance handling, and safe termination of stale or already-gone child orders.

## Why this is separate from the initial Chase change

The initial change implemented Chase placement and cancel/reprice. TAT-3813 then exercised the full Mobile flow, including the running-order UI, partial fills, stale snapshots, termination retries, and app backgrounding.

Backgrounding on a device exposed the main gap. The app can move to the background after a tick cancels the old child but before it rests the replacement. Stopping the tick at that point leaves no order on the book, even though Mobile reports a resting `backgrounded` Chase. This PR lets an admitted reprice finish, then stops the session and leaves the latest child resting as a limit order. It also prevents a timer from restarting the Chase later.

The same tests found the other lifecycle gaps covered here: partial-fill refresh, stable public snapshots, provider-aware cancellation, max-distance stopping, and idempotent termination when Mobile holds stale state.

## Validation

- Full Core build and perps-controller test suite
- Core lint, TypeScript, action-type, changelog, and security CI checks
- Independent review on the exact commit
- Mobile yalc integration with focused tests, component-view tests, lint, and TypeScript

The device recipes found these edge cases. Mobile will rerun the final uninterrupted recipe after it switches from yalc to the published package.

## Release handoff

After merge, publish a new `@metamask/perps-controller` version. Mobile will replace the yalc link with that version and rerun its tests and device proof.

Related: TAT-3813

---

> [!NOTE]
> **Medium Risk**
> Large changes to live order repricing, cancellation, and app-background teardown on the exchange path; incorrect concurrency could strand or duplicate resting orders, though behavior is heavily tested.
> 
> **Overview**
> Adds a **public Chase lifecycle surface** on `PerpsController`: **`getChaseOrders()`** for retained session snapshots and **`suspendChaseOrders()`** to stop repricing when the app backgrounds while leaving the latest post-only child resting. **`AggregatedPerpsProvider`** fans both calls across venues, attaches **`providerId`**, and fails suspension only if any venue cannot suspend (successful venues stay suspended for retry).
> 
> **HyperLiquid emulated Chase** gains stable **`ChaseOrder`** snapshots (including **`distanceChasedBps`**, optional **`chaseMaxDistanceBps`** adverse-touch stop, and exported **`CHASE_ORDER_STATUS`**), default repricing every **15s** with **unbounded** duration/reprice caps unless callers set limits, and **initial placement retries** for retryable post-only/oracle rejections. Background work is **serialized** (tick queue, pause during placement) so foreground orders and in-flight reprices do not race disconnect/suspend.
> 
> **Cancellation** is hardened: deduplicated/idempotent termination by strategy handle, **`termination_pending`** while a child cancel can retry, treating already-gone children as success, and routing **child exchange IDs** back to the owning Chase session. Validation adds **`ORDER_CHASE_MAX_DISTANCE_INVALID`**; analytics adds Chase background/termination interaction values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5df7885e9a9c0cae65841c1c9889a669b821d72a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


abretonc7s:
Ready for urgent Perps review. This unblocks the next `@metamask/perps-controller` release for Mobile Chase validation.

Validated at `e31ba042af25db99d7e73a738dab59ee6008a2f8`:
- full Core build passed
- perps-controller suite passed (332 tests)
- package changelog validation and all current CI gates passed
- independent isolated review approved with zero findings
- Mobile yalc integration passed focused tests, component-view tests, scoped lint, and TypeScript

Auto-merge is enabled once the required Perps/last-push approval lands.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -5525,6 +5756,98 @@ export class HyperLiquidProvider implements PerpsProvider {
     }
   }
 
+  /**
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 26a916ac95. An incomplete termination now retains the stable handle and reports CHASE_ORDER_STATUS.TerminationPending while the child remains cancellable; retrying the same handle can complete successfully.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -5525,6 +5756,98 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+        if (session.pendingReplacement) {
+          replacements.push(session.pendingReplacement);
+        }
+      }
+      await this.#chaseTickQueue;
+      await Promise.all(replacements);
+      return await this.getChaseOrders();
+    } finally {
+      this.#chasePlacementBlockers -= 1;
+    }
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 26a916ac95. Suspension now clears timers, drains the admitted Chase tick queue, and only then backgrounds active sessions. The adversarial test proves an admitted reprice rests its replacement before suspension returns.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -5519,12 +5756,115 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+          session.pendingReplacement === null &&
+          now - session.lastSnapshotSizeRefreshAt >= session.intervalMs
+        ) {
+          session.lastSnapshotSizeRefreshAt = now;
+          try {
+            const liveRemainder = await this.#readOrderRemainder(
+              session.orderId,
+            );
+            if (liveRemainder !== null) {
+              snapshotRemainingSize = liveRemainder;
```
Bigshmow, line 5794: liveRemainder is assigned to the local only, but line 5783 already stamped lastSnapshotSizeRefreshAt. The next read inside the interval window skips the refresh and falls back to session.size (line 5777) — still the last placed size — so remainingSize oscillates 1 → 0.2 → 1 → 0.2 on a partially filled child. Repro: call getChaseOrders() twice after a partial fill; the second returns 1. The existing partial-fill test only reads once.
Suggest session.size = liveRemainder alongside it — matches what the gone branch does at line 5424, and gives #resolveOwnRestingSizes a fresher size for netting.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -6130,6 +6502,25 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+    // Resolve ownership before the first await. A reprice can change the child
+    // ID while an ordinary cancel is in flight; cancelling through the stable
+    // session handle stops the loop, joins any replacement, and removes the
+    // current child instead of leaving a replacement chasing behind the user.
+    const owningChase = [...this.#chaseSessions.entries()].find(
+      ([, session]) =>
+        session.orderId === params.orderId ||
+        session.replacingOrderId === params.orderId,
+    );
+    if (owningChase) {
```
Bigshmow: This block sits outside the try at 6524, so a throw from #ensureReadyForTrading() or #cancelChaseChild rejects cancelOrder instead of returning createErrorResult. Confirmed with a mocked transport failure. Every other cancelOrder path — including the orderType: 'chase' route via #cancelStrategyOrder — returns {success: false, error}, so this breaks the "callers that omit orderType keep the behaviour they have today" contract at line 6499. #stopChaseSession has already run by then, so the session is left dead-but-resting and the caller never learns the cancel failed.
Moving it inside the existing try covers it.

abretonc7s: Fixed in `5df7885e9a`. Child-ID Chase cancellation now runs inside the existing error boundary. A regression verifies trading-setup failure resolves to `{ success: false, error }` instead of rejecting.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
abretonc7s: Fixed in `5df7885e9a`. Snapshot refresh now persists the live remainder only while the same child still owns the session. The regression reads twice inside the cache window and verifies both return `0.2`.

Bigshmow (approved):

PR #9971, opened 2026-08-26
fix(perps): align HIP-3 market contexts

abretonc7s:
## Explanation

Selective HIP-3 allowlists filtered the market universe without filtering the parallel asset-context array. Markets such as `io:ANTH` then inherited volume, open-interest, funding, and previous-price data from another market.

This change filters each market together with its same-index context. It also displays tiny nonzero funding rates as `<0.0001%` or `-<0.0001%` instead of `0.0000%`.

## References

- Jira: https://consensyssoftware.atlassian.net/browse/TAT-3841
- Mobile consumer PR: https://github.com/MetaMask/metamask-mobile/pull/35312

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes market-overview data correctness for allowlisted HIP-3 symbols and funding-rate UI text; trading logic is untouched, but wrong stats could mislead users before this fix.
> 
> **Overview**
> Fixes **wrong market stats on selectively allowlisted HIP-3 markets** by filtering each DEX’s `universe` entry together with its same-index `assetCtx` in `#mergeDexResultsInto`, instead of dropping markets from the list while still appending the full context array (which could attach another symbol’s volume, open interest, funding, and prior price to markets like `io:ANTH`).
> 
> Updates **`formatFundingRate`** so tiny nonzero rates that round to `0.0000%` at four decimal places render as **`<0.0001%`** or **`-<0.0001%`**, signaling activity below display precision rather than a flat zero.
> 
> Changelog and tests cover the alignment regression and the new funding threshold strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7b23c0180ed88a984aee660acb1ebe215d3715a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/perps-controller/tests/src/utils/perpsFormatters.test.ts
+++ b/packages/perps-controller/tests/src/utils/perpsFormatters.test.ts
@@ -283,6 +283,22 @@ describe('formatFundingRate', () => {
     expect(formatFundingRate(0)).toBe('0.0000%');
   });
 
+  it('shows a threshold for positive rates below display precision', () => {
+    const value = 0.000000059;
+
+    const result = formatFundingRate(value);
+
+    expect(result).toBe('<0.0001%');
+  });
+
+  it('shows a threshold for negative rates below display precision', () => {
+    const value = -0.000000059;
+
+    const result = formatFundingRate(value);
+
+    expect(result).toBe('>-0.0001%');
+  });
+
   it('returns empty string for undefined when showZero is false', () => {
     expect(formatFundingRate(undefined, { showZero: false })).toBe('');
   });
--- a/packages/perps-controller/tests/src/utils/perpsFormatters.test.ts
+++ b/packages/perps-controller/tests/src/utils/perpsFormatters.test.ts
@@ -296,7 +296,7 @@ describe('formatFundingRate', () => {
 
     const result = formatFundingRate(value);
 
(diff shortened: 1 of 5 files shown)


packages/perps-controller/tests/src/utils/perpsFormatters.test.ts
```diff
@@ -283,6 +283,22 @@ describe('formatFundingRate', () => {
...
+
+    expect(result).toBe('<0.0001%');
+  });
+
+  it('shows a threshold for negative rates below display precision', () => {
+    const value = -0.000000059;
+
+    const result = formatFundingRate(value);
+
+    expect(result).toBe('>-0.0001%');
```
michalconsensys: This may need to be doublechecked with product, as `>-0.0001` could mean any positive number

abretonc7s: Good catch. Funding rates can be negative, but `>-0.0001%` was ambiguous. The shared formatter now uses `-<0.0001%` to show a negative rate whose magnitude is below the display threshold. Tests and the changelog were updated in a7b23c0180.

michalconsensys (approved):

PR #9982, opened 2026-08-27
fix: reuse adaptOrderFromSDK for historical orders

geositta:
## Explanation

Hyperliquid historical orders were mapped through a separate inline path that dropped `triggerPx` and the normalized trigger order type. Consumers therefore received trigger orders without `Order.triggerPrice`, preventing clients such as MetaMask Mobile from displaying the trigger price in order history.

This change reuses the existing `adaptOrderFromSDK` mapping used for open orders, then overlays historical lifecycle fields such as status, timestamps, and filled/remaining size. It also preserves ordinary historical market-order classification when Hyperliquid supplies a non-empty `limitPx` as a slippage cap.

Provider tests now cover distinct trigger and limit prices for Take Profit Limit and Stop Market orders, plus a regular market order carrying a slippage-cap price. There are no public type changes or breaking changes.

Validation:
- `yarn workspace @metamask/perps-controller run test`
- `yarn eslint packages/perps-controller/src/providers/HyperLiquidProvider.ts packages/perps-controller/tests/src/providers/HyperLiquidProvider.history.test.ts`
- `yarn build`

## References

- Related to [MetaMask/metamask-mobile#35118](https://github.com/MetaMask/metamask-mobile/pull/35118)
- Addresses [the production history boundary review finding](https://github.com/MetaMask/metamask-mobile/pull/35118#discussion_r3868602007)

## Checklist

---

> [!NOTE]
> **Low Risk**
> Localized to HyperLiquid historical order mapping and adapter edge cases; no public API or breaking type changes, with expanded unit tests.
> 
> **Overview**
> **HyperLiquid order history** now goes through the same `adaptOrderFromSDK` path as open orders, instead of a separate inline mapper that dropped trigger metadata.
> 
> Historical rows still get lifecycle overlays (`status`, `statusTimestamp` / `lastUpdated`, and `remainingSize` from the exchange’s current `sz`). **`orderType`** for history is derived from HyperLiquid’s detailed type via `HISTORICAL_ORDER_TYPE_BY_DETAILED_TYPE`, so trigger and market orders are not mislabeled as limits when `limitPx` is only a slippage cap.
> 
> **`adaptOrderFromSDK`** tolerates missing `oid`, `orderType`, and empty `limitPx` so partial history payloads still map instead of failing silently.
> 
> Tests cover trigger/limit price preservation for TP/SL history, market orders with slippage-cap prices, and the detailed-type → limit/market mapping table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdd11cd0639a4f9fd143d2fa546f159f371b48c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/perps-controller/src/providers/HyperLiquidProvider.ts
+++ b/packages/perps-controller/src/providers/HyperLiquidProvider.ts
@@ -10535,9 +10535,7 @@ export class HyperLiquidProvider implements PerpsProvider {
         const adaptedOrder = adaptOrderFromSDK(order, undefined);
         let historicalOrderType = adaptedOrder.orderType;
         if (!adaptedOrder.triggerOrderType) {
-          historicalOrderType = order.orderType
-            ?.toLowerCase()
-            .includes('limit')
+          historicalOrderType = order.orderType?.toLowerCase().includes('limit')
             ? 'limit'
             : 'market';
         }
--- a/packages/perps-controller/src/providers/HyperLiquidProvider.ts
+++ b/packages/perps-controller/src/providers/HyperLiquidProvider.ts
@@ -10493,8 +10493,6 @@ export class HyperLiquidProvider implements PerpsProvider {
       // Transform HyperLiquid orders to abstract Order type
       const orders: Order[] = (rawOrders || []).map((rawOrder) => {
         const { order, status, statusTimestamp } = rawOrder;
-        // Normalize side: HyperLiquid uses 'A' (Ask/Sell) and 'B' (Bid/Buy)
-        const normalizedSide = order.side === 'B' ? 'buy' : 'sell';
 
         // Normalize status
         let normalizedStatus: Order['status'];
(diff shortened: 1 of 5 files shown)


packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -10534,24 +10532,24 @@ export class HyperLiquidProvider implements PerpsProvider {
...
           originalSize: order.origSz || order.sz,
-          price: order.limitPx || '0',
           filledSize: filledSize.toString(),
           remainingSize: currentSize.toString(),
           status: normalizedStatus,
           timestamp: statusTimestamp,
           lastUpdated: statusTimestamp,
-          detailedOrderType: order.orderType, // Full order type from exchange (e.g., 'Take Profit Limit', 'Stop Market')
-          isTrigger: order.isTrigger,
-          reduceOnly: order.reduceOnly,
```
michalconsensys, line 10554: Are these not needed anymore? Is it a breaking change if these get removed?

geositta: Removing them is not a breaking change, and they are still needed, they're just now supplied by adaptOrderFromSDK instead of being spelled out inline. hyperLiquidAdapter.ts:283-285, 362-364 assigns exactly the same three values from exactly the same raw fields:

  const detailedOrderType = rawOrder.orderType;
  const { isTrigger } = rawOrder;
  const { reduceOnly } = rawOrder;

The ...adaptedOrder spread carries them through unchanged, and the existing assertions at
HyperLiquidProvider.history.test.ts:640-673 still assert all three on the historical orders — they pass. I walked every field of the returned object against main and the shape is a strict superset: same values for the 15 pre-existing keys, plus newly populated triggerPrice, triggerOrderType, and (when present) takeProfitPrice/stopLossPrice/isPositionTpsl/strategyGroupId. Additive, so consumers pinned to the old shape keep working — which is what the mobile branch relies on (transactionTransforms.ts reads triggerOrderType and triggerPrice while still reading isTrigger and
detailedOrderType).

michalconsensys (dismissed):

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -10529,29 +10527,22 @@ export class HyperLiquidProvider implements PerpsProvider {
...
-        const originalSize = parseFloat(order.origSz || order.sz);
-        const currentSize = parseFloat(order.sz);
-        const filledSize = originalSize - currentSize;
+        const adaptedOrder = adaptOrderFromSDK(order, undefined);
+        // limitPx is also populated as a slippage cap for market orders, so the
+        // exchange's detailed type is the reliable execution-mode source.
+        const historicalOrderType = order.orderType
+          ?.toLowerCase()
+          .includes('limit')
+          ? 'limit'
```
abretonc7s: I wonder if we should now match the full order type here or if there could be issue with includes

geositta:
@metamaskbot publish-previews

packages/perps-controller/src/providers/HyperLiquidProvider.ts
geositta: Implemented exact order type mapping.

Covers all six HyperLiquid order types with a typed map.
Unknown/malformed values safely default to market.
Added table driven coverage, including Unexpected Limit to prevent substring classification.

abretonc7s (approved):

PR #9984, opened 2026-08-27
feat(notification-services-controller): add support for notifications categories

zelkibuilds:
## Explanation

**Current state:** The notifications API exposes `GET /api/v4/notifications/categories`, a server-driven manifest describing the notification-settings taxonomy (category IDs, platform visibility, and AUS preference keys). `NotificationServicesController` had no way to consume it. The internal `makeApiCall` helper only supported authenticated `POST`/`DELETE` requests with a positional body argument, making unauthenticated `GET` calls impossible to express without a workaround.

**Solution:**

- Reworked `makeApiCall` from positional args to an options object using a discriminated union: `GET`/`DELETE` accept an optional `bearerToken` and no body; `POST` requires a `body` and accepts an optional `bearerToken`. This makes missing-body/unexpected-body misuse a compile error and enables unauthenticated `GET` calls. `makeApiCall` is package-internal (not re-exported from the package index), so this is not a breaking change to consumers.

- Added `getNotificationsCategories()` service function that calls the new unauthenticated `GET /api/v4/notifications/categories` endpoint. Returns `[]` on HTTP or transport errors.

- Added two new state fields to `NotificationServicesControllerState`:
  - `metamaskNotificationsCategories: NotificationsCategory[]` — persisted, included in state logs/debug snapshots, used in UI.
  - `isFetchingMetamaskNotificationsCategories: boolean` — UI-only loading flag, not persisted; reset in `finally` so it cannot get stuck.

- Added public `fetchMetamaskNotificationsCategories()` controller method that fetches the categories manifest, stores it in state, and returns it. Exposed through the messenger as `NotificationServicesController:fetchMetamaskNotificationsCategories` (added to `MESSENGER_EXPOSED_METHODS`).

- Updated the notification-api schema types (`schema.ts`) to include the `/api/v4/notifications/categories` endpoint, `NotificationCategory`, and `NotificationsCategory` schemas. Updated mocks accordingly.

- Replaced the deprecated `ControllerStateChangeEvent` import with `ControllerStateChangedEvent`; the exported type alias `NotificationServicesControllerStateChangeEvent` is unchanged, so there is no breaking change for consumers.

## Checklist

### DIFF
--- a/packages/notification-services-controller/CHANGELOG.md
+++ b/packages/notification-services-controller/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `fetchMetamaskNotificationsCategories()` controller method that fetches the notification categories manifest from `GET /api/v4/notifications/categories`, stores the result in new `metamaskNotificationsCategories` state, and exposes it through the messenger as `NotificationServicesController:fetchMetamaskNotificationsCategories` ([#9984](https://github.com/MetaMask/core/pull/9984))
+- Add `metamaskNotificationsCategories` state field (`NotificationsCategory[]`) for the server-driven notification settings taxonomy ([#9984](https://github.com/MetaMask/core/pull/9984))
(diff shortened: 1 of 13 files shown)


zelkibuilds:
@metamaskbot publish-previews

PR #9990, opened 2026-08-27
refactor(platform-api-docs): use ts-morph globbing for file discovery

cryptodev-2s:
Stacked on #9913, targets its branch.

Replaces hand-rolled file discovery with `Project#addSourceFilesAtPaths`, which takes glob patterns directly. Drops `discovery.ts`, the per-file add/extract loops, `listTargetSubdirectories`, and the `glob` dependency. Also merges the two `Project` factories (they differed by one field) and removes the redundant `getSourceFile()` guard, closing https://github.com/MetaMask/core/pull/9913#discussion_r3857912028.

Net 254 lines deleted. Generation over this monorepo: **114s to 8.6s**.

Output is byte identical. `diff -r` on full docs trees generated before and after, for core, mobile and extension: all empty, sidebars included.

## Two traps, if you touch these patterns

Both produced wrong output while the whole suite stayed green.

- **Glob roots need `realpath`.** The matcher does not follow a symlinked *ancestor*. Under `/var/folders/...` the declaration glob returned 0 where `/private/var/...` returned 1980, so any project under `/tmp` or `/var` would have emitted empty docs.
- **Negations must be anchored.** `!**/*.test.ts` resolves against the process cwd, not against the pattern it accompanies. Since `yarn workspace` sets cwd to the package dir, every exclusion silently stopped matching and test fixtures leaked in (117 namespaces instead of 112).

Related: declaration-file exclusions are deliberately narrower than the source ones. A blanket `dist/**` matches the segment those files live under and drops all 1980.

## Behaviour change

A broken symlink used to emit a per-file parse warning. The matcher now resolves it away before ts-morph sees it, so it is skipped silently; valid files are still documented. One test rewritten to match.

## Follow-up

Second step makes each strategy self contained, which is what https://github.com/MetaMask/core/pull/9913#discussion_r3857882683 asked for. Worth knowing before that review: `extraction.ts` is a scan-only front end welded onto a shared core, so the split has to cut through it too, not just `generate.ts`.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Internal tooling refactor with byte-identical output claimed; main risk is glob/exclusion edge cases on unusual filesystem layouts, now covered by new tests.
> 
> **Overview**
> **Speeds up doc generation** by replacing per-file `glob` discovery and `addSourceFileAtPath` loops with a single shared ts-morph `Project` that bulk-loads matches via `addSourceFilesAtPaths`. Removes `discovery.ts`, the `glob` dependency, and merges the two project factories into shared `createProject()` in `ts-project.ts`.
> 
> The scan strategy now builds **resolved, forward-slashed glob roots** (`toGlobPath` + `realpath`) and **content-root–anchored exclusion negations** so test/dist skips work when cwd ≠ project root and monorepo packages named `test`/`dist` are not excluded wholesale. Added source files are **sorted by path** for stable dedup/source links; extraction uses a **realpath’d project root** for relative source paths.
> 
> **Behavior:** unreadable paths (e.g. broken symlinks) are skipped without per-file parse warnings; valid files still document. Tests cover the `packages/test` exclusion edge case and the symlink case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55c4a26b9ff023f7d83adc07a454900d6abfa05b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/platform-api-docs/src/generate.ts
```diff
@@ -253,71 +250,90 @@ function logScanPlan(sources: ScanSources): void {
...
+ * `!**‍/*.test.ts`. A matcher resolves an unanchored negation against the
+ * process's working directory, not against the pattern it accompanies, so an
+ * unanchored exclusion silently stops excluding anything the moment the scanned
+ * path falls outside the working directory — which is the normal case, since
+ * this runs from wherever the consumer invoked it.
+ *
+ * @param root - Resolved directory the positive pattern is rooted at.
+ * @returns The exclusion patterns.
+ */
+function tsSourceExclusions(root: string): string[] {
```
mcmire: Since this is a function, should its name start with a verb?

``` suggestion
function buildTsSourceExclusions(root: string): string[] {

File: packages/platform-api-docs/src/generate.ts [hunk 2]
```diff
@@ -253,71 +250,90 @@ function logScanPlan(sources: ScanSources): void {
...
+ * vendored inside a package's own `dist`.
+ *
+ * Deliberately narrower than {@link tsSourceExclusions}. A blanket `dist/**`
+ * exclusion would match the very `dist` segment these files live under and
+ * silently drop every one of them.
+ *
+ * @param root - Resolved `node_modules/@metamask` directory.
+ * @returns The exclusion patterns.
+ */
+function declarationFileExclusions(root: string): string[] {
```
mcmire:
``` suggestion
function buildDeclarationFileExclusions(root: string): string[] {

Comment (cryptodev-2s) [reply packages/platform-api-docs/src/generate.ts]: Make sense applied here https://github.com/MetaMask/core/pull/9990/changes/f491c56e2f217223dd08158633da723a7f15f54f

File: packages/platform-api-docs/src/generate.ts (hunk 1 above)
Comment (cryptodev-2s) [reply packages/platform-api-docs/src/generate.ts]: Make sense applied here https://github.com/MetaMask/core/pull/9990/changes/f491c56e2f217223dd08158633da723a7f15f54f

File: packages/platform-api-docs/src/generate.ts [hunk 3]
```diff
@@ -306,56 +355,60 @@ async function scanSources(
...
-    const srcDirs = await listTargetSubdirectories(
-      sources.packagesDir,
-      'src',
-      false,
+    const root = await toGlobPath(sources.packagesDir);
+    sourceFiles.push(
+      ...addSourceFiles(project, [
+        `${root}/*/src/**/*.ts`,
+        ...tsSourceExclusions(root),
+      ]),
```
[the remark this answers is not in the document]
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/9990/commits/13b1ee85dfc89c6db2a1c51304a67e21c7625a83 even though naming a package `test` or `dist` is unlikely.

packages/platform-api-docs/src/generate.ts
```diff
@@ -253,71 +250,103 @@ function logScanPlan(sources: ScanSources): void {
...
+ *
+ * Deliberately narrower than {@link buildTsSourceExclusions}. A blanket `dist/**`
+ * exclusion would match the very `dist` segment these files live under and
+ * silently drop every one of them.
+ *
+ * @param root - Resolved `node_modules/@metamask` directory.
+ * @returns The exclusion patterns.
+ */
+function buildDeclarationFileExclusions(root: string): string[] {
+  return [`!${root}/*/dist/**/node_modules/**`];
```
Mrtenz: Not sure I follow this one. Why would a `dist` folder contain `node_modules`?

packages/platform-api-docs/src/generate.ts
```diff
@@ -253,71 +250,103 @@ function logScanPlan(sources: ScanSources): void {
...
-    if (await directoryExists(candidate)) {
-      existing.push(candidate);
-    }
-  }
-  return existing;
+async function toGlobPath(...segments: string[]): Promise<string> {
+  // Safe to resolve without a fallback: `discoverScanSources` has already
+  // confirmed every directory reaching this point exists.
+  const resolved = await fs.realpath(path.join(...segments));
+  return resolved.replace(/\\/gu, '/');
```
Mrtenz, line 334: Should this use `path.posix.normalize`?

packages/platform-api-docs/src/generate.ts
```diff
@@ -335,56 +368,63 @@ async function scanSources(
...
+  for (const sourceFile of sourceFiles) {
+    try {
+      allItems.push(...extractFromSourceFile(sourceFile, resolvedProjectPath));
+    } catch (error) {
+      // istanbul ignore next: defensive. Files that can't be read or parsed
+      // are dropped by the matcher before they reach here, so this only
+      // catches a file whose types defeat the extractor — worth surviving
+      // when scanning thousands of files, but not reproducible in a test.
+      console.warn(
+        `Warning: failed to parse ${path.relative(resolvedProjectPath, sourceFile.getFilePath())}`,
```
Mrtenz: Nit:
```suggestion
        `Warning: Failed to parse ${path.relative(resolvedProjectPath, sourceFile.getFilePath())}`,
```

Mrtenz (changes requested):

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Defensive, I can remove it

packages/platform-api-docs/src/generate.ts
cryptodev-2s: `path.posix.normalize` treats backslashes as ordinary characters, so it wouldn't convert anything which could break Windows users.

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Updated here https://github.com/MetaMask/core/pull/9990/commits/c05f1143fda99b858d08a58dfc94df27fd57be27

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/9990/commits/c05f1143fda99b858d08a58dfc94df27fd57be27

packages/platform-api-docs/src/generate.ts
```diff
@@ -327,6 +341,10 @@ async function listTargetSubdirectories(
  * declaration in one file walking through an imported umbrella union into
  * an auto-generated `*-method-action-types.ts` sibling).
  *
+ * Locations are scanned in the order the previous implementation used — scan
+ * directories, then workspace packages, then published declaration files — as
+ * deduplication resolves ties in favour of whichever item it saw first.
```
mcmire: In the future we won't remember how this used to be implementated, so maybe we don't need to reference that? Also I'm not sure where the deduplication code is but it doesn't seem to be here 🤔 

``` suggestion
 * Scan every source location described by `sources` (scan directories first,
 * then workspace packages, then published declaration files) and return all
 * extracted messenger items.
 *
 * A single ts-morph Project is shared across every file so the type checker can
 * resolve cross-file references (e.g. a `*Messenger` declaration in one file
 * walking through an imported umbrella union into an auto-generated
 * `*-method-action-types.ts` sibling).
```

packages/platform-api-docs/src/generate.ts
```diff
@@ -335,56 +353,58 @@ async function scanSources(
...
+      // istanbul ignore next: defensive. Files that can't be read or parsed
+      // are dropped by the matcher before they reach here, so this only
+      // catches a file whose types defeat the extractor — worth surviving
+      // when scanning thousands of files, but not reproducible in a test.
+      console.warn(
+        `Warning: Failed to parse ${path.relative(resolvedProjectPath, sourceFile.getFilePath())}`,
       );
+      // istanbul ignore next: see above.
+      console.warn(error);
     }
```
mcmire: Hmm. Since it sounds like this error will be rare, maybe we don't want to swallow it? If there is a bug then we might want to know about it so we can fix it immediately. What are your thoughts?

``` suggestion
    allItems.push(...extractFromSourceFile(sourceFile, resolvedProjectPath));
```

packages/platform-api-docs/src/root-messenger-discovery.ts
```diff
@@ -131,11 +109,11 @@ function resolveMessengerCapabilitiesTypeReference({
...
   const absolutePath = path.resolve(projectPath, reference.filePath);
 
+  // `addSourceFileAtPath` is idempotent: the two references often name the same
+  // file, and the second call returns the source file added by the first.
   let sourceFile;
   try {
-    sourceFile =
-      project.getSourceFile(absolutePath) ??
-      project.addSourceFileAtPath(absolutePath);
+    sourceFile = project.addSourceFileAtPath(absolutePath);
```
mcmire: Nit: Should this comment be placed above the line it refers to?

``` suggestion
  let sourceFile;
  try {
    // `addSourceFileAtPath` is idempotent: the two references often name the same
    // file, and the second call returns the source file added by the first.
    sourceFile = project.addSourceFileAtPath(absolutePath);

File: packages/platform-api-docs/src/generate.ts [hunk 10]
```diff
@@ -335,56 +353,58 @@ async function scanSources(
...
-    for (const distDir of distDirs) {
-      allItems.push(
-        ...(await extractFromDirectory(
-          project,
-          distDir,
-          projectPath,
-          findDtsFiles,
-        )),
+    const root = await toGlobPath(sources.nodeModulesDir);
+    sourceFiles.push(...addSourceFiles(project, [`${root}/*/dist/**/*.d.cts`]));
```
mcmire, line 383: Do we have to call `addSourceFiles` multiple times? Can we assemble a big list of all of the source file paths we want to add and then add them in one call? Would that be even faster?

mcmire (commented):
Tried this out and it definitely speeds things up. Nice work! Just had a few comments below.

packages/platform-api-docs/src/root-messenger-discovery.ts
cryptodev-2s: Yes moved here https://github.com/MetaMask/core/pull/9990/commits/960cf509a4bd9ec45ce46130b8b870091b136323

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Make sense it would throw now after https://github.com/MetaMask/core/pull/9990/commits/960cf509a4bd9ec45ce46130b8b870091b136323

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Applied here https://github.com/MetaMask/core/pull/9990/commits/960cf509a4bd9ec45ce46130b8b870091b136323

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Tried it, it is faster (~2.6s → ~1.9s). Problem is one call sorts globally, so all the `node_modules` `.d.cts` files land ahead of `packages/*/src`, and dedup keeps whichever it sees first. 1022 capabilities lose their source link.

Real cause is the dedup scoring doesn't know source from build output. Fixing that makes the ordering irrelevant and this safe, so I'll do both in a follow up.

mcmire: Okay no problem. The current speedups are pretty good as is, so I'm happy with them. No further changes needed for this PR with respect to optimization.

mcmire (approved):
LGTM!

Mrtenz (approved):

PR #9976, opened 2026-08-26
chore!: Drop Node 18 and 20

Mrtenz:
Node 18 and 20 are EOL, we should not support them anymore. This PR makes a breaking change to every single package in the monorepo. This also reduces the number of jobs we need to run in CI.

This change was originally merged into the `esm` feature branch, but using a [stacked pull request](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs) seems cleaner here.

---

> [!NOTE]
> **Medium Risk**
> Semver-breaking `engines` change affects every package consumer and downstream apps still on Node 18/20; test-only crypto polyfill removal assumes CI/dev on Node 22+.
> 
> **Overview**
> **Breaking:** The monorepo no longer supports Node 18 or 20. Root and every package’s `engines.node` is now `^22.14.0 || ^24`, with matching **BREAKING** changelog entries across published packages.
> 
> CI runs fewer jobs: the prepare matrix and per-package test jobs only use **22.x** and **24.x** (the old `test-18` / `test-20` jobs are removed), and wallet-cli e2e drops Node 20 from its matrix. `@types/node` is bumped from 16.x to **^22.13.14** at the root and in packages that depended on the old types; `yarn.config.cjs` enforces the new engines string uniformly (including `@metamask/wallet-cli`, which previously allowed `>=20`).
> 
> Because Node 22 exposes Web Crypto globally, custom Jest environments and test `beforeAll` hooks that polyfilled `globalThis.crypto` are deleted or trimmed in packages such as **keyring-controller**, **passkey-controller**, **seedless-onboarding-controller**, **money-account-upgrade-controller**, and **wallet** tests. Related `jest-environment-node` devDependencies are dropped where they existed only for those environments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f27e444eec8bce236bdb00ceeddb183828e5abec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9994, opened 2026-08-27
feat: add script to auto-resolve CHANGELOG.md merge conflicts

Mrtenz:
## Explanation

Every consumer-facing change requires a `packages/*/CHANGELOG.md` entry under `## [Unreleased]`, and since many PRs land against `main` around the same time, that section conflicts constantly — usually because two branches added different bullet points to the same category, not because of any real disagreement about content. Resolving these by hand is repetitive and error-prone (easy to accidentally drop an entry).

This adds a `yarn changelog:merge` script that finds `packages/*/CHANGELOG.md` files with unresolved Git conflicts and computes the union of entries from both sides, using `@metamask/auto-changelog`'s parsing/stringification (`parseChangelog`/`Changelog`) rather than hand-rolling Markdown parsing, so category ordering (`Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`) and formatting stay correct.

Merge behaviour:
- Reads both sides of the conflict via `git show :2:<path>` (ours) and `git show :3:<path>` (theirs) — no manual conflict-marker parsing.
- Deduplicates entries by PR number (falling back to description) so an entry added by both sides isn't duplicated.
- Inserts new `**BREAKING:**` entries below any existing breaking entries in a category, and other new entries at the end.
- Inserts a release version that exists on only one side into its correct descending-SemVer position.
- Formats the result with `oxfmt`, matching the convention used by `scripts/update-changelog.sh`.
- Files it can't confidently merge (e.g. a structurally invalid side) are skipped and left with their conflict markers intact, and the script exits non-zero so it's clear something needs manual attention.

## References

* Related to `docs/processes/updating-changelogs.md`

## Checklist

---

> [!NOTE]
> **Low Risk**
> Dev-only merge helper with no runtime or published package behavior; main risk is incorrect changelog output during merges, mitigated by validation and skip-on-parse-failure.
> 
> **Overview**
> Adds **`yarn changelog:merge`**, a maintainer tool that finds conflicted `packages/*/CHANGELOG.md` files and merges them automatically instead of hand-editing conflict markers.
> 
> The implementation loads **ours** and **theirs** via Git stages (`:2` / `:3`), parses both sides with **`@metamask/auto-changelog`**, unions unique entries (deduped by PR + description), applies breaking-entry ordering and SemVer release ordering, then writes the formatted result back **unstaged**. Unparseable files are skipped with a non-zero exit; the CLI reminds you to run **`yarn changelog:validate`**.
> 
> Root **`package.json`** gains the script and **`@metamask/auto-changelog`** as a dev dependency, with library + CLI code and broad Jest coverage in **`scripts/lib/changelog-conflicts.ts`** and **`scripts/merge-changelog-conflicts.ts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95f1c80d4c2df7e8c7c6ba2dcc2833eb1c858ee9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,371 @@
...
+      theirs.addRelease(oursRelease);
+      const releases = theirs.getReleases();
+      const [inserted] = releases.splice(0, 1);
+
+      const sortedIndex = releases.findIndex(({ version }) =>
+        gt(inserted.version, version),
+      );
+
+      const insertIndex = sortedIndex === -1 ? releases.length : sortedIndex;
+      releases.splice(insertIndex, 0, inserted);
```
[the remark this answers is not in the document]
Mrtenz: `addToStart: true` is the default.

scripts/merge-changelog-conflicts.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Files that can't be automatically merged are left with their conflict
+ * markers intact, and cause the script to exit with a non-zero code.
+ *
+ * Usage: `tsx scripts/merge-changelog-conflicts.ts`
+ */
+export async function main(): Promise<void> {
+  const { resolved, skipped } = await resolveChangelogConflicts();
+
+  if (resolved.length === 0 && skipped.length === 0) {
+    console.log('No conflicted CHANGELOG.md files found.');
```
FrederikBolding:
```suggestion
    console.log('No CHANGELOG.md files with conflicts found.');
```
nit

Mrtenz: Applied your suggestion!

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+        repoUrl,
+        tagPrefix,
+      });
+
+      await fs.writeFile(
+        path.join(ROOT_WORKSPACE, changelogPath),
+        content,
+        'utf8',
+      );
+      await execa('git', ['add', changelogPath], { cwd: ROOT_WORKSPACE });
```
FrederikBolding: Unsure we want this to actively stage files 🤔

Mrtenz: How come? The idea behind the script is that if you run into a merge conflict, you just run `yarn changelog:merge` and then you can `git rebase --continue` (assuming no other conflicts).

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+  try {
+    const content = await fs.readFile(
+      path.join(ROOT_WORKSPACE, packageJsonPath),
+      'utf8',
+    );
+
+    return JSON.parse(content);
+  } catch {
+    const content = await readGitBlob(':2', packageJsonPath);
+    return JSON.parse(content);
```
FrederikBolding, line 139: Should we just blow up instead of this? Unsure of what usecase this has 🤔

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+ * @param changelogPath - The repo-relative path of the changelog file.
+ * @returns The package's name, repository URL, and changelog tag prefix.
+ */
+export async function resolvePackageMetadata(
+  changelogPath: string,
+): Promise<PackageMetadata> {
+  const packageJsonPath = path.posix.join(
+    path.posix.dirname(changelogPath),
+    'package.json',
+  );
```
FrederikBolding, line 156: Why are we using `path.posix` here?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+  // `ours` is the current branch (HEAD); during a rebase, it's the upstream
+  // branch being rebased onto. In both cases, that's the side more likely to
+  // already contain entries also present in `theirs`, so preserving its
+  // existing order (and only appending genuinely new entries from `theirs`)
+  // produces more intuitive results than the reverse.
+  const ours = parseChangelog({
+    changelogContent: oursContent,
+    repoUrl,
+    tagPrefix,
+    formatter: oxfmt,
```
FrederikBolding, line 323: How come we need to specify a formatter for parsing?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+  theirsContent,
+  repoUrl,
+  tagPrefix,
+}: {
+  oursContent: string;
+  theirsContent: string;
+  repoUrl: string;
+  tagPrefix: string;
+}): Promise<{ content: string; mergedEntryCount: number }> {
+  // `ours` is used as the base to mutate and stringify. During a Git merge,
```
FrederikBolding: I find the `ours`/`theirs` versus `base`/`ìncoming`naming differences across the file a bit confusing.

scripts/lib/changelog-conflicts.ts
FrederikBolding: Maybe just a personal preference then but I think it avoids some cases where you want to make further changes before staging and proceeding with rebase.

scripts/lib/changelog-conflicts.ts
Mrtenz: This is used when stringifying the changelog again (`ours.toString()`).

scripts/lib/changelog-conflicts.ts
Mrtenz: Yeah, good point. `ours`/`theirs` is Git terminology, but `base`/`incoming` may be easier to understand. I can rename everything to the latter if you're good with that.

FrederikBolding: Sounds good 👍

scripts/lib/changelog-conflicts.ts
Mrtenz: Since Git itself always uses POSIX-style paths. It's turned into an OS-dependent path before reading with `fs` in `path.join(ROOT_WORKSPACE, packageJsonPath)`.

scripts/lib/changelog-conflicts.ts
Mrtenz: It handles a case where `package.json` itself has conflicts (so `JSON.parse` fails), falling back to "ours" if it does.

scripts/lib/changelog-conflicts.ts
Mrtenz: Hmm, usually in a rebase you shouldn't need to make additional changes. But if you do, you can just stage it again?

I can remove it if you prefer though.

scripts/lib/changelog-conflicts.ts
FrederikBolding: Sure, but why not use a regular join which normalizes right away?

Mrtenz: Because it's potentially passed to `readGitBlob` which needs a Git POSIX-style path.

scripts/lib/changelog-conflicts.ts
FrederikBolding: Yeah, but is that likely? It adds a bunch of complexity, so I wonder if it ever would come up 😄

Mrtenz: Seems pretty likely, yeah. If a change on main bumps a dependency and adds a changelog entry for those, and you do the same in your branch, there can easily be a conflict for both files.

scripts/lib/changelog-conflicts.ts
FrederikBolding: Personally I would prefer if the script didn't mutate Git state yeah, just in case.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,376 @@
...
+      });
+
+      await fs.writeFile(
+        path.join(ROOT_WORKSPACE, changelogPath),
+        content,
+        'utf8',
+      );
+      await execa('git', ['add', changelogPath], { cwd: ROOT_WORKSPACE });
+
+      resolved.push({ path: changelogPath, mergedEntryCount });
```
FrederikBolding: Should the returned paths be normalized?

Mrtenz: It returns the path as used by Git. Do you think it should return OS-style paths instead? That would break when using Git on Windows with these paths.

FrederikBolding: Hmm, I was thinking these would get logged to the terminal and might not be clickable for example if they were in the wrong format. Unsure how Windows deals with that actually

scripts/lib/changelog-conflicts.ts
Mrtenz: Hmm, unsure about renaming everything, since "base" in Git means something different than "ours" so that could potentially be confusing too. Right now it uses "ours" and "theirs" in places where dealing with Git, and "base" and "incoming" in tool-agnostic merge logic.

scripts/lib/changelog-conflicts.ts
Mrtenz: Yeah, good point. I'm not able to test it (don't have a Windows dev machine), but can just add normalisation and see if anyone runs into issues. 😅

FrederikBolding: Looks like it works either way, probably not worth the trouble then

Mrtenz: Just pushed. Should I revert then?

FrederikBolding: Now that you've done it, I guess we can just keep it 😄

scripts/lib/changelog-conflicts.ts
FrederikBolding: Yeah, I'm not quite sure either. But the current naming is confusing as well IMO

FrederikBolding (dismissed):

scripts/merge-changelog-conflicts.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The entrypoint to this script.
+ *
+ * Automatically resolves Git merge conflicts in `packages/*\/CHANGELOG.md`
+ * files by taking the union of entries added on each side of the conflict.
+ * Files that can't be automatically merged are left with their conflict
+ * markers intact, and cause the script to exit with a non-zero code.
+ *
+ * Usage: `tsx scripts/merge-changelog-conflicts.ts`
```
mcmire, line 16: Side note: Once we merge https://github.com/MetaMask/core/pull/9976 then we can start on the path of removing `tsx` entirely, since Node now supports reading TypeScript files directly (we will have to enable the `erasableTypesOnly` option, which will force us to remove enums).

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+import { promises as fs } from 'fs';
+import path from 'path';
+import { gt } from 'semver';
+
+export const ROOT_WORKSPACE = path.resolve(__dirname, '../..');
+
+const CHANGELOG_PATH_PATTERN = /^packages\/[^/]+\/CHANGELOG\.md$/u;
+
+/**
+ * A regular expression for breaking entries. Assumes the entry is in the
```
mcmire, line 14: We don't strictly enforce this, but there are a few places where we assume this in `auto-changelog`, so it's probably a safe assumption here too.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+ * since it's normally not part of the conflict; if it can't be parsed there
+ * (e.g. it's also mid-merge and contains conflict markers), it's re-read
+ * from the "ours" conflict stage instead.
+ *
+ * @param packageJsonPath - The repo-relative path of the `package.json` file.
+ * @returns The parsed `package.json` contents.
+ */
+async function readPackageJson(packageJsonPath: string): Promise<PackageJson> {
+  try {
+    const content = await fs.readFile(
```
mcmire: Nit: There is also `readJsonFile` in `@metamask/utils` which sort of does this already.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+ * Resolve the package name and repository URL for the package that owns the
+ * given changelog file.
+ *
+ * @param changelogPath - The repo-relative path of the changelog file.
+ * @returns The package's name, repository URL, and changelog tag prefix.
+ */
+export async function resolvePackageMetadata(
+  changelogPath: string,
+): Promise<PackageMetadata> {
+  const packageJsonPath = path.posix.join(
```
mcmire, line 153: I've never used `path.posix.join` before. I always used `path.join`. What's the difference?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+  repoUrl: string;
+  tagPrefix: string;
+}): Promise<{ content: string; mergedEntryCount: number }> {
+  // `ours` is used as the base to mutate and stringify. During a Git merge,
+  // `ours` is the current branch (HEAD); during a rebase, it's the upstream
+  // branch being rebased onto. In both cases, that's the side more likely to
+  // already contain entries also present in `theirs`, so preserving its
+  // existing order (and only appending genuinely new entries from `theirs`)
+  // produces more intuitive results than the reverse.
+  const ours = parseChangelog({
```
mcmire: Nit: Maybe `ourChangelog` / `theirChangelog`?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+ * side.
+ * @param options.theirsContent - The changelog content on the "theirs"
+ * conflict side.
+ * @param options.repoUrl - The GitHub repository URL for the package.
+ * @param options.tagPrefix - The changelog tag prefix for the package.
+ * @returns The merged, re-serialized changelog content and the number of new
+ * entries that were merged in.
+ */
+export async function mergeChangelogs({
+  oursContent,
```
mcmire: Nit: Maybe `ourContent` / `theirContent`?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+  let mergedEntryCount = mergeReleaseChanges(
+    ours.getUnreleasedChanges(),
+    theirs.getUnreleasedChanges(),
+  );
+
+  const oursVersions = new Set(
+    ours.getReleases().map(({ version }) => version),
+  );
+
+  for (const theirsRelease of theirs.getReleases()) {
```
mcmire: Nit: Maybe `theirRelease`? etc.

scripts/lib/changelog-conflicts.test.ts
```diff
@@ -0,0 +1,609 @@
...
+### Added
+
+- Initial release ([#1](${REPO_URL}/pull/1))
+
+[Unreleased]: ${REPO_URL}/compare/${TAG_PREFIX}1.0.0...HEAD
+[1.0.0]: ${REPO_URL}/releases/tag/${TAG_PREFIX}1.0.0
+`;
+}
+
+describe('changelog-conflicts', () => {
```
mcmire: Nit: Jest already treats each test file as a module, so no need to wrap all tests for a file in a `describe`.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+  skipped: ConflictSkip[];
+};
+
+/**
+ * Find `packages/*\/CHANGELOG.md` files that currently have an unresolved
+ * Git merge conflict.
+ *
+ * @returns The repo-relative paths of the conflicted changelog files.
+ */
+export async function findConflictedChangelogFiles(): Promise<string[]> {
```
mcmire: Nit: "Conflicted" is a feeling. Maybe `findConflictingChangelogFiles` would sound better?

``` suggestion
export async function findConflictingChangelogFiles(): Promise<string[]> {
```

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+  if (!packageJson.name || !repositoryUrl) {
+    throw new Error(
+      `Could not resolve package name or repository for "${changelogPath}".`,
+    );
+  }
+
+  return {
+    name: packageJson.name,
+    repoUrl: repositoryUrl.replace(/\.git$/u, ''),
+    tagPrefix: `${packageJson.name}@`,
```
mcmire, line 169: Something to keep in mind: Not every package has a standard tag prefix. Some packages have been renamed over time and `auto-changelog` must be informed about the old name when updating the changelog so that constructed repo URLs are correct. Currently that knowledge — the old name vs. new name — is kept within the `changelog:validate` and `changelog:update`, as `--tag-prefix-before-rename` and `--version-before-rename` options to `auto-changelog`.

For instance, `@metamask/json-rpc-engine` used to be called `json-rpc-engine`: https://github.com/MetaMask/core/blob/20bd45ba9d296cbfb304727739992e7a026b5aa8/packages/json-rpc-engine/package.json#L59. And you can see that reflected in the URLs the bottom of the changelog: https://raw.githubusercontent.com/MetaMask/core/refs/heads/main/packages/json-rpc-engine/CHANGELOG.md. `eth-json-rpc-middleware` and `json-rpc-middleware-stream` are the same way. As it is, if there is a conflict in the changelogs for these packages then this script would likely rewrite those URLs.

I see two options for fixing this:

A. Copy the command line option values used for these packages into this file (passing them as arguments to `parseChangelog`). There would still be two places where this knowledge would be kept, but at least we'd be accounting for these packages.
B. For each of these packages, copy the command line option values into a new field within `package.json` called `autoChangelog` (or something similar). Then have this script read that field. (Later we can have `auto-changelog` look for this field directly.)

What are your thoughts?

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+  // branch being rebased onto. In both cases, that's the side more likely to
+  // already contain entries also present in `theirs`, so preserving its
+  // existing order (and only appending genuinely new entries from `theirs`)
+  // produces more intuitive results than the reverse.
+  const ours = parseChangelog({
+    changelogContent: oursContent,
+    repoUrl,
+    tagPrefix,
+    formatter: oxfmt,
+    shouldExtractPrLinks: true,
```
mcmire, line 324: Another thing to keep in mind: this option isn't enabled by default. The `check-changelog` action enables it whenever it parses changelogs, so as a result, most packages have had their changelogs accidentally reformatted so that PR links are formatted consistently. However, I don't know if this is the case for every package, so setting this option to `true` may force the rest of them to get reformatted as time goes on. That's probably okay, because I do eventually want to enable this option by default. Just letting you know.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+    theirs.getUnreleasedChanges(),
+  );
+
+  const oursVersions = new Set(
+    ours.getReleases().map(({ version }) => version),
+  );
+
+  for (const theirsRelease of theirs.getReleases()) {
+    if (!oursVersions.has(theirsRelease.version)) {
+      // `addRelease` can only add to the very start or end of the release
```
mcmire, line 345: Yeah, it'd be helpful to have an `insertRelease` method on Changelog, but this seems okay in the meantime.

scripts/lib/changelog-conflicts.ts
mcmire: I'm okay with keeping `ours`/`theirs` throughout the whole thing so there are fewer terms to keep straight (and `theirs` is understandable as the incoming change, to me), but up to you.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+ * Merge new entries from `incoming` into `base` across every category
+ * present on either side, mutating `base` in place.
+ *
+ * @param base - The release's changes (by category) to merge into.
+ * @param incoming - The release's changes (by category) to merge from.
+ * @returns The number of new entries added to `base`.
+ */
+function mergeReleaseChanges(
+  base: ReleaseChanges,
+  incoming: ReleaseChanges,
```
mcmire: Nit: It might be helpful to remind ourselves that these are release changes:

``` suggestion
  baseReleaseChanges: ReleaseChanges,
  incomingReleaseChanges: ReleaseChanges,

File: scripts/lib/changelog-conflicts.ts [hunk 22]
```diff
@@ -0,0 +1,380 @@
...
+    if (incomingEntries.length === 0) {
+      continue;
+    }
+
+    base[category] ??= [];
+
+    addedCount += mergeCategoryEntries(
+      base[category] as Change[],
+      incomingEntries,
+    );
```
mcmire: Would this mutate the `base` argument? Maybe it would be better to say:

``` suggestion
    addedCount += mergeCategoryEntries(
      base[category] ?? []) as Change[],
      incomingEntries,
    );

File: scripts/lib/changelog-conflicts.ts [hunk 23]
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * @param base - The release's changes (by category) to merge into.
+ * @param incoming - The release's changes (by category) to merge from.
+ * @returns The number of new entries added to `base`.
+ */
+function mergeReleaseChanges(
+  base: ReleaseChanges,
+  incoming: ReleaseChanges,
+): number {
+  let addedCount = 0;
```
mcmire: Nit: It might be helpful to remind ourselves what we are adding:

``` suggestion
  let addedEntriesCount = 0;
```

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+ * Merge new entries from `incoming` into `base`, mutating `base` in place.
+ * Breaking changes are inserted below any existing leading breaking changes;
+ * other changes are appended to the end. Relative order within `incoming` is
+ * preserved.
+ *
+ * @param base - The category's changes to merge into.
+ * @param incoming - The category's changes to merge from.
+ * @returns The number of new entries added to `base`.
+ */
+function mergeCategoryEntries(base: Change[], incoming: Change[]): number {
```
mcmire: Nit: It might be helpful to remind ourselves that these are changes (and to emphasize that they are not release changes):

``` suggestion
function mergeCategoryEntries(
  baseChanges: Change[],
  incomingChanges: Change[]
): number {

File: scripts/lib/changelog-conflicts.ts [hunk 25]
```diff
@@ -0,0 +1,380 @@
...
+  for (const change of incoming) {
+    const key = getChangeKey(change);
+    if (existingKeys.has(key)) {
+      continue;
+    }
+
+    existingKeys.add(key);
+
+    if (isBreakingChange(change)) {
+      const firstNonBreakingIndex = base.findIndex(
```
mcmire: Also to keep in mind: we currently do not enforce that breaking changes are listed first. So I'm curious if this would cause unintentional reordering of entries? I guess that wouldn't be the end of the world, so maybe it's okay if that happens. Just something to think about though.

scripts/lib/changelog-conflicts.ts
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * Determine whether a change entry is a breaking change, per the
+ * `**BREAKING:**` description prefix convention.
+ *
+ * @param change - The change entry.
+ * @returns Whether the change is a breaking change.
+ */
+function isBreakingChange(change: Change): boolean {
+  return BREAKING_CHANGE_PATTERN.test(change.description.trim());
```
mcmire, line 182: This is fine and makes sense, but I wonder if eventually we should have `auto-changelog` look for this prefix and include its presence as an `isBreaking` property on Change, so we don't keep this knowledge in two places.

mcmire (commented):
Had some comments but overall seems like a good approach.

scripts/lib/changelog-conflicts.ts
mcmire: Oh I see is that related to this? https://github.com/MetaMask/core/pull/9994#discussion_r3880620735

Mrtenz: Yes, and this comment too: https://github.com/MetaMask/core/pull/9994#discussion_r3880382048.

`path` by default uses OS-dependent formatting, so on Linux/macOS (POSIX) `path.join('some', 'path')` would be `some/path`, and on Windows `some\\path`. Git uses POSIX-style paths regardless of OS.

scripts/lib/changelog-conflicts.ts
Mrtenz: This won't reorder existing breaking entries. It just finds the first non-breaking entry and places the breaking entry on top of that, so if a changelog looks like this:

```
- First entry.
- **BREAKING:** Second entry.
- Third entry.
```

And a new breaking entry needs to be added, the resulting changelog will look like this:

```
- **BREAKING:** New breaking entry.
- First entry.
- **BREAKING:** Second entry.
- Third entry.
```

scripts/lib/changelog-conflicts.ts
Mrtenz: Makes sense to add something to `auto-changelog`. Eventually we can move this logic/script to `auto-changelog` too.

scripts/lib/changelog-conflicts.ts
Mrtenz: `mergeCategoryEntries` mutates the `baseChanges` array in-place, so changing it to an inline `baseChanges[category] ?? []` means we're mutating an array that's not stored anywhere, hence the nullish coalescing assignment before.

scripts/lib/changelog-conflicts.ts
Mrtenz: Good catch. Since I intend to move the merge logic into `auto-changelog` at some point, I went with option A for the time being. Deduplication of knowledge is of course better, but also a lot more effort. Once the logic lives in `auto-changelog`, we can just have a `changelog:merge` script in each `package.json` containing the flags.

scripts/lib/changelog-conflicts.ts
mcmire: Ah I see. Okay that makes sense.

scripts/lib/changelog-conflicts.ts
mcmire: Sounds good!

scripts/lib/changelog-conflicts.ts
mcmire: Ah gotcha. Makes sense.

mcmire (approved):
LGTM!

PR #10009, opened 2026-08-28
Enable typechecking in CI for base-data-service

mcmire:
## Explanation

To ensure all files in `base-data-service` — not just production files — are typechecked in CI now and in the future:

- Add `tsconfig.lint.json` for `base-data-service` and hook it into root `tsconfig.lint.json`
- Fix type errors in tests
  - Add missing `StorageService:*` actions to `ExampleDataService` messenger
  - Refactor test setup by adding `createRootMessenger` and `createServiceMessenger` helpers

## References

(N/A)

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes are limited to TypeScript lint config and test typing/setup; no production `base-data-service` logic is modified.
> 
> **Overview**
> Adds **`packages/base-data-service/tsconfig.lint.json`** and registers it in the root lint project so CI typechecks the package’s **tests** as well as production code.
> 
> To satisfy those checks, the test fixture **`ExampleDataService`** messenger type now includes **`StorageService:getItem`**, **`setItem`**, and **`removeItem`** actions. **`BaseDataService.test.ts`** centralizes setup in **`createRootMessenger`** and **`createServiceMessenger`**, wiring storage through **`rootMessenger.delegate`** instead of ad‑hoc **`buildChild`** / **`registerActionHandler`** blocks in each case.
> 
> Persistence specs now assert against injected storage handler mocks (and use **`messenger.waitUntil`** for rehydration) rather than spying on **`messenger.call`**. Runtime library behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42b2c73588e7c4d074d946dbb15750c14206418a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/base-data-service/src/BaseDataService.test.ts
+++ b/packages/base-data-service/src/BaseDataService.test.ts
@@ -1,10 +1,21 @@
-import { MOCK_ANY_NAMESPACE, Messenger } from '@metamask/messenger';
+import {
+  MOCK_ANY_NAMESPACE,
+  Messenger,
+  MessengerActions,
+  MockAnyNamespace,
+} from '@metamask/messenger';
+import {
+  StorageServiceGetItemAction,
+  StorageServiceRemoveItemAction,
+  StorageServiceSetItemAction,
+} from '@metamask/storage-service';
 import { hashKey } from '@tanstack/query-core';
 import { BrokenCircuitError } from 'cockatiel';
 import { cleanAll } from 'nock';
 
 import {
   ExampleDataService,
+  ExampleMessenger,
   serviceName,
 } from '../tests/ExampleDataService.js';
 import {
@@ -25,6 +36,14 @@ const MOCK_ASSETS = [
   'eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f',
 ];
 
+type RootMessenger = Messenger<
+  MockAnyNamespace,
+  | StorageServiceGetItemAction
+  | StorageServiceSetItemAction
+  | StorageServiceRemoveItemAction,
+  never
+>;
+
 describe('BaseDataService', () => {
   beforeAll(() => {
     jest.useFakeTimers({ doNotFake: ['nextTick', 'setImmediate'] });
@@ -42,7 +61,7 @@ describe('BaseDataService', () => {
   });
 
   it('handles basic queries', async () => {
-    const messenger = new Messenger({ namespace: serviceName });
+    const messenger = createServiceMessenger();
     const service = new ExampleDataService(messenger);
 
     expect(await service.getAssets(MOCK_ASSETS)).toStrictEqual([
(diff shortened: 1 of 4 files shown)


packages/base-data-service/tests/ExampleDataService.ts
```diff
@@ -22,7 +27,10 @@ export const serviceName = 'ExampleDataService';
 
 export type ExampleDataServiceActions =
   | ExampleDataServiceMethodActions
-  | DataServiceInvalidateQueriesAction<typeof serviceName>;
+  | DataServiceInvalidateQueriesAction<typeof serviceName>
+  | StorageServiceGetItemAction
```
mcmire, line 31: I found that this was necessary in order to remove the type errors above. It seems that `BaseDataService` does not verify that these actions are present on the messenger that it expects. Is that a bug?

mcmire:
Adding `no-changelog` since there are no changes to source files.

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -600,59 +573,97 @@ describe('BaseDataService', () => {
...
+ * Create an ExampleDataService messenger.
+ *
+ * @param rootMessenger - The root messenger to derive the ExampleDataService
+ * messenger from.
+ * @returns The ExampleDataService messenger.
+ */
+function createServiceMessenger(
+  rootMessenger = createRootMessenger(),
+): ExampleMessenger {
+  const messenger: ExampleMessenger = new Messenger({
```
mcmire, line 657: I opted not to use `buildChild` in this refactor because it doesn't seem to assign the right types by default: it will give the resulting messenger a type that includes the actions that we are delegating, but we actually need the resulting type to include `ExampleDataService` capability types too. 

Besides, adding a type annotation like this to "force" the type is consistent with how we generally create root messengers (see `createRootMessenger` above). Technically we _still_ could use `buildChild` in this case and get away with it, but that would work accidentally (because in specifying a return type we are actually assigning a type annotation to this function, so we're still "forcing" TypeScript to do what we want). But again, I like the consistency as it's easier for me to keep straight.

cryptodev-2s (dismissed):

packages/base-data-service/src/BaseDataService.test.ts
```diff
@@ -600,59 +573,97 @@ describe('BaseDataService', () => {
...
       const service = new ExampleDataService(messenger, {});
 
       service.init();
 
-      expect(callSpy).not.toHaveBeenCalledWith(
-        'StorageService:getItem',
-        expect.anything(),
-        expect.anything(),
-      );
+      expect(setItem).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
mcmire: Good point. Fixed in [42b2c73](https://github.com/MetaMask/core/pull/10009/commits/42b2c73588e7c4d074d946dbb15750c14206418a).

packages/base-data-service/tests/ExampleDataService.ts
FrederikBolding: Simply an oversight that we forgot to add this

FrederikBolding (approved):

PR #10010, opened 2026-08-28
Add typechecking in CI for react-data-query

mcmire:
## Explanation

To ensure all files in `react-data-query` — not just production files — are typechecked in CI now and in the future:

- Add `tsconfig.lint.json` for `react-data-query` and hook it into root `tsconfig.lint.json`
- Fix type errors in tests
  - Refactor `createUIQueryClient` test setup by porting `createRootMessenger` and `createServiceMessenger` helpers from `base-data-service` tests
  - Change `createUIQueryClient` test setup to create a messenger adapter instead of passing a messenger, since that isn't supported
- Remove `tstyche` test and dependency (it was only added as a stopgap until we had true typechecking in CI)

Also update `QueryKey` in `@metamask/base-data-service` to allow `readonly` so we aren't forced to use a typecast when defining `queryKey` in its own variable and passing it to `useQuery` or other functions.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes are mostly CI, tests, and an additive type widening; no production runtime behavior changes are introduced.
> 
> **Overview**
> Enables **full-package TypeScript checking** for `@metamask/react-data-query` in CI by adding `tsconfig.lint.json`, wiring it into the root lint project references, and dropping the **tstyche**-based `test:types` flow (scripts, devDependency, and `createUIQueryClient.tst.ts`). Jest is now the sole test entrypoint, and coverage config no longer ignores `.tst.ts` files.
> 
> Test fixes align with real APIs: **`createUIQueryClient` tests** use a messenger **adapter** (not a raw `Messenger`), with setup helpers ported from `base-data-service` tests; hook tests use **`as const`** on `queryKey`. A redundant “proxies requests to the messenger adapter” unit test was removed in favor of the shared setup.
> 
> In **`@metamask/base-data-service`**, **`QueryKey`** is widened to also accept **`readonly [string, ...Json[]]`**, so keys extracted with `as const` type-check for `fetchQuery` / `fetchInfiniteQuery` and related consumers (documented in the package changelog).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd2ec3031cffcc94681e47815308708cc6f29825. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/react-data-query/src/hooks.test.ts
```diff
@@ -12,7 +12,11 @@ jest.mock('@tanstack/react-query', () => ({
 
 describe('useQuery', () => {
   it('calls the underlying TanStack query function', () => {
-    const options = { queryKey: ['foo'] };
+    const options = {
+      // Type assertion: `useQuery` wants `[string, ...Json[]]`,
```
mcmire: Not great, but I think this is just how `useQuery` works: if you store options in a variable and pass the variable into `useQuery` you have to add some kind of type assertion to satisfy the strange types that `useQuery` has.

mcmire:
Adding `no-changelog` since there are no changes to source files.

packages/react-data-query/src/hooks.test.ts
```diff
@@ -12,7 +12,11 @@ jest.mock('@tanstack/react-query', () => ({
...
+      // which is not the inferred type of this key.
+      queryKey: ['foo'] as ['foo'],
```
Mrtenz: Does `as const` not work here?

Mrtenz (dismissed):

packages/react-data-query/src/hooks.test.ts
mcmire: Yes! We have to change `QueryKey` to allow `readonly` to accommodate this, but I think the result is much better. Updated here: [bd2ec30](https://github.com/MetaMask/core/pull/10010/commits/bd2ec3031cffcc94681e47815308708cc6f29825)

packages/react-data-query/src/hooks.test.ts
mcmire: I realized that we are the ones that defined this `QueryKey` type, it doesn't come from TanStack Query. So I've changed it so that we don't need a type assertion anymore: [bd2ec30](https://github.com/MetaMask/core/pull/10010/commits/bd2ec3031cffcc94681e47815308708cc6f29825)

Mrtenz (approved):

PR #10037, opened 2026-09-01
fix(perps-controller): order 0: Reduce only order would increase position

abretonc7s:
## Explanation

Fixes the remaining TAT-3252 failure paths. A stale position or half-up size rounding could submit a reduce-only order larger than the current position. #9719 fixed single-position close only.

Core now uses one position-source rule:

- symbol action: current DEX slice, otherwise HTTP;
- batch action: complete current WebSocket snapshot, otherwise complete REST snapshot;
- never merge REST and WebSocket positions.

This covers TP/SL, single and batch close, reduce-only edits, margin updates, and HIP-3 margin sizing. No public API changes.

## Proof

- **Reproduced first:** on pre-fix source, regressions submit an oversized TP/SL size and read stale position data, reproducing the Jira failure path.
- **Proved after:** the same regressions pass. A live HyperLiquid testnet recipe passed 47/47 nodes on `d4e527eb85`: `updateMargin` succeeded, a 100% partial TP rested at no more than the position size, the position stayed open, and cleanup closed it.
- The full Perps suite, TypeScript, ESLint, formatting, and changelog checks pass.

## References

- Fixes [TAT-3252](https://consensyssoftware.atlassian.net/browse/TAT-3252)
- Also fixes the close half of [TAT-3873](https://consensyssoftware.atlassian.net/browse/TAT-3873)
- Follows #9719, which covered only single-position close

## Checklist

- [x] Tests, JSDoc, and changelog updated
- [x] Breaking changes introduced

[TAT-3252]: https://consensyssoftware.atlassian.net/browse/TAT-3252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes core trading paths (close, batch close, TP/SL, margin) and WebSocket position cache semantics; mistakes could submit wrong reduce-only sizes or skip live positions.
> 
> **Overview**
> Hardens HyperLiquid **reduce-only** flows (close, batch close, TP/SL, margin updates, HIP-3 margin sizing) so order size and side always come from **one fresh source**, not caller snapshots or a frozen aggregate cache.
> 
> **Position freshness:** WebSocket per-DEX slices are stamped with a **connection epoch** that advances on socket close/reconnect; stale `clearinghouseState` from a replaced client is ignored. Symbol-scoped actions use the current DEX slice or a single HTTP read; `closeAll`/empty symbols require a **complete** all-DEX WebSocket snapshot or a strict REST fan-out (`PROVIDER_NOT_AVAILABLE` if any DEX is missing). REST and WebSocket positions are never mixed. `closePositions` loads only DEXes for the requested symbols and can return per-symbol unavailable results while still closing others.
> 
> **Sizing:** Partial TP/SL, whole-position TP/SL formatting, and **reduce-only** `editOrder` sizes are **floored** to the venue grid (with early `ORDER_TPSL_SIZE_INVALID` / `ORDER_SIZE_POSITIVE`); non-reduce edits keep half-up rounding.
> 
> **API/docs:** Optional batch-level `error` on `ClosePositionsResult`; optional `position` on close/TP/SL is documented as diagnostic only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ab8e682f5d750b3250da7e068864d292f15f898. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -8928,8 +8991,9 @@ export class HyperLiquidProvider implements PerpsProvider {
       // The provider-owned market policy is resolved from the positions below.
       await this.#ensureReadyForTrading({ requiresBuilderFee: false });
 
-      // Get all current positions from cache (avoids 429 rate limiting)
-      const positions = await this.getPositions();
+      // Use a complete current WebSocket snapshot or a strict all-DEX REST
+      // snapshot. Never mix sources whose relative recency is unknown.
+      const positions = await this.#getPositionsForOperation();
```
geositta: **Blocking — Closing one position depends on loading every market group**

This loads positions from every enabled market group before filtering to the symbols the user selected. If loading `xyz` stock positions fails, the app also refuses to close BTC—even though BTC and the main Hyperliquid market are working.

Please load position data only for the selected symbols. A complete snapshot across all market groups is necessary only for `closeAll` or an empty `symbols` list. Please add a regression test covering a selected BTC close when loading `xyz` positions fails.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2151,54 +2197,68 @@ export class HyperLiquidSubscriptionService {
...
+
+          // Stamp at delivery, not subscription creation. The SDK keeps this
+          // listener across automatic reconnects, so a captured creation
+          // epoch would reject every legitimate payload after the first
+          // reconnect. Delivery is synchronous from the raw socket `message`
+          // event through the SDK listener; the raw `close` event that retires
+          // the epoch cannot interleave with this callback. Only this
+          // clearinghouseState path stamps because openOrders carries no
+          // authoritative size or side.
+          this.#dexPositionsEpoch.set(
```
geositta, line 2265: **Blocking — An old position update can be marked as current after reconnect**

When reconnect replaces the WebSocket transport, a position update already queued from the old connection can run after the connection epoch has advanced. This line then stamps the old size and side with the new epoch, so close and margin operations can treat stale position data as current.

Please capture the subscription client used to create this callback and ignore the update when that client is no longer current. An automatic reconnect can continue using the same client, while a manually replaced transport will have a different client. Please add a regression test that invokes an old callback after transport replacement and verifies that its positions are not returned as fresh.

packages/perps-controller/src/types/index.ts
```diff
@@ -692,6 +679,7 @@ export type ClosePositionsResult = {
   success: boolean; // Overall success (true if at least one position closed)
   successCount: number; // Number of positions closed successfully
   failureCount: number; // Number of positions that failed to close
+  error?: string; // Batch-level failure before per-position results are available
```
geositta: **Non-blocking — The documented result contract does not match runtime behavior**

This says `error` is used when per-position results are unavailable, but `HyperLiquidProvider.closePositions` also returns a top-level `error` with populated per-position failures when batch submission throws. Please describe this as a batch-level operation error, or model the mutually exclusive result shapes with a discriminated union.

geositta (changes requested):
The reduce-only sizing and position freshness work is well tested, and the current CI suite passes. I am requesting changes for two exit-path regressions that affect production behavior:

1. Closing selected positions now depends on loading position data for every enabled market group.
2. A delayed position update from a replaced WebSocket connection can be accepted as current after reconnect.

Please scope selected position closes to the market groups represented by the requested symbols, and reject callbacks from replaced subscription clients. The batch level error documentation should also be aligned with the result shapes returned at runtime.

I found no additional TypeScript or numeric precision concerns, and this PR does not change React code.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
abretonc7s: Fixed in d30dfbeaa9. Selected closePositions now loads only the market groups for the requested symbols, so a failed xyz read no longer blocks a BTC close. Added a regression for that case; closeAll still requires a complete snapshot.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
abretonc7s: Fixed in d30dfbeaa9. The clearinghouseState callback captures the subscription client that created it and ignores updates when that client is no longer current. Added a regression that invokes the old callback after client replacement.

packages/perps-controller/src/types/index.ts
abretonc7s: Fixed in d30dfbeaa9. ClosePositionsResult.error is documented as a batch-level operation error that can appear with empty or populated per-position results. Left the single result shape so Mobile and Extension stay compatible.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -8997,16 +9060,35 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+              await Promise.all(
+                [
+                  ...new Set(
+                    requestedSymbols.map(
+                      (symbol) => parseAssetName(symbol).dex ?? '',
+                    ),
+                  ),
+                ].map((dexName) => this.#getPositionsForOperation(dexName)),
+              )
+            ).flat();
```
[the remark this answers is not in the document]
abretonc7s: Fixed in a65393e388. Selected closePositions now settles per-DEX loads independently, so a PROVIDER_NOT_AVAILABLE failure on xyz still closes BTC in the same symbols list. The xyz symbols are reported as per-position failures. Added a regression for that mixed selection.

Bigshmow (approved):

geositta (approved):

PR #10056, opened 2026-09-01
feat(perps-controller): add subscribeToTwapOrders

abretonc7s:
## Explanation

TWAP was read-only through `getTwapOrders()`, so a client wanting to monitor a running schedule had to poll — and each poll is two venue REST calls (`twapHistory` + `userTwapSliceFills`). HyperLiquid already pushes the same `twapHistory` payload over its `userTwapHistory` WebSocket channel; nothing in Core consumed it.

This adds `subscribeToTwapOrders` so clients can stream instead.

Found while building the Mobile TWAP management tab ([metamask-mobile#35544](https://github.com/MetaMask/metamask-mobile/pull/35544)), which currently polls at 5s purely because this contract was missing.

### Design notes

- **`SubscribeTwapOrdersParams` delivers the same `TwapOrder[]` shape `getTwapOrders()` returns**, so a client swaps a poll for a subscription without reshaping state.
- **The provider member is optional.** Providers without a native TWAP push channel omit it and the controller returns a no-op cleanup, so callers need not branch on provider identity and Lighter/MYX need no stubs. This also keeps the change non-breaking. It follows the precedent already set by `getTwapOrders?`.
- **The service owns transport and fan-out only; the provider supplies the adapter.** `#adaptTwapOrder` and its helpers are private to `HyperLiquidProvider`, so rather than move venue-shape knowledge into the subscription service (or duplicate it), the provider passes an `adapt` callback. `getTwapOrders()` is untouched.
- **Concurrent subscribers share one socket.** The first implementation opened a second WebSocket when two subscribers registered before either finished awaiting; a unit test caught it, and it now uses the same promise-dedup pattern as `#ensureOpenOrdersSubscription`.
- **Reconnect is handled.** Adapters are retained per account so the subscription is re-established after a reconnect, alongside the existing order-fill rehydration. `clearAll()` tears the channel down.

### Known limitation

The venue streams schedule state **without slice fills**, so pushed schedules carry an empty `fills` array. The Mobile consumer carries forward the fills a prior `getTwapOrders()` read supplied rather than blanking its Fill History view. Joining `userTwapSliceFills` into the same channel is possible (the SDK exposes it) but needs two accumulated caches and a join on every event — deliberately left out of this PR.

Separately: **HyperLiquid's `TwapState` carries no trigger price or max price** (`{ coin, executedNtl, executedSz, minutes, randomize, reduceOnly, side, sz, timestamp, user }`). Those fields cannot be surfaced by any client or Core change.

## References

Ref: https://consensyssoftware.atlassian.net/browse/TAT-3903
Mobile consumer: https://github.com/MetaMask/metamask-mobile/pull/35544

## Checklist

---

> [!NOTE]
> **Medium Risk**
> New live trading data path with async HIP-3 rebalance on terminal TWAPs and complex reconnect/generation handling; mistakes could strand collateral or deliver stale merged state, though behavior mirrors existing poll paths.
> 
> **Overview**
> Adds **`subscribeToTwapOrders`** so clients can stream TWAP schedule updates instead of polling `getTwapOrders()`. The hook is **optional** on `PerpsProvider`; when missing, the controller returns a **no-op unsubscribe** so callers can fall back to polling without provider checks.
> 
> **HyperLiquid** wires the venue `userTwapHistory` WebSocket through `HyperLiquidSubscriptionService` (per-account multiplexing, snapshot/delta merge, cached replay for late subscribers, reconnect restore, and a cap of **100 terminal** schedules while keeping all active ones). Pushed schedules omit slice fills; the service **preserves fills** from prior frames/REST when the stream sends empty `fills`. The provider adapts history to `TwapOrder[]` and runs the same **detached HIP-3 collateral rebalance** on terminal schedules as polled reads, with address/network captured before async work to avoid account-switch races.
> 
> **`AggregatedPerpsProvider`** delegates only to the **default** provider (no multi-provider mux) and stamps `providerId` on callbacks like `getTwapOrders`. Public exports include `SubscribeTwapOrdersParams` and the controller action type; tests cover controller no-ops and subscription lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 250e23bc4bb2cb7deb320862c6ff356cf61bbf59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2896,6 +2878,236 @@ export class HyperLiquidSubscriptionService {
...
+    },
+  ): () => void {
+    const { callback, accountId, adapt } = params;
+    const normalizedAccountId = accountId ?? 'default';
+    const unsubscribe = this.#createSubscription(
+      this.#twapOrderSubscribers,
+      callback,
+      normalizedAccountId,
+    );
+    this.#twapOrderAdapters.set(normalizedAccountId, adapt);
```
geositta, line 3029: Could we replay `#cachedTwapOrders.get(normalizedAccountId)` to each newly registered callback? Hyperliquid sends the snapshot once per underlying subscription. When a second consumer subscribes after that frame, `#ensureTwapOrderSubscription` returns because the socket already exists, and the new consumer receives no schedules until a future delta. A newly mounted or remounted Pro TWAP view therefore remains empty while another consumer keeps the shared socket alive. Please cover this with a test where the second subscriber registers after the initial snapshot and immediately receives the retained full set.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2896,6 +2878,236 @@ export class HyperLiquidSubscriptionService {
...
+        throw new Error('SubscriptionClient not available');
+      }
+    }
+
+    const userAddress =
+      await this.#walletService.getUserAddressWithDefault(accountId);
+
+    const subscription = await subscriptionClient.userTwapHistory(
+      { user: userAddress },
+      (data: UserTwapHistoryWsEvent) => {
```
geositta, line 3179: Could we check `startGeneration` inside this listener before processing the event? A frame queued before unsubscribe or reconnect can currently mutate the replacement cache and publish stale orders to newly registered subscribers. The post-subscription generation check protects registration, but it does not isolate later listener invocations from an invalidated socket.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2896,6 +2878,236 @@ export class HyperLiquidSubscriptionService {
...
+        // a long-running active schedule has an old `startedAt`, so a cap over
+        // the whole set would evict the very TWAP the user is monitoring in
+        // favour of newer dead ones.
+        const allOrders = [...merged.values()].sort(
+          (left, right) => right.startedAt - left.startedAt,
+        );
+        const activeOrders = allOrders.filter(
+          (order) => order.status === 'active',
+        );
+        const terminalOrders = allOrders
```
geositta, line 3222: Could we align this retention behavior with the public contract? `SubscribeTwapOrdersParams` states that callbacks receive the same full set as `getTwapOrders()`, while this slice drops every terminal schedule after 100. A trader replacing polling with this stream therefore loses older history retained by the REST read. If bounded retention is required, please make the limit part of the documented contract and align `getTwapOrders()`, or expose snapshot/delta events and let consumers choose their retention policy.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -4589,6 +4740,47 @@ export class HyperLiquidSubscriptionService {
...
+          return adapt ? [{ normalizedAccountId, adapt }] : [];
+        },
+      );
+      await Promise.all(
+        restorable.map(async ({ normalizedAccountId, adapt }) => {
+          const accountId =
+            normalizedAccountId === 'default'
+              ? undefined
+              : (normalizedAccountId as CaipAccountId);
+          await this.#ensureTwapOrderSubscription(accountId, adapt).catch(
```
geositta, line 4987: Could we schedule a per-account retry when restoration fails, following the market-data restoration path below? This code clears the dead subscription and swallows the replacement failure. If the transport remains connected afterward, no later reconnect occurs and the registered TWAP subscribers remain without a live socket.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -13557,6 +13391,93 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+  subscribeToTwapOrders(params: SubscribeTwapOrdersParams): () => void {
+    return this.#subscriptionService.subscribeToTwapOrders({
+      ...params,
+      adapt: (history) => {
+        const now = Date.now();
+        const ordersById = new Map<string, TwapOrder>();
+        for (const historyEntry of history) {
+          const order = this.#adaptTwapOrder({
+            historyEntry,
+            sliceFills: [],
```
geositta, line 13581: Could we preserve known fills or model this as a discriminated partial update? Emitting `fills: []` makes “not included by this channel” indistinguishable from “this order has no fills” and contradicts the documented drop-in replacement for `getTwapOrders()`. A consumer replacing state from this callback loses its Pro fill history. Hyperliquid exposes `userTwapSliceFills`; alternatively, the service can retain prior fills while merging lifecycle deltas.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -13557,6 +13391,93 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+    // inside the microtask it would resolve at run time, so an account switch
+    // racing a terminal push would key the reclaim to the newly selected
+    // account and silently skip the schedule it was meant to settle.
+    const userAddressPromise =
+      this.#walletService.getUserAddressWithDefault(accountId);
+
+    (async (): Promise<void> => {
+      const userAddress = await userAddressPromise;
+      await this.#rebalanceTerminalHip3Twaps(orders, {
+        network: this.#clientService.isTestnetMode() ? 'testnet' : 'mainnet',
```
geositta: Could we capture `network` synchronously alongside `userAddressPromise` before starting the detached task? The account scope is intentionally captured before the async boundary, but the network scope is read afterward. A network toggle while address resolution is pending makes cleanup search the opposite network's tracking key, skip the terminal TWAP, and leave transferred collateral unrebalanced.

packages/perps-controller/src/providers/AggregatedPerpsProvider.ts
```diff
@@ -906,6 +907,43 @@ export class AggregatedPerpsProvider implements PerpsProvider {
...
+   * Unlike `getTwapOrders`, this does not fan out: HyperLiquid is the sole
+   * venue with a native TWAP push channel today, so a mux would add
+   * multi-provider merge semantics with nothing to merge. Callers needing the
+   * aggregated view across providers keep using `getTwapOrders`.
+   *
+   * @param params - Subscription parameters including callback and account ID.
+   * @returns A cleanup function; a no-op when the default provider has no
+   * native TWAP push channel.
+   */
+  subscribeToTwapOrders(params: SubscribeTwapOrdersParams): () => void {
```
geositta, line 922: Could we keep this subscription's provider scope consistent with `getTwapOrders()`, or expose the narrower scope explicitly in the API? `getTwapOrders()` returns every active provider, but this method subscribes only to the default provider. An `AggregatedPerpsProvider` configured with MYX or Lighter as default returns a no-op even when its Hyperliquid provider has live TWAP data. Please either subscribe to every capable provider and merge by `providerId`, or return an explicit supported/unsupported result with a provider route.

packages/perps-controller/src/PerpsController.ts
```diff
@@ -5718,6 +5720,38 @@ export class PerpsController extends BaseController<
...
+   * Providers without a native TWAP push channel do not implement this; the
+   * returned no-op cleanup lets a client fall back to polling `getTwapOrders`
+   * without branching on provider identity.
+   *
+   * @param params - Subscription parameters including callback and account ID.
+   * @returns A cleanup function to remove the subscription.
+   */
+  subscribeToTwapOrders(params: SubscribeTwapOrdersParams): () => void {
+    const provider = this.getActiveProviderOrNull();
+    if (!provider?.subscribeToTwapOrders) {
```
geositta, line 5735: Could we return an explicit establishment result, such as `{ status: 'subscribed', unsubscribe } | { status: 'unsupported' | 'not_ready' }`, and expose asynchronous establishment failures? Returning the same no-op cleanup for an unavailable provider, an unsupported provider, and a setup failure is indistinguishable from a working subscription, so consumers cannot reliably continue polling as the contract requires.

packages/perps-controller/tests/src/services/HyperLiquidSubscriptionService.streams.test.ts
```diff
@@ -1564,4 +1591,372 @@ describe('HyperLiquidSubscriptionService', () => {
...
+        side: 'B',
+        sz: '10',
+        timestamp: 1_700_000_000_000,
+        user: '0x123',
+      },
+      status: { status: 'activated' },
+      twapId,
+    });
+
+    const statusAwareAdapt = (history: any) =>
```
geositta: Please replace the new `any` adapter and event fixtures with `TwapHistoryResponse` and `UserTwapHistoryWsEvent`, using `satisfies` for fixture validation. These tests exercise the Hyperliquid boundary, but `history: any` and `entry: any` allow invalid venue fields and status shapes to compile, removing the type-level protection that the production adapter depends on.

geositta (changes requested):
This is a strong foundation: the shared socket, account-scoped lifecycle handling, reconnect work, and active-order retention are thoughtfully structured, and the snapshot/delta merge follows Hyperliquid's documented stream behavior. I’m requesting changes because the current public contract can leave consumers without current state, erase fill history, publish stale frames after teardown, and silently disable updates when establishment or restoration fails. The aggregated and retention behavior also differs from the advertised parity with getTwapOrders(). Addressing these points will make this a reliable polling replacement for Pro traders and a safer API for React Native consumers.

Bigshmow (dismissed):

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
abretonc7s: Fixed in 4ff77bb8e6. `subscribeToTwapOrders` now replays the retained set to a subscriber that joins an already-open socket, guarded on a non-empty cache so a first subscriber doesn't get a spurious empty snapshot. Covered by a test where the second subscriber registers after the initial snapshot and immediately receives the full set.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
abretonc7s: Fixed in 4ff77bb8e6. The listener now compares `startGeneration` against the account's current generation and returns early, so a frame queued before an unsubscribe or reconnect can neither mutate the replacement's cache nor reach its subscribers.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
abretonc7s: Fixed in 4ff77bb8e6. Added `#scheduleTwapRestoreRetry`, mirroring `#scheduleRestoreRetry`: one deferred attempt per account, skipped while clearing or when one is already pending, and it no-ops if the adapter or subscribers are gone by the time it fires.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
abretonc7s: Fixed in 4ff77bb8e6. `network` is now captured synchronously alongside `userAddressPromise`, before the detached task starts, so a toggle mid-resolution can't key the reclaim to the opposite network.

packages/perps-controller/tests/src/services/HyperLiquidSubscriptionService.streams.test.ts
abretonc7s: Fixed in 4ff77bb8e6. The adapters are now typed `(history: TwapHistoryResponse): TwapOrder[]`, the fixture builder uses `satisfies TwapHistoryResponse[number]`, and the mock listeners take `UserTwapHistoryWsEvent`. This PR now adds zero `any` — verified by diffing `: any` occurrences against origin/main. You were right that this mattered: the untyped adapter is why a `startedAt` vs `lastUpdated` mix-up in the retention cap compiled cleanly and survived two review rounds.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
abretonc7s: Fixed in 4ff77bb8e6, in the service rather than the provider. The merge now carries forward the fills a prior frame or REST read resolved, so an empty `fills` from this channel no longer overwrites known history. Joining `userTwapSliceFills` into the same channel is the fuller answer — it needs a second accumulated cache and a join on every event, so I've left it out of this PR deliberately rather than by oversight.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
abretonc7s: Addressed in 250e23bc4b by making the limit part of the documented contract rather than changing the behaviour. `SubscribeTwapOrdersParams` now states that every active schedule is always delivered while terminal ones are bounded to the most recent 100, and points callers needing older history at `getTwapOrders()`. Unbounded retention on a long-lived stream was the alternative, and the cap exists because `userTwapHistory` is history — it would otherwise accumulate every schedule an account has ever run. Happy to switch to explicit snapshot/delta events if you'd prefer consumers own the policy.

packages/perps-controller/src/providers/AggregatedPerpsProvider.ts
abretonc7s: Not changed, and I want to flag the reasoning rather than quietly decline. I implemented the fan-out (subscribe to every capable provider, merge by `providerId`) and then reverted it: MYX is being removed, and HyperLiquid is the only venue with a native TWAP push channel, so the merge would have had nothing to merge. The default-provider scope is now stated explicitly in the method doc. If a second push-capable venue lands, the fan-out is the right shape and I'm happy to add it then.

packages/perps-controller/src/PerpsController.ts
abretonc7s: Not changed in this PR. I agree the three cases are indistinguishable, but a discriminated establishment result changes the shape of a public controller method, and every other `subscribeTo*` on this controller returns a bare `() => void`. Making TWAP the sole exception seemed worse than the ambiguity. If you'd like the richer result, I'd rather do it as a follow-up across all the subscription methods so the surface stays consistent — happy to open that.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2821,6 +2852,206 @@ export class HyperLiquidSubscriptionService {
...
+                'HyperLiquidSubscriptionService.subscribeToTwapOrders',
+              ),
+              this.#getErrorContext('subscribeToTwapOrders.unsubscribe'),
+            );
+          });
+          this.#twapOrderSubscriptions.delete(normalizedAccountId);
+        }
+        // Invalidate any open still in flight: without this its continuation
+        // installs a live socket that no subscriber is listening to.
+        this.#twapOrderSubscriptionGeneration += 1;
```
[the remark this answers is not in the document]
abretonc7s: Already fixed in 0867ef1fde, before this comment's snapshot. The counter is now `#twapOrderSubscriptionGenerations: Map<string, number>` keyed by account, so one account's teardown no longer invalidates another's in-flight open. Covered by a multi-account teardown-isolation test.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -13327,6 +13328,44 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+          const existing = ordersById.get(order.orderId);
+          if (!existing || order.lastUpdated >= existing.lastUpdated) {
+            ordersById.set(order.orderId, order);
+          }
+        }
+        return [...ordersById.values()].sort(
+          (left, right) => right.startedAt - left.startedAt,
+        );
+      },
+    });
```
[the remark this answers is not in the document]
abretonc7s: Already fixed in 842e0a00f9. `#rebalanceStreamedTerminalTwaps` runs from the stream path, so a terminal schedule seen on the subscription reclaims HIP-3 collateral the same way `getTwapOrders()` does.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2821,6 +2844,173 @@ export class HyperLiquidSubscriptionService {
...
+    );
+
+    // A concurrent caller may have established the subscription while this
+    // one awaited the client or the address; keep the first and drop ours so
+    // the venue is not left with an orphaned socket.
+    if (this.#twapOrderSubscriptions.has(normalizedAccountId)) {
+      subscription.unsubscribe().catch(() => undefined);
+      return;
+    }
+    this.#twapOrderSubscriptions.set(normalizedAccountId, subscription);
```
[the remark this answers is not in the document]
abretonc7s: Partly stale, partly fixed. The generation check does now exist and covers teardown, `clearAll`, and reconnect (0867ef1fde), the pending-map delete is identity-checked so a stale open can't evict a newer one, and registration additionally re-checks that subscribers still exist. Reconnect bumps the generation before clearing, so a pre-reconnect open discards itself rather than winning the first-writer check.

packages/perps-controller/src/services/HyperLiquidSubscriptionService.ts
```diff
@@ -2896,6 +2953,236 @@ export class HyperLiquidSubscriptionService {
...
+
+    this.#ensureTwapOrderSubscription(accountId, adapt).catch((error) => {
+      this.#logErrorUnlessClearing(
+        ensureError(
+          error,
+          'HyperLiquidSubscriptionService.subscribeToTwapOrders',
+        ),
+        this.#getErrorContext('subscribeToTwapOrders'),
+      );
+    });
```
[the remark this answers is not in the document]
abretonc7s: Fixed in 4ff77bb8e6 — this one was live, not stale. A subscriber joining an already-open socket now gets the retained set replayed immediately, so it no longer waits for a delta that may never arrive.

packages/perps-controller/src/providers/HyperLiquidProvider.ts
```diff
@@ -13557,6 +13558,93 @@ export class HyperLiquidProvider implements PerpsProvider {
...
+    // account and silently skip the schedule it was meant to settle.
+    const userAddressPromise =
+      this.#walletService.getUserAddressWithDefault(accountId);
+
+    (async (): Promise<void> => {
+      const userAddress = await userAddressPromise;
+      await this.#rebalanceTerminalHip3Twaps(orders, {
+        network: this.#clientService.isTestnetMode() ? 'testnet' : 'mainnet',
+        userAddress,
+      });
```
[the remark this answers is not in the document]
abretonc7s: Fixed across 4ff77bb8e6 and its predecessor. The address is resolved from the push's own scope before the task detaches, and `network` is now captured synchronously alongside it, so neither an account nor a network change mid-resolution can key the reclaim to the wrong scope.

abretonc7s:
Thanks @geositta — that was a genuinely useful review; several of these were real and one of them (the untyped fixtures) explains an earlier defect.

**Fixed** in 4ff77bb8e6 / 250e23bc4b:
- Cached replay for a subscriber joining an already-open socket
- Listener-level generation guard, so a frame queued before unsubscribe/reconnect can't mutate the replacement cache
- Fills carried across stream merges, so an empty `fills` no longer overwrites known history
- `network` captured synchronously with the address before the detached HIP-3 reclaim
- Per-account retry on failed TWAP restoration instead of swallowing it
- Test fixtures typed against `TwapHistoryResponse` / `UserTwapHistoryWsEvent`; this PR now adds zero `any`
- Retention bound is now part of the documented contract

**Deliberately not changed**, reasoning in the threads:
- Multi-provider fan-out — I built it, then reverted: MYX is being removed and HyperLiquid is the only push-capable venue, so there'd be nothing to merge
- Discriminated establishment result — worth doing, but across all `subscribeTo*` methods rather than making TWAP the lone exception

Several Bugbot comments were snapshots of earlier commits (the global generation counter and the missing HIP-3 rebalance were both already fixed); noted individually.

https://claude.ai/code/session_01WrdyHr1ELeWgsUGUjRrcK5

geositta (approved):

PR #10059, opened 2026-09-01
fix(assets-controller): clean accountTree event handlers

Kriys94:
## Explanation

Mobile
- Current behavior: Same as After because `AccountTreeController:initialized` was triggered at start
- After PR:
  - Onboarding: https://www.loom.com/share/5f5cdf90bfc2499eaf882bdc33c06365
  - After onboarding: https://www.loom.com/share/e48354b4869d42fb9d7070d377ce3a55

Extension:
- Current behavior: https://www.loom.com/share/55a51f14e91f4a6480145b959869dbd8 (behavior after https://github.com/MetaMask/core/pull/9892)
- After PR: https://www.loom.com/share/befb098c25584bfba2758464cc99e65e

### Context / motivation

Mobile and extension do **not** trigger `AccountTreeController:initialized` at the same point in the lifecycle:

- **Mobile** calls `AccountTreeController.init()` at startup via `AccountTreeInitService.initializeAccountTree()` (during authentication/login), so `:initialized` fires early in the app session.
- **Extension** does **not** call `init()` during controller construction (see FIXME in `account-tree-controller-init.ts` — the keyring must be unlocked first). Instead, it calls `AccountTreeController:init` later in `#initAccountsAfterUnlock()` after unlock.

Both clients delegate `AccountTreeController:initialized` to `AssetsController`, but the event timing differs. `AssetsController` could not reliably gate asset tracking on that event alone — mirroring `#accountTreeInitialized` from it was fragile across clients.

This PR adds `AccountTreeController:isInitialized` and has `AssetsController` read lifecycle state on demand so tracking starts/stops consistently regardless of when each client runs `init()`.

### Additional cleanup

While making that change, we also removed other sources of drift and complexity:

- **Local lifecycle mirrors:** `AssetsController` previously cached `#uiOpen`, `#keyringUnlocked`, and `#accountTreeInitialized` inside event handlers. Those flags could get out of sync if events were missed, reordered, or read before handlers ran. UI open and keyring unlock now use `ClientController:getState` and `KeyringController:isUnlocked` on demand as well.
- **`AccountTreeController:stateChange` subscription:** Used to detect mid-session tree mutations (e.g. a snap account added to the current group), with `#handleAccountTreeStateChange`, `#lastKnownAccountIds`, and overlap/diff logic to avoid redundant re-subscriptions during `init()`. That path is removed; explicit group switches stay on `selectedAccountGroupChange`, and snap accounts added mid-session without a group change are picked up on the next tracking restart.

### Solution

**`@metamask/account-tree-controller`**
- Add `AccountTreeController:isInitialized` so consumers can query whether `init()` has completed (returns `false` again after `clearState()` until the next successful `init()`).

**`@metamask/assets-controller`**
- Read lifecycle state **on demand** via messenger actions:
  - `#isUiOpen()` → `ClientController:getState`
  - `#isKeyringUnlocked()` → `KeyringController:isUnlocked`
  - `#isAccountTreeInitialized()` → `AccountTreeController:isInitialized`
- Remove local lifecycle mirrors and `#handleAccountTreeStateChange` / `#lastKnownAccountIds`.
- Stop subscribing to `AccountTreeController:stateChange`.
- Keep `AccountTreeController:selectedAccountGroupChange` for explicit group switches.
- Continue using `AccountTreeController:initialized` / `:uninitialized` to trigger `#updateActive()` (but gate on authoritative state via the actions above).

**Trade-off:** Snap accounts added mid-session without a group change are no longer picked up live via `:stateChange`; they are picked up on the next tracking restart (unlock / UI open cycle).

### Non-obvious details

- `#updateActive()` still reacts to `ClientController:stateChange`, `KeyringController:lock` / `:unlock`, and account-tree init events — but each gate now re-reads authoritative controller state instead of trusting cached flags.
- `#handleAccountGroupChanged` still ignores events while `#activeSubscriptions.size === 0`, so init-time `selectedAccountGroupChange` replays do not race with `:initialized`-owned startup.
- Test mocks were updated to publish `ClientController:stateChanged` (not `:stateChange`) so `ClientController:getState` reflects UI-open during lifecycle tests.

## References

- Consumer adoption (messenger delegation):
  - Extension: `app/scripts/messenger-client-init/messengers/assets/assets-controller-messenger.ts`
  - Mobile: `app/core/Engine/messengers/assets-controller/assets-controller-messenger.ts`
- Related prior work: `AccountTreeController:initialized` / `:uninitialized` (core#9892)

## Changelog

**`@metamask/account-tree-controller`**
- **Added:** `AccountTreeController:isInitialized` action

**`@metamask/assets-controller`**
- **Changed (BREAKING):** `AssetsControllerMessenger` now requires `AccountTreeController:isInitialized`, `ClientController:getState`, and `KeyringController:isUnlocked`
- **Changed (BREAKING):** No longer subscribes to `AccountTreeController:stateChange` — remove it from delegated events if listed explicitly
- **Changed:** Revert to `selectedAccountGroupChange` for account-group switches; snap accounts added mid-session are picked up on restart

## Manual testing

1. Build extension/mobile against this core branch.
2. Unlock wallet with UI open on the portfolio/home screen.
3. Verify balances load for the selected account group.
4. Switch account groups (Account 1 → Account 2) and confirm balances refresh.
5. Lock wallet → confirm tracking stops; unlock → confirm balances reload.
6. Close UI / background extension → reopen → confirm tracking resumes.
7. (Optional) Add a snap account to the current group mid-session and confirm it appears after lock/unlock or UI close/open (not immediately on tree mutation).

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking messenger wiring for extension/mobile plus a deliberate behavior change for mid-session account-tree updates; lifecycle gating touches when balances subscribe but not auth or funds logic directly.
> 
> **Overview**
> Adds **`AccountTreeController:isInitialized`** so consumers can tell whether `init()` finished (and it flips false again after `clearState()` until the next `init()`).
> 
> **`AssetsController`** no longer caches UI open, keyring unlock, or account-tree readiness from events. **`#updateActive()`** now gates tracking by calling **`ClientController:getState`**, **`KeyringController:isUnlocked`**, and **`AccountTreeController:isInitialized`** when it runs. It still listens to **`AccountTreeController:initialized` / `:uninitialized`** and client/keyring events, but only to re-evaluate those gates.
> 
> **Breaking for hosts:** the assets messenger must delegate the three new actions and can drop **`AccountTreeController:stateChange`** from allowed events. Mid-session account-tree mutations (e.g. snap added to the current group without a group switch) no longer trigger live refresh via **`stateChange`**; **`selectedAccountGroupChange`** still refreshes on explicit group switches, with snap pickup deferred until the next tracking restart (lock/unlock or UI cycle).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc7cfd33d428e5b11b99649418f5e54e8aded5a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/account-tree-controller/src/AccountTreeController.test.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.test.ts
@@ -531,6 +531,18 @@ describe('AccountTreeController', () => {
       expect(controller.isInitialized()).toBe(false);
     });
 
+    it('returns true after reinit', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+      controller.reinit();
+
+      expect(controller.isInitialized()).toBe(true);
+    });
+
     it('is exposed via AccountTreeController:isInitialized', () => {
       const { controller, messenger } = setup({
         accounts: [MOCK_HD_ACCOUNT_1],
--- a/packages/account-tree-controller/src/AccountTreeController.test.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.test.ts
@@ -498,6 +498,53 @@ describe('AccountTreeController', () => {
     jest.resetAllMocks();
   });
 
+  describe('isInitialized', () => {
+    it('returns false before init', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      expect(controller.isInitialized()).toBe(false);
+    });
+
+    it('returns true after init', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+
+      expect(controller.isInitialized()).toBe(true);
+    });
+
+    it('returns false after clearState until init is called again', () => {
+      const { controller } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+      controller.clearState();
+
+      expect(controller.isInitialized()).toBe(false);
+    });
+
+    it('is exposed via AccountTreeController:isInitialized', () => {
+      const { controller, messenger } = setup({
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      expect(messenger.call('AccountTreeController:isInitialized')).toBe(false);
+
+      controller.init();
+
+      expect(messenger.call('AccountTreeController:isInitialized')).toBe(true);
+    });
+  });
+
   describe('init', () => {
     it('groups accounts by entropy source, then snapId, then wallet type', () => {
       const { controller, messenger } = setup({
--- a/packages/account-tree-controller/src/AccountTreeController.ts
+++ b/packages/account-tree-controller/src/AccountTreeController.ts
@@ -51,6 +51,7 @@ import type { AccountWalletObject, AccountWalletObjectOf } from './wallet.js';
 export const controllerName = 'AccountTreeController';
 
 const MESSENGER_EXPOSED_METHODS = [
+  'isInitialized',
   'getSelectedAccountGroup',
   'setSelectedAccountGroup',
   'setSelectedAccountGroupByAccountId',
@@ -455,6 +456,16 @@ export class AccountTreeController extends BaseController<
     log('Re-initializing...');
     this.#initialized = false;
     this.init();
+  }
+
+  /**
+   * Whether `init()` has completed and the account tree is ready to consume.
+   * Returns false after `clearState()` until the next successful `init()`.
+   *
+   * @returns True when the controller has been initialized.
+   */
+  isInitialized(): boolean {
+    return this.#initialized;
   }
 
   /**
--- a/packages/assets-controller/src/AssetsController.ts
+++ b/packages/assets-controller/src/AssetsController.ts
@@ -1,8 +1,8 @@
 import type {
   AccountTreeControllerGetAccountsFromSelectedAccountGroupAction,
   AccountTreeControllerInitializedEvent,
+  AccountTreeControllerIsInitializedAction,
   AccountTreeControllerSelectedAccountGroupChangeEvent,
-  AccountTreeControllerStateChangeEvent,
   AccountTreeControllerUninitializedEvent,
 } from '@metamask/account-tree-controller';
 import type { AccountsControllerGetSelectedAccountAction } from '@metamask/accounts-controller';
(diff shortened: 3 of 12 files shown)


Kriys94:
@metamaskbot publish-preview

Kriys94:
Preview builds in extension: https://github.com/MetaMask/metamask-extension/pull/45953

Kriys94:
Preview builds in mobile: https://github.com/MetaMask/metamask-mobile/pull/35560

packages/account-tree-controller/src/AccountTreeController.ts
```diff
@@ -457,6 +458,16 @@ export class AccountTreeController extends BaseController<
...
   }
 
+  /**
+   * Whether `init()` has completed and the account tree is ready to consume.
+   * Returns false after `clearState()` until the next successful `init()`.
+   *
+   * @returns True when the controller has been initialized.
+   */
+  isInitialized(): boolean {
+    return this.#initialized;
```
salimtb, line 468: hhmmm we do this through the messenger now , this is better

packages/assets-controller/src/AssetsController.ts
```diff
@@ -805,6 +792,35 @@ export class AssetsController extends BaseController<
...
+  }
+
+  /**
+   * Whether the client (UI) is open, read from ClientController state.
+   *
+   * @returns True when the UI is open.
+   */
+  #isUiOpen(): boolean {
+    return clientControllerSelectors.selectIsUiOpen(
+      this.messenger.call('ClientController:getState'),
```
salimtb, line 811: ooh we replace everything with new actions now, could be good to export specific action instead of using getState here

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1322,92 +1322,16 @@ export class AssetsController extends BaseController<
...
   }
 
-  /**
-   * Handle AccountTreeController state changes after the tree is initialized.
-   * Re-subscribe only when the set of selected accounts has actually changed
-   * (e.g. a snap account was added after initial startup). Intermediate
-   * `:stateChange` events during `init()` are ignored — startup is driven by
-   * `:initialized` instead, so we do not fetch on every tree mutation.
-   */
-  #handleAccountTreeStateChange(): void {
```
salimtb, line 1340: can you quick me help me to understand why you're removing this function ?

packages/account-tree-controller/src/AccountTreeController.test.ts
```diff
@@ -498,6 +498,53 @@ describe('AccountTreeController', () => {
...
+        accounts: [MOCK_HD_ACCOUNT_1],
+        keyrings: [MOCK_HD_KEYRING_1],
+      });
+
+      controller.init();
+      controller.clearState();
+
+      expect(controller.isInitialized()).toBe(false);
+    });
+
```
ccharly, line 533: Nit: But having another small test with `reinit` could be nice. While we know the combo for `reinit` is clearing + calling `init` again, that would ensure that we don't have any regression for this action if someday we change it (clients do use `:reinit` sometimes)

salimtb (dismissed):

packages/account-tree-controller/src/AccountTreeController.test.ts
Kriys94: Thanks for the review :) Done could you look at https://github.com/MetaMask/core/pull/10059/changes/fc7cfd33d428e5b11b99649418f5e54e8aded5a9#diff-5ee70a3b38df69025fff2749ef20c331bf816d57f0ae5770a3645f736be6322fR534

ccharly (approved):

salimtb (approved):

packages/assets-controller/src/AssetsController.ts
Kriys94: Sure, selectedAccountGroupChange should be enough. We were listening stageChange to cover the case when a new non-EVM account was added with Snaps. Now we dont cover this case because there is almost no users that install third party non-EVM Snaps in runtime. And we win simplification of the codebase

juanmigdr (approved):

packages/assets-controller/src/AssetsController.ts
Kriys94: Yeah that's better but minor because uiOpen is the only state of the Controller right now

PR #10061, opened 2026-09-01
fix: read balances from RPC when Accounts API returns none

salimtb:
### The bug

The Accounts API omits tokens it does not index, and `AssetsController` commits balance responses to state with a `merge` update — anything absent from the response keeps its previous amount. Two failure modes result:

1. **Omitted token**: the API answers for a chain but leaves out a token it stopped indexing. The merge keeps the old amount in state **indefinitely**, and no other data source re-reads it because the API already claimed that chain as handled.
2. **Empty result**: the API returns no balances for an account at all, so the account is missing from `response.assetsBalance` entirely — the merge never even runs for it, with the same stale outcome.

A returned `0` is also indistinguishable from "not indexed", so it can't be trusted as a real zero either.

### The fix

`RpcFallbackMiddleware` (which already retries chains listed in `response.errors` on the RPC data source) now handles a second case: after the upstream sources respond, it scans controller state for **tracked EVM assets** (`state.assetsBalance` or `state.customAssets`) whose balance in the current response is empty — omitted, or reported as `0`. Those assets are handed to `RpcDataSource` as `customAssets`, so the balance fetcher includes them in its multicall and the on-chain amount overwrites the stale one (including a genuine `0`).

Guardrails on what gets re-read:

- **Staking vault assets** are excluded — their balances belong to `StakedBalanceDataSource`, and a plain ERC-20 `balanceOf` of the share token would clobber them.
- **Non-EVM assets** are excluded (RPC can't read them).
- Only chains that are **both in the request and supported by the owning account** — `RpcDataSource` fetches per account and silently drops anything else.
- Asset IDs are matched **case-insensitively** against the response, since state keys ERC-20s by checksummed address while some sources return them lower-cased.

The fix is contained entirely in `RpcFallbackMiddleware` (plus tests and changelog): the middleware reads state directly via `ctx.getAssetsState()`, so `DetectionMiddleware`, `TokenDataSource`, and the pipeline ordering in `AssetsController` are untouched. See the [PR comment](https://github.com/MetaMask/core/pull/10061#issuecomment-5501725996) for how this was consolidated from an earlier 11-file approach.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes balance merge behavior in the fast pipeline; incorrect filtering could leave stale balances or drop valid RPC recovery, but scope is limited to RpcFallbackMiddleware and RpcDataSource error propagation with extensive tests.
> 
> **Overview**
> Fixes **stale token balances** when the Accounts API omits unindexed assets or returns an untrusted `0`, because merge updates previously kept the old amount forever.
> 
> **`RpcFallbackMiddleware`** now triggers RPC balance reads in two situations: chains already in `response.errors`, and **tracked EVM assets** in controller state (`assetsBalance` / `customAssets`) whose entry in the current response is empty (missing, zero, or case-mismatched). Those assets are sent as `customAssets` for on-chain multicall, with exclusions for staking vault tokens, non-EVM assets, and chains outside the request or account support.
> 
> When merging RPC results, balances for chains where **RPC itself failed** are stripped before merge so failure stubs (native `0`) cannot overwrite good upstream balances or falsely clear errors. RPC errors are only kept for chains that were already errored upstream.
> 
> **`RpcDataSource.assetsMiddleware`** now copies per-chain `fetch` errors onto `context.response.errors`, which the fallback middleware uses to detect failed RPC chains.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ddcebe376162e67b0e945641988b167e9ae2218. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -45,17 +59,34 @@ export class RpcFallbackMiddleware {
       const erroredChains = new Set<ChainId>(
         Object.keys(ctx.response.errors ?? {}) as ChainId[],
       );
-      if (erroredChains.size === 0) {
+      const staleAssets = collectStaleTrackedAssets(ctx);
```
salimtb, line 62: **The core of the fix.** Previously this middleware only retried chains listed in `response.errors`. That never covered the stale-balance bug: when the Accounts API *answers* for a chain but omits a token it stopped indexing (or returns an empty result for the account), there is no error entry , the chain looks handled, the `merge` state update keeps the old amount, and nothing ever re-reads it.

`collectStaleTrackedAssets` (below) closes that gap by comparing the response against what state already tracks, instead of trusting the response to be complete.

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -45,17 +59,34 @@ export class RpcFallbackMiddleware {
...
+      const chainsToFetch = [
+        ...new Set([
+          ...ctx.request.chainIds.filter((id) => erroredChains.has(id)),
+          // Already restricted to requested chains. Their chain may not be
+          // errored: the Accounts API can answer for a chain while omitting a
+          // token it does not index.
+          ...staleAssets.map((assetId) => assetId.split('/')[0] as ChainId),
+        ]),
+      ];
+
```
salimtb, line 73: The stale assets' chains are added to the RPC fetch set **independently of `erroredChains`** , this is deliberate. A chain can be answered successfully by the Accounts API (so it never appears in `response.errors`) while still missing a token the API does not index. The set union also dedupes when a chain is both errored and hosts a stale asset.

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -45,17 +59,34 @@ export class RpcFallbackMiddleware {
...
+        chains: chainsToFetch,
       });
 
       const filteredRequest = {
         ...ctx.request,
-        chainIds: ctx.request.chainIds.filter((id) => erroredChains.has(id)),
+        chainIds: chainsToFetch,
+        customAssets: [
+          ...new Set([...(ctx.request.customAssets ?? []), ...staleAssets]),
+        ],
```
salimtb, line 89: Delivery mechanism: stale assets ride on `request.customAssets`, which `RpcDataSource` already includes in its per-chain multicall alongside the native asset (it filters to ERC-20s on the matching chain itself). The on-chain amount then overwrites the stale one via the normal response merge , including a genuine `0`, which is exactly the value the Accounts API could not be trusted to report.

Existing `customAssets` on the request are preserved and deduped. Because this middleware sits after `CustomAssetGraduationMiddleware` in the fast pipeline, the assets injected here can never trigger graduation.

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -99,3 +130,91 @@ export class RpcFallbackMiddleware {
...
+ * should re-read.
+ *
+ * Limited to chains both requested and supported by the owning account:
+ * RpcDataSource fetches per account and skips chains outside its supported
+ * set, so anything else would be queued and then silently dropped.
+ *
+ * @param ctx - Pipeline context.
+ * @returns Asset IDs to hand to the RPC data source.
+ */
+function collectStaleTrackedAssets(ctx: Context): Caip19AssetId[] {
```
salimtb, line 203: This reads controller state directly via `ctx.getAssetsState()`, which is what keeps the fix contained to this one file. "Tracked" means the union of `state.assetsBalance` (assets we hold a balance for) and `state.customAssets` (user-imported, possibly balance-less) per account.

The chain restriction (`supportedChains ∩ request.chainIds`) matters: `RpcDataSource` fetches per account and silently skips chains outside the account's supported set, so anything broader would be queued and then dropped without effect.

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -99,3 +130,91 @@ export class RpcFallbackMiddleware {
...
+    const chainsForAccount = supportedChains.filter((chainId) =>
+      ctx.request.chainIds.includes(chainId),
+    );
+
+    for (const assetId of trackedAssetIds) {
+      if (
+        isEvmAssetOnChains(assetId, chainsForAccount) &&
+        // Staked vault balances belong to StakedBalanceDataSource; an RPC
+        // ERC-20 read of the share token would clobber them.
+        !isStakingContractAssetId(assetId) &&
```
salimtb, line 226: Two exclusions here:

- `isEvmAssetOnChains` also filters out **non-EVM assets** (Solana, Bitcoin, …) , RPC cannot read them, so an empty upstream balance for those is not actionable.
- **Staking vault assets** are skipped because their balances are owned by `StakedBalanceDataSource`, which reads shares via the staking contract. A plain ERC-20 `balanceOf` of the share token through the fallback would clobber that value.

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -99,3 +130,91 @@ export class RpcFallbackMiddleware {
...
+ * Whether a balance response carries no positive amount for an asset. A
+ * returned `0` cannot be distinguished from "not indexed", so both count as
+ * empty. Asset IDs are matched case-insensitively: state keys ERC-20 assets by
+ * checksummed address, while some data sources return them lower-cased.
+ *
+ * @param balances - Balance map for a single account from the response.
+ * @param assetId - Asset ID to check.
+ * @returns True when the response holds no positive amount for the asset.
+ */
+function isBalanceEmpty(
```
salimtb, line 247: Two subtleties encoded here:

1. **`0` counts as empty.** The Accounts API cannot distinguish "balance is zero" from "token not indexed", so a returned `0` is untrusted and triggers an RPC re-read. If the balance really is zero, RPC confirms it and state is corrected either way.
2. **Case-insensitive matching.** State keys ERC-20 assets by checksummed address while some data sources return them lower-cased; an exact-key lookup would false-positive a "missing" balance and cause needless RPC reads. Exact match is tried first so the linear scan only runs on a case mismatch.

salimtb (commented):
Inline walkthrough of the fix — each comment explains one piece of the change. TL;DR: the Accounts API omits tokens it does not index (and can report an untrusted `0`), while state is committed with a `merge` update, so anything absent from the response silently kept its previous amount forever. This middleware now detects those tracked-but-empty assets and re-reads them on chain.

juanmigdr (dismissed):

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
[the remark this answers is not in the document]
salimtb: Confirmed and fixed in 6d0d431bb0 — verified each step of this finding against the code before fixing:

- `RpcDataSource` does write a native `0` stub for chains it fails on, and `mergeDataResponses` spreads the RPC result last, so the stub overwrote a correct upstream native amount.
- With `replaceCoveredChainBalances: true` in the fast pipeline, a chain "covered" only by that stub replaced the prior chain slice, wiping the tracked tokens on it. Worst case (Accounts API returns an empty result **and** RPC is down) the whole portfolio slice would be zeroed/wiped instead of left untouched.

**The fix:** balances from chains RPC itself reported in `errors` are discarded before the merge, and RPC errors are kept only for chains that were already errored upstream. A failed stale-asset re-read now leaves the authoritative upstream response untouched — the stale asset keeps its previous amount and is retried on the next pass (the safe pre-PR behavior). This also fixes a pre-existing quirk where the stub counted as a "recovered balance" and falsely cleared an errored chain's error.

Covered by three new regression tests: failed stale-asset chain (upstream survives), already-errored chain (stub dropped, error kept), and the mixed case (recovered chain merges while the failed one is discarded).

packages/assets-controller/src/middlewares/RpcFallbackMiddleware.ts
```diff
@@ -68,19 +99,45 @@ export class RpcFallbackMiddleware {
...
+      // upstream amounts and, with replaceCoveredChainBalances, wipe the
+      // chain's token slice from state. Drop them before merging — this also
+      // keeps failed chains from counting as "recovered" below.
+      const rpcFailedChains = new Set<ChainId>(
+        Object.keys(rpcResult.response.errors ?? {}) as ChainId[],
+      );
+      const rpcAssetsBalance = filterOutChainBalances(
+        rpcResult.response.assetsBalance,
+        rpcFailedChains,
+      );
```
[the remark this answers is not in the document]
salimtb: Confirmed and fixed in 5e2581448c. You're right: `RpcDataSource.assetsMiddleware` copied only `assetsBalance` and `assetsInfo` onto `context.response` and used the fetch errors purely internally (to compute `successfullyHandledChains`), so `rpcResult.response.errors` was always empty in the fallback and the discard filter from 6d0d431 was dead code in production.

**The fix:** `assetsMiddleware` now propagates the per-chain fetch errors onto `context.response.errors` (RpcDataSource.ts, after the `assetsInfo` copy), covered by a new test that mocks `fetch` returning errors and asserts they land on the context response — written first and watched fail against the old code. Downstream impact was audited: `#updateState` never consumes `response.errors` for state writes and `middlewareErrors` telemetry only counts thrown exceptions, so the propagation is inert outside `RpcFallbackMiddleware`, which is exactly the consumer that needed it.

Fair point on the mock fidelity — the fallback tests assign a full `DataResponse` including `errors`, which is the contract the real source didn't honor until now. With this commit the real middleware matches the mocked contract.

juanmigdr (approved):

Kriys94:
Personal comment: Does it goes through the subscribe flow?

PR #10074, opened 2026-09-02
chore(wallet): Update LICENSE

FrederikBolding:
## Explanation

Update LICENSE of `wallet` to match clients.

## References

https://consensyssoftware.atlassian.net/browse/WPC-1255

---

> [!NOTE]
> **High Risk**
> Breaking license terms restrict commercial redistribution and use; every consumer must re-evaluate compliance, not just upgrade dependencies.
> 
> **Overview**
> **Breaking legal change:** `@metamask/wallet` no longer ships under `(MIT OR Apache-2.0)`. The package now uses the ConsenSys Software Inc. **Non-Commercial Use** terms in `LICENSE`, with `package.json` set to `SEE LICENSE IN LICENSE`.
> 
> The standalone `LICENSE.MIT` and `LICENSE.APACHE2` files are removed, and the unreleased changelog documents that downstream use is limited to the non-commercial definition in `LICENSE` (commercial use requires contacting MetaMask).
> 
> No application or API code changes—this aligns the wallet package with client licensing (e.g. WPC-1255).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e130e2162769e9788fbbedde26770926d61ff44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/wallet/LICENSE
```diff
@@ -1,6 +1,18 @@
-This project is licensed under either of
-
- * MIT license ([LICENSE.MIT](LICENSE.MIT))
- * Apache License, Version 2.0 ([LICENSE.APACHE2](LICENSE.APACHE2))
-
-at your option. 
+Copyright ConsenSys Software Inc. 2026. All rights reserved.
```
cryptodev-2s, line 1: Nit: similar to the previously merged cryptography package should this be instead
```suggestion
Copyright 2026 ConsenSys Software Inc. All rights reserved.

File: packages/wallet/LICENSE [hunk 2]
```diff
@@ -1,6 +1,18 @@
...
+ 
+You acknowledge and agree that ConsenSys Software Inc. (“ConsenSys”) (or ConsenSys’s licensors) own all legal right, title and interest in and to the work, software, application, source code, documentation and any other documents in this repository (collectively, the “Program”), including any intellectual property rights which subsist in the Program (whether those rights happen to be registered or not, and wherever in the world those rights may exist), whether in source code or any other form.
```
cryptodev-2s, line 3: A side question we are referring here to `repository` instead of `package`, does this means this  is applied to all core packages ? Should we use “this package” ?

cryptodev-2s (dismissed):

packages/wallet/LICENSE
mcmire: Yes I think we should discuss with legal about changing this across the board. If we have ONE license that we can apply to all projects without needing to tweak it in various places then that would be ideal.

mcmire (changes requested):
Marking as "request changes" due to needed copy updates.

packages/wallet/LICENSE
cryptodev-2s: @ajeetd

mcmire (commented):
@cryptodev-2s Should we at least update the changelog? Usually we note a change in license in the changelog. I think we'd also want to note it as `**BREAKING:** as it removes rights from contributors.

mcmire (approved):
LGTM.

PR #10085, opened 2026-09-03
fix(platform-api-docs): prefer source over build output when deduplicating

cryptodev-2s:
## Explanation

A capability declared in a package's source is also visible in the `dist` built from it, and a cross-package import resolves to that `dist` rather than to the sibling's source. Whichever was reached first won, so 83 of 1164 source links pointed at `.d.cts` build output instead of code you can read and edit.

Deduplication now scores source above build output, so the outcome no longer depends on traversal order. All 1164 links point at source, and namespace, action and event counts are unchanged.

Both clients generate byte-identical docs, since they only ever see published packages and score every candidate the same way.

## References

* Follow up to #9990, as promised in https://github.com/MetaMask/core/pull/9990#discussion_r3916262625

## Checklist

---

> [!NOTE]
> **Low Risk**
> Changes only affect generated documentation links and dedup tie-breaking in the docs generator, with no runtime product or security impact.
> 
> **Overview**
> Generated platform API docs were often linking messenger capabilities to **`node_modules/.../dist/*.d.cts`** instead of the editable **`packages/*/src`** declaration, because cross-package imports resolve to build output and whichever duplicate was seen first won during deduplication.
> 
> **Deduplication scoring** now adds a **`sourceScore`** that favors paths outside **`/dist/`**, so when the same `typeString` appears in both source and compiled declarations, the source file wins regardless of scan order. JSDoc and “home package” preferences are unchanged.
> 
> **`scanSources`** collects all scan globs into one pattern list before a single **`addSourceFiles`** call (same bulk-load behavior, slightly cleaner structure). A regression test reproduces the monorepo **`a-controller` → `@metamask/b-controller` dist** ordering case; the changelog records the fix under **Fixed**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b32e92e6782f5003f2924785ab113ca467c52c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/platform-api-docs/src/generate.test.ts
+++ b/packages/platform-api-docs/src/generate.test.ts
@@ -338,6 +338,77 @@ export type FooMessenger = Messenger<'Foo', FooAction, never>;
     });
   });
 
+  it('deduplicates items preferring source over build output', async () => {
+    expect.assertions(2);
+
+    await withinSandbox(async ({ directoryPath }) => {
+      // Reproduces how build output wins in a real monorepo. `a-controller` is
+      // scanned before `b-controller`, and its import of `@metamask/b-controller`
+      // resolves through node_modules to the published `.d.cts`, so the first
+      // declaration seen for `B:get` is the built one. The source declaration is
+      // reached later, when `b-controller/src` is scanned, and should win.
+      const declaration = `
+/** Gets b. */
+export type BGetAction = {
+  type: 'B:get';
+  handler: () => string;
+};
+`;
+
+      const bSrc = path.join(directoryPath, 'packages', 'b-controller', 'src');
+      await fs.promises.mkdir(bSrc, { recursive: true });
+      await fs.promises.writeFile(
+        path.join(bSrc, 'BController.ts'),
+        `${declaration}
+export type BMessenger = Messenger<'B', BGetAction, never>;
+`,
+      );
+
+      const bDist = path.join(
+        directoryPath,
+        'node_modules',
+        '@metamask',
+        'b-controller',
+        'dist',
+      );
+      await fs.promises.mkdir(bDist, { recursive: true });
+      await fs.promises.writeFile(path.join(bDist, 'index.d.cts'), declaration);
+      await fs.promises.writeFile(
+        path.join(bDist, '..', 'package.json'),
+        JSON.stringify({
+          name: '@metamask/b-controller',
+          types: './dist/index.d.cts',
+        }),
+      );
+
+      const aSrc = path.join(directoryPath, 'packages', 'a-controller', 'src');
+      await fs.promises.mkdir(aSrc, { recursive: true });
+      await fs.promises.writeFile(
+        path.join(aSrc, 'AController.ts'),
+        `
+import type { BGetAction } from '@metamask/b-controller';
+
+export type AMessenger = Messenger<'A', BGetAction, never>;
+`,
+      );
+
+      const outputDir = path.join(directoryPath, '.docs');
+      await generate({
+        projectPath: directoryPath,
+        outputDir,
+        strategy: 'scan',
+        scanDirs: ['src'],
+      });
+
+      const actionsMd = await fs.promises.readFile(
+        path.join(outputDir, 'docs', 'B', 'actions.md'),
+        'utf8',
+      );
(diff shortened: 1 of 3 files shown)


cryptodev-2s:
@mcmire In case you have missed to look at this ?

mcmire:
@cryptodev-2s Ah sorry I did miss this, I haven't reviewed this yet. I will review this shortly.

packages/platform-api-docs/src/generate.ts
```diff
@@ -353,36 +360,31 @@ async function scanSources(
...
     );
   }
 
   if (sources.nodeModulesDir) {
     const root = await toGlobPath(sources.nodeModulesDir);
-    sourceFiles.push(...addSourceFiles(project, [`${root}/*/dist/**/*.d.cts`]));
+    patterns.push(`${root}/*/dist/**/*.d.cts`);
   }
 
+  const sourceFiles = addSourceFiles(project, patterns);
```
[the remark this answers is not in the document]
mcmire: This seems like a fair point, but is this an intentional change?

Are the changes to the way that source files are collected necessary to fix the bug with `deduplicationScore` mentioned in the PR description?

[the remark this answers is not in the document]
mcmire: > Are the changes to the way that source files are collected necessary to fix the bug with `deduplicationScore` mentioned in the PR description?

Ah sorry. The PR description reads as though we are fixing a bug. I see that we are not. It's a bit flipped around. The main point _is_ to change the way that source files are collected; the changes to the deduplication are a side effect.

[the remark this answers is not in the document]
mcmire: The point from Cursor still stands though. Is its claim legit?

packages/platform-api-docs/src/generate.test.ts
```diff
@@ -227,6 +227,60 @@ export type TestMessenger = Messenger<'Test', TestGetAction, never>;
...
   });
 
+  it('keeps published declaration files when a scan directory contains node_modules', async () => {
+    expect.assertions(2);
+
+    await withinSandbox(async ({ directoryPath }) => {
+      // Scanning `.` makes the project root a scan directory, so its
+      // `node_modules` exclusion covers the published declaration files too.
+      // Each location has to be collected in its own call for them to survive,
+      // since exclusions apply to every pattern in the call they belong to.
```
mcmire: Is this comment in the right place? Maybe it needs to be moved to above the `generate` call? On the other hand, to me, not knowing this fact doesn't seem to impact readability of this test, so I would also be fine with removing this:

``` suggestion
```

packages/platform-api-docs/src/generate.ts
```diff
@@ -360,31 +360,40 @@ async function scanSources(
   sources: ScanSources,
 ): Promise<MessengerCapabilityPacket[]> {
   const project = createProject();
-  const patterns: string[] = [];
+  const sourceFiles = [];
 
+  // One call per location, not one call carrying every pattern. Exclusions
+  // apply to a whole call, so combining them lets the source-tree
+  // `node_modules` and `dist` exclusions match the published declaration files
+  // and drop them.
```
mcmire: I think this could be written in a clearer way:

``` suggestion
  // NOTE: We are calling `addSourceFiles` for each kind of source instead of
  // calling it at the very end so that at each step we can make sure to exclude
  // `node_modules` and `dist`.

```

mcmire (commented):
Two more suggestions. Everything else looks good.

packages/platform-api-docs/src/generate.test.ts
cryptodev-2s: Yes finally did drop it here https://github.com/MetaMask/core/pull/10085/changes/38afbc9dd94a4c5a7370c0d466ace783bc367648

packages/platform-api-docs/src/generate.ts
cryptodev-2s: Make sense applied here https://github.com/MetaMask/core/pull/10085/changes/38afbc9dd94a4c5a7370c0d466ace783bc367648

mcmire (approved):
LGTM!

PR #10101, opened 2026-09-03
fix(perps-controller): revalidate TP/SL cache misses

geositta:
## Explanation

Immediately after a HyperLiquid position fills, the current WebSocket DEX slice can still omit the new symbol until the next `clearinghouseState` update. `updatePositionTPSL` treated that cache miss as authoritative and returned `POSITION_NOT_FOUND`, so post-order protection could fail even though the position existed on the venue.

This change lets only `updatePositionTPSL` opt into one target-DEX HTTP revalidation when its requested symbol is absent from the current WebSocket slice. A cache hit remains WebSocket-only. The HTTP response is used as a complete replacement rather than merged with cached data, and an unavailable response fails closed with `PROVIDER_NOT_AVAILABLE`.

Close and margin operations keep their existing cache-miss behavior, avoiding broader REST traffic and 429 pressure.

## References

- Related Mobile coordination fix: https://github.com/MetaMask/metamask-mobile/pull/35707
- HyperLiquid position state: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint/perpetuals
- HyperLiquid WebSocket recovery guidance: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/websocket

## Test plan

- `yarn workspace @metamask/perps-controller run jest --no-coverage tests/src/providers/HyperLiquidProvider.advanced-orders.test.ts`
- `yarn workspace @metamask/perps-controller run test`
- `yarn build`
- Changed-file ESLint and formatting checks
- `yarn workspace @metamask/perps-controller run changelog:validate`

## Checklist

### DIFF
--- a/packages/perps-controller/CHANGELOG.md
+++ b/packages/perps-controller/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Revalidate HyperLiquid positions over HTTP before reporting a WebSocket cache miss during a TP/SL update.
+- Revalidate HyperLiquid positions over HTTP before reporting a WebSocket cache miss during a TP/SL update ([#10101](https://github.com/MetaMask/core/pull/10101)).
 
 ## [16.0.0]
 
--- a/packages/perps-controller/CHANGELOG.md
+++ b/packages/perps-controller/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Revalidate HyperLiquid positions over HTTP before reporting a WebSocket cache miss during a TP/SL update ([#10101](https://github.com/MetaMask/core/pull/10101)).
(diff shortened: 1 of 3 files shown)


geositta:
@metamaskbot publish-previews

abretonc7s (changes requested):
Right fix, right layer.

Root cause: `updatePositionTPSL` used `getPositions({ skipCache: true })` in 11.0.0 through 15.0.0. 15.1.0 (#10037) swapped it for the cache-first lookup. Confirmed by unpacking each published version.

Validated on Mobile via a yarn patch of this exact change: revalidation fires, TP attaches, bug gone.

Two asks, both pushed in bbb6b00c43:

- Test for the transient window (empty slice, then populated, asserting one HTTP read). Existing tests only cover fixed cache states, which is the blind spot that let 15.1.0 ship.
- Changelog note on why only `updatePositionTPSL` opts in, and that cache hits and #10037's sizing guarantees are unaffected.

abretonc7s (approved):

abretonc7s:
For the record on why this asymmetry exists: #10037 bundled two changes. Ignoring the caller's `position` snapshot was required, that was TAT-3252. Reading the WebSocket cache first was a REST-traffic optimization, not required.

On the post-order path clients never pass `position`, so 15.0.0 already did an unconditional REST read there. This PR walks the optimization back for the one caller where the cache cannot answer the question.

abretonc7s (approved):

abretonc7s (approved):

Bigshmow (approved):

PR #10104, opened 2026-09-04
feat(core-backend): add timeout to retrieve auth token

Kriys94:
## Explanation

Current Behavior: https://www.loom.com/share/a6884ecd42d7475780838f4c9fdd2a7d

After PR: https://www.loom.com/share/aed73a550d934ac3ab319ba5d4a93a6d

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes when API calls are authenticated versus sent unauthenticated, which can affect rate limits and data visibility until a token arrives; behavior is configurable via `authTokenTimeout`.
> 
> **Overview**
> Adds **`authTokenTimeout`** on `ApiPlatformClientOptions` (default **`DEFAULT_AUTH_TOKEN_TIMEOUT`** = 500ms, exported from the package). `BaseApiClient` now resolves the bearer token via `#fetchBearerToken`, which races `getBearerToken` against that limit so slow token providers no longer block HTTP calls.
> 
> If the token is not ready in time, the request is sent **without** an `Authorization` header while the in-flight TanStack `fetchQuery` keeps running; once it resolves, later requests can attach the token. Set **`authTokenTimeout` to `0`** to restore blocking wait-for-token behavior.
> 
> Bearer-token lookup still uses `fetchQuery` for concurrent de-duplication, but **`staleTime` is now `0`** instead of the removed **`STALE_TIMES.AUTH_TOKEN`** (5-minute JWT cache), leaving token lifetime to AuthenticationController. Tests cover de-duplication, unauthenticated fallback, late auth, and infinite wait.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34ffa74c7859129796e41bfa13cf17454b49f13e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core-backend/src/api/base-client.ts
+++ b/packages/core-backend/src/api/base-client.ts
@@ -9,6 +9,7 @@ import {
   API_URLS,
   STALE_TIMES,
   GC_TIMES,
+  DEFAULT_AUTH_TOKEN_TIMEOUT,
   calculateRetryDelay,
   shouldRetry,
   HttpError,
@@ -44,6 +45,8 @@ export class BaseApiClient {
 
   protected readonly getBearerToken?: () => Promise<string | undefined>;
 
+  protected readonly authTokenTimeout: number;
+
   readonly #queryClientInstance: QueryClient;
 
   /**
@@ -73,6 +76,8 @@ export class BaseApiClient {
     this.clientProduct = options.clientProduct;
     this.clientVersion = options.clientVersion;
     this.getBearerToken = options.getBearerToken;
+    this.authTokenTimeout =
+      options.authTokenTimeout ?? DEFAULT_AUTH_TOKEN_TIMEOUT;
 
     this.#queryClientInstance =
       options.queryClient ??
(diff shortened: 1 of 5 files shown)


sahar-fehri (dismissed):
LGTM!

packages/core-backend/src/api/index.ts
```diff
@@ -15,6 +15,7 @@ export {
   API_URLS,
   STALE_TIMES,
   GC_TIMES,
+  DEFAULT_AUTH_TOKEN_TIMEOUT,
```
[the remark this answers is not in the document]
sahar-fehri: Nit but dnt think we need to export for now, should be fine

packages/core-backend/src/api/base-client.ts
```diff
@@ -90,6 +95,54 @@ export class BaseApiClient {
...
+          const result = await getBearerToken();
+          // Throw if no token - prevents caching null/undefined
+          // so subsequent requests can retry (e.g., after user logs in)
+          if (!result) {
+            throw new Error('No bearer token available');
+          }
+          return result;
+        },
+        staleTime: STALE_TIMES.AUTH_TOKEN,
+        retry: false, // Don't retry auth failures
```
Prithpal-Sooriya: Are we sure this is the behaviour we want?

1. staleTime - the underlying auth controller already manages this token caching, do we really want this second caching system?
2. retry - is this configurable
    - Should _all_ authenticated requests outright fail without retry?
    - Should authenticated be _queued_ until auth token is ready?

packages/core-backend/src/api/base-client.ts
```diff
@@ -90,6 +95,54 @@ export class BaseApiClient {
...
+      return await tokenPromise;
+    }
+
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+    const timeoutPromise = new Promise<undefined>((resolve) => {
+      timeoutId = setTimeout(() => resolve(undefined), this.authTokenTimeout);
+    });
+
+    try {
+      return await Promise.race([tokenPromise, timeoutPromise]);
```
Prithpal-Sooriya, line 141: Instead of promise races, can we use abort controllers instead?

```ts
const timeoutSignal = AbortSignal.timeout(timeoutMs);

// signal can be used for aborting requests
const signal = AbortSignal.any([mainSignal, timeoutSignal]);
```

packages/core-backend/src/api/base-client.ts
Prithpal-Sooriya: TBH this seems to only be used in 1 area, so this is fine for now, but these are possible things to consider as more teams use auth & core-backend.

Prithpal-Sooriya (dismissed):

packages/core-backend/src/api/base-client.ts
Prithpal-Sooriya: Actually good call. Instead of aborting, we keep the auth promise alive running in the background.

packages/core-backend/src/api/base-client.ts
Kriys94: I took these values from the current behavior, but agree that we can just have staleTime 0 and rely on AuthController cache. For the retry, I prefer to keep it as is for now.

Prithpal-Sooriya (approved):

PR #10109, opened 2026-09-04
feat(transaction-controller): add approval-time sponsorship/signing hooks

pedronfigueiredo:
## Related pull requests

- Extension: [MetaMask/metamask-extension#46060](https://github.com/MetaMask/metamask-extension/pull/46060)
- Mobile: [MetaMask/metamask-mobile#35754](https://github.com/MetaMask/metamask-mobile/pull/35754)

## Summary

- Add optional approval-time `isSponsored` and `shouldSign` hooks to `TransactionController`, with named object results for future extensibility.
- Default omitted hooks to non-sponsored transactions with local signing.
- Deprecate the legacy `isGasFeeSponsored` and `isExternalSign` lifecycle properties while retaining them in public types for compatibility; `isGasFeeSponsored` remains available as migration metadata.
- When simulation is enabled, refresh gas-fee-token and sponsorship availability after approval preparation and expose it as `TransactionMeta.isGasFeeSponsoredAvailable` before sponsorship policy runs.
- Reserve a nonce and sign locally only for non-sponsored transactions when `shouldSign` returns true.
- Preserve existing nonces when local signing is skipped and include newly reserved nonces in the `approved` status event.
- Require a client publish hook when local signing is skipped. Unsigned publication is owned by that hook, and the controller does not fall back to `eth_sendRawTransaction` with an empty payload or require the hook to return a transaction hash.

## Public API

```ts
hooks?: {
  afterAdd?: AfterAddHook;
  beforeCheckPendingTransaction?: (
    transactionMeta: TransactionMeta,
  ) => Promise<boolean>;
  beforePublish?: (
    transactionMeta: TransactionMeta,
  ) => Promise<boolean>;
  beforeSign?: BeforeSignHook;
  isSponsored?: (request: {
    transactionMeta: TransactionMeta;
  }) => Promise<{ isSponsored: boolean }>;
  publish?: (
    transactionMeta: TransactionMeta,
  ) => Promise<{ transactionHash: string }>;
  publishBatch?: PublishBatchHook;
  shouldSign?: (request: {
    transactionMeta: TransactionMeta;
  }) => Promise<{ shouldSign: boolean }>;
}
```

The new hooks are optional. Without them, transactions retain the existing non-sponsored, locally signed behavior.

## Approval behavior

1. Mark the transaction as `approved` and run `beforeSign` preparation.
2. Run the extracted approval-preparation helper, which refreshes gas-fee tokens and `isGasFeeSponsoredAvailable` when simulation is enabled (or selected-token preflight requires it), then performs the gas-fee-token preflight.
3. Call `isSponsored` (default: `false`).
4. If sponsored, skip `shouldSign`, nonce reservation, and local signing.
5. Otherwise call `shouldSign` (default: `true`); reserve a nonce and sign only when it returns `true`.
6. Publish the `approved` status after any local nonce reservation, then continue through signing and publication.
7. If local signing is skipped, require and defer to a publish hook. Do not invoke the provider fallback with the unsigned `0x` sentinel; the hook may return without a transaction hash.

`beforeSign` always runs, including when local signing is skipped. Existing nonces are retained on externally prepared transactions. Hook outcomes are transient policy results and are not persisted as transaction metadata.

## Validation

- `yarn workspace @metamask/transaction-controller run test`
- `yarn workspace @metamask/transaction-controller run build`
- `yarn eslint <changed TypeScript files>`
- `yarn prettier --check <changed TypeScript files>`
- `yarn changelog:validate`

The package test suite, package build, focused ESLint, changed-TypeScript formatting checks, changelog validation, and branch diff check pass.

---

> [!NOTE]
> **High Risk**
> Breaking changes to core transaction approval, nonce, signing, and publish behavior affect every client that relied on deprecated sponsorship/external-sign metadata or simulation-driven sponsorship.
> 
> **Overview**
> **Breaking:** Sponsorship and local signing are no longer driven by `isGasFeeSponsored`, `isExternalSign`, or simulation’s old `isGasFeeSponsored` flag. Integrators must use optional approval-time **`isSponsored`** and **`shouldSign`** hooks (defaults: not sponsored, sign locally).
> 
> On approval, the controller runs **`beforeSign`**, refreshes gas-fee tokens and sets **`TransactionMeta.isGasFeeSponsoredAvailable`** when simulation (or token preflight) requires it, then evaluates the hooks. Sponsored flows skip **`shouldSign`**, nonce reservation, and local signing. When **`shouldSign`** is false, existing nonces are kept, **`beforeSign`** still runs, and publication **requires** a **`publish`** hook—the controller will not fall back to **`eth_sendRawTransaction`** with an empty payload.
> 
> Simulation/gas-fee helpers now expose **sponsorship availability** (`isGasFeeSponsoredAvailable`) instead of treating simulation as the lifecycle policy. Legacy fields remain on types as **deprecated** migration metadata only. Tests and changelog document the new behavior, including batch gas-fee poller updates and stricter publish paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71e88f2d285a1d63ac55117e63d81d33fa555d0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jpuri (dismissed):

packages/transaction-controller/src/utils/prepare-transaction-for-approval.ts
```diff
@@ -0,0 +1,117 @@
...
+ * token, removing the nonce, and selecting external signing at publication.
+ *
+ * @param request - Transaction metadata and normalized preparation facts.
+ * @param request.signing - Signing facts for the selected publication path.
+ * @param request.sponsorship - Sponsorship facts determined by the caller.
+ * @param request.transactionMeta - Transaction metadata to prepare.
+ * @returns Prepared transaction metadata and the resulting decisions.
+ * @throws If sponsorship is required but no sponsored publication path is available.
+ */
+export function prepareTransactionForApproval({
```
matthewwalsh0: As discussed, these state properties have grown massively in complexity and currently being awkwardly changed during lifecycle to control the flow.

So better to experiment with a callback approach such as `isSponsored(tx)`, `shouldSign(tx)` and `isGasFeeSponsoredAvailable` and `isIntent` properties.

So we have a single source of truth per client, fixed metadata, and a logical injection point during the actual submission flow.

matthewwalsh0 (changes requested):

pedronfigueiredo:
Fixed the lint:misc:check failure (Prettier formatting in TransactionController.test.ts) and pushed 58ac50582e. GitHub checks are green again; @matthewwalsh0, could you re-review when you have a moment?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -413,6 +415,16 @@ export type TransactionControllerOptions = {
       transactionMeta: TransactionMeta,
     ) => Promise<boolean>;
 
+    /**
+     * Additional logic to determine whether a transaction is sponsored.
```
matthewwalsh0, line 419: I worry we're maintaining the ambiguity we're trying to avoid if this is optional, should these not the be the only source of truth?

Then we can remove the original properties entirely for maximum clarity and make this a breaking change?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -413,6 +415,16 @@ export type TransactionControllerOptions = {
...
+     */
+    isSponsored?: IsSponsoredHook;
```
matthewwalsh0, line 421: Minor, alphabetical order of these properties?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -413,6 +415,16 @@ export type TransactionControllerOptions = {
...
+
+    /**
+     * Additional logic to determine whether a transaction should be signed locally.
+     */
+    shouldSign?: ShouldSignHook;
```
matthewwalsh0, line 426: Should we clarify in the JSDoc the distinction between this and `beforeSign`?

That this is purely the decision where as `beforeSign` is extra logic to run at that point?

packages/transaction-controller/src/types.ts
```diff
@@ -2130,6 +2130,21 @@ export type AfterAddHook = (request: {
   updateTransaction?: (transaction: TransactionMeta) => void;
 }>;
 
+/**
+ * Custom logic to determine whether a transaction should be treated as sponsored.
+ */
+export type IsSponsoredHook = (request: {
```
matthewwalsh0, line 2136: Minor, worth putting at the bottom if chronological or keeping the hook types together?

packages/transaction-controller/src/types.ts
```diff
@@ -2130,6 +2130,21 @@ export type AfterAddHook = (request: {
...
+  transactionMeta: TransactionMeta;
+}) => Promise<boolean>;
```
matthewwalsh0, line 2138: Should we return objects for both of these (with properties like `isSponsored` and `shouldSign`) to keep them future-proof?

packages/transaction-controller/src/utils/prepare.ts
```diff
@@ -104,13 +104,7 @@ function normalizeAuthorizationList(
...
-  if (!value) {
-    return value;
-  }
-
-  if (value === '0x0') {
-    return '0x';
-  }
-
-  return (value.replace?.(/^0x(00)+/u, '0x') as Hex) ?? value;
+  return value === '0x0'
```
matthewwalsh0, line 107: Unrelated? Can we remove for minimum diff?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -833,6 +849,22 @@ export class TransactionController extends BaseController<
       /* istanbul ignore next */
       hooks?.beforeCheckPendingTransaction ??
       ((): Promise<boolean> => Promise.resolve(true));
+    this.#isSponsored =
+      hooks?.isSponsored ??
+      (async ({
+        transactionMeta,
+      }: {
+        transactionMeta: TransactionMeta;
+      }): Promise<boolean> => Boolean(transactionMeta.isGasFeeSponsored));
```
matthewwalsh0, line 858: As noted above, I'd vote to remove these properties entirely to avoid the ambiguity.

We discussed adding a new `isIntent` property though to keep the pay controller sign skip at the controller level, but I wonder if cleaner if we handle that entirely in the client so clients can just say don't sign as I see transaction pay quotes?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -1240,7 +1272,8 @@ export class TransactionController extends BaseController<
     } else {
       const newTransactionMeta = cloneDeep(addedTransactionMeta);
 
-      this.#updateGasProperties(newTransactionMeta)
+      // eslint-disable-next-line no-void
+      void this.#updateGasProperties(newTransactionMeta)
```
matthewwalsh0, line 1276: Is this better if we need the comment also?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -2386,21 +2420,20 @@ export class TransactionController extends BaseController<
       pickBy(transactionsToFilter, (transaction) => {
         // iterate over the predicateMethods keys to check if the transaction
         // matches the searchCriteria
+        const txParams = transaction.txParams as Record<string, unknown>;
```
matthewwalsh0, line 2423: Also unrelated, just auto-linting, can we remove for minimum diff?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
+          this.#updateTransactionInternal({ transactionId: txId }, fn),
+      });
 
       // eslint-disable-next-line require-atomic-updates
       transactionMeta = this.#getTransactionOrThrow(transactionId);
 
+      const isSponsored = await this.#isSponsored({ transactionMeta });
+      const shouldSign = await this.#shouldSign({
+        transactionMeta,
+        isSponsored,
```
matthewwalsh0, line 3187: Why do we pass `isSponsored` to `shouldSign`?

Wouldn't `isSponsored` always mean we don't want to sign so can skip the `shouldSign` call?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
+      });
+
+      // eslint-disable-next-line require-atomic-updates
+      transactionMeta = this.#updateTransactionInternal(
+        {
+          transactionId,
+        },
+        (draftTxMeta) => {
+          draftTxMeta.isGasFeeSponsored = isSponsored;
```
matthewwalsh0, line 3196: I'm torn on this. It feels wrong to start re-using an old property with a different meaning, especially given the risk to legacy code.

Should we at least add the discussed `isGasFeeSponsorshipAvailable` for clarity and to help the client hooks?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
+          draftTxMeta.isExternalSign = !shouldSign;
```
matthewwalsh0, line 3197: Hopefully this can be removed as too powerful be a property as discussed.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
+
+          if (!shouldSign) {
+            draftTxMeta.txParams.nonce = undefined;
```
matthewwalsh0, line 3200: Is it possible it has a `nonce` already?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
           draftTxMeta.txParams.gasLimit = gas;
-          draftTxMeta.txParams.nonce = nonce;
 
           if (!type && isEIP1559Transaction(txParams)) {
             draftTxMeta.txParams.type = TransactionEnvelopeType.feeMarket;
           }
         },
       );
 
-      this.#onTransactionStatusChange(transactionMeta);
```
matthewwalsh0, line 3148: Do we not still need to fire this for the `approved` status?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3156,87 @@ export class TransactionController extends BaseController<
...
+            transactionId,
+          },
+          (draftTxMeta) => {
+            draftTxMeta.txParams.nonce = nonce;
+          },
+        );
+
+        rawTx = await this.#trace(
+          { name: 'Sign', parentContext: traceContext },
+          () => this.#signTransaction(transactionMeta, true, true),
```
matthewwalsh0, line 3231: Is this not the only call, why do we need the two boolean args?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3691,21 +3772,37 @@ export class TransactionController extends BaseController<
...
-        this.#updateTransactionInternal({ transactionId: txId }, fn),
-    });
+    if (!skipBeforeSign) {
+      transactionMeta = await this.#applyBeforeSignHook(transactionMeta);
+    }
+
+    if (!skipGasFeeTokenCheck) {
+      const { networkClientId } = transactionMeta;
+
+      await checkGasFeeTokenBeforePublish({
```
matthewwalsh0, line 3793: Is this redundant now, as we already call this before signing?

matthewwalsh0 (changes requested):

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed. I made `isSponsored` and `shouldSign` required and removed `isGasFeeSponsored` / `isExternalSign` as lifecycle control properties, so approval policy now has a single client-owned source of truth. This is intentionally breaking.

Addressed in `fafc3e0ba9` ([commit](https://github.com/MetaMask/core/commit/fafc3e0ba942f01a0e6827b9f2e26b10a1b2dcb8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Good call. I reordered the hook properties alphabetically.

Addressed in `7e1183f7cc` ([commit](https://github.com/MetaMask/core/commit/7e1183f7ccac2effd1fbc1caa70b77ef1f5ae47b)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Clarified the distinction: `beforeSign` performs optional preparation and still runs when local signing is skipped, while `shouldSign` only decides whether the controller reserves a nonce and signs locally.

Addressed in `1c1099ff63` ([commit](https://github.com/MetaMask/core/commit/1c1099ff63b296dee5e133922d98f611ff546deb)).

packages/transaction-controller/src/types.ts
pedronfigueiredo: I kept the hook types together and ordered them alphabetically to match the controller options.

Addressed in `86b3554f65` ([commit](https://github.com/MetaMask/core/commit/86b3554f65c1093b74126b89d30a4ecaddbf20b6)).

packages/transaction-controller/src/types.ts
pedronfigueiredo: Updated both hooks to return named result objects (`{ isSponsored }` and `{ shouldSign }`), leaving room to add fields without changing the return shape again.

Addressed in `7d2e207851` ([commit](https://github.com/MetaMask/core/commit/7d2e207851ca7e50f7139b4d5dbb37a682a941f7)).

packages/transaction-controller/src/utils/prepare.ts
pedronfigueiredo: Removed the unrelated `removeLeadingZeroes` refactor to keep the PR focused.

Addressed in `f0b17fa009` ([commit](https://github.com/MetaMask/core/commit/f0b17fa0099d4d3b6e73c0f3d1c7a64ad3114a27)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Addressed by removing `isGasFeeSponsored` and `isExternalSign`, making the approval hooks authoritative, and leaving the signing decision with clients through `shouldSign`.

Addressed in `fafc3e0ba9` ([commit](https://github.com/MetaMask/core/commit/fafc3e0ba942f01a0e6827b9f2e26b10a1b2dcb8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed, removed the unnecessary `void` and lint suppression while retaining the existing rejection handling.

Addressed in `94abe2597b` ([commit](https://github.com/MetaMask/core/commit/94abe2597ba7cb27edb1bfc35d79624123fb0bfb)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Removed the unrelated predicate lookup refactor and restored the existing implementation to keep the diff focused.

Addressed in `3488ae0469` ([commit](https://github.com/MetaMask/core/commit/3488ae0469284eb3e774f495bb10dfeb20b366d7)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Updated the approval flow so sponsored transactions skip `shouldSign` and local signing automatically; `shouldSign` now handles only non-sponsored transactions.

Addressed in `df672f2ac8` ([commit](https://github.com/MetaMask/core/commit/df672f2ac858dd626560d72795aae6a18e7c89ed)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Addressed by removing the legacy sponsorship/signing properties entirely rather than reusing their meanings. Sponsorship policy now remains client-owned through the required `isSponsored` hook.

Addressed in `fafc3e0ba9` ([commit](https://github.com/MetaMask/core/commit/fafc3e0ba942f01a0e6827b9f2e26b10a1b2dcb8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Removed `isExternalSign` entirely. The `shouldSign` decision is now transient approval policy and is not persisted on transaction metadata.

Addressed in `fafc3e0ba9` ([commit](https://github.com/MetaMask/core/commit/fafc3e0ba942f01a0e6827b9f2e26b10a1b2dcb8)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Good point. Transactions that skip local signing now retain any existing nonce; the controller only assigns a nonce when it signs locally.

Addressed in `3eecb8d915` ([commit](https://github.com/MetaMask/core/commit/3eecb8d915315c748385eb6f957f970cf54e4eda)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Restored the `approved` status notification at the transition point. Locally signed transactions now publish `approved`, `signed`, then `submitted` without duplicating `signed`.

Addressed in `73f04e2735` ([commit](https://github.com/MetaMask/core/commit/73f04e2735e08aedb55552f2e7551cd7a44262dc)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Simplified the sole `#signTransaction` call and removed both skip flags. Approval now performs preparation once before invoking the focused signing method.

Addressed in `eaf345d7d7` ([commit](https://github.com/MetaMask/core/commit/eaf345d7d772164ece3bc361cee3f900d2052aeb)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Yes. Removed the redundant preprocessing from `#signTransaction`; `beforeSign` and the gas-fee-token check now run once in the approval flow.

Addressed in `eaf345d7d7` ([commit](https://github.com/MetaMask/core/commit/eaf345d7d772164ece3bc361cee3f900d2052aeb)).

packages/bridge-status-controller/src/strategy/batch-strategy.ts
```diff
@@ -45,7 +45,6 @@ export async function* submitBatchHandler(
       quoteResponse.quote.gasIncluded,
       isDelegatedAccount,
     ),
-    isGasFeeSponsored: Boolean(quoteResponse.quote.gasSponsored),
```
matthewwalsh0, line 48: Sorry Pedro, if this is causing many changes in other packages, maybe easier to mark as deprecated first and still remove any usages in `TransactionController`, then we can tidy up in a separate smaller PR with lower priority?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -420,15 +422,27 @@ export type TransactionControllerOptions = {
...
     publish?: (
       transactionMeta: TransactionMeta,
     ) => Promise<{ transactionHash: string }>;
     publishBatch?: PublishBatchHook;
+
+    /**
+     * Policy logic to determine whether to reserve a nonce and sign locally.
+     * Use `beforeSign` for transaction preparation instead.
+     */
+    shouldSign: ShouldSignHook;
```
matthewwalsh0: I'd still vote all new hooks be optional as a client may not support sponsorship nor require signing skip? So default to sign and not sponsored if absent.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3140,72 @@ export class TransactionController extends BaseController<
...
+          {
+            transactionId,
+          },
+          (draftTxMeta) => {
+            draftTxMeta.txParams.nonce = nonce;
+          },
+        );
+      }
+
+      this.#onTransactionStatusChange(transactionMeta);
```
matthewwalsh0, line 3199: Good placement, so this is for the `approved` status but we ensure `nonce` is included if generated?

packages/transaction-controller/src/types.ts
```diff
@@ -2143,6 +2130,22 @@ export type BeforeSignHook = (request: {
   | undefined
 >;
 
+/**
+ * Custom logic to determine whether a transaction should be treated as sponsored.
+ */
+export type IsSponsoredHook = (request: {
```
matthewwalsh0: Minor, `IsGasSponsored` more explicit?

packages/transaction-controller/src/types.ts
```diff
@@ -224,11 +224,6 @@ export type TransactionMeta = {
    */
   isFirstTimeInteraction?: boolean;
 
-  /**
-   * Whether the transaction is sponsored meaning the user does not pay the gas fee.
-   */
-  isGasFeeSponsored?: boolean;
```
matthewwalsh0, line 230: Can we replace with `isGasFeeSponsoredAvailable` as clients will need that in their constructor callbacks?

matthewwalsh0 (changes requested):

packages/bridge-status-controller/src/strategy/batch-strategy.ts
pedronfigueiredo: Done. I retained `isGasFeeSponsored` and `isExternalSign` in the public types as deprecated compatibility properties, preserved `isGasFeeSponsored` as migration metadata, removed their lifecycle control inside `TransactionController`, and reverted the cross-package cleanup from this PR.

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Done. Both hooks—and the outer `hooks` object—are now optional. Missing `isSponsored` defaults to `{ isSponsored: false }`; missing `shouldSign` defaults to `{ shouldSign: true }`.

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Yes. The approved status notification is emitted after local nonce reservation, and the test asserts that the first `approved` snapshot contains the generated nonce.

packages/transaction-controller/src/types.ts
pedronfigueiredo: Done. The exported hook type is now `IsGasSponsoredHook`.

packages/transaction-controller/src/types.ts
pedronfigueiredo: Done. `TransactionMeta.isGasFeeSponsoredAvailable` now exposes the simulation result, and approval preparation refreshes it before invoking `isSponsored` so the callback receives the current value.

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -4478,6 +4548,10 @@ export class TransactionController extends BaseController<
 
         ({ transactionHash } = await publishHook(transactionMeta, signedTx));
 
+        if (signedTx === '0x' && !transactionHash) {
```
matthewwalsh0: I'm not sure we can make this assumption here, should we not defer to the publish hooks to own this? Why did we need this?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3137,24 +3150,95 @@ export class TransactionController extends BaseController<
...
+      transactionMeta = await this.#applyBeforeSignHook(transactionMeta);
 
-      const rawTx = await this.#trace(
-        { name: 'Sign', parentContext: traceContext },
-        () => this.#signTransaction(transactionMeta),
-      );
+      const { networkClientId } = transactionMeta;
+
+      let gasFeeTokens = transactionMeta.gasFeeTokens ?? [];
+      const shouldRefreshGasFeeTokens =
```
matthewwalsh0: Is this new since last review?

Is this just so we can populate `gasFeeTokens` and `isGasFeeSponsoredAvailable` before we call our new hooks?

Could this not go in a separate function as feels like lots of duplication?

packages/transaction-controller/src/TransactionController.ts
```diff
@@ -3164,10 +3248,16 @@ export class TransactionController extends BaseController<
...
       }
 
-      if (!rawTx && !transactionMeta.isExternalSign) {
+      if (!rawTx && shouldSign) {
         return ApprovalState.NotApproved;
       }
 
+      if (!shouldSign && !publishHookOverride && !this.#publishHookProvided) {
+        throw new Error(
+          'Cannot publish a transaction without local signing or a publish hook',
```
matthewwalsh0: Minor, `local signing a no publish hook` for clarity.

matthewwalsh0 (dismissed):

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Good point. The check was added to prevent the controller from falling back to `eth_sendRawTransaction` with the `0x` sentinel when local signing is skipped. I’ll remove the requirement that the hook return a hash and instead suppress only the controller fallback for unsigned flows, leaving the publish hook responsible for publication.

Addressed in `88ebffe8ac` ([commit](https://github.com/MetaMask/core/commit/88ebffe8ac4f528a40a47a3c4198da822a36c524)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Yes, this is new since the last review. It refreshes `gasFeeTokens` and `isGasFeeSponsoredAvailable` immediately before the new policy hooks so they receive current sponsorship data, then runs the existing gas-fee-token preflight. I’ll extract that preparation into a focused private helper to keep the approval flow concise.

Addressed in `cc1a3b06a2` ([commit](https://github.com/MetaMask/core/commit/cc1a3b06a2a6a40d659522e36e99b5eec4dc45c9)).

packages/transaction-controller/src/TransactionController.ts
pedronfigueiredo: Agreed. I’ll reword the error to make it explicit that both conditions apply: `Cannot publish a transaction when there is no local signing and no publish hook`.

Addressed in `a3526649c5` ([commit](https://github.com/MetaMask/core/commit/a3526649c5af09f8d80ef1aa5c4778b43ea87be3)).

matthewwalsh0 (dismissed):

OGPoyraz (approved):

PR #10116, opened 2026-09-04
feat(ramps-controller): add neobank onboarding stage lookup

roz0n:
## Explanation

Adds `RampsController.hydrateNeobankStore({ walletAddress })` and persisted `state.neobank` so clients can resume Money Account onboarding from authoritative Core state.

The lookup refreshes KYC first, then checks wallet registration and autoramp readiness only after KYC completes. It returns one `NeobankOnboardingStage`. It does not navigate, launch SumSub, register a wallet, or create an autoramp.

Final KYC behavior:

- Missing vendor or provider terms stays in the matching terms stage.
- SumSub `abandoned` or retryable `failed` returns `KycStartedIncomplete`.
- SumSub `vendorProcessing` and user status `pending` return `KycPending` and must not relaunch the SDK.
- Controller phase `error` returns `LookupFailed`.
- `KycRejected` is reserved for user status `terminal-failure`.
- Mobile owns routing. A dedicated KYC page hook owns controller initialization when SumSub should be open.

Hosts must delegate the exported required KYC actions.

## References

- [TRAM-3916](https://consensyssoftware.atlassian.net/browse/TRAM-3916)
- [Mobile Engine wiring #35896](https://github.com/MetaMask/metamask-mobile/pull/35896)
- [Mobile routing #35894](https://github.com/MetaMask/metamask-mobile/pull/35894)

## Checklist

- [x] Tests updated
- [x] Documentation updated
- [x] Changelog updated

[TRAM-3916]: https://consensyssoftware.atlassian.net/browse/TRAM-3916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ramps-controller/src/neobank-onboarding.ts
```diff
@@ -0,0 +1,265 @@
...
+import type { RegistrationStatus } from './wallet-registration-service.js';
+
+/**
+ * Single Mobile-routable stage for NeoBank / Money Account onboarding.
+ *
+ * Ordered roughly by funnel position. Terms 1 = vendor disclaimers; Terms 2 =
+ * SumSub + idOS provider disclaimers (submitted as a batch — if either is
+ * missing after a partial accept, Mobile routes back to Terms 2).
+ */
+export enum NeobankOnboardingStage {
```
roz0n, line 27: Note to self: Review final states before merging

roz0n: Reviewed the final states against KYC controller statuses and the SumSub launch rule (mount KYCPage + page hook; do not launch from a router method).

Kept the funnel: NoUser / EmailOtp / VendorTerms / ProviderTerms → KYC stages → wallet → autoramp → LookupFailed. Core still returns a stage only.

Derivation fixes in c05e25b76:
- `abandoned` and retryable SumSub `failed` → `KycStartedIncomplete` (KYCPage hook may resume / relaunch)
- `vendorProcessing` and `userStatus === pending` → `KycPending` (do **not** launch the SDK)
- `phase === error` → `LookupFailed`, not a terminal KYC reject
- `form` / `check` count as in-progress, not `KycNotStarted`
- `KycRejected` is only `userStatus === terminal-failure`

Mobile should map `KycNotStarted` / `KycStartedIncomplete` / `KycNeedsReview` to KYCPage, and T&C stages elsewhere. No `LaunchSumsub` stage or Core launch action.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -3574,6 +3639,127 @@ export class RampsController extends BaseController<
...
+   * wallet registration or autoramp creation (see TRAM-3924). Clients must not
+   * persist UKYC `sessionId` or re-register; this method is lookup-then-derive
+   * only.
+   *
+   * @param params - Hydration parameters.
+   * @param params.walletAddress - Money Account wallet address used for
+   * registration / autoramp checks after KYC is complete.
+   * @returns The derived {@link NeobankOnboardingStage}.
+   */
+  async hydrateNeobankStore({
```
roz0n, line 3656: GW: do we need `walletAddress` passed here? (might be ok)

packages/ramps-controller/src/RampsController.ts
```diff
@@ -3574,6 +3639,127 @@ export class RampsController extends BaseController<
...
+      status: 'skipped',
+    };
+    let autoramp: NeobankOnboardingDerivationInput['autoramp'] = {
+      status: 'skipped',
+    };
+    let lookupError: string | null = null;
+
+    if (kycState.userStatus === 'completed') {
+      try {
+        const customerId = await this.resolveAutorampCustomerId();
```
georgeweiler, line 3703: see comment above

packages/ramps-controller/src/RampsController.ts
```diff
@@ -3574,6 +3639,127 @@ export class RampsController extends BaseController<
...
+      await this.messenger.call('KycController:refreshKycStatus');
+    } catch (error) {
+      return this.#commitNeobankStage(
+        NeobankOnboardingStage.LookupFailed,
+        String(error),
+      );
+    }
+
+    const kycState = this.messenger.call('KycController:getState');
+    const identity = this.messenger.call('KycController:getCustomerIdentity');
```
georgeweiler, line 3678: I think we have an ID issue here - this method will only return a moonpay customer ID and for Iron will be null

We should determine whether the user exists using the Iron KYC state. Then after KYC is complete we can continue using the existing Profile Sync ID customer lookup like you're doing below

georgeweiler: see: https://github.com/MetaMask/core/blob/main/packages/kyc-controller/src/KycController.ts#L1829-L1838

packages/ramps-controller/src/RampsController.ts
```diff
@@ -3574,6 +3639,127 @@ export class RampsController extends BaseController<
...
+    const kycInput: NeobankOnboardingDerivationInput['kyc'] = {
+      phase: kycState.phase,
+      userStatus: kycState.userStatus,
+      sumsubStatus: kycState.sumsub.status,
+      hasCustomerIdentity,
+      hasVendorTerms,
+      hasProviderTerms,
+    };
+
+    let wallet: NeobankOnboardingDerivationInput['wallet'] = {
```
georgeweiler, line 3693: I don't think it's necessary to stop here. Adding wallet signature & autoramp account creation in a separate PR & release will be slow.


The methods for those are already available - let's just finish up the state machine in this PR.


here's some pseudocode to keep a birds eye view of what we want from this. Remember this hydrate func will be called in a 5s poll and will update the controller state with the user's stage and then also take action to sign & create if needed.

```
hydrate() {
  // 1. terms / email / KYC  (already mostly here)
  if (!terms1) → VendorTermsRequired
  if (!terms2) → ProviderTermsRequired
  if (!email OTP) → EmailOtpRequired
  if KYC not done → KycNotStarted / Incomplete / Pending / Rejected / NeedsReview
  // 2. KYC success (this is what is missing)
  if wallet not registered → sign + register
  if autoramp missing → create
  if autoramp still pending → AutorampPending
  → AutorampCreated / ready
}
```

georgeweiler: >  Remember this hydrate func will be called in a 5s poll 

btw It would be great to add breakpoints so this function can be idempotent and we can avoid extra calls.

if the user has an autoramp account we can safely assume they've completed all previous steps. There's no need for 5s async poll to iron to confirm the user's T&C acceptance, for example.

packages/ramps-controller/src/RampsController.ts
```diff
@@ -3574,6 +3639,127 @@ export class RampsController extends BaseController<
...
+    const hasCustomerIdentity =
+      Boolean(identity?.id ?? kycState.email) || kycState.userStatus !== null;
+
+    const kycInput: NeobankOnboardingDerivationInput['kyc'] = {
+      phase: kycState.phase,
+      userStatus: kycState.userStatus,
+      sumsubStatus: kycState.sumsub.status,
+      hasCustomerIdentity,
+      hasVendorTerms,
+      hasProviderTerms,
```
jiexi, line 3690: IMO the ramps-controller should not be concerned about anything except the userStatus. If the userStatus is `not-started`, `new`, or `needs-more-information`, then you should be responsible for redirecting the user back to the kyc flow and allowing the kyc side (controller/components) to resume

packages/ramps-controller/src/RampsController.ts
jiexi: see my comment below about how RampsController should not have to consume anything but the userStatus

PR #10123, opened 2026-09-07
fix: keep persisted feature flags for empty segmentation id

mikesposito:
## Explanation

`init()` re-derives the effective flags from the persisted `rawRemoteFeatureFlags`. For threshold flags that means grouping the user by `getMetaMetricsId` or `getCanonicalProfileId`, and both are documented as returning an empty string when the identifier isn't available. When that happened we stored the raw threshold array as the flag value, so a flag that had a resolved value last session came back unresolved.

That isn't hypothetical. On the extension, `wallet.init()` runs before the controllers backing both getters are registered, so both return an empty string on every startup, and consumers reading a threshold flag get an array instead of a value until the next fetch.

With this change, when there's no identifier we keep the value from the previous session rather than the array, plus its entry in `featureFlagThresholdGroups` so the value and its group stay in sync. Flags with nothing to preserve, as on a fresh install, still fall back to the array. Same reasoning already applied to the no-raw-flags branch in #9816: leaving a flag ungrouped beats flipping it for the end user.

## References

See: https://github.com/MetaMask/metamask-extension/pull/45749#discussion_r3947558519

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes threshold flag resolution during `init`, which affects effective flag values for all consumers until the next fetch, though the behavior is intentionally conservative.
> 
> **Overview**
> Fixes a startup bug where **`init()`** could replace resolved threshold feature flag values with the raw threshold **array** when `getMetaMetricsId` or `getCanonicalProfileId` return an empty string (common before their backing controllers are ready).
> 
> During threshold processing, if there is no segmentation id the controller now **reuses the last persisted resolved value** from `#processedRemoteFeatureFlags` and restores the matching **`featureFlagThresholdGroups`** entry so value and group stay aligned. Fresh installs with nothing to preserve still get the unresolved array.
> 
> Adds a regression test for empty canonical profile id on `init` and documents the fix in the package changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 652a1b81e8416b44455b736548d08f8985aec627. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remote-feature-flag-controller/CHANGELOG.md
+++ b/packages/remote-feature-flag-controller/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Keep the previously selected value for a threshold feature flag when the segmentation identifier is unavailable, instead of exposing the unresolved threshold array ([#0000](https://github.com/MetaMask/core/pull/0000))
+- Keep the previously selected value for a threshold feature flag when the segmentation identifier is unavailable, instead of exposing the unresolved threshold array ([#10123](https://github.com/MetaMask/core/pull/10123))
   - `getMetaMetricsId` and `getCanonicalProfileId` may legitimately return an empty string, as they do when `init` runs before whatever backs them is ready. Previously `init` would then replace resolved threshold values with the raw arrays, flipping those flags for the end user until the next fetch. Flags with no previously selected value, as on a fresh install, still fall back to the raw array.
 
 ## [6.1.0]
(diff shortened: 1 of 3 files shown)


packages/remote-feature-flag-controller/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/utils` from `^11.11.0` to `^11.12.0` ([#10076](https://github.com/MetaMask/core/pull/10076))
 
+### Fixed
+
+- Keep the previously selected value for a threshold feature flag when the segmentation identifier is unavailable, instead of exposing the unresolved threshold array ([#0000](https://github.com/MetaMask/core/pull/0000))
```
mikesposito:
```suggestion
- Keep the previously selected value for a threshold feature flag when the segmentation identifier is unavailable, instead of exposing the unresolved threshold array ([#10123](https://github.com/MetaMask/core/pull/10123))
```

tommasini (approved):

PR #10131, opened 2026-09-07
fix(assets-controller): make #start() re-entrancy safe

juanmigdr:
## Explanation

`AssetsController#start()` guards against running twice with `#activeSubscriptions.size > 0`, but that only becomes true after the initial `#runStartupRefresh()` call resolves — including its forced `getAssets()` call, which can take many seconds for accounts with many chains/snaps.

`#start()` is triggered by several independent events (keyring unlock, account tree init, client state change). If a second one of these fires before the first `getAssets()` call resolves, the existing guard doesn't catch it, and the entire startup refresh — including its forced `getAssets()` call — runs a second time for the same accounts.

This adds a synchronous `#startInFlight` flag, set before the async work starts and cleared once it settles, so a second trigger during that window is a no-op instead of re-running the whole refresh.

## References

https://consensyssoftware.atlassian.net/browse/ASSETS-3955

## Checklist

---

> [!NOTE]
> **Low Risk**
> Narrow lifecycle guard around existing startup path; behavior unchanged once the first refresh completes, with test coverage for the race.
> 
> **Overview**
> Fixes a race where **multiple lifecycle events** (unlock, account tree init, UI open) could each call `#start()` while the first startup `#runStartupRefresh()` was still waiting on a slow forced `getAssets()`. The old guard only checked `#activeSubscriptions`, which stays empty until that fetch finishes, so a second trigger could run the full startup refresh twice.
> 
> **`#startInFlight`** is set synchronously before `#runStartupRefresh()` and cleared in `.finally()`, so overlapping `#start()` calls are ignored until the in-flight startup completes. A regression test holds `getAssets` pending and asserts a second trigger does not add duplicate forced fetches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5c6d5a92670ea4900455bf0e205cb911cdf6cbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/assets-controller/src/AssetsController.test.ts
+++ b/packages/assets-controller/src/AssetsController.test.ts
@@ -3319,6 +3319,68 @@ describe('AssetsController', () => {
         },
       );
     });
+
+    it('does not run the startup refresh twice if a second start trigger fires before the first getAssets resolves', async () => {
+      await withController(
+        { clientControllerState: { isUiOpen: true } },
+        async ({ controller, messenger }) => {
+          let resolveGetAssets: (() => void) | undefined;
+          const getAssetsSpy = jest
+            .spyOn(controller, 'getAssets')
+            .mockImplementation(
+              () =>
+                new Promise((resolve) => {
+                  resolveGetAssets = (): void => resolve({});
+                }),
+            );
+
+          // First trigger: unlock + account tree ready. #start() kicks off
+          // #runStartupRefresh(), whose forced getAssets() call is still
+          // pending (it never resolves until we call resolveGetAssets below).
+          messenger.publish('KeyringController:unlock');
+          (messenger.publish as CallableFunction)(
+            'AccountTreeController:initialized',
+            {},
+          );
+          await flushPromises();
+
+          expect(getAssetsSpy).toHaveBeenCalledTimes(1);
+
+          // Second, independent trigger fires while the first getAssets()
+          // call is still in flight — #activeSubscriptions is still empty at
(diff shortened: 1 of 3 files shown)


sahar-fehri (approved):

packages/assets-controller/src/AssetsController.test.ts
```diff
@@ -3319,6 +3319,68 @@ describe('AssetsController', () => {
...
+          // Second, independent trigger fires while the first getAssets()
+          // call is still in flight — #activeSubscriptions is still empty at
+          // this point, so only the in-flight guard can prevent a duplicate
+          // #runStartupRefresh() (and its own forced getAssets() call).
+          (
+            messenger as unknown as {
+              publish: (topic: string, payload?: unknown) => void;
+            }
+          ).publish('ClientController:stateChanged', { isUiOpen: true });
+          await flushPromises();
```
Prithpal-Sooriya, line 3358: Nit (not a blocker) - this was a very bad practice we initially added to our unit tests. It spams our tests with useless promise flushing.

Instead I would prefer if we used the `waitFor` utility.

Prithpal-Sooriya (approved):

packages/assets-controller/src/AssetsController.test.ts
juanmigdr: Just seen this, will amend in a future pr

PR #10132, opened 2026-09-07
fix: filter Price API v3 requests to supported networks

Prithpal-Sooriya:
## Explanation

The Price API v3 spot-prices endpoint returns HTTP 400 when it is called with CAIP-19 asset IDs from networks it does not support. `PriceDataSource` currently forwards every priceable asset from balance state / middleware without checking that list.

This change fetches the Price API supported-networks list (`fullSupport` + `partialSupport`) and filters asset IDs to those networks before `fetchV3SpotPrices`. If the supported-networks request fails, the filter fails open and sends the original asset IDs so existing behavior is preserved.

## References

* Related to https://github.com/MetaMask/metamask-extension/issues/46051
* Related to https://consensyssoftware.atlassian.net/browse/ASSETS-3951

## Checklist

<div><a href="https://cursor.com/agents/bc-2a63e86e-4f1b-4bbf-bbe6-09d301f75a14?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2a63e86e-4f1b-4bbf-bbe6-09d301f75a14&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[ASSETS-2825]: https://consensyssoftware.atlassian.net/browse/ASSETS-2825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/assets-controller/src/data-sources/PriceDataSource.test.ts
```diff
@@ -1125,6 +1134,10 @@ describe('PriceDataSource', () => {
     const promise1 = controller.assetsMiddleware(context1, next);
     const promise2 = controller.assetsMiddleware(context2, next);
 
+    // Flush microtasks so #getSupportedNetworks resolves and the deduper
+    // issues the (hanging) fetchV3SpotPrices call before we assert.
+    await jest.advanceTimersByTimeAsync(0);
```
Prithpal-Sooriya: @cursor Can we please try using `waitFor` util instead? This is a better practice.

packages/assets-controller/src/data-sources/PriceDataSource.test.ts
```diff
@@ -1315,4 +1329,305 @@ describe('PriceDataSource', () => {
...
+          return setup;
+        },
+        assert: ({ apiClient }): void =>
+          expectApiSentAssets(apiClient, [
+            'eip155:137/slip44:1' as Caip19AssetId,
+          ]),
+      },
+      {
+        name: 'fetch fails open (sends all assets) when supported networks fetch fails',
+        arrange: (): FilterTestSetup => {
```
Prithpal-Sooriya: @cursor See if you can DRY up the arrange statements. Shared util or smthn

packages/assets-controller/src/data-sources/PriceDataSource.ts
```diff
@@ -466,6 +478,56 @@ export class PriceDataSource {
     }
   }
 
+  /**
+   * Gets the supported networks from the Price API.
+   * Caching is handled by ApiPlatformClient.
+   *
+   * @returns Set of supported chain IDs in CAIP-2 format. Fails open (empty
+   * set) on error so callers can decide whether to proceed.
+   */
```
Prithpal-Sooriya, line 484: @cursor keep jsdoc comment short

packages/assets-controller/src/data-sources/PriceDataSource.ts
```diff
@@ -466,6 +478,56 @@ export class PriceDataSource {
...
+
+  /**
+   * Filters asset IDs to only include those from supported networks.
+   *
+   * @param assetIds - Array of CAIP-19 asset IDs.
+   * @param supportedNetworks - Set of supported CAIP-2 chain IDs.
+   * @returns Array of asset IDs from supported networks. When
+   * `supportedNetworks` is empty (fetch failed), returns all asset IDs
+   * (fail open) to preserve existing behaviour.
+   */
```
Prithpal-Sooriya, line 507: @cursor keep jsdoc short

packages/assets-controller/src/data-sources/PriceDataSource.ts
```diff
@@ -491,7 +553,18 @@ export class PriceDataSource {
...
 
     // Filter out non-priceable assets (e.g., Tron bandwidth/energy resources)
-    const assetIds = rawAssetIds.filter(isPriceableAsset);
+    const priceableAssetIds = rawAssetIds.filter(isPriceableAsset);
+
+    if (priceableAssetIds.length === 0) {
+      return response;
+    }
+
+    // Filter to only assets from networks the Price API supports
```
Prithpal-Sooriya: @cursor remove unnecessary comments!

Prithpal-Sooriya:
@cursor /babysit

packages/assets-controller/src/data-sources/PriceDataSource.test.ts
```diff
@@ -1315,4 +1328,120 @@ describe('PriceDataSource', () => {
...
+      async ({ assets, expectAPI }) => {
+        expect.hasAssertions();
+        const setup = arrangeMiddleware({
+          detectedAssets: assets as Caip19AssetId[],
+          supportedNetworks: [SUPPORTED_SOLANA_CHAIN],
+        });
+
+        await setup.act().catch(() => {
+          // Do nothing
+        });
```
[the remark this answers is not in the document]
Prithpal-Sooriya: Neh this is fine... unless you think we should capture detectAssets that are malformed?

packages/assets-controller/src/data-sources/PriceDataSource.ts
```diff
@@ -491,7 +548,17 @@ export class PriceDataSource {
...
 
     // Filter out non-priceable assets (e.g., Tron bandwidth/energy resources)
-    const assetIds = rawAssetIds.filter(isPriceableAsset);
+    const priceableAssetIds = rawAssetIds.filter(isPriceableAsset);
+
+    if (priceableAssetIds.length === 0) {
+      return response;
+    }
+
+    const supportedNetworks = await this.#getSupportedNetworks();
```
Kriys94, line 557: What about getting the supported Networks at start and not at every fetch?. Even if that is behind TanStack cache

juanmigdr (dismissed):

Kriys94:
Even if that helps, but the root cause of it is that the Snaps is sending non-selected and non-supported networks right?

packages/assets-controller/src/data-sources/PriceDataSource.ts
Prithpal-Sooriya: Hmm. I think I prefer it during runtime.

1. We need to consider basic functionality, doing at start might mean making this call when BF is off.
2. The supportedNetworks call is fast, so isn't really a bottleneck.
3. I like the idea of leveraging core-backend caches. Easy implementation.

Honestly... I'm tempted to say that core-backend should expose compound queries. Otherwise anyone that wants to use token API / price API / accounts API must run their own supportedNetworks calls.

packages/assets-controller/src/data-sources/PriceDataSource.ts
```diff
@@ -184,6 +191,7 @@ export class PriceDataSource {
         this.#executeBatchFetch(assetIds),
       freshnessTtlMs: options.priceFreshnessTtlMs ?? this.#pollInterval,
     });
+    this.#supportedNetworksInflight = this.#loadSupportedNetworks();
```
Kriys94: That should not be necessary, TanStack is already doing request dedup

Prithpal-Sooriya:
> Even if that helps, but the root cause of it is that the Snaps is sending non-selected and non-supported networks right?

@Kriys94

Correct, snaps is sending an unsupported network to the PriceDataSource.

Likewise I will check if custom networks also call the PriceDataSource 🤔 

EDIT: confirmed this also happen for custom networks too!
https://price.api.cx.metamask.io/v3/spot-prices?assetIds=eip155%3A2818%2Fslip44%3A1&vsCurrency=usd&includeMarketData=true&cacheOnly=false

Screenshot 2026-09-08 at 09 33 41

salimtb (dismissed):

packages/assets-controller/src/data-sources/PriceDataSource.test.ts
```diff
@@ -526,9 +535,7 @@ describe('PriceDataSource', () => {
 
     expect(apiClient.prices.fetchV3SpotPrices).toHaveBeenCalledTimes(1);
 
-    jest.advanceTimersByTime(5000);
-    await Promise.resolve();
-
+    await jest.advanceTimersByTimeAsync(5000);
```
Prithpal-Sooriya, line 538: Existing UTs are terrible... I'll do a better overall revamp as a global test initiative.

packages/assets-controller/src/data-sources/PriceDataSource.test.ts
```diff
@@ -560,8 +567,8 @@ describe('PriceDataSource', () => {
     // Advance one tick at a time, flushing microtasks between each so the
     // async pollFn completes and inflight promises settle before the next tick.
     for (let i = 2; i <= 7; i++) {
-      jest.advanceTimersByTime(10000);
-      await jest.advanceTimersByTimeAsync(0);
```
Prithpal-Sooriya, line 564: Lets make sure we do not do this during tests with fake times.
If we are using fake times use the async `jest.advanceTimersByTimeAsync` rather than both synchronous and asynchronous versions.

packages/assets-controller/src/data-sources/PriceDataSource.ts
Prithpal-Sooriya: fixed.

packages/assets-controller/src/data-sources/PriceDataSource.ts
Kriys94: I'm good with keeping it for each fetch. What is the Stale time then?

Kriys94: > I'm tempted to say that core-backend should expose compound queries.

That could be a good idea

salimtb (approved):

Prithpal-Sooriya:
TY @salimtb!

PR #10142, opened 2026-09-08
feat(profile-metrics-controller): add batched proof-of-ownership signing

hmalik88:
## Explanation

This PR adds batched proof-of-ownership signing to `@metamask/profile-metrics-controller`. Previously, `ProfileMetricsController` requested one proof signature per queued account by calling `ProofOfOwnershipService:sign` for each account. For snap-backed non-EVM accounts, that meant one `SnapController:handleRequest` call per account, causing large proof-signing fan-out during profile metrics sync.

This PR adds `ProofOfOwnershipService:signBatch` and updates `ProfileMetricsController` to call it once per entropy-source batch. The service keeps EVM behavior unchanged by signing each EVM account through `KeyringController:signPersonalMessage`, but groups snap-backed non-EVM accounts by `snapId` and sends one `signProofOfOwnershipBatch` request per snap.

Batch responses preserve input order. Per-item snap errors are handled per account, so successful proofs in the same batch are still submitted. Whole snap batch failures or malformed responses are allowed to throw, so the sync queue is not cleared on a broken batch integration.

## References

Requires wallet snaps to support `signProofOfOwnershipBatch`:
- Solana: https://github.com/MetaMask/internal-snaps/pull/256
- Tron: https://github.com/MetaMask/internal-snaps/pull/265
- Bitcoin: https://github.com/MetaMask/internal-snaps/pull/266
- Stellar: https://github.com/MetaMask/internal-snaps/pull/267

## Breaking Change

`ProfileMetricsController` now calls `ProofOfOwnershipService:signBatch` when attaching ownership proofs during profile metrics sync.

Consumers must update their controller messenger wiring to delegate `ProofOfOwnershipService:signBatch` to `ProfileMetricsController`. Clients must also ensure the involved wallet snaps support the `signProofOfOwnershipBatch` JSON-RPC method before adopting this core version, because this change does not fall back to singleton `signProofOfOwnership` calls.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Breaking messenger and snap contract for proof-of-ownership signing during profile sync; rollout must align with snap batch support, though behavior and failure modes are heavily tested.
> 
> **Overview**
> Profile metrics sync now signs ownership proofs in **one batch per entropy-source queue** instead of calling `ProofOfOwnershipService:sign` for each account. **`ProofOfOwnershipService:signBatch`** is added: EVM accounts still use `KeyringController:signPersonalMessage` per item, while snap-backed accounts are **grouped by snap ID** and sent a single **`signProofOfOwnershipBatch`** RPC per snap.
> 
> **`ProfileMetricsController.#attachProofs`** maps batch results back onto queued accounts—successful items get a `proof`, per-item failures are logged and submitted without a proof. If **`signBatch` rejects** or returns the **wrong number of results**, the batch **stays on the sync queue** (no metrics submit for that group).
> 
> This is a **breaking** change: messengers must delegate **`ProofOfOwnershipService:signBatch`**, and wallet snaps must implement **`signProofOfOwnershipBatch`** (there is **no fallback** to the single-account snap method).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8de74884bdf50f95a5dcdc83195d63d63c5d03cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hmalik88:
@metamaskbot publish-previews

packages/profile-metrics-controller/src/ProfileMetricsController.test.ts
```diff
@@ -1574,6 +1596,73 @@ describe('ProfileMetricsController', () => {
...
+
+                await expect(controller._executePoll()).rejects.toThrow(
+                  'ProofOfOwnershipService:signBatch returned 0 results for 1 requests.',
+                );
+
+                expect(mockSubmitMetrics).not.toHaveBeenCalled();
+                expect(controller.state.syncQueue).toStrictEqual(accounts);
+              },
+            );
+          });
```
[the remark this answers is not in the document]
hmalik88: [e377b45](https://github.com/MetaMask/core/pull/10142/commits/e377b454c10b0ca0fa53bdb5b1891a5d07a4dfae)

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
```diff
@@ -223,6 +320,85 @@ export class ProofOfOwnershipService {
...
+              ),
+            },
+          };
+        } catch (error) {
+          results[request.index] = { error: getErrorMessage(error) };
+        }
+      }),
+      ...[...snapRequestsBySnapId.entries()].map(async ([snapId, requests]) => {
+        await this.#signViaSnapBatch(snapId, requests, results);
+      }),
```
mathieuartu, line 396: What happens if one snap fails at `signProofOfOwnershipBatch` here?

packages/profile-metrics-controller/src/ProfileMetricsController.ts
mathieuartu, line 406: Do we still "submit without a proof" now? I still have to understand better what happens but this feels at risk

packages/profile-metrics-controller/src/ProofOfOwnershipService.ts
hmalik88: The snap returns an error in its response and it's logged in the `ProfileMetricsController` same as singular failures today

packages/profile-metrics-controller/src/ProfileMetricsController.ts
hmalik88: Yeah, current behavior is submitting without a proof

mathieuartu: Something is still giving me a weird feeling about this but my brain couldn't follow the whole train of thought today. I asked Fable to have a look, what do you think? 

> There are two distinct failure paths here and only one degrades gracefully. A per-item { error } inside a successful signProofOfOwnershipBatch response is logged and the account is submitted without a proof, as described. But if the snap request itself rejects (snap doesn't implement signProofOfOwnershipBatch yet, snap crash, timeout), signBatch throws: ProofOfOwnershipService.ts:394-396 lets the rejection escape Promise.all, and the PR's own tests assert this (surfaces snap batch request errors without singleton fallback → rejects.toThrow; keeps the batch in the queue when signBatch rejects → submitMetrics never called, syncQueue unchanged).
The consequence is that the whole mnemonic batch, EVM accounts included, is retained and retried every poll (10 s on clients), each retry calling fetchNonces (burning single-use nonces) and re-signing every EVM account, until the snap is updated. On main this exact case submits without a proof, which is what the untouched docblock at ProfileMetricsController.ts:404-407 still promises. Fix: catch the rejection per snap at L394-396 and write { error } into that snap's result slots, so a stale snap costs its proofs, not the batch.

This feels a bit far fetched though, what do you think?

PR #10143, opened 2026-09-08
fix(notification-services-controller): prevent unsibscribing if keyring is locked

zelkibuilds:
## Explanation

Prevents calls to NSC when keyring is locked from unwillingly unsubscribing the authenticated user from wallet activity notifications.

## References
Fixes https://consensyssoftware.atlassian.net/browse/GE-478

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes when wallet-activity push registration and on-chain trigger setup run; incorrect guards could delay setup, but the change is defensive and avoids silent push unregister during lock/empty keyring.
> 
> **Overview**
> **`createOnChainTriggers`** and **`enablePushNotifications`** now return early when the keyring is locked or yields no accounts, instead of treating that as “no enabled addresses” and unregistering push or writing AUS preferences prematurely.
> 
> That keeps existing device push links and leaves first-time preference seeding for a later run when accounts are readable (fixes unwanted unsubscription while locked or before the keyring is populated).
> 
> Tests call **`controller.init()`** so keyring unlock state is wired like production, and add coverage for locked/empty keyring paths on both methods. The package changelog records the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9509a238a630c7b3e300cbeab22ef59e7e9a0256. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/notification-services-controller/src/NotificationServicesController/NotificationServicesController.ts
+++ b/packages/notification-services-controller/src/NotificationServicesController/NotificationServicesController.ts
@@ -913,7 +913,7 @@ export class NotificationServicesController extends BaseController<
   public async enablePushNotifications(): Promise<void> {
     try {
       const { accounts } = this.#accounts.listAccounts();
-      if (accounts.length === 0) {
+      if (!this.#keyringController.isUnlocked || accounts.length === 0) {
         // no keyring accounts could be caused by calling this function while keyring is locked
         // it does not mean the user does not have any enabled accounts on Trigger API
         return;
@@ -1032,7 +1032,7 @@ export class NotificationServicesController extends BaseController<
       this.#setIsUpdatingMetamaskNotifications(true);
 
       const { accounts } = this.#accounts.listAccounts();
-      if (accounts.length === 0) {
+      if (!this.#keyringController.isUnlocked || accounts.length === 0) {
         // no keyring accounts could be caused by calling this function while keyring is locked
         // it does not mean the user does not have any enabled accounts on Trigger API
         return;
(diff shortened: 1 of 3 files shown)


packages/notification-services-controller/src/NotificationServicesController/NotificationServicesController.ts
```diff
@@ -912,8 +912,14 @@ export class NotificationServicesController extends BaseController<
    */
   public async enablePushNotifications(): Promise<void> {
     try {
-      const { bearerToken } = await this.#getBearerToken();
       const { accounts } = this.#accounts.listAccounts();
+      if (accounts.length === 0) {
+        // no keyring accounts could be caused by calling this function while keyring is locked
+        // it does not mean the user does not have any enabled accounts on Trigger API
+        return;
```
mikesposito: Calling `KeyringController:getState` via messenger would return more precise information on the wallet lock state:

```typescript
const { isUnlocked } = this.messenger.call('KeyringController:getState');
if (!isUnlocked) {
  return;
}
```

oilnam (dismissed):

packages/notification-services-controller/src/NotificationServicesController/NotificationServicesController.ts
zelkibuilds: I'll extend the guard to include this but we still do not want to update subscriptions if we don't get any accounts from Keyring 👍

oilnam (approved):

PR #10148, opened 2026-09-09
Feat/pwd change failure recovery

lwin-kyaw:
## Explanation

Seedless password changes span multiple independently persisted states (remote Seedless/TOPRF, local Seedless vault, local KeyringController vault, the stored Keyring encryption key, and a lifecycle marker). These cannot be committed atomically, so a crash, lost response, or partial local update could previously leave the wallet in an ambiguous state where neither the old nor the new password reliably unlocks, and the client had no way to tell that recovery was needed.

This PR adds a **server-first password-change recovery model** to `SeedlessOnboardingController`:

- A persisted, non-sensitive `passwordChangePhase` field (`SeedlessPasswordChangePhase`: `SEEDLESS_CHANGE_PENDING` → `SEEDLESS_COMMITTED` → `LOCAL_KEYRING_PENDING` → `KEY_SYNC_PENDING`, plus `UNKNOWN`) acts as a **recovery signal**. An unset/`undefined` phase means no change is in progress. The phase is not proof of remote or local state — recovery always re-verifies actual state before acting.
- `changePassword` is now lifecycle-aware: it writes each phase at the irreversible boundaries, **preserves the last known phase on error**, and rejects a second concurrent change with `PasswordChangeInProgress`.
- Two new public methods own the Seedless-side recovery sequencing (the controller owns the Seedless side; the client owns the Keyring side):
  - `resolvePasswordSyncState({ skipCache })` — password-less, called at unlock (render + submit). Replaces the public `checkIsPasswordOutdated` read and returns a `PasswordChangeRecoveryStatus` that tells the client which recovery step to run next.
  - `reconcilePassword({ globalPassword })` — password-consuming. Internally runs password-chain unlock and local vault rewrite, re-encrypts `encryptedKeyringEncryptionKey` under the new wrapping key so `loadKeyringEncryptionKey` keeps working, and advances to `LOCAL_KEYRING_PENDING`. Used both for an interrupted local password change and for a password change made on another device.
- Lifecycle-advance methods: `markPasswordChangeKeySyncPending` after the Keyring encryption key is stored, and `clearPasswordChangePhase` once key synchronization and local persistence are verified. `clearPasswordChangePhase` is the only way back to "no change in progress".
- A lost/ambiguous `changeEncKey` response is classified via `fetchAuthPubKey` comparison into **old** / **new** / **unknown**; ambiguous results stay `UNKNOWN` and keep the wallet locked.

**Breaking changes vs `main`:**

- Remove `checkIsPasswordOutdated` and `SeedlessOnboardingControllerCheckIsPasswordOutdatedAction`. Call `resolvePasswordSyncState({ skipCache })` instead (`true` maps to `PasswordChangeRecoveryStatus.PasswordOutdated`).
- Remove `submitGlobalPassword`, `syncLatestGlobalPassword`, and their messenger actions. Call `reconcilePassword({ globalPassword })` instead.
- `changePassword` now writes lifecycle phases and rejects a concurrent change with `PasswordChangeInProgress`. Clients must not start a second password change while a lifecycle is unfinished, and must drive it to completion with `clearPasswordChangePhase`.

There is **no awaitable durability hook** on the controller for lifecycle writes — the phase is persisted as ordinary debounced controller state, so recovery re-verifies actual state (a stale/missing marker is recoverable via the outdated check + cryptographic Keyring verification). The controller does not call `KeyringController` (`AllowedActions = never`); clients own the Keyring-side steps and wallet locking.

Full design, recovery flow, and a step-by-step client integration guide are in [`docs/0002-password-change-recovery-flow.md`](packages/seedless-onboarding-controller/docs/0002-password-change-recovery-flow.md). The ADR is in [`docs/0001-seedless-password-change-recovery.md`](packages/seedless-onboarding-controller/docs/0001-seedless-password-change-recovery.md).

## Test plan

- [ ] `yarn workspace @metamask/seedless-onboarding-controller run test`
- [ ] Confirm `changePassword` writes `SEEDLESS_CHANGE_PENDING` → `SEEDLESS_COMMITTED` → `LOCAL_KEYRING_PENDING` and rejects a second concurrent change with `PasswordChangeInProgress`
- [ ] Interrupted local change: `resolvePasswordSyncState` → `reconcilePassword` → Keyring old/new branch → `storeKeyringEncryptionKey` / `markPasswordChangeKeySyncPending` → `clearPasswordChangePhase`
- [ ] Another-device outdated password: `resolvePasswordSyncState` returns `password-outdated`, `reconcilePassword` returns `reconcile-keyring`, and `loadKeyringEncryptionKey` still decrypts after vault rewrite
- [ ] Ambiguous remote `changeEncKey` result stays `UNKNOWN` and does not infer success from a rejected Promise
- [ ] Client unlock path no longer calls `checkIsPasswordOutdated`, `submitGlobalPassword`, or `syncLatestGlobalPassword`

## References

- ADR: [`docs/0001-seedless-password-change-recovery.md`](packages/seedless-onboarding-controller/docs/0001-seedless-password-change-recovery.md)
- Recovery flow + client integration guide: [`docs/0002-password-change-recovery-flow.md`](packages/seedless-onboarding-controller/docs/0002-password-change-recovery-flow.md)
- Future Option B migration plan (controller owns the Keyring side too): [`docs/0003-controller-owned-password-change-recovery-plan.md`](packages/seedless-onboarding-controller/docs/0003-controller-owned-password-change-recovery-plan.md)
- Client/consumer PRs to adopt the breaking removals: _(to be linked)_

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes wallet password, vault encryption, and TOPRF recovery paths with multiple breaking public APIs; incorrect recovery sequencing could affect unlock or lock users out.
> 
> **Overview**
> Adds a **recoverable password-change and cross-device password-sync model** to `@metamask/seedless-onboarding-controller` via persisted `seedlessOperationLifecycle` checkpoints (`REMOTE_PASSWORD_PENDING` → `LOCAL_STATE_PENDING` → `LOCAL_PASSWORD_PENDING` → `KEY_SYNC_PENDING`).
> 
> **New client-facing API:** `resolvePasswordSyncState({ skipCache })` returns `PasswordSyncInstruction` at unlock without a password; `reconcilePassword({ globalPassword })` runs Seedless-side vault rewrite/unlock and advances checkpoints; `markPasswordChangeKeySyncPending` and `completePasswordChange` finish the lifecycle. **`changePassword`** now writes checkpoints and rejects overlapping changes with `PasswordChangeInProgress`. **`clearState`** is async, waits for in-flight work, and drops in-memory decrypted vault data.
> 
> **Breaking removals:** `checkIsPasswordOutdated`, `submitGlobalPassword`, and `syncLatestGlobalPassword` (and their messenger actions)—clients should use `resolvePasswordSyncState` / `reconcilePassword` instead. Changelog, flow docs, and generated messenger action types are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a1dd02adead857eb375882207067b192c6dcc06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lwin-kyaw:
@codex review

lwin-kyaw:
@metamaskbot publish-preview

lwin-kyaw:
@codex review

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1484,12 +1498,21 @@ export class SeedlessOnboardingController<
...
+   * advancing the lifecycle phase.
+   *
    * @param keyringEncryptionKey - The keyring encryption key.
    */
   async storeKeyringEncryptionKey(keyringEncryptionKey: string): Promise<void> {
     const { toprfPwEncryptionKey: encKey } =
       await this.#unlockVaultAndGetVaultData();
-    await this.#storeKeyringEncryptionKey(encKey, keyringEncryptionKey);
+    const encryptedKeyringEncryptionKey = this.#encryptKeyringEncryptionKey(
+      keyringEncryptionKey,
```
grvgoel81, line 1497: `storeKeyringEncryptionKey` awaits vault access and writes `encryptedKeyringEncryptionKey` without the controller mutex. Consider this sequence: `reconcilePassword` calls `#syncLatestGlobalPasswordInner` which reads the key under the old `toprfPwEncryptionKey` before the vault rewrite, then calls `#commitPasswordChangeState` which installs the new `toprfPwEncryptionKey`. A concurrent client call to `storeKeyringEncryptionKey` that started after the old-key decrypt but before `#commitPasswordChangeState` would write a ciphertext encrypted under the old wrapping key into a state now using the new one. `loadKeyringEncryptionKey` will then fail to decrypt it.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1504,27 +1527,6 @@ export class SeedlessOnboardingController<
     return await this.#loadKeyringEncryptionKey(encKey);
```
grvgoel81, line 1525: loadKeyringEncryptionKey is used inside #syncLatestGlobalPasswordInner (called from within reconcilePassword which holds the controller lock) but is also public. A client calling loadKeyringEncryptionKey concurrently with a password-change vault rewrite will read a toprfPwEncryptionKey that no longer matches the stored ciphertext.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2238,6 +2357,292 @@ export class SeedlessOnboardingController<
...
+   * Clear the password-change lifecycle.
+   *
+   * Used after a definitive remote failure (server did not commit) or once
+   * Keyring encryption-key synchronization is verified and all required local
+   * writes have succeeded. The controller clears the phase so the next unlock
+   * is normal.
+   *
+   * @returns A promise that resolves once the lifecycle has been cleared.
+   */
+  async clearPasswordChangePhase(): Promise<void> {
```
grvgoel81: clearPasswordChangePhase has no guard on which phase it clears. A delayed client call (e.g. from a timed-out request that was retried in a new session) can clear a newer transaction's phase. Example: client A starts changePassword → SEEDLESS_CHANGE_PENDING. Client crashes, restarts, changePassword runs again → LOCAL_KEYRING_PENDING. Delayed clear from session A acquires the lock and wipes LOCAL_KEYRING_PENDING, silently skipping Keyring reconciliation.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2238,6 +2357,292 @@ export class SeedlessOnboardingController<
...
+  /**
+   * Mark the password-change lifecycle as `KEY_SYNC_PENDING`.
+   *
+   * Called by the client coordinator before it synchronizes the current
+   * Keyring encryption key to Seedless. The controller only records the
+   * boundary; it does not perform or verify synchronization.
+   *
+   * @returns A promise that resolves once the phase has been persisted.
+   */
+  async markPasswordChangeKeySyncPending(): Promise<void> {
```
grvgoel81, line 2439: markPasswordChangeKeySyncPending only no-ops when already KEY_SYNC_PENDING, so it will happily advance from undefined (no change in progress) to KEY_SYNC_PENDING. A client bug or replay can corrupt clean state. The guard should assert that the current phase is LOCAL_KEYRING_PENDING before allowing this transition.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2238,6 +2357,292 @@ export class SeedlessOnboardingController<
...
+   * Return the recovery status for phases that require no Seedless-side
+   * mutation. Shared by `resolvePasswordSyncState` (read) and
+   * `reconcilePassword` (apply) so both route the terminal phases
+   * identically.
+   *
+   * @param phase - The persisted password-change phase.
+   * @returns The status for the phase. An unrecognized persisted value is
+   * treated as no change in progress and returns `InSync`.
+   */
+  #statusForTerminalPhase(
```
grvgoel81: The default branch silently maps any unrecognized SeedlessPasswordChangePhase value to InSync, which means a new phase added in future (or a persisted value from a newer app version running on an older code version) will be interpreted as "nothing to recover." This should return PasswordSyncStatus.Unknown (fail-safe) or throw, rather than silently skip recovery. An unrecognized phase is not evidence that the wallet is clean.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Addressed here, https://github.com/MetaMask/core/pull/10148/commits/43532054c17586faf380e8dc9b83945709ae235d

Apply the controller lock to `storeKeyringEncryptionKey` and `loadKeyringEncryptionKey`

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: `loadKeyringEncryptionKey` is now protected with controller lock.
https://github.com/MetaMask/core/commit/43532054c17586faf380e8dc9b83945709ae235d

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: `clearPasswordChangePhase` is only supposed to be called when the password change/sync or unlock is successful. So, if I understand your comment correctly, this isn't possible with the correct client implementation. 

For example,

1. Session A completes the password sync
2. Before session A clear the phase, session B starts the request concurrently.
3. session B request will be rejected as the password change phase (`KEY_SYNC_PENDING`) is still in the state.

lwin-kyaw: However, this is possible with the incorrect implementation in the client side. 
That's why I'm planning to gather all of these moving pieces and put them all in the controller side. 

That's the next step we will do. I've documented [here](https://github.com/MetaMask/core/pull/10148/changes#diff-6f4bb185e40b240f9dad3d4539eec5ec16df2d660ffb69aa5c8830c713e1d1b2), feel free to check it too. 🙏

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Yes, client is responsible to call these based on the state machine checkpoint. 
We will migrate these to inner controller methods so that clients do not need to call these manually, in the later PRs.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: This is removed [here](https://github.com/MetaMask/core/pull/10148/commits/2861dd96e4f74ba8c8a8c00f04772cbb9f3d0007)

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -1706,6 +1715,15 @@ export class SeedlessOnboardingController<
         keyShareIndex: globalKeyIndex,
       } = await this.#recoverEncKey(oldPassword));
     }
+
+    // Persist the lifecycle before the first remote mutation so a later
+    // crash or lost response leaves a recovery signal. The password change
+    // is never retried; recovery reconciles local state via the existing
+    // password-sync flow.
+    this.#writeSeedlessOperationLifecycle(
```
grvgoel81, line 1732: #writeSeedlessOperationLifecycle(undefined, ...) clears the lifecycle before #runPasswordSyncFlow runs. If #submitGlobalPassword succeeds but #syncLatestGlobalPasswordInner throws, the device restarts with no checkpoint and no recovery signal, the next resolvePasswordSyncState must blindly re-check the remote to determine state.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
grvgoel81: clearState() still bypasses #withControllerLock. A concurrent logout calling this while changePassword or reconcilePassword is mid-flight wipes the lifecycle and vault while the in-flight operation still holds decrypted vault data.

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
```diff
@@ -2238,6 +2385,294 @@ export class SeedlessOnboardingController<
...
+
+      if (
+        !checkpoint ||
+        checkpoint === SeedlessOnboardingCheckpoint.RemotePasswordPending
+      ) {
+        const shouldSkipCache =
+          checkpoint === SeedlessOnboardingCheckpoint.RemotePasswordPending
+            ? true
+            : options?.skipCache;
+        const outdated = await this.#checkIsPasswordOutdated({
```
grvgoel81, line 2519: #checkIsPasswordOutdated → toprfClient.fetchAuthPubKey is a network call. While this is waiting, every other controller operation (including refreshAuthTokens calling #reAuthenticate + #updateVault) is serialized behind it. If the TOPRF node is slow, this creates a long mutex-hold that starves token refreshes, meaning by the time the lock is released and a pending refresh runs, the token used to start the flow is already expired. Consider using the token-refresh retry wrapper (#executeWithTokenRefresh) inside the lock for this call,

lwin-kyaw: `#checkIsPasswordOutdated` is already wrapped within the `#executeWithTokenRefresh` refresh.
See here, https://github.com/MetaMask/core/blob/014997e5fd787fe06fb331888300081e27ae625e/packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts#L1316

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: Good find. Addressed here, https://github.com/MetaMask/core/pull/10148/commits/0a1dd02adead857eb375882207067b192c6dcc06

packages/seedless-onboarding-controller/src/SeedlessOnboardingController.ts
lwin-kyaw: If it throws, we just fall back to normal password-sync flow right?

ieow (commented):
nits improvement:
Do we need state LOCAL_STATE_PENDING ?
if local state ( seedless controller is not persisted ) the check with server will show outated pubkey
having the state can may be reduce 1 api call but will it complicate the states?

PR #10156, opened 2026-09-09
Fix partial build step in CI to pass correct args to tsc

mcmire:
## Explanation

The step in CI which only builds changed packages is not working because `tsc` is called with both `--build` and `--project`. Supplying both seems to have allowed prior to TypeScript 7, but this no longer works. The fix is to remove `--project`, as `--build` fills the same role.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Single-line CI workflow change with no runtime or application code impact.
> 
> **Overview**
> Fixes the **partial package build** step in `lint-build-test.yml` so changed-package CI builds work again under current TypeScript.
> 
> When a merge base is available and the change set is not a full-repo build, the workflow generates a temporary solution-style tsconfig and runs `yarn tsc --build` against it. The command previously used **both** `--build` and `--project`, which TypeScript no longer accepts (it worked before TS 7). The fix drops `--project` and passes the generated config path as the positional argument to `--build`, matching how full builds invoke `tsc` elsewhere in the repo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb56c93bbf7d3d8d95758161bc8ad353c0fb9607. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -220,7 +220,7 @@ jobs:
               echo "Building changed packages:"
               jq -r '"- " + (.references[].path | ltrimstr("./") | rtrimstr("/tsconfig.build.json"))' "$TSCONFIG"
               echo ""
-              yarn tsc --build --project "$TSCONFIG" --verbose
+              yarn tsc --build "$TSCONFIG" --verbose
             else
               echo "No packages to build."
             fi


cryptodev-2s (approved):

Mrtenz:
Oops, should've tested it.

Either way we can probably just remove the partial build logic now, since a full build only takes ~12 seconds in CI.

PR #9324, opened 2026-06-30
Update BaseDataService to accommodate mutations, not just queries

mcmire:
## Explanation

Currently, `BaseDataService` has a `fetchQuery` method which is best used for making read-only requests ("queries" in TanStack Query parlance), but does not work as well for requests that change state on the server side ("mutations"). For instance, it makes sense for queries to be cached and retried, but not so much for mutations.

This commit adds a separate method, `executeMutation`, which accommodates mutations better. Besides the differences mentioned above, it also uses the mutation cache instead of the query cache.

## References

Closes https://consensyssoftware.atlassian.net/browse/WPC-1118.

## Manual testing

I've created https://github.com/MetaMask/core/pull/10176, which converts `AuthenticatedUserStorageService.setAssetsWatchlist` to use `executeMutation` under the hood, and https://github.com/MetaMask/metamask-mobile/pull/36083, which loads those changes into the mobile app. You can check out the mobile branch and run through the manual testing steps there to confirm that `executeMutation` still exhibits the same basic behavior as `fetchQuery`.

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> Touches shared cache/event contracts and cross-process mutation sync; incorrect handling could desync UI state or mishandle `:cacheUpdated` listeners, though no-retry mutations reduce duplicate side-effect risk.
> 
> **Overview**
> Adds **mutation support** to the MetaMask data-service stack so write operations no longer go through query-oriented `fetchQuery`.
> 
[...]

PR #10176, opened 2026-09-10
[DON'T MERGE] Apply executeMutation pattern across some data services for testing

mcmire:
This is a PR which allows us to smoke-test the changes being introduced in #9324 — specifically the introduction of `executeMutation` in BaseDataService — within clients.

It will be removed once this smoke test is complete.


mcmire:
@metamaskbot publish-previews

PR #10179, opened 2026-09-10
chore: remove `@metamask/utils` files owned by the monorepo root

cryptodev-2s:
## Explanation

Third step of migrating `@metamask/utils` (PR#8 of Phase B in the [migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md)).

The package arrived from its own repo carrying a full set of repo level tooling. This monorepo owns all of it centrally, so those files are removed: `.github/`, `.yarn/`, `.yarnrc.yml`, `yarn.lock`, `scripts/`, `eslint.config.mjs`, `knip.config.mts`, `constraints.pro`, and the `.editorconfig`, `.gitattributes`, `.gitignore`, `.nvmrc`, `.oxfmtrc.json` and `.prettierrc.cjs` dotfiles.

I checked each one before removing it: the root has an equivalent for every file except `constraints.pro`, which has no counterpart because this repo uses `yarn.config.cjs` instead. No package under `packages/` carries its own copy of any of them.

What remains is the guide's keep list: `src/`, `CHANGELOG.md`, `LICENSE`, `package.json`, `README.md`, `jest.config.cjs`, `tsconfig*.json` and `typedoc.json`. There is no `tests/` directory because this package keeps its tests alongside the sources.

Deletions only, no code changes. This also clears the zizmor and Socket alerts on #10175, which came from the migrated workflow files and lockfile.

## References

- Follows #10175 (git history migration) and #10178 (changelog reset)

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Deletion-only cleanup of duplicate config and CI; no application logic changes, though contributors must use root workflows until follow-up wiring is confirmed.
> 
> **Overview**
[...]

PR #10180, opened 2026-09-10
chore: extend `@metamask/utils` configs from the monorepo root

cryptodev-2s:
## Explanation

Fourth step of migrating `@metamask/utils` (PR#9 of Phase B in the [migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md)).

The package's config files were standalone. They now extend the monorepo's, matching the other packages:

- `tsconfig.json` extends `../../tsconfig.packages.json`
- `tsconfig.build.json` extends `../../tsconfig.packages.build.json`
- `tsconfig.lint.json` extends `./tsconfig.json` and `../../tsconfig.packages.lint.json`, emitting into `./.tsc-lint-cache`
- `jest.config.cjs` is a `deepmerge` of `../../jest.config.packages.cjs`, so `deepmerge` is added as a devDependency
- `typedoc.json` matches the other packages, outputting to `api-docs`

Two things are deliberately preserved rather than inherited away:

**The stricter compiler flags.** This package sets `exactOptionalPropertyTypes`, `forceConsistentCasingInFileNames`, `noErrorTruncation` and `noUncheckedIndexedAccess`, none of which are in `tsconfig.base.json`. The sources were written against them, and `noUncheckedIndexedAccess` is load bearing: several functions branch on an index access being `undefined`. `eth-json-rpc-provider` sets the same four, so this follows existing practice.

**The 100% coverage thresholds**, which the package has met since before the migration.

No `references` are needed: none of this package's dependencies live in this repo.

Two details worth flagging for review. `exclude` replaces rather than extends, so `tsconfig.build.json` repeats the root's entries alongside its own. And `tsconfig.lint.json` excludes `*.test-d.ts`, because those belong to `tsd` and pulling them in drags in `tsd`'s own types, which reference `lib="esnext"` and widen built in types like `ReadonlySet` beyond this package's target.

## References

- Follows #10179

## Checklist

---

> [!NOTE]
> **Low Risk**
> Config and build-tooling only; no runtime or library API changes, with intentional preservation of strict TS flags and coverage gates.
> 
> **Overview**
> Aligns **`@metamask/utils`** tooling with the rest of the monorepo by inheriting shared configs instead of maintaining full copies locally.
> 
> **TypeScript:** `tsconfig.json`, `tsconfig.build.json`, and `tsconfig.lint.json` now extend `../../tsconfig.packages*.json`, with package-specific **`include`/`exclude`** (shared `types`, build excludes for fixtures/mocks/`test-d`) and **lint cache paths** unchanged in spirit. The package still overrides **stricter compiler flags** (`exactOptionalPropertyTypes`, `noUncheckedIndexedAccess`, etc.) that the sources depend on.
> 
> **Jest:** `jest.config.cjs` shrinks to a **`deepmerge`** of `../../jest.config.packages.cjs`, adding **`deepmerge`** as a devDependency. Overrides keep **100% global coverage thresholds** and extra **`coveragePathIgnorePatterns`** for fixtures and `*.test-d.ts`.
> 
> **Typedoc:** output directory moves from `docs` to **`api-docs`**, with an explicit `./tsconfig.build.json` path—matching other packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cffab65b12771c33ad8c971d44d50c2ae669724. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


merged-packages/utils/tsconfig.json
```diff
@@ -1,17 +1,16 @@
 {
+  "extends": "../../tsconfig.packages.json",
   "compilerOptions": {
-    "esModuleInterop": true,
+    /**
+     * These are stricter than the monorepo defaults and the sources were
+     * written against them, so they are set here rather than inherited away.
+     * `noUncheckedIndexedAccess` in particular is load bearing: several
+     * functions branch on an index access being `undefined`.
+     */
```
mcmire: Maybe we don't need this comment? We aren't taking anything away that isn't already there.

``` suggestion
```

merged-packages/utils/jest.config.cjs
```diff
@@ -53,179 +28,6 @@ module.exports = {
...
-  // testRegex: [],
-
-  // This option allows the use of a custom results processor
-  // testResultsProcessor: undefined,
-
-  // This option allows use of a custom test runner
-  // testRunner: "jest-circus/runner",
-
   // Reduce the default test timeout from 5s to 2.5s
   testTimeout: 2500,
```
mcmire: Maybe we don't need this either? If we really want to, we can reduce the global default to 2.5s. But in the meantime I don't see why we need to make an exception for `utils`. It's rare IMO that we run into timeout issues in tests.

``` suggestion
```

cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/10180/commits/7cffab65b12771c33ad8c971d44d50c2ae669724

merged-packages/utils/tsconfig.json
cryptodev-2s: Removed here https://github.com/MetaMask/core/pull/10180/commits/7cffab65b12771c33ad8c971d44d50c2ae669724

GuillaumeRx (approved):

PR #10182, opened 2026-09-10
chore: update the `@metamask/utils` README for the monorepo

cryptodev-2s:
## Explanation

Final step of Phase B for `@metamask/utils` (PR#11 of the [migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md)).

The README is rewritten for its new status as a non-root package. The opening description is preserved, `Installation` is kept, and a `Contributing` section pointing at the monorepo README is added, matching the 95 packages with `Installation` and 94 with `Contributing`.

Removed: the migration notice banner (no longer accurate now that the package lives here), and the development instructions the monorepo owns centrally, covering setup, testing, linting, documentation, the release process and preview builds.

The `API` section is also dropped. It linked to `metamask.github.io/utils`, which will no longer be updated now that docs are generated from this repo.

## References

- Follows #10181

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Documentation-only README edits with no runtime or API changes.
> 
> **Overview**
> Rewrites the **`@metamask/utils` README** now that the package lives in the **MetaMask core monorepo**, instead of the old standalone repo.
> 
> The **migration warning banner** and long **local dev docs** (setup, test/lint, docs generation, release, preview builds) are removed in favor of a short **Contributing** section that points to the [monorepo README](https://github.com/MetaMask/core#readme). The **API** section linking to `metamask.github.io/utils` is removed because docs will come from this repo. The **package description** and **Installation** (`yarn` / `npm`) sections are unchanged; the title is updated to **`# @metamask/utils`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b405a756e4c405b17f02c13adcab0d44729c8249. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #10181, opened 2026-09-10
chore: align `@metamask/utils` deps and scripts with the monorepo

cryptodev-2s:
## Explanation

Fifth step of migrating `@metamask/utils` (PR#10 of Phase B in the [migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md)).

**Scripts** now match the other packages: `build:all`, `changelog:update`, `changelog:validate`, `lint:tsconfigs`, `since-latest-release`, and the `NODE_OPTIONS=--experimental-vm-modules` test variants. The lint scripts (`lint`, `lint:eslint`, `lint:constraints`, `lint:misc`, `lint:dependencies`, `lint:tsc`) are gone because the root owns linting, and `prepack` is gone with the `scripts/` directory removed in #10179.

`test` keeps running `test:types` after Jest, since this package has `tsd` type tests that nothing else would run.

**Dependencies**: 21 devDependencies drop to 17. Removed are the ones the root already provides and no package declares for itself: the `@lavamoat/*`, `@metamask/eslint-config*` and `eslint*` sets, plus `@types/node`, `@types/semver`, `knip`, `oxfmt`, `prettier` and `typescript-eslint`. Every one was already at a version identical to the root's, so nothing changes on disk.

`jest-it-up` is also removed: it existed only to bump coverage thresholds from the old `test:source` script, which no longer exists.

Kept, matching the convention in all 99 packages: `typescript`, `@typescript/native`, `deepmerge`, `jest`, `ts-jest`, `rimraf`, `typedoc`, `@metamask/auto-changelog`, `@types/jest`, and `tsx` for the new `lint:tsconfigs` script. Kept as genuinely package specific: `jest-when`, `@types/jest-when`, `@types/uuid`, `stdio-mock`, `tsd` and `web3`. I verified each of those is actually imported by the sources.

`packageManager` is dropped because no package declares it (the root does). `engines` stays, as it does in all 99. The `lavamoat` block stays: 8 packages have one, including the same `unrs-resolver` entry.

Runtime dependencies are untouched.

## References

- Follows #10180

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
[...]

PR #10178, opened 2026-09-10
chore: reset the migrated `@metamask/utils` changelog

cryptodev-2s:
## Explanation

Second step of migrating `@metamask/utils`, following [the package migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md) (PR#7 of Phase B).

The migrated package arrived with the full changelog from its old repo, covering 59 releases up to `12.0.0`. Releases from here on are made from this repo, so that history no longer belongs in this file. It is reset to a fresh changelog with a single `Unreleased` entry recording the migration and linking to the original, which stays available in `MetaMask/utils`.

No code changes.

## References

- Follows #10175, which migrated the git history into `merged-packages/utils/`
- Original changelog: https://github.com/MetaMask/utils/blob/main/CHANGELOG.md

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Documentation-only changelog edit with no code or dependency changes.
> 
> **Overview**
> Resets **`merged-packages/utils/CHANGELOG.md`** for post-migration releases in **`MetaMask/core`**, replacing the copied history (59 releases through **`12.0.0`**) and version footer links with a minimal **[Unreleased]** section.
> 
> The new entry records the move from **`MetaMask/utils`** ([#10175](https://github.com/MetaMask/core/pull/10175)) and points readers to the [original changelog](https://github.com/MetaMask/utils/blob/main/CHANGELOG.md) for prior releases. The **`[Unreleased]`** compare link now targets **`MetaMask/core`** instead of the old repo.
> 
> **No runtime or API changes**—changelog maintenance only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e103757c0130ad651d1ebc8264e6765d0f7c226f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #10175, opened 2026-09-10
chore: migrate `@metamask/utils` git history into `merged-packages/`

cryptodev-2s:
## Explanation

First step of migrating `@metamask/utils` into this repo, following [the package migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md).

This brings the source repo's git history into `merged-packages/utils/` using `git filter-repo --to-subdirectory-filter`, so the 229 commits of history are preserved rather than squashed into one. Nothing outside `merged-packages/` is touched, and no config or build wiring is done yet: that follows in the later PRs of Phase B.

Phase A is complete, and `@metamask/utils@12.0.0` is published from the source repo. It aligns with this repo already: ESM only, `engines` of `^22.14.0 || ^24`, and an ES2022 target.

> [!IMPORTANT]
> Per the guide, this PR must be **merged without squashing** to preserve the migrated history, which needs a maintainer to temporarily enable merge commits into `main`. The branch must also **not be rebased**. Please merge promptly once approved, so superfluous merge commits from `main` do not pollute the migrated history.

After this lands, a tag for the last source-repo release needs creating so `action-publish-release` does not try to publish `12.0.0` again:

```shell
git tag -a @metamask/utils@12.0.0 <release-commit-sha-in-core>
git push origin @metamask/utils@12.0.0
```

## References

- Source repo: https://github.com/MetaMask/utils
- Migration process guide: `docs/processes/package-migration-process-guide.md`

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Additive history import under a staging path with no monorepo build or publish wiring; runtime behavior of `@metamask/core` is unchanged until follow-up migration PRs land.
> 
> **Overview**
[...]

PR #10185, opened 2026-09-10
chore: integrate `@metamask/utils` into `packages/`

cryptodev-2s:
## Explanation

Phase C of migrating `@metamask/utils`, following [the package migration process guide](https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md). This moves the package from `merged-packages/` into `packages/` and wires it into the monorepo.

- `merged-packages/utils` → `packages/utils`, registered as a workspace
- Reference paths added to the root `tsconfig.json`, `tsconfig.build.json` and `tsconfig.lint.json`
- The version bump for dependents is split into a follow up (#10192) to keep this diff reviewable. This PR adds a temporary `@metamask/utils` entry to `ALLOWED_INCONSISTENT_DEPENDENCIES` in `yarn.config.cjs` so constraints pass while dependents are still on the published version; #10192 removes it. The two are meant to land together.
- Downstream tsconfig references added via `yarn lint:tsconfigs:fix:all` (174 files). Without them `yarn lint:tsc` reported 536 errors; it is now clean
- `yarn constraints --fix` applied, which split the test scripts into `test:unit` and `test:types`
- CODEOWNERS entry added through `codeowners.ts` and `teams.json`, assigning the package to `@MetaMask/core-platform` as with the other foundational packages
- README package list and dependency graph regenerated with `yarn readme-content:update`

### Type tests moved from `tsd` to `tstyche`

The package's type tests import from source (`./index.js`, `./misc.js`), so they never needed built declarations. `tsd` requires a typings file as a structural precondition regardless, which forced a build purely to satisfy the tool. `tstyche` type checks source directly and is already used by `@metamask/messenger`, so the four `*.test-d.ts` files are now `*.tst.ts`, with all 120 assertions preserved.

That removed three workarounds: a `composite: false` override for `tsd`, a `--typings` flag, and the `*.test-d.ts` exclusion in `tsconfig.lint.json`. That last one existed because `tsd`'s own types pull in `lib="esnext"`, widening built in types like `ReadonlySet` past this package's target. With that gone the type tests are covered by `lint:tsc` rather than excluded from it, and `tsd` leaves the lockfile entirely.

### Root tooling now resolves the package from source

Nine files at the repository root import `@metamask/utils`, which previously arrived prebuilt from npm. As a workspace its `dist` does not exist until something builds it, and the lint jobs do not build, so `constraints`, `lint:teams` and `lint:tsconfigs:all` all failed on a fresh checkout.

Adding a `paths` entry to the root `tsconfig.json` fixes the scripts. `yarn.config.cjs` is the exception, since Yarn loads it as plain CommonJS where `paths` do not apply and a `.ts` file cannot be required. It only used `hasProperty`, whose runtime body is a single expression, so it now calls `Object.hasOwn` directly. That is available on every supported Node version.

All of the following pass with no build artifacts present: `constraints`, `lint:teams`, `lint:tsconfigs:all`, `readme-content:check`, `codeowners:check`, `lint:tsc`, `lint:misc:check`, and the package's own tests.

### Inherited lint findings

86 pre-existing findings in the package, mostly `no-explicit-any` in tests, are recorded in `eslint-suppressions.json` rather than fixed here, matching how this repo tracks that debt. They came from rules the package's own ESLint config had disabled before the migration.

## References

- Follows #10175, #10178, #10179, #10180, #10181 and #10182
- Migration process guide: `docs/processes/package-migration-process-guide.md`

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Wide monorepo integration touches build graphs and resolution for a foundational dependency; temporary dual resolution (workspace v12 vs published v11) must land with the follow-up bump to avoid inconsistent runtime/types.
> 
> **Overview**
> **`@metamask/utils` is now a workspace package at `packages/utils` (v12.0.0)** instead of living under `merged-packages/` or npm-only. Repo metadata, CODEOWNERS/`teams.json`, README list/graph, and root `tsconfig` project references are updated so the package builds and lints with the rest of core.
> 
> **Monorepo wiring:** Jest and root/script `tsconfig` paths resolve `@metamask/utils` (and `/node`) to **TypeScript source** so tests and lint work without a prior build. Package `tsconfig` references to `../utils` were added across dependents. **`tsconfig.packages.json` still maps `@metamask/utils` to published v11 in `node_modules`** until a follow-up bumps dependents; `yarn.config.cjs` temporarily allows that version skew and drops the `@metamask/utils` import in favor of **`Object.hasOwn`**.
> 
> **Package tooling:** Type tests move from **`tsd` (`*.test-d.ts`) to `tstyche` (`*.tst.ts`)**; build excludes `*.tst.ts`. Unused **`@metamask/utils` dependencies are removed** from a few packages (e.g. messenger, ai-controllers). Pre-existing ESLint issues in `packages/utils` are **recorded in `eslint-suppressions.json`**. Stale **`merged-packages/utils`** type-test and lint config files are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd045569381c45f128305b0c1cef5190e067a4c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/accounts-controller/package.json
```diff
@@ -61,7 +61,7 @@
     "@metamask/messenger": "^3.0.0",
     "@metamask/network-controller": "^37.0.0",
     "@metamask/superstruct": "^3.4.1",
-    "@metamask/utils": "^11.12.0",
+    "@metamask/utils": "^12.0.0",
```
mcmire: Hmm, this bumps to 12.0.0 at the same time. Should we do that in a separate PR?

packages/utils/package.json
```diff
@@ -43,10 +47,13 @@
     "lint:tsconfigs": "tsx ../../scripts/lint-tsconfigs/lint-tsconfigs.mts",
     "lint:tsconfigs:fix": "tsx ../../scripts/lint-tsconfigs/lint-tsconfigs.mts --fix",
     "since-latest-release": "../../scripts/since-latest-release.sh",
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest --reporters=jest-silent-reporter && yarn test:types",
-    "test:clean": "NODE_OPTIONS=--experimental-vm-modules jest --clearCache",
-    "test:types": "tsd --files 'src/*.test-d.ts'",
-    "test:verbose": "NODE_OPTIONS=--experimental-vm-modules jest --verbose",
+    "test": "yarn test:unit && yarn test:types",
+    "test:clean": "yarn test:unit:clean && yarn test:types",
+    "test:types": "tstyche",
```
mcmire, line 52: This bumps to `tstyche` at the same time. Should we do that in a separate PR?

eslint-suppressions.json
```diff
@@ -2007,5 +2007,95 @@
     "@typescript-eslint/naming-convention": {
       "count": 1
     }
+  },
```
mcmire, line 2010: Oof 😬 Maybe we should have another PR that fixes these type errors? Ideally we shouldn't introduce the package with a bunch of suppressions.

tsconfig.json
```diff
@@ -5,7 +5,11 @@
    */
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
-    "noEmit": true
+    "noEmit": true,
+    "paths": {
```
mcmire, line 9: Hmm, why is this necessary? TypeScript should be able to find the exports we've defined in `utils`'s `package.json`.

packages/accounts-controller/package.json
cryptodev-2s: Yes because the local version is 12 now

packages/utils/package.json
cryptodev-2s: Normally yes, but tsd can't work here. The package test job runs `yarn workspace <pkg> run test` with no build step, and `tsd` requires a built typings file even though these tests import from source, so CI would fail.

It worked in the utils repo because `test:types` ran inside the build job right after `yarn build`, while the test job only ran `test:source`. Here both halves run in one job that never builds, and constraints mandate `test` = `test:unit && test:types`, so splitting them across jobs isn't an option either.

Doing the swap separately would also mean landing three tsd workarounds and reverting them right after: `composite: false`, `--typings ./dist/index.d.ts`, and excluding `*.test-d.ts` from `tsconfig.lint.json`.

I should have done this in the old `utils` repo, but I have missed this one

eslint-suppressions.json
cryptodev-2s: Yes definitely I was planning to follow up on these, without growing the scope here

tsconfig.json
cryptodev-2s: Because they are used by scripts, and otherwise `lint:teams` was failing, it looks like `tsconfig.scripts.json` is there but not used after all.
[This configuration is intended for the scripts/ directory... It's currently not actually used for that purpose, but it will be in a future PR.](https://github.com/MetaMask/core/blob/main/tsconfig.scripts.json#L3)

cryptodev-2s:
@metamaskbot update-changelogs

packages/accounts-controller/package.json
cryptodev-2s: I am investigating the split... as it looks like the scope did grow up

cryptodev-2s: @mcmire this has now been splited into two PRs stacked

packages/utils/package.json
mcmire: Yeah in the future I think we should basically do everything we can in the source repo to match `core` before moving it over to `core`. I believe that's the intended goal of the package migration guide.

eslint-suppressions.json
mcmire: Again, it might have been better to address these in the `utils` repo before moving over to `core`.

tsconfig.json
mcmire: Ah, strange, okay. It's okay for now. I wonder if a good way to fix this would be to make `create-package` a private package, then move all scripts that aren't `create-package` into a second private package. Then we "shouldn't" need a separate `tsconfig.scripts.json`. Maybe we can worry about that later though.

packages/account-tree-controller/tsconfig.build.json
```diff
@@ -22,6 +22,9 @@
     },
     {
       "path": "../profile-sync-controller/tsconfig.build.json"
+    },
+    {
+      "path": "../utils/tsconfig.build.json"
```
mcmire, line 27: We don't need to modify the `tsconfig.json` and `tsconfig.build.json` across packages until we upgrade packages to 12.0.0, right?

packages/build-utils/tsconfig.lint.json
```diff
@@ -4,5 +4,9 @@
     "outDir": "./.tsc-lint-cache",
     "tsBuildInfoFile": "./.tsc-lint-cache/tsconfig.tsbuildinfo"
   },
-  "references": []
+  "references": [
+    {
+      "path": "../utils/tsconfig.lint.json"
```
mcmire, line 9: Same goes for `tsconfig.lint.json` files — we shouldn't need to modify these yet right?

packages/account-tree-controller/tsconfig.build.json
mcmire: Oh, hmm, it seems that `lint:tsconfigs` doesn't realize that we aren't depending on the workspace version of `utils`. Okay. This is alright I guess, there's no harm in including these.

packages/build-utils/tsconfig.lint.json
mcmire: Never mind, I guess this is necessary :)

mcmire (approved):
LGTM.

packages/account-tree-controller/tsconfig.build.json
cryptodev-2s: The docs are explicit on doing this in one PR https://github.com/MetaMask/core/blob/main/docs/processes/package-migration-process-guide.md#1-move-the-migration-target-directory-from-merged-packages-into-packages
The split though was done here because they would both be merged at once but the review will be a bit easier

mcmire: Oh I see, I forgot about that, thank you.

PR #10192, opened 2026-09-11
chore: bump dependents to `@metamask/utils@^12.0.0`

cryptodev-2s:
## Explanation

Follow up to #10185, which moved `@metamask/utils` into `packages/`. This bumps its dependents to the local version.

- `@metamask/utils` bumped from `^11.12.0` to `^12.0.0` in 79 packages and the root
- Removes the temporary `@metamask/utils` entry from `ALLOWED_INCONSISTENT_DEPENDENCIES` in `yarn.config.cjs`, which #10185 added so that constraints would pass while the dependents were still on the published version
- Changelog entries recorded for the bump, plus the lockfile dedupe and README dependency graph that follow from it

These two PRs are meant to land together, so the intermediate state where dependents still resolve the published copy never reaches `main`.

Splitting it this way keeps #10185 reviewable: the move, the config rewiring and the type test change are separate from 79 mechanical version bumps.

## References

- Follows #10185

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> A major-version bump of a shared utility touches most packages; runtime risk is low if v12 is compatible, but consumers and lingering v11 transitive deps (keyring) can cause type or duplicate-package issues until fully aligned.
> 
> **Overview**
> This PR aligns the monorepo with the in-repo **`@metamask/utils` v12** after it was moved into `packages/` (follow-up to #10185). **79 packages** plus the root bump the dependency from **`^11.12.0` to `^12.0.0`**, with matching **Unreleased changelog** entries and an updated **README dependency graph**.
> 
> Tooling is cleaned up so builds use the workspace package instead of the published copy: **`tsconfig.packages.json`** now maps `@metamask/utils` to local source (the temporary `node_modules` override is removed), and **`yarn.config.cjs`** drops the **`ALLOWED_INCONSISTENT_DEPENDENCIES`** exception for `@metamask/utils`.
> 
[...]

PR #10216, opened 2026-09-11
fix: keep the keyring cast valid in preview builds

cryptodev-2s:
## Explanation

Preview builds have been failing since #10192 with one TS error:

```
packages/keyring-controller/src/KeyringController.ts(644,7): error TS2578: Unused '@ts-expect-error' directive.
```

`Keyring` in that builder comes from `@metamask/keyring-utils`, which still depends on `@metamask/utils@^11`, while this package depends on the workspace copy at v12. Normally that leaves two identities for the same type, so the cast does not overlap and the directive is needed.

Preview builds resolve it differently. The workflow writes a `portal:` resolution for every workspace into the root manifest, and since that applies globally it also replaces the nested v11 copy `@metamask/keyring-utils` brings along. Both copies collapse onto the workspace one, the cast overlaps again, and the directive is flagged as unused.

So it is required in a normal build and wrong in a preview build. Casting through `unknown` holds either way. The comment notes it can go once the keyring packages depend on `@metamask/utils@12`.

## References

Failing run: https://github.com/MetaMask/core/actions/runs/34633477377/job/103375769956

Related to #10192

## Checklist

### DIFF
--- a/packages/keyring-controller/src/KeyringController.ts
+++ b/packages/keyring-controller/src/KeyringController.ts
@@ -641,12 +641,16 @@ const hdKeyringV2Builder: KeyringV2Builder = Object.assign(
 const simpleKeyringV2Builder: KeyringV2Builder = Object.assign(
   (keyring: Keyring): KeyringV2 =>
     new SimpleKeyringV2({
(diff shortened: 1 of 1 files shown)


cryptodev-2s:
@metamaskbot publish-preview

PR #10224, opened 2026-09-14
feat(transaction-pay-controller): atomic quoting for subsidized max Money Account deposits

OGPoyraz:
## Explanation

Enable atomic `EXACT_OUTPUT` Relay execution for subsidized max deposits, with vault/transaction calls embedded instead of submitted as a separate second leg.

### Atomic max quoting

- When atomic max is enabled, `atomic: true` (or omitted) requests atomic `EXACT_OUTPUT` first. The destination amount comes from the source-token budget rescaled to destination decimals, rounding down, for fixed-spread subsidized stablecoin routes that exchange 1:1 without additional fees. It does not reuse the pre-Max required amount or require a discovery quote.
- If that atomic quote is subsidized, return it directly. Otherwise, request a non-atomic max `EXACT_INPUT` quote for separate second-leg execution.
- With `atomic: false`, start with non-atomic `EXACT_INPUT`. If subsidized, use its output amount to rebuild calldata and required assets on a clone before requesting atomic `EXACT_OUTPUT`.
- Preserve `isMaxAmount` throughout and leave the original transaction unchanged while preparing atomic calls. Non-max behavior is unchanged.

### Feature flag and error handling

Eligibility is controlled by `confirmations_pay_extended.payStrategies.relay.atomicMaxEnabled`, replacing the earlier `atomicMaxPromotionEnabled` name. Its shape remains `{ default?: boolean; transactionTypes?: Partial<Record<TransactionType, boolean>> }`. It defaults to disabled, and configured overrides match top-level or nested transaction types. There is no hardcoded Money Account eligibility fallback.

When disabled, existing max transaction-embedding restrictions still apply; clients requesting a non-atomic max deposit should set `atomic: false`.

Atomic promotion failures retain the `Atomic promotion failed` prefix through standard quote-error handling and pay-strategy fallback. The special terminal-error branch in the quote orchestrator is removed. Atomic preparation/fetch errors are not silently converted into non-atomic retries.

Max routing lives in `relay-max.ts`. Legacy gas-station handling remains intact; removing it is deferred to a separate PR. No public API signatures changed.

### Client coordination

Companion draft: [metamask-mobile#36412](https://github.com/MetaMask/metamask-mobile/pull/36412). Clients can use the `atomic` hint to reflect expected route subsidy. Correct atomic hints now avoid the initial discovery quote. Client package integration and live route validation remain pending.

## Validation Status

- Package test suite: **54 suites, 1,479 tests passed**.
- Package build, including workspace dependencies: passed.
- Changed-file ESLint, formatting, and package changelog validation: passed.
- Built-module smoke checks: passed for decimal rescaling, one-quote subsidized execution, unsubsidized fallback, transaction immutability, and rejection of sub-unit output.
- Regression coverage includes explicit/default atomic hints, stale pre-Max amounts, both decimal conversion directions, disabled flags, non-Money Account eligibility, strategy fallback, and max ERC-20 transfer recipients.
- Test-inclusive TypeScript checking still reports pre-existing errors. Comparison against the previous PR head found no newly introduced diagnostics in changed files.
- Live client/Relay and simulator verification have not been performed for this revision. Existing screenshots and recording below are retained as earlier flow evidence.

## Screenshot & recordings

Exact Input for Max ARB token:
Exact Input for Max ARB token

Exact output for Max USDC token:
Exact Output for Max USDC token

Recording:

https://github.com/user-attachments/assets/6403ed48-67d2-4900-8335-235e1fea099a

## References

* Background: [metamask-mobile#33489](https://github.com/MetaMask/metamask-mobile/pull/33489), [core#9497](https://github.com/MetaMask/core/pull/9497)
* Companion client PR: [metamask-mobile#36412](https://github.com/MetaMask/metamask-mobile/pull/36412) (draft)
* https://consensyssoftware.atlassian.net/browse/CONF-1949

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes max-deposit quoting, amount embedding, validation, and submit routing for subsidized Relay routes; mistakes could mis-size deposits or leave users without quotes, but behavior is feature-flag gated and heavily tested.
> 
> **Overview**
> Adds **subsidized max Relay deposits** that can run as atomic **`EXACT_OUTPUT`** quotes with embedded transaction calls, behind **`payStrategies.relay.atomicMaxEnabled`** (off by default, with per-`TransactionType` overrides).
> 
> Max quoting now routes through **`getRelayMaxQuote`**: when the flag is on, the client **`atomic`** hint drives whether to request atomic max first (source budget rescaled to destination decimals, rounded down) or start with non-atomic **`EXACT_INPUT`** and **promote** to atomic when Relay returns a subsidized quote. Promotion rebuilds calldata via **`getAmountData`** on a cloned transaction without mutating the original. Unsubsidized atomic attempts fall back to non-atomic max; promotion/validation failures surface as **`QuoteError`** with an **`Atomic promotion failed`** prefix so orchestration can fall back to other pay strategies and clear stale quotes on refresh.
> 
> Relay quoting loosens the “max + embedded tx” guard when atomic max is enabled, validates promoted atomic max quotes in a separate batch (wrapping validation errors as promotion failures), and **submit** skips the separate Money Account vault step when the promoted quote already inlines the deposit calls in Relay execute.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fd8ff1b5227286567b473d1799ef4153502dbdd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


OGPoyraz:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/types.ts
```diff
@@ -496,6 +496,7 @@ export type TransactionPaySourceAmount = {
  * and `detail` on {@link QuoteErrorInfo}.
  */
 export type QuoteErrorReason =
+  | 'atomic-promotion-failed'
```
matthewwalsh0, line 499: Do we need an explicit reason for this? Or is this a subset of `no-quotes` and we could use our error prefix pattern to get a good error message?

packages/transaction-pay-controller/src/strategy/relay/relay-submit-execute.ts
```diff
@@ -75,7 +75,7 @@ async function submitViaRelayExecuteInternal(
   return FALLBACK_HASH;
 }
 
-function isSubsidizedRelayQuote(quote: RelayQuote): boolean {
+export function isSubsidizedRelayQuote(quote: RelayQuote): boolean {
```
matthewwalsh0, line 78: Minor, exported above private?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -150,6 +155,11 @@ export async function getRelayQuotes(
     return quotes;
   } catch (error) {
     log('Error fetching quotes', { error });
+
+    if (quotes.some(isPromotedSubsidizedMaxMoneyAccountQuote)) {
```
matthewwalsh0, line 159: Can we be sure any error under these conditions is because of those conditions?

Should we instead just add an error prefix around the re-quote?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+  );
+}
+
+function shouldAttemptAtomicPromotion(
+  request: QuoteRequest,
+  transaction: TransactionMeta,
+  discoveryQuote: TransactionPayQuote<RelayQuote>,
+): boolean {
+  return (
+    isMoneyAccountDepositTransaction(transaction) &&
```
matthewwalsh0, line 290: Is this needed?

Can we avoid the coupling as atomic and subsidised should always be okay?

Could this be a feature flag keyed by transaction type?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+  );
+
+  return maybePromoteSubsidizedMaxMoneyAccountQuote({
+    discoveryQuote,
+    fullRequest,
+    request,
+  });
+}
+
+async function maybePromoteSubsidizedMaxMoneyAccountQuote({
```
matthewwalsh0, line 190: Minor, can some of these utils go in `relay-max-gas-station.ts`?

Maybe we rename to `relay-max.ts`?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+    const transactionClone = cloneTransactionForPromotion(
+      fullRequest.transaction,
+    );
+    const promotionTransaction = await applyAmountDataUpdates({
+      amount: targetAmount,
+      messenger: fullRequest.messenger,
+      transaction: transactionClone,
+    });
+
+    const promotedQuote = await getSingleQuote(
```
matthewwalsh0, line 223: We discussed which way to default this, as if the subsidized flow is the more common, we could do the initial quote with `atomic` and then do the second quote only if resulting quote is not subsidized?

And we could "hint" this from the client by setting atomic based on fixed spreads feature flag and checking route, so it's already off if route cannot be subsidized?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+    throw new Error('Promoted quote request atomic flag was not set');
+  }
+
+  /* istanbul ignore next: re-quote hardcodes the discovery target; defensive guard. */
+  if (promotedQuote.request.targetAmountMinimum !== targetAmount) {
+    throw new Error('Promoted quote target amount minimum was not preserved');
+  }
+
+  if (discoveryQuote.original.details.currencyOut.amount !== targetAmount) {
+    throw new Error('Discovery quote target amount changed before promotion');
```
matthewwalsh0, line 433: I don't think we're handling enough error states 😄

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+
+  /* istanbul ignore next: re-quote hardcodes isMaxAmount false; defensive guard. */
+  if (promotedQuote.request.isMaxAmount !== false) {
+    throw new Error(
+      'Promoted quote request isMaxAmount was not cleared before restore',
+    );
+  }
+
+  /* istanbul ignore next: re-quote hardcodes atomic true; defensive guard. */
+  if (promotedQuote.request.atomic !== true) {
```
matthewwalsh0, line 423: Are some of these necessary? Are we guarding against Relay returning a response that simply ignores our request params?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -164,7 +174,285 @@ async function getQuoteWithMaxAmountHandling(
...
+  /* istanbul ignore next: atomic re-quotes always embed funding and delegation calls; defensive guard. */
+  if (!promotedQuote.original.request.txs?.length) {
+    throw new Error('Promoted quote is missing embedded calls');
+  }
+
+  const sourceCost = getPromotedSourceCost({ promotedQuote, request });
+
+  const budget = new BigNumber(request.sourceTokenAmount);
+
+  if (sourceCost.isGreaterThan(budget)) {
```
matthewwalsh0, line 411: This is already validated in quote validation phase generically, checking the amount in the quote is less or equal to balance.

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: > We discussed which way to default this, as if the subsidized flow is the more common, we could do the initial quote with atomic and then do the second quote only if resulting quote is not subsidized?

I think it make sense, we can do that.

> And we could "hint" this from the client by setting atomic based on fixed spreads feature flag and checking route, so it's already off if route cannot be subsidized?

To clarify here, in clients we will say;

```ts
const expectSubsidy =
  isFixedSpreadsEnabled && isSubsidizedRoute(sourceChain, sourceToken);

config.atomic = expectSubsidy; // "I expect atomic to work" - not "do atomic"
```

And in controller: 

```ts
const quote = await getSingleQuote({ ...request, atomic: hint }, fullRequest);

if (hint === true && !isSubsidizedRelayQuote(quote.original)) {
  // Hint was wrong: fall back to sponsored second leg.
  return getSingleQuote({ ...request, atomic: false }, fullRequest);
}

if (hint === false && isSubsidizedRelayQuote(quote.original)) {
  // Bonus case: unexpected subsidy - upgrade to atomic (this is our max promotion, generalized).
  return promoteToAtomic(...);
}

return quote;
```

Mobile PR: https://github.com/MetaMask/metamask-mobile/pull/36343

OGPoyraz:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Updating my comment after tracing it through: trying atomic first doesn't work for Max deposits. To build an atomic request we need to know the output amount, but for Max that's exactly what we're trying to discover - so there's nothing valid to put in the request. The code refuses that combination outright:

```ts
// in `processTransactions`
if (isMaxAmount) {
    throw new Error('Max amount quotes do not support included transactions');
  }
```

And it should: without the refusal it would build quotes for ~$0 and read subsidy answers off meaningless results. So for Max, look-first-then-go-atomic isn't a preference, it's the only order that works.

Related point as I remember from past: this also decouples subsidized routes from client code - clients carry no "hint" logic, and atomic promotion stays a controller-side decision driven by actual subsidy. That keeps the surface in one place (the controller flag) instead of spreading route knowledge across clients.

packages/transaction-pay-controller/src/types.ts
OGPoyraz: Dropped the custom reason; terminal failures are `no-quotes` with an `Atomic promotion failed: ` prefix now.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Added `payStrategies.relay.atomicMaxPromotionEnabled` (`{default?, transactionTypes?}`), same shape as the validation flag.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Promotion and utils now in `relay-max.ts`

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Agreed, unnecessarily tight to implement these guards, removed all

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Removed these

OGPoyraz:
@metamaskbot publish-preview

OGPoyraz:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -167,7 +193,34 @@ async function getQuoteWithMaxAmountHandling(
     return getQuoteWithPostQuoteGasHandling(request, fullRequest);
   }
 
-  return getRelayMaxGasStationQuote(request, fullRequest, getSingleQuote);
+  const isAtomicPromotionEligible = isAtomicMaxPromotionEnabled(
+    fullRequest.messenger,
```
matthewwalsh0: Minor, destructure?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -167,7 +193,34 @@ async function getQuoteWithMaxAmountHandling(
...
+    fullRequest.transaction,
+  );
+  // The required target amount may predate Max. An atomic hint does not
+  // establish the output obtainable from the source budget; discover it first.
+  const discoveryRequest =
+    request.atomic !== false &&
+    request.isPostQuote !== true &&
+    (isAtomicPromotionEligible ||
+      hasTransactionType(fullRequest.transaction, [
+        TransactionType.moneyAccountDeposit,
```
matthewwalsh0: Isn't the benefit of a feature flag to avoid coupling to specific transaction types?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -519,7 +572,13 @@ async function processTransactions(
     return true;
   }
 
-  if (isMaxAmount) {
+  // Eligible atomic max requests carry a destination amount, so the calls
+  // below use EXACT_OUTPUT rather than the usual max EXACT_INPUT quote.
+  if (
+    isMaxAmount &&
+    (request.isPostQuote === true ||
+      !isAtomicMaxPromotionEnabled(messenger, transaction))
```
matthewwalsh0: Minor, are the semantics / function name a little confusing here?

The promotion aspect doesn't matter, just whether we support "atomic max"?

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
```diff
@@ -167,7 +193,34 @@ async function getQuoteWithMaxAmountHandling(
...
+  // The required target amount may predate Max. An atomic hint does not
+  // establish the output obtainable from the source budget; discover it first.
+  const discoveryRequest =
+    request.atomic !== false &&
+    request.isPostQuote !== true &&
+    (isAtomicPromotionEligible ||
+      hasTransactionType(fullRequest.transaction, [
+        TransactionType.moneyAccountDeposit,
+      ]))
+      ? { ...request, atomic: false }
```
matthewwalsh0: So we're forcing `atomic` to `false` for the discovery quote, but what does the discovery quote even do for our flow?

If it's a max subsidized deposit, we just need one `EXACT_OUTPUT` quote?

And if it's not subsidzed, we just need one `EXACT_INPUT` quote and we'll do the vault / transaction in the second non-atomic leg?

Does this force two quotes minimum?

Is this a relic from the legacy gas station support which had to subtract the cost of the Sentinel API gas fee token? Should we remove that entirely in a separate PR since we have the simpler Relay execute flow now?

packages/transaction-pay-controller/src/strategy/relay/relay-max.ts
```diff
@@ -466,3 +473,173 @@ function markQuoteAsMaxGasStation(
...
+  getSingleQuote,
+  request,
+}: {
+  discoveryQuote: TransactionPayQuote<RelayQuote>;
+  fullRequest: PayStrategyGetQuotesRequest;
+  getSingleQuote: GetSingleQuoteFn;
+  request: QuoteRequest;
+}): Promise<TransactionPayQuote<RelayQuote>> {
+  if (
+    !shouldAttemptAtomicPromotion(
```
matthewwalsh0: Does this mean we always do the first quote as non-atomic?

I thought we were going to use the `atomic` flag from client to match that to avoid an extra quote in most cases?

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -195,10 +195,16 @@ export type RelayValidationEnabledConfig = {
...
+export type AtomicMaxPromotionEnabledConfig = {
+  default?: boolean;
+  transactionTypes?: Partial<Record<TransactionType, boolean>>;
+};
+
 type FeatureFlagsExtendedRaw = {
   excludeChainIdsFromInfura?: Hex[];
   payStrategies?: {
     relay?: {
+      atomicMaxPromotionEnabled?: AtomicMaxPromotionEnabledConfig;
```
matthewwalsh0: Is this technically `atomicMaxEnabled` and the promotion is a side effect of that?

packages/transaction-pay-controller/src/utils/feature-flags.ts
```diff
@@ -689,6 +695,54 @@ export function isRelayValidationEnabled(
...
+  for (const [type, enabled] of Object.entries(transactionTypes)) {
+    if (hasTransactionType(transaction, [type as TransactionType])) {
+      return enabled;
+    }
+  }
+
+  // Absent flag preserves the original scope: direct Money Account deposits.
+  return (
+    promotionEnabled?.default ??
+    hasTransactionType(transaction, [TransactionType.moneyAccountDeposit])
```
matthewwalsh0: I thought the flag would remove the coupling?

packages/transaction-pay-controller/src/utils/quotes.ts
```diff
@@ -758,6 +759,25 @@ async function getQuotes(
         ? caughtError.info
         : { message: (caughtError as Error).message, reason: 'no-quotes' };
 
+      if (
```
matthewwalsh0: Can we remove this now since we'll get the prefix in the standard quote error already?

And if we want logs, we could do that where we `try catch` and apply our prefix?

matthewwalsh0 (changes requested):

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
matthewwalsh0: Afraid I don't follow, the amount in the quote would be the user balance as with any `isMaxAmount` flow.

Except this special atomic max flow and the subsidized quote means we can use it with `EXACT_OUTPUT` (after adjusting decimals) confident that the user can afford it as there will be no additional fees added and the output is 1:1 as fixed spreads only support stablecoins.

OGPoyraz:
@metamaskbot publish-preview

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Got it, thanks for clarifying. Atomic max now uses the source budget adjusted to destination decimals for the 1:1 subsidized route, rather than the pre-Max required amount. It retries non-atomically if the quote isn’t subsidized.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: Removed the Money Account check to entirely check flags

packages/transaction-pay-controller/src/strategy/relay/relay-max.ts
OGPoyraz: Now honoring the hint. `true` or omitted tries atomic first; `false` starts non-atomically. 

A subsidized non-atomic quote can still be promoted.

packages/transaction-pay-controller/src/utils/feature-flags.ts
OGPoyraz: Renamed to `payStrategies.relay.atomicMaxEnabled`, along with the config type and helper.

packages/transaction-pay-controller/src/utils/quotes.ts
OGPoyraz: Removed the special branch and log. The prefixed error now goes through standard quote-error handling and strategy fallback.

packages/transaction-pay-controller/src/strategy/relay/relay-quotes.ts
OGPoyraz: You’re right. The mandatory discovery quote was there to obtain the destination amount before rebuilding the vault calldata. That isn’t necessary for the subsidized 1:1 stablecoin route: we can use the source budget adjusted to destination decimals. I’ve changed the flow accordingly.

With `atomicMaxEnabled` enabled, the Relay execute flow now behaves as follows:
- Atomic hint + subsidized: one `EXACT_OUTPUT` quote, with the vault calls embedded.
- Non-atomic hint + unsubsidized: one `EXACT_INPUT` quote, with the vault/transaction handled in the separate second leg.
- Atomic hint + unsubsidized: two quotes, first `EXACT_OUTPUT`, then a non-atomic `EXACT_INPUT` retry.
- Non-atomic hint + subsidized: two quotes, first `EXACT_INPUT`, then promotion to atomic `EXACT_OUTPUT` using the returned output amount.

So it no longer forces two quotes minimum. When the hint matches the subsidy result, we only need one.

The mandatory discovery step and legacy gas-station adjustment were separate concerns. I’ve left its removal for a separate PR so we don’t change the non-execute fallback paths here.

OGPoyraz: Preview tests looking good

https://github.com/user-attachments/assets/03312f68-b9cc-49a9-9aae-bcd338807806

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

PR #10230, opened 2026-09-14
feat(assets-controller): add transient per-account loading state for user-visible fetches

Prithpal-Sooriya:
## Description

Adds a **transient, non-persisted per-account loading state** to `AssetsController`, published only while assets are being fetched for specific user-visible UX moments (account switch, unlock), plus selectors for the UX layer to read it.

### Problem

The UI has no signal for "an assets fetch is in flight *because the user just did something*" (switched accounts / unlocked the wallet). Balance state can't be used to infer this (stale balances are indistinguishable from fresh ones), and `getAssets` is called constantly in the background (polling, price/currency/tx refreshes) — so "any fetch in flight" is not the same as "the user is waiting".

### Approach

- **New state field:** `assetsLoadingStatus: Record<AccountId, AssetsLoadingTrigger>` where `AssetsLoadingTrigger = 'accountSwitch' | 'unlock'`. A string-literal union was chosen over `Record<_, boolean>`: the key's *presence* already encodes "loading", the *value* tells the UI which moment is loading (so it can render the right skeleton/label), and the union is extensible for future triggers without a second field or a breaking change.
- **Transient by construction:** the field's state metadata is `persist: false` (with `includeInStateLogs`, `includeInDebugSnapshot`, `usedInUi` set). A restarted client always begins with no fetch in flight, so nothing is persisted and **no migration is needed**.
- **Scoped opt-in:** `getAssets` accepts a new optional `trigger` option. Only the two user-action call sites pass it — `#handleAccountGroupChanged` (`'accountSwitch'`) and `#runStartupRefresh` (`'unlock'`). Every other `getAssets` caller (polling, price updates, tx confirmations, currency changes, network events) is unchanged and never publishes loading state, so the indicator cannot flap due to background work.
- **Lifecycle in `getAssets`:** the public method is now a thin wrapper — set the entries, `try { await } finally { clear }` — around the extracted, unchanged fetch pipeline (`#getAssetsInternal`). Entries are cleared conditionally (only if they still hold this fetch's trigger), so an older fetch settling can never clobber a newer fetch's marker for the same account.
- **Selectors:** `getAccountLoadingStatus(state, accountId)`, `isAccountLoading(state, accountId)`, `getAccountsLoadingStatus(state)`, `isAnyAccountLoading(state)` are exported from the package index for the UX team, mirroring the existing `selectors/balance.ts` pattern.

### `getAssets` complexity & decorator evaluation (design note)

The requested analysis of `getAssets` complexity and whether a TypeScript decorator would be cleaner:

- `getAssets` was already a long, multi-stage pipeline (force-update fast/slow paths, middleware chain, merge/replace modes). The loading concern is *not* deeply entangled with that logic — it is a strict request-scoped lifecycle (set → run → clear). Modeling it inside the pipeline body would have raised cyclomatic complexity and added error-handling paths through ~180 lines that today have none.
- A decorator (`@withLoadingStatus`) was **evaluated and rejected**:
  - **No precedent:** there are zero decorators in the entire monorepo; the codebase's established idiom for wrapping controller methods is functional composition (e.g. `withTrace` in `src/utils/trace.ts`). Introducing the first decorator raises toolchain risk (the repo's tsconfig does not enable `experimentalDecorators`; the TC39 stage-3 decorators proposal has evolving semantics) and review cost.
  - **Not a generic concern:** the wrapper is parameter-aware (per-account keys from the *first argument*, `trigger` from an *options bag property*) — a decorator would need bespoke argument inspection anyway, buying no reuse for other methods.
  - **Discoverability:** decorators hide control flow from readers and from the generated action types; an explicit wrapper keeps the lifecycle visible at the method signature, which is what `AssetsController-method-action-types.ts` (auto-generated from `AssetsController['getAssets']`) naturally picks up.
- The chosen **standard implementation** adds ~10 lines to `getAssets` and two small private helpers, with zero cyclomatic-complexity growth in the fetch pipeline itself (its body moved verbatim to `#getAssetsInternal`). This is the smallest diff that satisfies the lifecycle requirement while staying within repo idioms.

## Related issues

- https://consensyssoftware.atlassian.net/browse/ASSETS-3907

## Checklist

- [x] I've updated the changelog (Unreleased → Added, `packages/assets-controller/CHANGELOG.md`).
- [x] Tests added/updated: 7 new controller tests covering set/clear per trigger, no-trigger fetches, event wiring (`accountSwitch`, `unlock`), `stateChanged` emission, non-persistence via `deriveStateFromMetadata`, and concurrent-fetch marker safety; 12 selector tests.
- [x] `yarn workspace @metamask/assets-controller run test` — 1034/1034 passing.
- [x] Monorepo `yarn build` (type check) passing.
- [x] `yarn eslint … --fix --prune-suppressions` and `yarn lint:misc:check` clean on touched files; `AssetsController-method-action-types.ts` regenerated via `yarn workspace @metamask/assets-controller run messenger-action-types:generate`.

## Test plan

1. `yarn workspace @metamask/assets-controller run test` — all suites pass.
2. Gated-client unit tests observe `assetsLoadingStatus` mid-flight (Accounts API calls frozen via an armed gate) and after release — entries present exactly while the fetch is in flight and cleared on settle.
3. `deriveStateFromMetadata` test asserts the field is excluded from persisted state.

---

> [!NOTE]
> **Low Risk**
> UI-only transient state around existing `getAssets` force refreshes; balance/persist logic is unchanged aside from a small migration merge fix.
> 
> **Overview**
> Adds **transient, non-persisted** per-account loading signals so the UI can tell when a forced assets fetch is in flight.
> 
> **`assetsLoadingStatus`** (`'loading' | 'loaded'`) and internal **`assetsLoadingTokens`** are new controller state (not persisted). A **`@trackAssetsLoading`** decorator on **`getAssets`** marks requested accounts as loading before the fetch and settles them to loaded in a `finally` block when **`forceUpdate: true`**; cache-only calls and empty account lists are ignored. Token ownership prevents an older overlapping fetch from flipping an account to loaded while a newer fetch still owns the marker.
> 
> New **loading selectors** (per account, account group, and selected group) are exported from the package. Tests cover gated API clients, queued account switches, failures, state events, and non-persistence. Also fixes the temporary **`assetsInfo`** heal migration to merge healed fields via a full state spread instead of replacing state wholesale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f87e9ae4170e03cafa45daf10b17590577b65be0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/AssetsController-method-action-types.ts
```diff
@@ -5,6 +5,34 @@
 
 import type { AssetsController } from './AssetsController.js';
 
+/**
```
Prithpal-Sooriya: Remove this JSDoc addition, I want to keep PR minimal

packages/assets-controller/src/AssetsController.ts
```diff
@@ -261,6 +262,13 @@ export type AssetsControllerState = {
   assetPreferences: { [assetId: string]: AssetPreferences };
   /** Currently-active ISO 4217 currency code */
   selectedCurrency: SupportedCurrency;
+  /**
```
Prithpal-Sooriya: Go through the PR changes, remove all these unnecessary comments please

packages/assets-controller/src/types.ts
```diff
@@ -408,6 +408,18 @@ export type DataResponse = {
  */
 export type AssetsUpdateMode = 'full' | 'merge' | 'update';
 
+/**
```
Prithpal-Sooriya: remove comment please...

packages/assets-controller/src/types.ts
```diff
@@ -408,6 +408,18 @@ export type DataResponse = {
...
+ * User-visible moments for which an assets fetch is surfaced as a loading
+ * state in {@link AssetsControllerState.assetsLoadingStatus}.
+ *
+ * - **accountSwitch**: The selected account group changed (e.g. the user
+ *   switched accounts), so balances for the newly selected accounts are
+ *   being fetched.
+ * - **unlock**: The wallet was unlocked (or tracking started after unlock),
+ *   so balances are being fetched before anything is renderable.
+ */
+export type AssetsLoadingTrigger = 'accountSwitch' | 'unlock';
```
Prithpal-Sooriya: IDK, I kinda wanted the loading state to be a bit simpler.

The loading states be like `'loading' | 'loaded'`. it is meant to be a 1-time loading for now.

packages/assets-controller/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add a transient, non-persisted per-account assets loading state. `AssetsControllerState` now includes `assetsLoadingStatus: Record<AccountId, AssetsLoadingTrigger>` (`'accountSwitch' | 'unlock'`), populated while `getAssets` runs for user-visible moments and cleared when the fetch settles; the field is never persisted. `getAssets` accepts a new optional `trigger` option that opts a fetch into this behavior — only account switches and unlock-time startup refreshes pass it ([#10230](https://github.com/MetaMask/core/pull/10230))
+- Add loading-state selectors `getAccountLoadingStatus`, `isAccountLoading`, `getAccountsLoadingStatus`, and `isAnyAccountLoading` for the UX layer to read `assetsLoadingStatus` ([#10230](https://github.com/MetaMask/core/pull/10230))
```
Prithpal-Sooriya: This changelog is massive, keep to 1 sentence. We don't need to explain everything new exported.

Prithpal-Sooriya (commented):
Look through self-review comments

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3637,6 +3776,7 @@ export class AssetsController extends BaseController<
         await this.getAssets(accounts, {
           chainIds: [...this.#enabledChains],
           forceUpdate: true,
+          trigger: 'accountSwitch',
```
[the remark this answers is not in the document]
Prithpal-Sooriya: Fixed in 1b282cc5: the loading marker is now set in `#handleAccountGroupChanged` / `#runStartupRefresh` before acquiring `#accountRefreshMutex`, so a queued switch marks its newly selected accounts immediately while the previous refresh still holds the mutex. The status settles to `'loaded'` in a `finally` once the fetch completes. Added a regression test ("marks a queued account switch as loading immediately, while a previous refresh still holds the refresh mutex") that freezes the first switch fetch mid-flight and asserts the second group's account is marked without waiting.

packages/assets-controller/src/AssetsController-method-action-types.ts
Prithpal-Sooriya: Done — the loading lifecycle moved out of `getAssets` into the account-switch/unlock handlers, so `getAssets` is back to its original form and this generated file is now identical to `main` (zero diff).

packages/assets-controller/src/AssetsController.ts
Prithpal-Sooriya: Done — removed the doc comments on the state field, metadata entry, default state, helpers, and `getAssets`. The only comments left in the diff are pre-existing ones that were re-indented by the handler nesting.

packages/assets-controller/src/types.ts
Prithpal-Sooriya: Done — it is now just `export type AssetsLoadingStatus = 'loading' | 'loaded';` with no comment.

packages/assets-controller/src/types.ts
Prithpal-Sooriya: Done — the field is now `assetsLoadingStatus: Record<AccountId, AssetsLoadingStatus>` with `AssetsLoadingStatus = 'loading' | 'loaded'`. Accounts are marked `'loading'` when a switch/unlock fetch starts and flip to `'loaded'` when it settles (success or failure) — a one-time, never-persisted marker. The trigger-specific encoding is gone, and `getAssets` no longer takes a `trigger` option.

packages/assets-controller/CHANGELOG.md
Prithpal-Sooriya: Done — collapsed to one sentence: "Add a transient, non-persisted per-account assets loading state (`assetsLoadingStatus`) that marks accounts as loading during account switches and unlock, with selectors to read it."

packages/assets-controller/src/migrations/healAssetsInfoMetadata.ts
```diff
@@ -99,9 +99,11 @@ export type AssetsInfoHealingPatch = {
 
 const log = createModuleLogger(projectLogger, 'tempHealAssetsInfoMetadata');
 
-export type TempHealAssetsInfoMetadataOptions = {
+export type TempHealAssetsInfoMetadataOptions<
+  State extends AssetsControllerStateInternal = AssetsControllerStateInternal,
+> = {
   /** Current `AssetsController` state the healing patch is computed against. */
-  state: AssetsControllerStateInternal;
+  state: State;
```
Prithpal-Sooriya: Was this type change necessary in this file? Is it possible to keep the same?

packages/assets-controller/src/selectors/loading.ts
Prithpal-Sooriya, line 1: Is it possible to create selectors here that also go against accountGroup - that way engineers don't need to check against the evm accountID, then sol accountID, then other accountIDs for the account group

Prithpal-Sooriya, line 1: Also thoughts on making a more developer friendly state. `getIsAssetsLoadingForSelectedAccountGroup`? So they don't need to inject anything?

packages/assets-controller/src/AssetsController.test.ts
```diff
@@ -3765,6 +3826,215 @@ describe('AssetsController', () => {
...
+        async ({ controller, messenger }) => {
+          await activateTracking(messenger);
+
+          arm();
+          (messenger.publish as CallableFunction)(
+            'AccountTreeController:selectedAccountGroupChange',
+            'entropy:mock-keyring-id-1/1',
+            'entropy:mock-keyring-id-1/0',
+          );
+          await flushPromises();
```
Prithpal-Sooriya: This test is awful with code smells.
Honestly I'd prefer if we followed integration tests approach where we can fake nock some endpoints and force loading that way.

Also never use flush promises for this new tests, its a terrible code-smell. Use the `waitFor` util.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1357,27 +1370,36 @@ export class AssetsController extends BaseController<
    * @param accounts - Selected accounts to refresh.
    */
   async #runStartupRefresh(accounts: InternalAccount[]): Promise<void> {
-    const releaseLock = await this.#accountRefreshMutex.acquire();
+    const loadingToken = this.#setAssetsLoadingStatus(accounts);
```
Prithpal-Sooriya: Location for this logic seems wrong. Can we just add the logic to the `this.getAssets` method instead of these 2 places? Likewise I don't like the additional logic and code changes in assets-controller, this is why I'm interested if this can be pulled out as a possible decorator. Try it please, if it fails you can reply to this comment

Prithpal-Sooriya: Likewise I don't like the additional logic and code changes in assets-controller, this is why I'm interested if this can be pulled out as a possible decorator. Try it please, if it fails you can reply to this comment

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1798,6 +1816,24 @@ export class AssetsController extends BaseController<
...
+        state.assetsLoadingStatus[account.id] = 'loading';
+      }
+    });
+  }
+
+  #markAssetsLoaded(accounts: InternalAccount[]): void {
+    this.update((state) => {
+      for (const account of accounts) {
+        if (state.assetsLoadingStatus[account.id] === 'loading') {
+          state.assetsLoadingStatus[account.id] = 'loaded';
```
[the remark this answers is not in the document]
Prithpal-Sooriya: Fixed in 8f92a386d91: the handler now records an ownership token per account when it sets `loading`, and the settle step only flips accounts still owned by that token. A regression test covers the exact case here — an older refresh settling while a newer queued refresh owns the marker stays `loading`, and flipping the guard in the test reproduces this finding.

packages/assets-controller/src/migrations/healAssetsInfoMetadata.ts
Prithpal-Sooriya: Good call — reverted the migration file to exactly what's on main (zero diff now), and the call site in the controller instead spread-merges the heal result over `this.state`, so no type in this file needed to change.

packages/assets-controller/src/selectors/loading.ts
Prithpal-Sooriya: Added in ec06807b7e: `getAccountGroupLoadingStatus(state, accountTreeState, groupId)` and `isAccountGroupLoading(state, accountTreeState, groupId)` resolve the group's accounts from the account tree (reusing `getAccountIdsForGroup` from the balance selectors), so consumers no longer need to enumerate per-namespace account ids themselves.

Prithpal-Sooriya: Added in ec06807b7e: `getIsAssetsLoadingForSelectedAccountGroup(state, accountTreeState)` resolves the selected account group from the account tree state internally and returns whether any of its accounts is loading, so nothing has to be injected — only the two state slices. This also replaced the earlier explicit-id helpers, which are now redundant (`getAccountsLoadingStatus`/`isAnyAccountLoading` are gone).

packages/assets-controller/src/AssetsController.test.ts
Prithpal-Sooriya: Fair points, addressed in ec06807b7e:

- No `flushPromises` anywhere in the new tests — every wait now goes through `waitFor`, including waiting for the startup fetch to settle before arming.
- The `getAssets` spy is gone. The fake client now freezes and releases fetches one at a time (`arm()` installs a fresh gate, `release()` resolves the oldest pending one), so the race test holds the older refresh in flight, lets it settle, and keeps the newer queued refresh in flight — all deterministically, driven purely by events and the faked endpoints.
- The failure-path test now fails through the client (balance endpoints reject) instead of stubbing `getAssets`.

I kept the fake at the client boundary rather than nock because a refresh fans out across several API Platform endpoints (accounts V2/V5/V6, prices, staked balance, sweep) — faking just the accounts client keeps each test freezing exactly one fetch. Happy to convert these to full nock-based integration tests if you'd still prefer that.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -1357,27 +1371,36 @@ export class AssetsController extends BaseController<
    * @param accounts - Selected accounts to refresh.
    */
   async #runStartupRefresh(accounts: InternalAccount[]): Promise<void> {
-    const releaseLock = await this.#accountRefreshMutex.acquire();
+    const loadingToken = this.#setAssetsLoadingStatus(accounts);
```
Prithpal-Sooriya: We're almost there....

Location for this loading logic seems wrong.

Can we just add the logic to the this.getAssets method instead of these 2 places (runStartupRefresh, and handleAccountGroupChanged)?

Likewise I don't like the additional logic and code changes in assets-controller, this is why I'm interested if this can be pulled out as a possible decorator.


Try the decorator approach, if it fails you can reply to this comment. But then at least move the logic to the this.getAssets method.

Prithpal-Sooriya: The decorator worked, so it's in: new `trackAssetsLoading` method decorator (`src/trackAssetsLoading.ts`) on `getAssets` — marks the requested accounts `'loading'` at method entry and settles them to `'loaded'` in `finally`, whether the fetch succeeds or fails. Both handlers are back to their original shapes and all the marker logic lives in the decorator.

I kept the ownership-token logic (per-controller account→token map): when a newer `getAssets` call for the same account starts mid-flight, it re-claims the marker, so the older fetch settling can only flip accounts it still owns. There's a regression test with two overlapping direct `getAssets` calls where the older one settles first and the account stays `'loading'` until the newer one finishes.

Accepted tradeoffs of the new location: queued switches now mark when `getAssets` actually runs (after the refresh mutex frees), and every `getAssets` fetch marks — including direct action calls — consistent with the 1-time loading scope. The method signature is unchanged, so the generated action types are untouched.

packages/assets-controller/src/trackAssetsLoading.ts
```diff
@@ -0,0 +1,131 @@
...
+      assetsLoadingStatus: Record<AccountId, AssetsLoadingStatus>;
+    }) => void,
+  ) => void;
+};
+
+const loadingTokens = new WeakMap<
+  AssetsLoadingStateUpdater,
+  Map<AccountId, number>
+>();
+let loadingTokenCounter = 0;
```
Prithpal-Sooriya: I love this work you have created for this decorator 😍 

Some notes if you can try attempting.

1. Can we see if we improve the type of `AssetsLoadingStateUpdater`, ideally I want it to be the Actual AssetsController. If we can import the type, or pick on the type to reduce the size/scope of interface.

2. I'm wondering if we need this weakmap for `loadingTokens` here. Can we use the `state` from inside the controller? Is this possible? If yes I think the logic to `getTokens`/etc might be much cleaner and simpler.

3. Now that this is nicely decoupled, do you think we can offer some nice tests in isolation?

Prithpal-Sooriya: All three attempted, and all three landed:

1. The type now imports the real controller: `Pick<AssetsController, 'state'>` for the state, with `update` spelled out next to it — it's protected on the controller, so it can't be picked, but everything else comes straight from the actual type.

2. The WeakMap is gone. The ownership tokens now live in the controller's non-persisted state as `assetsLoadingTokens`, written and deleted in the same `update` calls that set `assetsLoadingStatus`, so there are no extra state change events. `markAssetsSettled` just reads `controller.state` to check ownership — no map bookkeeping left.

3. Added `src/trackAssetsLoading.test.ts` with isolated tests around a minimal fake controller: in-flight marking and settling, settling on failure, the older/newer overlap ownership rule, no accounts, and cache reads.

One extra from the latest Bugbot pass: calls that do not force an update (cache-only reads via `getAssetsBalance`/`getAssetsPrice`) never actually fetch, so the decorator now skips tracking them instead of emitting a momentary `loading` → `loaded` flap.

packages/assets-controller/src/trackAssetsLoading.ts
```diff
@@ -0,0 +1,131 @@
...
+): (this: This, ...args: Args) => Promise<Return> {
+  return async function (this: This, ...args: Args): Promise<Return> {
+    const [accounts] = args;
+    const controller = this as unknown as AssetsLoadingStateUpdater;
+    const token = markAssetsLoading(controller, accounts);
+    try {
+      return await target.call(this, ...args);
+    } finally {
+      markAssetsSettled(controller, accounts, token);
+    }
```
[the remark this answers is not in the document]
Prithpal-Sooriya: Partially fixed, and the rest is the accepted scope of this PR.

Fixed: calls that do not force an update are cache-only reads — they return state without fetching anything — and the decorator now skips tracking them entirely, so they no longer emit a momentary `loading` → `loaded` transition or extra state change events. This covers the `getAssetsBalance`/`getAssetsPrice` read paths and any direct `getAssets` call without `forceUpdate`.

Accepted for now: background fetches that do force an update (pollers, currency changes, transaction refreshes, network events, price follow-ups) still mark `loading` transiently. That is the direct consequence of moving the marker into `getAssets` itself per review direction in this PR ("1-time loading for now"), and the status is per-account and settles as soon as the fetch completes. Restricting marking to UX-moment fetches only (e.g. an opt-in flag the unlock/account-switch paths pass) is a straightforward follow-up if the flap on background refreshes turns out to matter in practice.

Prithpal-Sooriya:
@metamaskbot publish-previews

PR #10228, opened 2026-09-14
chore: make dependency range alignment opt-in

cryptodev-2s:
## Explanation

Follow up to #10225, addressing review feedback on it.

#10225 made the dependency range consistency constraint auto-fixable, which also made it fixable from `yarn lint:fix`. That risks a team hitting a constraints error, reaching for `yarn lint:fix` to clear it, and folding a monorepo-wide dependency bump into an unrelated pull request. Upgrades should stay in their own Dependabot pull requests.

This makes the alignment opt in and off by default. Every local invocation now behaves exactly as it did before #10225, same error, same wording. Only CI repairing a bot's pull request will opt in.

The mechanism is an environment variable rather than a flag because `yarn constraints` is a Yarn builtin and rejects anything other than `--fix` and `--json`:

```
$ yarn constraints --fix --align
Unknown Syntax Error: Unsupported option name ("--align").
```

There is deliberately no package script wrapping it. A convenient front door is what would tempt someone into bundling an upgrade into unrelated work, which is the thing being fixed.

## Verification

| Case | Result |
| --- | --- |
| Default, one package bumped | errors with "Pick one", no other manifest rewritten |
| `ALIGN_DEPENDENCY_RANGES=true` | aligns all 25 manifests, exit 0 |
| Clean tree, either mode | passes, nothing modified |
| Incomparable range (`npm:uuid@^14.0.2`) | errors regardless of the variable |

## References

Follows #10225. Part of WPC-1161.

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate (no harness exists for `yarn.config.cjs`)
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md) (no published package is touched)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Low Risk**
> Only changes Yarn constraint fix behavior for monorepo manifest consistency; no runtime package code or published APIs.
> 
> **Overview**
[...]

PR #10231, opened 2026-09-14
feat: repair Dependabot pull requests automatically

cryptodev-2s:
## Explanation

Dependabot's pull requests cannot merge on their own. Its security updates walk manifests one at a time rather than treating the Yarn workspace as one project, so a dependency ends up with different ranges in different packages and `yarn constraints` fails. It also does not deduplicate `yarn.lock` and does not write changelog entries. #10147, #10157, #10177 and #9369 are all stuck on this.

This adds a workflow that repairs all three and pushes the result back. It runs `yarn constraints --fix` with `ALIGN_DEPENDENCY_RANGES=true` (the opt in from #10228, whose only intended caller is this workflow), reinstalls, deduplicates, then runs the changelog fixer that already exists. Order matters: aligning ranges rewrites manifests but leaves `yarn.lock` stale.

It also adds `@lavamoat/*` to the npm allowlist. Both that and `@metamask/*` are already in `npmPreapprovedPackages`, so neither can trip the Yarn age gate during the repair install. Note that `allow` only governs version updates, so the workflow will also act on security pull requests for other packages, which is where the constraints failures come from.

## Notes

- `pull_request`, not `pull_request_target`. Dependabot runs get a read only token and no secrets, so the push is authorised by the OIDC exchange, the same way snaps does it.
- `opened` and `reopened` only. The job pushes to the pull request branch, so `synchronize` would loop.
- `[dependabot skip]` on the commit, otherwise Dependabot rebases and wipes the repair.

## Verification

The workflow cannot run until it is on the default branch. Checked locally:

| Check | Result |
| --- | --- |
| `constraints --fix` leaves `yarn.lock` stale | confirmed, hence the reinstall step |
| `ALIGN_DEPENDENCY_RANGES=true yarn constraints --fix` on a split range | aligns all 25 manifests, exit 0 |
| Default, without the variable | unchanged, still errors |

## References

Follows #10225 and #10228. Part of WPC-1161. Pattern borrowed from MetaMask/snaps `update-pull-request.yml`.

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate (workflow and config only)
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md) (no published package is touched)

---

> [!NOTE]
> **Low Risk**
> Changes are limited to Dependabot and CI workflow configuration; no runtime application or auth logic is modified, though the workflow does push commits to Dependabot branches via OIDC token exchange.
> 
> **Overview**
> Adds a GitHub Actions workflow that runs when **Dependabot** opens or reopens a PR, then fixes the common reasons those PRs fail CI: misaligned dependency ranges across the Yarn workspace, a stale or duplicated **yarn.lock**, and missing changelog entries.
> 
> The job exchanges an OIDC token for write access (Dependabot’s default token cannot push), runs **`yarn constraints --fix`** with **`ALIGN_DEPENDENCY_RANGES=true`**, reinstalls with **`--no-immutable`**, dedupes, and runs the existing **`changelog:validate --fix`** flow before committing with **`[dependabot skip]`** so Dependabot does not rebase away the repair. It only listens to **`opened`** / **`reopened`** (not **`synchronize`**) to avoid a push loop.
> 
> **Dependabot** config now allows version bumps for **`@lavamoat/*`** npm packages, matching **`@metamask/*`**, so those updates can go through the same repair path without Yarn age-gate issues during install.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c90706dc3a3462dd8c3357e3f3ab787a9cc18941. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/repair-dependabot-pull-requests.yml
```diff
@@ -0,0 +1,93 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  repair:
+    name: Repair constraints, lockfile and changelogs
+    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' && github.actor != 'metamaskbot' }}
+    runs-on: ubuntu-latest
+    environment: default-branch
```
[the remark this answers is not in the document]
mcmire: This is true. Since this workflow is triggered on the `pull_request` event, I don't think we need an environment?

.github/workflows/repair-dependabot-pull-requests.yml
```diff
@@ -0,0 +1,93 @@
...
+          is-high-risk-environment: false
+
+      - name: Align dependency ranges
+        env:
+          ALIGN_DEPENDENCY_RANGES: 'true'
+        run: yarn constraints --fix
+
+      # Aligning ranges rewrites manifests but not the lockfile, so it has to be resolved again before deduplicating.
+      - name: Resolve the lockfile
+        run: yarn install --no-immutable
```
mcmire: Will `--no-immutable` work here? My understanding is that we use this in workflows when we've updated the lockfile in Git but we want Yarn to install dependencies, and we don't want Yarn to update the lockfile anymore. However in this case we _do_ need to update the lockfile (because we're changing versions of dependencies), so maybe we need to drop `--no-immutable`? (Also, if we rephrase this step as "apply dependency range updates" maybe we don't need the comment?)

``` suggestion
      - name: Apply dependency range updates
        run: yarn install

File: .github/workflows/repair-dependabot-pull-requests.yml [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Deduplicate the lockfile
+        run: yarn dedupe
```
mcmire: Nit: We aren't deduplicating the lockfile per se, we're deduplicating the dependency tree, which the lockfile represents.

``` suggestion
      - name: Deduplicate dependency tree
        run: yarn dedupe

File: .github/workflows/repair-dependabot-pull-requests.yml (hunk 1 above)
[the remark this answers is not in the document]
Comment (cryptodev-2s) [reply .github/workflows/repair-dependabot-pull-requests.yml]: Yes make sense

File: .github/workflows/repair-dependabot-pull-requests.yml [hunk 4]
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Configure Git
+        run: |
+          git config user.name 'MetaMask Bot'
+          git config user.email 'metamaskbot@users.noreply.github.com'
```
mcmire: Nit: Since the token comes from `metamask-ci` do we want to use that instead? (this is copied from https://github.com/MetaMask/core-platform-metrics/blob/7b14be72749417587c6ab6b423b9455114292adc/.github/workflows/remove-dashboard-preview.yml#L49C1-L51C1)

``` suggestion
          git config user.name "metamask-ci[bot]"
          git config user.email "271559518+metamask-ci[bot]@users.noreply.github.com"
```

mcmire (commented):
Makes sense, just left a few comments/suggestions.

.github/workflows/repair-dependabot-pull-requests.yml
cryptodev-2s: Applied here https://github.com/MetaMask/core/pull/10231

.github/workflows/repair-dependabot-pull-requests.yml
cryptodev-2s: Reworded here https://github.com/MetaMask/core/pull/10231/changes/67b5b17d56697d79e4e8e310c81ddfe407bad2ce

.github/workflows/repair-dependabot-pull-requests.yml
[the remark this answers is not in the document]
cryptodev-2s: Fixed here https://github.com/MetaMask/core/pull/10231/changes/67b5b17d56697d79e4e8e310c81ddfe407bad2ce

.github/workflows/repair-dependabot-pull-requests.yml
cryptodev-2s: Actually why we can't use`yarn install` as ub the CI it defaults to `--immutable` and forbids lockfile changes, We need`--no-immutable` to allow this as the previous step we regenerate the lock.

mcmire: Oh, right, I got mixed up. Thanks.

.github/workflows/repair-dependabot-pull-requests.yml
```diff
@@ -0,0 +1,92 @@
...
+      - opened
+      - reopened
+
+permissions:
+  contents: read
+
+jobs:
+  repair:
+    name: Repair constraints, lockfile and changelogs
+    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' && github.actor != 'metamask-ci[bot]' }}
```
mcmire: Do you need to check that this doesn't come from `metamask-ci`? Is it enough to check that it comes from Dependabot?

``` suggestion
    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}

mcmire (commented):
One more question.

.github/workflows/repair-dependabot-pull-requests.yml
cryptodev-2s: Indeed its redundant and we don't use `synchronise` here

cryptodev-2s: Dropped here https://github.com/MetaMask/core/pull/10231/commits/c90706dc3a3462dd8c3357e3f3ab787a9cc18941

mcmire (approved):
LGTM!

PR #8836, opened 2026-05-15
feat(authenticated-user-storage): add watchlist support

Prithpal-Sooriya:
## Explanation

https://consensyssoftware.atlassian.net/browse/ASSETS-3219

Add watchlist support to the AUS SDK.

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [x] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> Adds new authenticated API calls and caching/validation paths for a new `assets-watchlist` blob; mistakes could cause client-side validation failures or unexpected overwrites, but changes are localized to this SDK.
> 
> **Overview**
> Adds **assets watchlist** support to `@metamask/authenticated-user-storage` via new `AuthenticatedUserStorageService.getAssetsWatchlist()` (returns `null` on 404) and `setAssetsWatchlist()` (PUT overwrite semantics) methods, exposed over the messenger as `AuthenticatedUserStorageService:getAssetsWatchlist`/`...:setAssetsWatchlist`.
> 
> Introduces an `AssetsWatchlistBlob` runtime/schema type with a `version: 1` literal plus a write-side `ASSETS_WATCHLIST_MAX_ASSETS` (100) cap enforced via superstruct before sending requests, and wires cache invalidation so `setAssetsWatchlist` refreshes subsequent `getAssetsWatchlist` reads.
> 
> Updates public exports plus README/CHANGELOG, and adds comprehensive nock-backed fixtures/mocks and unit tests for headers, error handling, caching behavior, and validation edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b041e27224541d4f966b564e6301b83ece1965cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #10233, opened 2026-09-14
feat(authenticated-user-storage): add user custom tokens feature (imported/hidden assets)

Prithpal-Sooriya:
**Jira:** [ASSETS-3937](https://consensyssoftware.atlassian.net/browse/ASSETS-3937)
**Pattern references:** #8836 (assets-watchlist domain), #9441 (singleton-blob preferences pattern)<br>**Backend contract:** [va-mmcx-notify-users-storage#128](https://github.com/consensys-vertical-apps/va-mmcx-notify-users-storage/pull/128) (`GET`/`PUT /api/v1/custom-tokens`)
**Design doc:** ADR `0002-authenticated-user-storage-for-user-imported-tokens` (aus-implementation notes)

## Summary

Adds a **user custom tokens** domain to the Authenticated User Storage SDK: a per-user singleton blob
recording which tokens the user chose to **import** and which they chose to **hide**, keyed by CAIP-19
asset identifiers. The PR contains both the endpoint layer (HTTP contract + strict schema validation)
the high-level SDK API (`importTokens` / `hideTokens`), and a clean-slate wipe (`clearUserAssets`).

```jsonc
// GET/PUT /api/v1/custom-tokens
{
  "version": 1,
  "importedAssets": [
    "eip155:1/erc20:0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
    "eip155:8453/erc20:0x833589fcd6edb6e08f4c7c32d4f71b54bda02913"
  ],
  "hiddenAssets": [
    "eip155:10/erc20:0xa0b86991c6218b36c1d19d4a2e9eb0ce36000000"
  ]
}
```

## API

| Method | Kind | Description |
|---|---|---|
| `getUserAssets(): Promise<UserAssetsBlob \| null>` | Low-level | `GET /custom-tokens`; returns `null` on 404 (first read), mirroring `getAssetsWatchlist` |
| `setUserAssets(blob, clientType?): Promise<void>` | Low-level | `PUT /custom-tokens`; validates (strict) → normalizes → PUTs the normalized blob → invalidates the get cache |
| `importTokens(ids, clientType?): Promise<UserAssetsBlob>` | High-level | Merges `ids` into `importedAssets` (dedup, order-preserving), removes them from `hiddenAssets` (import wins), persists, returns the resolved blob |
| `hideTokens(ids, clientType?): Promise<UserAssetsBlob>` | High-level | Merges `ids` into `hiddenAssets` (dedup), removes them from `importedAssets` (mutual exclusivity), persists, returns the resolved blob |
| `clearUserAssets(clientType?): Promise<void>` | High-level | Wipes the blob by persisting empty lists — a clean slate; idempotent, works with or without an existing record |

All five are exposed as messenger actions (`AuthenticatedUserStorageService:getUserAssets`,
`:setUserAssets`, `:importTokens`, `:hideTokens`, `:clearUserAssets`) via `MESSENGER_EXPOSED_METHODS`; the action-types
file is regenerated with `yarn messenger-action-types:generate`.

## Semantics & guarantees

- **Fail-open conflict resolution (ticket requirement):** if a CAIP-19 identifier exists in both
  `importedAssets` and `hiddenAssets`, the user's intent to import wins — it stays in
  `importedAssets` and is removed from `hiddenAssets`. The write is never rejected because of a
  conflict. Implemented as a pure function (`normalizeUserAssetsBlob`) applied on **every** write, so
  even direct `setUserAssets` calls and pre-existing server-side conflicts are resolved.
- **Deduplication safeguard (ticket requirement):** all lists are de-duplicated (order-preserving,
  first occurrence wins) before every write, both in the high-level merge and in `setUserAssets`.
- **Mutual exclusivity:** `importTokens` removes ids from `hiddenAssets`; `hideTokens` removes ids
  from `importedAssets`. Each identifier lives in at most one list.
- **Strict schema validation (ticket requirement):** write-side every entry must be a CAIP-19 asset
  identifier (`CaipAssetTypeStruct` from `@metamask/utils`); `version` is a `literal(1)`. Malformed
  blobs throw a superstruct `StructError` **before** the request is sent. Read-side stays lenient
  (plain strings) so existing server data is never rejected — same philosophy as the watchlist blob.
- **Fail-fast input validation:** `importTokens`/`hideTokens` validate `ids` before any network I/O.
- **Clean slate (wipe):** `clearUserAssets` persists the empty blob (`{ version: 1, importedAssets: [], hiddenAssets: [] }`) —
  the wipe mechanism the current API contract supports (the backend exposes record deletion only via internal
  auth-event/GDPR flows, not a public `DELETE` endpoint).
- **Per-list cap:** writes enforce at most `USER_ASSETS_MAX_ASSETS` (100) entries per list, mirroring the API's
  server-side cap — same convention as `ASSETS_WATCHLIST_MAX_ASSETS`; the lenient read schema stays uncapped.

## Backend contract (aligned with va-mmcx-notify-users-storage#128)

`GET`/`PUT /api/v1/custom-tokens`. On write the server enforces: `version === 1`, both lists required,
deduplicated, mutually exclusive, at most 100 entries per list — violations are **rejected with 400**
(not normalized: a full-blob PUT carries no ordering information, so the server cannot resolve
cross-list precedence — the SDK applies move/dedupe semantics pre-write, which it does on every write).

The SDK's normalize-then-assert pipeline guarantees every PUT satisfies these invariants, so the
server's 400 path is a safety net, not a dependency. A public `DELETE` endpoint (the store already has
a `DeleteCustomTokens` primitive used by account-deletion flows) would be a server-side follow-up if
true record removal is wanted; `clearUserAssets` works with the contract as deployed.

## Known limitations

- Read-modify-write race: like the existing watchlist/preferences endpoints, there is no ETag/If-Match
  optimistic concurrency in the current API. Concurrent writers can lose updates; same trade-off as
  the existing endpoints, out of scope here.

## Test plan

- 91 unit tests pass (package total), 100% coverage maintained (threshold-enforced).
- New coverage: messenger round-trips for all five actions (incl. `clearUserAssets`); 404→null; 401/500 → `HttpError`; malformed
  response body → `StructError`; caching + cache invalidation (incl. refetch after write, and after
  `importTokens`/`hideTokens`); request-body assertions proving dedup and fail-open normalization
  happen **before** the request; mutual-exclusivity assertions both directions; CAIP-19 strictness on
  write; fail-fast invalid ids (no network call made); `X-Client-Type` header handling; per-list cap (100)
  boundary + overflow (incl. `importTokens`/`hideTokens` merged-blob overflow); `clearUserAssets` wipes to
  empty lists and invalidates the cache.
- `yarn build` (full monorepo `tsc --build`): clean. `yarn lint:eslint`, `yarn lint:misc:check`
  (oxfmt), `yarn constraints`, `yarn changelog:validate`, `yarn readme-content:check`,
  `yarn messenger-action-types:check`: all clean.

## Changelog

Entry added under `## [Unreleased]` → `### Added`, linking this PR: [#10233](https://github.com/MetaMask/core/pull/10233).

[ASSETS-3937]: https://consensyssoftware.atlassian.net/browse/ASSETS-3937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> New authenticated user-preference persistence with merge semantics and list caps; mistakes could affect which tokens users see, but patterns mirror existing watchlist flows and validation runs client-side before writes.
> 
> **Overview**
> Adds a **user custom tokens** domain to `@metamask/authenticated-user-storage`: persisted `importedAssets` and `hiddenAssets` CAIP-19 lists via **`GET`/`PUT /custom-tokens`**, with **`getUserAssets`** returning `null` on 404 like the watchlist.
> 
> **`AuthenticatedUserStorageService`** gains **`setUserAssets`**, **`importTokens`**, **`hideTokens`**, and **`clearUserAssets`**, plus messenger actions and the **`UserAssetsBlob`** / **`USER_ASSETS_MAX_ASSETS`** exports. Writes validate CAIP-19 ids and cap each list at 100, then **normalize** (dedupe, mutual exclusivity with **import wins** on overlap) before PUT; **`importTokens`** / **`hideTokens`** read-merge-write and return the resolved blob.
> 
> Docs, changelog, validators, and broad unit coverage (HTTP, caching, invalidation, fail-fast validation) ship with the feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10c9e446357cac1df1c866a4401ef7d44be0073f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/authenticated-user-storage/src/authenticated-user-storage-method-action-types.ts
```diff
@@ -85,6 +85,95 @@ export type AuthenticatedUserStorageServiceSetAssetsWatchlistAction = {
   handler: AuthenticatedUserStorageService['setAssetsWatchlist'];
 };
 
+/**
+ * Returns the user-assets (custom tokens) blob for the authenticated user.
+ *
+ * @returns The user-assets blob, or `null` if none has been set (404).
+ */
```
Prithpal-Sooriya, line 92: I've mentioned this before, maybe we can make a memory update around this.

I dislike these overly verbose comments and JSDoc. See if they are necessary, but most of my new code I don't want it!

packages/authenticated-user-storage/src/authenticated-user-storage.test.ts
Prithpal-Sooriya, line 1: New test changes you added here feel primed for test tables/`testCases` array + `it.each`.
So some patterns hopefully around it.each that you can use in this repo.

You aren't limited to 1 massive testCases array, you can split this where it seems appropriate.

packages/authenticated-user-storage/src/authenticated-user-storage-method-action-types.ts
Prithpal-Sooriya: Addressed in a09928ee55: trimmed the JSDoc on `setUserAssets`, `importTokens`, and `hideTokens` (plus the `validators.ts`/`types.ts` docs) to concise one-to-two-line summaries with only the necessary `@param`/`@returns`/`@throws` tags, and regenerated the action-types file to match — net −109 lines of comment prose, no functional change. `getUserAssets` was already terse and is unchanged. The key semantics (fail-open conflict resolution, dedup) stay documented in one line each. Also filed the preference as a memory-update issue so future code follows it from the start.

packages/authenticated-user-storage/src/validators.ts
```diff
@@ -235,3 +236,126 @@ export function assertAssetsWatchlistBlobForWrite(
...
+ * every entry must be a CAIP-19 asset identifier
+ * (e.g. `eip155:1/erc20:0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48`),
+ * enforced via {@link CaipAssetTypeStruct} from `@metamask/utils`.
+ * Validation failures throw a `StructError` before the request is sent.
+ */
+const UserAssetsBlobWriteSchema = type({
+  version: literal(1),
+  importedAssets: array(CaipAssetTypeStruct),
+  hiddenAssets: array(CaipAssetTypeStruct),
+});
```
Prithpal-Sooriya, line 262: Let me know if this is stupid, but wondering if we can bake in the mutual exclusivity mentioned in the decisions doc for imported and hidden asset?

```
const MutuallyExclusive = refine(
  object({
    a: optional(array(string())),
    b: optional(array(string())),
  }),
  'MutuallyExclusive',
  (value) => {
    const hasA = Boolean(value.a?.length);
    const hasB = Boolean(value.b?.length);

    // XOR: exactly one must be present
    if (hasA === hasB) {
      return "Exactly one of 'a' or 'b' must be provided (not both, not neither)";
    }
    return true;
  }
);
```


If stupid, thats fine. But maybe gives us protection on write? If yes then I want to make sure when we perform the write it should never blow up on us. (safely handle mutual exclusion, if in both default for it to be in includedAssets)

packages/authenticated-user-storage/src/authenticated-user-storage.test.ts
Prithpal-Sooriya: Done in 43305bea6d2 — the repetitive new tests are now table-driven `it.each`, split per scenario rather than one massive array:

- `importTokens`/`hideTokens` mirrors collapsed into one describe with five 2-row tables (fresh blob on 404, merge+dedup, mutual exclusivity, invalid-id fail-fast, X-Client-Type header), using `as const` rows so `service[method]` stays type-safe.
- `getUserAssets` 401/500 error cases → `it.each([401, 500])`.
- `setUserAssets` malformed-blob cases → case-object table (also added a `hiddenAssets`-side invalid-entry row for symmetry).
- Cache-invalidation trio (`setUserAssets`/`importTokens`/`hideTokens`) → one table with a `run` callback per mutator.

All original assertions preserved (one `toContain` strengthened to `toStrictEqual`); 82/82 passing with the enforced 100% coverage, eslint/oxfmt/`lint:tsc` clean. Net −22 lines.

packages/authenticated-user-storage/src/validators.ts
Prithpal-Sooriya: Not stupid at all — good instinct, and I've baked it in (78fdd35a8c7), with two semantic adjustments:

1. **Per-identifier, not per-field.** The blob always carries both lists (they're required and may legitimately be empty — e.g. a fresh blob), so an XOR of `exactly one list present` would reject every normal blob. The real invariant from the decisions doc is: *no single identifier may appear in both lists*, so the refine is a set-intersection check:

```ts
const UserAssetsBlobNormalizedWriteSchema = refine(
  UserAssetsBlobWriteSchema,
  'MutuallyExclusiveUserAssets',
  (blob) =>
    blob.hiddenAssets.every(
      (assetId) => !blob.importedAssets.includes(assetId),
    ) || 'An identifier may not appear in both importedAssets and hiddenAssets',
);
```

2. **It validates the normalized blob, not the raw input.** Applying the refine to the raw blob would violate the ticket's fail-open rule — an identifier in both lists must never reject the write. So the write pipeline is now: structural assert on the raw input (version literal + CAIP-19 entries, still fails fast) → `normalizeUserAssetsBlob` (dedup + fail-open conflict resolution, import wins) → `assertUserAssetsBlobNormalized` with the refined schema, right before the PUT.

That gives exactly the guarantee you asked for: the exclusivity check **cannot blow up on user input** (normalization has already removed any overlap — if in both, it ends up in `importedAssets` only); it only trips if an internal bug ever produced an overlapping blob, which is precisely the write-side protection we want. The existing fail-open test (identifier deliberately in both lists) still passes and asserts the request body contains it only in `importedAssets`; two new unit tests cover the refine's pass and reject paths. 84/84 green, 100% coverage held, eslint/oxfmt/tsc clean.

PR #10241, opened 2026-09-15
feat: consume and return reserve value from the bridge API

Julink-eth:
## Explanation

Add optional quote.feeData.reserve on V2 quotes so clients can consume the native minimum-balance reserve returned by the Bridge API (
[va-mmcx-bridge-api#1327](https://github.com/consensys-vertical-apps/va-mmcx-bridge-api/pull/1327)).
reserve is a sibling of feeData.network, not a FeeType. Fee aggregators iterate FeeType keys (network, relayer, metabridge, txFee), so treating reserve as a fee would inflate quote cost. It is a balance that must remain in the source account after execution, not an amount spent.
Normalize reserve in toNormalizedAmounts the same way as other amount fields, and preserve it when coercing V1 ↔ V2.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Additive optional quote field and metadata helpers with tests; reserve is explicitly excluded from FeeType aggregation to avoid inflating quoted costs.
> 
> **Overview**
> Adds optional **`quote.feeData.reserve`** on V2 bridge quotes so clients can surface the Bridge API’s native minimum balance that must stay on the source account after execution.
> 
> **`reserve` is modeled as a sibling of `feeData.network`, not a `FeeType`**, so existing fee aggregation over `FeeType` keys does not treat it as spendable network/relayer cost. The quote validator accepts optional `reserve` entries (`AmountsAndAssetSchema`), and **`toNormalizedAmounts`** / **`toCurrencyValues`** now normalize atomic amounts and derive fiat for `reserve` the same way as other amount fields, without folding it into the `FeeType` loop. Tests cover reserve normalization and fiat conversion; the changelog documents the addition.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66154f406f29e66d1c2ee813b2a13d6dabbcf4e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bridge-controller/src/utils/quote-metadata/to-currency-values.ts
```diff
@@ -59,16 +59,23 @@ export const toCurrencyValues = (
...
+                  toCurrency(fee, usdToFiatExchangeRate),
+                ),
+              ]),
+          ),
+          ...(feeData.reserve && {
+            reserve: feeData.reserve.map(
+              (reserve) => toCurrency(reserve, usdToFiatExchangeRate) ?? {},
+            ),
+          }),
+        } as DeepPartial<QuoteResponse['quote']['feeData']>),
```
micaelae, line 78: Please add `reserve` to FeeType so there's no need to change this block

```suggestion
        feeData &&
        Object.fromEntries(
          Object.values(FeeType)
            .filter((feeType) => feeData[feeType])
            .map((feeType) => [
              feeType,
              feeData[feeType]?.map((fee) =>
                toCurrency(fee, usdToFiatExchangeRate),
              ),
            ]),
        ),
```

packages/bridge-controller/src/validators/quote.ts
```diff
@@ -96,6 +96,11 @@ export const QuoteSchema = intersection([
       [FeeType.TX_FEE]: optional(
         intersection([FeeDataSchema, TxFeeGasLimitsSchema]),
       ),
+      /**
+       * Native balance that must remain in the source account after execution.
+       * Copied through from V2; not a FeeType.
+       */
+      reserve: optional(array(AmountsAndAssetSchema)),
```
micaelae: This can be removed since reserve is not part of the QuoteResponse v1 object. Changes to the coercers , calculator and merge can also be removed along with this

```suggestion
```

packages/bridge-controller/src/utils/quote-metadata/to-currency-values.ts
Julink-eth: hmm, I know it would remove that change but I'd rather keep reserve off FeeType, the reason being all the keys in FeeType are spendable feeData keys where as reserve is not spendable per se, it's the balance that needs to stay in the account post tx.#
Plus we kept the key off FeeType in the API, that would make it a discrepancy between controller and API.

packages/bridge-controller/src/validators/quote.ts
Julink-eth: True, will remove it.

micaelae (approved):

PR #10243, opened 2026-09-15
fix: bump `uuid` to `^11.1.1` across the monorepo

cryptodev-2s:
## Explanation

Clears GHSA-w5hq-g745-h8pq, which accounts for 25 of the repo's 155 open Dependabot alerts. The advisory is a missing buffer bounds check in `v3`/`v5`/`v6` when `buf` is supplied; core only uses `v1` and `v4`, so nothing here was exploitable, but the alerts need clearing.

Dependabot proposed `^14.0.2` in #10147. This goes to `^11.1.1` instead, the lowest patched version, because uuid is ESM-only from v12 onwards and our Jest setup cannot load an ESM-only dependency without transforming it: `jest.requireActual('uuid')` fails with `SyntaxError: Unexpected token 'export'`. Moving to v14 is doable but needs `transformIgnorePatterns` work, which does not belong in a security fix.

Also drops `@types/uuid` from 11 manifests, since uuid has bundled its own types since v10.

`packages/utils/src/fs.test.ts` needed a small helper: `v4` is overloaded and `jest.spyOn` resolves to the last overload, which returns `Uint8Array` rather than `string`. That is true in both v11 and v14, so it is not specific to the version chosen.

## Client impact

Checked against the clients' lockfiles rather than assumed. Both already resolve uuid 11, so this introduces nothing new and core's `9.0.1` copy should collapse into the `11.1.1` already present.

| client | uuid versions already resolved |
| --- | --- |
| metamask-extension | 3.2.1, 3.4.0, 8.3.2, 9.0.1, 11.0.3, **11.1.1** |
| metamask-mobile | 3.3.2, 7.0.3, 8.3.2, 9.0.0, 9.0.1, 11.0.3, 11.1.0, 14.0.1 |

Worth noting for anyone tempted by v14 later: metamask-extension has no uuid 14 anywhere and pins a transitive `gridplus-sdk/uuid@^13.0.0` back down to `^8.3.2`.

## References

Supersedes #10147. Part of WPC-1161.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Monorepo-wide dependency and typing cleanup with no production API changes; usage stays on `v4`/`v1`-style imports clients already resolve at 11.x.
> 
> **Overview**
> Raises **`uuid`** from **`^9.0.1`** (and changelog references from older ranges) to **`^11.1.1`** on the root devDependency and across controller/utils packages, with matching **`yarn.lock`** and **Unreleased** changelog entries. This targets the patched line for advisory **GHSA-w5hq-g745-h8pq** without jumping to ESM-only **v12+**, which the PR notes would break Jest’s **`requireActual('uuid')`** pattern.
> 
> **`@types/uuid`** is removed from the manifests that had it, since **`uuid` ≥ v10** ships its own types.
> 
> The only non-manifest code change is in **`packages/utils/src/fs.test.ts`**: tests mock **`v4`** via a **`jest.fn`** wrapper and a **`mockUuidV4`** helper instead of **`jest.spyOn`** on a namespace import, avoiding **`v4`**’s overload typing returning **`Uint8Array`** when spied.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 197dee52d49d0a0bde9198a54e1a1e788e6a8a1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryptodev-2s:
@metamaskbot update-changelogs

packages/utils/src/fs.test.ts
```diff
@@ -31,6 +31,13 @@ jest.mock('uuid', () => {
   };
 });
 
+const mockUuidV4 = (value: string): void => {
```
mcmire, line 33: Can we change the way that we mock `v4` to do something like this?

``` typescript
import { v4 } from 'uuid';

jest.mock('uuid');

const v4Mock = jest.mocked(v4);

// ...

v4Mock.mockReturnValue(...);
```

This way we don't have to have a custom function for mocking `v4`.

cryptodev-2s: I am getting the same error
```
fs.test.ts(682,30): error TS2345: Argument of type 'string' is not assignable to parameter of type 'Uint8Array<ArrayBufferLike>'

cryptodev-2s: I was able to simplify it here using a named import https://github.com/MetaMask/core/pull/10243/changes/fff9fb8a5476b7db78ba7f639c9354a0306b178a

mcmire: Gotcha, sounds good!

mcmire (approved):
Checked over `uuid`'s changelog and there doesn't seem to be any changes that impact us. LGTM.

PR #10252, opened 2026-09-15
feat: present a bearer token on requests to the built-in Infura endpoints

cryptodev-2s:
## Explanation

Requests to our built-in Infura endpoints should carry a bearer token in the `Authorization` header. The token identifies the user and can be refreshed at any time, so it has to be resolved per request rather than captured when a network client is built.

`RpcService` had no way to do that. It accepts a `fetch` and a static `fetchOptions`, and `request` takes per-call options, but nothing resolves a header value at the moment a request goes out. The obvious workaround is to hand `RpcService` a wrapped `fetch` that mutates the options on the way through, and that does work, but it pays for a header with a transport substitution: a caller who supplies their own `fetch` silently gets a different function than the one they passed, and the wrapper has to make assumptions about the shape of `init.headers` from outside the class that owns it.

This PR adds the missing capability at the HTTP boundary instead, then uses it.

### 1. A dynamic header hook on `RpcService`

```ts
getRequestHeaders?: () => Promise<Record<string, string> | undefined>;
```

Called immediately before each `fetch`, inside the Cockatiel policy, so every retry attempt resolves headers again. Precedence is `getRequestHeaders` > `request` options > `fetchOptions` > defaults. A rejection fails the request attempt; resolve `undefined` to proceed without the headers.

Two deliberate details:

- **Resolution happens after the `REQUEST INITIATED` debug log**, which already receives the full fetch options. Anything this hook returns is a good candidate for being sensitive, so injecting afterwards keeps it out of loglevel output.
- **Merging uses the same `deepmerge` the class already uses** to fold in the Basic credential derived from URL userinfo. No second assumption about header shape, and no assumption made from outside the class.

### 2. `NetworkController.getInfuraAuthToken`

```ts
getInfuraAuthToken?: () => Promise<string | undefined>;
```

Core owns the gating, the client owns the token.

Core attaches the credential only when `configuration.type === 'infura'` and the endpoint is not a failover. That is the security-critical half and it keys on the client type and the failover flag, not on the URL, so it cannot be fooled by an endpoint that merely looks like Infura. A custom endpoint pointing at Infura with somebody else's key gets nothing, failover endpoints get nothing, and under `forced` failover mode there is no Infura endpoint in the chain so nothing is issued at all.

The client supplies where the token comes from, which keeps the auth controller and the consent policy out of `network-controller`. No new messenger action, no new dependency, nothing breaking. Clients that pass nothing behave exactly as before.

Because an `RpcService` is bound to a single `endpointUrl` fixed at construction, a header set for one endpoint cannot reach another. This is also why the hook belongs on the constructor rather than on `request`: `RpcServiceChain.request` forwards one options object to *every* service in the chain, so a credential passed that way would also be sent to the failovers.

### Client usage

```ts
new NetworkController({
  state,
  messenger,
  infuraProjectId,
  getInfuraAuthToken: async () => {
    try {
      // The token identifies the user, so only send it with consent.
      if (!analyticsController.state.optedIn) {
        return undefined;
      }
      return await authenticationController.getBearerToken();
    } catch {
      // Throws while the wallet is locked. Proceed without the credential.
      return undefined;
    }
  },
});
```

Resolve the token inside the callback, never in a closure around it. `getInfuraAuthToken` is invoked per request; anything read at construction is stale for the life of the client.

## References

N/A

## Checklist

- [x] I've prepared solutions for platform tests, or opened PRs on [platform repositories](https://github.com/MetaMask/contributor-docs/blob/main/docs/how-to-consume-core-changes.md) to handle them


cryptodev-2s:
@metamaskbot publish-preview

PR #10260, opened 2026-09-16
fix(assets-controller): skip no-op state assignments in updateState

juanmigdr:
## Explanation

`AssetsController#updateState` writes `assetsInfo`, `assetsBalance`, and `assetsPrice` on every response from a data source (RPC poll, websocket push, Accounts API), unconditionally reassigning each key even when the incoming value is identical to what's already in state.

`this.update()` is backed by Immer, which emits a patch — and publishes `stateChange` — whenever a draft key is assigned a different object reference, regardless of whether the new value is deeply equal to the old one. Since data sources re-report unchanged metadata, balances, and prices on virtually every poll (chains/accounts idle far more often than they change), this meant the controller published a no-op `stateChange` and triggered a full state persist on nearly every poll tick, not just when something actually moved.

This surfaced while investigating wallet-unlock JS-thread contention on mobile: with an idle wallet sitting on the home screen, `AssetsController` was persisting its entire state roughly every poll cycle, competing for the JS thread with everything else that reacts to `stateChange`/persistence.

The fix adds `assignIfChanged()`, a small module-level helper alongside the existing `mergeAccountBalances()`/`normalizeResponse()` helpers, that only assigns `target[key]` when it differs from the value already in state (before this update began), so that the existing `changedMetadata`/`changedBalances` tracking (used for logging and downstream events like `balanceChanged`) stays accurate. It's used for all three writes in `#updateState`.

## References

* Fixes ASSETS-3978: https://consensyssoftware.atlassian.net/browse/ASSETS-3978

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core state-merge paths for balances, metadata, and prices on every data-source update; incorrect equality checks could skip legitimate writes or leave event tracking out of sync, though behavior is covered by new tests.
> 
> **Overview**
> Stops **AssetsController** from treating repeated poll data as state updates when nothing actually changed. In `#updateState`, writes to **`assetsInfo`**, **`assetsBalance`**, and **`assetsPrice`** now happen only when the new value is **deeply unequal** to what was already in state (using `lodash` `isEqual`), so Immer does not emit a patch—and **`AssetsController:stateChange`** plus full persistence—on every unchanged RPC/websocket/API tick.
> 
> For metadata, change detection runs against the **merged** value (including the existing “preserve symbol/name when the API returns empty fields” path), so no-op repeats are skipped without losing that merge behavior. Account balances are assigned only when the merged **`effective`** map differs from the prior account slice; prices are updated per asset only when that price entry changed.
> 
> Tests cover both sides: no **`stateChange`** when the same balance, metadata, or price is re-reported, and **`stateChange`** when any of those genuinely change. The package changelog documents the fix under **Unreleased → Fixed**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34ebf946ca125fb54c78857def287ce120e9fcb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/src/AssetsController.ts
```diff
@@ -637,6 +637,38 @@ function mergeAccountBalances(
...
+  target: Record<string, Value>,
+  key: string,
+  previousValue: Value | undefined,
+  nextValue: Value,
+): boolean {
+  if (isEqual(previousValue, nextValue)) {
+    return false;
+  }
+  target[key] = nextValue;
+  return true;
```
Prithpal-Sooriya: Hmm lets discuss this.

1. Do we have measurements on what this impact for this might be?
    - Should we eat the deep-eq cost on every balance update?

2. Where does immer fail for us? Maybe we should try to leverage immer internally for our nested objects?

juanmigdr: I don't have an isolated benchmark, but from the mobile investigation that motivated this, ~90% of updateState calls on an idle wallet were reassigning already-identical data, each one still triggering a full Immer patch and persist, so isEqual on these shallow maps is cheap relative to what it's replacing. Immer isn't failing us here, it's just diffing by reference like it's supposed to. It can't know two different objects have the same values without a check like this one; the real fix would be making mergeAccountBalances itself return the same reference when nothing changed, but that felt like a bigger refactor for the same outcome, so I went with the narrower guard.

juanmigdr: If you are referring to the isEqual cost, it's a generic deep-equality check that scales with the number of assets, so it's slower than a hand-rolled comparator for this specific shape. But it's still cheap relative to what it replaces: the write it skips triggers an Immer patch, a stateChange fanout, and a full JSON.stringify + disk persist of the whole controller state, which touches way more data than the single balance map we're comparing. So we are paying a real cost on every update, just a smaller one than the write we avoid most of the time.

Prithpal-Sooriya (approved):

PR #10263, opened 2026-09-16
fix(multichain-account-service): fix `:removeMultichainAccountWallet` + use `deleteAccounts` on providers

ccharly:
## Explanation

The EVM provider rely on the `HdKeyring` which only allow to remove the last account. Which means we have to go from last-to-first in that case.

Instead of sorting for each providers, we now introduce a `deleteAccounts` which let the provider implement it according to its own constraints.

Ideally, we would want to also have a proper `deleteAccounts` on the keyring API V2, so that Snaps could also batch account deletion, but that's beyond the scope of this PR.

## References

N/A

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes wallet removal semantics and EVM HD keyring deletion ordering, which can leave wallets registered or strand accounts if callers assumed the old always-remove behavior.
> 
> **Overview**
> Fixes multichain wallet removal by moving cleanup off per-provider `deleteAccount` loops and onto a new **`deleteAccounts`** provider API (via `@metamask/account-api` **2.1.0**), so each provider can enforce its own deletion rules—especially **EVM HD last-account-first** under one keyring lock.
> 
> **`MultichainAccountWallet.deleteAllMultichainAccountGroups`** now owns the full teardown: it snapshots group/account IDs (including optional **`getAccountIds(provider)`** filtering), deletes EVM in batch, then best-effort non-EVM cleanup. **`removeMultichainAccountWallet`** only unregisters the wallet when that succeeds; **partial EVM failure throws**, keeps the wallet, prunes emptied groups, and removes non-EVM accounts only where the paired EVM account was already deleted. Sentry failure payloads use **`id`** instead of **`accountId`**.
> 
> Providers gain a default best-effort **`deleteAccounts`** result type; **`EvmAccountProvider`** implements ordered batch deletion; **`AccountProviderWrapper`** forwards batches even when disabled. **`alignAccounts`** re-checks alignment under the wallet mutex after concurrent deletes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75eccb819ebe7cd6c38e52e9bcfd9b441dc844b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -625,9 +642,9 @@ export class MultichainAccountService {
       // failure in `context`. The shape is pinned by
       // `RemoveMultichainAccountWalletFailureContext`.
       const context: RemoveMultichainAccountWalletFailureContext = {
-        failures: failures.map(({ provider, accountId, error }) => ({
+        failures: failures.map(({ provider, id, error }) => ({
           provider,
-          accountId,
```
ccharly, line 630: Decided to align with the new type too, that's fine since it's only used on Sentry, so not a breaking change here

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -453,4 +458,86 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+      Bip44Account<KeyringAccount>[]
+    >();
+
+    for (const id of ids) {
+      try {
+        const account = this.getAccount(id);
+        const entropySource = account.options.entropy.id;
+        const group = byEntropy.get(entropySource) ?? [];
+        group.push(account);
+        byEntropy.set(entropySource, group);
```
gantunesr: I think for now we should only allow deleting accounts coming from the same entropy, wdyt? We could probably omit some logic if we can ensure this assumption

ccharly: Yep, right, that would make things simpler I agree, I'll change that :)

ccharly: https://github.com/MetaMask/core/pull/10263/commits/81c5306c9fcaa39c0df5e65c97ab8ac5fb28b593

packages/multichain-account-service/src/MultichainAccountService.ts
```diff
@@ -562,83 +561,23 @@ export class MultichainAccountService {
...
-      };
-      reportError(
-        this.#messenger,
-        `Failed to delete one or more accounts during wallet removal`,
-        new Error('Wallet removal partially failed'),
-        context,
-      );
-    }
-
+    await wallet.deleteAllMultichainAccountGroups();
```
ccharly, line 565: Decided to move that to the `wallet` itself. This way it's easier for the wallet instance to mutate its internal map of groups.

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -453,4 +458,92 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+    const failures: DeleteAccountsFailure[] = [];
+    const accounts: Bip44Account<KeyringAccount>[] = [];
+
+    for (const id of ids) {
+      try {
+        accounts.push(this.getAccount(id));
+      } catch (error) {
+        failures.push({ id, error });
+      }
+    }
```
hmalik88: You can use `this.getAccounts` here

ccharly: Problem is, if somehow the `id` is missing from the `AccountsController`, it would be just `undefined`, while using `getAccount` here would throw (meaning it would endup in the `failures`).

Another alternative would be to treat `undefined` as failure "explicitly" too, but that'd be a bit more code, not sure how better that it 🤔

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -143,6 +166,184 @@ export class MultichainAccountWallet<
...
+   * drop the wallet.
+   *
+   * NOTE: This operation WILL lock the wallet's mutex.
+   *
+   * @throws If one or more EVM accounts cannot be deleted.
+   * @returns A promise that resolves once all groups cleanup is complete.
+   */
+  async deleteAllMultichainAccountGroups(): Promise<void> {
+    return await this.#withLock(
+      // TODO: Add a proper status for account deletion in `MultichainAccountWalletStatus`.
```
hmalik88: Can we do this now?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -453,4 +458,92 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+
+    const [first] = accounts;
+    if (!first) {
+      // Nothing could be resolved, so there is no keyring to lock.
+      return toDeleteAccountsResult(failures);
+    }
+
+    // A single lock only guarantees the deletion order for one keyring, so
+    // callers must group by entropy source themselves.
```
hmalik88: I'm not sure why this is required? Mind explaining a bit more?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
```diff
@@ -453,4 +458,92 @@ export class EvmAccountProvider extends BaseBip44AccountProvider {
...
+    const entropySource = first.options.entropy.id;
+    assert(
+      accounts.every((account) => account.options.entropy.id === entropySource),
+      'Expected all accounts to be owned by the same entropy source',
+    );
+
+    // We need to sort accounts by descending group index to ensure that the
+    // last account in the HD keyring is deleted first.
```
hmalik88, line 519: Why do we want to ensure that the last account in the keyring is deleted first?

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Let me try to speedrun that! 😄
- https://github.com/MetaMask/accounts/pull/632

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: I'd prefer the extra bit of code then `n` round trips haha 😛

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: I think the phrasing is limited but, we can only lock the `KeyringController`'s mutex for 1 keyring ID at a time.

Unless we use `:withController` which let us combine multiple operations.

The simplification of using only 1 entropy source was suggested here btw:
- https://github.com/MetaMask/core/pull/10263#discussion_r4030421925

And IMO that makes more sense "semantically speaking".

Do you want me to rephrase this to mention single lock with `:withKeyringV2` instead?

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: That's an old "requirement" we had on the HD keyring. The idea was, we could always remove the last accounts so the "internal index" (when creating the next account with the old `addAccounts`) could use the highest index among its list of accounts and do `+1`.

If accounts were removed out-of-order, there was no real way to fill the gaps.

So that's mostly an "old constraint" that we never changed (despite having everything we need on the keyring API v2 🙈)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: Those roundtrips should be cheap though 😄  but fair enough, I can change that, np! 😉

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: Yes please :)

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
hmalik88: Gotcha!

packages/multichain-account-service/src/MultichainAccountWallet.ts
hmalik88: Approved!

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: https://github.com/MetaMask/core/pull/10263/commits/a9f55deb1a42c7c9440b9474da834bbec7f4176d

packages/multichain-account-service/src/providers/EvmAccountProvider.ts
ccharly: https://github.com/MetaMask/core/pull/10263/commits/07f2263d13126763a9fe397644f29edcc21fe810

packages/multichain-account-service/src/MultichainAccountWallet.ts
ccharly: Released in:

- https://github.com/MetaMask/accounts/pull/633

hmalik88 (dismissed):

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -143,6 +166,183 @@ export class MultichainAccountWallet<
...
+
+      if (failures.length > 0) {
+        // That's not a hard failure, non-EVM are best effort, but we still report something
+        // went wrong, to further investigate.
+        const error = new Error(
+          `Failed to delete one or more accounts during wallet removal`,
+        );
+        reportFailures(error.message, error);
+      }
+    }
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/10263/commits/4a69bf205e5ec4bacc0974fbdcaba6602ab347b3

packages/multichain-account-service/src/MultichainAccountWallet.ts
```diff
@@ -143,6 +166,183 @@ export class MultichainAccountWallet<
...
+      // was removed, then stop reserving those now-orphaned group indexes.
+      const evmRemovedGroupIndexes = new Set(
+        [...evmGroups.entries()]
+          .filter(
+            ([_, evmGroupAccountIds]) =>
+              // This group has no remaining EVM accounts, we keep it.
+              !evmGroupAccountIds.some((id) => evmRemainingAccountIds.has(id)),
+          )
+          .map(([groupIndex, _]) => groupIndex),
+      );
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/core/pull/10263/commits/75eccb819ebe7cd6c38e52e9bcfd9b441dc844b1

hmalik88 (approved):

Mrtenz (approved):
Dependency bumps and changelogs look good to me. Haven't reviewed the actual implementation.

PR #10265, opened 2026-09-16
feat(profile-sync): add MFA authentication services

mathieuartu:
## Explanation

Wires the MFA HTTP client into the SRP JWT auth SDK on top of the validation foundation.

- `mfa/services` calls `/api/v2/mfa/*` (enroll, enroll complete, verify, verify complete, credentials), validates request/response bodies, maps server codes to `MfaError` subclasses, and handles OTP cooldown / `Retry-After`.
- `SRPJwtBearerAuth` exposes begin/complete enrollment and verification, credential listing, and assertion-to-token exchange.
- Public `JwtBearerAuth` forwards those methods and rejects non-SRP auth types.
- Nock fixtures and unit tests cover happy paths and error mapping.

Clients still go through `AuthenticationController` in later PRs; this layer is not UI-facing.

## References

Depends on #10264. Follow-ups: #10266, #10267. Stack: [stack #10268](https://github.com/MetaMask/core/pull/10268).
Related to: https://consensyssoftware.atlassian.net/browse/MUL-2261

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **High Risk**
> Touches authentication, step-up tokens, and passkey/OTP handling; mistakes could weaken session elevation or mishandle credentials, though coverage is extensive.
> 
> **Overview**
> Adds **passkey and email OTP MFA** to the profile-sync JWT auth SDK: a new HTTP layer calls `/api/v2/mfa/*` (enroll, complete, verify, credentials), validates payloads, maps server codes to `MfaError` subclasses, and handles OTP cooldown plus `Retry-After`.
> 
> **`SRPJwtBearerAuth`** and public **`JwtBearerAuth`** expose enrollment/step-up flows, credential listing, and **`exchangeMfaAssertion`** (AAL2 JWT → elevated access token via existing OIDC). MFA is **SRP-only**; email enrollment uses a separate `email` option from `entropySourceId`.
> 
[...]

PR #10266, opened 2026-09-16
feat(profile-sync): add MFA credential enrollment

mathieuartu:
## Explanation

Exposes UI-driven MFA enrollment on `AuthenticationController` (no ApprovalController).

- In-memory `enrolledCredentials` cache (optional on the state type so partial-state selectors stay assignable) with redacted emails in state logs.
- `refreshEnrolledCredentials`, `beginCredentialEnrollment`, `completeCredentialEnrollment`.
- `AuthenticationController:credentialsChanged` when the cache actually changes.
- Automatic credential refresh after sign-in and unlock when MFA is enabled; refresh failure does not fail sign-in.
- Cache cleared (and the change event published) on lock, sign-out, and wallet reset.
- Regenerated messenger action types; README and changelog updates.

Enrollment is UI-orchestrated: the controller returns challenges; the client collects the passkey or OTP proof and calls complete.

## References

Depends on #10265. Follow-up: #10267. Stack: [stack #10268](https://github.com/MetaMask/core/pull/10268).
Related to: https://consensyssoftware.atlassian.net/browse/MUL-2262

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes authentication session lifecycle and MFA enrollment, including SRP token invalidation after email enroll and race handling around lock/sign-out—security-sensitive behavior that clients must wire correctly.
> 
> **Overview**
> Adds **UI-orchestrated MFA enrollment** on `AuthenticationController`: `refreshEnrolledCredentials`, `beginCredentialEnrollment`, and `completeCredentialEnrollment`, wired to the existing JWT bearer MFA SDK and new messenger action types.
> 
> Enrollment state lives in a **non-persisted** `enrolledCredentials` cache (UI-visible; state logs allow-list type/status/enrolledAt only). An optional **`trace`** constructor hook (`@metamask/controller-utils`) wraps MFA network steps with operation/credential tags and outcomes.
> 
> **Session and concurrency guards** bump an auth-session epoch on lock, sign-out, and wallet reset so in-flight MFA work cannot apply after the session ends; credential refreshes are sequenced so slower responses cannot overwrite newer lists. **Email OTP completion** invalidates the primary SRP session after a post-enroll refresh (or even when the session ends mid-flow) so cached tokens are not reused without the new email claim; passkey enroll does not touch the session. Sign-out/`clearState` clear the credential cache.
> 
> Also adds MFA nock fixtures and mock response helpers, depends on `@metamask/controller-utils`, and tightens `ElevatedTokenClaimsStruct` as a single sensitive struct.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d70d671c45fdc45bc04254aff3d740702f8adce9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -129,6 +147,28 @@ const metadata: StateMetadata<AuthenticationControllerState> = {
...
+        }
+        const separatorIndex = credential.email.indexOf('@');
+        const domain =
+          separatorIndex >= 0 ? credential.email.slice(separatorIndex) : '';
+        return {
+          ...credential,
+          email: `${credential.email.slice(0, 1)}••${domain}`,
+        };
+      });
+    },
```
ccharly: I feel like we could mistakenly have new credentials in the state logs if we forget about this implementation.

Do we really want them here? We could just go with `includeInStateLogs: false`

If we really want to keep the `email_otp` ones, then maybe we should use `if (credential.type === `email_otp`)` instead, so we apply the custom logic ONLY for this one. And for all others we either return `null` or something less sensitive? (no data, no PII)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -343,6 +406,14 @@ export class AuthenticationController extends BaseController<
     }
   }
 
+  #assertAuthSessionEpoch(epoch: number, methodName: string): void {
+    if (!this.#isUnlocked || this.#authSessionEpoch !== epoch) {
+      throw new Error(
+        `${methodName} - unable to proceed, the authenticated session ended`,
```
ccharly, line 402: Question: Do we want to use `!==` or `<` here? Since we mention "ended". Just trying to make sure we don't consider more recent sessions (> a certain epoch) as invalid

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -727,7 +798,185 @@ export class AuthenticationController extends BaseController<
...
+      async () => {
+        try {
+          const result = await fn();
+          data.outcome = 'success';
+          return result;
+        } catch (error) {
+          data.outcome = 'error';
+          const mfaCode = getMfaErrorCode(error);
+          if (mfaCode) {
+            data.mfaCode = mfaCode;
```
ccharly: Nit: Maybe `mfaErrorCode` instead? (I was a bit scared at first thinking we were dumping some factor code into our tracing system 😅)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -727,7 +798,185 @@ export class AuthenticationController extends BaseController<
...
+      'MFA Credentials Refresh',
+      'credentials.refresh',
+      'all',
+      async () => await this.#auth.getMfaCredentials(primaryEntropySourceId),
+    );
+    this.#assertAuthSessionEpoch(sessionEpoch, 'refreshEnrolledCredentials');
+
+    // A newer refresh started while this one was in flight; its result is at
+    // least as fresh, so defer to it rather than overwrite with older data.
+    if (refreshSeq !== this.#credentialsRefreshSeq) {
```
ccharly, line 853: Same remark than for `#authSessionEpoch` but this one is probably ok though 🤔

Since those counters are monotonic, we should never really have an issue using `!==` but worth double-checking I guess 😅

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -727,7 +798,185 @@ export class AuthenticationController extends BaseController<
...
+      return this.state.enrolledCredentials ?? [];
```
ccharly: Very nit (cosmetic): What do you think about having `#getEnrolledCredentials` that would return this? So we avoid repeating the pattern every time. Like we could even use this in `#clearEnrolledCredentials`.

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Good question! They're equivalent here since the epoch only ever increments. A newer epoch means the session this request started in has ended, which is exactly what we want to catch. I kept `!==` because it's really an identity check ("am I still in the same session?") rather than an ordering one. Happy to add a "monotonic" note to the JSDoc if that helps 🙂

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Same here 😄

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: [d70d671](https://github.com/MetaMask/core/pull/10266/commits/d70d671c45fdc45bc04254aff3d740702f8adce9)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: [d70d671](https://github.com/MetaMask/core/pull/10266/commits/d70d671c45fdc45bc04254aff3d740702f8adce9)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: [d70d671](https://github.com/MetaMask/core/pull/10266/commits/d70d671c45fdc45bc04254aff3d740702f8adce9)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
ccharly: I guess that's fine :) I was worried about potential race where things could get re-ordered, but as you said, that's an identity check and we always capture the current value before incrementing it.

We can always add it to the jsdoc on the final PR, but that's fine IMO!

ccharly (approved):
LGTM (went quickly over tests, relying on coverage threshold here)

PR #10267, opened 2026-09-16
feat(profile-sync): add MFA step-up sessions

mathieuartu:
## Explanation

Adds MFA step-up verification and a memory-only elevated AAL2 session on `AuthenticationController`.

- `beginStepUp`, `completeStepUp`, `getElevatedProfileToken`, `clearStepUpSession`.
- In-memory elevated token with TTL; `AuthenticationController:stepUpSession` on open/close.
- Session is torn down on lock, sign-out, wallet reset, successful enrollment, and when `getElevatedProfileToken` observes expiry.
- Completing enrollment or step-up re-asserts the wallet is still unlocked after network calls so a lock during the request cannot reopen an AAL2 session.
- MFA network steps are traced with `operation` / `credentialType` tags and span attributes for outcome / `mfaCode`.

## References

Depends on #10266. Stack: [stack #10268](https://github.com/MetaMask/core/pull/10268).
Related to: https://consensyssoftware.atlassian.net/browse/MUL-2263

## Checklist

---

> [!NOTE]
> **High Risk**
> Changes authentication and elevated-token handling for sensitive operations; incorrect session TTL or teardown logic could block legitimate actions or briefly retain AAL2 access after lock/sign-out.
> 
> **Overview**
> Adds **MFA step-up verification** on `AuthenticationController`: `beginStepUp` / `completeStepUp` run passkey or email OTP ceremonies, exchange the MFA assertion for an **AAL2 elevated profile token**, and expose it via `getElevatedProfileToken` while a short-lived session is open. The token stays **memory-only**; persisted/UI state is limited to non-secret `stepUpSessionExpiresAt` (60s TTL clamped to JWT `exp`).
> 
> **Session lifecycle** clears elevated access on wallet lock, sign-out, wallet reset, successful credential enrollment, base-session `authentication_required`, and hard expiry (timer or lazy check in `getElevatedProfileToken`). Callers can require fresher proof with `maxSessionAgeMs` without tearing down the session for others.
> 
> **Hardening** reuses auth-session epoch checks so lock/sign-out during in-flight MFA cannot apply results; invalid exchanged tokens fail with `elevated_token_invalid`. MFA network tracing now sets span **attributes** (`outcome`, `mfaErrorCode`) via the trace callback context instead of mutating shared trace data. A 401 on credential refresh also invalidates the SRP session and drops any step-up session.
> 
> Docs/changelog and messenger action types are updated; tests cover step-up flows, TTL edge cases, and enrollment lock races.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f0582fdba75005886855828948348af81ec5aea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -805,30 +855,51 @@ export class AuthenticationController extends BaseController<
...
           return result;
         } catch (error) {
-          data.outcome = 'error';
+          this.#setTraceAttribute(context, 'outcome', 'error');
           const mfaCode = getMfaErrorCode(error);
           if (mfaCode) {
-            data.mfaErrorCode = mfaCode;
+            this.#setTraceAttribute(context, 'mfaErrorCode', mfaCode);
+            if (mfaCode === 'authentication_required' && this.#isUnlocked) {
+              this.#invalidateSrpSession(this.#getPrimaryEntropySourceId());
```
ccharly, line 867: If we invalidate the main session, should we also invalidate the elevated one? Maybe that's already implicit, but I was wondering if `#invalidateSrpSession` would naturally invalidate it too?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -939,6 +1002,7 @@ export class AuthenticationController extends BaseController<
       }
       throw error;
     }
+    this.clearStepUpSession();
```
ccharly, line 1009: Are we clearing this step-up session to force the refresh since a new "credential" has been registered? If yes, can you add a comment to explain this?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -951,6 +1015,159 @@ export class AuthenticationController extends BaseController<
...
+      request.reason.operation,
+      request.type,
+      async () =>
+        await this.#auth.beginMfaVerification(
+          request.type,
+          primaryEntropySourceId,
+        ),
+    );
+    this.#assertAuthSessionEpoch(sessionEpoch, 'beginStepUp');
+    return challenge;
```
ccharly, line 1046: Question (cause I'm not 100% sure): the credential enrollment does not require an elevated token, but here, we still the `#authSessionEpoch`, which means that if we:
1. initiate a credential enrollment with a "normal token"
2. initiate a step-up session slightly after

Then the session from 1 could get invalidated even though there are technically 2 different "kind" of sessions (normal vs elevated).

I guess, for simplicity, we don't allow 2 tokens here, we just expose 1 (normal or elevated depending on if we started the step-up dance flow).

Am I right? We're ok with potentially invalidating in-flight sessions "dances" (credential in this case) for now?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -951,6 +1015,159 @@ export class AuthenticationController extends BaseController<
...
+      type,
+      async () => await this.#auth.exchangeMfaAssertion(assertion.token),
+    );
+    this.#assertAuthSessionEpoch(sessionEpoch, 'completeStepUp');
+
+    let decodedClaims: unknown;
+    try {
+      decodedClaims = decodeJwtPayload(accessToken.accessToken);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
```
ccharly: We have a new util for this

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -951,6 +1015,159 @@ export class AuthenticationController extends BaseController<
...
+    const now = Date.now();
+    if (now >= session.expiresAt) {
+      // The hard-expiry timer has not fired yet (e.g. a suspended tab).
+      this.clearStepUpSession();
+      return null;
+    }
+    // `>=` so a zero max age always forces a fresh ceremony.
+    const maxAge = request.maxSessionAgeMs;
+    if (maxAge !== undefined && now - session.token.obtainedAt >= maxAge) {
+      return null;
```
ccharly, line 1130: Shouldn't we clear the session here?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -951,6 +1015,159 @@ export class AuthenticationController extends BaseController<
...
+    try {
+      decodedClaims = decodeJwtPayload(accessToken.accessToken);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      throw new ElevatedTokenInvalidError(message);
+    }
+    const claims = parseElevatedTokenClaims(decodedClaims);
+    if (claims.exp * 1000 <= Date.now()) {
+      throw new ElevatedTokenInvalidError('Elevated token is expired');
+    }
```
ccharly, line 1097: Should we clamp here too? And re-use `STEP_UP_SESSION_TTL_MS`? So we also consider it expired before reaching `#openStepUpSession`?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
```diff
@@ -951,6 +1015,159 @@ export class AuthenticationController extends BaseController<
     }
   }
 
+  /**
+   * Begins step-up verification with an enrolled credential.
+   *
+   * @param request - Credential type and trace reason.
+   * @returns A challenge for the client-owned ceremony.
+   */
+  public async beginStepUp(
```
ccharly, line 1028: Nit: But since we use `clearStepUpSession` we could use `{being,complete}StepUpSession` naming too?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Good catch, I'll do that in the next commit!

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: The epoch is not bumped during the flow you described, it's only bumped on `lock / sign-out / reset`.
Both flows use the same AAL1 bearer for their API calls, and the elevated token is a separate one that only lives in #stepUpSession. So an enrollment in flight survives a step-up starting next to it

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Riiihgt, forgot to bump the last stack! Thanks!

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: Deliberately not: `maxSessionAgeMs` is this caller's freshness requirement, not a global invalidation. If we cleared, one strict caller (e.g. `maxSessionAgeMs: 0`) would destroy a session other callers can still legitimately use within the TTL. Hard expiry is the TTL/exp timer's job. Will add a comment so it's more clear 😄

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: The clamp would be a noop here I think. For a freshly exchanged token `obtainedAt ≈ now`, so `obtainedAt + TTL` is always in the future. The only way `min(…)` could be past is `exp` already being past, which is exactly what this check catches

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
mathieuartu: it's intentional: you begin/complete a step-up (the verification ceremony), and completing it opens a session, which you then clear. `beginStepUpSession` would suggest a challenge starts a session. WDYT?

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
ccharly: Ah true, I missed the part it was not bumped 🤦‍♂️ since I saw a similar pattern in other PRs I just assumed we were bumping it here too... Sorry for the noise, all good then 💪

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
ccharly: Ah yes, makes sense, the `maxSessionAge` is tied to the request here, so we definitely don't want to clear up the main elevated that might still be valid 👍

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
ccharly: Yep, that makes sense, otherwise we would need another suffix to avoid thinking this is a "transactional flow" (e.g you start, you do your thing, then you end it)

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.ts
ccharly: I think yes too. This could be an issue if we were getting served an old token (which is unlikely since we have the combo befin/complete with a brand new challenge everytime 😄).

So that should not be possible to hit indeed.

Anyway, we still have the timer in `#openStepUpSession` that would clear up the session automatically in any case!

We're good. Thanks 😄

packages/profile-sync-controller/src/controllers/authentication/AuthenticationController.test.ts
```diff
@@ -2780,6 +2846,495 @@ describe('MFA credential enrollment', () => {
...
+      controller.getElevatedProfileToken({ maxSessionAgeMs: 0 }),
+    ).toBeNull();
+    expect(controller.getElevatedProfileToken()).not.toBeNull();
+    expect(
+      controller.getElevatedProfileToken({ maxSessionAgeMs: 600_000 }),
+    ).not.toBeNull();
+    expect(() =>
+      controller.getElevatedProfileToken({ maxSessionAgeMs: -1 }),
+    ).toThrow(/MFA\[invalid_request\]/u);
+  });
```
ccharly, line 3067: Nit: Should we have a test to also ensure the session is not clear when the request ends up as "invalid request"? e.g. we're never clearing the token in that case?

ccharly (dismissed):

ccharly (approved):

PR #10298, opened 2026-09-18
feat: pilot Renovate with dashboard approval

cryptodev-2s:
## Explanation

Pilots Renovate on this repo, self-hosted, to answer the question the Dependabot allowlist decision has been waiting on: how much work is actually behind those 219 external dependencies currently in use in Core.

Nothing is opened automatically. `dependencyDashboardApproval` means Renovate writes a single Dependency Dashboard issue listing every pending update and then stops; a pull request appears only when someone ticks a checkbox. So this can run alongside Dependabot without duplicate pull requests and without turning anything off.

Renovate is `AGPL-3.0-only`. Running it in our own CI is not distribution, but worth flagging before this goes beyond a pilot.

## Blocked on

1. The `renovate` environment being created in this repo, with no branch policy and no protection rules.
2. consensys-vertical-apps/token-exchange-service#180.

## Not covered yet

Renovate handles version ranges, the lockfile and deduplication on its own, so of the three repairs the Dependabot workflow performs only changelog entries remain. That is deliberately left until the first approved update, when we can see what else it gets right or wrong. The fix is either `postUpgradeTasks` running the existing fixer, or widening the `dependabot[bot]` gate on `repair-dependabot-pull-requests.yml`.

## References

Part of WPC-1161.

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate (workflow and config only)
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/core/tree/main/docs/processes/updating-changelogs.md) (no published package is touched)
- [ ] I've introduced [breaking changes](https://github.com/MetaMask/core/tree/main/docs/processes/breaking-changes.md) in this PR and have prepared draft pull requests for clients and consumer packages to resolve them

---

> [!NOTE]
> **Medium Risk**
> The workflow can create PRs and modify repo content once the renovate environment and token exchange are configured; misconfiguration could broaden write scope beyond intended dependency updates.
> 
> **Overview**
> Adds a **self-hosted Renovate pilot** alongside existing Dependabot: a daily (06:00 UTC) and manually triggerable GitHub Actions workflow that exchanges an OIDC token for scoped GitHub access and runs `renovatebot/github-action` against this repository only.
> 
[...]

PR #10311, opened 2026-09-21
fix: let Renovate link node_modules when updating lockfiles

cryptodev-2s:
## Explanation

Approving an update on the Dependency Dashboard (#10300) produced no pull request. Renovate built the branch, hit a `lock file error`, and reported `No files to commit`. Underneath that:

```
➤ YN0000: ┌ Link step
➤ YN0073: │ Skipped due to mode=update-lockfile
➤ YN0000: └ Completed
Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)
$ yarn run [--inspect] ... <scriptName>
```

Renovate runs `yarn install --mode=update-lockfile`, which skips the Link step, so `node_modules` and its install state file are never written. Our `.yarnrc.yml` loads LavaMoat's [`plugin-allow-scripts`](https://github.com/LavaMoat/LavaMoat/tree/main/packages/yarn-plugin-allow-scripts), whose `afterAllInstalled` hook runs `yarn run allow-scripts`. `yarn run` needs that state file to resolve the binary, so it exits non zero, and Renovate reads that as a lock file error and throws the updated lockfile away.

Isolated it by running `yarn install --mode=update-lockfile` cold on `main` with each piece removed in turn:

| `.yarnrc.yml` plugin | root `postinstall` | Result |
| --- | --- | --- |
| present | present | fails |
| present | removed | fails |
| removed | present | exit 0 |

The third row also shows Yarn does not run root lifecycle scripts in this mode at all, so the root `postinstall` is not involved.

Setting `skipInstalls: false` makes Renovate use `yarn install --mode=skip-build` instead ([source](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/npm/post-update/yarn.ts#L166-L207)): the Link step runs, so the state file exists and `allow-scripts` succeeds, while dependency build scripts stay skipped. Cold run on `main`, exit 0 in 26 seconds.

This also reverts #10303. The `! Corepack is about to download` line is a notice on stderr, not a blocking prompt, and was never the cause.

## References

Reverts #10303. Follows #10298. Part of WPC-1161.

## Checklist


renovate.json
```diff
@@ -5,6 +5,7 @@
   "dependencyDashboardApproval": true,
   "minimumReleaseAge": "3 days",
   "postUpdateOptions": ["yarnDedupeHighest"],
+  "skipInstalls": false,
```
Mrtenz, line 8: Can't we just disable the post-install for Renovate? There's no reason to run it.

Bit hacky, but this is how we do it in the extension:

https://github.com/MetaMask/metamask-extension/blob/4cc1e269a524007ed3f633d0399b48429b91be22/package.json#L41

Mrtenz: Where are we doing this `yarn run postinstall` then? 🤔

cryptodev-2s: we're not, yarn is. it's a root lifecycle script so yarn runs it itself after the Link step, that's where the `yarn run ... <scriptName>` in the error comes from.

and it never gets to the script body anyway. swapped it for a bare `node -e "console.log('hi')"` and same failure.

Mrtenz: So with `yarn install --mode=update-lockfile`, Yarn still runs `postinstall`, even though it isn't able to? That seems confusing.

cryptodev-2s: nah you were right to doubt it, i had it wrong. yarn doesn't run postinstall in that mode at all.

it's the lavamoat allow-scripts plugin in `.yarnrc.yml`. its `afterAllInstalled` hook shells out to `yarn run allow-scripts`, which needs the node_modules state file update-lockfile never writes. dropping the postinstall changes nothing, dropping the plugin makes it pass.

same fix though, skip-build runs the Link step. PR description updated.

Mrtenz: Right, that makes more sense.

Maybe we should add an environment variable to the allow-scripts plugin (or detect the install mode, if that's possible) so we can skip it when it's not needed instead?

I can approve this for now, but maybe look into that as follow-up.

Mrtenz (approved):

renovate.json
Gudahtt: >Maybe we should add an environment variable to the allow-scripts plugin (or detect the install mode, if that's possible) so we can skip it when it's not needed instead?

I had the same problem when using Renovate on my Prettier plugin, and this is the same idea I landed on 😄. I forgot to mention it, and haven't implemented it yet. Very easy to do though.

PR #10317, opened 2026-09-21
fix: stop Renovate posting commit statuses

cryptodev-2s:
## Explanation

#10316 broke the Renovate job outright. Asking the token exchange for `statuses: write` now fails before Renovate even starts:

```
##[error]Token exchange failed: 500 Internal Server Error
{"status":"error","message":"Failed to create installation token from GitHub"}
```

That is GitHub refusing to mint the token, not the policy rejecting it. consensys-vertical-apps/token-exchange-service#182 allows the permission, but the `metamask-ci` App installation does not hold Commit statuses, and a policy can only narrow what the App already has. Granting it means an org level App permission change that every installation has to accept, which is a lot of process for four informational checks.

So this reverts the permission request and stops Renovate posting the statuses at all. With every entry in `statusCheckNames` set to null, Renovate skips the POST rather than 403ing on it ([`status-checks.ts:91`](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/update/branch/status-checks.ts#L91)), which was what aborted the run before the pull request got created.

What we lose is the checks appearing on the pull request. Artifact and config errors still show up in the run log and on the dependency dashboard. `minimumReleaseAge` is unaffected, it is enforced internally whether or not the status is posted.

## References

Reverts #10316. Follows #10311. Part of WPC-1161.

## Checklist

---

> [!NOTE]
> **Low Risk**
> CI-only changes to Renovate permissions and config with no runtime or application code impact.
> 
> **Overview**
> Fixes the Renovate workflow failing at token exchange by **removing `statuses: write`** from the MetaMask token-exchange permissions in `.github/workflows/renovate.yml`. The `metamask-ci` app installation does not have commit-status scope, so requesting it caused token minting to fail before Renovate could run.
> 
> **Disables Renovate’s GitHub commit statuses** by adding `statusCheckNames` in `renovate.json` with `artifactError`, `configValidation`, `mergeConfidence`, and `minimumReleaseAge` set to `null`, so Renovate skips posting those checks instead of erroring. Dependency updates and internal rules like `minimumReleaseAge` still apply; only the PR-visible status checks are dropped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5288f2c82ffd76ba19ab84e963acde54ce0b8ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mrtenz (approved):
This is fine, but we can also ask TechOps to grant this permission to the app.

PR #10325, opened 2026-09-21
fix(transaction-pay-controller): re-derive max source amounts after balance refresh

jpuri:
## Explanation

`updateQuotes` refreshes the payment token balance before building quote requests, but source amounts were only recalculated when the payment token **address or chain** changed. On a max amount the source amount *is* the payment token balance, so a token selected before its balance had loaded kept a zero source amount even though the refresh had since written the real balance.

The result was a quote request that contradicted itself — the refreshed balance next to a zero input:

```
{ "isMaxAmount": true, "sourceBalanceRaw": "5855729", "sourceTokenAmount": "0", ... }
```

Relay was then asked for an `EXACT_INPUT` of `0`, returned nothing, and the client surfaced "no payment route available" for a route that was in fact fundable. It reproduced on a Money Account deposit by leaving the amount screen and returning, which re-selects the pay token before its balance resolves.

Two coordinated changes fix it:

- `TransactionPayController` now re-derives source amounts when the refreshed balance differs on a non-post-quote max transaction. It deliberately does **not** request quotes again from that branch: the balance is refreshed by `updateQuotes` itself, which consumes the re-derived amounts in the same pass. Post-quote flows are excluded because their source is the required token, not the payment token, so a payment-token balance change is irrelevant there.
- `updateQuotes` reads the current source amounts from controller state after the refresh instead of the snapshot destructured before it. The snapshot is a frozen pre-update copy, so it could never observe the recalculation.

## References

- Jira: [CONF-2063](https://consensyssoftware.atlassian.net/browse/CONF-2063)

## Checklist

[CONF-2063]: https://consensyssoftware.atlassian.net/browse/CONF-2063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes max-amount quote input derivation in Transaction Pay; incorrect behavior would still block fundable routes, but the change is narrow and well covered by tests.
> 
> **Overview**
> Fixes **max-amount** MetaMask Pay flows where the pay token was chosen before its balance loaded, leaving a **zero** source amount even after a live balance refresh—Relay then got an `EXACT_INPUT` of `0` and the UI showed no route.
> 
> **TransactionPayController** now treats a change to `paymentToken.balanceRaw` like a max-amount input update: it calls `updateSourceAmounts` when `isMaxAmount` is set and the flow is not post-quote, but **does not** fire another quote pass (that refresh already happens inside `updateQuotes`). Post-quote and non-max cases are unchanged.
> 
> **updateQuotes** builds quote requests from **current** `sourceAmounts` in controller state (via new `getSourceAmounts`) after `refreshPaymentTokenBalance`, instead of the pre-refresh snapshot passed into the function.
> 
> Tests and an Unreleased changelog entry cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d16c593efb2493d7831bff86f65e770633eada7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/transaction-pay-controller/src/TransactionPayController.ts
```diff
@@ -382,6 +385,22 @@ export class TransactionPayController extends BaseController<
...
+      } else if (
+        isPaymentTokenBalanceUpdated &&
+        current.isMaxAmount &&
+        !current.isPostQuote
+      ) {
+        // On max the source amount is the payment token balance, so a
+        // refreshed balance must re-derive it, otherwise the amount stays at
+        // whatever the balance was when the token was selected. Quotes are not
+        // requested again here: the balance is refreshed by `updateQuotes`
+        // itself, which reads the re-derived amounts before building requests.
```
OGPoyraz, line 397: Not sure I follow here, assuming we don't force `updateQuotes` by `shouldUpdateQuotes = true`, let's say payment token balance is decreased quote will fail if user doesn't wait for auto-quote refresh isn't it? 

Shouldn't we make another case above and force `updateQuote`? 
```ts
///
const isPaymentTokenBalanceUpdatedForMax =  isPaymentTokenBalanceUpdated && current.isMaxAmount && !current.isPostQuote;

 if (
        isPaymentTokenUpdated ||
        isIsMaxUpdated ||
        isTokensUpdated ||
        isPostQuoteUpdated ||
        isAccountOverrideUpdated || 
        isPaymentTokenBalanceUpdatedForMax
      ) {
        updateSourceAmounts(
          transactionId,
          current as never,
          this.messenger,
          this.#getBalance,
        );

        shouldUpdateQuotes = true;
      }
  ```

matthewwalsh0: I understood we're already in a quote update when it refreshes the token balance, so we don't need to trigger another, since one is already in progress.

The source amounts should update synchronously, so `buildQuoteRequests` can use it immediately.

OGPoyraz (dismissed):

OGPoyraz (approved):

PR #8582, opened 2026-04-24
feat: assets controller resilience

salimtb:
## Explanation

This PR makes the `AssetsController` more resilient to upstream failures across three related fronts.

**1. Custom asset graduation (EVM only)**

`customAssets` is meant for tokens that the regular detectors don't yet know about — by design, `RpcDataSource` is the sole balance fetcher for anything in that list. Once `AccountsApiDataSource` or `BackendWebsocketDataSource` reports a balance for an asset that's currently in `customAssets[selectedAccount]`, the asset is no longer "custom" — it's just a regularly-detected token. The new `CustomAssetGraduationMiddleware` removes such assets from `customAssets` so they aren't double-tracked.

The rule applies only to EVM chains; non-EVM custom assets (Solana, BTC, Tron via Snap) are left untouched. Graduation is gated by `sourceId` in the subscription pipeline so RPC pushes (which intentionally carry custom assets) and Snap pushes (non-EVM) never trigger graduation.

**2. RPC fallback for non-custom tokens**

When `AccountsApiDataSource` or `BackendWebsocketDataSource` can't deliver a balance for some chain — network error, `unprocessedNetworks` in the API response, or a timeout — the affected chains end up in `response.errors`. The new `RpcFallbackMiddleware`, inserted right after the parallel balance middleware in the fast pipeline, retries those chains on `RpcDataSource` with the request filtered down to just the failed chains. Successful RPC balances are merged into the response and the recovered chains are cleared from `response.errors`. The middleware is a no-op when nothing failed upstream, so the happy path is unchanged.

**3. 15s fetch timeout on data sources that hit external APIs**

`AccountsApiDataSource`, `PriceDataSource`, and `TokenDataSource` had no timeout on their HTTP calls, so a hung connection could stall an entire balance/enrichment pipeline. Each now exposes a configurable `fetchTimeoutMs` option (default `15000`) and wraps its API calls with a shared `fetchWithTimeout` utility. On timeout:

- AccountsAPI marks every requested chain as errored (same path as a network failure), so `RpcFallbackMiddleware` picks them up.
- Price and token enrichment degrade gracefully — the surrounding pipeline continues with whatever data did arrive.

## References

Fixes #<insert-ticket-number>

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [x] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
[...]

PR #10335, opened 2026-09-21
fix(assets-controller): stop graduating custom assets out of customAssets

Prithpal-Sooriya:
## Description

`CustomAssetGraduationMiddleware` removed an EVM asset from `customAssets` as soon as the Accounts API / Account Activity websocket reported a balance for it ("graduation"). Losing the `customAssets` entry also stripped the token's **exemption from `TokenDataSource` occurrence / Blockaid spam filtering**, so a low-occurrence token that the API had just started indexing was auto-filtered out of balances and metadata — the user lost the very token they had imported:

1. User adds a custom token.
2. Token becomes supported by the API → "graduates" (was removed from `customAssets`).
3. Token enters `TokenDataSource` and gets filtered out (occurrences below floor / Blockaid).
4. User loses their custom token. 😞

**Fix:** remove the graduation middleware entirely. Custom assets are now never removed from `customAssets` automatically, so they keep both their RPC supplemental polling and their spam-filter exemption. The `AssetsController:removeCustomAsset` action is unchanged and remains the way to remove one explicitly.

## Test plan

Test-driven: all new assertions were confirmed **red** against the old behavior first, then green after the fix.

- **Fast-lane integration suite** (`buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts`): extended with the graduation scenario — CDOGE seeded as a custom asset, positive Accounts API balance, `occurrences: 1` (below the BNB Chain floor of 3) → now survives balances, metadata and `detectedAssets` in the pipeline response.
- **Controller integration suite** (`AssetsController.bsc-spam-token-filtering.integration.test.ts`): same scenario end-to-end — the imported token stays in `customAssets`, `assetsBalance` and `assetsInfo` after a full fetch.
- **Unit** (`AssetsController.test.ts`): the "custom asset graduation" block is now "custom asset retention" — custom assets stay put when AccountsApi / AccountActivity / RPC report balances.
- Full `@metamask/assets-controller` suite passes; monorepo `yarn build` clean; ESLint / Prettier clean; `yarn changelog:validate` passes.

## Changelog

- `packages/assets-controller/CHANGELOG.md`: **BREAKING** `Removed` entry — the `CustomAssetGraduationMiddleware` / `CustomAssetGraduationMiddlewareOptions` exports are gone.

---

> [!NOTE]
> **Medium Risk**
> Breaking removal of exported middleware and a behavior change in how custom imports interact with spam filtering and the asset pipeline; well-covered by integration/unit tests but affects wallet-visible token lists.
> 
> **Overview**
> Removes **automatic “graduation”** of user-imported tokens: **`CustomAssetGraduationMiddleware` is deleted** and no longer runs in the fast fetch lane or on Accounts API / Account Activity updates. Entries in **`customAssets` stay put** even when upstream balance sources report a positive balance, so those tokens **keep their spam-filter exemption** in `TokenDataSource` and are not dropped after low-occurrence / Blockaid filtering.
> 
> This is a **breaking public API change**: `CustomAssetGraduationMiddleware` and `CustomAssetGraduationMiddlewareOptions` are removed from package exports. **`removeCustomAsset` remains** the only way to clear a custom import. Pipeline order is now balances → RPC fallback → detection → enrichment (graduation step removed).
> 
> Tests are updated from expecting graduation to **retention**, with new BNB Chain (CDOGE) integration coverage for spam tokens **imported as custom assets**, plus fixture support to omit balances from mocked Accounts API responses and an RPC fallback test when the API already reports the custom asset balance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 664e582590c61681ae47af87151ebe29dec05321. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/assets-controller/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/transaction-controller` from `^70.0.1` to `^70.1.0` ([#10262](https://github.com/MetaMask/core/pull/10262))
 
+### Removed
+
+- **BREAKING:** Remove `CustomAssetGraduationMiddleware` (and its fast-lane source role), which removed an EVM asset from `customAssets` as soon as the Accounts API / Account Activity websocket reported a balance for it. Custom assets are now never removed from `customAssets` automatically, even once upstream sources start reporting them, so they keep their exemption from occurrence / Blockaid spam filtering and can no longer be auto-filtered out of balances and metadata after "graduating" ([#10335](https://github.com/MetaMask/core/pull/10335))
```
Prithpal-Sooriya: [bot] Shorten changlog pls.

packages/assets-controller/src/middlewares/CustomAssetGraduationMiddleware.ts
Prithpal-Sooriya, line 1: [bot] Reply to this comment with your safety & judgement score (out of 100) on how confident you feel on removing this.

Lemme know the code paths and reasons why this isn't necessary.
- Do we ensure we're not making extra additional calls?
- Do we make sure custom assets that custom assets that return by AccountsAPI don't need to make the RPC fallback calls?
- Do we make sure that custom assets not returned by AccountsAPI do need to make the RPC fallback calls?

Can we create some unit tests and integration tests for more confidence?

packages/assets-controller/src/pipeline/buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -239,3 +230,28 @@ describe('assets pipeline: BNB Chain spam token (CDOGE)', () => {
...
+  afterEach(() => {
+    cleanAll();
+  });
+
+  // Graduation scenario: the user imported CDOGE by hand before the Accounts
+  // API indexed it. The API now reports a positive balance for it, while the
+  // Tokens API still carries only 1 occurrence — below the default floor of
+  // 3 for BNB Chain. The token must survive the fast lane: custom assets are
+  // exempt from occurrence filtering and must never be removed from
+  // `customAssets` automatically, or the user loses the token they imported.
```
Prithpal-Sooriya: [bot] you know my take on comments - we shouldn't need them

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -21,14 +21,17 @@ import { createTestApiClient } from './__fixtures__/mockTokenApi.js';
...
 /**
  * Integration coverage for `AssetsController` against the BNB Chain wallet
  * from the `$$$DOGECHAIN` (`CDOGE`) spam-token report.
  *
  * Boots the real controller, answers the same captured APIs as
  * `buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts`, and
- * asserts CDOGE never lands in persisted state.
+ * asserts CDOGE never lands in persisted state — unless the user imported it
+ * as a custom asset, in which case it must survive (see the custom-asset
+ * suite below).
```
Prithpal-Sooriya, line 34: [bot] Remove this comment, makes the test spec a little confusing.

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -111,6 +114,32 @@ async function fetchWallet(
   });
 }
 
+/**
+ * The wallet state `addCustomAsset` leaves behind when the user imports
+ * CDOGE by hand: the checksummed asset ID registered under their account, a
+ * seeded zero balance and stub metadata so the token renders immediately.
+ *
+ * @returns The starting state for the custom-asset fetch scenario.
+ */
```
Prithpal-Sooriya: [bot] You know my take on comments please :)

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -168,3 +197,28 @@ describe('AssetsController: BNB Chain spam token (CDOGE)', () => {
...
+  afterEach(() => {
+    cleanAll();
+  });
+
+  // Graduation scenario: the user imported CDOGE by hand before the Accounts
+  // API indexed it. The API now reports a positive balance for the token
+  // while the Tokens API still carries only 1 occurrence — below the default
+  // floor of 3 for BNB Chain. The import must survive the fetch: custom
+  // assets are exempt from occurrence filtering and are never removed from
+  // `customAssets` automatically, or the user loses the token they imported.
```
Prithpal-Sooriya: [bot] comment cleanup/removal pls

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
```diff
@@ -168,3 +197,28 @@ describe('AssetsController: BNB Chain spam token (CDOGE)', () => {
...
+  it('keeps the imported token in customAssets, balances and metadata', async () => {
+    const state = await fetchWallet(buildCustomAssetWalletState());
```
Prithpal-Sooriya, line 198: [bot] Since this `buildCustomAssetWalletState` is only used here, lets please scope this function/util inside this describe block.

packages/assets-controller/src/AssetsController.ts
```diff
@@ -3810,26 +3792,17 @@ export class AssetsController extends BaseController<
...
-        // those.
-        const shouldGraduateCustomAssets =
-          sourceId === 'AccountsApiDataSource' ||
-          sourceId === 'AccountActivityDataSource';
-
         // Websocket updates can carry brand-new spam airdrops: enrich them
         // with Token API occurrences and drop below-floor tokens BEFORE
         // detection, so spam is never detected, enriched, priced or persisted.
+        // Custom assets are exempt — see the candidate scan in
+        // `TokenDataSource.occurrenceFilterMiddleware`.
```
Prithpal-Sooriya: [bot] remove comment

packages/assets-controller/CHANGELOG.md
Prithpal-Sooriya: Done — shortened to one sentence in 664e582590.

packages/assets-controller/src/pipeline/buildFastFetchSources.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya: Done — removed the narrating block in 664e582590.

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya: Done — dropped the expectation line from the header in 664e582590.

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya: Done — removed the docblock in 664e582590.

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya: Done — removed the narrating block in 664e582590.

packages/assets-controller/src/AssetsController.bsc-spam-token-filtering.integration.test.ts
Prithpal-Sooriya: Done — moved `buildCustomAssetWalletState` inside the custom-asset describe block in 664e582590.

packages/assets-controller/src/AssetsController.ts
Prithpal-Sooriya: Done — removed the comment block in 664e582590.

packages/assets-controller/src/middlewares/CustomAssetGraduationMiddleware.ts
Prithpal-Sooriya: Actioned in 664e582590. Answers to your questions:

**Why delete rather than neuter?** The middleware's only job was removing EVM assets from `customAssets` once upstream sources started reporting them. Its premise that "the regular detection flow keeps it fresh" is exactly the bug: graduation subjects the token to occurrence/Blockaid spam filtering, which a user import must be exempt from. It had a single caller (fast-lane wiring) and no other behavior to preserve, so deleting it is the minimal change that guarantees custom assets are never auto-removed. Explicit removal stays available via `AssetsController:removeCustomAsset`.

**Extra calls per fetch?** None. `RpcFallbackMiddleware.collectStaleTrackedAssets` already unions `customAssets` with `assetsBalance` keys when computing stale assets, so per-fetch call counts are unchanged by this removal. The only cost is that subscription-level `pickRpcCustomAssetsSupplement` polling persists for graduated tokens — the same polling the token had before graduating, deduped against `rpcAssignedChains`.

**RPC fallback behavior, both cases (now pinned by tests):**
- Custom asset the Accounts API reports with a positive balance → never stale → **no RPC call**.
- Custom asset the Accounts API omits (or reports zero) → **one RPC re-read**; the asset is passed via the request's `customAssets` field for multicall, alongside state-tracked assets.

**Tests added in 664e582590:**
- Unit (`RpcFallbackMiddleware.test.ts`): "skips custom assets the response already reports a balance for" — pairs with the existing "fetches custom assets from state that the response left empty" case.
- Integration (fast-lane BSC suite; real AccountsApi/StakedBalance/RpcFallbackMiddleware/DetectionMiddleware/TokenDataSource/PriceDataSource with a recorded RPC source): "does not re-read the custom asset on RPC when the Accounts API reports its balance" (asserts zero RPC calls) and "re-reads the custom asset on RPC when the Accounts API omits it" (asserts one RPC call, the asset ID in the request's `customAssets`, and the merged balance in the final response).

Full JSON-RPC e2e isn't feasible in this suite — `RpcDataSourceConfig` has no endpoint injection — so the recorded-source seam is the closest equivalent.

Safety self-assessment: 94/100 — single-caller removal with no remaining behavior to preserve, no added network calls, both fallback branches pinned by unit + integration tests; residual risk is the persistent subscription-level polling for graduated tokens noted above.

packages/assets-controller/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump `@metamask/transaction-controller` from `^70.0.1` to `^70.1.0` ([#10262](https://github.com/MetaMask/core/pull/10262))
 
+### Removed
+
+- **BREAKING:** Remove `CustomAssetGraduationMiddleware` and its exports, so custom assets are never removed from `customAssets` automatically and keep their spam-filtering exemption instead of being auto-filtered out once upstream sources start reporting them ([#10335](https://github.com/MetaMask/core/pull/10335))
```
Prithpal-Sooriya, line 16: This is an internal breaking change, not exposed interfaces have changed.

Prithpal-Sooriya: I kept the `**BREAKING:**` prefix, because this PR does remove public exports — the evidence:

- `src/index.ts` is the package entrypoint (`package.json` exports `.` → `dist/index.js`, built from `src/index.ts`), and on `main` it re-exported `CustomAssetGraduationMiddleware` and `CustomAssetGraduationMiddlewareOptions`. This PR removes both from that entrypoint.
- Those exports have shipped to npm since `6.2.0` ([#8582](https://github.com/MetaMask/core/pull/8582)): `@metamask/assets-controller@16.1.0/dist/index.d.ts` still exports them today, so any consumer importing either symbol breaks on this upgrade.
- Per the repo's changelog guidelines, "a change is breaking if it removes … any public export", and the `**BREAKING:**` prefix is what makes `create-release-branch` cut `17.0.0` instead of `16.2.0`. Dropping the prefix would ship the removal under a minor that `^16.x` consumers auto-pull.

The behavior change also stands on its own under the same guideline ("changes default behavior that consumers rely on"): auto-removal of custom assets on graduation has been the documented default since `6.2.0`.

That said — if the team considers all consumers first-party and wants this treated as non-breaking anyway, say the word and I'll drop the prefix and adjust the entry.

